PR #843, opened 2019-11-05
Enable publish to maven central via Sonatype

jayv:
__Motivation__

Directly publishing to sonatype maven central gives us more control over the artifacts to be promoted to central.

__Modifications__

- add POM metadata as required by maven central
- add signing of artifacts as per maven central requirements
- update Docker Compose to pass PGP and Sonatype credentials from the ENV

__Result__

Gradle can now publish to maven central.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -71,9 +71,17 @@ services:
       - CI
       - BINTRAY_USER
       - BINTRAY_KEY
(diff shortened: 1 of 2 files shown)


idelpivnitskiy:
@jayv Can we close this in favor of #849?

jayv:
Yes closing, altho both need work

Issue #828, opened 2019-10-30
ProtocolCompatibilityTest.serviceTalkToServiceTalkErrorViaServiceFilter test failure


Scottmitch:
failed while building https://github.com/apple/servicetalk/pull/827

ci link: https://ci.servicetalk.io/job/servicetalk-java11-prb/801/testReport/junit/io.servicetalk.grpc.netty/ProtocolCompatibilityTest/serviceTalkToServiceTalkErrorViaServiceFilter/

```
Error Message
org.junit.experimental.theories.internal.ParameterizedAssertionError: serviceTalkToServiceTalkErrorViaServiceFilter("false" <from ssl[0]>, "true" <from streaming[1]>)
Stacktrace

org.junit.experimental.theories.internal.ParameterizedAssertionError: serviceTalkToServiceTalkErrorViaServiceFilter("false" <from ssl[0]>, "true" <from streaming[1]>)
	at org.junit.experimental.theories.Theories$TheoryAnchor.reportParameterizedError(Theories.java:288)
	at org.junit.experimental.theories.Theories$TheoryAnchor$1$1.evaluate(Theories.java:237)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithCompleteAssignment(Theories.java:218)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithAssignment(Theories.java:204)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithIncompleteAssignment(Theories.java:212)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithAssignment(Theories.java:202)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithIncompleteAssignment(Theories.java:212)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithAssignment(Theories.java:202)
	at org.junit.experimental.theories.Theories$TheoryAnchor.evaluate(Theories.java:187)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
[...]

PR #850, opened 2019-11-07
Support for Single.fromCallable(), Single.fromSupplier(), Completable.fromRunnable()

blake-bauman:
Motivation:

Currently, if we want to take a piece of non-asynchronous code and create a Single out of it for a response, we’d have to do one of the following:

* Use CompletableFuture or a Mono to encapsulate the code, and then use Single.fromStage()
* Use Single.defer(() -> Single.succeeded(…))

While this works, it’s an extra step that can be provided as a shortcut in the Single class itself.  Both Mono (Project Reactor) and CompletableFuture offer ways of creating deferred executions via a Supplier, Callable, or Runnable.  However, to use the streaming server it has to be of type Single.  It makes sense for Single to have these capabilities directly.

Modifications:

- Added fromCallable() and fromSupplier() to Single
- Added fromRunnable() to Completable
- Added related tests

Result:

Increased simplicity


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CallableSingle.java
```diff
@@ -0,0 +1,54 @@
...
+
+import static io.servicetalk.concurrent.Cancellable.IGNORE_CANCEL;
+import static io.servicetalk.concurrent.internal.SubscriberUtils.handleExceptionFromOnSubscribe;
+
+final class CallableSingle<T> extends AbstractSynchronousSingle<T> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SucceededSingle.class);
+    private final Callable<T> callable;
+
+    CallableSingle(final Callable<T> callable) {
+        this.callable = callable;
```
normanmaurer: missing `Objects.requireNotNull(...)` check

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1187,6 +1187,15 @@ public static Completable completed() {
         return CompletedCompletable.INSTANCE;
     }
 
+    /**
+     * Creates a realized completed {@code Completable} which first runs the supplied {@code Runnable}.
+     *
+     * @return A new {@code Completable}.
+     */
+    public static Completable fromRunnable(final Runnable runnable) {
+        return new RunnableCompletable(runnable);
```
normanmaurer: missing `Objects.requireNotNull(...)` check

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RunnableCompletable.java
```diff
@@ -0,0 +1,53 @@
...
+
+import static io.servicetalk.concurrent.Cancellable.IGNORE_CANCEL;
+import static io.servicetalk.concurrent.internal.SubscriberUtils.handleExceptionFromOnSubscribe;
+
+public class RunnableCompletable extends AbstractSynchronousCompletable {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SucceededSingle.class);
+    private final Runnable runnable;
+
+    RunnableCompletable(final Runnable runnable) {
+        this.runnable = runnable;
```
normanmaurer: missing `Objects.requireNotNull(...)` check

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SupplierSingle.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static io.servicetalk.concurrent.Cancellable.IGNORE_CANCEL;
+import static io.servicetalk.concurrent.internal.SubscriberUtils.handleExceptionFromOnSubscribe;
+
+final class SupplierSingle<T> extends AbstractSynchronousSingle<T> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SucceededSingle.class);
+    private final Supplier<T> supplier;
+
+    SupplierSingle(final Supplier<T> supplier) {
+        this.supplier = supplier;
```
normanmaurer: missing `Objects.requireNotNull(...)` check

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SupplierSingle.java
```diff
@@ -0,0 +1,57 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.function.Supplier;
+
+import static io.servicetalk.concurrent.Cancellable.IGNORE_CANCEL;
+import static io.servicetalk.concurrent.internal.SubscriberUtils.handleExceptionFromOnSubscribe;
+
+final class SupplierSingle<T> extends AbstractSynchronousSingle<T> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SucceededSingle.class);
```
normanmaurer: this should be `SupplierSingle.class`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RunnableCompletable.java
```diff
@@ -0,0 +1,53 @@
...
+package io.servicetalk.concurrent.api;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static io.servicetalk.concurrent.Cancellable.IGNORE_CANCEL;
+import static io.servicetalk.concurrent.internal.SubscriberUtils.handleExceptionFromOnSubscribe;
+
+public class RunnableCompletable extends AbstractSynchronousCompletable {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SucceededSingle.class);
```
normanmaurer: This should be `RunnableCompletable.class`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CallableSingle.java
```diff
@@ -0,0 +1,54 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.Callable;
+
+import static io.servicetalk.concurrent.Cancellable.IGNORE_CANCEL;
+import static io.servicetalk.concurrent.internal.SubscriberUtils.handleExceptionFromOnSubscribe;
+
+final class CallableSingle<T> extends AbstractSynchronousSingle<T> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SucceededSingle.class);
```
normanmaurer: This should be `CallableSingle.class`

blake-bauman: my bad

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CallableSingle.java
blake-bauman: You mean `Objects.requireNonNull(...)`?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
blake-bauman: Wouldn't this be redundant with the check in `RunnableCompletable` itself?

normanmaurer: Yes sorry...

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CallableSingle.java
normanmaurer: yep :)

blake-bauman:
Updated.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CallableSingle.java
normanmaurer: nit: you could merge both lines as `requireNonNull(...)` returns `callable`. Same comment for all the other usages.

blake-bauman: oops.  yeah.  it's late here.  :-)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1087,6 +1088,32 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
...
+     * Creates a realized {@link Single} which always completes successfully with the result of the provided
+     * {@code supplier}.
+     *
+     * @param supplier {@link Supplier} which supplies the result of the {@link Single}.
+     * @param <T>      Type of the {@link Single}.
+     *
+     * @return A new {@link Single}.
+     */
+    public static <T> Single<T> fromSupplier(final Supplier<T> supplier) {
+        return new SupplierSingle<>(supplier);
```
NiteshKant: Do we need both `fromCallable` and `fromSupplier` methods? It seems like `fromCallable` should be sufficient (since `Callable#call()` accomodates for throwing an `Exception`) and if users have a `Supplier` they can use `fromCallable(supplier::get)`.

In case you think both methods are important to provide, we should reduce implementations by delegating to `fromCallble()` from here, like:

```java
public static <T> Single<T> fromSupplier(final Supplier<T> supplier) {
    return fromCallable(supplier::get);
}
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CallableSingle.java
```diff
@@ -0,0 +1,55 @@
...
+
+import static io.servicetalk.concurrent.Cancellable.IGNORE_CANCEL;
+import static io.servicetalk.concurrent.internal.SubscriberUtils.handleExceptionFromOnSubscribe;
+
+final class CallableSingle<T> extends AbstractSynchronousSingle<T> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(CallableSingle.class);
+    private final Callable<T> callable;
+
+    CallableSingle(final Callable<T> callable) {
+        this.callable = Objects.requireNonNull(callable);
```
NiteshKant: nit: Use static imports for `Objects.requireNonNull()`

NiteshKant:

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1187,6 +1187,15 @@ public static Completable completed() {
         return CompletedCompletable.INSTANCE;
     }
 
+    /**
+     * Creates a realized completed {@code Completable} which first runs the supplied {@code Runnable}.
+     *
```
NiteshKant: Add an `@param` for the arguments and follow the guidance on the javadocs from `Single` above

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CallableSingle.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    void doSubscribe(final Subscriber<? super T> subscriber) {
+        try {
+            subscriber.onSubscribe(IGNORE_CANCEL);
```
NiteshKant: Since the call to `Callable` may block, we should make sure that the calling thread is interrupted if the `Cancellable` is cancelled. We have a utility `ThreadInterruptingCancellable`[1] to handle such scenarios and you can refer to `BlockingUtils#blockingToSingle()`[2] on how to use it. Let me know if you need help on this.

Once you do this change, please also add a unit test to validate.

[1] https://github.com/apple/servicetalk/blob/master/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/ThreadInterruptingCancellable.java#L31
[2] https://github.com/apple/servicetalk/blob/master/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingUtils.java#L68

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RunnableCompletable.java
```diff
@@ -0,0 +1,55 @@
...
+    private final Runnable runnable;
+
+    RunnableCompletable(final Runnable runnable) {
+        this.runnable = Objects.requireNonNull(runnable);
+    }
+
+    @Override
+    void doSubscribe(final Subscriber subscriber) {
+        try {
+            subscriber.onSubscribe(IGNORE_CANCEL);
```
NiteshKant: Same comment regarding using `ThreadInterruptingCancellable` as above for `CallableSingle`

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SupplierSingleTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+import static io.servicetalk.concurrent.api.SourceAdapters.toSource;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+
+public class SupplierSingleTest {
```
NiteshKant: For every new source we create, we also write reactive streams TCK test. Please add those for the new implementations => `CallableSingle` and `RunnableCompletable`. As a reference see:

https://github.com/apple/servicetalk/blob/master/servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/SingleDeferTckTest.java

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/CallableSingleTest.java
```diff
@@ -0,0 +1,81 @@
...
+    private Callable<Integer> factory;
+
+    @SuppressWarnings("unchecked")
+    @Before
+    public void setUp() throws Exception {
+        factory = mock(Callable.class);
+    }
+
+    @Test
+    public void testEverySubscribeCalls() throws Exception {
```
NiteshKant: Can you also add a test to verify the case when `Subscriber#onSubscribe()` throws an `Exception`?

jayv:
Do we want a TCK test @NiteshKant ?

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SupplierSingleTest.java
blake-bauman: will do

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
NiteshKant: Sounds good!

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/CallableSingleTest.java
NiteshKant: You can use the following:

```java
@Test
public void onSubscribeThrows() {
    final Single<Integer> source = Single.fromCallable(factory);
    final SingleSource.Subscriber<Integer> subscriber = mock(SingleSource.Subscriber.class);
    doThrow(DELIBERATE_EXCEPTION).when(subscriber).onSubscribe(any());
    toSource(source).subscribe(subscriber);
    verify(subscriber).onSubscribe(any());
    verify(subscriber).onError(DELIBERATE_EXCEPTION);
    verifyNoMoreInteractions(subscriber);
}
```

Also you can create the `Subscriber` mock in the constructor just like the factory.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CallableSingle.java
NiteshKant: The test is a bit specific to this case, so I wrote a sample for you:

```java
@Test
public void cancelInterrupts() throws Exception {
    final Single<Integer> source = Single.fromCallable(factory);
    CountDownLatch latch = new CountDownLatch(1);
    when(factory.call()).then(invocation -> {
        try {
            // await till interrupted.
            latch.await();
            latch.countDown(); // signal that we got interrupted for the test to proceed
        }
        return 1;
    });
    toSource(source).subscribe(new SingleSource.Subscriber<Integer>() {
        @Override
        public void onSubscribe(final Cancellable cancellable) {
            cancellable.cancel();
        }

        @Override
        public void onSuccess(@Nullable final Integer result) {
            // noop
        }

        @Override
            // noop
        }
    });
    latch.await();
}
```

The complexity here is that `subscribe()` will not return as `Callable#call()` blocks but we need access to the `Cancellable` to cancel. Let me know if you need clarification on the code.

blake-bauman:
Updated.  Thanks for the suggestions.

NiteshKant:

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1187,6 +1187,24 @@ public static Completable completed() {
         return CompletedCompletable.INSTANCE;
     }
 
+    /**
+     * <p>
+     * Creates a {@link Single} which when subscribed will invoke {@link Runnable#run()} on the passed
```
NiteshKant: `{@link Single}` => `{@link Completable}` in this entire doc.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1187,6 +1187,24 @@ public static Completable completed() {
         return CompletedCompletable.INSTANCE;
     }
 
+    /**
+     * <p>
```
NiteshKant: We generally just use a `<p/>` to indicate a new paragraph, so you do not have to wrap the paragraph between `<p>...</p>`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1087,6 +1088,32 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
         return new SucceededSingle<>(value);
     }
 
+    /**
+     * Creates a realized {@link Single} which always completes successfully with the result of the provided
```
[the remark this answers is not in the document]
blake-bauman: QQ.  Single.defer() has the same situation as Single.fromCallable() and Single.fromSupplier() in that there can be blocking calls or an exception thrown during Supplier.get().  Yet the Javadoc for defer() doesn't include this text, the implementation isn't using ThreadInterruptingCancellable, and the unit tests don't test for an interrupt.  What am I missing in the SingleDefer implementation?

NiteshKant:

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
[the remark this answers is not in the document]
NiteshKant: Good question!

There is a subtle difference between `defer()` and `fromCallable()`/`fromSupplier()`:

The provided entity in `defer()` is __providing the source__ (potentially new instance) for each subscribe call. It is typically use to convert eager/hot sources to lazy/cold. However, the provided entity in `fromCallable()`/`fromSupplier()` is __implementing the source__. We expect implementation of sources to be potentially blocking whereas the act of providing a new source is less likely to be blocking.

From an implementation point of view, `defer()` does not interact with the `Subscriber`  so there is not much it is expected to do as compared to `fromCallable()`/`fromSupplier()` which interact with the `Subscriber`. 

Hope this clarifies.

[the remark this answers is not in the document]
blake-bauman: It does, from a theoretical standpoint.  From a practical standpoint, while this would be the intended use of defer() going forward, and since Single was missing fromCallable() and fromSupplier() until now, I've seen the following code:

```
   Single.defer(() -> {
       final SomeObj result = << make blocking call >>
       return Single.succeeded(result);
   });
```

Going forward, you'd want to do that in Single.fromCallable() or fromSupplier(), of course, but I can't help but think there's a bunch of code out there already doing the above.

NiteshKant:
Build failure attributed to know flaky/unstable gRPC test: #828 

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1187,6 +1187,23 @@ public static Completable completed() {
         return CompletedCompletable.INSTANCE;
     }
 
+    /**
+     * Creates a {@link Completable} which when subscribed will invoke {@link Runnable#run()} on the passed
+     * {@link Runnable} and emit the value returned by that invocation from the returned {@link Completable}. Any error
+     * emitted by the {@link Runnable} will terminate the returned {@link Completable} with the same error.
+     * <p />
```
NiteshKant:
```suggestion
     * <p>
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1087,6 +1088,42 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
         return new SucceededSingle<>(value);
     }
 
+    /**
+     * Creates a {@link Single} which when subscribed will invoke {@link Callable#call()} on the passed
+     * {@link Callable} and emit the value returned by that invocation from the returned {@link Single}. Any error
+     * emitted by the {@link Callable} will terminate the returned {@link Single} with the same error.
+     * <p />
```
NiteshKant:
```suggestion
     * <p>
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1087,6 +1088,42 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
...
+     */
+    public static <T> Single<T> fromCallable(final Callable<T> callable) {
+        return new CallableSingle<>(callable);
+    }
+
+    /**
+     * Creates a {@link Single} which when subscribed will invoke {@link Supplier#get()} on the passed
+     * {@link Supplier} and emit the value returned by that invocation from the returned {@link Single}. Any error
+     * emitted by the {@link Supplier} will terminate the returned {@link Single} with the same error.
+     * <p />
```
NiteshKant:
```suggestion
     * <p>
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
blake-bauman: Ok.  I thought it was strange you asked me to use `<p/>`, but I went with it.  :-)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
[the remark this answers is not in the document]
NiteshKant: Hmm ok. I think that usage is "unexpected" from a design point of view but I hear what you are saying. It may be a good idea to add the "blocking details" to `defer()` javadoc also. Do you want to do that?

To clarify; what we did for these newly added methods was just to make sure that if cancellation is done from a different thread it can interrupt the thread running `Callable#call()`. From the point of view of protecting from blocking code, we still have to offload the subscribe call, which is expected out of the users.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
NiteshKant: woops, ya sorry about that

NiteshKant:

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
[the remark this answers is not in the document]
blake-bauman: I'm fine leaving the Javadoc the same if you're not concerned.

PR #859, opened 2019-11-09
BOM is not publishing artifacts

jayv:
__Motivation__

We need to publish BOM

__Modifications__

Duplicate publish settings for BOM which diverges from library tasks.

__Result__

BOM will be published.


jayv:

jayv:

PR #861, opened 2019-11-11
Introduce automatic retries

NiteshKant:
__Motivation__

There are a few expected scenarios where ServiceTalk may use a connection for making request when the connection is already closed. This happens because there is an inherent race between a peer closing a connection and the load balancer choosing a connection. Due to this race, load balancer may end up picking a connection which is already closed but the close notification has not reached the load balancer. In such scenarios, when load balancer chooses an unusable connection writing a request will fail and the users are expected to retry these exceptions. Since in these cases we will never write the request on the wire, these requests are always safe to retry automatically.

__Modification__

We already do a form of automatic retry which is when the load balancer does not have any usable hosts. Since, this new addition is just another form of automatic retry, instead of adding a top-level builder config, now introduced an abstraction called `AutomaticRetryStrategyProvider` which provides a retry strategy for automatic retries. The client builders then can disable or modify this strategy if required.

__Result__

ServiceTalk now automatically retry expected exceptions which are safe to retry.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AutomaticRetryFilter.java
```diff
@@ -31,55 +27,31 @@
...
 
 /**
  * A {@link StreamingHttpClient} filter that will account for transient failures introduced by a {@link LoadBalancer}
  * not being ready for {@link #request(HttpExecutionStrategy, StreamingHttpRequest)} and retry/delay requests until the
  * {@link LoadBalancer} is ready.
  */
-final class LoadBalancerReadyStreamingHttpClientFilter extends StreamingHttpClientFilter {
-    private final LoadBalancerReadySubscriber loadBalancerReadySubscriber;
-    private final int maxRetryCount;
+final class AutomaticRetryFilter extends StreamingHttpClientFilter {
```
idelpivnitskiy: It looks very similar to `RetryingHttpRequesterFilter` now. And `RetryingHttpRequesterFilter` already handles `RetryableConnectException` by default.
It's harder for users to understand the difference between using `DefaultAutomaticRetryStrategyProvider.Builder` and `RetryingHttpRequesterFilter.Builder`. And it's harder to discover that `RetryingHttpRequesterFilter` exists when `AutomaticRetryStrategyProvider` exposed on the builder.

I see that `AutomaticRetryStrategyProvider` needs access to `LoadBalancer`. Could we explore an opportunity to merge it with `RetryingHttpRequesterFilter`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -93,7 +96,8 @@
     private StreamingHttpConnectionFilterFactory connectionFilterFactory;
     @Nullable
     private StreamingHttpClientFilterFactory clientFilterFactory;
-    private boolean lbReadyFilterEnabled = true;
+    @Nullable
+    private AutomaticRetryStrategyProvider autoRetry = new Builder().build();
```
jayv: Suggestion: call everything `Auto*` instead of `Automatic` ?

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerReadyEvent.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018-2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
```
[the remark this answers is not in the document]
NiteshKant: hmmm ... looks like this happened because I moved the class to a different package 😢 
will fix, good catch!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -405,8 +409,9 @@ private AbsoluteAddressHttpRequesterFilter proxyAbsoluteAddressFilterFactory() {
     }
 
     @Override
-    public DefaultSingleAddressHttpClientBuilder<U, R> disableWaitForLoadBalancer() {
-        lbReadyFilterEnabled = false;
+    public SingleAddressHttpClientBuilder<U, R> automaticRetryStrategy(
+            final AutomaticRetryStrategyProvider automaticRetryStrategyProvider) {
+        autoRetry = automaticRetryStrategyProvider == DISABLE_AUTO_RETRIES ? null : autoRetry;
```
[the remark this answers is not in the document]
NiteshKant: ugghh ... good catch

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AutomaticRetryFilter.java
NiteshKant: I agree that they are similar in terms of the fact that they are doing retries, however, the type of retries they do is very different:

- _auto-retry_: This is for expected errors in regular paths when due to races in closing and selecting a connection, we select a bad connection. Such errors are always retryable and since they do not write anything on the network, there is no need to add backoff, jitter, etc.
- _user-retries_: This is for failures which happen during client processing or from the server. Since there is work done on the network, one needs to introduce jitters, backoff, etc.

From a functionality wise, auto-retry also needs the `LoadBalancer` as you noticed. So, I believe they are different enough to be configured differently.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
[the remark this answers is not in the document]
NiteshKant: I also added a unit test to verify that the update is effective 🙏

NiteshKant:
It seems like the newly added test is flaky, will look into it

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AutomaticRetryFilter.java
idelpivnitskiy: Agreed, and as discussed offline, it's also better to have auto-retries closer to transport, when user-retries may handle more logic from other filters.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/AutoRetryStrategyProvider.java
```diff
@@ -0,0 +1,54 @@
...
+
+import io.servicetalk.concurrent.api.AsyncCloseable;
+import io.servicetalk.concurrent.api.BiIntFunction;
+import io.servicetalk.concurrent.api.Completable;
+
+import static io.servicetalk.concurrent.api.Completable.completed;
+import static io.servicetalk.concurrent.api.Completable.failed;
+
+/**
+ * A provider for {@link AutomaticRetryStrategy}.
```
idelpivnitskiy: `AutomaticRetryStrategy` -> `AutoRetryStrategy` ?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/SingleAddressGrpcClientBuilder.java
```diff
@@ -95,6 +96,14 @@
      */
     GrpcClientSecurityConfigurator<U, R> secure();
 
+    /**
+     * Updates the automatic retry strategy for the clients generated by this builder.
```
idelpivnitskiy: Consider provide more motivation/context for this type of retries. We can clarify a few more things in javadoc about these auro retries:
1. designed for framework internals that happen before a request touches network;
2. will be handled as a first filter => other users filters won't see failures that were successfully retried;
3. are protocol-agnostic and do not have request meta;

Should we clarify that other retries should be handled by custom `StreamingHttpClientFilter`s?

NiteshKant: Good idea, lemme add some information without leaking details about the implementation.

NiteshKant:
FYI: the tests were flaky because they were counting retries based on connect request on the connection factory. However, we also retry when there are no hosts available for the load balancer. This led to mismatch in retries reported and retries expected.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuilder.java
```diff
@@ -156,11 +159,18 @@
             Function<U, CharSequence> unresolvedAddressToHostFunction);
 
     /**
-     * Disables automatically delaying {@link StreamingHttpRequest}s until the {@link LoadBalancer} is ready.
+     * Updates the automatic retry strategy for the clients generated by this builder. Automatic retries are done by
+     * the clients automatically when allowed by the passed {@link AutoRetryStrategyProvider}. These retries are not a
+     * substitute for user level retries which are designed to infer retry decisions based on request/error information.
+     * Typically such user level retries are done using filters (eg:
```
jayv: It's good that we're trying to provide more context, but I feel like the description is too vague in what an auto-retry could/should be used for. It's heavily tied to load-balancing from an API perspective already, maybe we should just call it out.  ?

NiteshKant: I hear what you are saying but there is a balance about prescribing implementation details at this level or staying neutral to have the strategy define the behaviours. I thought this language was somehow walking that line. 

The API requires a `LoadBalancer` as the LB events are not otherwise accessible. However, it is not limited to load balancer as we are retrying transport errors.  

Do you think you can suggest an alternative language? I will be happy to update.

jayv: We need to address the underlying problem which is the real implementation quirk users shouldn't have to know about. It's not clear to me why we'd want to disable/configure this and LB-ready. More so it's surprising to users that we don't do this behind the scenes as it's not under their control and all other libraries do this.  

Maybe you envisioned this API enabling other features in the future, but I'm not yet convinced this kind of abstraction is helpful for a user. At the very least it should be clear what _kind_ of behaviors to expect or not-expect should people want to implement their own. We traded an implementation detail for vagueness in this PR and neither is ideal. If we're not sure or not comfortable with describing the intended uses, maybe reconsider introducing as public API.  

As far as wording, I'd consider something like:
> The Client consults this strategy in case a request couldn't be made, for ex. due to state of the LoadBalancer or its Connection(s). It may allow the Client to automatically retry requests that failed when it is guaranteed safe to do so according to the strategy. Not all failures caused by an issue with the Connection are safe to retry without side-effects and therefore application-level error handling is recommended.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DefaultAutoRetryStrategyProvider.java
```diff
@@ -0,0 +1,159 @@
...
+
+        @Override
+        public Completable apply(final int count, final Throwable cause) {
+            if (count > maxRetryCount) {
+                return failed(cause);
+            }
+            if (loadBalancerReadySubscriber != null && cause instanceof NoAvailableHostException) {
+                return loadBalancerReadySubscriber.onHostsAvailable();
+            }
+            if (retryAllRetryableExceptions && cause instanceof RetryableException) {
```
idelpivnitskiy: Should we retry all `RetryableException`s by default? It will also retry `io.servicetalk.client.api.ConnectTimeoutException` that may cause very high latency.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuilder.java
idelpivnitskiy: I also have mixed fillings about this API. It's nice that users can control and disable these retries, but do we know the use case when it's necessary? Can we defer this level of control until we know why these retries cause a problem? And until that just do it internally. It looks like for 99.9% of cases users will be happy if ST solves these type of problems automatically.

I'm agreed that we may want to increase level of visibility for these type of exceptions and notify users about what is happening under the hood. It may help them debug server-side or service discovery problems. But for disabling these retries there are other ways: they can always wrap LB interface and this wrapper may wrap specific exceptions to hide the `RetryableException` marker. Or, if `NoAvailableHostException` causes undesired high latency for the first request users can apply timeout filter.

NiteshKant: I may not have been clear about this in the description but 

__This is not a problem due to ServiceTalk design, it is a problem due to concurrent actions from users and network__

The intentional vaguenss in the builder methods is to defer the details to the supplied strategy. `DefaultAutoRetryStrategyProvider` gives relevant details about the behaviors, eg:

```java
        /**
         * Connection closures (by the peer or locally) and new requests may happen concurrently. This means that it is
         * possible for a {@link LoadBalancer} to select a connection which is already closed (concurrently) but the
         * close signal has not yet been seen the {@link LoadBalancer}. In such cases, requests fail with a
         * {@link RetryableException}. By default, automatic retries always retries these {@link RetryableException}s.
         * This method disables the default behavior.
         *
         * @return {@code this}.
         */
        public Builder disableRetryAllRetryableExceptions() {
```

So I would not say that "we're not sure or not comfortable with describing the intended uses" but just being less opinionated on a generic API with deferring specifics to specific API.

As a reference grpc-java has similar control knobs in form of [waitForReady](https://github.com/grpc/grpc-java/blob/ce2b0174b2be207cb316faf49e923afe09e0f264/api/src/main/java/io/grpc/CallOptions.java#L156-L160) which we already provided before this PR in the form of `disableWaitForLoadBalancer`. When disabled, the clients display a "fail-fast" behavior which may be required in cases to reject traffic before the system is warmed up (indicated through composing multiple client statuses) as opposed to holding requests in memory.
 
As a principle, I think if we make a decision to provide default for a behavior which isn't otherwise stated in the specification, we should provide a way to disable the behavior when reasonably possible. The motivation to extract a separate interface is to consolidated auto-retry behaviors for different type of failures which seems better than providing two top-level methods that are non-intuitively related.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DefaultAutoRetryStrategyProvider.java
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * A builder for {@link DefaultAutoRetryStrategyProvider}.
+     */
+    public static final class Builder {
+        private boolean waitForLb = true;
+        private boolean retryAllRetryableExceptions = true;
+        private int maxRetries = 4;
+
+        /**
+         * By default, automatic retries waits for the associated {@link LoadBalancer} to be ready before triggering a
```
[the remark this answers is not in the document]
NiteshKant: `waits` looks like correct grammar to me.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DefaultAutoRetryStrategyProvider.java
NiteshKant: Are you assuming that all subsequent connects also will timeout?

For a degenerate case, when all servers fail connect then yes it would cause additional latency but otherwise client-side load balancing is expected to move off such "faulty hosts"

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuilder.java
idelpivnitskiy: After thinking about your comment and discussing offline it makes sense to expose this config. It helps with discoverability of this behavior to control it and gives a way to configure max number of retries that could be useful in envs with slow ServiceDiscovery.

PR #866, opened 2019-11-14
gRPC clients do not provide `ExecutionContext`

NiteshKant:
__Motivation__

HTTP clients provides a way to fetch the `ExecutionContext` with which the client was created. This was missed for gRPC.

__Modification__

- Add `executionContext()` method for `GrpcClient` and `BlockingGrpcClient`.
- In order for the generated clients to provide this information, we need to provide the same in `GrpcClientCallFactory`.
- For generated clients that delegate to filters, in order to provide this information, we need to make it as part of generated filterable clients. Generated filterable clients were not using any interface which made it hard to add this new method. Added a new interface `FilterableGrpcClient` which provides contract for filterable clients.
- Generated filterable clients (and filters) were incorrectly having the blocking `close()` method,which is not carried over to the new `FilterableGrpcClient` interface.

__Result__

Parity between HTTP and gRPC clients.


NiteshKant:
@ddossot can you sanity check the codegen changes?

ddossot:
@NiteshKant looks pretty sane 😉 🚢

PR #868, opened 2019-11-20
Declare character encoding in responses returned from fallback service

sonicloong:
Motivation:

Currently `DefaultFallbackServiceStreaming` returns response without specifying the charset parameter in the `Content-Type` header. This would confuse the client when it needs to parse the payload. Extra response handling logic is needed on client side to either abort parsing or guess the character encoding.

Modifications:

- Set charset to UTF-8 in the `Content-Type` header.

Result:

Character encoding is specified explicitly.

### DIFF
--- a/servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/DefaultFallbackServiceStreaming.java
+++ b/servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/DefaultFallbackServiceStreaming.java
@@ -24,7 +24,7 @@ import io.servicetalk.http.api.StreamingHttpService;
 
 import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_LENGTH;
 import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_TYPE;
-import static io.servicetalk.http.api.HttpHeaderValues.TEXT_PLAIN;
+import static io.servicetalk.http.api.HttpHeaderValues.TEXT_PLAIN_UTF_8;
 import static io.servicetalk.http.api.HttpHeaderValues.ZERO;
 import static io.servicetalk.http.api.HttpResponseStatus.NOT_FOUND;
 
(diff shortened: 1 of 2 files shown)


idelpivnitskiy:
Thanks for the contribution @sonicloong!
Could you please elaborate a little bit more in the motivation section why the existing behavior causes an issue.

@NiteshKant I see that the payload body is empty in this case. Should we omit the `content-type` header? [RFC7231](https://tools.ietf.org/html/rfc7231#section-3.1.1.5) does not require it if the payload body is empty:

> A sender that generates a message containing a payload body SHOULD generate a Content-Type header field in that message unless the intended media type of the enclosed representation is unknown to the sender.

NiteshKant:
@idelpivnitskiy thats a good idea, however, we should make sure that our deserializers can handle `content-length:0` with no `content-type` request/response. It does not look like they do today.

sonicloong:
@idelpivnitskiy sure. I've updated the commit. Please let me know if that makes sense. Omitting the content type header also works for me, and seems more appropriate.

idelpivnitskiy:
The current fix will help only if the client uses default text deserializer:
https://github.com/apple/servicetalk/blob/03a7b0a48eb0d2192ddf2fdfd0ed09d409b5b6c9/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpSerializationProviders.java#L166-L174

However, the issue remains the same if users use `formUrlEncodedDeserializer()` or JSON deserializer.

I think we should address deserializer issues and then omit the `content-type` header for 404. But I don't mind to merge this PR first, wdyt @NiteshKant?

@servicetalk-bot ok to test

NiteshKant:
@idelpivnitskiy 

>> the issue remains the same if users use formUrlEncodedDeserializer() or JSON deserializer.

I may have conflated this discussion by the mention of deserializers. Specifying the charset in the `content-type` is a good practice so this PR should be good to go, let me merge it.

Our deserializers being strict about presence/absence of charset is a different issue that we can handle separately.

PR #874, opened 2019-11-22
Introduce `servicetalk-router-api` module

idelpivnitskiy:
Motivation:

`servicetalk-http-router-jersey` introduces annotations that helps users
to control execution strategies per route. The same functionality will
be useful not only for jersey but for other routing modules as well.

Modifications:

- Create `servicetalk-router-api` module;
- Move `@RouteExecutionStrategy` and `@NoOffloadsRouteExecutionStrategy`
annotations to that module;

Result:

Annotations that control execution strategy could be used in non-jersey
modules.


servicetalk-router-api/src/main/java/io/servicetalk/router/api/NoOffloadsRouteExecutionStrategy.java
```diff
@@ -27,8 +25,8 @@
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 /**
- * A special {@link RouteExecutionStrategy} that indicates that a resource class or method needs to use
- * the {@link HttpExecutionStrategies#noOffloadsStrategy()}.
+ * A special {@link RouteExecutionStrategy} that indicates that a resource class or method needs to use a strategy that
+ * does not offload user code to a worker thread pool.
```
NiteshKant: nit: remove usage of word "worker" => `does not offload user code using a different thread pool`

(here and below)

PR #875, opened 2019-11-22
Fix evaluation order in HeaderUtils.hasContentType

sonicloong:
Motivation:

Currently `HeaderUtils.hasContentType` returns true if headers contain `application/x-www-form-urlencoded; charset=UTF-8`, no matter what the expected charset is. This causes `FormUrlEncodedHttpDeserializer` with UTF-16 charset attempting to deserialize UTF-8 encoded payload.

Modifications:

- Fix evaluation order in `HeaderUtils.hasContentType`

Result:

Method behaves as expected.


idelpivnitskiy:
@servicetalk-bot ok to test

PR #880, opened 2019-11-25
Add http headers values to serialization exception message

### LINKED ISSUE #878: SerializationException from `http-api` does not list `content-type`
When users attempt to deserialize a response that has the wrong `content-type`, they receive an exception that lists all the headers, but with `<filtered>` as the value, making it hard to diagnose.

This may happen if the `content-type` is wrong, but the exception doesn’t include the `content-type`, so it’s hard to debug. This may also happen if the `content-type` is missing, such as often the case with a non-200 response.

We should enhance the message of `SerializationException` in `http-api` module to include the received `content-type` and `content-length` values.

Current message:
```
Unexpected headers, can not deserialize. Headers: DefaultHttpHeaders[content-type: <filtered>
content-length: <filtered>]
```
Desired message (with alternatives):
```
Unexpected headers, can not deserialize. DefaultHttpHeaders[content-type: text/plain
content-length: 12, some-other-header: <filtered>]
```
Header names that should not be filtered for `SerializationException`:
- content-type
- content-length
- transfer-encoding

karim-elngr:
Motivation:

When users attempt to deserialize a response that has the wrong content-type, they receive an exception that lists all the headers, but with `<filtered>` as the value, making it hard to diagnose.

Modifications:

- Add a new debug filter `DEFAULT_DEBUG_HEADER_FILTER` to the class `HeaderUtils`.
- Modify the following classes [`DefaultClassHttpDeserializer`, `DefaultTypeHttpDeserializer`, `FormUrlEncodedHttpDeserializer`, `HttpStringDeserializer`] and change the method `checkContentType` to call `headers.toString(DEFAULT_DEBUG_HEADER_FILTER)`.
- Add tests to cover the new SerializationException messages.

Result:

Better debugging info in `SerializationException` when attempting to deserialize invalid
`content-type`.

Fixes #878

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultClassHttpDeserializer.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultClassHttpDeserializer.java
@@ -64,7 +64,7 @@ final class DefaultClassHttpDeserializer<T> implements HttpDeserializer<T> {
     private void checkContentType(final HttpHeaders headers) {
         if (!checkContentType.test(headers)) {
             throw new SerializationException("Unexpected headers, can not deserialize. Headers: "
-                    + headers.toString());
+                    + headers.toString(HeaderUtils.DEFAULT_DEBUG_HEADER_FILTER));
         }
     }
 }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultClassHttpDeserializer.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultClassHttpDeserializer.java
@@ -23,6 +23,8 @@ import io.servicetalk.serialization.api.Serializer;
 
 import java.util.function.Predicate;
 
+import static io.servicetalk.http.api.HeaderUtils.DEFAULT_DEBUG_HEADER_FILTER;
+
 /**
  * A {@link HttpDeserializer} that can deserialize to a {@link Class} of type {@link T}.
  *
@@ -64,7 +66,7 @@ final class DefaultClassHttpDeserializer<T> implements HttpDeserializer<T> {
     private void checkContentType(final HttpHeaders headers) {
         if (!checkContentType.test(headers)) {
             throw new SerializationException("Unexpected headers, can not deserialize. Headers: "
-                    + headers.toString(HeaderUtils.DEFAULT_DEBUG_HEADER_FILTER));
+                    + headers.toString(DEFAULT_DEBUG_HEADER_FILTER));
         }
     }
 }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
@@ -18,9 +18,13 @@ package io.servicetalk.http.api;
 import io.servicetalk.buffer.api.ByteProcessor;
 
 import java.nio.charset.Charset;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.regex.Pattern;
 import javax.annotation.Nullable;
@@ -60,16 +64,16 @@ public final class HeaderUtils {
     static final int HASH_CODE_SEED = 0xc2b2ae35;
     public static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence> DEFAULT_HEADER_FILTER =
             (k, v) -> "<filtered>";
+    private static final Set<CharSequence> DEFAULT_DEBUG_HEADER_NAMES = Collections.unmodifiableSet(
+            new HashSet<>(Arrays.asList(CONTENT_TYPE, CONTENT_LENGTH, TRANSFER_ENCODING)));
     static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence>
             DEFAULT_DEBUG_HEADER_FILTER = (key, value) -> {
-                switch (key.toString()) {
-                    case "content-type":
-                    case "content-length":
-                    case "transfer-encoding":
+                for (CharSequence headerName : DEFAULT_DEBUG_HEADER_NAMES) {
+                    if (CharSequences.contentEqualsIgnoreCase(key, headerName)) {
                         return value;
-                    default:
-                        return "<filtered>";
+                    }
                 }
+                return "<filtered>";
             };
     private static final ByteProcessor HEADER_NAME_VALIDATOR = value -> {
         validateHeaderNameToken(value);
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
@@ -60,7 +60,7 @@ public final class HeaderUtils {
     static final int HASH_CODE_SEED = 0xc2b2ae35;
     public static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence> DEFAULT_HEADER_FILTER =
             (k, v) -> "<filtered>";
-    public static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence>
+    static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence>
             DEFAULT_DEBUG_HEADER_FILTER = (key, value) -> {
                 switch (key.toString()) {
                     case "content-type":
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
@@ -60,6 +60,17 @@ public final class HeaderUtils {
     static final int HASH_CODE_SEED = 0xc2b2ae35;
     public static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence> DEFAULT_HEADER_FILTER =
             (k, v) -> "<filtered>";
+    public static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence>
+            DEFAULT_DEBUG_HEADER_FILTER = (key, value) -> {
+                switch (key.toString()) {
+                    case "content-type":
+                    case "content-length":
+                    case "transfer-encoding":
+                        return value;
+                    default:
+                        return "<filtered>";
+                }
+            };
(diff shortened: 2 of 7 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultClassHttpDeserializer.java
```diff
@@ -64,7 +64,7 @@ public T deserialize(final HttpHeaders headers, final Buffer payload) {
     private void checkContentType(final HttpHeaders headers) {
         if (!checkContentType.test(headers)) {
             throw new SerializationException("Unexpected headers, can not deserialize. Headers: "
-                    + headers.toString());
+                    + headers.toString(HeaderUtils.DEFAULT_DEBUG_HEADER_FILTER));
```
NiteshKant: nit: please use static import for `DEFAULT_DEBUG_HEADER_FILTER` to improve readability.
(here and other places where you are using this constant.)

NiteshKant:

servicetalk-http-api/src/test/java/io/servicetalk/http/api/FormUrlEncodedHttpDeserializerTest.java
```diff
@@ -84,7 +84,42 @@ public void invalidContentTypeThrows() {
         final HttpHeaders headers = DefaultHttpHeadersFactory.INSTANCE.newHeaders();
         headers.set(CONTENT_TYPE, "invalid/content/type");
 
+        final String expectedMessage = "Unexpected headers, can not deserialize. Headers: DefaultHttpHeaders" +
```
NiteshKant: As we are only interested in making sure the desired content-type is present in the message, you can just check for the presence of that message:

```java
expectedException.expectMessage(containsString("invalid/content/type"));
```

This will limit changing tests if/when we change the exception message.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
```diff
@@ -60,6 +60,17 @@
     static final int HASH_CODE_SEED = 0xc2b2ae35;
     public static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence> DEFAULT_HEADER_FILTER =
             (k, v) -> "<filtered>";
+    public static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence>
```
NiteshKant: Make this package private to reduce API surface area. We generally limit creating new public APIs unless they are absolutely required, the intention is to avoid breaking changes (API/behavior) in the future while giving us the flexibility to modify internally used APIs.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/FormUrlEncodedHttpDeserializerTest.java
karim-elngr: This makes more sense :+1

idelpivnitskiy:

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
```diff
@@ -60,6 +60,17 @@
     static final int HASH_CODE_SEED = 0xc2b2ae35;
     public static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence> DEFAULT_HEADER_FILTER =
             (k, v) -> "<filtered>";
+    static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence>
+            DEFAULT_DEBUG_HEADER_FILTER = (key, value) -> {
+                switch (key.toString()) {
+                    case "content-type":
+                    case "content-length":
+                    case "transfer-encoding":
```
idelpivnitskiy: The header names are case-insensitive, see [FRC7230, section 3.2](https://tools.ietf.org/html/rfc7230#section-3.2).
We have a special utility `CharSequences.contentEqualsIgnoreCase(...)` for comparing header names.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/HeaderUtilsTest.java
```diff
@@ -39,6 +42,17 @@
 import static org.junit.Assert.assertTrue;
 
 public class HeaderUtilsTest {
+    @Test
+    public void DEFAULT_DEBUG_HEADER_FILTER() {
```
idelpivnitskiy: Consider using camel case for method names: `defaultDebugHeaderFilter()`

servicetalk-http-api/src/test/java/io/servicetalk/http/api/HeaderUtilsTest.java
```diff
@@ -39,6 +42,17 @@
...
+        assertEquals(APPLICATION_JSON, HeaderUtils.DEFAULT_DEBUG_HEADER_FILTER.apply(CONTENT_TYPE, APPLICATION_JSON));
+
+        assertEquals("3495", HeaderUtils.DEFAULT_DEBUG_HEADER_FILTER.apply(CONTENT_LENGTH, "3495"));
+
+        assertEquals(GZIP, HeaderUtils.DEFAULT_DEBUG_HEADER_FILTER.apply(TRANSFER_ENCODING, GZIP));
```
idelpivnitskiy: Consider using `CHUNKED` here instead of `GZIP` as a more relevant for this use-case.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
idelpivnitskiy: Please, also use our constants from `HttpHeaderNames` class.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
```diff
@@ -60,16 +64,16 @@
     static final int HASH_CODE_SEED = 0xc2b2ae35;
     public static final BiFunction<? super CharSequence, ? super CharSequence, CharSequence> DEFAULT_HEADER_FILTER =
             (k, v) -> "<filtered>";
+    private static final Set<CharSequence> DEFAULT_DEBUG_HEADER_NAMES = Collections.unmodifiableSet(
+            new HashSet<>(Arrays.asList(CONTENT_TYPE, CONTENT_LENGTH, TRANSFER_ENCODING)));
```
idelpivnitskiy: I would recommend keeping only `asList` that already returns an unmodifiable list of elements. When you use for-each loop it iterates over elements of the collection. The internal data structure is much simpler in the list (simple array) vs set (nodes with reference to the next element in memory). Therefore, the traversal will be quicker.

The other aspect is that when you use for-each loop with a collection java should create an iterator under the hood to be able to iterate. However, java 6u23 introduces escape analysis that will help to avoid creation of iterator for simple cases (it works for array-list) [1]. Therefore, java will optimize it to a simple memory counter and eliminate iterator allocation.

Here is a difference between unmodifiable set and list:

```java
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Level;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;

import java.util.HashSet;
import java.util.List;
import java.util.Set;
import java.util.concurrent.TimeUnit;

import static io.servicetalk.http.api.CharSequences.contentEqualsIgnoreCase;
import static io.servicetalk.http.api.CharSequences.newAsciiString;
import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_LENGTH;
import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_TYPE;
import static io.servicetalk.http.api.HttpHeaderNames.TRANSFER_ENCODING;
import static java.util.Arrays.asList;
import static java.util.Collections.unmodifiableSet;

@Fork(value = 1)
@State(Scope.Benchmark)
@BenchmarkMode(Mode.Throughput)
@Warmup(iterations = 5, time = 3, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 10, time = 3, timeUnit = TimeUnit.SECONDS)
public class ForEachBenchmark {

    private List<CharSequence> list;
    private Set<CharSequence> set;
    private CharSequence header;

    @Setup(Level.Trial)
    public void setup() {
        list = asList(CONTENT_TYPE, CONTENT_LENGTH, TRANSFER_ENCODING);
        set = unmodifiableSet(new HashSet<>(list));
        header = newAsciiString("some-header-name");
    }

    @Benchmark
    public boolean overList() {
        for (CharSequence headerName : list) {
            if (contentEqualsIgnoreCase(header, headerName)) {
                return true;
            }
        }
        return false;
    }

    @Benchmark
    public boolean overSet() {
        for (CharSequence headerName : set) {
            if (contentEqualsIgnoreCase(header, headerName)) {
                return true;
            }
        }
        return false;
    }
}
```

Results:

```
Benchmark                   Mode  Cnt         Score         Error  Units
ForEachBenchmark.overList  thrpt   10  96009696.616 ± 1279098.553  ops/s
ForEachBenchmark.overSet   thrpt   10  25110467.290 ±  294802.130  ops/s
```

Iteration over the list ~4 times faster.

[1] http://psy-lob-saw.blogspot.com/2014/12/the-escape-of-arraylistiterator.html

karim-elngr: That makes a lot of sense.

One note though, reading from the documentation for `asList` 

> Returns a fixed-size list backed by the specified array. Changes to the returned list "write through" to the array.

therefore, the original list can be changed, I benchmarked `unmodifiableList` but it's about 10-20% slower.

```
Benchmark                          Mode  Cnt         Score          Error  Units
MyBenchmark.overList              thrpt   10  97029293.383 ± 14275818.096  ops/s
MyBenchmark.overUnmodifiableList  thrpt   10  85589843.703 ±   258289.995  ops/s
```

If we should keep only `asList`, how about using normal arrays `CharSequence[]` instead, since it will be faster, but either way, we sacrifice immutability.

idelpivnitskiy: Because this is a private list, I won't worry too much about immutability. Nobody will try to alter its state if the scope is limited. I think `asList` is fine.

idelpivnitskiy:

karim-elngr:
Thanks for all the help @idelpivnitskiy @NiteshKant

PR #882, opened 2019-11-28
Move PlatformDependent to `servicetalk-utils-internal` #879

### LINKED ISSUE #879: Move PlatformDependent to `servicetalk-utils-internal`
`PlatformDependent` class lives in `servicetalk-concurrent-internal` module, but is not designed only for concurrent primitives. Different modules use it and therefore depend on `servicetalk-concurrent-internal` (and transitively on `servicetalk-concurrent`). This is not ideal, because requires unnecessary dependencies.

Because we recently introduced `servicetalk-utils-internal` module, we can move `PlatformDependent` and all related pkg-private utility classes there.

After this move, we should investigate which modules do not need to depend on `servicetalk-concurrent-internal` module anymore. For example, `servicetalk-buffer-netty`.

muthupalaniappan:
Motivation:

PlatformDependent class lives in `servicetalk-concurrent-internal` module, but is not designed only for concurrent primitives. Different modules use it and therefore depend on servicetalk-concurrent-internal (and transitively on servicetalk-concurrent). This is not ideal, because requires unnecessary dependencies.

Modifications:

* Moved files `PlatformDependent.java`, `PlatformDependent0.java`, `ReflectionUtils.java` from `servicetalk-concurrent-internal` module to `servicetalk-utils-internal` module.
* Moved `gradle/checkstyle/suppressions.xml` from `servicetalk-concurrent-internal` module to `servicetalk-utils-internal` module.
* Removed `jctools` dependency from `servicetalk-concurrent-internal`.
* Added `jctools` and `sl4j` dependencies to `servicetalk-utils-internal`.
* Added `servicetalk-utils-internal` project as implementation dependency to `servicetalk-buffer-netty`, `servicetalk-concurrent-api`, `servicetalk-concurrent-api-internal`, `servicetalk-concurrent-internal`, `servicetalk-data-jakson`, `servicetalk-grps-api`, `servicetalk-transport-netty-internal`.
* Added `servicetalk-utils-internal` project as `testImplementation` dependency to `servicetalk-grpc-netty` and `servicetalk-http-netty`.
* Removed `servicetalk-concurrent-internal` dependency from `servicetalk-buffer-netty`.
* Modified a lot of class files import and static import statements to use `io.servicetalk.util.internal...` instead of `io.servicetalk.concurrent.internal...` for `PlatformDependent`, `PlatformDependent0`, `ReflectionUtils` classes and its inner static members.

Result:

`PlatformDependent` class moved into `servicetalk-utils-internal` module.

Fixes #879


idelpivnitskiy:

PR #891, opened 2019-12-10
Allow default deserializers to handle payload without specified encoding

sonicloong:
Motivation:

Many existing services don't include charset in `Content-Type` header.
Therefore it would be convenient to have a default (UTF-8) deserializer
that can handle payload if charset is UTF-8 or no charset is specified.

Modifications:

- Add a rule to `hasContentType` method to accept no charset case when
  the expected charset is UTF-8.

Result:

The default `textDeserializer` and `formUrlEncodedDeserializer` can
deserialize payload of request/response with no charset in
`Content-Type` header.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
@@ -668,14 +668,15 @@ public final class HeaderUtils {
         }
 
         if (UTF_8.equals(expectedCharset)) {
+            if (!hasCharset(contentTypeHeader)) {
+                return true;
+            }
             if (contentEqualsIgnoreCase(expectedContentType, TEXT_PLAIN) &&
-                    (contentEqualsIgnoreCase(contentTypeHeader, TEXT_PLAIN_UTF_8) ||
-                            !hasCharset(contentTypeHeader))) {
+                    contentEqualsIgnoreCase(contentTypeHeader, TEXT_PLAIN_UTF_8)) {
                 return true;
             }
             if (contentEqualsIgnoreCase(expectedContentType, APPLICATION_X_WWW_FORM_URLENCODED) &&
-                    (contentEqualsIgnoreCase(contentTypeHeader, APPLICATION_X_WWW_FORM_URLENCODED_UTF_8) ||
-                            !hasCharset(contentTypeHeader))) {
+                    contentEqualsIgnoreCase(contentTypeHeader, APPLICATION_X_WWW_FORM_URLENCODED_UTF_8)) {
                 return true;
             }
         }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
@@ -36,11 +36,7 @@ import static io.servicetalk.http.api.CharSequences.regionMatches;
 import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_LENGTH;
 import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_TYPE;
 import static io.servicetalk.http.api.HttpHeaderNames.TRANSFER_ENCODING;
-import static io.servicetalk.http.api.HttpHeaderValues.APPLICATION_X_WWW_FORM_URLENCODED;
-import static io.servicetalk.http.api.HttpHeaderValues.APPLICATION_X_WWW_FORM_URLENCODED_UTF_8;
 import static io.servicetalk.http.api.HttpHeaderValues.CHUNKED;
-import static io.servicetalk.http.api.HttpHeaderValues.TEXT_PLAIN;
-import static io.servicetalk.http.api.HttpHeaderValues.TEXT_PLAIN_UTF_8;
 import static io.servicetalk.http.api.NetUtils.isValidIpV4Address;
 import static io.servicetalk.http.api.NetUtils.isValidIpV6Address;
 import static java.lang.Math.min;
@@ -669,15 +665,13 @@ public final class HeaderUtils {
         }
 
         if (UTF_8.equals(expectedCharset)) {
+            if (contentTypeHeader.length() == expectedContentType.length()) {
+                return true;
+            }
+            if (regionMatches(contentTypeHeader, true, expectedContentType.length(), "; charset=UTF-8", 0, 15)) {
+                return true;
+            }
             if (!hasCharset(contentTypeHeader)) {
-                return true;
-            }
-            if (contentEqualsIgnoreCase(expectedContentType, TEXT_PLAIN) &&
-                    contentEqualsIgnoreCase(contentTypeHeader, TEXT_PLAIN_UTF_8)) {
-                return true;
-            }
-            if (contentEqualsIgnoreCase(expectedContentType, APPLICATION_X_WWW_FORM_URLENCODED) &&
-                    contentEqualsIgnoreCase(contentTypeHeader, APPLICATION_X_WWW_FORM_URLENCODED_UTF_8)) {
                 return true;
             }
         }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
@@ -80,6 +80,7 @@ public final class HeaderUtils {
         return true;
     };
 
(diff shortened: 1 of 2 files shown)


NiteshKant:

NiteshKant:
@jayv the case that this PR is trying to handle is when some one specifies a deserializer with the charset as `UTF-8` but receives a `content-type` with no `charset` attribute. Will your suggestion also handle this case?

NiteshKant:

jayv:
Oh yes that makes sense @NiteshKant. 

In that case the downside is that we're now doing an `AsciiBuffer -> String` conversion on every such request and a second scan over the buffer from the start for which we expect the prefix to be identical. 

Wondering if we should create some utility method that does a case-insensitive scan for a prefix with optional suffix on `CharSequence` that leverages a single fast-path `ByteProcessor` on underlying `AsciiBuffer` that implements this algo in pseudo-code: 
```java
if (contentType.startsWith("text/plain")) {
    if (contentType.length() == "text/plain".length()) return true;
    else return contentType.containsIgnoreCase(";charset=utf8", "text/plain".length()); // continue from last offset
}
```
I understand there is more code towards the bottom of the method that uses regexes for other exotic charsets, which could leverage this utility if it existed. 

Maybe it's fine to leave it as it as-is and open an issue to improve or should we consider improving it now?

NiteshKant:
@jayv I am not opposed to optimizing code but yes as you suggested we can decouple optimization from this particular change.

For this specific change (/cc @sonicloong) we are adding another check which is doing the similar string comparisions as the previous check (eg: `UTF_8.equals(expectedCharset)`) which can be reduced by storing the results in a local variable.

jayv:
Created https://github.com/apple/servicetalk/issues/896 to follow-up on performance concerns.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
```diff
@@ -675,6 +675,13 @@ static boolean hasContentType(final HttpHeaders headers,
             return true;
         }
 
+        if (UTF_8.equals(expectedCharset) &&
```
NiteshKant: @sonicloong we are doing this and other checks above too, can we avoid doing multiple checks here, eg: by storing prior results in a local variable?

sonicloong: @NiteshKant that makes sense. I've consolidated the duplicate `UTF_8.equals`/`contentEqualsIgnoreCase` checks.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
```diff
@@ -667,12 +667,17 @@ static boolean hasContentType(final HttpHeaders headers,
             return false;
         }
 
-        if (UTF_8.equals(expectedCharset) &&
-                ((contentEqualsIgnoreCase(expectedContentType, TEXT_PLAIN) &&
-                        contentEqualsIgnoreCase(contentTypeHeader, TEXT_PLAIN_UTF_8)) ||
-                (contentEqualsIgnoreCase(expectedContentType, APPLICATION_X_WWW_FORM_URLENCODED) &&
-                        contentEqualsIgnoreCase(contentTypeHeader, APPLICATION_X_WWW_FORM_URLENCODED_UTF_8)))) {
-            return true;
+        if (UTF_8.equals(expectedCharset)) {
```
NiteshKant: It looks like what is required is:

```java
if (UTF_8.equals(expectedCharset)) {
    if (!hasCharset(contentTypeHeader)) {
        return true;
    }
    if (contentEqualsIgnoreCase(expectedContentType, TEXT_PLAIN) &&
            contentEqualsIgnoreCase(contentTypeHeader, TEXT_PLAIN_UTF_8)) {
        return true;
    }
    if (contentEqualsIgnoreCase(expectedContentType, APPLICATION_X_WWW_FORM_URLENCODED) &&
            contentEqualsIgnoreCase(contentTypeHeader, APPLICATION_X_WWW_FORM_URLENCODED_UTF_8)) {
        return true;
    }
}
```

I would also suggest adding another test to make sure we are checking for content-type equality without charset:

```java
assertFalse(HeaderUtils.hasContentType(
        headersWithContentType(of("image/png")), APPLICATION_X_WWW_FORM_URLENCODED, UTF_8));
```

sonicloong: @NiteshKant applied to any content type and added tests.

NiteshKant:

NiteshKant:
JDK 8 build timed out

NiteshKant:
@Scottmitch Can you verify this makes sense to do?

jayv:

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
```diff
@@ -667,12 +667,18 @@ static boolean hasContentType(final HttpHeaders headers,
...
+            if (!hasCharset(contentTypeHeader)) {
+                return true;
+            }
+            if (contentEqualsIgnoreCase(expectedContentType, TEXT_PLAIN) &&
+                    contentEqualsIgnoreCase(contentTypeHeader, TEXT_PLAIN_UTF_8)) {
+                return true;
+            }
+            if (contentEqualsIgnoreCase(expectedContentType, APPLICATION_X_WWW_FORM_URLENCODED) &&
+                    contentEqualsIgnoreCase(contentTypeHeader, APPLICATION_X_WWW_FORM_URLENCODED_UTF_8)) {
```
idelpivnitskiy: This is preexisting, but I think we don't need the full `contentEqualsIgnoreCase` here. My suggestion will be to add a length check as the first check inside `if (UTF_8.equals(expectedCharset))`. Because we already verified that we match the type and subtype in line 666, that will help to avoid unnecessary `hasCharset` and first `contentEqualsIgnoreCase`. 

The second suggestion will be to use `regionMatches` to find `; charset=UTF-8`. Because this is a fastlane shortcut, we should not worry about various OWS combinations, they will be handled after this `if`.

The final code for fastlane will be:
```java
if (UTF_8.equals(expectedCharset)) {
    if (contentTypeHeader.length() == expectedContentType.length()) {
        return true;
    }
    if (regionMatches(contentTypeHeader, true, expectedContentType.length(), "; charset=UTF-8", 0, 15)) {
        return true;
    }
    if (!hasCharset(contentTypeHeader)) {
        return true;
    }
}
```

servicetalk-http-api/src/test/java/io/servicetalk/http/api/HeaderUtilsTest.java
```diff
@@ -78,15 +78,26 @@ public void hasContentType() {
         assertFalse(HeaderUtils.hasContentType(
                 headersWithContentType(TEXT_PLAIN), APPLICATION_JSON, null));
 
-        assertFalse(HeaderUtils.hasContentType(
+        assertTrue(HeaderUtils.hasContentType(
```
idelpivnitskiy: Great!
Please, also add these tests while you are here:
```java
        assertTrue(HeaderUtils.hasContentType(
                headersWithContentType(of("Text/Plain")), TEXT_PLAIN, null));

        assertTrue(HeaderUtils.hasContentType(
                headersWithContentType(of("Text/Plain")), TEXT_PLAIN, UTF_8));
```

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
```diff
@@ -701,6 +707,10 @@ private static Pattern compileCharsetRegex(String charsetName) {
         return compile(".*;\\s*charset=\"?" + quote(charsetName) + "\"?\\s*(;.*|$)", CASE_INSENSITIVE);
     }
 
+    private static boolean hasCharset(final CharSequence contentTypeHeader) {
+        return contentTypeHeader.toString().contains("charset=");
```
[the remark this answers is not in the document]
sonicloong: Good catch. I'll fix it and add this test.

> where CHARSET_PATTERN = compile(".+charset=.+", CASE_INSENSITIVE)

Probably we should use `".+;\\s*charset=.+"`, in case the header contains unknown fields that end with *charset?

idelpivnitskiy:
@servicetalk-bot ok to test

idelpivnitskiy:
JDK11 build failed due to #911 

idelpivnitskiy:
Thanks for the contribution, @sonicloong!
❤️

PR #893, opened 2019-12-11
DirectByteBuf implementations should not force free memory

Scottmitch:
Motivation:
ServiceTalk doesn't expose reference counting and relies upon GC for memory
reclamation. Buffer objects are resizable an in the event a buffer is resized
(e.g. a write operation needs to expand the buffer) this will result in
allocating new memory and Netty will attempt to deallocate the old memory.
Applications may be multi-threaded and it is possible another thread will still
have a reference to this old memory location due to visibility constraints. We
should prevent Netty from force-freeing this old memory and instead rely upon GC
to avoid undefined behavior.

Modifications:
- ServiceTalk DirectByteBuf implementations to override freeDirect(ByteBuffer)
  and do nothing.

Result:
Less risk for resize operation resulting in referencing freed memory.


Scottmitch:
this change is mostly reverted by https://github.com/apple/servicetalk/pull/894 after discussion on https://github.com/netty/netty/pull/9850

PR #898, opened 2019-12-13
HTTP/1.x and HTTP/2.0 decoder Buffer visibility and data corruption

Scottmitch:
Motivation:
ServiceTalk attempts to effectivley disable Netty's reference counting by using
a strategy similar to Netty's `UnreleasableByteBuf` to short circit operations
which attempt to modify the reference count, and therefore prevent the reference
count from reaching zero and prematruley freeing memory. ServiceTalk's
ByteBuf instances will therefore always have a `refCnt() == 1`. However Netty's
ByteToMessageDecoder uses `refCnt() > 1` to determine if the cumulation ByteBuf
maybe shared, and if not it may compact data in place. This compacting data in
place may result in a corrupted view of data from the perspective of offloaded
threads in ServiceTalk, and for example result in failures to decode HTTP/2 data
such as gRPC frames.

ServiceTalk's `ByteToMessageDecoder` uses Netty's
`ByteToMessageDecoder#MERGE_CUMULATOR` to manage accumulation of data. As of
Netty 4.1.43.Final this may result in writing into the `cumulation` ByteBuf such
that a reallocation opearation occurs. This reallocation operation is not done
in a thread safe way and this may result in visibility issues when referencing
the same `Buffer` object from another thread (e.g. when offloading is enabled).
This may result in the applicaiton thread observing an empty `Buffer` if the
reference to the newly allocated memory is visibile before the copy operation is
visible.

Modifications:
- All `ByteBuf` objects generated by `ServiceTalkBufferAllocator` should have a
`refCnt() > 1` to avoid any optimizations which may assume exclusive ownership
of a `ByteBuf` in Netty. This will prevent corrupted content for HTTP/2.x and
gRPC use cases. This may result in more reallocation/copy operations in Netty's
`ByteToMessageDecoder` and there is a Netty PR
https://github.com/netty/netty/pull/9877 intended to improve this behavior.
- ServiceTalk's `ByteToMessageDecoder` should not use
`ByteToMessageDecoder#MERGE_CUMULATOR` and instead use a strategy similar to
https://github.com/netty/netty/pull/9877 to do the accumulation of bytes. It is
also unecessary to attempt to compact the buffer after accumulation has been
attempted, as any compaction can be handled in the accumulation.

Result:
No more decode related visibility issues for HTTP/1.x and no more content
corruption and visibility issues for HTTP/2.0 / gRPC.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ByteToMessageDecoder.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ByteToMessageDecoder.java
@@ -28,7 +28,6 @@ import io.netty.channel.ChannelProgressivePromise;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.EventLoop;
 import io.netty.channel.socket.ChannelInputShutdownEvent;
-import io.netty.handler.codec.ByteToMessageDecoder.Cumulator;
 import io.netty.handler.codec.DecoderException;
 import io.netty.handler.codec.DelimiterBasedFrameDecoder;
 import io.netty.handler.codec.FixedLengthFrameDecoder;
@@ -44,8 +43,7 @@ import java.util.List;
 import javax.annotation.Nullable;
 
 import static io.netty.buffer.Unpooled.EMPTY_BUFFER;
-import static io.netty.handler.codec.ByteToMessageDecoder.MERGE_CUMULATOR;
-import static java.lang.Math.min;
+import static java.lang.Integer.MAX_VALUE;
 
 /**
  * {@link ChannelInboundHandlerAdapter} which decodes bytes in a stream-like fashion from one {@link ByteBuf} to an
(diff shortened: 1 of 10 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ByteToMessageDecoder.java
```diff
@@ -217,39 +195,30 @@ public void channelReadComplete(ChannelHandlerContext ctx) {
...
      * @param cumulation The {@link ByteBuf} that accumulates across socket read operations.
-     * @param allocator Used to allocate a new {@link ByteBuf} if necessary.
-     * @return the {@link ByteBuf} that is responsible for accumulated socket reads.
+     * @param in The bytes to copy.
+     * @return the result of the swap and copy operation.
      */
-    protected ByteBuf swapCumulation(ByteBuf cumulation, ByteBufAllocator allocator) {
+    protected ByteBuf swapAndCopyCumulation(final ByteBufAllocator alloc, final ByteBuf cumulation, final ByteBuf in) {
+        ByteBuf newCumulation = alloc.buffer(alloc.calculateNewCapacity(
+                cumulation.readableBytes() + in.readableBytes(), MAX_VALUE));
```
idelpivnitskiy: Should we take care about `int` overflow here? Otherwise, users may get `IllegalArgumentException` from `calculateNewCapacity`

Scottmitch: if we overflow `int` then we wouldn't be able to accommodate all the data in a single buffer. We could attempt to fall back to a multi-buffer approach, or provide a pre-check for a more informative exception, but since this case is unlikely to occur I think the current "check in a single spot" approach is sufficient for now.

PR #902, opened 2019-12-21
Publishing of docs didn't update SNAPSHOT versions.

jayv:
__Motivation__

A prior fix to avoid overwriting javadoc indexes for older versions with the antora placeholders introduced an issue where SNAPSHOT docs were no longer being updated.

__Modifications__

- Change the approach to filtering which ensure only the javadocs indices are reset
- Fix an issue with javadocs for SNAPSHOT not cleaning up older versions of removed or renamed classes

__Result__

Correct docs.

### DIFF
--- a/scripts/publish-docs.sh
+++ b/scripts/publish-docs.sh
@@ -81,7 +81,7 @@ if [ ! -z "$version" ]; then
 fi
 
 pushd gh-pages
-# Do not override older javadoc with anotra's placeholder:
+# Do not override older javadoc with Antora's placeholder:
 git diff --name-only | grep 'javadoc/index.html' | grep -v $version | grep -v SNAPSHOT | xargs git checkout --
 
 git add * .nojekyll
--- a/scripts/publish-docs.sh
+++ b/scripts/publish-docs.sh
@@ -71,6 +71,8 @@ git worktree add gh-pages gh-pages
 touch gh-pages/.nojekyll
 \cp -r $DOCS_FOLDER/* gh-pages
 echo "Copy javadoc to gh-pages/servicetalk/SNAPSHOT"
+# Avoid accumulating old javadocs for classes that have been moved, renamed or deleted.
+rm -rf gh-pages/servicetalk/SNAPSHOT/javadoc
 \cp -r $JAVADOC_FOLDER gh-pages/servicetalk/SNAPSHOT
 if [ ! -z "$version" ]; then
     echo "Copy javadoc to gh-pages/servicetalk/$version"
@@ -79,10 +81,7 @@ fi
 
 pushd gh-pages
 # Do not override older javadoc with anotra's placeholder:
-for file in $(git diff --name-only | grep -v "servicetalk/SNAPSHOT/javadoc/index.html" | \
-  grep -v "servicetalk/$version/javadoc/index.html"); do
-    git checkout -- $file
-done
+git diff --name-only | grep 'javadoc/index.html' | xargs git checkout --
 
 git add * .nojekyll
 if [ -z "$version" ]; then
@@ -94,8 +93,10 @@ fi
 git push docs gh-pages
 popd
 
-rm -rf gh-pages
-git worktree prune
+# Cleanup gh-pages state for future runs based on remote
+git worktree remove gh-pages
+# above takes care of removal: rm -rf gh-pages
+git branch -D gh-pages
 
 if [ -z "$version" ]; then
     echo "Docs website for the SNAPSHOT version successfully updated"
--- a/scripts/publish-docs.sh
+++ b/scripts/publish-docs.sh
@@ -76,12 +76,13 @@ rm -rf gh-pages/servicetalk/SNAPSHOT/javadoc
 \cp -r $JAVADOC_FOLDER gh-pages/servicetalk/SNAPSHOT
 if [ ! -z "$version" ]; then
     echo "Copy javadoc to gh-pages/servicetalk/$version"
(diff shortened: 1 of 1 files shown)


jayv:
We can merge later, new docs were published already.

scripts/publish-docs.sh
```diff
@@ -71,6 +71,8 @@ git worktree add gh-pages gh-pages
 touch gh-pages/.nojekyll
 \cp -r $DOCS_FOLDER/* gh-pages
 echo "Copy javadoc to gh-pages/servicetalk/SNAPSHOT"
+# Avoid accumulating old javadocs for classes that have been moved, renamed or deleted.
+rm -rf gh-pages/servicetalk/SNAPSHOT/javadoc
```
idelpivnitskiy: Great catch! 👍 
For the cases when we want to update the release version as well (for example, when we release a patch version like 0.21.1 and want to update 0.21 javadoc) we need to do the same for `gh-pages/servicetalk/$version/javadoc` below, inside the `if` statement.

scripts/publish-docs.sh
```diff
@@ -79,10 +81,7 @@ fi
 
 pushd gh-pages
 # Do not override older javadoc with anotra's placeholder:
-for file in $(git diff --name-only | grep -v "servicetalk/SNAPSHOT/javadoc/index.html" | \
-  grep -v "servicetalk/$version/javadoc/index.html"); do
-    git checkout -- $file
-done
+git diff --name-only | grep 'javadoc/index.html' | xargs git checkout --
```
idelpivnitskiy: Thanks for simplifying with `xargs` 👍 
However, you have to keep excluding `servicetalk/SNAPSHOT/javadoc/index.html` and `servicetalk/$version/javadoc/index.html, otherwise they won't be updated and the old version from the git history will be recovered.

scripts/publish-docs.sh
```diff
@@ -94,8 +93,10 @@ fi
 git push docs gh-pages
 popd
 
-rm -rf gh-pages
-git worktree prune
+# Cleanup gh-pages state for future runs based on remote
+git worktree remove gh-pages
+# above takes care of removal: rm -rf gh-pages
+git branch -D gh-pages
```
idelpivnitskiy: Could you please provide more context why it should be changed and how does it change existing behavior? Not a `worktree` master :)
Just don't understand how does it aligns with `git worktree add gh-pages gh-pages` in line 69.

scripts/publish-docs.sh
jayv: This code seems to have updated the currently published docs AFAICT:

Docs:
https://apple.github.io/servicetalk/servicetalk/SNAPSHOT/javadoc/index.html
https://apple.github.io/servicetalk/servicetalk/0.21/javadoc/index.html

Commit:
https://github.com/apple/servicetalk/commit/b9ab3f02c5ac66a2e0256a4cbdcb157ce742c182

scripts/publish-docs.sh
jayv: While attempting to publish and fix the bugs my repo got wedged multiple times, requiring to do these manual steps as instructed in our docs. This way we avoid unpredictable state and just start clean from wherever we publish.

scripts/publish-docs.sh
jayv: FYI this was not a simplification, this was actually the bug ;)

idelpivnitskiy: I don't think so, your `rm -rf` fixes the issue described in the `modifications` section as
> Fix an issue with javadocs for SNAPSHOT not cleaning up older versions of removed or renamed classes

And this is important fix, thank for catching it.

The `servicetalk/SNAPSHOT/javadoc/index.html` and `servicetalk/$version/javadoc/index.html` will be updated by that `rm` and the following `cp`.

If you won't exclude them from this `git checkout --`, git will reset all changes for `SNAPSHOT/javadoc/index.html` (it won't be updated, git will restore the old version).

In your first commit (before these changes) I see that `SNAPSHOT/javadoc/index.html` was updated correctly: https://github.com/apple/servicetalk/commit/b9ab3f02c5ac66a2e0256a4cbdcb157ce742c182#diff-09ce3dd0f73c5e798b81c37a99da5b94
In the following commit (https://github.com/apple/servicetalk/commit/08c2e3b57fc84f83e5a40752bf73674091fd203c) I see that the timestamp was not updated in `SNAPSHOT/javadoc/index.html`, because you did `git checkout --` for it. However, other files in SNAPSHOT javadoc has updated timestamp: https://github.com/apple/servicetalk/commit/08c2e3b57fc84f83e5a40752bf73674091fd203c#diff-03104bc52fa54b8040ffd814d55d6402
This is not correct. We have to exclude `servicetalk/SNAPSHOT/javadoc/index.html` and `servicetalk/$version/javadoc/index.html` from `git checkout --`.

jayv: I understand the intent of this section of the script, however it was resetting too much and that was the bug. 

The timestamp doesn't update because we don't run `clean`, hence you'll see the same timestamps in both commits. 

You are correct however that we still need to filter `SNAPSHOT` and `$version`, it worked out well in my earlier testing due to the previous build output not revealing any changes for those files.

scripts/publish-docs.sh
```diff
@@ -79,10 +81,7 @@ fi
 
 pushd gh-pages
 # Do not override older javadoc with anotra's placeholder:
```
ddossot: Since this code is receiving TLC, consider fixing "anotra's" into "Antora's" 😊

PR #903, opened 2019-12-23
gRPC services to support closing

NiteshKant:
__Motivation__

gRPC implementation does not wire close implementations (by implementing methods in `AsyncCloseable`, `GracefulAutoCloseable` or `AutoCloseable`) defines by services such that server closure will invoke the appropriate close methods on the service. Also, there is no way to specify close semantics while implementing interfaces for each RPC method.

__Modification__

- Defined two new interfaces; `Rpc` and `BlockingRpc` that are appropriately implemented by individual RPC methods.
- Introduced a `wrap()` method for each existing `Route` interface variant such that a lambda implementing the route can also define a detached close implementation and this method will attach the close implementation to the route implementation by wrapping.
- Modified gRPC code generation to utilize the above changes.
- Modified gRPC router to use these new functionality and wire close together through various route layers.
- Added tests to verify each close implementation variant.
- Fixed a bug in blocking -> async service conversions that were not implementing graceful close (required for this change).

__Result__

gRPC services can now implement close methods when required.


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -98,30 +108,48 @@ private GrpcRouter(final Map<String, RouteProvider> routes,
...
-            if (request.method() != HttpRequestMethod.POST || (service = allRoutes.get(request.path())) == null) {
-                return notFound.handle(ctx, request, responseFactory);
-            } else {
-                return service.handle(ctx, request, responseFactory);
+        return binder.bindStreaming(new StreamingHttpService() {
+            @Override
+            public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
+                                                        final StreamingHttpRequest request,
+                                                        final StreamingHttpResponseFactory responseFactory) {
+                StreamingHttpService service;
```
Scottmitch: nit while you are here: make this `final`?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -184,24 +212,38 @@ RouteProviders drainRoutes() {
...
+                            try {
+                                final GrpcServiceContext serviceContext =
+                                        new DefaultGrpcServiceContext(request.path(), ctx);
+                                final HttpDeserializer<Req> deserializer =
+                                        serializationProvider.deserializerFor(readGrpcMessageEncoding(request),
+                                                requestClass);
+                                return route.handle(serviceContext, request.payloadBody(deserializer))
+                                        .map(rawResp -> newResponse(responseFactory,
+                                                ctx.executionContext().bufferAllocator())
+                                                .payloadBody(uncheckedCast(rawResp),
```
Scottmitch: while you are here: `uncheckedCast` shouldn't be necessary (`Class<Resp>` should be used to give type safety) ... consider removal.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -214,19 +256,37 @@ RouteProviders drainRoutes() {
...
+                                                                final StreamingHttpResponseFactory responseFactory) {
+                        try {
+                            final GrpcServiceContext serviceContext =
+                                    new DefaultGrpcServiceContext(request.path(), ctx);
+                            final HttpDeserializer<Req> deserializer =
+                                    serializationProvider.deserializerFor(readGrpcMessageEncoding(request),
+                                            requestClass);
+                            final Publisher<Resp> response = route.handle(serviceContext,
+                                    request.payloadBody(deserializer))
+                                    .map(GrpcUtils::uncheckedCast);
```
Scottmitch: while you are here: `uncheckedCast` shouldn't be necessary (`Class<Resp>` should be used to give type safety) ... consider removal.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -214,19 +256,37 @@ RouteProviders drainRoutes() {
...
-                        return succeeded(newErrorResponse(responseFactory, t,
-                                ctx.executionContext().bufferAllocator()));
+                StreamingHttpService service = new StreamingHttpService() {
+                    @Override
+                    public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
+                                                                final StreamingHttpRequest request,
+                                                                final StreamingHttpResponseFactory responseFactory) {
+                        try {
+                            final GrpcServiceContext serviceContext =
+                                    new DefaultGrpcServiceContext(request.path(), ctx);
```
Scottmitch: future consideration for perf (/cc @jayv) ... `path()` takes some extra parsing and by this point we should have already gone through the router layer (which currently also does `path()`) but I wonder if we could just use the raw request target and avoid the parsing all together (and just bail if there are any additional components of the URL).

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -292,19 +382,33 @@ public HttpExecutionStrategy serviceInvocationStrategy() {
...
-                            final Resp response = route.handle(serviceContext, request.payloadBody(deserializer));
-                            return newResponse(responseFactory, ctx.executionContext().bufferAllocator())
-                                    .payloadBody(response,
-                                            serializationProvider.serializerFor(serviceContext, responseClass));
-                        } catch (Throwable t) {
-                            return newErrorResponse(responseFactory, t, ctx.executionContext().bufferAllocator());
+                    toStreamingHttpService(new BlockingHttpService() {
+                        @Override
+                        public HttpResponse handle(final HttpServiceContext ctx, final HttpRequest request,
+                                                   final HttpResponseFactory responseFactory) throws Exception {
```
Scottmitch: nit: remove `throws Exception`

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
```diff
@@ -208,7 +208,7 @@ protected GrpcService newServiceFromRoutes(final AllGrpcRoutes routes) {
             final String path, final GrpcExecutionStrategy executionStrategy,
             final StreamingRoute<Req, Resp> route, final Class<Req> requestClass,
             final Class<Resp> responseClass, final GrpcSerializationProvider serializationProvider) {
-        routeBuilder.addStreamingRoute(path, null, route, requestClass, responseClass,
+        routeBuilder.addStreamingRoute(path, executionStrategy, route, requestClass, responseClass,
```
Scottmitch: good catch ... unit test here?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
```diff
@@ -446,6 +446,7 @@ protected GrpcService newServiceFromRoutes(final AllGrpcRoutes routes) {
      * @param <Req> Type of request.
      * @param <Resp> Type of response.
      */
+    @FunctionalInterface
```
Scottmitch: @idelpivnitskiy 🎁 👍

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/Rpc.java
```diff
@@ -0,0 +1,31 @@
...
+
+import io.servicetalk.concurrent.api.AsyncCloseable;
+import io.servicetalk.concurrent.api.Completable;
+
+import static io.servicetalk.concurrent.api.Completable.completed;
+
+/**
+ * A contract for generated RPC method interface.
+ */
+public interface Rpc extends AsyncCloseable {
```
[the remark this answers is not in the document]
NiteshKant: The intent for these new interfaces as well as existing `GrpcService` and `BlockingGrpcService` is to provide a contract for these services such that we can add behaviors in the future (with default implementations) without modifying generated code.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
NiteshKant: We do cache the parsed result for the `URI` inside the request implementation[1] so calling `path()` multiple times would not incur additional parsing. Lets see if this shows up in perf and we can fix, thanks for pointing out 🙏 

[1] https://github.com/apple/servicetalk/blob/master/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequestMetaData.java#L293-L295

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
NiteshKant: I have created an issue #907 to add tests to verify the strategy used by routes as @idelpivnitskiy is working on it ATM.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/BlockingRpc.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+package io.servicetalk.grpc.api;
+
+import io.servicetalk.concurrent.GracefulAutoCloseable;
+
+/**
+ * A contract for generated blocking RPC method interface.
+ */
+public interface BlockingRpc extends GracefulAutoCloseable {
```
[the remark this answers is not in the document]
NiteshKant: Actually using the `[Blocking]GrpcService` interface here is sufficient, let me do that

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ClosureTest.java
```diff
@@ -0,0 +1,271 @@
...
+    public static Collection<Boolean> data() {
+        return asList(true, false);
+    }
+
+    @Test
+    public void serviceImplIsClosed() throws Exception {
+        CloseSignal signal = new CloseSignal();
+        TesterService svc = setupCloseMock(mock(TesterService.class), signal);
+        startServerAndClose(new ServiceFactory(svc), signal);
+        verifyClosure(svc, 4 /* 4 rpc methods */);
```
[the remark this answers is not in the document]
NiteshKant: It has been our assumption all the way that close methods are idempotent like `AutoCloseable` but ya it is a good idea to be explicit about it in the javadoc. I will do that as a follow up.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/BlockingRpc.java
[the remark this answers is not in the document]
idelpivnitskiy: Can we kill these interfaces now?

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ClosureTest.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    @Parameterized.Parameters(name = "graceful? => {0}")
+    public static Collection<Boolean> data() {
+        return asList(true, false);
+    }
+
+    @Test
+    public void serviceImplIsClosed() throws Exception {
+        CloseSignal signal = new CloseSignal(1);
```
idelpivnitskiy: Should it be 4?

idelpivnitskiy:
@NiteshKant btw, do you want to annotate generated `*Rpc` interfaces with `@FunctionalInterface` as part of this work?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/BlockingRpc.java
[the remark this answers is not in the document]
NiteshKant: Uggh yes ...

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/Rpc.java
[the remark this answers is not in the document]
NiteshKant: Update: 

As part of [this dicussion](https://github.com/apple/servicetalk/pull/903#discussion_r362641889) I have removed this new interfaces and using the existing `[Blocking]GrpcService` interfaces instead.

NiteshKant:
>> do you want to annotate generated *Rpc interfaces with @FunctionalInterface as part of this work?

@idelpivnitskiy I would prefer not to increase the scope of this PR now as it is close to merging.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ClosureTest.java
```diff
@@ -0,0 +1,267 @@
...
+        CloseSignal signal = new CloseSignal(4);
+        TesterService svc = setupCloseMock(mock(TesterService.class), signal);
+        startServerAndClose(new ServiceFactory(svc), signal);
+        verifyClosure(svc, 4 /* 4 rpc methods */);
+        signal.verifyCloseAtLeastCount(closeGracefully);
+    }
+
+    @Test
+    public void blockingServiceImplIsClosed() throws Exception {
+        CloseSignal signal = new CloseSignal(1);
```
idelpivnitskiy: This also should be `4` :)

NiteshKant: Hmm actually I take it back. This and the one in `serviceImplIsClosed()` should be `1` as we only expect 1 close call but due to current implementation detail it is called `4` times (as you correctly pointed out)

PR #909, opened 2020-01-06
Update protobuf to 3.11.2

Scottmitch:


Scottmitch:
hmm guess they only did a php release https://github.com/protocolbuffers/protobuf/releases

PR #910, opened 2020-01-07
Move RPC paths to generated RPC interfaces

NiteshKant:
__Motivation__

gRPC generated code uses an `Enum` to define all RPC paths. It is convenient to have all paths together but they are detached from the actual RPC methods for which the paths are defined.

__Modification__

- Remove `RpcPaths` `Enum` from generated code.
- Generate a `PATH` constant for all individual RPC interfaces.
- Async interfaces defines the `PATH` constant with the actual path `String`.
- Blocking interfaces define the `PATH` constant which referr to the async `PATH` constant.

__Result__

RPC paths are attached to the associate RPC interfaces.

### DIFF
--- a/servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
+++ b/servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
@@ -29,15 +29,12 @@ import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 import java.util.function.BiFunction;
-import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import javax.lang.model.element.Modifier;
 
 import static com.squareup.javapoet.MethodSpec.constructorBuilder;
 import static com.squareup.javapoet.MethodSpec.methodBuilder;
-import static com.squareup.javapoet.TypeSpec.anonymousClassBuilder;
 import static com.squareup.javapoet.TypeSpec.classBuilder;
-import static com.squareup.javapoet.TypeSpec.enumBuilder;
 import static com.squareup.javapoet.TypeSpec.interfaceBuilder;
 import static io.servicetalk.grpc.protoc.Generator.NewRpcMethodFlag.BLOCKING;
 import static io.servicetalk.grpc.protoc.Generator.NewRpcMethodFlag.CLIENT;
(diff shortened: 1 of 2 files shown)


servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -250,26 +222,50 @@ private void addSerializationProviderInit(final State state, final TypeSpec.Buil
 
     private void addServiceRpcInterfaces(final State state, final TypeSpec.Builder serviceClassBuilder) {
         state.serviceRpcInterfaces = new ArrayList<>(2 * state.serviceProto.getMethodCount());
-
-        state.serviceProto.getMethodList().forEach(methodProto -> Stream.of(false, true).forEach(blocking -> {
-            final String name = context.deconflictJavaTypeName((blocking ? Blocking : "") +
+        state.serviceProto.getMethodList().forEach(methodProto -> {
+            final String name = context.deconflictJavaTypeName(
```
idelpivnitskiy: What was the reason to decouple async and blocking interfaces generation? The code for them is almost similar. With your changes you will need to do different `pathSpecBuilder.initializer` depending on the `blocking` flag. For everything else all conditions were preexisting. Consider reverting the previous logic.

NiteshKant: The primary motivation is to have the ability to refer to the corresponding async RPC interface `PATH` constant. Since we deconflict the name, there is no way to get the same name as was used by the async RPC interface apart from what I am doing here.
I can extract common code between these two lambdas into a separate method if you prefer but we will have to do the following:
- Iterate over the methods in proto file to generate async interfaces.
- Iterate over the async interfaces to generate the corresponding blocking interfaces.

This also has a nice side-effect that async and blocking interfaces for the same RPC method has similar names. eg: if the RPC method name is `foo` but the service name is also `foo` then with existing code, async RPC name will be `Foo0` but blocking interface will be `BlockingFoo`. After this change the names will be `Foo0` and `BlockingFoo0` which is nicer.

idelpivnitskiy: Agreed, thanks for clarification.

PR #912, opened 2020-01-08
Support `router-api` annotations for gRPC services

idelpivnitskiy:
Motivation:

gRPC should support annotations that help to configure the execution
strategy per service or per route. Also, gRPC should compute the
effective execution strategy based on the main execution strategy
from the builder configuration and from the route configuration.

Modifications:

- Parse annotations from `route-api` module;
- Compute the difference between execution strategy on the builder
and on the route;
- Add tests to verify that offloading works correctly for different
execution strategy configurations;
- Add tests to verify that misconfiguration of `route-api` annotations
will be explained to the user;
- Adjust gRPC codegen for new features;
- Change `difference` functions to compare with `noOffloadsStrategy`
by configuration, not by reference;
- Make `NoOffloadsHttpExecutionStrategy` no op;
- Fix tests;

Result:

gRPC routes correctly compute per-route execution strategy
configuration.


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategies.java
```diff
@@ -117,10 +117,10 @@ public static Builder customStrategyBuilder() {
     public static HttpExecutionStrategy difference(final Executor fallback,
                                                    final HttpExecutionStrategy left,
                                                    final HttpExecutionStrategy right) {
-        if (left.equals(right) || right == noOffloadsStrategy()) {
+        if (left.equals(right) || noOffloads(right)) {
             return null;
         }
-        if (left == noOffloadsStrategy()) {
+        if (noOffloads(left)) {
```
idelpivnitskiy: We can not compare with `noOffloadsStrategy()` by reference, because we may receive a `noOffloadsStrategy()` which is wrapped by `DefaultGrpcExecutionStrategy` here or users may give their own implementation of the strategy without offloading.
@NiteshKant WDYT about introducing something like `boolean HttpExecutionStrategy.isNoOffloading()`? Or maybe even promoting this method to `ExecutionStrategy` interface. That will help us to avoid duplication of `noOffloads` method impl, we will be able to use it in `DefaultHttpExecutionStrategy.offloadService` in addition to `diff == null` check, and users will have an easy way to check that there is no offloading.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ExecutionStrategyTestServices.java
```diff
@@ -0,0 +1,270 @@
...
+
+    /// Async API:
+
+    private static class EsAsyncService implements TesterService {
+
+        @Override
+        public Single<TestResponse> test(final GrpcServiceContext ctx, final TestRequest request) {
+            final ThreadInfo threadInfo = new ThreadInfo(ctx);
+            return succeeded(threadInfo)
+                    .afterOnSubscribe(__ -> threadInfo.responseOnSubscribeThreadName = threadName())
```
idelpivnitskiy: I see that `Single` and `Completable` have `before/afterOnSubscribe`, but doesn't have `whenOnSubscribe` (but `Publisher` has it). If they were missed, I can add them in a follow-up.
@NiteshKant wdyt?

Also, we don't have `whenSubscriber` for all 3 async sources and `Publisher.whenSubscription`, but `before/after*` variants are exist.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
```diff
@@ -716,45 +717,56 @@ private CompatResponse response(final int value) {
...
-            }
+        };
+        final ServiceFactory serviceFactory = strategy == defaultStrategy() ? new ServiceFactory(compatService) :
+                new ServiceFactory.Builder().bidirectionalStreamingCall(strategy, compatService)
+                        .clientStreamingCall(strategy, compatService)
+                        .scalarCall(strategy, compatService)
+                        .serverStreamingCall(strategy, compatService)
+                        .build();
+
+        // FIXME(idel): remove condition around appendServiceFilter when WSS or filters execution strategy will be fixed
```
NiteshKant: I am not yet convinced that the failure is due to the race you mention here, can we dig a bit deeper to see why are we consistently seeing exceptions without this change in the test here?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategies.java
NiteshKant: I do not think that we will have many users checking presence/absence of no-offloads so I do not think we should promote this to a public API. The code repetition in jersey code is perhaps we can avoid by using this `difference()` method in jersey? There shouldn't be a reason for us to be different in behavior between routers. Is changing jersey to use the common `difference()` method here something you are thinking of doing?

servicetalk-router-utils-internal/src/main/java/io/servicetalk/router/utils/internal/RouteExecutionStrategyUtils.java
```diff
@@ -0,0 +1,161 @@
...
+import java.lang.reflect.Method;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+import static java.util.Arrays.stream;
+
+/**
+ * Utilities to handle {@link RouteExecutionStrategy} annotation.
+ */
+public final class RouteExecutionStrategyUtils {
```
NiteshKant: I am assuming that code here is just refactored from existing jersey classes, so haven't reviewed this here. Is there anything specific you want me to look at?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategies.java
idelpivnitskiy: I see that the PR #682 changes jersey router to use its own impl of `difference` function instead of the `HttpExecutionStrategies.difference`. If I change `EndpointEnhancingRequestFilter` to use `HttpExecutionStrategies.difference`, some tests in jersey module fail.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -65,9 +65,7 @@
     }
 
     private static HttpExecutionStrategy serviceInvocationStrategy(final HttpExecutionStrategyInfluencer influencer) {
-        HttpExecutionStrategy httpExecutionStrategy = influencer.influenceStrategy(DEFAULT_STRATEGY);
-        assert httpExecutionStrategy.isMetadataReceiveOffloaded() : "This will deadlock!";
-        return httpExecutionStrategy;
+        return influencer.influenceStrategy(DEFAULT_STRATEGY);
```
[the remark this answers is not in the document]
idelpivnitskiy: Previously, we used this conversion service as part of the computation of the main strategy for the builder. Therefore, we tried to prevent the case when users could misconfigure execution strategy for the blocking service.
For gRPC, we use this service between the router and the final RPC method and we may already be on executor thread.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -342,9 +356,9 @@ private void addServiceFactory(final State state, final TypeSpec.Builder service
                             .addModifiers(PUBLIC)
                             .addParameter(rpcInterface.className, rpc, FINAL)
                             .returns(builderClass)
-                            .addStatement("$L($T.$L, $L.wrap($L::$L, $L), $T.class, $T.class, $L)",
-                                    addRouteMethodName, rpcInterface.className, RPC_PATH, routeInterfaceClass,
-                                    rpc, routeName, rpc, inClass, outClass, serializationProvider)
+                            .addStatement("$L($T.$L, $L.getClass(), $S, $L.wrap($L::$L, $L), $T.class, $T.class, $L)",
```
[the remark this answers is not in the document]
idelpivnitskiy: We cannot use `$T.class` here because we need an actual implementation provided by the user (that is annotated), instead of just an rpc static interface.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
idelpivnitskiy: As we discussed offline, when the offloading is enabled, it's expected that the client may receive either grpc-exception first or see that server closed the connection. See `grpcJavaToServiceTalkErrorInStreamingResponse` as an example. 
Tests with `noOffloadsStrategy()` always expect grpc-exception. Only tests with `*NoOffload` suffix are affected by this `FIXME` and it will be removed when we fix wrapping routes with filters because currently, filters override route's execution strategy.

servicetalk-router-api/src/main/java/io/servicetalk/router/api/RouteExecutionStrategyFactory.java
```diff
@@ -0,0 +1,38 @@
...
+public interface RouteExecutionStrategyFactory<ES> {
+
+    /**
+     * Gets an execution strategy for the specified {@code id}{@link RouteExecutionStrategy#id() id} of
+     * {@link RouteExecutionStrategy} annotation.
+     *
+     * @param id of {@link RouteExecutionStrategy}
+     * @return {@link ES} implementation or {@code null}
+     */
+    @Nullable
```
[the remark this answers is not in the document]
idelpivnitskiy: This is a good point. We use `__ -> null` as a default factory in Jersey. However, we have a error processing check that makes sure the factory provides an execution strategy for all ids. Let me try to use `IllegalArgumentException` instead.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ExecutionStrategyTestServices.java
NiteshKant: It may have got missed, creating a `good first issue` SGTM

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategies.java
NiteshKant: We can fix the tests if they are inaccurate or do not align with gRPC, there isn't a reason why they should be different. It is OK to do this in a follow up but we should look at reducing special cases between modules which makes it hard to reason about behavior later.

servicetalk-router-utils-internal/src/main/java/io/servicetalk/router/utils/internal/DefaultRouteExecutionStrategyFactory.java
```diff
@@ -0,0 +1,65 @@
...
+import io.servicetalk.router.api.RouteExecutionStrategyFactory;
+import io.servicetalk.transport.api.ExecutionStrategy;
+
+/**
+ * Implementation of {@link RouteExecutionStrategyFactory} that always throws {@link IllegalArgumentException}.
+ * This should be used as a default strategy factory when users did not provide their own implementation.
+ *
+ * @param <ES> protocol-specific execution strategy implementation
+ */
+public final class DefaultRouteExecutionStrategyFactory<ES extends ExecutionStrategy>
```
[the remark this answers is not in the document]
idelpivnitskiy: No, it's not only for tests, we use it as a default factory in jersey router and grpc `ServiceFactory`.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategies.java
idelpivnitskiy: Created an issue https://github.com/apple/servicetalk/issues/924

servicetalk-router-utils-internal/src/main/java/io/servicetalk/router/utils/internal/DefaultRouteExecutionStrategyFactory.java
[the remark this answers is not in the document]
NiteshKant: Hmm .. somehow did not see that. I still think the whole static raw-types and methods for rawtypes casting is more of a pain than convenience. I would suggest just use the lambda in the different routers.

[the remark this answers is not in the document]
idelpivnitskiy: I also use `getUsingDefaultStrategyFactory` method in jersey to handle it's conversion between the map to factory and handle `@Nonnull` properly: https://github.com/apple/servicetalk/pull/912/files/21cdc386803b4c7792264b7966ec103f27ca1983#diff-25d1a53373d0d3bfc4ab2d26e22c13fdR223
This class is in `servicetalk-router-utils-internal` module that helps to consolidate repeated lambda and utils.
Can we keep it?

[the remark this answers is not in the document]
NiteshKant: Yep did not notice that it is an internal class, did not notice before. Sounds good to keep.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ExecutionStrategyTestServices.java
idelpivnitskiy: https://github.com/apple/servicetalk/issues/926

PR #918, opened 2020-01-13
Remove LegacyBlockingSubscriber

Scottmitch:
Motivation:
LegacyBlockingSubscriber was intended for removal but still had some remaining
utility methods used to orchestrate sequential blocking test execution. This
functionality can be replaced by a TestCollectingPublisherSubscriber.

Modifications:
- Remove LegacyBlockingSubscriber
- Add TestCollectingPublisherSubscriber which can be used in combination with
TestPublisherSubscriber or independently.

Result:
Legacy code removed, more integrated sequential blocking test utilities.


servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCollectingPublisherSubscriber.java
```diff
@@ -0,0 +1,213 @@
...
+
+    @Override
+    public void onSubscribe(final Subscription subscription) {
+        this.subscription = subscription;
+        onSubscribeLatch.countDown();
+    }
+
+    @Override
+    public void onNext(@Nullable final T t) {
+        items.add(t == null ? NULL_ON_NEXT : t);
```
NiteshKant: Consider adding 

`assert onTerminal == null;`

to make sure there is no `onNext` after a terminal since we verify consumed items post terminal with
 `awaitOn[Error/Complete](boolean verifyOnNextConsumed)`

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCollectingPublisherSubscriber.java
```diff
@@ -0,0 +1,213 @@
...
+    /**
+     * Block until {@link #onSubscribe(Subscription)}.
+     *
+     * @return The {@link Subscription} from {@link #onSubscribe(Subscription)}.
+     * @throws InterruptedException if an interrupt occurs while blocking for waiting for
+     * {@link #onSubscribe(Subscription)}.
+     */
+    public Subscription awaitSubscription() throws InterruptedException {
+        onSubscribeLatch.await();
+        if (subscription == null) {
```
NiteshKant: Should we store `subscription` in a local var so that we are null-checking and returning the same value?

Alternatively, add an `assert subscription == null;` in `onSubscribe()`.

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCollectingPublisherSubscriber.java
```diff
@@ -0,0 +1,213 @@
...
+    /**
+     * Consume all currently available {@link #onNext(Object)} signals.
+     *
+     * @return {@link List} containing all currently available {@link #onNext(Object)} signals.
+     */
+    @SuppressWarnings("unchecked")
+    public List<T> pollAllOnNext() {
+        List<T> consumedItems = new ArrayList<>();
+        Object item;
+        while ((item = items.poll()) != null) {
```
NiteshKant: Consider using `drainTo()` and `stream()` for a more succinct representation as we can absorb higher allocations in test code:

```java
List<Object> consumedItems = new ArrayList<>();
items.drainTo(consumedItems);
return consumedItems.stream().map(item -> item == NULL_ON_NEXT ? null : (T) item).collect(toList());
```

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapSingleTest.java
```diff
@@ -515,31 +517,37 @@ public void onComplete() {
...
-        subscriber.awaitAndVerifyAwaitingItem(2); // Second item would only be sent once the first thread is free.
-        subscriber.unblock(2);
-        subscriber.verifyReceived(2);
+        Integer nextItem = blockingSubscriber.takeOnNext();
+        assertNotNull(nextItem);
+        assertEquals(2, nextItem.intValue());
+        nextItem = blockingSubscriber.takeOnNext();
+        assertNotNull(nextItem);
+        assertEquals(3, nextItem.intValue());
+        assertFalse(blockingSubscriber.pollTerminal(200, MILLISECONDS));
```
NiteshKant: Do we need this check? 

I am questioning usefullness more since we are adding `pollTerminal()` on `TestCollectingPublisherSubscriber` for this reason. Generally since we have test timeouts, we do not add APIs for time based polls, it will be better to avoid.

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCollectingPublisherSubscriber.java
```diff
@@ -0,0 +1,213 @@
...
+ *     sub.awaitOnComplete(); // this will verify that all onNext signals have been consumed
+ *
+ *     // Approach 2 - wait for terminal, verify items in bulk.
+ *     sub.awaitOnComplete(false); // wait for the terminal signal, ignore if there are unconsumed onNext signals.
+ *     List&lt;String&gt; onNextSignals = sub.pollAllOnNext();
+ *     // verify all onNextSignals occurred in the expected order
+ * </pre>
+ * @param <T> The type of data in {@link #onNext(Object)}.
+ */
+public final class TestCollectingPublisherSubscriber<T> implements Subscriber<T> {
```
NiteshKant: Should we add some tests for this class?

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCollectingPublisherSubscriber.java
```diff
@@ -0,0 +1,213 @@
...
+     * @param unit The unit of time to apply to the duration.
+     * @return {@code true} if a terminal event has been received before the timeout duration.
+     * @throws InterruptedException If an interrupt occurs while blocking for the terminal event.
+     */
+    public boolean pollTerminal(long timeout, TimeUnit unit) throws InterruptedException {
+        return onTerminalLatch.await(timeout, unit);
+    }
+
+    private void verifyAllOnNextProcessed() {
+        if (!items.isEmpty()) {
```
NiteshKant: You can use `pollAllOnNext()` to simplify:

```java
List<T> items = pollAllOnNext();
if (!items.isEmpty()) {
    throw new IllegalStateException(items.size() + " onNext items were not processed: " + items);
}
```

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCollectingPublisherSubscriber.java
Scottmitch: I shifted the null check to the `onSubscribe`, made sure the value wasn't set twice, and changed this to an assert. PTAL

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCollectingPublisherSubscriber.java
Scottmitch: done

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapSingleTest.java
Scottmitch: This is to test the affirmative that a terminal event isn't delivered until we expect it to be. Just checking `blockingSubscriber.awaitOnComplete();` would be "sufficient" but the event may have been delivered for some reason other than the source completing (e.g. operator bug).

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCollectingPublisherSubscriber.java
Scottmitch: will do

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCollectingPublisherSubscriber.java
Scottmitch: I still have to iterate to build the string, so I'll just keep this for now.

Scottmitch:
test this please

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCollectingPublisherSubscriber.java
NiteshKant: I think the above code I provided would output the string pretty close to what is built by the iteration but it is not a big deal, I was just intending to reduce code.

PR #919, opened 2020-01-14
Protocol emits payload boundary event

jayv:
__Motivation__

WriteStreamSubscriber is not aware of payload boundaries, hence when a Channel becomes suddenly inactive (closed or output shutdown) it could only assume the write failed. In case of H2 the codec would perfom an output shutdown after observing and emitting trailers. This output shutdown would race with write publisher completion and occasionally result in WriteStreamSubscriber failing the write with an error even though it actually completed normally.

__Modifications__

- H1/H2 codecs emit payload boundaries via CloseHandler
- CloseHandler triggers a new UserEvent on outbound payload end
- DefaultNettyConnection intercepts above event and notifies WriteListener (WriteStreamSubscriber)
- WriteStreamSubscriber responds to 2 triggers for write source stream completion

__Result__

Writes no longer fail due to race between closure of write publisher and H2 end of stream

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
@@ -67,7 +67,6 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
     @Override
     public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
         if (msg instanceof HttpRequestMetaData) {
-            closeHandler.protocolPayloadBeginOutbound(ctx);
             HttpRequestMetaData metaData = (HttpRequestMetaData) msg;
             HttpHeaders h1Headers = metaData.headers();
             CharSequence host = h1Headers.getAndRemove(HOST);
@@ -105,7 +104,6 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
             Http2Headers h2Headers = headersFrame.headers();
             final HttpResponseStatus httpStatus;
             if (!readHeaders) {
-                closeHandler.protocolPayloadBeginInbound(ctx);
                 CharSequence status = h2Headers.getAndRemove(STATUS.value());
                 if (status == null) {
                     throw new IllegalArgumentException("a response must have " + STATUS + " header");
@@ -139,7 +137,6 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
                 } else {
                     ctx.fireChannelRead(h2HeadersToH1HeadersClient(h2Headers, null));
                 }
-                closeHandler.protocolPayloadEndInbound(ctx);
             } else if (httpStatus == null) {
                 throw new IllegalArgumentException("a response must have " + STATUS + " header");
             } else {
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
@@ -23,6 +23,7 @@ import io.servicetalk.http.api.HttpRequestMetaData;
 import io.servicetalk.http.api.HttpRequestMethod;
 import io.servicetalk.http.api.HttpResponseStatus;
 import io.servicetalk.http.api.StreamingHttpResponse;
+import io.servicetalk.transport.netty.internal.CloseHandler;
 
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPromise;
@@ -56,8 +57,9 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
     @Nullable
     private HttpRequestMethod method;
 
-    H2ToStH1ClientDuplexHandler(boolean sslEnabled, BufferAllocator allocator, HttpHeadersFactory headersFactory) {
-        super(headersFactory);
+    H2ToStH1ClientDuplexHandler(boolean sslEnabled, BufferAllocator allocator, HttpHeadersFactory headersFactory,
+                                CloseHandler closeHandler) {
+        super(headersFactory, closeHandler);
         this.scheme = sslEnabled ? HttpScheme.HTTPS : HttpScheme.HTTP;
         this.allocator = allocator;
     }
@@ -65,6 +67,7 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
     @Override
     public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
         if (msg instanceof HttpRequestMetaData) {
+            closeHandler.protocolPayloadBeginOutbound(ctx);
             HttpRequestMetaData metaData = (HttpRequestMetaData) msg;
             HttpHeaders h1Headers = metaData.headers();
             CharSequence host = h1Headers.getAndRemove(HOST);
@@ -102,6 +105,7 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
             Http2Headers h2Headers = headersFrame.headers();
             final HttpResponseStatus httpStatus;
             if (!readHeaders) {
+                closeHandler.protocolPayloadBeginInbound(ctx);
                 CharSequence status = h2Headers.getAndRemove(STATUS.value());
                 if (status == null) {
                     throw new IllegalArgumentException("a response must have " + STATUS + " header");
@@ -135,6 +139,7 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
                 } else {
                     ctx.fireChannelRead(h2HeadersToH1HeadersClient(h2Headers, null));
                 }
+                closeHandler.protocolPayloadEndInbound(ctx);
             } else if (httpStatus == null) {
                 throw new IllegalArgumentException("a response must have " + STATUS + " header");
             } else {
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
@@ -28,11 +28,8 @@ import io.servicetalk.http.api.StreamingHttpService;
 import io.servicetalk.http.netty.NettyHttpServer.NettyHttpServerConnection;
 import io.servicetalk.tcp.netty.internal.TcpServerChannelInitializer;
 import io.servicetalk.transport.api.IoExecutor;
-import io.servicetalk.transport.netty.internal.CloseHandler;
 
-import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
-import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelOutboundHandlerAdapter;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.embedded.EmbeddedChannel;
@@ -129,23 +126,7 @@ public class FlushStrategyOnServerTest {
 
         final ReadOnlyHttpServerConfig config = new HttpServerConfig().asReadOnly();
         serverConnection = initChannel(channel, httpExecutionContext, config,
-                new TcpServerChannelInitializer(config.tcpConfig()) {
-                    @Override
-                    public void init(final Channel channel) {
-                        super.init(channel);
-                        channel.pipeline().addLast(new ChannelInboundHandlerAdapter() {
-                            @Override
-                            public void userEventTriggered(final ChannelHandlerContext ctx,
-                                                           final Object evt) throws Exception {
-                                if (evt == CloseHandler.ProtocolPayloadEndEvent.OUTBOUND) {
-                                    // Mute payload boundary events for this test with repeated read
-                                    return;
-                                }
-                                super.userEventTriggered(ctx, evt);
-                            }
-                        });
-                    }
-                }, service, true,
+                new TcpServerChannelInitializer(config.tcpConfig()), service, true,
                 UNSUPPORTED_PROTOCOL_CLOSE_HANDLER)
                 .toFuture().get();
         serverConnection.process(true);
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
@@ -28,8 +28,11 @@ import io.servicetalk.http.api.StreamingHttpService;
 import io.servicetalk.http.netty.NettyHttpServer.NettyHttpServerConnection;
 import io.servicetalk.tcp.netty.internal.TcpServerChannelInitializer;
 import io.servicetalk.transport.api.IoExecutor;
+import io.servicetalk.transport.netty.internal.CloseHandler;
 
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelOutboundHandlerAdapter;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.embedded.EmbeddedChannel;
@@ -43,6 +46,7 @@ import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.BlockingQueue;
@@ -66,7 +70,6 @@ import static io.servicetalk.http.api.StreamingHttpRequests.newTransportRequest;
 import static io.servicetalk.http.netty.NettyHttpServer.initChannel;
 import static io.servicetalk.transport.netty.NettyIoExecutors.createIoExecutor;
 import static io.servicetalk.transport.netty.internal.CloseHandler.UNSUPPORTED_PROTOCOL_CLOSE_HANDLER;
-import static java.util.Arrays.asList;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasSize;
 
@@ -87,7 +90,17 @@ public class FlushStrategyOnServerTest {
     public final Timeout timeout = new ServiceTalkTestTimeout();
     private final HttpHeadersFactory headersFactory;
 
-    public FlushStrategyOnServerTest(final HttpExecutionStrategy executionStrategy) throws Exception {
+    private enum Param {
+        NO_OFFLOAD(noOffloadsStrategy()),
+        DEFAULT(defaultStrategy()),
+        OFFLOAD_ALL(customStrategyBuilder().offloadAll().build());
+        private final HttpExecutionStrategy executionStrategy;
+        Param(HttpExecutionStrategy executionStrategy) {
+            this.executionStrategy = executionStrategy;
+        }
+    }
+
+    public FlushStrategyOnServerTest(final Param param) throws Exception {
         writeEvents = new LinkedBlockingQueue<>();
         channel = new EmbeddedChannel(new ChannelOutboundHandlerAdapter() {
             @Override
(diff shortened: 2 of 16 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -459,6 +459,21 @@ public FlushStrategy defaultFlushStrategy() {
...
+         * This helps coordinate state between the writer and the codec/transport. Even if the writing source may not
+         * have completed from a Reactive Streams perspective, it means that as far as the codec and transport is
+         * concerned it considers the write is complete and no further writes can happen. Closing of the transport after
+         * receiving this event should not be considered a write error since the final buffer was written and
+         * potentially flushed. A failed write promise after observing this event however indicates the end of payload
+         * marker may not have been written correctly and should result in an error.
+         * <p>
+         * This is event is expected be called from the eventloop.
+         */
+        void protocolPayloadComplete();
```
NiteshKant: I think we can leverage pre-existing concepts here instead of introducing `protcol` and `payload` knowledge inside `WritableListener`. What we are after here is essentially "graceful closure" which in this case means wait for enqueued writes to finish before closing the listener.

I would suggest renaming this method to `closeGracefully()` and renaming existing `channelClosed(Throwable)` to `close(Throwable)` so that we clearly can define behaviors for abrupt and graceful closures without introducing the "protocol concept" here. If you get around unifying `channelClosedOutbound()` and `channelClosed()` as mentioned above then we can also remove `channelClosedOutbound()`.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CloseHandler.java
```diff
@@ -260,6 +267,11 @@ public void protocolPayloadBeginOutbound(final ChannelHandlerContext ctx) {
 
         @Override
         public void protocolPayloadEndOutbound(final ChannelHandlerContext ctx) {
+            ctx.pipeline().fireUserEventTriggered(ProtocolPayloadEndEvent.OUTBOUND);
```
NiteshKant: I would avoid adding supposedly protocol specific logic like this to `UnsupportedProtocolHandler`. IIUC, this is to support H2 so lets create a `CloseHandler` implementation for H2 and for now we can simply implement this method.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
```diff
@@ -112,19 +125,36 @@ public void flush(final ChannelHandlerContext ctx) {
...
+                new TcpServerChannelInitializer(config.tcpConfig()) {
+                    @Override
+                    public void init(final Channel channel) {
+                        super.init(channel);
+                        channel.pipeline().addLast(new ChannelInboundHandlerAdapter() {
+                            @Override
+                            public void userEventTriggered(final ChannelHandlerContext ctx,
+                                                           final Object evt) throws Exception {
+                                if (evt == CloseHandler.ProtocolPayloadEndEvent.OUTBOUND) {
+                                    // Mute payload boundary events for this test with repeated read
```
NiteshKant: Can you explain the motivation for this change?

`ProtocolPayloadEndEvent.OUTBOUND` is for write but the comment here suggests it is due to "repeated reads".

In case you end up removing this change, we can make  the class `ProtocolPayloadEndEvent` package-private.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
```diff
@@ -112,19 +125,36 @@ public void flush(final ChannelHandlerContext ctx) {
...
                 UNSUPPORTED_PROTOCOL_CLOSE_HANDLER)
                 .toFuture().get();
         serverConnection.process(true);
         headersFactory = DefaultHttpHeadersFactory.INSTANCE;
     }
 
     @Parameters(name = "{index}: strategy = {0}")
-    public static Collection<HttpExecutionStrategy> data() {
-        return asList(noOffloadsStrategy(), defaultStrategy(), customStrategyBuilder().offloadAll().build());
+    public static Param[][] data() {
```
NiteshKant: Is the use of `enum` here just a personal choice or there is a difference between the two approaches?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
```diff
@@ -56,15 +57,17 @@
...
+                                CloseHandler closeHandler) {
+        super(headersFactory, closeHandler);
         this.scheme = sslEnabled ? HttpScheme.HTTPS : HttpScheme.HTTP;
         this.allocator = allocator;
     }
 
     @Override
     public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
         if (msg instanceof HttpRequestMetaData) {
+            closeHandler.protocolPayloadBeginOutbound(ctx);
```
NiteshKant: We are using `UnsupportedProtocolHandler` which does not implement these callbacks so we are not really verifying whether these callbacks are correct. Should we hold off adding calls to `CloseHandler` other than `protocolPayloadEndOutbound()` till we need them?

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/DefaultNettyConnectionTest.java
```diff
@@ -500,4 +523,34 @@ public void testChannelInactiveWithNoSubscriberIsQueued() throws Exception {
...
+        Buffer hello1 = newBuffer("Hello1");
+        publisher.onNext(hello1);
+        publisher.onNext(TRAILER);
+
+        pollChannelAndVerifyWrites("Hello1", TRAILER_MSG);
+
+        channel.pipeline().fireChannelInactive();
+        publisher.onComplete();
+
+        writeListener.verifyCompletion();
```
NiteshKant: Should we add `channel.close()` in this test?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
```diff
@@ -163,6 +163,13 @@ public void protocolPayloadBeginOutbound(final ChannelHandlerContext ctx) {
 
     @Override
     public void protocolPayloadEndOutbound(final ChannelHandlerContext ctx) {
+        if (isClient || has(state, CLOSING)) {
```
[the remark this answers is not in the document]
jayv: Unless I misunderstood our prior discussing, I assumed you suggested implementing it in `RequestResponseCloseHandler` as opposed to having a more elaborate state machine inside `WriteStreamSubscriber`.  What was your expectation?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -189,6 +189,11 @@ public void channelWritable() {
         requestMoreIfRequired(subscription);
     }
 
+    @Override
+    public void protocolPayloadComplete() {
+        onComplete();
+    }
+
     @Override
     public void channelClosedOutbound() {
```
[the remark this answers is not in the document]
jayv: It didn't seem necessary to solve the problem at hand, but I can look into it.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteSingleSubscriber.java
```diff
@@ -73,6 +73,11 @@ public void channelWritable() {
         // No op.
     }
 
+    @Override
+    public void protocolPayloadComplete() {
+        // No op.
```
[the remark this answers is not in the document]
jayv: Made it noop because it doesn't seem relevant to Single and not sure we should complicate the state machine for this.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -189,6 +189,11 @@ public void channelWritable() {
         requestMoreIfRequired(subscription);
     }
 
+    @Override
+    public void protocolPayloadComplete() {
+        onComplete();
```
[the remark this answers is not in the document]
jayv: IIUC `WriteStreamSubscriber` already takes care of this by not requesting more data via `isWritable()` ?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
jayv: Not personal choice, this improves tooling (IDE/gradle) support. With the existing approach you can't run/debug individual test-cases due to unstable name.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
jayv: The repeated read, single perpetual write mechanism is root cause for some of the items you had remarks on. If the protocol event were to be propagated here the `WriteStreamSubscriber` would be terminated which effectively stops requesting data from the source. Some tests send multiple requests on the same connection. 

It's the same reason we conditionally emit the event from the `RequestResponseCloseHandler` (cfr. `isClient` mentioned above) and the cause for piling up of listeners on server write futures until channel close.

If I were to split out an H2 variant from the `UnsupportedProtocolCloseHandler`, this would go away.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CloseHandler.java
jayv: I assumed all protocols would need to provide some notion of payload completion for `WriteStreamSubscriber` to work properly, so it seemed reasonable to provide it here.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
jayv: I wanted to make it more consistent with H1, but I'm OK with leaving them out.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
jayv: OK I'll try

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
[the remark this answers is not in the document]
NiteshKant: I am not questioning the changes to `RequestResponseCloseHandler` but why do we need the special case for `isClient`. Is `has(state, CLOSING)` sufficient?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteSingleSubscriber.java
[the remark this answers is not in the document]
NiteshKant: If we are adding a generic method to the `WritableListener` we should implement it in a way that makes sense for the contract. In this case it does seem like we would want to emit an error if we get `protocolPayloadComplete()` without `onSuccess()`, it will make it easier in the future to reason about the code.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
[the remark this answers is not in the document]
NiteshKant: `isWritable()` only checks for `!hasFlag(CHANNEL_CLOSED) && !hasFlag(SUBSCRIBER_TERMINATED);` so we would have to also check for `SOURCE_TERMINATED`.

Additionally, not requesting more data does not necessarily mean we will not get any more data if we are synthetically generating `onComplete()` as we may have already requested more than what we have got.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
NiteshKant: >> If I were to split out an H2 variant from the UnsupportedProtocolCloseHandler, this would go away.

Ok I think we should split out an H2 variant and not emit the event from `UnsupportedProtocolCloseHandler `.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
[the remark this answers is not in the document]
jayv: I feel like there shouldn't be any condition at all, not `isClient()` nor `has(CLOSING)`. `end of payload` should mean what it is and emit every time the codec sees one. 
But if we did that, then we'd get into the issue down the line with H1 repeated read/single write reusing a terminated `WriteStreamSubscriber`. So we have to suppress this event before it reaches `WSS` in the case of H1 server; for H2 it doesn't matter.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
NiteshKant: ok, great. thanks

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
NiteshKant: Ok Ya either we validate that we are doing the rite thing or leave it out. I would prefer leaving them out.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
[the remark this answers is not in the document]
NiteshKant: I understand why `has(CLOSING)` is required and I think it is reasonable as this event is only valuable to determine if we expect any other write on this connection. Since `CloseHandler` has all the other information (outbound closing) it makes sense to keep this knowledge inside `CloseHandler`.

I am not clear about why we need `isClient` check here, can you explain?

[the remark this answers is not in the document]
jayv: The discrepancy is the client `WriteStreamSubscriber` lifetime is tied to the request where on the server side its lifetime is aligned with the connection. Hence to be correct the client needs to trigger this event for every write, no matter if it'll be the last write on the connection. 
In practice it doesn't matter because close handler does proper sequencing and attaches listeners to the write promise so we can't ever see a premature channel closed event.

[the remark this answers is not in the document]
jayv: Also `has(CLOSING)` is tied to the connection, not the lifetime of `WSS` which differs between client and server.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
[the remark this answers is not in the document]
jayv: We have the same check in `doWrite()` actually so, adding `SOURCE_TERMINATED` should be sufficient then?

jayv:
@NiteshKant tried to address most/all concerns except for open question on `isClient`

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -307,28 +306,13 @@ void sourceTerminated(@Nullable Throwable cause) {
 
         void channelClosed(Throwable cause) {
             assert eventLoop.inEventLoop();
-            if (hasFlag(CHANNEL_CLOSED) || hasFlag(SUBSCRIBER_TERMINATED)) {
+            if (hasAnyFlags(CHANNEL_CLOSED, SUBSCRIBER_TERMINATED)) {
                 setFlag(CHANNEL_CLOSED);
                 return;
             }
             tryFailure(!written ? new AbortedFirstWrite(cause) : cause);
```
NiteshKant: If there are pending writes, we should not prematurely terminate the promise but wait for the promise results which will fail if the channel closed before writes were done. The following would be reasonable to do:

```java
void channelClosed(Throwable cause) {
    if (hasFlag(CHANNEL_CLOSED)) {
        return;
    }
    if (hasFlag(SUBSCRIBER_TERMINATED)) {
        setFlag(CHANNEL_CLOSED);
        // We have already terminated the subscriber (all writes have finished (one has failed)) then we
        // just close the channel now.
        closeHandler.closeChannelOutbound(channel);
    } else if (activeWrites > 0) {
        // Writes are pending, we will close the channel once writes are done.
        setFlag(CLOSE_OUTBOUND_ON_SUBSCRIBER_TERMINATION);
    } else {
        setFlag(CHANNEL_CLOSED);
        // subscriber has not terminated, no writes are pending and channel has closed so terminate the
        // subscriber with a failure.
        tryFailure(!written ? new AbortedFirstWrite(cause) : cause);
    }
}
```

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -391,7 +375,7 @@ private boolean setFailure0(Throwable cause) {
         }
 
         private void terminateSubscriber(@Nullable Throwable cause) {
-            if (hasFlag(CLOSE_OUTBOUND_ON_SUBSCRIBER_TERMINATION) || cause != null) {
```
NiteshKant: This will be reverted if we update the `channelClosed()` method.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -307,28 +306,13 @@ void sourceTerminated(@Nullable Throwable cause) {
 
         void channelClosed(Throwable cause) {
```
NiteshKant: Maybe just call this method `close()` now?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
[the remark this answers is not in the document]
NiteshKant: Ok ya makese sense, thanks for the explanation.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CloseHandler.java
```diff
@@ -262,6 +270,69 @@ public void protocolPayloadBeginOutbound(final ChannelHandlerContext ctx) {
...
+        @Override
+        public void protocolClosingInbound(final ChannelHandlerContext ctx) {
+        }
+
+        @Override
+        public void protocolClosingOutbound(final ChannelHandlerContext ctx) {
+        }
+    }
+
+    private static final class H2ProtocolHandler extends CloseHandler {
```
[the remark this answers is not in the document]
jayv: No can do, we deliberately made some methods pkg private on `CloseHandler` so it can only be implemented in this pkg.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteSingleSubscriber.java
```diff
@@ -74,21 +82,28 @@ public void channelWritable() {
     }
 
     @Override
-    public void channelClosed(Throwable closedException) {
+    public void closeGracefully() {
+        if (terminated == UNSET) {
```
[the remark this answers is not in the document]
jayv: OK I was under assumption that it would come back as a close-with-exception, but we can do it this way too.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CloseHandler.java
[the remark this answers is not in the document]
NiteshKant: Got it, thanks

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -459,6 +458,16 @@ public FlushStrategy defaultFlushStrategy() {
          */
         void channelWritable();
 
+        /**
+         * Close the channel after the pending writes complete.
+         *
+         * <p>
+         * Calling {@link #close(Throwable)} after {@link #closeGracefully()} will be ignored.
+         * <p>
+         * This is event is expected be called from the eventloop.
```
NiteshKant: `is event is` => `event is`

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteSingleSubscriber.java
```diff
@@ -51,21 +58,29 @@ public void onSubscribe(Cancellable cancellable) {
...
+            channel.writeAndFlush(result).addListener(future -> {
+                Throwable cause = future.cause();
+                if (cause == null) {
+                    notifyComplete();
+                } else {
+                    notifyError(cause);
+                }
+            });
+        } else {
+            LOGGER.error("Failed to write data after permature close of the WriteListener.");
```
NiteshKant: Also log the `result`?

```suggestion
            LOGGER.error("Ignoring write {} as the listener is already closed.", result);
```

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteSingleSubscriber.java
```diff
@@ -51,21 +58,29 @@ public void onSubscribe(Cancellable cancellable) {
...
+        }
     }
 
     @Override
     public void onError(Throwable t) {
-        notifyError(t);
+        if (terminatedUpdater.compareAndSet(this, AWAITING_RESULT, TERMINATED)) {
+            notifyError(t);
+        } else {
+            LOGGER.error("Failed to fail subscriber after permature close of the WriteListener.");
```
NiteshKant:
```suggestion
            LOGGER.error("Ignoring emitted error as the listener is already closed.", t);
```

PR #922, opened 2020-01-17
Update dependencies to the latest versions

idelpivnitskiy:


servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent.java
```diff
@@ -322,7 +322,7 @@ private Queues() {
         }
 
         static <T> Queue<T> newUnboundedLinkedMpscQueue() {
-            return USE_UNSAFE_QUEUES ? MpscLinkedQueue.newMpscLinkedQueue()
+            return USE_UNSAFE_QUEUES ? new MpscLinkedQueue<>()
```
NiteshKant: Was this an API change in jctools?

Scottmitch: jctools update was a major version bump (e.g. 2.x -> 3.x) so API breaking is fair. However the package name and maven coordinates didn't change so this may result in collisions at runtime during the transition period... /cc @nitsanw

idelpivnitskiy: yep, they had minor API changes: https://github.com/JCTools/JCTools/blob/master/RELEASE-NOTES.md#300
@Scottmitch @NiteshKant will you suggest to not update for now? Netty 4.1.45 depends on jctools 2.x, the next netty version will use 3.0.0. We can update it later to make it consistent with netty: https://github.com/netty/netty/commit/8410b083b551d78d7c03081cd49b48987176a4fe

idelpivnitskiy: however, there were no api changes made in netty, so we should be good either way

NiteshKant: AFAIR netty shades jctools so we should be fine. Lets update

PR #929, opened 2020-01-29
Add trailers after gRPC `BlockingStreamingRoute` response

idelpivnitskiy:
Motivation:

gRPC protocol requires a server to send trailers after the response
payload body to notify a client about the status of this response
processing. We add trailers for all service APIs except
`BlockingStreamingRoute` when service responds successfully.

Modifications:

- Add trailers with `grpc-status: 0` if `BlockingStreamingRoute`
responds without errors;

Result:

Server sends a complete response with trailers from
`BlockingStreamingRoute`.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
@@ -71,6 +71,7 @@ import static io.servicetalk.grpc.api.GrpcUtils.newErrorResponse;
 import static io.servicetalk.grpc.api.GrpcUtils.newResponse;
 import static io.servicetalk.grpc.api.GrpcUtils.readGrpcMessageEncoding;
 import static io.servicetalk.grpc.api.GrpcUtils.setStatus;
+import static io.servicetalk.grpc.api.GrpcUtils.setStatusOk;
 import static io.servicetalk.http.api.HttpApiConversions.toStreamingHttpService;
 import static io.servicetalk.http.api.HttpExecutionStrategies.defaultStrategy;
 import static io.servicetalk.http.api.HttpRequestMethod.POST;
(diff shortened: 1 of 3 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -444,8 +445,12 @@ public void handle(final HttpServiceContext ctx, final BlockingStreamingHttpRequ
                             final DefaultGrpcPayloadWriter<Resp> grpcPayloadWriter =
                                     new DefaultGrpcPayloadWriter<>(response.sendMetaData(serializer));
                             try {
+                                // Set status OK before invoking handle methods because users can close PayloadWriter
+                                final HttpPayloadWriter<Resp> payloadWriter = grpcPayloadWriter.payloadWriter();
+                                setStatusOk(payloadWriter.trailers(), ctx.executionContext().bufferAllocator());
```
NiteshKant: Did you consider setting a status in `grpcPayloadWriter.close()` if it is not set?

NiteshKant:
@servicetalk-bot test this please.

PRB does not seem to have picked this up

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
idelpivnitskiy: Yes, I considered that approach, but then we have to store the ref for `BufferAllocator` inside `DefaultGrpcPayloadWriter` and do a look up for the header before setting `OK`.
Because `OK` is the most common case and we do not expose trailers in `handle` method (users can't change them), this approach is good enough.

NiteshKant: >> this approach is good enough.

Agreed, thanks for the explanation

idelpivnitskiy:

PR #932, opened 2020-02-04
gRPC filters override configured execution strategy for the route

idelpivnitskiy:
Motivation:

When users append a service filter using `ServiceFactory`, it
reconstructs the `GrpcService` interface from registered routes,
wraps it with the filter, and re-registers the service again as
independent routes. This logic does not preserve the execution
strategy that was used to register the original route and it
uses the default strategy for a route.

Modifications:

- Save the route execution strategy if it was configured for the
original route;
- Use the saved execution strategy when we re-register filtered
routes;
- Add tests to verify the execution strategy configuration when
users apply a filter;
- Add tests to verify grpc router does not extract execution strategy
annotations from filters;
- Close `GrpcPayloadWriter` in the conversion layer from
`BlockingStreamingRoute` to `StreamingRoute`, because users of
`BlockingRequestStreamingRoute` does not have access to it;

Result:

gRPC filters do not override configured execution strategy for
the route.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
@@ -165,22 +165,26 @@ final class GrpcRouter {
         private final Map<String, RouteProvider> streamingRoutes;
         private final Map<String, RouteProvider> blockingRoutes;
         private final Map<String, RouteProvider> blockingStreamingRoutes;
+        private final Map<String, GrpcExecutionStrategy> executionStrategies;
 
         Builder() {
             routes = new HashMap<>();
             streamingRoutes = new HashMap<>();
             blockingRoutes = new HashMap<>();
             blockingStreamingRoutes = new HashMap<>();
+            executionStrategies = new HashMap<>();
         }
 
         Builder(final Map<String, RouteProvider> routes,
                 final Map<String, RouteProvider> streamingRoutes,
                 final Map<String, RouteProvider> blockingRoutes,
-                final Map<String, RouteProvider> blockingStreamingRoutes) {
+                final Map<String, RouteProvider> blockingStreamingRoutes,
+                final Map<String, GrpcExecutionStrategy> executionStrategies) {
             this.routes = routes;
             this.streamingRoutes = streamingRoutes;
             this.blockingRoutes = blockingRoutes;
             this.blockingStreamingRoutes = blockingStreamingRoutes;
+            this.executionStrategies = executionStrategies;
         }
 
         RouteProviders drainRoutes() {
@@ -196,18 +200,25 @@ final class GrpcRouter {
             return new RouteProviders(allRoutes);
         }
 
+        GrpcExecutionStrategy alreadyRegisteredExecutionStrategy(final String path,
+                                                                 final GrpcExecutionStrategy defaultValue) {
+            return executionStrategies.getOrDefault(path, defaultValue);
+        }
+
         static GrpcRouter.Builder merge(final GrpcRouter.Builder... builders) {
             final Map<String, RouteProvider> routes = new HashMap<>();
             final Map<String, RouteProvider> streamingRoutes = new HashMap<>();
             final Map<String, RouteProvider> blockingRoutes = new HashMap<>();
             final Map<String, RouteProvider> blockingStreamingRoutes = new HashMap<>();
+            final Map<String, GrpcExecutionStrategy> executionStrategies = new HashMap<>();
             for (Builder builder : builders) {
                 mergeRoutes(routes, builder.routes);
                 mergeRoutes(streamingRoutes, builder.streamingRoutes);
                 mergeRoutes(blockingRoutes, builder.blockingRoutes);
                 mergeRoutes(blockingStreamingRoutes, builder.blockingStreamingRoutes);
+                executionStrategies.putAll(executionStrategies);
             }
-            return new Builder(routes, streamingRoutes, blockingRoutes, blockingStreamingRoutes);
+            return new Builder(routes, streamingRoutes, blockingRoutes, blockingStreamingRoutes, executionStrategies);
         }
 
         private static void mergeRoutes(final Map<String, RouteProvider> first,
@@ -263,6 +274,7 @@ final class GrpcRouter {
                     // aggregated and streaming. Therefore, verify that there is no blocking-aggregated route registered
                     // for the same path:
                     path, blockingRoutes);
+            executionStrategies.put(path, executionStrategy);
             return this;
         }
 
@@ -320,6 +332,7 @@ final class GrpcRouter {
                     // aggregated and streaming. Therefore, verify that there is no blocking-streaming route registered
                     // for the same path:
                     path, blockingStreamingRoutes);
+            executionStrategies.put(path, executionStrategy);
             return this;
         }
 
@@ -434,6 +447,7 @@ final class GrpcRouter {
                     // aggregated and streaming. Therefore, verify that there is no async-aggregated route registered
                     // for the same path:
                     path, routes);
+            executionStrategies.put(path, executionStrategy);
             return this;
         }
 
@@ -497,6 +511,7 @@ final class GrpcRouter {
                     // aggregated and streaming. Therefore, verify that there is no async-streaming route registered
                     // for the same path:
                     path, streamingRoutes);
+            executionStrategies.put(path, executionStrategy);
             return this;
         }
 
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
@@ -200,8 +200,7 @@ final class GrpcRouter {
             return new RouteProviders(allRoutes);
         }
 
-        GrpcExecutionStrategy alreadyRegisteredExecutionStrategy(final String path,
-                                                                 final GrpcExecutionStrategy defaultValue) {
+        GrpcExecutionStrategy executionStrategyFor(final String path, final GrpcExecutionStrategy defaultValue) {
             return executionStrategies.getOrDefault(path, defaultValue);
         }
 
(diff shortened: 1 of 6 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
```diff
@@ -173,9 +187,18 @@ protected GrpcService newServiceFromRoutes(final AllGrpcRoutes routes) {
...
+        // Check if we already have a computed GrpcExecutionStrategy for this path. This may happen when we re-register
+        // filtered routes:
+        final GrpcExecutionStrategy saved = executionStrategies.getOrDefault(path, NULL);
+        if (saved != NULL) {
+            return saved;
+        }
+        final GrpcExecutionStrategy computed = getAndValidateRouteExecutionStrategyAnnotationIfPresent(method, clazz,
+                strategyFactory, errors, noOffloadsStrategy());
+        // Save computed value even if it's null to prevent analysing annotations on a service filter class
+        executionStrategies.put(path, computed);
```
idelpivnitskiy: @NiteshKant should we account for users that may try-catch routes registration? If they register the same path 2+ times with different route implementations, we will override the strategy here, but later we throw an exception saying that this path was already registered. If users try-catch this exception, internal state will be modified. I can move this `executionStrategies.put` after we register a route if necessary.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
```diff
@@ -173,9 +187,18 @@ protected GrpcService newServiceFromRoutes(final AllGrpcRoutes routes) {
     }
 
     @Nullable
-    private GrpcExecutionStrategy executionStrategy(final Method method, final Class<?> clazz) {
-        return getAndValidateRouteExecutionStrategyAnnotationIfPresent(method, clazz, strategyFactory, errors,
-                noOffloadsStrategy());
+    private GrpcExecutionStrategy executionStrategy(final String path, final Method method, final Class<?> clazz) {
+        // Check if we already have a computed GrpcExecutionStrategy for this path. This may happen when we re-register
+        // filtered routes:
+        final GrpcExecutionStrategy saved = executionStrategies.getOrDefault(path, NULL);
```
NiteshKant: Just use `get()` and eliminate defining `NULL` as default?

```java
final GrpcExecutionStrategy saved = executionStrategies.get(path);
if (saved != null) {
    return saved;
}
```

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
NiteshKant: >> but later we throw an exception saying that this path was already registered.

Did you mean that we will ignore the strategy for a subsequent registration? We are not using `executionStrategies` to check for duplicate registration, are we?

I think we should account for cases when a registration exception is caught and retried unless we have a strong reason to believe that is invalid. AFAIR, this is required only for disallowing filters to specify the strategy annotation, rite? I have not thought of that case as to whether it is good or bad, so if we need special code to disallow it, lets refrain from doing that for now and have "behavior when strategy annotation is defined on filters is undefined" as our answer.

idelpivnitskiy: I was thinking about a use-case like this:
```java
ServiceFactory.Builder builder = new ServiceFactory.Builder();
builder.testBlocking(noOffloadsStrategy(), serviceImpl);
try {
    builder.testBlocking(anotherStrategy, anotherServiceImpl);
} catch (IllegalStateException e) {
    // ignore that this route was already registered
}
```
Don't think we need to verify duplication for the `executionStrategies` map, because we already do that for routes. An additional check will be useless.
In my first impl I stored execution strategy before I verified routes duplication. That's why in the example above users will be in the state when they have `serviceImpl` with `anotherStrategy`.
Now I moved `executionStrategies.put(...)` after we verify routes duplication. So, nothing will corrupt the internal state.

> AFAIR, this is required only for disallowing filters to specify the strategy annotation, rite? I have not thought of that case as to whether it is good or bad.

Right. As we discussed before, ST policy is that if you configure an execution strategy for the route, ST will use it regardless of filters/influencers/etc.
The case here is that grpc filter implements the same service interface. Therefore, users can accidentally add `@RouteExecutionStrategy` annotation on the filter. I think we should ignore it.

> so if we need special code to disallow it, lets refrain from doing that for now and have "behavior when strategy annotation is defined on filters is undefined" as our answer.

The code to disallow annotation inspection for filters is trivial. It's just `getOrDefault` vs `get` on the map to distinguish `key-null` entry vs entry for this key does not exist.

Defined behavior is always better than undefined. I've added this test to demonstrate the problem: https://github.com/apple/servicetalk/pull/932/files#diff-4fbfd0eed0161987f94f03bc6d9e3203R262-R279

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
idelpivnitskiy: Let's continue this discussion here: https://github.com/apple/servicetalk/pull/932#discussion_r374957744

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -196,18 +200,25 @@ RouteProviders drainRoutes() {
...
             final Map<String, RouteProvider> streamingRoutes = new HashMap<>();
             final Map<String, RouteProvider> blockingRoutes = new HashMap<>();
             final Map<String, RouteProvider> blockingStreamingRoutes = new HashMap<>();
+            final Map<String, GrpcExecutionStrategy> executionStrategies = new HashMap<>();
             for (Builder builder : builders) {
                 mergeRoutes(routes, builder.routes);
                 mergeRoutes(streamingRoutes, builder.streamingRoutes);
                 mergeRoutes(blockingRoutes, builder.blockingRoutes);
                 mergeRoutes(blockingStreamingRoutes, builder.blockingStreamingRoutes);
+                executionStrategies.putAll(executionStrategies);
```
NiteshKant:
```suggestion
                executionStrategies.putAll(builder.executionStrategies);
```

If it is simple, add a test to verify, else it is OK.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -196,18 +200,25 @@ RouteProviders drainRoutes() {
             return new RouteProviders(allRoutes);
         }
 
+        GrpcExecutionStrategy alreadyRegisteredExecutionStrategy(final String path,
```
NiteshKant: Rename `alreadyRegisteredExecutionStrategy` => `executionStrategyFor()`

It does not seem like "already registered" is providing any additional value here.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
```diff
@@ -173,7 +177,15 @@ protected GrpcService newServiceFromRoutes(final AllGrpcRoutes routes) {
     }
 
     @Nullable
-    private GrpcExecutionStrategy executionStrategy(final Method method, final Class<?> clazz) {
+    private GrpcExecutionStrategy executionStrategy(final String path, final Method method, final Class<?> clazz) {
+        // Check if we already have a computed GrpcExecutionStrategy for this path. This happens when we re-register
+        // filtered routes and have to use the original execution strategy for the route instead of analysing
+        // annotations on a service-filter class. Because previously registered strategy could be null, we use NULL
```
NiteshKant: `null` executionStrategy has no semantic meaning from a user point of view. We use it as a short-cut internally to avoid method overloads. If we only add non-null strategy for a path in `GrpcRouter` methods, can we simplify this as `null` value meaning no strategy is defined?

idelpivnitskiy: This `null` is not visible for users. `null` will be here only when users use `ServiceFactory.Builder` methods that do not take execution strategy and they do not add annotations on the service. For a `null` use-case we define execution strategy based on the method API: https://github.com/apple/servicetalk/blob/412222cb49cb54101f8d8afc79c52a5c3f1f6e6a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java#L259

NiteshKant: I understand that. I am questioning the need for this awkward `NULL` marker used here which is trying to distinguish between whether the value stored in the `executionStrategies` `Map` for a `path` is `null` or there is no mapping for that `path`.

If we do not put a `null` value in the `executionStrategies` `Map` then we do not need this `NULL` marker here, rite?

idelpivnitskiy: The stored `path<->null` pair helps to understand that we already processed annotations for this route and an execution strategy for this route was `null` (therefore, we have to infer an execution strategy based on the route API).
If we won't store `path<->null` pair, we don't have a way to distinguish between "we already know that the execution strategy for this route is null" (this is when we re-register route with a filter) and "we don't know the execution strategy for this route and have to process annotations" (first-time registration of the route).

idelpivnitskiy: We need to know this difference because the same code path is used for initial route registration and re-registration of the filtered route.

idelpivnitskiy: @NiteshKant if the main concern is the awkward `NULL` marker, I've pushed an alternative impl using a `boolean` flag: https://github.com/apple/servicetalk/pull/932/commits/fd575fdf701a6da07743d95068edec6df9525b6f
LMK if this approach looks better.

NiteshKant: IMHO the `boolean` approach is slightly worse than what you had earlier as it stores state(`registerFilters`) assuming interactions from `GrpcServiceFactory#applyFilterToRoutes()`.

It seems we are working around code to support "filters should not have strategy annotation" scenario. Since you seem to be convinced that scenario needs to be validated, lets go with what you had before (`NULL` marker approach) which looks like a lesser evil.

PR #934, opened 2020-02-07
gRPC-client should fail responses without grpc-status code

idelpivnitskiy:
Motivation:

gRPC protocol requires server to send grpc-status code as part
of the response. It may come in headers (when there is no
payload body) or in trailers. Current implementation considers
responses without `grpc-status` as legit responses.

Modifications:

- Ensure that the response object contains `grpc-status` code;
- Tests to verify that client throws an exception when server
does not send `grpc-status`;

Result:

gRPC-client throws an exception when server does not send
`grpc-status`.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019-2020 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
@@ -193,8 +193,8 @@ final class GrpcUtils {
         final GrpcStatusCode statusCode = extractGrpcStatusCodeFromHeaders(headers);
         if (statusCode == null) {
             // This is a protocol violation as we expect to receive grpc-status.
-            throw new GrpcStatus(INTERNAL, null, "Response does not contain " + GRPC_STATUS_CODE_TRAILER + " header or trailer.")
-                    .asException();
+            throw new GrpcStatus(INTERNAL, null, "Response does not contain " + GRPC_STATUS_CODE_TRAILER +
+                    " header or trailer").asException();
         }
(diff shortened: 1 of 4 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -138,50 +150,56 @@ static void setStatus(final HttpHeaders trailers, final Throwable cause, final B
...
-                    throw ex;
-                }
-            } else if (!(o instanceof Buffer)) {
-                throw new IllegalArgumentException("Unexpected payload type: " + o.getClass());
-            }
-            return o;
-        }).filter(o -> !(o instanceof HttpHeaders)).map(o -> (Buffer) o));
+
+        response.transformRaw(ENSURE_GRPC_STATUS_RECEIVED);
+        return deserializer.deserialize(headers, response.payloadBodyAndTrailers()
```
idelpivnitskiy: This helps to catch the case when server does not send trailers at all. 
In this case `payloadBodyAndTrailers()` returns only `Buffer`s. `transformRaw` helps to create empty trailers and catch that there is no `grpc-status`.
I did not add a comment here, because the test will fail if this changes, but I can add it if it will be useful.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -138,50 +150,56 @@ static void setStatus(final HttpHeaders trailers, final Throwable cause, final B
...
+        // There was no grpc-status in the trailers, so it must be in headers.
+        ensureGrpcStatusReceived(response.headers());
+        return response.payloadBody(deserializer);
+    }
+
+    private static void ensureGrpcStatusReceived(final HttpHeaders headers) {
+        final GrpcStatusCode statusCode = extractGrpcStatusCodeFromHeaders(headers);
+        if (statusCode == null) {
+            // This is a protocol violation as we expect to receive grpc-status.
+            throw new GrpcStatus(INTERNAL, null, "Response does not contain grpc-status code")
```
NiteshKant:
```suggestion
            throw new GrpcStatus(INTERNAL, null, "Response does not contain " + GRPC_STATUS_CODE_TRAILER + " header or trailer.")
```

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
idelpivnitskiy: Good idea, I will keep it with `transformRaw` for now and we can optimize if it show ups as a problem later.

idelpivnitskiy:
Build failed due to some docker issue:
> ERROR: for servicetalk-java8-prb_runtime-setup_1  Cannot start service runtime-setup: network servicetalk-java8-prb_default not found

PR #936, opened 2020-02-12
`HttpObjectEncoder` should always allocate direct outbound buffers

idelpivnitskiy:
Motivation:

`HttpObjectEncoder` internally allocates buffers that will be
passed to the transport or `OPENSSL` handler, both require direct
memory for processing. We should explicitly allocate direct memory
in outbound handler to prevent from slowing down (by introducing
additional copying from heap to direct memory) when users specify
`io.netty.noPreferDirectFlag=true` system property.

Modifications:

- Explicitly allocate direct pooled buffers in `HttpObjectEncoder`;

Result:

Buffers allocated in `HttpObjectEncoder` will always use direct
pooled memory and won't be copied by transport or `OPENSSL` handler.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectEncoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectEncoder.java
@@ -125,7 +125,7 @@ abstract class HttpObjectEncoder<T extends HttpMetaData> extends ChannelOutbound
             // We prefer a direct allocation here because it is expected the resulted encoded Buffer will be written
             // to a socket. In order to do the write to the socket the memory typically needs to be allocated in direct
             // memory and will be copied to direct memory if not. Using a direct buffer will avoid the copy.
-            ByteBuf byteBuf = POOLED_ALLOCATOR.buffer((int) headersEncodedSizeAccumulator);
+            ByteBuf byteBuf = POOLED_ALLOCATOR.directBuffer((int) headersEncodedSizeAccumulator);
             Buffer stBuf = newBufferFrom(byteBuf);
 
             // Encode the message.
@@ -256,7 +256,7 @@ abstract class HttpObjectEncoder<T extends HttpMetaData> extends ChannelOutbound
                                              PromiseCombiner promiseCombiner) {
         if (contentLength > 0) {
             String lengthHex = toHexString(contentLength);
-            ByteBuf buf = POOLED_ALLOCATOR.buffer(lengthHex.length() + 2);
+            ByteBuf buf = POOLED_ALLOCATOR.directBuffer(lengthHex.length() + 2);
             buf.writeCharSequence(lengthHex, US_ASCII);
             writeShortBE(buf, CRLF_SHORT);
             promiseCombiner.add(ctx.write(buf));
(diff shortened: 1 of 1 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectEncoder.java
```diff
@@ -125,7 +125,7 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
             // We prefer a direct allocation here because it is expected the resulted encoded Buffer will be written
             // to a socket. In order to do the write to the socket the memory typically needs to be allocated in direct
             // memory and will be copied to direct memory if not. Using a direct buffer will avoid the copy.
-            ByteBuf byteBuf = POOLED_ALLOCATOR.buffer((int) headersEncodedSizeAccumulator);
+            ByteBuf byteBuf = POOLED_ALLOCATOR.directBuffer((int) headersEncodedSizeAccumulator);
```
idelpivnitskiy: Transport needs direct memory for the outgoing data and after this handler users do not touch buffers anymore. So, this helps to prevent additional copy in transport when users configure buffer allocator to prefer heap memory.

@normanmaurer as there any specific reason to use `buffer` method here instead of `directBuffer` or `ioBuffer` in similar handler in netty?
https://github.com/netty/netty/blob/fcf55fcf718af1e4cae85eb86cdb679094186aed/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java#L93

normanmaurer: usually we use `buffer` so the user has pretty much control over what type of buffer is allocated. And by default `buffer(...)` will return a direct buffer in netty anyway

idelpivnitskiy: Agreed, this is by default. But this "control" leads to overhead, because transport needs to copy to direct memory anyway. Not sure about netty, because users may have other handlers between transport and `HttpObjectEncoder` with their logic, but for ST I think it makes sense to always use direct buffers here.

normanmaurer:
@idelpivnitskiy yep... I wonder if we could do something smart in netty 5 and choose the buffertype based on where the handler is in the pipeline and what the next handler prefers

idelpivnitskiy:
@normanmaurer I think ability to configure buffer allocator for the handler should do the trick in netty 5.

idelpivnitskiy:
@normanmaurer on the other hand, propagation of the buffer type preference also sounds interesting 🤔

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectEncoder.java
NiteshKant: @idelpivnitskiy looks like @normanmaurer is saying that `buffer()` will anways return direct buffers. Is this change just to be explicit or you found an issue?

idelpivnitskiy: @NiteshKant he is saying that this is by default in netty, but the actual type depends on the configuration of the provided buffer allocator or `io.netty.noPreferDirectFlag=true` system property. This is outgoing path and transport anyway requires direct memory. So, to reduce overhead of copying from heap to direct memory at transport level, we should ignore what users prefer and always use direct buffers here. Their preference of heap memory makes sense only when they allocate buffers in code they control, but not in transport or this encoder.

NiteshKant: got it, thanks!

PR #937, opened 2020-02-12
Avoid potential memory leak in `ByteToMessageDecoder`

idelpivnitskiy:
Motivation:

`ByteToMessageDecoder` does not release the received
`ByteBuf` message when it copies data to the cumulation
buffer. It's not an issue at this moment, because we always
see unpooled buffers here (because `CopyByteBufHandler`
copies data to unpooled memory), but we should release
any way to prevent memory leaks if this assumption changes.

Modifications:

- Add `try-finally` block to release inbound `ByteBuf` after
we copied it to the cumulation buffer;

Result:

No memory leaks in `ByteToMessageDecoder`.


NiteshKant:

PR #940, opened 2020-02-17
H2 client should send `ENABLE_PUSH=false` setting for new connections

idelpivnitskiy:
Motivation:

To notify server-side that ST client currently does not support
server push, client should send `ENABLE_PUSH=false` setting when it
connects to the server.
If server sends a `PUSH_PROMISE` frame, client should send `GO_AWAY`
before closing the connection.

Modifications:

- Add `ENABLE_PUSH=false` and `MAX_CONCURRENT_STREAMS=0` to
the set of initial h2 settings;
- Send `GO_AWAY` frame with `PROTOCOL_ERROR` if client receives a
`PUSH_PROMISE` frame from server;

Result:

H2 client notifies server that it does not support server push and
sends `GO_AWAY` frame before closing the connection, if server sends
a push.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019-2020 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
```diff
@@ -72,7 +84,14 @@ private H2PushStreamHandler() {
...
         @Override
         public void channelRegistered(final ChannelHandlerContext ctx) {
-            ctx.close(); // push streams are not supported
+            // See SETTINGS_ENABLE_PUSH in https://tools.ietf.org/html/rfc7540#section-6.5.2
+            ctx.writeAndFlush(new DefaultHttp2GoAwayFrame(PROTOCOL_ERROR))
+                    .addListener((ChannelFutureListener) future -> {
+                if (!future.isSuccess()) {
+                    LOGGER.debug("Failed to send a GO_AWAY frame for received PUSH_PROMISE frame", future.cause());
+                }
+                ctx.close(); // push streams are not supported
```
NiteshKant: Netty should be closing the channel and log appropriately:

https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java#L905-L929

I do not think we need to add a listener here

PR #941, opened 2020-02-17
Use `payloadBodyAndTrailers()` to drain payload body

idelpivnitskiy:
Motivation:

In some cases we need to consume payload body and trailers.
We usually use `payloadBody().ignoreElements()` pattern.
However, since the request or response comes from the network,
it's more efficient to do `payloadBodyAndTrailers().ignoreElements()`
to avoid splitting trailers from payload body.

Modifications:

- Replace all `payloadBody().ignoreElements()` usages with
`payloadBodyAndTrailers().ignoreElements()`;
- Do the same in tests for consistency;
- Update `performance.adoc` to demonstrate `payloadBodyAndTrailers`
for the draining use-case;

Result:

Less overhead when we need to drain payload body.

### DIFF
--- a/servicetalk-http-router-jersey/src/testFixtures/java/io/servicetalk/http/router/jersey/CancellationTest.java
+++ b/servicetalk-http-router-jersey/src/testFixtures/java/io/servicetalk/http/router/jersey/CancellationTest.java
@@ -239,7 +239,7 @@ public class CancellationTest {
                         errorRef.compareAndSet(null, new NullPointerException("result == null not expected."));
                         cancelledLatch.countDown();
                     } else {
-                        result.payloadBody().ignoreElements().afterFinally(cancelledLatch::countDown).subscribe();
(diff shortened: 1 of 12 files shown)


servicetalk-http-router-jersey/src/testFixtures/java/io/servicetalk/http/router/jersey/CancellationTest.java
```diff
@@ -239,7 +239,7 @@ public void onSuccess(@Nullable final StreamingHttpResponse result) {
                         errorRef.compareAndSet(null, new NullPointerException("result == null not expected."));
                         cancelledLatch.countDown();
                     } else {
-                        result.payloadBody().ignoreElements().afterFinally(cancelledLatch::countDown).subscribe();
+                        result.payloadBodyAndTrailers().ignoreElements().afterFinally(cancelledLatch::countDown).subscribe();
```
NiteshKant: > 120

PR #943, opened 2020-02-20
Optimize `ProtoBufSerializationProvider` by minimizing copying

idelpivnitskiy:
Motivation:

Current implementation of `ProtoBufSerializationProvider` has a
few performance issues:
1. `ProtoDeserializer` may create an intermediate copy of the composite
buffers and then copy it again inside proto parser.
2. `DefaultSerializer.DEFAULT_SIZE_ESTIMATOR` inaccurately predicts
size for destination `Buffer`. For aggregated API we may use
`getSerializedSize()` to allocate enough bytes for destination
`Buffer`.
3. `MessageLite.writeTo(OutputStream)` creates an internal copy of
data that could be avoided if we create `CodedOutputStream` from
internal data storage of destination `Buffer`.

Modifications:

- Create optimized version of `CodedInputStream` for
`ProtoDeserializer` to do less copying during proto deserialization;
- Provide a number of `bytesEstimate` for serialization for aggregated
API;
- Create optimized version of `CodedOutputStream` for
`ProtoSerializer` to do less copying during proto serialization;

Results:

Less copying during protobuf serialization/deserialization leads to
improved throughput by 5-15% on the gRPC client and server when
they (de)serialize 16Kb payload body.

### DIFF
--- a/servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProvider.java
+++ b/servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019-2020 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProvider.java
+++ b/servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProvider.java
@@ -24,16 +24,20 @@ import io.servicetalk.serialization.api.StreamingDeserializer;
 import io.servicetalk.serialization.api.StreamingSerializer;
 import io.servicetalk.serialization.api.TypeHolder;
 
+import com.google.protobuf.CodedInputStream;
+import com.google.protobuf.CodedOutputStream;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.MessageLite;
 import com.google.protobuf.Parser;
 
 import java.io.IOException;
-import java.io.OutputStream;
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
 import javax.annotation.Nullable;
 
+import static com.google.protobuf.CodedOutputStream.newInstance;
+import static com.google.protobuf.UnsafeByteOperations.unsafeWrap;
 import static io.servicetalk.buffer.netty.BufferAllocators.DEFAULT_ALLOCATOR;
 import static java.lang.Math.max;
 import static java.util.Collections.emptyList;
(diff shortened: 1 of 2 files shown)


servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProvider.java
```diff
@@ -146,7 +150,22 @@ private static boolean isCompressed(Buffer buffer) throws SerializationException
...
+                                    lengthOfData));
+                        } else {
+                            // Aggregated payload body may consist of multiple Buffers. In this case,
+                            // CompositeBuffer.toNioBuffer(idx, length) may return a single ByteBuffer (when requested
+                            // length < components[0].length) or create a new ByteBuffer and copy multiple components
+                            // into it. Later, proto parser will copy data from this temporary ByteBuffer again.
+                            // To avoid unnecessary copying, we use newCodedInputStream(buffers, lengthOfData).
+                            final ByteBuffer[] buffers = toDeserialize.toNioBuffers(toDeserialize.readerIndex(),
+                                    lengthOfData);
+                            in = buffers.length == 1 ? CodedInputStream.newInstance(buffers[0]) :
```
NiteshKant: Is this check necessary since we will only be in this `else` block if `toDeserialize.nioBufferCount() > 1`?

servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProvider.java
```diff
@@ -186,6 +205,23 @@ private static boolean isCompressed(Buffer buffer) throws SerializationException
             }
         }
 
+        private static CodedInputStream newCodedInputStream(final ByteBuffer[] buffers, final int lengthOfData) {
+            // Because we allocated a new internal ByteBuffer that will never be mutated we may just wrap it and
+            // enable aliasing to avoid an extra copying inside parser for a deserialized message.
+            final CodedInputStream in = unsafeWrap(mergeByteBuffers(buffers, lengthOfData)).newCodedInput();
```
NiteshKant: Wondering why we can not use `ByteString.unsafeWrap()` and aliasing for `toDeserialize.nioBufferCount() == 1` case also? We are pretty sure that the buffer will not be mutated between creation of `CodedInputStream` and usage by the parser.

servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProviderBuilder.java
```diff
@@ -155,7 +157,7 @@ public GrpcSerializationProvider build() {
         @Override
         public Buffer serialize(final HttpHeaders headers, final T value, final BufferAllocator allocator) {
             addContentHeaders(headers);
-            return serializer.serialize(value, allocator);
+            return serializer.serialize(value, allocator, METADATA_SIZE + value.getSerializedSize());
```
NiteshKant: If the size is not memoized, `getSerializedSize()` will use reflection to calculate the size. Is that worth the size optimization here?

servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProvider.java
idelpivnitskiy: If `toDeserialize.nioBufferCount() == 1`, the `Buffer` will not create a new copy when we do `toDeserialize.toNioBuffer(...)`. Instead, it will return a `ByteBuffer` wrapper that shares the same internal array.
Because users have a way to access the original `Buffer` using http filters, we can not guarantee that they do not modify the internal array. Therefore, by default we need to be safe. However, we can provide a configuration option for `ProtoBufSerializationProviderBuilder` and `GrpcClientBuilder/GrpcServerBuilder` to do zero-copying deserialization when users sure that they do not touch the payload body in any other way.

servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProvider.java
idelpivnitskiy: Yes, this is described in the comment above. `toDeserialize.nioBufferCount()` does not account for a requested length. We may be in a situation when the `CompositeBuffer` has 2 or more buffers internally (and those buffers may contain multiple proto messages), but we request only a few bytes (single proto message) that could be read from the first internal buffer. In this case, `toDeserialize.toNioBuffers(..., length)` may return an array that has only the first internal `ByteBuffer`, not all the internal buffers.

servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProviderBuilder.java
idelpivnitskiy: We invoke `getSerializedSize()` anyway to fill in the metadata later. So, we do not pay the reflection cost twice. Without this, profiling shows that we frequently expand the buffer during `writeTo` because it can not fit the serialized proto.

PR #946, opened 2020-02-23
Disable netty's go away handling for H2.

NiteshKant:
__Motivation__

By default when an H2 channel is closed, netty sends go away frames as required by the spec. However, ServiceTalk provides separate close and graceful close modes for closing a connection and implements go away frame handling when graceful closure is requested by the user.
Currently, as we do not disable netty's default of go away handling, a force close also behaves like graceful closure (by sending go away frames) and we set the graceful closure timeout to `-1` for netty. This makes it such that if a request is pending (magnified for long lived streams) a client/server never closes.

__Modification__

Disable graceful closure for netty and add tests to verify that server and clients are eventually closed.

__Result__

H2 servers and clients can be closed even with pending requests.


NiteshKant:
@Scottmitch PTAL again

PR #947, opened 2020-02-25
gRPC `BlockingResponseStreamingRoute` expects only one request item

idelpivnitskiy:
Motivation:

gRPC `BlockingResponseStreamingRoute` is implemented as a conversion
based on `BlockingStreamingRoute`, but must ensure that there is
only one proto item in the request and fail if there are more.
Otherwise, it does not drain request `BlockingIterable` and may lead
to the leaked resources.

Modifications:

- Throw if `BlockingResponseStreamingRoute` sees the second item
in the request `BlockingIterable`;
- Throw if `BlockingResponseStreamingRoute` sees no items in the
request `BlockingIterable`;
- Make exception messages of `ResponseStreamingRoute` consistent
with `BlockingResponseStreamingRoute`;
- Add tests to verify that these cases work as expected and
blocking/async API behavior is consistent;

Result:

`BlockingResponseStreamingRoute` fails if a client sends more than
one proto items in the request.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
@@ -154,13 +154,13 @@ final class DefaultGrpcClientCallFactory implements GrpcClientCallFactory {
         final BlockingStreamingClientCall<Req, Resp> streamingClientCall =
                 newBlockingStreamingCall(serializationProvider, requestClass, responseClass);
         return (metadata, request) -> {
-            final BlockingIterator<Resp> iterator = streamingClientCall.request(metadata, request).iterator();
-            final Resp firstItem = iterator.next();
-            if (iterator.hasNext()) {
-                throw new IllegalArgumentException("Only a single item expected, but saw the second value: " +
-                        iterator.next());
+            try (BlockingIterator<Resp> iterator = streamingClientCall.request(metadata, request).iterator()) {
+                final Resp firstItem = requireNonNull(iterator.next(), "Response item is null");
+                if (iterator.hasNext()) {
+                    throw new IllegalStateException("Only a single response item is expected, but saw the second one");
+                }
+                return firstItem;
             }
-            return firstItem;
         };
     }
 
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
@@ -158,7 +158,7 @@ final class DefaultGrpcClientCallFactory implements GrpcClientCallFactory {
                 final Resp firstItem = requireNonNull(iterator.next(), "Response item is null");
                 if (iterator.hasNext()) {
                     iterator.next(); // Consume the next item to make sure it's not a TerminalNotification with an error
-                    throw new IllegalStateException("More than one response message received");
+                    throw new IllegalArgumentException("More than one response message received");
                 }
                 return firstItem;
             }
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
@@ -158,7 +158,7 @@ final class DefaultGrpcClientCallFactory implements GrpcClientCallFactory {
                 final Resp firstItem = requireNonNull(iterator.next(), "Response item is null");
                 if (iterator.hasNext()) {
                     iterator.next(); // Consume the next item to make sure it's not a TerminalNotification with an error
-                    throw new IllegalStateException("Only a single response item is expected, but saw the second one");
+                    throw new IllegalStateException("More than one response message received");
                 }
                 return firstItem;
             }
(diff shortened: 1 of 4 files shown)


servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/BlockingApiCorrectnessTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Rule
+    public Timeout timeout = new ServiceTalkTestTimeout();
+
+    private final ServerContext serverContext;
+
+    public BlockingApiCorrectnessTest() throws Exception {
+        serverContext = GrpcServers.forAddress(localAddress(0))
+                // HTTP filter to modify paths to workaround API restrictions:
+                .appendHttpServiceFilter(service -> new StreamingHttpServiceFilter(service) {
```
NiteshKant: nit: perhaps better to do this switch on the client as the client is intending to make "incorrect requests"

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -572,7 +574,18 @@ public void closeGracefully() throws Exception {
...
                                            final GrpcPayloadWriter<Resp> responseWriter) throws Exception {
-                            route.handle(ctx, requireNonNull(request.iterator().next()), responseWriter);
+                            final Req firstItem;
+                            try (BlockingIterator<Req> requestIterator = request.iterator()) {
+                                firstItem = requireNonNull(requestIterator.next(), "Request item is null");
+                                if (requestIterator.hasNext()) {
+                                    // Consume the next item to make sure it's not a TerminalNotification with an error
+                                    requestIterator.next();
+                                    throw new GrpcStatus(INVALID_ARGUMENT, null,
+                                            "Only a single request item is expected, but saw the second one")
```
NiteshKant: Not changed in this PR but as you are changing code; this would be a better message.

```suggestion
                                            "More than one request message received.")
```

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
```diff
@@ -154,13 +154,14 @@
...
-            final BlockingIterator<Resp> iterator = streamingClientCall.request(metadata, request).iterator();
-            final Resp firstItem = iterator.next();
-            if (iterator.hasNext()) {
-                throw new IllegalArgumentException("Only a single item expected, but saw the second value: " +
-                        iterator.next());
+            try (BlockingIterator<Resp> iterator = streamingClientCall.request(metadata, request).iterator()) {
+                final Resp firstItem = requireNonNull(iterator.next(), "Response item is null");
+                if (iterator.hasNext()) {
+                    iterator.next(); // Consume the next item to make sure it's not a TerminalNotification with an error
+                    throw new IllegalStateException("Only a single response item is expected, but saw the second one");
```
NiteshKant: Not changed in this PR but as you are changing code; this would be a better message.

```suggestion
                    throw new IllegalStateException("More than one response message received.");
```

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -572,7 +574,18 @@ public void closeGracefully() throws Exception {
                         @Override
                         public void handle(final GrpcServiceContext ctx, final BlockingIterable<Req> request,
                                            final GrpcPayloadWriter<Resp> responseWriter) throws Exception {
-                            route.handle(ctx, requireNonNull(request.iterator().next()), responseWriter);
+                            final Req firstItem;
+                            try (BlockingIterator<Req> requestIterator = request.iterator()) {
+                                firstItem = requireNonNull(requestIterator.next(), "Request item is null");
```
NiteshKant: It isn't changed in this PR but I think `null` should pass-through.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
```diff
@@ -154,13 +154,14 @@
         final BlockingStreamingClientCall<Req, Resp> streamingClientCall =
                 newBlockingStreamingCall(serializationProvider, requestClass, responseClass);
         return (metadata, request) -> {
-            final BlockingIterator<Resp> iterator = streamingClientCall.request(metadata, request).iterator();
-            final Resp firstItem = iterator.next();
-            if (iterator.hasNext()) {
-                throw new IllegalArgumentException("Only a single item expected, but saw the second value: " +
-                        iterator.next());
+            try (BlockingIterator<Resp> iterator = streamingClientCall.request(metadata, request).iterator()) {
+                final Resp firstItem = requireNonNull(iterator.next(), "Response item is null");
```
NiteshKant: I think we should pass through `null` values.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
idelpivnitskiy: Our API for gRPC currently does not annotate client calls with `@Nullable` annotation. If I don't do a null check, intellij idea warns that we return a possible `null` value from `@Nonnull` method.
And the `ProtoBufSerializationProvider` never returns null, even if the `Buffer` was empty: https://github.com/apple/servicetalk/blob/c4ab8877ace59689f67af2f4e4ad8e9bcb1939e1/servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProvider.java#L121-L123
Therefore, it never will be `null`. Only `NoSuchElementException` is possible here.

Btw, on the client-side we can simply throw `NoSuchElementException`. But on the server-side, do we need to generate a `GrpcStatusException` with `INVALID_ARGUMENT` if there were no items in the request?

NiteshKant: >> Therefore, it will never be null. Only NoSuchElementException is possible here.

Can a filter modify the request payload to pass `null` since we allow `null` for our async sources?
The meta point here is that unless we require the item to be `non-null`, we should allow for `null` values to pass through as we do in all our async sources. If the intent is only to avoid code style/spotbugs warnings then we can ignore/suppress.

NiteshKant: Emitting a `GrpcStatusException` SGTM

idelpivnitskiy:
JDK11 build failed because of https://github.com/apple/servicetalk/pull/949

PR #948, opened 2020-02-25
Reduce probability of LB selecting an unusable connection

NiteshKant:
__Motivation__

`NettyChannelPublisher` closes the connection if it sees a `cancel()` before all data for the current `Subscriber` is read. However, the client layer makes the connection eligible for selection by load balancer if it sees a `cancel()` being unaware that the connection MAY be eventually closed by the transport. This creates a race condition where load balancer may select a connection which is going to be closed.

__Modification__

Pessimistically assume that if we see a `cancel()` for a request, transport will close the connection so force close the connection. This is pessimistic because it may so happen that completion of read is racing with the cancel and transport may see read completion before cancel and hence not close the connection. At the client layer (where we control LB selection eligibility), it is impossible to discern whether transport will close the connection or not so the safest option is to assume closure. This reduces the possibility of selecting an unusable connection hence the pains associated with debugging those situations at the cost of closing the connection.
For H2, this will be a stream, which is cheap to close but for H1 this will close the actual connection.

__Result__

Reduce possibility of selecting an unusable connection. As `cancel()` is not the only reason for closure, we still have the case where a connection can be selected only to be closed later but we reduce that probability with this change.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ConcurrentRequestsHttpConnectionFilter.java
```diff
@@ -1,149 +0,0 @@
...
-import io.servicetalk.transport.api.ConnectionContext;
-
-import static io.servicetalk.client.api.internal.RequestConcurrencyControllers.newController;
-import static io.servicetalk.client.api.internal.RequestConcurrencyControllers.newSingleController;
-import static io.servicetalk.concurrent.Cancellable.IGNORE_CANCEL;
-import static io.servicetalk.concurrent.api.Executors.immediate;
-import static io.servicetalk.concurrent.api.SourceAdapters.toSource;
-import static io.servicetalk.http.api.HttpEventKey.MAX_CONCURRENCY;
-
-final class ConcurrentRequestsHttpConnectionFilter implements FilterableStreamingHttpConnection {
```
NiteshKant: This was dead code.

PR #950, opened 2020-02-27
Expose `SocketOption` querying on `ConnectionContext` API 

idelpivnitskiy:
Motivation:

Users should be able to read socket options for created connections.

Modifications:

- Add `ConnectionContext.socketOption` API to query socket options;
- Add `ConnectionContextSocketOptionTest` to test new API;

Results:

Users can read socket options of the connections.


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionContextSocketOptionTest.java
```diff
@@ -0,0 +1,244 @@
...
+    @Parameters(name = "protocol={0}")
+    public static Object[] data() {
+        return Protocol.values();
+    }
+
+    @Test
+    public void tcpStandardSocketOptionIsNotNull() throws Exception {
+        testSocketOption("TCP_NODELAY", Boolean.class, is(notNullValue()),
+                // HTTP_2 stream channel does not have TCP_NODELAY option
+                protocol == Protocol.HTTP_2 ? equalTo("null") : not(equalTo("null")));
```
idelpivnitskiy: Server connection context uses child-stream for h2 that does not allow users to query multiple TCP options for the original parent connection, like `TCP_NODELAY` / `SO_RCVBUF` / `SO_SNDBUF` / etc.
Do we want to keep it as-is? We can expose options for the original TCP connection by keeping ref for the parent's `ChannelConfig`.

idelpivnitskiy:
JDK11 build failed due to docker issue

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SocketOptionUtils.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        if (option == ServiceTalkSocketOptions.CONNECT_TIMEOUT) {
+            return (T) config.getOption(ChannelOption.CONNECT_TIMEOUT_MILLIS);
+        }
+        if (option == ServiceTalkSocketOptions.WRITE_BUFFER_THRESHOLD) {
+            final WriteBufferWaterMark result = config.getOption(ChannelOption.WRITE_BUFFER_WATER_MARK);
+            return result == null ? null : (T) Integer.valueOf(result.high());
+        }
+        if (option == ServiceTalkSocketOptions.IDLE_TIMEOUT) {
+            return idleTimeoutMs == null ? null : (T) idleTimeoutMs;    // TODO: return (T) idleTimeoutMs;
```
NiteshKant: Can you address this TODO?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -125,22 +130,25 @@
...
+                                   ExecutionStrategy executionStrategy) {
+        this(channel, allocator, executor, terminalMsgPredicate, closeHandler, flushStrategy, idleTimeoutMs,
+                executionStrategy, null);
     }
 
     private DefaultNettyConnection(Channel channel, BufferAllocator allocator, Executor executor,
                                    TerminalPredicate<Read> terminalMsgPredicate, CloseHandler closeHandler,
-                                   FlushStrategy flushStrategy, ExecutionStrategy executionStrategy,
-                                   @Nullable SSLSession sslSession) {
+                                   FlushStrategy flushStrategy, @Nullable Long idleTimeoutMs,
```
NiteshKant: I am somewhat apprehensive about individually adding options here which are otherwise not stored by netty. Anything else we add will mean we add more arguments here.

Can we instead take+store the `TcpConfig` here and pass the same to `SocketOptionUtils` for querying? If we do use `TcpConfig` we can use that to also get the `FlushStrategy`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionContextSocketOptionTest.java
NiteshKant: Are there any options for H2 that are applicable for a stream? If not, should we always just query the options from the parent context?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionContextSocketOptionTest.java
```diff
@@ -0,0 +1,244 @@
...
+                        .protocols(protocol.config);
+        if (idleTimeoutMs != null) {
+            builder.socketOption(ServiceTalkSocketOptions.IDLE_TIMEOUT, idleTimeoutMs);
+        }
+        return builder
+                .buildBlocking();
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <T> SocketOption<T> socketOption(String name, Class<T> type) {
```
NiteshKant: Since you are creating a new `HttpServer` for each test, using the headers indirection is more work (String -> option translations). Instead you can simply pass the socket option and expected value to `startServer` and use that inside the service and then fail the response with the failure message.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
idelpivnitskiy: Thought about that but decided to defer until we add anything else to better understand how to structure it (maybe it will be better to have an object to keep these options instead of using the full `TcpConfig`).

We do not expose `AbstractReadOnlyTcpConfig` as a public class for now. If you think we can make `AbstractReadOnlyTcpConfig` public (it lives in `tcp-netty-internal` module) I can do the change now. Otherwise, let's defer?

We may also consider using a single `ExecutionContext` instead of other 3 separate arguments: `BufferAllocator`, `Executor`, `ExecutionStrategy`. Can do here or as a follow-up.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SocketOptionUtils.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        // Try to look for a ChannelOption with the same name and type:
```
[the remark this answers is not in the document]
idelpivnitskiy: Netty has a lot more other options that users might be interested in to query, especially for different transports, like `TCP_FASTOPEN_CONNECT` for epoll. See `EpollChannelOption` class from netty.

If we will follow the policy to be consistent between `get` and `set`, every time users request a new option for querying we will need to provide one more `ServiceTalkSocketOption` constant and implement support for _setting_ this option. However, users may not bother with setting it, just querying. Also, maintaining a single or separate lists of `ServiceTalkSocketOption` constants that are applicable only for a specific transport may be not ideal.

`SocketOption<T>` is a public JDK API. Users can use it to create the required options for querying without asking us. The only risk here is that they can take internal `ByteBufAllocator` from netty. But that will be their risky choice or we can prevent specific options from querying. The list is short and contains only 3 options: `ALLOCATOR`, `RCVBUF_ALLOCATOR`, `MESSAGE_SIZE_ESTIMATOR`.

Considering that overhead of keeping `get` and `set` in consistency for every new option, I decided to let users query what they want with existing API.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionContext.java
```diff
@@ -51,7 +55,21 @@
...
      * @return the {@link ExecutionContext} for this {@link ConnectionContext}.
      */
     ExecutionContext executionContext();
+
+    /**
+     * Get the {@link SocketOption} value of type {@code T} for this {@link ConnectionContext}.
+     *
+     * @param option {@link SocketOption} to get.
+     * @param <T> the type of the {@link SocketOption} value.
+     * @return the {@link SocketOption} value of type {@code T} for this {@link ConnectionContext} or {@code null} if
```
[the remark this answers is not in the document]
idelpivnitskiy: With existing approach (design choice described in the discussion below) we support any `SocketOption` users implement. But they will see `null` for options that are not known by transport.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionContextSocketOptionTest.java
idelpivnitskiy: From what I see in netty, it creates a new `DefaultChannelConfig` for each stream. And these config is a fresh new and is not related to the parent's `ChannelConfig`. Therefore, each channel has its own subset of options. In theory (but not in the current ST) some code may modify that options for each stream independently (like `AUTO_READ, `AUTO_CLOSE`, etc.) and it will not have side-effects for parent channel.

I think it will be more practical for users to be interested in parent channel options. It will require one more reference in `NettyHttpServerConnection`. Let me try and will see the overhead.

idelpivnitskiy: Done: https://github.com/apple/servicetalk/pull/950/commits/4fbb1f9e2a8430c1baab762f1e16217c842875a3

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
NiteshKant: Can we just use `TcpConfig` for now as these are internal classes and we can make sure we do not update config? If you want, we can also wrap the config to throw from the setters.

I am also OK to defer, your call.

idelpivnitskiy: Will defer that, thanks

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SocketOptionUtils.java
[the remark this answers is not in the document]
idelpivnitskiy: As we discussed offline, I made get/set consistent in https://github.com/apple/servicetalk/pull/950/commits/09d09184e14c782c61beb22fa1107a3827c90670.
Here is an issue to enhance these two methods to support custom options: https://github.com/apple/servicetalk/issues/953

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionContext.java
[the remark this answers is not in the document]
idelpivnitskiy: Done https://github.com/apple/servicetalk/commit/09d09184e14c782c61beb22fa1107a3827c90670

PR #954, opened 2020-03-03
Fix flakiness in `ResponseCancelTest`

NiteshKant:
__Motivation__

`ResponseCancelTest` is flaky due to a race condition that successful response or cancel will race to get to the connection filter and the result will either be success or failure.

__Modification__

- `cancel()` and `cancelAfterSuccessOnTransport()` tests were both doing same actions unintentionally. Modified the test to reflect the intentions.
- `cancelAfterSuccessOnTransport()` can either result in success or failure after the first request, modified test to verify the same.

__Result__

Fix tests.


idelpivnitskiy:
Looks like CI is happy. I will pull this in and update other PRs

PR #957, opened 2020-03-04
Explicit set MacOSDnsServerAddressStreamProvider when it can be used …

normanmaurer:
…to ensure we return the correct nameservers when on macOS

Motivation:

When on MacOS we need to use MacOSDnsServerAddressStreamProvider explicitly to ensure we return the correct nameservers when a VPN is used.

Modifications:

Check if MacOSDnsServerAddressStreamProvider can be used and if so explicit enable it

Result:

Resolution will use the correct nameservers on MacOS in all cases


idelpivnitskiy:
@normanmaurer I guess this PR can be closed now, because users will be able to use `netty-resolver-dns-native-macos` automatically when it's in a classpath, right?

normanmaurer:
@idelpivnitskiy right... as long as you have it on the class path... so you still need the dependency

idelpivnitskiy:
Let me close this PR then. We will think how/where to document it. Thanks @normanmaurer

Issue #758, opened 2019-09-03
Flaky test: io.servicetalk.concurrent.reactivestreams.tck.PublisherPublishOnOverrideTckTest.required_createPublisher1MustProduceAStreamOfExactly1Element


jayv:
Observed with https://github.com/apple/servicetalk/pull/757

```
java.lang.AssertionError: Async error during test execution: Rule 3.9 states non-positive request signals are illegal, but got: -1821637576
	at org.testng.Assert.fail(Assert.java:78)
	at org.reactivestreams.tck.TestEnvironment.verifyNoAsyncErrorsNoDelay(TestEnvironment.java:374)
	at org.reactivestreams.tck.TestEnvironment.subscribe(TestEnvironment.java:306)
	at org.reactivestreams.tck.TestEnvironment.newManualSubscriber(TestEnvironment.java:321)
	at org.reactivestreams.tck.TestEnvironment.newManualSubscriber(TestEnvironment.java:316)
	at org.reactivestreams.tck.PublisherVerification$1.run(PublisherVerification.java:165)
	at org.reactivestreams.tck.PublisherVerification.activePublisherTest(PublisherVerification.java:1142)
	at org.reactivestreams.tck.PublisherVerification.required_createPublisher1MustProduceAStreamOfExactly1Element(PublisherVerification.java:162)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:74)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:673)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:846)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1170)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:125)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:109)
	at org.testng.TestRunner.runWorkers(TestRunner.java:1147)
[...]

PR #958, opened 2020-03-05
Don't re-use *Exceptions when stored in static fields if enableSuppre…

normanmaurer:
…ssion can not be set to false

Motivation:

Have a static instance of an Exception which does not set enableSuppression to false when constructed is considered to be dangerous as it may escape to the user where an user may add more exceptions to it. The problem with this is that the internal storage will grow and grow over time and never be GC'ed. This can lead to OOME situations.

Modifications:

Create new *Exceptions when needed and remove usage of static stored *Exceptions if needed.

Result:

No more risk of OOME due user calling `addSuppressed` on exceptions that are propagated through the code

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RepeatStrategies.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RepeatStrategies.java
@@ -35,8 +35,7 @@ public final class RepeatStrategies {
     public static final class TerminateRepeatException extends Exception {
         private static final long serialVersionUID = -1725458427890873886L;
 
-        // It is fine to reuse this instance and let it escape to the user as we enableSuppression is set to false
-        // while constructing it.
+        // It is fine to reuse this instance and let it escape to the user as enableSuppression is set to false.
         static final TerminateRepeatException INSTANCE = new TerminateRepeatException();
 
         // Package-private as the user should never instance it.
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RepeatStrategies.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RepeatStrategies.java
@@ -35,6 +35,8 @@ public final class RepeatStrategies {
     public static final class TerminateRepeatException extends Exception {
         private static final long serialVersionUID = -1725458427890873886L;
 
+        // It is fine to reuse this instance and let it escape to the user as we enableSuppression is set to false
+        // while constructing it.
         static final TerminateRepeatException INSTANCE = new TerminateRepeatException();
 
         // Package-private as the user should never instance it.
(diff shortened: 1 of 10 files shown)


NiteshKant:
@normanmaurer Same test:

`io.servicetalk.transport.netty.internal.DefaultNettyConnectionTest.testNoErrorEnrichmentWithoutCloseHandlerOnError`

failed for both PRBs, looks like something isn't happy 😢

normanmaurer:
@NiteshKant will check

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RepeatStrategies.java
```diff
@@ -35,6 +35,8 @@
     public static final class TerminateRepeatException extends Exception {
         private static final long serialVersionUID = -1725458427890873886L;
 
+        // It is fine to reuse this instance and let it escape to the user as we enableSuppression is set to false
+        // while constructing it.
```
NiteshKant:
```suggestion
        // It is fine to reuse this instance and let it escape to the user as enableSuppression is set to false.
```

NiteshKant:
@normanmaurer JDK 11 build failure is due to a known flaky test:

https://github.com/apple/servicetalk/issues/758

Can you add this failure details to the issue?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/StacklessClosedChannelException.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.transport.netty.internal;
+
+import io.servicetalk.concurrent.internal.ThrowableUtils;
+
+import java.nio.channels.ClosedChannelException;
+
+public final class StacklessClosedChannelException extends ClosedChannelException {
```
NiteshKant: Interesting that the quality task did not complain of the lack of javadoc in a public class \cc @idelpivnitskiy 

Can you add javadocs here and for `newInstance()` method?

normanmaurer: done... not resolving this so @idelpivnitskiy can still see it tho...

idelpivnitskiy: Interesting, checkstyle introduced new improved rules for javadoc checks:
MissingJavadocMethod
MissingJavadocPackage
MissingJavadocType
See https://checkstyle.sourceforge.io/config_javadoc.html#MissingJavadocMethod

The old rules (JavadocMethod, JavadocPackage, JavadocType) work as regex, when new rules work on checkstyle token parsing. And it looks like the old rules do not work correctly anymore. I tried locally this PR, when I activate `MissingJavadocMethod` rule, it captures this class (for the previous commit).

normanmaurer:
Merged... thanks :)

PR #961, opened 2020-03-10
Publisher#range operator

Scottmitch:
Motivation:
The canonocal way to do an asynchornous for lop with integer indexes in reactive streams is to use a range() operator. We currently don't have a range operator.

Modifications:
- Add Publisher#range(int begin, int end)
- Add Publisher#range(int begin, int end, int stride)

Result:
For loops with integer indexes can be done directly with Publisher#range(..)

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RangeIntPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RangeIntPublisher.java
@@ -41,7 +41,7 @@ final class RangeIntPublisher extends AbstractSynchronousPublisher<Integer> {
             throw new IllegalArgumentException("begin(" + begin + ") > end(" + end + ")");
         }
         if (stride <= 0) {
-            throw new IllegalArgumentException("stride: " + stride);
+            throw new IllegalArgumentException("stride: " + stride + " (expected >0)");
         }
         this.begin = begin;
         this.end = end;
@@ -85,7 +85,7 @@ final class RangeIntPublisher extends AbstractSynchronousPublisher<Integer> {
                     return;
                 }
             }
-            if (index >= end) {
+            if (index == end) {
                 sendComplete();
             }
         }
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RangeIntPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RangeIntPublisher.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright © 2020 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RangeIntPublisher.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    void doSubscribe(final Subscriber<? super Integer> subscriber) {
+        subscriber.onSubscribe(new RangeIntSubscription(subscriber));
+    }
+
+    private final class RangeIntSubscription implements PublisherSource.Subscription {
+        private final PublisherSource.Subscriber<? super Integer> subscriber;
+        private long pendingN;
+        private int index = RangeIntPublisher.this.begin;
```
normanmaurer: nit: consider passing begin into the constructor so you can make the class static.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RangeIntPublisher.java
```diff
@@ -0,0 +1,116 @@
...
+            pendingN = addWithOverflowProtection(pendingN, n);
+            for (; pendingN > 0 && index < end; --pendingN, index += min(stride, (long) end - index)) {
+                try {
+                    subscriber.onNext(index);
+                } catch (Throwable cause) {
+                    sendOnError(cause);
+                    return;
+                }
+            }
+            if (index >= end) {
```
normanmaurer: nit:  `==` should be enough ?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RangeIntPublisher.java
```diff
@@ -0,0 +1,116 @@
...
+    RangeIntPublisher(int begin, int end) {
+        this(begin, end, 1);
+    }
+
+    RangeIntPublisher(int begin, int end, int stride) {
+        if (begin > end) {
+            throw new IllegalArgumentException("begin(" + begin + ") > end(" + end + ")");
+        }
+        if (stride <= 0) {
+            throw new IllegalArgumentException("stride: " + stride);
```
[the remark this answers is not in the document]
Scottmitch: done

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RangeIntPublisher.java
Scottmitch: the inner class also needs to access `end` and `stride` from the outer class. I will leave as is for now.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RangeIntPublisher.java
Scottmitch: sure, let me update.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/FromIntRangePublisherTest.java
```diff
@@ -0,0 +1,179 @@
...
+    public void reentrantCancelStopsDelivery() {
+        reentrantCancelStopsDelivery(true);
+    }
+
+    @Test
+    public void reentrantCancelStopsDeliveryStride() {
+        reentrantCancelStopsDelivery(false);
+    }
+
+    private void reentrantCancelStopsDelivery(boolean stride) {
```
NiteshKant: An additional test case may be useful to verify `request-n` post `cancel` are ignored.

PR #962, opened 2020-03-10
MultiAddressUrl client fails to send http request after https request

idelpivnitskiy:
Motivation:

If `MultiAddressUrl` was used to send an HTTPS request over secure
connection, all other requests to new non-secure servers will fail
with `DecoderException`. It happens because we incorrectly copy the
`HttpClientBuilder` template. We did not create a new instance of
`TcpClientConfig` and `HttpConfig`.

Modifications:

- Add copy ctor for `HttpConfig`;
- Use copy ctors for `TcpClientConfig` and `HttpConfig` in
`HttpClientConfig`;
- Add test to verify that `MultiAddressUrl` client can connect to
insecure server after connecting to a secure server and vice versa;

Result:

`HttpClientConfig` makes a full copy of `TcpClientConfig` and
`HttpConfig` that prevents the `HttpClientBuilder` template from
modifications. `MultiAddressUrl` client correctly copies the
builder template without internal state modifications.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
@@ -289,8 +289,7 @@ final class DefaultMultiAddressUrlHttpClientBuilder
             this.executionContext = requireNonNull(executionContext);
         }
 
-        private FilterableStreamingHttpClient selectClient(
-                HttpRequestMetaData metaData) {
+        private FilterableStreamingHttpClient selectClient(final HttpRequestMetaData metaData) {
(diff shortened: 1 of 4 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -289,8 +289,7 @@ public StreamingHttpClient apply(final UrlKey urlKey) {
             this.executionContext = requireNonNull(executionContext);
         }
 
-        private FilterableStreamingHttpClient selectClient(
-                HttpRequestMetaData metaData) {
+        private FilterableStreamingHttpClient selectClient(final HttpRequestMetaData metaData) {
```
normanmaurer: nit: consider reverting changes which are just about formatting

PR #963, opened 2020-03-10
Add a separate page for `ServiceDiscoverer` in docs

idelpivnitskiy:
Motivation:

`ServiceDiscoverer` is another core abstraction in ServiceTalk
what does not coupled to any protocol and should be well documented.

Modifications:

- Add a separate page for docs website about `ServiceDiscoverer`;
- Update links in all other places to refer to this new page;
- Describe how SD works, the default implementation, and how macOS
users may address DNS resolution issues;

Result:

Better and more comprehensive doc for `ServiceDiscoverer`.


idelpivnitskiy:
If you want to look at it locally, this is the task to generate to docs:
```
 ./gradlew validateLocalDocSite
```

servicetalk-dns-discovery-netty/docs/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,71 @@
+// Configure {source-root} values based on how this document is rendered: on GitHub or not
+ifdef::env-github[]
+:source-root:
+endif::[]
+ifndef::env-github[]
+ifndef::source-root[:source-root: https://github.com/apple/servicetalk/blob/{page-origin-refname}]
+endif::[]
+
+= Service Discovery
```
NiteshKant: It does not seem correct to have all docs related to Service Discovery under DNS as we may add more implementations for service discovery later. If you really want to extract Service Discovery into a separate doc then adding it to `client-api` would be more appropriate. 

I would suggest leaving this definition of `Service Discovery` in HTTP for now and add the doc describing DNS implementation here, optionally referring to the HTTP Service Discovery section in HTTP. On the other hand, if you want to move this definition under `client-api`, its fine by me.

servicetalk-dns-discovery-netty/docs/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+Another core abstraction in ServiceTalk is its
```
NiteshKant:
```suggestion
A core abstraction in ServiceTalk is its
```

servicetalk-dns-discovery-netty/docs/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,71 @@
...
+link:{source-root}/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java[ServiceDiscoverer]
+underpinning the various protocol _Clients_ (eg. HTTP/1.1, HTTP/2, gRPC, etc.).
```
NiteshKant:
```suggestion
used by the protocol _Clients_ (eg. HTTP/1.1, HTTP/2, gRPC, etc.).
```

servicetalk-dns-discovery-netty/docs/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+For scenarios where a client communicates with multiple homogeneous footnote:[belonging to the same service, providing
+same capabilities] server instances, managing addresses of multiple server instances for each _Client_ is complex.
+_ServiceDiscoverer_ makes this easy for users by transparently communicating with mechanisms that know about the actual
+location of remote servers and may provide updates when the network topology is changing.
```
NiteshKant: "location of remote servers" and "network topology" terms are misleading here as an SD does not really have to know about network topology or physical server locations. As all clients in ST use service discovery I would suggest using the following language:

> A ServiceTalk client leverages _ServiceDiscoverer_ to find server instances that it can use for processing requests. A  _ServiceDiscoverer_  finds server instances associated with a logical name<TODO: Add link to #discover() method>. Additionally, ServiceTalk assumes that the client operates in a dynamic ephemeral environment and the server instances associated with a logical name _may_ change over time. So, a _ServiceDiscoverer_ returns a `Publisher` of server instances and their current relationship to the logical name i.e. whether it is added or removed from the logical name. Using this information a _Client_ can maintain an active collection of server instances that it can use.

idelpivnitskiy: > <TODO: Add link to #discover() method>

It's hard to add a link to the line number of the source code file because the link will be broken when we update the source code. I slightly reworded your suggestion to include `ServiceDiscovererEvents`.

servicetalk-dns-discovery-netty/docs/modules/ROOT/pages/index.adoc
idelpivnitskiy: 💯 `client-api` is a better place.

PR #967, opened 2020-03-13
Ensure `TcpConnector#connect` does not emit error after success

NiteshKant:
__Motivation__

`TcpConnector#connect` eagerly terminates the `Single` when the channel is registered. It also terminates the `Single` with failure if connect fails.
This violates the contract of the `Subscriber` that there should only be a single termination signal.

__Modification__

As we have to complete the returned `Single` with the channel as soon as the channel is registered to ensure we do not miss channel events and we also have to propagate connect failures, we can not abide by the `Single<Channel>` contract. Instead of returning `Single<Channel>` now returning `Single<Connection>` and do the translation of `Channel` -> `Connection` using a supplied `Function`.

__Result__

`TcpConnector#connect` abides by the `Subscriber` contract.


servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpConnector.java
```diff
@@ -68,67 +76,74 @@ private TcpConnector() {
...
-            protected void handleSubscribe(final Subscriber<? super Channel> subscriber) {
-                connectFutureToListener(localAddress, resolvedRemoteAddress, subscriber,
-                        connect0(localAddress, resolvedRemoteAddress, config, autoRead, executionContext, subscriber));
+            protected void handleSubscribe(final Subscriber<? super C> subscriber) {
+                ConnectHandler<C> connectHandler = new ConnectHandler<>(subscriber, connectionFactory);
+                try {
+                    Future<?> connectFuture = connect0(localAddress, resolvedRemoteAddress, config, autoRead,
+                            executionContext, connectHandler);
+                    connectHandler.connectFuture(connectFuture);
+                    connectFuture.addListener(f -> {
```
idelpivnitskiy: Consider verifying if `connectFuture` is already done before applying the listener, as we do in other places:
https://github.com/apple/servicetalk/blob/e5f4c563e0b5818fc8415d64a3ddd1ba1b90b393/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscoverer.java#L370-L376

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpConnector.java
```diff
@@ -209,4 +206,87 @@ private NoopNettyAddressResolverGroup() {
...
+                    .subscribeShareContext())
+                    .subscribe(new Subscriber<C>() {
+                        @Override
+                        public void onSubscribe(final Cancellable cancellable) {
+                            flatMapCancellable.delayedCancellable(cancellable);
+                        }
+
+                        @Override
+                        public void onSuccess(final C connection) {
+                            requireNonNull(connection);
```
idelpivnitskiy: `SingleSource.Subscriber` defines a parameter for `onSuccess` as `@Nullable`. Instead of throwing from here, should we propagate `onError` instead?

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpConnector.java
NiteshKant: We do that when we are already on the eventloop and no ordering guarantees are required. In this case, adding a listener makes sure that the `Single` is completed on the eventloop.

(This code was just refactored, so I would continue with the approach what was present before but if you feel we need to change this in general in all places, we can discuss/do that separately)

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpConnector.java
NiteshKant: Good point, the `null-check` is only required for when we want to call `close()` so instead of disallowing `null`, lemme guard the `close`.

servicetalk-tcp-netty-internal/src/testFixtures/java/io/servicetalk/tcp/netty/internal/TcpClient.java
```diff
@@ -82,13 +83,24 @@ public TcpClient(TcpClientConfig config) {
      */
     public NettyConnection<Buffer, Buffer> connectBlocking(ExecutionContext executionContext, SocketAddress address)
             throws ExecutionException, InterruptedException {
-        return TcpConnector.connect(null, address, config, false, executionContext)
-                .flatMap(channel -> DefaultNettyConnection.<Buffer, Buffer>initChannel(channel,
+        return connect(executionContext, address).toFuture().get();
+    }
+
+    /**
+     * Connect and await for the connection.
```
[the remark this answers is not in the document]
NiteshKant: Yep .. copy-pasta, lemme fix

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpConnector.java
```diff
@@ -244,15 +244,16 @@ public void onSubscribe(final Cancellable cancellable) {
                         }
 
                         @Override
-                        public void onSuccess(final C connection) {
-                            requireNonNull(connection);
+                        public void onSuccess(@Nullable final C connection) {
                             if (terminatedUpdater.compareAndSet(ConnectHandler.this, 0, 1)) {
                                 target.onSuccess(connection);
```
idelpivnitskiy: Do we assume that downstream target `Subscriber` will handle `null` instead of a connection? I think it's easy to assume from the caller side that it's expected to receive non-null connection. Was suggesting to invoke `target.onError` instead of throwing from `onSuccess`.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpConnector.java
```diff
@@ -209,4 +206,88 @@ private NoopNettyAddressResolverGroup() {
...
+                        }
+
+                        @Override
+                        public void onSuccess(@Nullable final C connection) {
+                            if (terminatedUpdater.compareAndSet(ConnectHandler.this, 0, 1)) {
+                                target.onSuccess(connection);
+                            } else {
+                                LOGGER.info("Connection {} created for a channel: {} but connect failed previously. " +
+                                                "Closing connection",
+                                        connection, channel);
```
normanmaurer: nit: This logging message will read a bit strange and may be confusing if `connection == null`.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpConnector.java
```diff
@@ -209,4 +206,88 @@ private NoopNettyAddressResolverGroup() {
...
+                            flatMapCancellable.delayedCancellable(cancellable);
+                        }
+
+                        @Override
+                        public void onSuccess(@Nullable final C connection) {
+                            if (terminatedUpdater.compareAndSet(ConnectHandler.this, 0, 1)) {
+                                target.onSuccess(connection);
+                            } else {
+                                LOGGER.info("Connection {} created for a channel: {} but connect failed previously. " +
+                                                "Closing connection",
```
normanmaurer: also nit.... All other log messages end with a `.`.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpConnector.java
NiteshKant: >> Do we assume that downstream target Subscriber will handle null instead of a connection?
Yes, in general in operators we pass-through what we get unless we use it.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpConnector.java
NiteshKant: mmm ya but as we do not expect the connection to be `null` and I want to add a log if we are closing the connection, I think it is OK to not add more conditional to handle `null`. I will change the level to `debug` though as there isn't anything a user is expected to do here.

PR #968, opened 2020-03-16
JDK provider incorrectly configures default ciphers for TLSv1.3

idelpivnitskiy:
Motivation:

We use the old approach of configuring secure protocols in netty via
`SSLEngine#setEnabledProtocols(...)`. In this case, `SslContextBuilder`
does not know if users configure TLSv1.3 or not and therefore, sets
incorrect set of default ciphers which does not support TLSv1.3.
As the result, users see:
javax.net.ssl.SSLHandshakeException: No appropriate protocol
(protocol is disabled or cipher suites are inappropriate).

Modifications:

- Use `SslContextBuilder#protocols(...)` instead of
`SSLEngine#setEnabledProtocols(...)` to set secure protocols;
- Delete `WrappingSslContext` that is not required anymore;

Result:

`SslContextBuilder` correctly sets default ciphers when TLSv1.3
is present.


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/Tls13Test.java
```diff
@@ -40,61 +50,88 @@
...
+            serverSecurityConfigurator.ciphers(singletonList(cipher));
+        }
+        try (ServerContext serverContext = serverSecurityConfigurator
                 .commit(DefaultTestCerts::loadServerPem, DefaultTestCerts::loadServerKey)
-                .listenBlockingAndAwait((ctx, request, responseFactory) -> responseFactory.ok()
-                        .payloadBody(request.payloadBody(textDeserializer()), textSerializer()))) {
+                .listenBlockingAndAwait((ctx, request, responseFactory) -> {
+                    assertThat(request.payloadBody(textDeserializer()), equalTo("request-payload-body"));
+                    SSLSession sslSession = ctx.sslSession();
+                    assertThat(sslSession, is(notNullValue()));
```
NiteshKant: Can we also verify the cipher here?

PR #969, opened 2020-03-17
More strict decoding of HTTP/1.x start-line

idelpivnitskiy:
Motivation:

RFC7230 allows only a single SP between start-line components:
https://tools.ietf.org/html/rfc7230#section-3.1
And it adds additional restrictions for which characters are
allowed for each individual component.

Modifications:

- Skip only limited number of prefacing control characters;
- Do not allow leading/trailing WS for start-line;
- Expect only a single SP/HTAB as a delimiter between start-line
components;
- Verify that request-target and reason-phrase has only allowed
characters;
- Allow only upper-case letters for HTTP-method;
- Simplify header field-name/field-value parsing;
- Improved parsing for HTTP version;
- Share common code between `HttpRequestDecoderTest` and
`HttpResponseDecoderTest`;
- Address issues with existing tests and add more tests
for decoders;

Result:

More strict decoding of HTTP/1.x start-line.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestDecoder.java
```diff
@@ -58,19 +62,27 @@ protected boolean isDecodingRequest() {
...
-        for (int i = 0; i < len; ++i) {
-            byte b = buffer.getByte(buffer.readerIndex() + i);
-            if (b == ' ' && i > 0) {
-                // If we find a space after at least one capital letter, accept this as valid.
+        final int writerIndex = buffer.writerIndex();
+        int readerIndex = buffer.readerIndex();
+        if (readerIndex == writerIndex) {   // ByteBuf is not readable
+            return;
+        }
+        byte b = buffer.getByte(readerIndex);
```
Scottmitch: any reason not to use a `ByteProcessor` here to reduce calls to `getByte`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestDecoder.java
```diff
@@ -79,17 +91,15 @@ protected HttpRequestMetaData createMessage(final ByteBuf buffer,
...
         return newRequestMetaData(nettyBufferToHttpVersion(buffer, thirdStart, thirdLength),
-                decodeHttpMethod(buffer.toString(firstStart, firstLength, US_ASCII)),
+                decodeHttpMethod(buffer, firstStart, firstLength),
                 buffer.toString(secondStart, secondLength, US_ASCII),
                 headersFactory().newHeaders());
     }
 
-    private static HttpRequestMethod decodeHttpMethod(final String methodName) {
+    private static HttpRequestMethod decodeHttpMethod(final ByteBuf buffer, final int start, final int length) {
+        buffer.forEachByte(start, length, ENSURE_UPPER_CASE);
```
Scottmitch: consider doing validation only for custom header names, most of the time we shouldn't have to do the validation because it is handled in the lookup (e.g. if it matches a cached entry it is upper case). It is worth paying the string conversion costs assuming most of the time we will get a valid string.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpResponseDecoder.java
```diff
@@ -38,13 +41,23 @@
...
 import static java.util.Objects.requireNonNull;
 
 final class HttpResponseDecoder extends HttpObjectDecoder<HttpResponseMetaData> {
 
     private static final byte[] FIRST_BYTES = "HTTP".getBytes(US_ASCII);
+    private static final ByteProcessor ENSURE_NO_CONTROL_CHARS = value -> {
+        if (value == HT) {
+            return true;    // allow HTAB
+        }
+        if (isISOControl(value)) {
```
Scottmitch: nit: consider combining conditionals:

```java
if (value != HT && isISOControl(value) {
  throw ...
}
return true;
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpResponseDecoder.java
```diff
@@ -84,10 +97,19 @@ protected HttpResponseMetaData createMessage(final ByteBuf buffer, final int fir
...
+                        reasonPhrase(buffer, thirdStart, thirdLength)),
                 headersFactory().newHeaders());
     }
 
+    private static String reasonPhrase(final ByteBuf buffer, final int start, final int length) {
+        if (length <= 0) {
+            return EMPTY_STRING;
+        }
+        // Verify reason-phrase = *( HTAB / SP / VCHAR / obs-text )
+        buffer.forEachByte(start, length, ENSURE_NO_CONTROL_CHARS);
```
Scottmitch: https://tools.ietf.org/html/rfc7230#section-3.1.2
> A client SHOULD ignore the reason-phrase content.

sad to waste cycles validating something that should be ignored 😢

idelpivnitskiy: We can optimize the "most-likely path" of hitting the known status code if we will make `HttpResponseStatus.valueOf(int)` public. Right know, there is no cheap way to check if we already have a constant for the `status-code` because we always fallback to `new HttpResponseStatus` that is relatively expensive. We can do that in a follow up. Something like:
1. Check if the `status-code` is known.
2. If yes, return the constant. If no, do the validation and create a new `HttpResponseStatus`.

PR #971, opened 2020-03-18
Add a classifier for `netty-resolver-dns-native-macos` in docs

idelpivnitskiy:
Motivation:

`netty-resolver-dns-native-macos` dependency requires the `osx-x86_64`
classifier to be set.

Modifications:

- Add `osx-x86_64` classifier for `netty-resolver-dns-native-macos`
dependency in `Service Discovery` docs;

Result:

Correct dependency declaration for `netty-resolver-dns-native-macos`.


idelpivnitskiy:
@Scottmitch we don't. I just a way to tell users that they have to have a version between artifact name and classifier. Assume the build will fail if there is not `$nettyVersion` defined and they will fix it.

PR #973, opened 2020-03-18
Enforce JDK8 when running `publish-docs.sh`

idelpivnitskiy:
Motivation:

Different JDK versions produce different website style for
javadoc.

Modifications:

- Make sure the current Java version is 1.8 before running
`publish-docs.sh`;

Result:

Javadoc website always published with the consistent format
and style.

### DIFF
--- a/scripts/publish-docs.sh
+++ b/scripts/publish-docs.sh
@@ -49,7 +49,7 @@ function clean_up_gh_pages() {
 }
 
 # Enforce JDK8 to keep javadoc format consistent for all versions:
-java_version=$(java -version 2>&1 | grep 'version' 2>&1 | awk -F\" '{ split($2,a,"."); print a[1]"."a[2]}')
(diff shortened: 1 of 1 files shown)


scripts/publish-docs.sh
```diff
@@ -48,6 +48,13 @@ function clean_up_gh_pages() {
   rm -rf gh-pages
 }
 
+# Enforce JDK8 to keep javadoc format consistent for all versions:
+java_version=$(java -version 2>&1 | grep 'version' 2>&1 | awk -F\" '{ split($2,a,"."); print a[1]"."a[2]}')
```
NiteshKant: As we use `gradle` to create javadocs, should we be considering if `org.gradle.java.home` is set?

idelpivnitskiy: You are right, we need to extract the version of JVM that gradlew sees. I've updated the script to use `./gradlew -versions`.

PR #976, opened 2020-03-23
NettyChannelPublisher cancel active subscriber should terminate

Scottmitch:
Motivation:
NettyChannelPublisher allows for resubscribes, and delivers queued data to the new Subscriber. If the previous Subscriber consumed some data and cancelled this may lead to the new Subscriber receiveing partial data from the last request.

Modifications:
- NettyChannelPublisher should discard any pending data, and deliver an error to new Subscribers if a previously active Subscriber cancels.

Result:
Resubscribes to NettyChannelPublisher won't get partial data for previous requests, like in the case of client pipelinining.

### DIFF
--- a/servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/NettyChannelPublisherTest.java
+++ b/servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/NettyChannelPublisherTest.java
@@ -27,9 +27,12 @@ import io.servicetalk.concurrent.internal.ServiceTalkTestTimeout;
 import io.servicetalk.transport.api.ConnectionContext.Protocol;
 import io.servicetalk.transport.netty.internal.NettyConnection.TerminalPredicate;
 
+import io.netty.channel.ChannelDuplexHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelOutboundHandlerAdapter;
+import io.netty.channel.ChannelPromise;
 import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.channel.socket.ChannelInputShutdownReadComplete;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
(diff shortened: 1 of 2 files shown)


servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/NettyChannelPublisherTest.java
```diff
@@ -111,6 +113,27 @@ public void tearDown() throws Exception {
...
+        subscription1.request(1);
+
+        assertFalse(channel.writeInbound(1));
+        Integer next = subscriber1.takeOnNext();
+        assertThat(next, is(1));
+        assertFalse(channel.writeInbound(2)); // this write should be queued, because there isn't any requestN demand.
+
+        subscription1.cancel(); // cancel of active subscription should clear the queue and fail future Subscribers.
+
+        toSource(publisher).subscribe(subscriber2);
```
NiteshKant: Can you also add a test that does a `channelRead()` between `cancel()` and re-subscribe? That will re-initialize the `pending` queue.

Scottmitch: The `cancel` operation does a `shutdownInput` so no more data should be delivered by Netty via `channelRead`. `EmbeddedChannel` throws `ClosedChannelException` if I try to deliver more data, and I will add a test for this.

NiteshKant: The case I was trying to get to was when data is enqueued in the channel when `cancel()` happens in which case Netty should deliver data first and then the shutdown. We do not have to go out of our way to verify the scenario but looks like having a handler that buffers data could do the trick?

Scottmitch: good call ... I was trying to get away with avoiding additional conditional in `channelRead` but it is helpful to avoid this case.

Scottmitch:
@NiteshKant - ptal as some interesting changes made in followup

Issue #336, opened 2019-02-23
InsufficientlySizedExecutorHttpTest


Scottmitch:
failed while building locally

```
org.junit.runners.model.TestTimedOutException: test timed out after 10000 milliseconds
	at java.lang.Object.wait(Native Method)
	at java.lang.Object.wait(Object.java:502)
	at io.servicetalk.concurrent.api.ExecutorCompletionStage$GetListener.blockingWait(ExecutorCompletionStage.java:820)
	at io.servicetalk.concurrent.api.ExecutorCompletionStage.waitingGet(ExecutorCompletionStage.java:652)
	at io.servicetalk.concurrent.api.ExecutorCompletionStage.get(ExecutorCompletionStage.java:509)
	at io.servicetalk.http.netty.InsufficientlySizedExecutorHttpTest.insufficientServerCapacityStreaming(InsufficientlySizedExecutorHttpTest.java:128)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.rules.ExpectedException$ExpectedExceptionStatement.evaluate(ExpectedException.java:239)
	at io.servicetalk.concurrent.internal.ServiceTalkTestTimeout$TimeoutStatement$CallableStatement.call(ServiceTalkTestTimeout.java:167)
	at io.servicetalk.concurrent.internal.ServiceTalkTestTimeout$TimeoutStatement$CallableStatement.call(ServiceTalkTestTimeout.java:159)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:748)
```

Scottmitch:
[...]

PR #977, opened 2020-03-23
Initial support for windows development

### LINKED ISSUE #575: Support building on windows
Attempt to build on Windows with both OpenJDK 8 and 11 produces 9 errors:

- 7 due to Checkstyle crashes (`Unable to create Root Module`),
- 1 due to failure to load test PEM files,
- 1 due to some weird file system permission issue when writing jUnit result.

To be able to build on Windows and add `gradlew.bat` scripts more investigation of the root causes for these issues required.

Scottmitch:
Motivation:
ServiceTalk currently doesn't build on windows. This can be limiting for
folks who are comfortable with that development environment.

Modifications:
- grpc-protoc project needs refactoring of the plugin in order to
execute on unix and windows platforms. The distribution is now
consistent with a script as the executable and co-located uber jar
containing the plugin logic.
- ignore/modify tests that failed when running locally on windows, there
will be followup PRs for other tests that sporadically fail.

Result:
ServiceTalk can now be built on windows.

Resolves #575.

### DIFF
--- a/buildSrc/src/main/java/io/servicetalk/internal/build/ExecutableBuilder.java
+++ b/buildSrc/src/main/java/io/servicetalk/internal/build/ExecutableBuilder.java
@@ -41,10 +41,10 @@ public final class ExecutableBuilder {
     public static void buildWindowsExecutable(String uberJarName, File outputFile) throws IOException {
         prepareOutputFile(outputFile);
         try(FileOutputStream execOutputStream = new FileOutputStream(outputFile)) {
-            execOutputStream.write(("@ECHO OFF\n" +
-                    "pushd %~dp0\n" +
-                    "java -jar " + uberJarName + " %*\n" +
-                    "popd\n").getBytes(StandardCharsets.US_ASCII));
+            execOutputStream.write(("@ECHO OFF\r\n" +
+                    "pushd %~dp0\r\n" +
+                    "java -jar " + uberJarName + " %*\r\n" +
+                    "popd\r\n").getBytes(StandardCharsets.US_ASCII));
         }
         finalizeOutputFile(outputFile);
     }
--- a/buildSrc/src/main/java/io/servicetalk/internal/build/ExecutableBuilder.java
+++ b/buildSrc/src/main/java/io/servicetalk/internal/build/ExecutableBuilder.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright © 2020 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.internal.build;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
+
+public final class ExecutableBuilder {
+
+    private ExecutableBuilder() {
+        // no instances
+    }
+
+    public static void buildUnixExecutable(String uberJarName, File outputFile) throws IOException {
+        prepareOutputFile(outputFile);
+        try(FileOutputStream execOutputStream = new FileOutputStream(outputFile)) {
+            execOutputStream.write(("#!/bin/sh\n" +
+                    "pushd $(dirname \"$0\") > /dev/null\n" +
+                    "exec java -jar " + uberJarName + " \"$@\"\n" +
+                    "popd > /dev/null\n").getBytes(StandardCharsets.US_ASCII));
+        }
+        finalizeOutputFile(outputFile);
+    }
+
+    public static void buildWindowsExecutable(String uberJarName, File outputFile) throws IOException {
+        prepareOutputFile(outputFile);
+        try(FileOutputStream execOutputStream = new FileOutputStream(outputFile)) {
+            execOutputStream.write(("@ECHO OFF\n" +
+                    "pushd %~dp0\n" +
+                    "java -jar " + uberJarName + " %*\n" +
+                    "popd\n").getBytes(StandardCharsets.US_ASCII));
+        }
+        finalizeOutputFile(outputFile);
+    }
+
+    public static String addExecutablePostFix(String rawName, boolean isWindows) {
+        return rawName + (isWindows ? ".bat" : ".sh");
+    }
+
+    private static void prepareOutputFile(File outputFile) throws IOException {
+        if (!outputFile.exists()) {
+            if (!outputFile.getParentFile().isDirectory() && !outputFile.getParentFile().mkdirs()) {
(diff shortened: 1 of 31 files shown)


servicetalk-grpc-protoc/build.gradle
```diff
@@ -47,42 +50,78 @@ shadowJar {
...
+  inputs.files shadowJar.outputs.files
+  outputs.file outputFile
+
+  doLast {
+    buildWindowsExecutable(grpcPluginUberJarName, outputFile)
+  }
+}
+
+// we attempt to generate both grpc executables when on windows, and don't publish from windows anyways.
+tasks.withType(PublishToMavenRepository) {
```
idelpivnitskiy: `PublishToMavenRepository` does not cover `PublishToMavenLocal`. Should we use `AbstractPublishToMaven` type instead?
https://docs.gradle.org/current/javadoc/org/gradle/api/publish/maven/tasks/AbstractPublishToMaven.html

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MultiAddressUrlHttpClientTest.java
```diff
@@ -92,6 +93,7 @@ public static void beforeClass() throws Exception {
         afterClassCloseables = newCompositeCloseable();
 
         client = afterClassCloseables.append(HttpClients.forMultiAddressUrl()
+                .socketOption(CONNECT_TIMEOUT, 1) // windows default connect timeout is seconds, we want to fail fast.
```
idelpivnitskiy: Could you please elaborate why this is necessary? Does it cause issues with the test if windows has different default value here?

buildSrc/src/main/java/io/servicetalk/internal/build/ExecutableBuilder.java
```diff
@@ -0,0 +1,73 @@
...
+        finalizeOutputFile(outputFile);
+    }
+
+    public static void buildWindowsExecutable(String uberJarName, File outputFile) throws IOException {
+        prepareOutputFile(outputFile);
+        try(FileOutputStream execOutputStream = new FileOutputStream(outputFile)) {
+            execOutputStream.write(("@ECHO OFF\n" +
+                    "pushd %~dp0\n" +
+                    "java -jar " + uberJarName + " %*\n" +
+                    "popd\n").getBytes(StandardCharsets.US_ASCII));
```
idelpivnitskiy: Should we use `\r\n` for windows?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/InsufficientlySizedExecutorHttpTest.java
```diff
@@ -116,6 +116,7 @@ public void insufficientServerCapacityStreaming() throws Exception {
         insufficientServerCapacityStreaming0();
     }
 
+    // TODO Windows seeing CHANNEL_CLOSED_INBOUND. Is 1 thread enough with control events (e.g. close)?
```
idelpivnitskiy: Consider creating an issue, otherwise it's hard to track.

servicetalk-grpc-protoc/build.gradle
```diff
@@ -47,42 +50,78 @@ shadowJar {
...
   publications {
     mavenJava {
       artifact(buildExecutable.outputs.files.singleFile) {
         classifier = "linux-x86_64"
-        extension = "exe"
+        extension = "sh"
         builtBy buildExecutable
       }
       artifact(buildExecutable.outputs.files.singleFile) {
         classifier = "osx-x86_64"
```
idelpivnitskiy: What is the expected output for linux and osx classifiers? After running it locally, in `servicetalk-grpc-protoc/build/buildExecutable` folder I see only:
- protoc-gen-servicetalk_grpc.sh
- servicetalk-grpc-protoc-0.26.0-SNAPSHOT-all.jar
Should I see any files with classifiers?

idelpivnitskiy: Weird, in `servicetalk-grpc-protoc/build/buildExecutable` I see only `protoc-gen-servicetalk_grpc.sh.
But when I do `./gradlew publishToMavenLocal` then I see that it publishes two scripts:
1. `servicetalk-grpc-protoc-0.26.0-SNAPSHOT-linux-x86_64.sh`
2. `servicetalk-grpc-protoc-0.26.0-SNAPSHOT-osx-x86_64.sh`
But their content is identical. I think we may have only one `.sh` script.

idelpivnitskiy: Tested with the `mavenLocal`, looks like os-classifier is necessary and we need to have separate scripts for all platforms.

servicetalk-grpc-protoc/build.gradle
Scottmitch: I actually want `PublishToMavenLocal` to work for local testing. This is just meant to highlight the fact that we can generate all artifacts from a single architecture (e.g. just a jar and a script for windows). It could be removed but the hope is that it helps clarify expectations.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MultiAddressUrlHttpClientTest.java
Scottmitch: yes, tests which expect connection failures to trigger retries and ultimately failures on my windows installation were taking 2 seconds to fail each connection attempt. This often lead to test timeouts when we do 5 retries and have a 10 second test timeout.

buildSrc/src/main/java/io/servicetalk/internal/build/ExecutableBuilder.java
Scottmitch: I tested with the `\n` but yes let me change to `\r\n`.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/InsufficientlySizedExecutorHttpTest.java
Scottmitch: actually one already exists :) https://github.com/apple/servicetalk/issues/336

Scottmitch: I'll just ignore this test for now...

servicetalk-grpc-protoc/build.gradle
Scottmitch: yes for local development/builds we use a single script targeted at the current platform in `servicetalk-grpc-protoc/build/buildExecutable`. However for maven we publish both scripts with classifiers to disambiguate and this is also what the grpc-gradle-plugin expects too.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpServerMultipleRequestsTest.java
```diff
@@ -65,6 +66,7 @@
     @Rule
     public final Timeout timeout = new ServiceTalkTestTimeout();
 
+    @Ignore("todo NettyHttpServer repeat WriteStreamSubscriber issues")
```
Scottmitch: will be replaced by reference to https://github.com/apple/servicetalk/pull/977

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MultiAddressUrlHttpClientTest.java
idelpivnitskiy: Thanks!

PR #980, opened 2020-03-23
HTTP Client Pipelining fullduplex

Scottmitch:
Motivation:
The queue which orders and sequences HTTP pipelining on the client side
has been shown to lead to duplicate subscribe exceptions and also
doesn't support full duplex read/write.

Modifications:
- The first was related to DefaultNettyPipelinedConnection
which may have enqueued out of order which lead to a duplicate subscribe
exception on NettyChannelPublisher. The DefaultNettyPipelinedConnection
was rewritten to avoid using the SequentialTaskQueue (which relies upon
thread locals for single item execution and async completion) and also
doesn't allow for full duplex (write must finish before read starts),
and those issues are both fixed.
- ClientClosureRaceTest is intentionally closing the server connection
abrubtley, which may result in writing a request that is not
automatically retryable. The retry strategy should always retry in this
case or else the test may hang/fail.

Result:
HTTP client now support full duplex pipelining.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,628 @@
...
+            Node next = prev.next;
+            while (next != null) {
+                prev = next;
+                next = next.next;
+            }
+            return prev;
+        }
+    }
+
+    private static final class DelayedSubscribeCompletable extends Completable {
```
Scottmitch: we could consider moving this to `concurrent.internal` in the future, but for now I just left them internal here.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/PipelinedStreamingHttpConnection.java
```diff
@@ -19,38 +19,35 @@
...
                 config.maxPipelinedRequests(), executionContext, reqRespFactory, config.headersFactory());
-        this.nettyConnection = connection;
     }
 
     @Override
     protected Publisher<Object> writeAndRead(Publisher<Object> requestStream,
                                              @Nullable final FlushStrategy flushStrategy) {
         if (flushStrategy == null) {
-            return connection.request(requestStream);
+            return connection.write(requestStream, connection::defaultFlushStrategy,
```
Scottmitch: I wanted to keep the new API exposure for the `NettyPipelinedConnection` low for now, but I think this is worth revisiting as part of the larger "how is state propagated on the server https://github.com/apple/servicetalk/issues/981" which may provide opportunities to clean this up.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -537,6 +503,8 @@ public void channelWritabilityChanged(ChannelHandlerContext ctx) {
             if (ctx.channel().isWritable()) {
                 connection.channelOutboundListener.channelWritable();
             } else if (connection.flushStrategyHolder.currentStrategy().shouldFlushOnUnwritable()) {
+                // TODO(scott): if we have a flush per write operation, shouldFlushOnUnwritable is more challenging.
```
Scottmitch: this is also related to the larger "state management on the server" discussion too https://github.com/apple/servicetalk/issues/981

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/NettyPipelinedConnectionTest.java
```diff
@@ -0,0 +1,379 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class NettyPipelinedConnectionTest {
```
NiteshKant: I haven't mapped myself but have we made sure that all previous tests from the previous test(`DefaultNettyPipelinedConnectionTest`) are represented here?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteDemandEstimators.java
```diff
@@ -0,0 +1,34 @@
...
+        // no instance
+    }
+
+    /**
+     * Returns a new instance of a default implementation of {@link WriteDemandEstimator}.
+     *
+     * @return A new instance of a default implementation of {@link WriteDemandEstimator}.
+     */
+    public static WriteDemandEstimator newDefaultEstimator() {
+        return new MaxSizeBasedWriteDemandEstimator();
```
idelpivnitskiy: Since we have only a single implementation of `WriteDemandEstimator` atm, and both classes (this and `MaxSizeBasedWriteDemandEstimator`) are in the same `io.servicetalk.transport.netty.internal` package, can we just make `MaxSizeBasedWriteDemandEstimator` public instead of introducing new factory? Because the code is internal, we can add it later if necessary.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
Scottmitch: more investigation is required, so wanted to make a mental note. however if this is the only thing preventing shifting to a model of flush-on-request I would be willing to compromise on this (assuming we can just always flush in this scenario or otherwise avoid deadlock).

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteDemandEstimators.java
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.transport.netty.internal;
+
+/**
+ * Utility methods associated with {@link WriteDemandEstimator}.
```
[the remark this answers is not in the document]
Scottmitch: the motivation was to decouple utility/factory methods from the interface definition. there is currently only a single factory method but if there are alternative implementations or utilities later this provides a place for them to be exposed.

servicetalk-tcp-netty-internal/src/test/java/io/servicetalk/tcp/netty/internal/TcpServerBinderConnectionAcceptorTest.java
```diff
@@ -147,7 +148,7 @@ public void testAcceptConnection() {
         try {
             NettyConnection<Buffer, Buffer> connection = client.connectBlocking(CLIENT_CTX, serverAddress);
             final Buffer buffer = connection.executionContext().bufferAllocator().fromAscii("Hello");
-            connection.writeAndFlush(buffer).toFuture().get();
+            connection.write(Publisher.from(buffer)).toFuture().get();
```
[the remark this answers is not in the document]
Scottmitch: `flushOnEach()` is the default strategy, and this test isn't trying to test flushing strategies, so I think it makes things simpler to not try to change the flush strategy.

servicetalk-tcp-netty-internal/src/test/java/io/servicetalk/tcp/netty/internal/TcpConnectorTest.java
```diff
@@ -58,7 +59,8 @@ public void testWriteAndRead() throws Exception {
 
     private static void testWriteAndRead(NettyConnection<Buffer, Buffer> connection)
             throws ExecutionException, InterruptedException {
-        connection.writeAndFlush(connection.executionContext().bufferAllocator().fromAscii("Hello")).toFuture().get();
+        connection.write(
```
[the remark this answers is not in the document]
Scottmitch: (lets discuss above on https://github.com/apple/servicetalk/pull/980#discussion_r397540030)

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ClientClosureRaceTest.java
```diff
@@ -122,76 +122,68 @@ public void stopServer() throws Exception {
...
     private void runIterations(Callable<Single<?>> test) throws Exception {
         int count = 0;
         try {
-            while (!receivedExpectedError && count < ITERATIONS) {
-                try {
-                    count++;
-                    Object response = test.call().toFuture().get();
-                    LOGGER.debug("Response {} = {}", count, response);
-                } catch (Exception e) {
-                    if (isAllowableError(e)) {
```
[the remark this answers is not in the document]
Scottmitch: The retry strategy changed [1]. We no longer expect exceptions and instead rely upon timeouts.

[1]
> ClientClosureRaceTest is intentionally closing the server connection
abrubtley, which may result in writing a request that is not
automatically retryable. The retry strategy should always retry in this
case or else the test may hang/fail.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/PipelinedStreamingHttpConnection.java
```diff
@@ -19,38 +19,35 @@
...
         } else {
-            // Using the Writer abstraction here defers updating the flush strategy until just before this request is
-            // written.
-            return connection.request(() -> {
+            // TODO(scott): if we can remove the flush state on the connection we can simplify the control flow here.
+            return Publisher.defer(() -> {
                 final Cancellable resetFlushStrategy = connection.updateFlushStrategy(
                         (prev, isOriginal) -> isOriginal ? flushStrategy : prev);
-                return nettyConnection.write(requestStream).afterFinally(resetFlushStrategy::cancel);
+                return connection.write(requestStream, connection::defaultFlushStrategy,
```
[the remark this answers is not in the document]
Scottmitch: the control flow here is complicated, and something I want to explore improving as part of #981, but I wanted to keep the scope of this PR limited.

the flush strategy passed in will get conditionally applied via `connection.updateFlushStrategy` and may not always be applied depending upon the current connection level flush strategy which maybe set externally on the connection [see FlushStrategyForClientApiTest#aggregatedApiShouldNotOverrideExplicit](https://github.com/apple/servicetalk/blob/master/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyForClientApiTest.java#L100-L101).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/PipelinedStreamingHttpConnection.java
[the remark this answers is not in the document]
Scottmitch: done. minor clarification we are dealing with a `NettyPipelinedConnection` which is not a `NettyConnection`, also discussed above https://github.com/apple/servicetalk/pull/980#discussion_r396835362

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,628 @@
...
+     *
+     * @param connection {@link NettyConnection} requests to which are to be pipelined.
+     */
+    NettyPipelinedConnection(NettyConnection<Resp, Req> connection) {
+        this.connection = requireNonNull(connection);
+    }
+
+    /**
+     * Do a write operation in a pipelined fashion.
+     * @param requestPublisher {@link Publisher} producing the request(s) to write.
```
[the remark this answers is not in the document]
Scottmitch: reworded. ptal.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,628 @@
...
+                            break;
+                        }
+                    }
+                } else if (stateUpdater.compareAndSet(this, currentState, DRAINING_SUBSCRIBERS)) {
+                    assert currentState != DRAINING_SUBSCRIBERS;
+                    @SuppressWarnings("unchecked")
+                    PublisherSource.Subscriber<? super T>[] queue =
+                            (PublisherSource.Subscriber<? super T>[]) currentState;
+                    for (PublisherSource.Subscriber<? super T> next : queue) {
+                        publisher.subscribe(next);
```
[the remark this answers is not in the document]
Scottmitch: `DelayedSubscribePublisher` is meant to be a generic implementation which supports multiple subscribes, and lets the downstream `Publisher` determine how multiple subscribes are handled. In this case multiple subscribes are not expected.

discussed in https://github.com/apple/servicetalk/pull/980#discussion_r396834951 ... I could extract this to `concurrent.internal` package and add tests targeted at this class but not sure it is worth the additional API exposure with just a single use case. wdyt?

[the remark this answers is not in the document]
Scottmitch: I will at least extract these delayed* classes and write tests for them ... we can then discuss if they should live in `concurrent.api.internal` or just stay local to `http` package for now.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,628 @@
...
+                          final Supplier<WriteDemandEstimator> writeDemandEstimatorSupplier) {
+        return Publisher.defer(() -> {
+            // Lazy modification of local state required (e.g. nodes, delayed subscriber, queue modifications)
+
+            // Setup read side publisher and nodes
+            DelayedSubscribePublisher<Resp> delayedResponsePublisher = new DelayedSubscribePublisher<>(
+                    toSource(connection.read()));
+            Node readNode = new Node(delayedResponsePublisher::processSubscribers);
+            Publisher<Resp> composedResponsePublisher =
+                    delayedResponsePublisher.liftSync(new ReadPopNextOperator(readNode));
```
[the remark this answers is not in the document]
Scottmitch:
```java
    /**
     * Logically equivalent to {@link Publisher#afterFinally(Runnable)} but relies upon internal queue CAS operations
     * to prevent multiple executions (e.g. reduces a CAS operation).
     */
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,628 @@
...
+ * <p>
+ * Pipelining allows to have concurrent requests processed on the server but still deliver responses in order.
+ * This eliminates the need for request-response correlation, at the cost of head-of-line blocking.
+ * @param <Req> Type of requests sent on this connection.
+ * @param <Resp> Type of responses read from this connection.
+ */
+final class NettyPipelinedConnection<Req, Resp> implements NettyConnectionContext {
+    private static final Logger LOGGER = LoggerFactory.getLogger(NettyPipelinedConnection.class);
+    @SuppressWarnings("rawtypes")
+    private static final AtomicReferenceFieldUpdater<NettyPipelinedConnection, Node>
```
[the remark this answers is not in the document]
normanmaurer: +1... I would keep it simple if possible. At first look this looks like over-engineered imho.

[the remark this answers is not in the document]
Scottmitch: I don't believe the complexity in the PR is substantially increased/decreased from what was here previously. There are two high level components of this PR (both of which existed previously):

A. relationship between read/write operations
B. Mpsc queue that only consumes and executes a single element at a time, and async completion of the current events triggers a pop of the current element and execution of the next element

(A) This PR modifies the previous control flow in [DefaultNettyPipelinedConnection](https://github.com/apple/servicetalk/blob/master/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyPipelinedConnection.java) to allow for full duplex, where previously the read could only start after the write completely finished.

(B) This PR replaces [SequentialTaskQueue](https://github.com/apple/servicetalk/blob/master/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SequentialTaskQueue.java) which attempted todo the same thing (Mpsc Single Runnable Active, async completion starts the next Runnable). However the previous queue relied upon `ThreadLocals` and may not make progress if the completion thread is different than the thread that started the task. This PR fixes a potential correctness issue.

The interesting characteristics of the queue are:
- offer needs to know when the queue was previously empty -> the first element needs to be executed/run
- pop needs to be correlated to the current active task (e.g. only a task that is executed/run should trigger pop, or else sequencing is off) and it also needs to execute the next task (if one exists)

this queue implementation achieves those characteristics with atomic operations to update linked nodes, and a tail pointer with trailing atomic updates. It could be done in other ways (e.g. `synchronized`) but using existing `MPSC` queues won't give us the trigger points to know when "this offer was the first element" and "this specific element needs to be popped, and I need the next element in order relative to the related read/write".

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/NettyPipelinedConnectionTest.java
Scottmitch: Yes, the test sequences scenarios were adapted from `DefaultNettyPipelinedConnectionTest` and additional ones were added on top.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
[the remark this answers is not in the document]
Scottmitch: (discussed offline with @NiteshKant and clarified control flow)

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,628 @@
...
+            // 1. Another thread appended this head, but has not yet updated the tail. In this case the tail will be
+            // stale (e.g. pointing to head node that has already been processed) and corrected by future inserts.
+        }
+    }
+
+    private void safeProcessSubscribers(Runnable delayedSource) {
+        try {
+            delayedSource.run();
+        } catch (Throwable cause) {
+            connection.closeAsync().subscribe();
```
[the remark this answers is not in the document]
Scottmitch: discussed offline, added some additional comments, and exception handling.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,353 @@
...
+                    } catch (Throwable cause) {
+                        closeOnError(connection, delayedWriteCompletable, cause);
+                    }
+                }
+            };
+
+            try {
+                writeQueue.offer(writeNode);
+            } catch (Throwable cause) {
+                // The writeNode's run method handles exceptions as a result of Subscribe/external calls. An Exception
```
Scottmitch: @NiteshKant - this exception handling was added for demonstration purposes but is somewhat superfluous bcz the queue offer doesn't do any allocations (e.g. node allocation done outside the scope), and we don't always exhaustively protect against OOME type errors.

also the try/catch around the insertions are also somewhat much, because the `Publisher`  (or completable) will catch exceptions thrown during `subscribe` (in the internal `handleSubscribe` method). Let me know if you think it is worth keeping all of these cases.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/MpmcSequentialRunQueue.java
```diff
@@ -0,0 +1,185 @@
...
+ * This queue allows for Muli-Producer Multi-Consumer (Mpmc) threading semantics while also invoking the head
+ * {@link Node}'s {@link Node#run()} in a serial fashion. The {@link Node#run()} will eventually trigger a
+ * {@link #poll(Node)} (possibly asynchronously on another thread) which will invoke {@link Node#run()} on the next head
+ * (assuming one exists).
+ * <p>
+ * Below is the expected interaction pattern and lifecycle of a Node:
+ *     <pre>{@link #offer(Node)} -> {@link Node#run()} -> {@link #poll(Node)}</pre>
+ * <p>
+ * Although this queue supports Multi-Consumer threading semantics the {@link #poll(Node)} is typically only invoked
+ * from a single thread (assuming successful runnable completion), it maybe invoked multiple times (potentially from
```
normanmaurer, line 33: nit: s/maybe/may be/

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/MpmcSequentialRunQueue.java
```diff
@@ -0,0 +1,185 @@
...
+            tailUpdater = newUpdater(MpmcSequentialRunQueue.class, Node.class, "tail");
+
+    @Nullable
+    private volatile Node tail;
+
+    /**
+     * Offer {@link Node} to this queue.
+     * @param node The {@link Node} to append. Must only exclusively be offered to this queue instance.
+     */
+    void offer(final Node node) {
```
normanmaurer, line 47: nit: do we need to guard against `null` ? As otherwise we could end up updating the `tail` with null and then fail once we try to run `node.run()` (if node is null). I think at least we should add an `assert`.

normanmaurer, line 47: nit: I think `offer` is a bit of an odd name for what is happening in this method as it may queue or may run the node. So why not call it `execute(...)` just as `Executor` is doing (Which is very similar in terms of behaviour).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/MpmcSequentialRunQueue.java
```diff
@@ -0,0 +1,185 @@
...
+                // to see if it changed, and if not attempt to walk the list and update the tail pointer.
+                final Node newTail = this.tail;
+                if (newTail == currentTail && offerUpdateStaleTail(node, currentTail)) {
+                    break;
+                }
+            }
+        }
+    }
+
+    private boolean offerUpdateStaleTail(final Node node, final Node oldTail) {
```
normanmaurer, line 88: same as above... I think the method name is a bit odd as it may queue but also may run.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/MpmcSequentialRunQueue.java
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * Pop {@code node} from this queue. This will invoke {@link Node#run()} on the next head of the queue
+     * (if one exists).
+     * <p>
+     * Typically invoked by a single thread but maybe invoked by multiple threads, and maybe invoked multiple times.
+     * Re-entry from {@link Node#run()} from another thread is permitted.
+     * @param head A {@link Node} which has been passed to {@link #offer(Node)}, and whose {@link Node#run()} has been
+     * invoked by this queue.
+     */
+    void poll(final Node head) {
```
normanmaurer, line 118: same comment with the method name.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/MpmcSequentialRunQueue.java
```diff
@@ -0,0 +1,185 @@
...
+            tailUpdater.compareAndSet(this, oldTail, currentTail);
+        }
+        // Best effort to update/clear the tail, and failure is OK because:
+        // 1. Another thread is in offer and already patched up the tail pointer and we will read the new
+        // tail on the next loop iteration.
+        return false;
+    }
+
+    /**
+     * Pop {@code node} from this queue. This will invoke {@link Node#run()} on the next head of the queue
```
normanmaurer, line 110: nit: `{@code head}`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/MpmcSequentialRunQueue.java
```diff
@@ -0,0 +1,185 @@
...
+         * the EMPTY_NODE's next must point to itself so that attempts to {@link #append(Node)} on it will fail,
+         * and also be detected as {@link #isPopped()} and not prevent any new nodes from being
+         * {@link #offer(Node) offered}.
+         */
+        private static final Node EMPTY_NODE = new Node(false) {
+            @Override
+            void run() {
+            }
+        };
+        private static final AtomicReferenceFieldUpdater<Node, Node> nextUpdater =
```
normanmaurer, line 148: nit: you could just have `Node` extend `AtomicReference<Node>` and remove the whole updater dance. This will also remove some overhead in terms of access checks.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    @Override
+    public Channel nettyChannel() {
+        return connection.nettyChannel();
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getName() + '(' + connection + ')';
```
normanmaurer, line 180: nit: `getClass.getSimpleName()`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/MpmcSequentialRunQueue.java
```diff
@@ -0,0 +1,185 @@
...
+                // 2. The consumer thread has seen the link from the old tail to the new node, processed node,
+                // popped node from the list (updated node's next to point to EMPTY_NODE), another producer thread
+                // appends a new node, sees the tail is popped, and updates the tail reference via CAS.
+                tailUpdater.compareAndSet(this, currentTail, node);
+                break;
+            } else if (currentTail.isPopped()) {
+                // A previously appended node was processed, and popped before updating the tail after append. In that
+                // case the tail maybe pointing to an invalid node and we clear it out.
+                if (tailUpdater.compareAndSet(this, currentTail, node)) {
+                    node.run();
```
NiteshKant, line 70: Looks like this can race with `poll()` and run concurrent tasks.


```java
tail = node 1; (node1 currently running)
Thread 1:  
offer (node 2)
 - node1.append(node2) <- success
    - suspended
Thread 2 (concurrent with Thread 1):
  offer(node3)
  fails node1.append() as raced with Thread 1
  suspended

Thread 3:
  node 1 returns; node1.pop() returns node 2;
     node 2.run();
     returns;

Thread 2 resumes:
    node1.isPopped() <- true
    tailUpdater.cas(node1, node3) <- true
      node3.run(); <-- we are running 2 nodes concurrently (Thread 2 => node 3, Thread 3 => node 2)
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,378 @@
...
+                                } catch (Throwable cause) {
+                                    // We started the write, but failed to setup the read. This is considered fatal as
+                                    // we will be out of sync for delivering future read responses.
+                                    closeConnection(rSubscriber, cause);
+                                }
+                            }
+                        }));
+            } catch (Throwable cause) {
+                // We failed to setup the write operation, which means we also failed to setup the read operation.
+                // This failure maybe recoverable as our internal state isn't corrupted, so just propagate the error.
```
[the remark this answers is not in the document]
Scottmitch: `connection.close()` sgtm

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,378 @@
...
+                deliverTerminalFromSource(subscriber, cause);
```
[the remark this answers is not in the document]
Scottmitch: lets just close the connection as you suggested on your next comment.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,378 @@
...
+                            writeDemandEstimatorSupplier);
+                } catch (Throwable cause) {
+                    deliverTerminalFromSource(subscriber, cause);
+                    return;
+                }
+
+                try {
+                    writeQueue.offer(node);
+                } catch (Throwable cause) {
+                    // The queue offer is not expected to throw, but if it does we cannot poll the WriteNode to recover.
```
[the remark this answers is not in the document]
Scottmitch: If a node is added and not removed yes the queue would accumulate. The queue/executor can be made more robust to catch exceptions when synchronously calling `run()` but if the "poll" method isn't called in the async control flow the same issue will occur. Do you see code paths which may lead to this? we are also closing the connection which should lead to the queues being discarded.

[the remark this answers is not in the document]
Scottmitch: I adjusted the queue APIs to:
- capture synchronous exceptions
- make the "execute next task" condition more clear (e.g. provided as an argument to the `run(..)` method).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,378 @@
...
+
+        @Override
+        void run() {
+            final PublisherSource<Resp> src;
+            try {
+                src = toSource(connection.read().liftSync(new ReadPopNextOperator(this)));
+            } catch (Throwable cause) {
+                // We started the write, but failed to setup the read. This is considered fatal as we will be out of
+                // sync for delivering future read responses.
+                closeConnection(subscriber, cause);
```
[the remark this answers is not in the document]
Scottmitch: yes! fixed.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/MpmcSequentialRunQueue.java
Scottmitch: great find 🔥🔐! let me just use the simplified queue approach as you suggested offline. that way we can get this in and optimize the queue elements later if necessary.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/ConcurrentUtils.java
```diff
@@ -31,11 +31,46 @@
...
 
+    /**
+     * Acquire a lock that is exclusively held with no re-entry, but attempts to acquire the lock while it is
+     * held can be detected by {@link #releasePendingLock(AtomicIntegerFieldUpdater, Object)}.
+     * @param lockUpdater The {@link AtomicIntegerFieldUpdater} used to control the lock state.
+     * @param owner The owner of the lock object.
+     * @param <T> The type of object that owns the lock.
+     * @return {@code true} if the lock was acquired, {@code false} otherwise.
+     */
+    public static <T> boolean acquirePendingLock(AtomicIntegerFieldUpdater<T> lockUpdater, T owner) {
```
Scottmitch, line 48: @NiteshKant - note these are utilities that are also shared with https://github.com/apple/servicetalk/pull/1011 and https://github.com/apple/servicetalk/pull/1014 ... which ever PR lands first will bring them in and I will rebase to avoid duplication.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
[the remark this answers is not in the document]
Scottmitch: custom queue implementation has since been removed ... thanks for the feedback!

Scottmitch:
@normanmaurer @NiteshKant - thanks for reviewing the custom queue ... I took your advice and reverted back to an off-the-shelf queue to simplify. PTAL.

servicetalk-benchmarks/src/jmh/java/io/servicetalk/http/netty/NettyPipelinedConnectionBenchmark.java
```diff
@@ -0,0 +1,268 @@
...
+import static io.servicetalk.concurrent.api.SourceAdapters.toSource;
+
+@Fork(value = 1)
+@State(Scope.Benchmark)
+@Warmup(iterations = 5, time = 3)
+@Measurement(iterations = 5, time = 3)
+@BenchmarkMode(Mode.Throughput)
+public class NettyPipelinedConnectionBenchmark {
+    static {
+        AsyncContext.disable(); // reduce noise in benchmarks.
```
normanmaurer, line 72: nit: Should we enable it again after we are done with the benchmark ? IF so we may just do it in `setup()` and `tearDown()` methods

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,344 @@
...
+
+    /**
+     * New instance.
+     *
+     * @param connection {@link NettyConnection} requests to which are to be pipelined.
+     */
+    NettyPipelinedConnection(NettyConnection<Resp, Req> connection) {
+        this.connection = requireNonNull(connection);
+        writeQueue = newUnboundedMpscQueue();
+        readQueue = newUnboundedMpscQueue();
```
normanmaurer, line 81: nit: I wonder if unbounded may be risky here... But I guess it should be fine ?

Scottmitch: We apply limits at a higher level in the client.

servicetalk-benchmarks/src/jmh/java/io/servicetalk/http/netty/NettyPipelinedConnectionBenchmark.java
Scottmitch: we currently don't expose an "enable" it is just intended as a one-time "disable" ... I think this is OK for benchmarks.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,362 @@
...
+            this.flushStrategySupplier = flushStrategySupplier;
+            this.writeDemandEstimatorSupplier = writeDemandEstimatorSupplier;
+        }
+
+        void run() {
+            final PublisherSource<Resp> src;
+            try {
+                src = toSource(connection.write(requestPublisher, flushStrategySupplier,
+                        writeDemandEstimatorSupplier)
+                        .afterFinally(() -> {
```
NiteshKant, line 243: Does it have to be `afterFinally()` or we can use `beforeFinally()`? A positive of using `before*()` here is that any exception will bubble up to the subscriber.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,362 @@
...
+                            WriteTask nextWriteTask = pollWithLockAcquired(writeQueue, writeQueueLockUpdater);
+                            if (nextWriteTask != null) {
+                                nextWriteTask.run();
+                            }
+                        }).merge(new Publisher<Resp>() {
+                            @Override
+                            protected void handleSubscribe(final Subscriber<? super Resp> rSubscriber) {
+                                final Subscriber<? super Resp> firstReadSubscriber;
```
NiteshKant, line 251: `firstReadSubscriber` => `nextReadSubscriber`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,362 @@
...
+    @Override
+    public FlushStrategy defaultFlushStrategy() {
+        return connection.defaultFlushStrategy();
+    }
+
+    private void closeConnection(final Subscriber<? super Resp> subscriber, final Throwable cause) {
+        toSource(connection.closeAsync().concat(Publisher.<Resp>failed(cause))).subscribe(subscriber);
+    }
+
+    private void tryStartRead(@Nullable Subscriber<? super Resp> subscriber) {
```
NiteshKant, line 207: nit: move this method to the `WriteTask` as thats where it is used.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,362 @@
...
+
+    private void handleWriteSetupError(Subscriber<? super Resp> subscriber, Throwable cause) {
+        try {
+            closeConnection(subscriber, cause);
+        } finally {
+            // the lock has been acquired!
+            do {
+                WriteTask nextWriteTask;
+                while ((nextWriteTask = writeQueue.poll()) != null) {
+                    deliverTerminalFromSource(nextWriteTask.subscriber, cause);
```
NiteshKant, line 279: Can we avoid calling external methods inside `finally`? eg: `deliverTerminalFromSource` can throw if both `subscriber.onSubscribe()`  and `subscriber.onError()` throw. Alternative:

```java
private void handleWriteSetupError(Subscriber<? super Resp> subscriber, Throwable cause) {
    try {
        closeConnection(subscriber, cause);
    } catch (Throwable t) {
        subscriber.onSubscribe(EMPTY_SUBSCRIPTION);
        subscriber.onError(t);
    }
    // the lock has been acquired!
    do {
        WriteTask nextWriteTask;
        while ((nextWriteTask = writeQueue.poll()) != null) {
            deliverTerminalFromSource(nextWriteTask.subscriber, cause);
        }
    } while (!releaseLock(writeQueueLockUpdater, this) && tryAcquireLock(writeQueueLockUpdater, this));
}
```

OR just make sure `closeConnection` does not throw and remove the `catch` close above.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,362 @@
...
+     * {@link ConcurrentUtils#tryAcquireLock(AtomicIntegerFieldUpdater, Object)}.
+     * @param item The item to {@link Queue#offer(Object)}.
+     * @param <T> The type of item in the {@link Queue}.
+     * @return {@code null} if the queue was empty, or the lock couldn't be acquired. otherwise the lock has been
+     * acquired and it is the caller's responsibility to release!
+     */
+    @Nullable
+    private <T> T offerTryAcquireLock(final Queue<T> queue,
+          @SuppressWarnings("rawtypes") final AtomicIntegerFieldUpdater<NettyPipelinedConnection> lockUpdater, T item) {
+        queue.add(item);
```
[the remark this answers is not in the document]
Scottmitch: switched to `addAndTryPoll`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
```diff
@@ -0,0 +1,362 @@
...
+     * @return Response {@link Publisher} for this request.
+     */
+    Publisher<Resp> write(final Publisher<Req> requestPublisher,
+                          final Supplier<FlushStrategy> flushStrategySupplier,
+                          final Supplier<WriteDemandEstimator> writeDemandEstimatorSupplier) {
+        // Lazy modification of local state required (e.g. nodes, delayed subscriber, queue modifications)
+        return new Publisher<Resp>() {
+            @Override
+            protected void handleSubscribe(final Subscriber<? super Resp> subscriber) {
+                final WriteTask firstWriteTask;
```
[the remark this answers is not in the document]
Scottmitch: I will change to "next"

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
Scottmitch: `beforeFinally` can trigger before write stream subscriber is cleaned up in some error situations so I'll just stick with `after*` for now.

Scottmitch: to provide more context ....
-  in the event the write publisher terminates with an error `WriteStreamSubscriber.AllWritePromise#terminateSubscriber` closes the closeChannelOutbound before propagating exception to the subscriber which results in `A write is already active on this connection.` exception on `DefaultNettyConnection`. I think the sequencing in incorrect and `closeChannelOutbound` should be after the subscriber is terminated and I will fix this. however then `merge` triggers a cancellation which results in starting the next read task, which subscribes to the `channel.read() ` publisher before the previous subscriber is "detached" from `NettyChannelPublisher` which results in a `DuplicateSubscribeException` ... in this case we want an operator like "do subscriber before, but subscription after" to invoke `tryStartRead`.
- it is also a questioning of sequencing ... if we dequeue the next element before notifying the status of the current element, and there is a synchronous type of failure/completion we will notify completion in reverse order as the queue. we have a queue so "cleaning up local state first" may result in out of order notification

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyPipelinedConnection.java
Scottmitch: `closeConnection` already shouldn't throw, but since there is a lock acquired the cleanup code is put in a finally block.

I don't think `deliverTerminalFromSource` should throw. if it does it may result in duplicate notification from our sources. For example:

```
NeverPublisher#subscribe()
subscriber.onSubscribe(..)
catch(cause)
handleExceptionFromOnSubscribe(..)
subscriber.onError(cause)
catch(cause2)  (in Publisher#handleSubscribe(..))
deliverTerminalFromSource(..)
subscriber.onSubscribe(..) // duplicate call introduced by our error recovery
catch(cause3)
handleExceptionFromOnSubscribe(..)
subscriber.onError(cause3) // duplicate call introduced by our error recovery
```

I'll submit a PR.

NiteshKant: > closeConnection already shouldn't throw, 

Unless `connection.closeAsync()` itself throws? I think we should generally avoid calling external code in `finally` block, it complicates control-flow and hides exception if any code in `try` and in the `finally` block throws.

Scottmitch: I'll just remove the `finally` as it is defensive programming and is targeted at programming errors or runtime failures (e.g. OOME). However I don't think the "don't call external code in finally block" is a hard rule despite it potentially complicating control flow. The goal of "finally" is two independent code blocks need to execute regardless if the first throws. This can be particularly useful to guarantee asynchronous control flow is preserved if something unexpected happens. If exceptions aren't expected from either code block (e.g. programming error occurs as opposed to data flow error), I would argue it doesn't really matter which exception is propagated (if it did matter exceptions are no longer "unexpected"), but we achieved the goal of preserving async control flow. Granted this is a defensive, but if it does occur debugging in an async control flow will be challenging.

In this case if `deliverTerminalFromSource` throws then the control flow isn't correct here because we wouldn't propagate the exception to all pending subscribers, and we should use a safer alternative instead (not convinced this is the desired behavior though #1034). If the `queue.poll()` throws (isn't expect to) then we cannot reliably continue to drain the queue (may get another exception on the next poll attempt) so this is considered fatal and we let the exception propagate.

> Unless connection.closeAsync() itself throws?

Yes any method in `closeConnection()` could throw. However the method returns an async source and throwing isn't the expected way to propagate an error in this case, but may happen in extreme situations (e.g. OOME).

NiteshKant: >> I'll just remove the finally 

👌 thanks

>>  I don't think the "don't call external code in finally block" is a hard rule 

Agreed.

Scottmitch:
thanks for review @NiteshKant @normanmaurer @idelpivnitskiy !

PR #983, opened 2020-03-25
gRPC gradle plugin rework

Scottmitch:
Motivation:
0ebe48ed3948f22bfc003c8a0c551572bb70bbbf divided the
servicetalk-grpc-gradle plugin into two files:
1. an executable script
2. an uber jar with the plugin logic
The executable script assumed the uber jar would be co-located in the
same directory as the uber jar, but that isn't the case in gradle
caches. This means the plugin may fail to execute outside of the maven
m2 repository structure.

Modifications:
- Instead of publishing a static script for each platform which assumes
a co-located uber jar, dynamically generate the executable script
depending upon where the uber jar is resolved from for the local build.

Result:
servicetalk-grpc-gradle works with gradle cache directory structure and
local development.


servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
```diff
@@ -67,6 +62,49 @@ class ServiceTalkGrpcPlugin implements Plugin<Project> {
...
+        uberJarFile = project.configurations.compileOnly.find { it.name.startsWith("servicetalk-grpc-protoc") }
+        if (uberJarFile == null) {
+          throw new IllegalStateException("failed to find the servicetalk-grpc-protoc:$serviceTalkVersion:all")
+        }
+      }
+
+      File scriptExecutableFile
+      try {
+        if (org.gradle.internal.os.OperatingSystem.current().isWindows()) {
+          scriptExecutableFile = File.createTempFile(scriptNamePrefix, ".bat")
```
Scottmitch: generation of temp files is not ideal ... but a reliable way to make sure we use the jar file that is actually resolved for each build. I'm open to alternatives @idelpivnitskiy

servicetalk-examples/grpc/helloworld/build.gradle
```diff
@@ -33,7 +33,7 @@ serviceTalkGrpc {
 
   // The following setting must be omitted in users projects and is necessary here
   // only because we want to use the locally built version of the plugin
-  serviceTalkProtocPluginPath = "${project.rootProject.rootDir}/servicetalk-grpc-protoc/build/buildExecutable/" +
-      io.servicetalk.internal.build.ExecutableBuilder.addExecutablePostFix("protoc-gen-servicetalk_grpc",
-          org.gradle.internal.os.OperatingSystem.current().isWindows())
+  serviceTalkProtocPluginPath =
+      "${project.rootProject.rootDir}/servicetalk-grpc-protoc/build/buildExecutable/servicetalk-grpc-protoc-" +
+          project.version + "-all.jar"
```
idelpivnitskiy: nit (here and in other places): You can avoid concatenation in this way: `"${project.rootProject.rootDir}/servicetalk-grpc-protoc/build/buildExecutable/servicetalk-grpc-protoc-${project.version}-all.jar"`

servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
```diff
@@ -67,6 +62,49 @@ class ServiceTalkGrpcPlugin implements Plugin<Project> {
...
+      // If this project is outside of ServiceTalk's gradle build we need to add an explicit dependency on the
+      // uber jar which contains the protoc logic, as otherwise the grpc-gradle-plugin will only add a dependency
+      // on the executable script
+      File uberJarFile
+      String scriptNamePrefix
+      if (serviceTalkProtocPluginPath) {
+        scriptNamePrefix = "servicetalk-grpc-protoc-" + project.version
+        uberJarFile = new File(serviceTalkProtocPluginPath)
+      } else {
+        scriptNamePrefix = "servicetalk-grpc-protoc-$serviceTalkVersion"
```
idelpivnitskiy: Why `if` branch should use `project.version` but `else` branch uses `$serviceTalkVersion`? Are they different?

servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
```diff
@@ -67,6 +62,49 @@ class ServiceTalkGrpcPlugin implements Plugin<Project> {
...
+        def stGrpcProtocDep =
+            project.getDependencies().create("io.servicetalk:servicetalk-grpc-protoc:$serviceTalkVersion:all")
+        compileOnlyDeps.add(stGrpcProtocDep)
+        testCompileOnlyDeps.add(stGrpcProtocDep)
+
+        uberJarFile = project.configurations.compileOnly.find { it.name.startsWith("servicetalk-grpc-protoc") }
```
idelpivnitskiy: `servicetalk-grpc-protoc` string is used multiple times in this class, consider using a constant

servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
idelpivnitskiy: how grpc-java solves this issue?

servicetalk-examples/grpc/helloworld/build.gradle
Scottmitch: I was getting a check style error in some cases bcz the line was too long so I'll just break it up for now.

idelpivnitskiy: ok, np

servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
Scottmitch: > Are they different?

Yes. if `serviceTalkProtocPluginPath` is set then `serviceTalkVersion == ${serviceTalkVersion}` (or something similar). Basically if `serviceTalkProtocPluginPath` is set the variable isn't reliably available and instead we use the current project version (because you are assumed to be building in the same project anyways ... IIUC this should be true even if using `--include-build`)

servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
Scottmitch: they don't have to deal with this because they write their plugin in c++ and publish a single executable for each platform.

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/744

Scottmitch:

Scottmitch:
another failure attributed to #744

PR #984, opened 2020-03-26
gRPC gradle plugin write script to buildDir

Scottmitch:
Motivation:
servicetalk-grpc-gradle-plugin was writing the executable script files
to the temp directory and not cleaning them up. There was also a bug in
a variable substitution which lead to compile errors.

Modifications:
- Write to the buildDir instead of temp, and don't rewrite the file on
every build if it already exists since the version information is in the
file name.
- Fix the variable naming bug

Result:
No more temp files for servicetalk-grpc-gradle-plugin and one less
variable naming bug.

### DIFF
--- a/servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
+++ b/servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
@@ -21,6 +21,7 @@ import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.tasks.Delete
 import org.gradle.plugins.ide.eclipse.EclipsePlugin
 import org.gradle.plugins.ide.idea.IdeaPlugin
 import org.gradle.plugins.ide.idea.model.IdeaModel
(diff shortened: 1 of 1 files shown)


servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
```diff
@@ -21,6 +21,7 @@ import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.internal.os.OperatingSystem
```
normanmaurer: I am not a gradle export but just from the package name I would suspect this is not really something we should use. Is there nothing that we can use that is not in an internal package ?

Scottmitch: Agreed, using gradle internal APIs is not ideal but was not newly introduced by this PR. We can consider improving in a followup PR after we get the build stable again.

Scottmitch: actually looking at the internal code it is just using `os.name` ... let me try this quick replace for windows and not-windows.

PR #985, opened 2020-03-26
Follow-up enhancements for `servicetalk-grpc-gradle-plugin`

idelpivnitskiy:
Motivation:

There are a few issues with `servicetalk-grpc-gradle-plugin`.

Modifications:

- Fix incorrect plugin name in exception message;
- Correctly verify that `serviceTalkVersion` has correct value as
it's always not empty, but may have unprocessed value;
- Do not add `serviceTalkVersion` to the `scriptExecutableFile`
name as it's not necessary;
- Generate script only for projects that use
`servicetalk-grpc-gradle-plugin` instead of the root project;
- Do not looks for `serviceTalkGrpcProtocGenerateScript` task
before creating it;
- Delete generated script file when users run `clean` task;
- Recreate executable script file on every run to avoid having
an outdated file (for example if gradle changes the hash for
cached artifacts or if users run another project with and
without `--include-build`);

Result:

`servicetalk-grpc-gradle-plugin` behaves correctly.

### DIFF
--- a/servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
+++ b/servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
@@ -79,14 +79,9 @@ class ServiceTalkGrpcPlugin implements Plugin<Project> {
           // uber jar which contains the protoc logic, as otherwise the grpc-gradle-plugin will only add a dependency
           // on the executable script
           File uberJarFile
-          String scriptNamePrefix
           if (serviceTalkProtocPluginPath) {
-            scriptNamePrefix = serviceTalkGrpcProtoc + "-" +
-                // fallback to the current project.version if there is no version in plugin meta
-                (isVersion(serviceTalkVersion) ? serviceTalkVersion : project.version)
             uberJarFile = new File(serviceTalkProtocPluginPath.toString())
           } else {
-            scriptNamePrefix = serviceTalkGrpcProtoc + "-" + serviceTalkVersion
             def stGrpcProtocDep =
                 project.getDependencies().create("io.servicetalk:$serviceTalkGrpcProtoc:$serviceTalkVersion:all")
             compileOnlyDeps.add(stGrpcProtocDep)
@@ -100,7 +95,7 @@ class ServiceTalkGrpcPlugin implements Plugin<Project> {
           }
 
           final boolean isWindows = System.getProperty("os.name").toLowerCase().contains("windows")
-          File scriptExecutableFile = new File("${project.buildDir}/scripts/${scriptNamePrefix}." +
+          File scriptExecutableFile = new File("${project.buildDir}/scripts/${serviceTalkGrpcProtoc}." +
               (isWindows ? "bat" : "sh"))
 
           doFirst {
--- a/servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
+++ b/servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
@@ -36,7 +36,7 @@ class ServiceTalkGrpcPlugin implements Plugin<Project> {
     }
 
     if (project.plugins.hasPlugin("com.google.protobuf")) {
-      throw new InvalidUserCodeException("Plugin `servicetalk-grpc` needs to be applied " +
+      throw new InvalidUserCodeException("Plugin `io.servicetalk.servicetalk-grpc-gradle-plugin` needs to be applied " +
           "*before* plugin `com.google.protobuf`. (The latter is applied automatically by the former.)")
     }
 
@@ -54,7 +54,7 @@ class ServiceTalkGrpcPlugin implements Plugin<Project> {
       def serviceTalkGrpcProtoc = "servicetalk-grpc-protoc"
       def serviceTalkVersion = pluginProperties."implementation-version"
       def serviceTalkProtocPluginPath = extension.serviceTalkProtocPluginPath
-      if (!serviceTalkVersion && !serviceTalkProtocPluginPath) {
+      if (!isVersion(serviceTalkVersion) && !serviceTalkProtocPluginPath) {
         throw new IllegalStateException("Failed to retrieve ServiceTalk version from plugin meta " +
           "and `serviceTalkGrpc.serviceTalkProtocPluginPath` is not set.")
       }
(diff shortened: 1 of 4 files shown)


idelpivnitskiy:
quality validation failed due to docker error

servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
```diff
@@ -64,12 +64,11 @@ class ServiceTalkGrpcPlugin implements Plugin<Project> {
         throw new InvalidUserDataException("Please set `serviceTalkGrpc.protobufVersion`.")
       }
 
-      Project rootProject = project.rootProject != null ? project.rootProject : project
       String serviceTalkGrpcProtocGenerateScriptTaskName = "serviceTalkGrpcProtocGenerateScript"
-      if (rootProject.getTasksByName(serviceTalkGrpcProtocGenerateScriptTaskName, false).isEmpty()) {
-        rootProject.task(serviceTalkGrpcProtocGenerateScriptTaskName, {
+      if (project.getTasksByName(serviceTalkGrpcProtocGenerateScriptTaskName, false).isEmpty()) {
```
Scottmitch: the intention with `rootProject` is that we only need to have a single task which generates the executable script, and have a single script generated in the root project (regardless of how many subprojects apply the plugin). can you clarify why this changed?

servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
```diff
@@ -80,26 +79,23 @@ class ServiceTalkGrpcPlugin implements Plugin<Project> {
           // uber jar which contains the protoc logic, as otherwise the grpc-gradle-plugin will only add a dependency
           // on the executable script
           File uberJarFile
-          String scriptNamePrefix
```
Scottmitch: the rational for putting the version suffix was that if folks update their version of ServiceTalk and the script generation changes, they won't have to clean their build in order to get the updates. Can you clarify the motivation for this change?

servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
```diff
@@ -54,7 +54,7 @@ class ServiceTalkGrpcPlugin implements Plugin<Project> {
       def serviceTalkGrpcProtoc = "servicetalk-grpc-protoc"
       def serviceTalkVersion = pluginProperties."implementation-version"
       def serviceTalkProtocPluginPath = extension.serviceTalkProtocPluginPath
-      if (!serviceTalkVersion && !serviceTalkProtocPluginPath) {
+      if (!isVersion(serviceTalkVersion) && !serviceTalkProtocPluginPath) {
```
idelpivnitskiy: `serviceTalkVersion` is always not empty. But it could have `${projectVersion}` as a value instead of the actual version (e.g. `0.26.0-SNAPSHOT`) when this code executes before `processResources`.

servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
```diff
@@ -224,9 +220,6 @@ class ServiceTalkGrpcPlugin implements Plugin<Project> {
   }
 
   private static boolean createNewScriptFile(File outputFile) throws IOException {
-    if (outputFile.exists()) {
```
Scottmitch: can you clarify the motivation for this change? if the file already exists the assumption is we don't need to try to re-write the file. The script files are (or were before this PR) versioned so no need to re-write. If folks are using SNAPSHOT it is possible the contents have changed, but this will mostly be just for us during development and we can force a clean if we are working on the scripts.

idelpivnitskiy: If we do not generate a new file on every run users can be in a situation when the previously generated file is not correct and contains path to the old `*-all.jar` that does not exist anymore. Examples:
1. gradle changes the hash for the cached file. Because the generated script contains the full path to the cached artifact, it won't run with the old one:
```
#!/bin/sh
exec java -jar ~/.gradle/caches/modules-2/files-2.1/io.servicetalk/servicetalk-grpc-protoc/0.26.0-SNAPSHOT/a2b7606f481385d75bc8fc419d6dfe244435a6b4/servicetalk-grpc-protoc-0.26.0-SNAPSHOT-all.jar "$@"
```
2. Users run with `--include-build`, then run the same project without `--include-build`. The script file contains value from the `serviceTalkProtocPluginPath` provided for `--include-build`. This makes script incorrect when users run it without `--include-build` later.

The overhead of writing a new file is very low. It will be cleaner and more reliable to override the old file on every run.

servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
idelpivnitskiy: This is not an issue if we generate a new file every time (see motivation described in the discussion below).
When you run project with `--include-build` the version is taken from `project.version` variable instead of `implementation-version` property. But `project.version` with `--include-build` is equal to the version of the project that runs with `--include-build`, it's not a ST version. For example, if I have a project with version `5.5` and run it with `--include-build=/path/to/servicetalk` I will see `5.5` in the script name.

servicetalk-grpc-gradle-plugin/src/main/groovy/io/servicetalk/grpc/gradle/plugin/ServiceTalkGrpcPlugin.groovy
idelpivnitskiy: Understood that intention before making this change. With the `rootProject` approach: 
1. gradle adds `build` folder at the root and does not clean it after with `./gradlew clean`;
2. it increases the exposure of the servicetalk-grpc-protoc plugin for the whole project, that is not necessary if could be isolated in the module that is using it. Because the overhead of this task is very low, it's ok to run it for each module that needs it. Optimization doesn't worth it. Debugging of the gradle's task dependency graph and sequencing of tasks is easier when everything executes as part of a single module.

Scottmitch: > gradle adds build folder at the root and does not clean it after with ./gradlew clean

This is not the behavior I observe. Can you describe your setup and Is this related to `--include-build` behavior?

Scottmitch: if we are adding to every sub-project we can get rid of this conditional check and always add the task. which avoids an "expensive" operation for gradle [1].

[1] https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#getTasksByName-java.lang.String-boolean-
> NOTE: This is an expensive operation since it requires all projects to be configured.

PR #989, opened 2020-03-30
Add support for SameSite attribute in Cookies

Scottmitch:
Motivation:
There is a revision of the HTTP cookies RFC (6265) in review which adds
a new SameSite attribute
https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05#section-5.3.7.

Modifications:
- Add the SameSite attribute support to the SetCookies related APIs

Result:
HttpSetCookies now support the SameSite attribute.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
@@ -42,7 +42,7 @@ public final class DefaultHttpSetCookie implements HttpSetCookie {
     private static final String ENCODED_LABEL_SAMESITE = "; samesite=";
 
     /**
-     * An underlying size of 8 has been shown with the current {@link AsciiBuffer} hash algorithm to have no collisions
+     * An underlying size of 16 has been shown with the current {@link AsciiBuffer} hash algorithm to have no collisions
      * with the current set of supported cookie names. If more cookie names are supported, or the hash algorithm changes
      * this initial value should be re-evaluated.
      * <p>
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
@@ -20,13 +20,13 @@ import javax.annotation.Nullable;
 
 import static io.servicetalk.http.api.CharSequences.caseInsensitiveHashCode;
 import static io.servicetalk.http.api.CharSequences.contentEqualsIgnoreCase;
+import static io.servicetalk.http.api.CharSequences.equalsIgnoreCaseLower;
 import static io.servicetalk.http.api.CharSequences.newAsciiString;
 import static io.servicetalk.http.api.HeaderUtils.validateCookieNameAndValue;
 import static io.servicetalk.http.api.HeaderUtils.validateCookieTokenAndHeaderName;
 import static io.servicetalk.http.api.HttpSetCookie.SameSite.Lax;
 import static io.servicetalk.http.api.HttpSetCookie.SameSite.None;
 import static io.servicetalk.http.api.HttpSetCookie.SameSite.Strict;
-import static java.lang.Character.toLowerCase;
 import static java.lang.Long.parseLong;
 
 /**
@@ -402,7 +402,7 @@ public final class DefaultHttpSetCookie implements HttpSetCookie {
         }
         if (sameSite != null) {
             sb.append(ENCODED_LABEL_SAMESITE);
-            sb.append(sameSite.toString());
+            sb.append(sameSite);
         }
         if (httpOnly) {
             sb.append(ENCODED_LABEL_HTTP_ONLY);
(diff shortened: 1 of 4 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpSetCookie.java
```diff
@@ -55,6 +55,15 @@
...
     CharSequence expires();
 
+    /**
+     * Get the value for the
+     * <a href="https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05#section-5.3.7">SameSite attribute</a>.
+     * @return The value for the
+     * <a href="https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05#section-5.3.7">SameSite attribute</a>.
+     */
+    @Nullable
+    SameSite sameSite();
```
Scottmitch: note this API breaks the public APIs in a few spots

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
```diff
@@ -45,13 +50,14 @@
      * if/else block we lean on {@link HttpHeaders} which provides an associative array that compares keys in a case
      * case-insensitive manner.
      */
-    private static final HttpHeaders AV_FIELD_NAMES = new DefaultHttpHeaders(8, false, false);
+    private static final HttpHeaders AV_FIELD_NAMES = new DefaultHttpHeaders(16, false, false);
```
normanmaurer: I just wondered why 16 now... Seems like 8 should be enough ?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
```diff
@@ -111,18 +119,21 @@ public DefaultHttpSetCookie(final CharSequence name, final CharSequence value,
...
      * <a href="http://www.owasp.org/index.php/HTTPOnly">HTTP-only</a>).
+     * @param sameSite the
+     * <a href="https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05#section-5.3.7">SameSite attribute</a>.
      */
     public DefaultHttpSetCookie(final CharSequence name, final CharSequence value, @Nullable final CharSequence path,
                                 @Nullable final CharSequence domain, @Nullable final CharSequence expires,
-                                @Nullable final Long maxAge, final boolean wrapped, final boolean secure,
-                                final boolean httpOnly) {
+                                @Nullable final Long maxAge, @Nullable final SameSite sameSite, final boolean wrapped,
+                                final boolean secure, final boolean httpOnly) {
```
normanmaurer: not really related to this PR but I wondered if you may want to consider adding a Builder at some point as these are quite a few params for a constructor.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
```diff
@@ -277,7 +295,9 @@ static HttpSetCookie parseSetCookie(final CharSequence setCookieString, boolean
             }
         }
 
-        return new DefaultHttpSetCookie(name, value, path, domain, expires, maxAge, isWrapped, isSecure, isHttpOnly);
+        assert name != null && value != null; // these are checked at runtime in the constructor
```
normanmaurer: why you even need this if its checked in the constructor ?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
```diff
@@ -373,6 +400,10 @@ public CharSequence encoded() {
             sb.append(ENCODED_LABEL_MAX_AGE);
             sb.append(maxAge);
         }
+        if (sameSite != null) {
+            sb.append(ENCODED_LABEL_SAMESITE);
+            sb.append(sameSite.toString());
```
normanmaurer: nit: the `toString()` is not needed.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpSetCookie.java
normanmaurer: You could also add a `default` implementation if you want to minimise breakage... not sure this is needed tho.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
```diff
@@ -447,6 +479,35 @@ public CharSequence subSequence(final int start, final int end) {
         }
     }
 
+    @Nullable
+    private static SameSite fromSequence(CharSequence cs, int begin, int end) {
+        final int length = end - begin;
+        if (length == Lax.toString().length()) {
```
normanmaurer: nit: you could optimize this with a switch statement:

```
switch (length) {
    // lax
    case 3: 
       ....
    // none
    case 4:
       ....
    // strict
    case 6:
       ...
    default:
        return null;

}

```

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
Scottmitch: 8 and 16 are both functional, but 8 lead to overlap in a bucket between 2 entries and 16 makes it so there is no overlap.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
Scottmitch: lets consider for a followup PR.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
Scottmitch: IntelliJ flags this as a warning (e.g. passing in arguments that maybe null that are no annotated with `@Nullable`).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
Scottmitch: done

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
Scottmitch: done

Scottmitch:
test this please

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
NiteshKant: javadoc mentions `8` which is stale now.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpSetCookie.java
NiteshKant: Ack; I think it is OK to break APIs now.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
Scottmitch: javadoc fixed

PR #990, opened 2020-03-31
Add `whenFinally(TerminalSignalConsumer)` for Publisher, Single, Completable

idelpivnitskiy:
Motivation:

`whenFinally(Runnable)` does not allow users to understand which final event
happened. When they add multiple operators for each individual event users
have to manually protect from possible concurrency between `whenCancel` and
`whenOnComplete/whenOnError`.

Modification:

- Add `whenFinally(TerminalSignalConsumer)` operator for Publisher, Single,
Completable;
- Add `Single.TerminalSignalConsumer` that may handle `Single`'s result;
- Update javadoc;
- Enhance tests to account for different final signals for `whenFinally`;

Result:

Users can pass `TerminalSignalConsumer` for `whenFinally` operator and make
sure only a single callback will be invoked.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -757,8 +798,45 @@ public final Completable beforeCancel(Runnable onCancel) {
      * </ul>
      * for Subscriptions/{@link Subscriber}s of the returned {@link Completable}. <strong>MUST NOT</strong> throw.
      * @return The new {@link Completable}.
+     *
+     * @see <a href="http://reactivex.io/documentation/operators/do.html">ReactiveX do operator.</a>
      */
     public final Completable beforeFinally(Runnable doFinally) {
+        return beforeFinally(TerminalSignalConsumer.from(doFinally));
```
idelpivnitskiy: Do we need to keep `beforeFinally(Runnable)`? Should we remove it and users can use `beforeFinally(TerminalSignalConsumer.from(doFinally))`  themselves.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -294,6 +294,46 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
         return afterFinally(doFinally);
     }
 
+    /**
+     * Invokes an appropriate terminal signal of the {@link TerminalSignalConsumer}, when any of the following terminal
+     * methods are called:
+     * <ul>
+     *     <li>{@link Subscriber#onSuccess(Object)} - invokes {@link TerminalSignalConsumer#onComplete()}</li>
```
idelpivnitskiy: Here we have an inconsistency between `Single.whenOnSuccess(Consumer<? super T>)` and `TerminalSignalConsumer.onComplete()` that does not provide `T`. Should we have a different `TerminalSignalConsumer` for `Single`?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -1499,6 +1540,42 @@ public final Completable flatMapCompletableDelayError(Function<? super T, ? exte
...
+        return beforeFinally(TerminalSignalConsumer.from(doFinally));
+    }
+
+    /**
+     * Invokes an appropriate terminal signal of the {@link TerminalSignalConsumer} <strong>before</strong> any of the
+     * following terminal methods are called:
+     * <ul>
+     *     <li>{@link Subscriber#onComplete()} - invokes {@link TerminalSignalConsumer#onComplete()}</li>
+     *     <li>{@link Subscriber#onError(Throwable)} - invokes {@link TerminalSignalConsumer#onError(Throwable)}</li>
+     *     <li>{@link Subscription#cancel()} - invokes {@link TerminalSignalConsumer#onCancel()}</li>
```
idelpivnitskiy: Here we have a small naming inconsistency between `TerminalSignalConsumer` API and other `when*` operators:
- `whenOnComplete(Runnable)` matches with `TerminalSignalConsumer#onComplete()`;
- `whenOnError(Consumer<Throwable>)` matches with `TerminalSignalConsumer#onError(Throwable)`;

`whenCancel(Runnable)` does not have `On` in the middle to match with `TerminalSignalConsumer#onCancel()`. 

It's ok for me, just want to highlight this naming inconsistency. WDYT?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumer.java
```diff
@@ -40,4 +42,14 @@ default void onError(Throwable throwable) {
...
     }
+
+    /**
+     * Creates a new {@link TerminalSignalConsumer} from {@link Runnable}.
+     *
+     * @param onFinally a {@link Runnable} to run on any terminal signal
+     * @return a new {@link TerminalSignalConsumer} from {@link Runnable}
+     */
+    static TerminalSignalConsumer from(Runnable onFinally) {
+        return new RunnableTerminalSignalConsumer(onFinally);
```
idelpivnitskiy: Tried to place `RunnableTerminalSignalConsumer` in some internal module to avoid adding this API, but the module structure did not allow me to reuse this class in `concurrent-api` and `http-utils`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
NiteshKant: `Runnable` flavors are useful for state cleanups where type of termination does not matter. So, I will suggest to keep these methods.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumer.java
NiteshKant: It seems Ok to me to add as a `public` method.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumer.java
```diff
@@ -40,4 +42,14 @@ default void onError(Throwable throwable) {
      */
     default void onCancel() {
     }
+
+    /**
+     * Creates a new {@link TerminalSignalConsumer} from {@link Runnable}.
```
NiteshKant, line 47:
```suggestion
     * Adapts the passed {@link Runnable} to a {@link TerminalSignalConsumer} such that {@link Runnable#run()} 
     * is invoked for each call to {@link TerminalSignalConsumer#onError()}, {@link TerminalSignalConsumer#onComplete()} and {@link TerminalSignalConsumer#onCancel()}.
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumer.java
```diff
@@ -40,4 +42,14 @@ default void onError(Throwable throwable) {
...
+     *
+     * @param onFinally a {@link Runnable} to run on any terminal signal
```
NiteshKant, line 49:
```suggestion
     * @param onFinally a {@link Runnable} to adapt to the returned {@link TerminalSignalConsumer} such that {@link Runnable#run()} is invoked for each call to {@link TerminalSignalConsumer#onError()}, {@link TerminalSignalConsumer#onComplete()} and {@link TerminalSignalConsumer#onCancel()}
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumer.java
```diff
@@ -40,4 +42,14 @@ default void onError(Throwable throwable) {
...
+     * @return a new {@link TerminalSignalConsumer} from {@link Runnable}
```
NiteshKant, line 50:
```suggestion
     * @return a {@link TerminalSignalConsumer} that adapts the passed {@link Runnable}.
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
```diff
@@ -0,0 +1,83 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.concurrent.api;
+
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+
+import static java.util.Objects.requireNonNull;
+
+final class TerminalSignalConsumers implements TerminalSignalConsumer {
```
NiteshKant, line 22: Does this have to `implement TerminalSignalConsumer`?

Also, this does not follow the typical pattern where in such classes we provide static factory methods instead of exposing the concrete implementations.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
```diff
@@ -0,0 +1,83 @@
...
+
+    static final class CompleteTerminalSignalConsumer implements TerminalSignalConsumer {
+
+        private static final AtomicIntegerFieldUpdater<TerminalSignalConsumers.CompleteTerminalSignalConsumer>
+                completeUpdater = AtomicIntegerFieldUpdater.newUpdater(
+                        TerminalSignalConsumers.CompleteTerminalSignalConsumer.class, "complete");
+        @SuppressWarnings("unused")
+        private volatile int complete;
+
+        private final TerminalSignalConsumer deletage;
```
NiteshKant, line 56: `deletage` -> `delegate`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
```diff
@@ -0,0 +1,83 @@
...
+            onFinally.run();
+        }
+
+        @Override
+        public void onCancel() {
+            onFinally.run();
+        }
+    }
+
+    static final class CompleteTerminalSignalConsumer implements TerminalSignalConsumer {
```
NiteshKant, line 48: `CompleteTerminalSignalConsumer` => `DelegatingTerminalSignalConsumer`?

I am on the fence for wrapping here to make sure that we only invoke the said methods once. Altenratively you can leave the state in respective operators as they are today and if you want to share code of invoking these methods, you can create static methods here like:

```java

static <T> void invokeOnCompleteIfNotDone(TerminalSignalConsumer target, AtomicIntegerFieldUpdater<T> doneUpdater,
                                          final T owner) {
    if (doneUpdater.compareAndSet(owner, 0, 1)) {
        target.onComplete();
    }
}
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
```diff
@@ -0,0 +1,83 @@
...
+
+        private static final AtomicIntegerFieldUpdater<TerminalSignalConsumers.CompleteTerminalSignalConsumer>
+                completeUpdater = AtomicIntegerFieldUpdater.newUpdater(
+                        TerminalSignalConsumers.CompleteTerminalSignalConsumer.class, "complete");
+        @SuppressWarnings("unused")
+        private volatile int complete;
```
NiteshKant, line 54: nit: name `complete` is misleading as there is a method here `onComplete()` which may indicate that this state resembles calling that method. You can instead use a name like `done`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -757,8 +798,45 @@ public final Completable beforeCancel(Runnable onCancel) {
...
+    }
+
+    /**
+     * Invokes the {@code whenFinally} {@link Runnable} argument <strong>before</strong> any of the following terminal
```
NiteshKant, line 809: Similar comment for other methods here and in other sources.

```suggestion
     * Invokes the corresponding method on {@code whenFinally} {@link TerminalSignalConsumer} argument <strong>before</strong> any of the following terminal
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -757,8 +798,45 @@ public final Completable beforeCancel(Runnable onCancel) {
...
+     *     <li>{@link Cancellable#cancel()} - invokes {@link TerminalSignalConsumer#onCancel()}</li>
+     * </ul>
+     * for Subscriptions/{@link Subscriber}s of the returned {@link Completable}.
+     * <p>
+     * From a sequential programming point of view this method is roughly equivalent to the following:
+     * <pre>{@code
+     *  try {
+     *      resultOfThisCompletable();
+     *  } finally {
+     *      doFinally.onTerminalSignal(); // Invokes an appropriate terminal signal
```
NiteshKant, line 823: Similar comment for other methods here and in other sources.

The semantics here do not really match `try-finally` now as the code paths are different for error vs completion:

```java
     *  try {
     *      resultOfThisCompletable();
     *  } catch(Throwable t) {
     *      doFinally.onError(t);
     *      nextOperation(); // Maybe notifying of cancellation, or termination
     *      return;
     *  }
     *  doFinally.onComplete(); 
     *  nextOperation(); // Maybe notifying of cancellation, or termination
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
NiteshKant: Ya I think so as `onComplete()` does not match the name (`onSuccess()`) too.

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TerminalSignalConsumerMock.java
```diff
@@ -0,0 +1,94 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Mocked {@link TerminalSignalConsumer} that helps to verify that only a single terminal method was invoked.
+ */
+public final class TerminalSignalConsumerMock implements TerminalSignalConsumer {
+
+    private Runnable runnable;
```
NiteshKant, line 30: Fields can be `final`

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TerminalSignalConsumerMock.java
```diff
@@ -0,0 +1,94 @@
...
+    private TerminalSignalConsumer signalConsumer;
+
+    /**
+     * Creates a new instance.
+     */
+    public TerminalSignalConsumerMock() {
+        runnable = Mockito.mock(Runnable.class);
```
NiteshKant, line 37: What is the use of the `Runnable` here?

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TerminalSignalConsumerMock.java
```diff
@@ -0,0 +1,94 @@
...
+    public void onCancel() {
+        signalConsumer.onCancel();
+    }
+
+    /**
+     * Verifies that only {@link TerminalSignalConsumer#onComplete()} was invoked and no other methods.
+     */
+    public void verifyOnComplete() {
+        verify(signalConsumer).onComplete();
+        verify(signalConsumer, never()).onError(any(Throwable.class));
```
NiteshKant, line 61: You can use `Mockito.verifyNoMoreInteractions(signalConsumer);` to verify that no other methods are invoked on the mock.

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TerminalSignalConsumerMock.java
```diff
@@ -0,0 +1,94 @@
...
+
+import static org.mockito.AdditionalAnswers.delegatesTo;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Mocked {@link TerminalSignalConsumer} that helps to verify that only a single terminal method was invoked.
+ */
+public final class TerminalSignalConsumerMock implements TerminalSignalConsumer {
```
NiteshKant, line 28: I am not really sure about this class if you remove the `Runnable` reference and use `Mockito.verifyNoMoreInteractions(signalConsumer);`. It is more intuitive if the verification is done inside the test as opposed to inside a common method like this.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
NiteshKant: The operator names are following the pattern `before/when/after[actual method name]` so I think `whenCancel()` is correct. We can rename `TerminalSignalConsumer#onCancel()` to `cancel()` to match the pattern

normanmaurer: `cancel()` sounds like it "cancels" something which seems to be no the case or I am missing something ?

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TerminalSignalConsumerMock.java
normanmaurer: @NiteshKant I think it is just used in the line below. @idelpivnitskiy consider to adjust scope to constructor.

normanmaurer, line 37: nit: static import ?

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/AbstractWhenFinallyTest.java
```diff
@@ -39,46 +38,41 @@
     @Rule
     public final ExpectedException thrown = ExpectedException.none();
 
-    private Runnable doFinally;
-
-    @Before
-    public void setUp() throws Exception {
-        doFinally = mock(Runnable.class);
-    }
+    private TerminalSignalConsumerMock doFinally = new TerminalSignalConsumerMock();
```
normanmaurer, line 41: nit: final

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/AbstractWhenFinallyTest.java
```diff
@@ -46,13 +45,8 @@
...
     final TestPublisher<String> publisher = new TestPublisher<>();
     final TestPublisherSubscriber<String> subscriber = new TestPublisherSubscriber<>();
-    private Runnable doFinally;
     final TestSubscription subscription = new TestSubscription();
-
-    @Before
-    public void setUp() throws Exception {
-        doFinally = mock(Runnable.class);
-    }
+    private TerminalSignalConsumerMock doFinally = new TerminalSignalConsumerMock();
```
normanmaurer, line 49: nit: final

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/completable/AbstractWhenFinallyTest.java
```diff
@@ -38,46 +37,41 @@
     @Rule
     public final ExpectedException thrown = ExpectedException.none();
 
-    private Runnable doFinally;
-
-    @Before
-    public void setUp() throws Exception {
-        doFinally = mock(Runnable.class);
-    }
+    private TerminalSignalConsumerMock doFinally = new TerminalSignalConsumerMock();
```
normanmaurer, line 40: nit: final

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
```diff
@@ -0,0 +1,83 @@
...
+
+        private final TerminalSignalConsumer deletage;
+
+        CompleteTerminalSignalConsumer(final TerminalSignalConsumer deletage) {
```
normanmaurer, line 58: `deletage` -> `delegate` (just as @NiteshKant also mentioned above)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
```diff
@@ -0,0 +1,83 @@
...
+        public void onCancel() {
+            onFinally.run();
+        }
+    }
+
+    static final class CompleteTerminalSignalConsumer implements TerminalSignalConsumer {
+
+        private static final AtomicIntegerFieldUpdater<TerminalSignalConsumers.CompleteTerminalSignalConsumer>
+                completeUpdater = AtomicIntegerFieldUpdater.newUpdater(
+                        TerminalSignalConsumers.CompleteTerminalSignalConsumer.class, "complete");
```
normanmaurer, line 52: nit: as this one is internal anyway you could also extend `AtomicInteger` directly.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
```diff
@@ -0,0 +1,83 @@
...
+            this.onFinally = requireNonNull(onFinally);
+        }
+
+        @Override
+        public void onComplete() {
+            onFinally.run();
+        }
+
+        @Override
+        public void onError(final Throwable throwable) {
```
normanmaurer, line 38: nit: remove final to be consistent ?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumer.java
normanmaurer: This should be `{@link {@link TerminalSignalConsumer#onError(Throwable)}}`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumer.java
normanmaurer: This should be `{@link TerminalSignalConsumer#onError(Throwable)}`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -215,6 +215,47 @@ public final Completable whenFinally(Runnable doFinally) {
...
+     *     <li>{@link Cancellable#cancel()} - invokes {@link TerminalSignalConsumer#onCancel()}</li>
+     * </ul>
+     * for Subscriptions/{@link Subscriber}s of the returned {@link Completable}.
+     * <p>
+     * The order in which {@code whenFinally} will be invoked relative to the above methods is undefined. If you need
+     * strict ordering see {@link #beforeFinally(TerminalSignalConsumer)} and
+     * {@link #afterFinally(TerminalSignalConsumer)}.
+     * <p>
+     * From a sequential programming point of view this method is roughly equivalent to the following:
+     * <pre>{@code
```
normanmaurer, line 233: @idelpivnitskiy cool... I didn't know you can use `{@code...} with multiple lines..

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
NiteshKant: Ya its just a conflict between following names of the callbacks these methods represent (`TerminalSignalConsumer#cancel()` here is invoked when `Subscription#cancel()` is called) or changing names to make it more meaningful in this context. We went with following callback names which seems easier to correlate.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
idelpivnitskiy: I tried this approach first and it works well for `(Before|After)FinallyPublisher`, but similar operators for `Completable` and `Single` use `BeforeCancellable`. It required me to create a new `Cancellable` implementation or pass a lambda to `BeforeCancellable` that will keep references to the state outside of that lambda. Not a big deal but introduces a similar need for wrapping. 
This approach also wraps, but at the same time helps to reduce duplication between operators of maintaining atomic state and allows me to pass only references to the `doFinally::onCancel` for `BeforeCancellable`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
idelpivnitskiy: If we will have something like `SingleTerminalSignalConsumer`, WDYT about having a `TerminalSignalConsumer` for each async source independently?
- `CompletableTerminalSignalConsumer`
- `SingleTerminalSignalConsumer`
- `PublisherTerminalSignalConsumer`

Also, WDYT about promoting all of them to `servicetalk-concurrent` module where we keep all interfaces for async sources?
- `CompletableSource.TerminalSignalConsumer`
- `SingleSource.TerminalSignalConsumer`
- `PublisherSource.TerminalSignalConsumer`

idelpivnitskiy: Moving them to `servicetalk-concurrent` module will also help to hide `from(Runnable)` from public API.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
idelpivnitskiy: > Does this have to implement `TerminalSignalConsumer`?

No, this is a leftover from intermediate iteration. This upper-level class used to be `RunnableTerminalSignalConsumer`. Thanks for the catch.

idelpivnitskiy: Btw, because `TerminalSignalConsumer` has all its methods as `default` noop impl, it's very easy to forget about them. Was bitten by this twice. Should we remove default implementations? If users don't need all 3 signals they can use alternative operators. If they need only 2 signals, it's not a big deal to provide a noop impl for the signal you are not interested in,

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
idelpivnitskiy: Thanks for the tip. Didn't know `AtomicInteger` is not final.
Let me use the pattern with `AtomicIntegerFieldUpdater` we have everywhere else in the codebase.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
idelpivnitskiy: I will add `final` in `RunnableTerminalSignalConsumer` ctor to be consistent.

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TerminalSignalConsumerMock.java
idelpivnitskiy: Trying to test that `RunnableTerminalSignalConsumer` delegates to this `Runnable` correctly and invokes it only once.

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TerminalSignalConsumerMock.java
idelpivnitskiy: Agreed, if I will have a different `TerminalSignalConsumer` for `Single` it will have even less sense. Let me remove it.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
NiteshKant: I am hesitant to push this to `concurrent` module just for reuse of an implementation. This interface is specific to operators and not to the sources.

As we do not expect new callbacks on `Completable` and the terminal callbacks are the same b/w `Publisher` and `Completable`, I think it is less justified to add different interfaces for `Publisher` and `Completable`. I would suggest keeping the current `TerminalSignalConsumer ` as-is, use it for `Publisher` and `Completable` and create a `Single` specific interface which can be an inner class in `concurrent.api.Single`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
idelpivnitskiy: Renamed to `TerminalSignalConsumer#cancel()`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumers.java
idelpivnitskiy: Because we have `Single.TerminalSignalConsumer` now this wrapped doesn't make sense anymore. Reverted the previous approach.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AfterFinallyCompletable.java
```diff
@@ -23,43 +23,49 @@
...
+        AfterFinallyCompletableSubscriber(Subscriber original, TerminalSignalConsumer doFinally) {
             this.original = original;
-            this.runnable = runnable;
+            this.doFinally = doFinally;
         }
 
         @Override
         public void onSubscribe(Cancellable originalCancellable) {
-            original.onSubscribe(new BeforeCancellable(originalCancellable, this::afterFinally));
+            original.onSubscribe(new BeforeCancellable(originalCancellable, () -> {
```
Scottmitch, line 54: (pre-existing): `BeforeCancellable` -> consider renaming to `ComposedCancellable` or as the "before" naming convention clashes with the different operators that use is (e.g. this is an `After` operator that uses a `Before` class).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -64,6 +64,76 @@
...
  */
 public abstract class Single<T> {
+
+    /**
+     * Callback interface on which only a single method is ever called matching the terminal outcome of the associated
+     * {@link Single} and {@link Cancellable}.
+     *
+     * @param <T> Type of the result of the {@link Single}.
+     */
+    public interface TerminalSignalConsumer<T> {
```
Scottmitch, line 74: consider moving this new code down in the class as this interface isn't a central feature of the outer class.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -670,6 +780,45 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
      * @return The new {@link Single}.
      */
     public final Single<T> beforeFinally(Runnable doFinally) {
+        return beforeFinally(TerminalSignalConsumer.from(doFinally));
```
Scottmitch, line 783: if we still expose the `Runnable` variants can we hide `TerminalSignalConsumer.from` for now?

idelpivnitskiy:
jdk11 build failed because of #999

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumer.java
```diff
@@ -16,28 +16,25 @@
...
     /**
      * Callback to signal completion of the {@code Subscription} for this {@code Subscriber}.
      */
-    default void onComplete() {
-    }
+    void onComplete();
 
     /**
-     * Callback to receive an {@link Throwable error} for this {@code Subscriber}.
+     * Callback to signal an {@link Throwable error} of the {@code Subscription} for this {@code Subscriber}.
```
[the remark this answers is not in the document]
idelpivnitskiy: I tried to follow the pattern we have above for `onComplete()` to make them consistent:
> Callback to signal completion of the `{@code Subscription}` for this `{@code Subscriber}`.

Only `Subscriber` has `onComplete/onError` but it owns the `Subscription` for the source. Should we change `Subscription` to `Source` in both places?

> Callback to signal completion of the `{@code Source}` for this `{@code Subscriber}`.
> Callback to signal `{@link Throwable error}` of the `{@code Source}` for this `{@code Subscriber}`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -757,8 +798,48 @@ public final Completable beforeCancel(Runnable onCancel) {
...
+     *     <li>{@link Cancellable#cancel()} - invokes {@link TerminalSignalConsumer#cancel()}</li>
+     * </ul>
+     * for Subscriptions/{@link Subscriber}s of the returned {@link Completable}.
+     * <p>
+     * From a sequential programming point of view this method is roughly equivalent to the following:
+     * <pre>{@code
+     *  try {
+     *      resultOfThisCompletable();
+     *  } catch(Throwable t) {
+     *      doFinally.onError(t);
```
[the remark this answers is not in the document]
idelpivnitskiy: I used this pattern for `beforeFinally(TerminalSignalConsumer)` for Completable, Single, and Publisher as it represents logic of `BeforeFinallyCompletableSubscriber`. 
`whenFinally(TerminalSignalConsumer)` and `afterFinally(TerminalSignalConsumer)` use another control flow, like in `AfterFinallyCompletableSubscriber`:
```java
@Override
public void onComplete() {
    try {
        original.onComplete();
    } finally {
        doFinally.onComplete();
    }
}
```
Javadoc for `Completable.afterFinally(TerminalSignalConsumer)` written in the same way:
```
     * <pre>{@code
     *  try {
     *      resultOfThisCompletable();
     *  } finally {
     *      nextOperation(); // Maybe notifying of cancellation, or termination
     *      doFinally.onTerminalSignal(); // Invokes the corresponding terminal signal
     *  }
     * }</pre>
```
We do not try-catch for `afterFinally`/`whenFinally` and always touch `TerminalSignalConsumer` in `finally` block.
I use `doFinally.onTerminalSignal()` instead of `doFinally.onCompete()` because the same patters works for `onError` and `cancel`.

Can you please clarify how the javadoc for `afterFinally`/`whenFinally` should change?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TerminalSignalConsumer.java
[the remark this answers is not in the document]
NiteshKant: Oh didn't notice the above usage, its a :dumpster fire: 😄 

```java
/**
 * A contract that provides discrete callbacks for various ways in which a
 * {@link PublisherSource.Subscriber} or a {@link CompletableSource.Subscriber} can terminate.
 */
public interface TerminalSignalConsumer {

    /**
     * Callback to indicate termination via {@link PublisherSource.Subscriber#onComplete()} or
     * {@link CompletableSource.Subscriber#onComplete()}.
     */
    default void onComplete() {
    }

    /**
     * Callback to indicate termination via {@link PublisherSource.Subscriber#onError(Throwable)} or
     * {@link CompletableSource.Subscriber#onError(Throwable)}.
     *
     * @param throwable the {@link Exception} observed.
     */
    default void onError(Throwable throwable) {
    }

    /**
     * Callback to indicate termination via {@link Cancellable#cancel()}.
     */
    default void onCancel() {
    }
}
```

Note that I have removed the "exactly once semantics" from the doc here as it is more appropriate in the operators that use this (and it is dependent on a subscribe call).

Also, use similar language on the `Single.TerminalSignalConsumer`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -294,6 +295,46 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
...
+     * <pre>{@code
+     *  try {
+     *      T result = resultOfThisSingle();
+     *  } finally {
+     *      // NOTE: The order of operations here is not guaranteed by this method!
+     *      nextOperation(); // Maybe notifying of cancellation, or termination
+     *      doFinally.onTerminalSignal(); // Invokes the corresponding terminal signal
+     *  }
+     * }</pre>
+     * @param doFinally an appropriate method of {@link TerminalSignalConsumer} is invoked exactly once, when any of the
```
NiteshKant, line 322: (Same suggestion for `Publisher` and `Completable` methods)

```suggestion
     * @param doFinally For each subscribe of the returned {@link Single}, at most one method of this {@link TerminalSignalConsumer} will be invoked.
```

I think we can remove the below method mappings now as I have suggested to add them to `TerminalSignalConsumer`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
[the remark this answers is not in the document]
idelpivnitskiy: Discussed offline, the control flow in javadoc is not related to the actual implementation of `Subscriber` and should represent a similar behavior for sequential programming. Updated.

PR #991, opened 2020-03-31
Do not wrap unreleasable ByteBuf(s) with `Unpooled.unreleasableBuffer`

idelpivnitskiy:
Motivation:

For cases when `AbstractByteBuf` wraps `ByteBuf`(s) allocated by ST
with another interface (slice/duplicate/readOnly) we wrap them one
more time with `UnreleasableByteBuf`. However, this is not necessary
as our `ByteBuf`(s) are always unreleasable and final. There is no
way it can be unwrapped and released. Therefore, `UnreleasableByteBuf`
is useless and just creates more garbage for GC.

Modifications:

- Remove all methods that apply `Unpooled.unreleasableBuffer` for
returned results;
- Remove unnecessary overrides for methods that return retained
slice or duplicate;
- Add missed `final` modifiers for methods of
`UnreleasableHeapByteBuf`;
- Enhance tests to cover all API that may return a new `ByteBuf`
or a new wrapped `ByteBuf`;
- Assert that all `Buffer`s allocated by ST have `refCnt() > 1`;

Result:

Less memory allocation for slice/duplicate/readOnly methods of
already unreleasable `ByteBuf`(s).

### DIFF
--- a/servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/UnreleasableDirectByteBuf.java
+++ b/servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/UnreleasableDirectByteBuf.java
@@ -44,26 +44,6 @@ final class UnreleasableDirectByteBuf extends UnpooledDirectByteBuf {
     }
 
     @Override
-    public ByteBuf readRetainedSlice(int length) {
-        return readSlice(length);
-    }
-
-    @Override
-    public ByteBuf retainedSlice() {
-        return slice();
-    }
-
-    @Override
-    public ByteBuf retainedSlice(int index, int length) {
-        return slice(index, length);
-    }
-
-    @Override
-    public ByteBuf retainedDuplicate() {
-        return duplicate();
-    }
-
-    @Override
     public ByteBuf retain(int increment) {
         return this;
     }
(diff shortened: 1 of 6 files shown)


servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/UnreleasableDirectByteBuf.java
```diff
@@ -44,31 +43,11 @@
...
-    }
-
-    @Override
-    public ByteBuf readSlice(int length) {
-        return Unpooled.unreleasableBuffer(super.readSlice(length));
-    }
-
     @Override
     public ByteBuf readRetainedSlice(int length) {
         return readSlice(length);
```
idelpivnitskiy: Actually, we can also remove these methods that are here to avoid an additional call to `retain()`. As `retain()` it just noon for our implementations, it's ok to use `readRetainedSlice` from `AbstractByteBuf`. Tests are all green if I remove these methods. WDYT?

NiteshKant: (I am not sure if you are saying the same thing)

I guess it was here before since this was calling `readSlice()` which was wrapping into the `unreleasableBuffer`. Since `readSlice()` override is removed, this becomes redundant.

PR #993, opened 2020-04-02
Replace ConcurrentSubscription with DelayedCancellable when synchronous request(Long.MAX_VALUE)

Scottmitch:
Motivation:
There are some operators which convert from Publisher to Completable or
Single. Since the demand for Completable and Single is implicit on
subscribe(..) these operators need to simulate the demand in
onSubscribe. In order to prevent concurrent access on the Subscription
with the downstream Subscriber and the synchornous requestN call it is
wrapped in a ConcurrentSubscription. However ConcurrentSubscription has
to coordinate between two methods and is inheritly more complicated than
DelayedCancellable. We can use DelayedCancellable which is a simpler
operation at the cost of requesting demand before propagating
synchronous cancel operation. This isn't perceived as a deal breaker
because demand is implicit anyways, and cancel is best effort.

Modifications:
- When converting from Publisher to Completable or Single use
DelayedCancellable instead of ConcurrentSubscription

Result:
Less complicated and less expensive conversion from Subscription to
Cancellable when converting from Publisher to Single or Completable.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ForEachSubscriber.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ForEachSubscriber.java
@@ -18,7 +18,6 @@ package io.servicetalk.concurrent.api;
 import io.servicetalk.concurrent.PublisherSource.Subscriber;
 import io.servicetalk.concurrent.PublisherSource.Subscription;
 import io.servicetalk.concurrent.internal.DelayedCancellable;
-import io.servicetalk.concurrent.internal.SequentialCancellable;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -27,7 +26,7 @@ import java.util.function.Consumer;
 
 import static java.util.Objects.requireNonNull;
 
-final class ForEachSubscriber<T> extends SequentialCancellable implements Subscriber<T> {
+final class ForEachSubscriber<T> extends DelayedCancellable implements Subscriber<T> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ForEachSubscriber.class);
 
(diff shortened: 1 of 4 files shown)


Scottmitch:
Uncovered this while benchmarking https://github.com/apple/servicetalk/pull/980

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ForEachSubscriber.java
```diff
@@ -40,11 +39,12 @@
 
     @Override
     public void onSubscribe(Subscription s) {
-        // We wrap the subscription into ConcurrentSubscription as cancel on SequentialCancellable (this) can be
+        // We wrap the subscription into DelayedCancellable as cancel on SequentialCancellable (this) can be
         // concurrent with the request-n below.
-        ConcurrentSubscription cs = wrap(s);
-        nextCancellable(cs::cancel);
-        cs.request(Long.MAX_VALUE);
+        DelayedCancellable delayedCancellable = new DelayedCancellable();
```
NiteshKant, line 44: Looks like we can just extend `DelayedCancellable` now.

```java
final class ForEachSubscriber<T> extends DelayedCancellable implements Subscriber<T> {

    @Override
    public void onSubscribe(Subscription s) {
        s.request(Long.MAX_VALUE);
        delayedCancellable(s);
    }
}
```

Also for all other classes in this PR

Scottmitch: hum good point ... I will change this class to extend `DelayedCancellable`. however in the other cases since the `Cancellable` isn't exposed externally I think it makes more sense to keep to the local allocation in `onSubscribe` as the memory can be reclaimed asap, and may even be stack allocated because it is local scope and doesn't escape.

Scottmitch: actually forget what I said ... the `DelayedCancellable` does escape to the `Subscriber` ... change made!

PR #994, opened 2020-04-02
Introduce `HttpLoadBalancerFactory`

NiteshKant:
__Motivation__

Today `HttpClientBuilder` takes two separate implementations of `LoadBalancerFactory` as well as `Function` as a protocol binder. As `LoadBalancerFactory` is protocol neutral, it does make sense to have the protocol binding separate but having two separate entities makes the generic ugly and it is hard to understand what the parameters stand for. Moreover the protocol binder is closely related to the load balancer anyways.

__Modification__

Introduce `HttpLoadBalancerFactory` that combines these two aspects and makes the usage easier.
Add a `DefaultHttpLoadBalancerFactory` to provide the default implementation using Round robin load balancer and in the future more load balancer implementations geared towards HTTP.

__Result__

Easier to understand interface for HTTP layer load balancing.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractFilterableStreamingHttpLoadBalancedConnection.java
```diff
@@ -0,0 +1,88 @@
...
+import io.servicetalk.http.api.HttpExecutionContext;
+import io.servicetalk.http.api.HttpExecutionStrategy;
+import io.servicetalk.http.api.HttpRequestMethod;
+import io.servicetalk.http.api.StreamingHttpRequest;
+import io.servicetalk.http.api.StreamingHttpResponse;
+import io.servicetalk.http.api.StreamingHttpResponseFactory;
+
+import static java.util.Objects.requireNonNull;
+
+abstract class AbstractFilterableStreamingHttpLoadBalancedConnection
```
idelpivnitskiy, line 34: This is a pkg-private abstract class that is used only in one place - `DefaultHttpLoadBalancerFactory`. Consider implementing it as a private static final class of `DefaultHttpLoadBalancerFactory`. Extend `StreamingHttpConnectionFilter` to avoid duplicating delegate logic.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -0,0 +1,141 @@
...
+import io.servicetalk.client.api.LoadBalancer;
+import io.servicetalk.client.api.LoadBalancerFactory;
+import io.servicetalk.client.api.ServiceDiscovererEvent;
+import io.servicetalk.concurrent.api.Publisher;
+import io.servicetalk.http.api.FilterableStreamingHttpConnection;
+import io.servicetalk.http.api.FilterableStreamingHttpLoadBalancedConnection;
+import io.servicetalk.http.api.HttpExecutionStrategy;
+import io.servicetalk.http.api.HttpExecutionStrategyInfluencer;
+import io.servicetalk.http.api.HttpLoadBalancerFactory;
+import io.servicetalk.loadbalancer.RoundRobinLoadBalancer;
```
idelpivnitskiy, line 28: Unused import statement

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -0,0 +1,141 @@
...
+        return config.rawFactory.newLoadBalancer(eventPublisher, cf);
+    }
+
+    @Override
+    public FilterableStreamingHttpLoadBalancedConnection toLoadBalancedConnection(
+            final FilterableStreamingHttpConnection connection) {
+        return new AbstractFilterableStreamingHttpLoadBalancedConnection(connection) {
+            @Override
+            public float score() {
+                return 1;
```
idelpivnitskiy, line 60: Looking at #996, should it be `Integer.MAX_VALUE`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -0,0 +1,141 @@
...
+        public static <ResolvedAddress> Builder<ResolvedAddress> from(
+                final LoadBalancerFactory<ResolvedAddress, FilterableStreamingHttpLoadBalancedConnection> rawFactory) {
+            final HttpExecutionStrategyInfluencer strategyInfluencer;
+            if (rawFactory instanceof HttpExecutionStrategyInfluencer) {
+                strategyInfluencer = (HttpExecutionStrategyInfluencer) rawFactory;
+            } else if(rawFactory instanceof RoundRobinLoadBalancerFactory) {
+                strategyInfluencer = strategy -> strategy; // RoundRobinLoadBalancer is non-blocking.
+            } else {
+                /* user provided load balancer assumed to be blocking unless told otherwise */
+                strategyInfluencer = defaultStreamingInfluencer();
```
idelpivnitskiy, line 125: The LB provided by users may be implemented in the same way as `RoundRobinLoadBalancerFactory`: it may not have a dependency on `http-api`. Therefore, it won't be able to implement `HttpExecutionStrategyInfluencer`. Consider providing a builder method that can take `HttpExecutionStrategyInfluencer`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -0,0 +1,141 @@
...
+import static io.servicetalk.loadbalancer.RoundRobinLoadBalancer.newRoundRobinFactory;
+
+/**
+ * Default implementation of {@link HttpLoadBalancerFactory}.
+ *
+ * @param <ResolvedAddress> The type of address after resolution.
+ */
+public final class DefaultHttpLoadBalancerFactory<ResolvedAddress>
+        implements HttpLoadBalancerFactory<ResolvedAddress>, HttpExecutionStrategyInfluencer {
+    private final Config<ResolvedAddress> config;
```
Scottmitch, line 41: consider killing the `Config` object and unrolling to just pass the fields directly from the builder:

```java
private final LoadBalancerFactory<ResolvedAddress, FilterableStreamingHttpLoadBalancedConnection> rawFactory;
private final HttpExecutionStrategyInfluencer strategyInfluencer;
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -0,0 +1,141 @@
...
+         */
+        public static <ResolvedAddress> Builder<ResolvedAddress> from(
+                final LoadBalancerFactory<ResolvedAddress, FilterableStreamingHttpLoadBalancedConnection> rawFactory) {
+            final HttpExecutionStrategyInfluencer strategyInfluencer;
+            if (rawFactory instanceof HttpExecutionStrategyInfluencer) {
+                strategyInfluencer = (HttpExecutionStrategyInfluencer) rawFactory;
+            } else if(rawFactory instanceof RoundRobinLoadBalancerFactory) {
+                strategyInfluencer = strategy -> strategy; // RoundRobinLoadBalancer is non-blocking.
+            } else {
+                /* user provided load balancer assumed to be blocking unless told otherwise */
```
Scottmitch, line 124: nit: consider using `//` comments

Scottmitch: nit: consider clarifying, the only supported way to override the `rawFactory` to implement `HttpExecutionStrategyInfluencer`.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -331,6 +329,25 @@ public Completable closeAsyncGracefully() {
         return asyncCloseable.closeAsyncGracefully();
     }
 
+    /**
+     * {@link LoadBalancerFactory} for {@link RoundRobinLoadBalancer}.
+     *
+     * @param <ResolvedAddress> The resolved address type.
+     * @param <C> The type of connection.
+     */
+    public static final class RoundRobinLoadBalancerFactory<ResolvedAddress, C extends LoadBalancedConnection>
```
Scottmitch, line 338: can you clarify why this needs to be exposed as public type and have a static factory method to create a new one?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
idelpivnitskiy: Should we provide any way for users to configure this value? For example, consider adding a method for the builder that will take `Function<FilterableStreamingHttpConnection, FilterableStreamingHttpLoadBalancedConnection>`.

NiteshKant: >> Looking at #996, should it be Integer.MAX_VALUE?

Yes! I will change it here or in the other PR depending on which gets merged first.

How scores are used/specified depends on the load balancer requirements. eg: the only load balancer we have ATM does not use the scores. I will avoid adding new APIs speculating usage. Once we have something that uses the scores, we can decide the APIs. Until then, one can create their own `HttpLoadBalancerFactory` implementation.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
NiteshKant: >>  it may not have a dependency on http-api. 

If someone wants to influence the strategy then the only way to do it is to have dependency on the `http-api`. If there is a generic load balancer implementation, it can be wrapped and the strategy influencer can be added.

```java
public class NoInfluenceLBFactory<R, C extends LoadBalancedConnection> 
        implements LoadBalancerFactory<R, C>, HttpExecutionStrategyInfluencer {
    
    private final LoadBalancerFactory<R, C> delegate;

    public NoInfluenceLBFactory(final LoadBalancerFactory<R, C> delegate) {
        this.delegate = delegate;
    }

    @Override
    public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strategy) {
        return strategy;
    }

    @Override
    public LoadBalancer<? extends C> newLoadBalancer(
            final Publisher<? extends ServiceDiscovererEvent<R>> eventPublisher, 
            final ConnectionFactory<R, ? extends C> connectionFactory) {
        return delegate.newLoadBalancer(eventPublisher, connectionFactory);
    }
}
```

Decoupling strategy influencing from actual code creates scopes for bugs as the implementations change.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
NiteshKant: The complexity here is that we want to make sure a `LoadBalancerFactory` can specify that it does not block (implement `HttpExecutionStrategyInfluencer`) but `LoadBalancerFactory` is protocol agnostic hence it can not implement that. 

Now, users can specify `RoundRobinLoadBalancer` factory to the client without realizing that they need to wrap with `HttpExecutionStrategyInfluencer`. This is an attempt to identify that the passed `LoadBalancerFactory` is provided by ST and hence we can make some assumptions.

Alternatively, we can say that if a user provides any LB factory (even if it is RRLB) we will offload and they should make sure to wrap, it seemed we can do better than that.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -0,0 +1,200 @@
...
+                // user provided load balancer assumed to be blocking unless it implements
+                // HttpExecutionStrategyInfluencer
+                strategyInfluencer = defaultStreamingInfluencer();
+            }
+            return new Builder<>(rawFactory, strategyInfluencer);
+        }
+    }
+
+    private static final class DefaultFilterableStreamingHttpLoadBalancedConnection
+            implements FilterableStreamingHttpLoadBalancedConnection {
```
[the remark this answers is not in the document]
NiteshKant: But it is not a connection filter! 

I expect this to be made public later as we do not have a delegating `FilterableStreamingHttpLoadBalancedConnection` implementation, which was my original intent but did not do it due to lack of usage.

PR #996, opened 2020-04-02
Use `int`s as scores for `ScoreSupplier`

NiteshKant:
__Motivation__

`ScoreSupplier#score()` returns a `float` today with an intent to have precisions for slightly varying scores. However floating point arithmentic adds more complexity in score computations. Instead we can use `int`s and use bigger scores for precision (eg: multiply by `1000`)

__Modification__

Make `ScoreSupplier#score()` return an `int`.

__Result__

Simpler score calculations.


NiteshKant:
CI setup issues for JDK 8 PRB

normanmaurer:
LGTM... just wondering about trouble caused by API breakage

NiteshKant:
@normanmaurer ya we are taking liberties to change APIs ATM

PR #998, opened 2020-04-03
Work around SpotBugs JDK11 bug, make code more robust

Scottmitch:
Motivation:
SpotBugs has a bug impacting builds on JDK11 https://github.com/spotbugs/spotbugs/issues/756. This also hinted at an area that could be made more robust.

Modifications:
- Ignore SpotBugs warning due to bug
- Improve robustness of handling null element from iterator

Result:
SpotBugs no longer fails on JDK11 builds.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
@@ -29,6 +29,7 @@ import io.servicetalk.http.api.HttpResponse;
 import io.servicetalk.http.api.StreamingHttpClient;
 import io.servicetalk.http.api.StreamingHttpRequest;
 
+import java.util.NoSuchElementException;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.internal.BlockingIterables.singletonBlockingIterable;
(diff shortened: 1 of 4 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
```diff
@@ -155,7 +156,10 @@
                 newBlockingStreamingCall(serializationProvider, requestClass, responseClass);
         return (metadata, request) -> {
             try (BlockingIterator<Resp> iterator = streamingClientCall.request(metadata, request).iterator()) {
-                final Resp firstItem = iterator.next();
+                final Resp firstItem;
+                if (!iterator.hasNext() || (firstItem = iterator.next()) == null) {
```
NiteshKant, line 160: `null` value here is essentially OK from the context of this call and we generally pass through `null`.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -593,11 +593,10 @@ public void handle(final GrpcServiceContext ctx, final BlockingIterable<Req> req
                                            final GrpcPayloadWriter<Resp> responseWriter) throws Exception {
                             final Req firstItem;
                             try (BlockingIterator<Req> requestIterator = request.iterator()) {
-                                if (!requestIterator.hasNext()) {
+                                if (!requestIterator.hasNext() || (firstItem = requestIterator.next()) == null) {
```
NiteshKant, line 596: Same here, we should pass through `null` values.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
Scottmitch: Can you clarify in which situations we expect to return a `null` value for a scalar blocking RPC call?

NiteshKant: The scalar/streaming blocking/async variant is immaterial here and no we do not expect a `null` value but neither do we use the `Resp` here in a way that it can not be `null`. 

Hypothesizing situations, `streamingClientCall` that is giving the `BlockingIterator` here is created using a `StreamingHttpClient` that is provided by the user of `GrpcClientCallFactory` API. It is nowhere we check that the payload of the response returned by the client can not be `null` unless we actually use it, eg: while deserializing. `Deserializer` here also is provided by the user. So really a user can pass `null` from everywhere and get back a `null` response and we really do not care.

So the point here isn't whether it is fair to make that assumption, it is really about whether there is any positive of making that assumption. Same argument as passing `null` through async sources, we do not care about it being `null` unless we use it.

PR #1002, opened 2020-04-04
`ProxyConnectConnectionFactoryFilter` leaks connection in case of errors

idelpivnitskiy:
Motivation:

In case of any error or cancellation, `ProxyConnectConnectionFactoryFilter`
does not close the connection to the proxy.

Modifications:

- Add tests to verify different failure scenarios for
`ProxyConnectConnectionFactoryFilter`;
- Close connection when `CONNECT` request or SSL handshake failures;
- Do not drain the response payload body in case of error, because the
connection will be closed anyway;

Result:

`ProxyConnectConnectionFactoryFilter` does not leak connections.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
@@ -20,7 +20,6 @@ import io.servicetalk.client.api.ConnectionFactoryFilter;
 import io.servicetalk.client.api.DelegatingConnectionFactory;
 import io.servicetalk.concurrent.SingleSource;
 import io.servicetalk.concurrent.api.Single;
-import io.servicetalk.concurrent.api.SingleTerminalSignalConsumer;
 import io.servicetalk.http.api.FilterableStreamingHttpConnection;
 import io.servicetalk.http.api.HttpExecutionStrategy;
 import io.servicetalk.http.api.HttpExecutionStrategyInfluencer;
@@ -32,8 +31,6 @@ import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.handler.ssl.SslHandshakeCompletionEvent;
-
-import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.api.Processors.newSingleProcessor;
 import static io.servicetalk.concurrent.api.Single.failed;
@@ -78,26 +75,8 @@ final class ProxyConnectConnectionFactoryFilter<ResolvedAddress, C extends Filte
                     // connection filters we use the default strategy which should offload everything to be safe.
                     return c.request(defaultStrategy(), c.connect(connectAddress).addHeader(CONTENT_LENGTH, ZERO))
                             .flatMap(response -> handleConnectResponse(c, response))
-                            // Close recently created connection in case of any error while it connects to the proxy
-                            // or cancellation:
-                            .recoverWith(t -> c.closeAsync().concat(failed(t)))
-                            // Use afterFinally to prevent handling cancel event after termination:
-                            .afterFinally(new SingleTerminalSignalConsumer<C>() {
-                                @Override
-                                public void cancel() {
-                                    c.closeAsync().subscribe();
-                                }
-
-                                @Override
-                                public void onSuccess(@Nullable final C result) {
-                                    // noop
-                                }
-
-                                @Override
-                                public void onError(final Throwable throwable) {
-                                    // noop
-                                }
-                            });
+                            // Close recently created connection in case of any error while it connects to the proxy:
+                            .recoverWith(t -> c.closeAsync().concat(failed(t)));
                 } catch (Throwable t) {
                     return c.closeAsync().concat(failed(t));
                 }
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
@@ -82,41 +82,41 @@ final class ProxyConnectConnectionFactoryFilter<ResolvedAddress, C extends Filte
                 }
             });
         }
-    }
 
-    private Single<C> handleConnectResponse(final C connection, final StreamingHttpResponse response) {
-        if (response.status().statusClass() != SUCCESSFUL_2XX) {
-            return failed(new ProxyResponseException("Non-successful response from proxy CONNECT " +
-                    connectAddress, response.status()));
-        }
-
-        final Channel channel = ((NettyConnectionContext) connection.connectionContext()).nettyChannel();
-        final SingleSource.Processor<C, C> processor = newSingleProcessor();
-        channel.pipeline().addLast(new ChannelInboundHandlerAdapter() {
-            @Override
-            public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) {
-                if (evt instanceof SslHandshakeCompletionEvent) {
-                    SslHandshakeCompletionEvent event = (SslHandshakeCompletionEvent) evt;
-                    if (event.isSuccess()) {
-                        processor.onSuccess(connection);
-                    } else {
-                        processor.onError(event.cause());
-                    }
-                }
-                ctx.fireUserEventTriggered(evt);
+        private Single<C> handleConnectResponse(final C connection, final StreamingHttpResponse response) {
+            if (response.status().statusClass() != SUCCESSFUL_2XX) {
+                return failed(new ProxyResponseException("Non-successful response from proxy CONNECT " +
+                        connectAddress, response.status()));
             }
-        });
 
-        final DeferSslHandler deferSslHandler = channel.pipeline().get(DeferSslHandler.class);
-        if (deferSslHandler == null) {
-            return failed(new IllegalStateException("Failed to find a handler of type " +
-                    DeferSslHandler.class + " in channel pipeline."));
+            final Channel channel = ((NettyConnectionContext) connection.connectionContext()).nettyChannel();
+            final SingleSource.Processor<C, C> processor = newSingleProcessor();
+            channel.pipeline().addLast(new ChannelInboundHandlerAdapter() {
+                @Override
+                public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) {
+                    if (evt instanceof SslHandshakeCompletionEvent) {
+                        SslHandshakeCompletionEvent event = (SslHandshakeCompletionEvent) evt;
+                        if (event.isSuccess()) {
+                            processor.onSuccess(connection);
+                        } else {
+                            processor.onError(event.cause());
+                        }
+                    }
+                    ctx.fireUserEventTriggered(evt);
+                }
+            });
+
+            final DeferSslHandler deferSslHandler = channel.pipeline().get(DeferSslHandler.class);
+            if (deferSslHandler == null) {
+                return failed(new IllegalStateException("Failed to find a handler of type " +
+                        DeferSslHandler.class + " in channel pipeline."));
+            }
+            deferSslHandler.ready();
+
+            // There is no need to apply offloading explicitly (despite completing `processor` on the EventLoop)
+            // because `payloadBody()` will be offloaded according to the strategy for the request.
+            return response.payloadBodyAndTrailers().ignoreElements().concat(fromSource(processor));
         }
-        deferSslHandler.ready();
-
-        // There is no need to apply offloading explicitly (despite completing `processor` on the EventLoop)
-        // because `payloadBody()` will be offloaded according to the strategy for the request.
-        return response.payloadBodyAndTrailers().ignoreElements().concat(fromSource(processor));
     }
 
     @Override
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019-2020 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,12 +19,11 @@ import io.servicetalk.client.api.ConnectionFactory;
 import io.servicetalk.client.api.ConnectionFactoryFilter;
 import io.servicetalk.client.api.DelegatingConnectionFactory;
 import io.servicetalk.concurrent.SingleSource;
-import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
 import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.concurrent.api.Single.TerminalSignalConsumer;
 import io.servicetalk.http.api.FilterableStreamingHttpConnection;
 import io.servicetalk.http.api.HttpExecutionStrategy;
 import io.servicetalk.http.api.HttpExecutionStrategyInfluencer;
-import io.servicetalk.http.api.StreamingHttpRequestResponseFactory;
 import io.servicetalk.transport.netty.internal.DeferSslHandler;
 import io.servicetalk.transport.netty.internal.NettyConnectionContext;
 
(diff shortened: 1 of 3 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -100,28 +103,43 @@ public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt
...
+                    @Override
+                    public void cancel() {
+                        closeConnection(c);
+                    }
+                 });
+            } catch (Exception e) {
+                closeConnection(c);
+                return failed(e);
+            }
+            });
```
normanmaurer, line 135: nit: looks like formatting is off...

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -74,14 +70,21 @@ private ProxyFilter(final ConnectionFactory<ResolvedAddress, C> delegate) {
...
                 // We currently only have access to a StreamingHttpRequester, which means we are forced to provide an
                 // HttpExecutionStrategy. Because we can't be sure if there is any blocking code in the connection
                 // filters we use the default strategy which should offload everything to be safe.
-                c.request(defaultStrategy(),
-                            reqRespFactory.connect(connectAddress).addHeader(CONTENT_LENGTH, ZERO))
+                return c.request(defaultStrategy(), c.connect(connectAddress).addHeader(CONTENT_LENGTH, ZERO))
                  .flatMap(response -> {
-                if (SUCCESSFUL_2XX.contains(response.status())) {
+                    // Drain response payload body asynchronously as we are not interested in it:
+                    response.payloadBodyAndTrailers().ignoreElements().subscribe();
```
NiteshKant, line 81: Dangling subscribes are usually a bad idea. As this code is not composing the subscribe, if the returned `Single` from `newConnection` is cancelled, we will still keep consuming the response. Worse if the response does not complete, we will keep this subscribe around forever (and so the resources associated with the subscribe). 

What we were doing before, i.e. composing the response draining with the `Single` returned from this `flatMap` is the correct thing to do.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -100,28 +103,43 @@ public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt
...
+            } catch (Exception e) {
+                closeConnection(c);
+                return failed(e);
+            }
+            });
         }
     }
 
+    private static <C extends FilterableStreamingHttpConnection> void closeConnection(C connection) {
+        connection.closeAsync().subscribe();
```
NiteshKant, line 140: Same as above, we should not be doing dangling subscribes, instead compose connection closure with error response.
Cancel path will be an exception to this rule as cancels are not composable.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -100,28 +103,43 @@ public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt
...
+                    public void onError(final Throwable ignore) {
+                        closeConnection(c);
+                    }
+
+                    @Override
+                    public void cancel() {
+                        closeConnection(c);
+                    }
+                 });
+            } catch (Exception e) {
```
NiteshKant, line 131:
```suggestion
            } catch (Throwable t) {
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -100,28 +103,43 @@ public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt
...
+                    }
+
+                    @Override
+                    public void onError(final Throwable ignore) {
+                        closeConnection(c);
+                    }
+
+                    @Override
+                    public void cancel() {
+                        closeConnection(c);
```
NiteshKant, line 128: It seems cancel path may leak connection in general if the connection got emitted from here but the source got cancelled afterwards. eg: any operator added after this proxy filter may not send the connection further if it sees a `cancel()`.

idelpivnitskiy: Interesting observation. I was focused on cancel handling before `onSuccess`. For example, before the handshake completes: https://github.com/apple/servicetalk/pull/1002/files#diff-09ab1790ede42935bdec34cd0d96560cR234-R251

Will you suggest to use `whenCancel` operator and always close the connection on cancellation? Because close is idempotent it should not matter if this connection is already closed by someone else. Is there any use-case when the connection is still useful after cancel?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
idelpivnitskiy: Good catch! Two questions:
1. If we concat with failed Single, there is a chance that `connection.closeAsync()` may also fail. Should we worry about handling that case and adding the original exception as a suppressed one for the failed closure?
2. Should we make the same change for RRLB as a follow-up? Currently, we just subscribe to closeAsync and return an exception. For example: https://github.com/apple/servicetalk/blob/52532d402b8f7e33670942794f54388cae6844c0/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java#L279-L285

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
NiteshKant: This is a more broader problem and it can not be handled specifically in this class. As `cancel()`s can come after a success, we really do not know if the cancel has come due to the operation was cancelled by the user or due to some operators sending a cancel for the previous source when they move on to the next source (eg: `concat()`).

I commented here to highlight the problem but this PR does not have to be blocked for this change. What you have here is sufficient in isolation for this class.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
NiteshKant: A1: As `closeAsync()` is idempotent and not expected to fail I will avoid handling that case and make the control flow complex here.

A2: Good catch, yes please do a follow-up

idelpivnitskiy: Follow-up for RRLB: https://github.com/apple/servicetalk/pull/1006

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -74,51 +68,63 @@ private ProxyFilter(final ConnectionFactory<ResolvedAddress, C> delegate) {
...
+                try {
+                    // We currently only have access to a StreamingHttpRequester, which means we are forced to provide
+                    // an HttpExecutionStrategy. Because we can't be sure if there is any blocking code in the
+                    // connection filters we use the default strategy which should offload everything to be safe.
+                    return c.request(defaultStrategy(), c.connect(connectAddress).addHeader(CONTENT_LENGTH, ZERO))
+                            .flatMap(response -> handleConnectResponse(c, response))
+                            // Close recently created connection in case of any error while it connects to the proxy
+                            // or cancellation:
+                            .recoverWith(t -> c.closeAsync().concat(failed(t)))
+                            .whenCancel(() -> c.closeAsync().subscribe());
```
idelpivnitskiy, line 81: > we really do not know if the cancel has come due to the operation was canceled by the user or due to some operators sending a cancel for the previous source when they move on to the next source (eg: `concat()`).

Can any operator cancel after success? IIUC they cancel the previous source only for non-success/non-complete cases.

LMK if I need to revert `whenFinally` here to prevent closure on cancel after success.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -74,51 +68,63 @@ private ProxyFilter(final ConnectionFactory<ResolvedAddress, C> delegate) {
...
-                                    processor.onSuccess(c);
-                                } else {
-                                    processor.onError(event.cause());
-                                }
-                            }
-                            ctx.fireUserEventTriggered(evt);
-                        }
-                    });
+    private Single<C> handleConnectResponse(final C connection, final StreamingHttpResponse response) {
+        try {
```
idelpivnitskiy, line 90: Moved this logic to the different method because nested `try` blocks make code indentation awful.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
NiteshKant: > Can any operator cancel after success? 

Yes they do, consider this `connFactory.newConnection().concat(executor.timer(1, MILLISECONDS)` (eg: to add a delay to respond to connect)

`concat()` uses `SequentialCancellable` which [cancels the old](https://github.com/apple/servicetalk/blob/master/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/SequentialCancellable.java#L59-L62) `Cancellable` when the new `Cancellable` is received, which in this case will be after the successful completion of `connFactory.newConnection()`.

More generally, we should not assume anywhere that `cancel` is only received before `success()` as `Cancellable` and `Subscriber` code paths are concurrent.

Scottmitch: `whenCancel` will unconditionally execute the callback when `cancel` is called regardless if the connection has been delivered downstream. If we have already delivered the `connection` we shouldn't later close it (regardless if someone cancels or not). In addition to this being the expected control flow, the RS spec has some rules which discuss cancel being a no-op after a terminal signal is delivered [1][2].

`afterFinally(SingleTerminalSignalConsumer<T> doFinally)` happens to enforce "only a single callback will be executed" but may still result in invoking the `onCancel()` call back and also calling the downstream `Subscriber#onSuccess(...)` for the following reasons:
- Subscription can be invoked on a different thread
- Data/terminal signals may still be delivered after `cancel` [3]

So `afterFinally` is an improvement over `afterCancel`, but still isn't ideal because we may deliver a closed object (and/or invoke `closeAsync()` concurrently).

[1] https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#1.6
> If a Publisher signals either onError or onComplete on a Subscriber, that Subscriber’s Subscription MUST be considered cancelled.

[2] https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#3.7
> After the Subscription is cancelled, additional Subscription.cancel() MUST be NOPs.

[3] https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#2.8
> A Subscriber MUST be prepared to receive one or more onNext signals after having called Subscription.cancel() if there are still requested elements pending [see 3.12]. Subscription.cancel() does not guarantee to perform the underlying cleaning operations immediately.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -74,51 +68,63 @@ private ProxyFilter(final ConnectionFactory<ResolvedAddress, C> delegate) {
...
+            if (response.status().statusClass() != SUCCESSFUL_2XX) {
+                return response.payloadBodyAndTrailers().ignoreElements().concat(failed(
```
Scottmitch, line 92: this could lead to unnecessary resource (CPU, memory, network, etc...) consumption if we continuously read data that we don't care about (potential DOS vector). Can you clarify why this is necessary when we are hard-closing the connection?

idelpivnitskiy: Good catch!
I thought about this while was working on this PR. But because it was deferred until after #990 is merged I forgot to ask about it. Fixed.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
NiteshKant: >>  TBH don't see where it cancels the old Cancellable.

Aah, you are correct. I misread under an older assumption that we `cancel()` the previous `Cancellable`. 

Anyways, for other reasons me and Scott mention, unconditional `close()` upon `cancel()` isn't correct.

idelpivnitskiy: Updated to use `whenFinally` https://github.com/apple/servicetalk/pull/1002/commits/d2bf22f6b3dd9d9f8a82a7193fd159b4ca069d50

idelpivnitskiy:
Build failed because of an issue with docker.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
NiteshKant: > So afterFinally is an improvement over afterCancel, but still isn't ideal because we may deliver a closed object (and/or invoke closeAsync() concurrently).

Ok ya this seems to be a problem. Can we remove the close-on-cancel part for now? 

Connection lifetime is anyways a problem in such situation out of the context of this filter as mentioned in https://github.com/apple/servicetalk/pull/1002#discussion_r404474939. 

Lets fix the obvious issue of leaking connection for non-200 responses and then handle lifecycle on cancel/early termination later.

idelpivnitskiy: Removed in https://github.com/apple/servicetalk/pull/1002/commits/57e470b5307862d543c41e8274b3fc5ee634cfa7 and created #1010.

idelpivnitskiy:
jdk8 build failed with timeout

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -74,51 +68,59 @@ private ProxyFilter(final ConnectionFactory<ResolvedAddress, C> delegate) {
...
-                    return response.payloadBodyAndTrailers().ignoreElements().concat(
-                            failed(new ProxyResponseException("Bad response from proxy CONNECT " + connectAddress,
-                                    response.status())));
-                }
-            }));
+            // There is no need to apply offloading explicitly (despite completing `processor` on the
+            // EventLoop) because `payloadBody()` will be offloaded according to the strategy for the
+            // request.
+            return response.payloadBodyAndTrailers().ignoreElements().concat(fromSource(processor));
+        } catch (Throwable t) {
```
NiteshKant, line 122: This `try-catch` isn't required as it is called from the `flatMap` which has a `recoverWith`, so if anything does throw from here, `flatMap` will correctly terminate the control flow with the same error which will then be intercepted by `recoverWith()` closing the connection.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -74,51 +68,59 @@ private ProxyFilter(final ConnectionFactory<ResolvedAddress, C> delegate) {
...
+            return delegate().newConnection(resolvedAddress).flatMap(c -> {
+                try {
+                    // We currently only have access to a StreamingHttpRequester, which means we are forced to provide
+                    // an HttpExecutionStrategy. Because we can't be sure if there is any blocking code in the
+                    // connection filters we use the default strategy which should offload everything to be safe.
+                    return c.request(defaultStrategy(), c.connect(connectAddress).addHeader(CONTENT_LENGTH, ZERO))
+                            .flatMap(response -> handleConnectResponse(c, response))
+                            // Close recently created connection in case of any error while it connects to the proxy:
+                            .recoverWith(t -> c.closeAsync().concat(failed(t)));
+                } catch (Throwable t) {
```
NiteshKant, line 80: I think we are being overly paranoid here about the calls to `c.request()` or `c.connect()` throwing. Any method returning an asynchronous source is not expected to throw. Having said that it is not a big deal so its ok as it is, I will leave it to you to take a call on this.

PR #1004, opened 2020-04-07
AsyncContextInMemoryScopeManager to restore previous Scope on close()

Scottmitch:
Motivation:
AsyncContextInMemoryScopeManager currently doesn't restore the previous
Scope when the current Scope is closed. This behavior is provided for
ThreadLocalScope and allows users to use try-with-resources type
constructs to create sub-spans more easily.

Modifications:
- AsyncContextInMemoryScopeManager to save/restore the previous Scope
when a new Scope is activated
- AsyncContextInMemoryScopeManager to remove the currentScope() hack so
lifetime of Scopes is more narrowly/correctly constrainted. This means
that filter ordering is important and use of after* operators should be
done with care!

Result:
AsyncContextInMemoryScopeManager has more consistent behavior with
ThreadLocalScope, and lifetimes of Scopes are more narrowly/correctly
constrainted.

### DIFF
--- a/servicetalk-log4j2-mdc-utils/src/testFixtures/java/io/servicetalk/log4j2/mdc/utils/LoggerStringWriter.java
+++ b/servicetalk-log4j2-mdc-utils/src/testFixtures/java/io/servicetalk/log4j2/mdc/utils/LoggerStringWriter.java
@@ -19,9 +19,9 @@ import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.core.Appender;
 import org.apache.logging.log4j.core.LoggerContext;
-import org.apache.logging.log4j.core.StringLayout;
 import org.apache.logging.log4j.core.appender.WriterAppender;
 import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.config.LoggerConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -42,6 +42,7 @@ import static org.hamcrest.Matchers.greaterThanOrEqualTo;
 
 public final class LoggerStringWriter {
     private static final Logger LOGGER = LoggerFactory.getLogger(LoggerStringWriter.class);
+    private static final String APPENDER_NAME = "writer";
     @Nullable
     private static StringWriter logStringWriter;
 
@@ -54,6 +55,13 @@ public final class LoggerStringWriter {
      */
     public static void reset() {
         getStringWriter().getBuffer().setLength(0);
+    }
+
+    /**
+     * Remove the underlying in-memory log appender.
+     */
+    public static void remove() {
+        removeStringWriter();
     }
 
     /**
@@ -157,21 +165,41 @@ public final class LoggerStringWriter {
         return logStringWriter;
     }
 
+    private static synchronized void removeStringWriter() {
+        if (logStringWriter == null) {
+            return;
+        }
+        removeWriterAppender((LoggerContext) LogManager.getContext(false));
+        logStringWriter = null;
+    }
+
     private static StringWriter addWriterAppender(final LoggerContext context, Level level) {
         final Configuration config = context.getConfiguration();
         final StringWriter writer = new StringWriter();
 
         final Map.Entry<String, Appender> existing = config.getAppenders().entrySet().iterator().next();
         final WriterAppender writerAppender = WriterAppender.newBuilder()
-                .setName("writer")
-                .setLayout((StringLayout) existing.getValue().getLayout())
+                .setName(APPENDER_NAME)
+                .setLayout(existing.getValue().getLayout())
                 .setTarget(writer)
                 .build();
-        writerAppender.start();
 
-        config.addAppender(writerAppender);
+        writerAppender.start();
         config.getRootLogger().addAppender(writerAppender, level, null);
 
         return writer;
+    }
+
+    private static void removeWriterAppender(final LoggerContext context) {
+        final Configuration config = context.getConfiguration();
+        LoggerConfig rootConfig = config.getRootLogger();
+        // Stopping the logger is subject to race conditions where logging during cleanup on global executor
+        // may still try to log and raise an error.
+        WriterAppender writerAppender = (WriterAppender) rootConfig.getAppenders().get(APPENDER_NAME);
+        if (writerAppender != null) {
+            writerAppender.stop(0, NANOSECONDS);
+        }
+        // Don't remove directly from map, because the root logger also cleans up filters.
+        rootConfig.removeAppender(APPENDER_NAME);
     }
 }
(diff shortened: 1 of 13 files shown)


servicetalk-opentracing-http/src/main/java/io/servicetalk/opentracing/http/TracingHttpServiceFilter.java
```diff
@@ -40,6 +43,12 @@
 
 /**
  * A {@link StreamingHttpService} that supports open tracing.
+ * <p>
+ * Append this filter before others that are expected to see {@link Scope} for this request/response. Filters appended
```
Scottmitch, line 47: discussed with @NiteshKant offline on possibility of splitting this filter into two:
- TracerExtractor
- TracerCloser

however the complexity of implementation would be high to hand-off ownership between filters. Also I will come up with a PR to change `when*` operators to default to `before*` (instead of `after*`) to reduce the risk of filters inadvertently applying an "after" operator which would mean their filter response processing may run without trace info.

servicetalk-opentracing-asynccontext/src/main/java/io/servicetalk/opentracing/asynccontext/AsyncContextInMemoryScopeManager.java
```diff
@@ -31,62 +31,53 @@
  */
 public final class AsyncContextInMemoryScopeManager implements InMemoryScopeManager {
     private static final AsyncContextMap.Key<AsyncContextInMemoryScope> SCOPE_KEY = newKey("opentracing");
```
NiteshKant, line 33:
```suggestion
    private static final AsyncContextMap.Key<InMemoryScope> SCOPE_KEY = newKey("opentracing");
```

servicetalk-opentracing-http/src/main/java/io/servicetalk/opentracing/http/TracingHttpRequesterFilter.java
```diff
@@ -34,13 +35,21 @@
...
 import static io.opentracing.tag.Tags.HTTP_URL;
 import static io.opentracing.tag.Tags.SPAN_KIND;
 import static io.opentracing.tag.Tags.SPAN_KIND_CLIENT;
 
 /**
  * An HTTP filter that supports open tracing.
+ * <p>
+ * Append this filter before others that are expected to to see {@link Scope} for this request/response. Filters
+ * appended after this filter that use operators with the <strong>after*</strong> prefix on the
+ * {@link StreamingHttpResponse#transformRawPayloadBody(UnaryOperator) response payload body} (e.g.
```
NiteshKant, line 50: This disclaimer also applied for metadata in case of error, you may want to add that.

servicetalk-opentracing-http/src/main/java/io/servicetalk/opentracing/http/TracingHttpServiceFilter.java
NiteshKant: Yes agreed! 

It seems logical to define the scope of an operation within the context of a filter. `After*` operators seem to extend that scope after everything else outside that scope. Drawing an analogy to an imperative programming model:

```java
public Result someMethod(Caller caller) {

    State state;
    try {
        state = new State();
        Result result = doWork();
        return result;
    } finally {
         // Code here is analogous to beforeFinally async operators.
         state.cleanup(); 
        // In order to demonstrate cleanup from the code that is calling this method. 
        // Assume caller is passed to this method.
        caller.cleanup();
    }
// Code below this is analogous to async afterFinally operators.
}
```

servicetalk-opentracing-http/src/test/java/io/servicetalk/opentracing/http/TracingHttpRequesterFilterTest.java
```diff
@@ -189,4 +221,54 @@ private static ServerContext buildServer() throws Exception {
     private static String toStringOrNull(@Nullable CharSequence cs) {
         return cs == null ? null : cs.toString();
     }
+
+    private static final class TestTracingLoggerFilter implements StreamingHttpClientFilterFactory {
+        private final String[] logLinePrefix;
+
+        TestTracingLoggerFilter(final String[] logLinePrefix) {
+            this.logLinePrefix = requireNonNull(logLinePrefix);
```
normanmaurer, line 229: I guess it doesn't matter as its just used in the test but usually you should call `.clone()` on an array when it is passed via the constructor to ensure it can't be modified later on.

servicetalk-opentracing-http/src/test/java/io/servicetalk/opentracing/http/TracingHttpServiceFilterTest.java
```diff
@@ -176,4 +210,54 @@ public void tracerThrowsReturnsErrorResponse() throws Exception {
             }
         }
     }
+
+    private static final class TestTracingLoggerFilter implements StreamingHttpServiceFilterFactory {
+        private final String[] logLinePrefix;
+
+        private TestTracingLoggerFilter(final String[] logLinePrefix) {
+            this.logLinePrefix = requireNonNull(logLinePrefix);
```
normanmaurer, line 218: same comment as above

servicetalk-opentracing-http/src/test/resources/log4j2-test.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright © 2018 Apple Inc. and the ServiceTalk project authors
```
normanmaurer, line 3: 2020 ?

servicetalk-log4j2-mdc-utils/src/testFixtures/java/io/servicetalk/log4j2/mdc/utils/LoggerStringWriter.java
```diff
@@ -157,21 +165,41 @@ private static synchronized StringWriter getStringWriter() {
         return logStringWriter;
     }
 
+    private static synchronized void removeStringWriter() {
```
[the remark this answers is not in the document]
Scottmitch: I'll leave this as is to avoid exposing the `static synchronized` on the public method and for consistency with the `reset()` method structure.

servicetalk-opentracing-http/src/test/java/io/servicetalk/opentracing/http/TestUtils.java
```diff
@@ -43,6 +54,36 @@ private static String getRandomHexString(int numchars) {
...
+                    } catch (Throwable cause) {
+                        cause.addSuppressed(new AssertionError("failed on prefix: " + prefix));
+                        throw cause;
+                    }
+                    break;
+                }
+            }
+            if (!foundMatch) {
+                throw new AssertionError("could not find log line with prefix: " + prefix);
+            }
```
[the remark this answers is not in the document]
Scottmitch: follow-up PR https://github.com/apple/servicetalk/pull/1009

PR #1005, opened 2020-04-07
Change when* operators to before* instead of after*

Scottmitch:
Motivation:
The when* operators provide call backs to take action when that action
isn't sensitive to execution order. The current implementation of the
when* operators uses after* semantics, despite the order being explicitly
undefiend by the API. The after* semantics may result in user code
executing in unexpected times, espeically when local state is involved.
For example tracing filters utilize AsyncContext and trigger the state
cleanup on the response payload Publisher termination. This filter must
be appended first in order for the trace information to be initialized
correctly and if any subsequent filters apply a payload Publisher
transformation with after* semantics the terminal methods will be
invoked after the tracing Scope has been closed. If subsequent operators
use before* semantics they will not be subject to this ordering issue.

Modifications:
- Change Publisher, Single, Completable when* operators to use before*
semantics instead of after* semantics.

Result:
when* operators use before* semantics to reduce the risk that local
state managed outside the scope of the operator flow is cleaned up when
user code executes.

### DIFF
--- a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
+++ b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
@@ -122,7 +122,7 @@ public final class TcpServerBinder {
                                     // subscribeOn is required to offload calls to connectionAcceptor#accept
                                     .subscribeOn(executionContext.executor()));
                 }
-                connectionSingle.afterOnError(cause -> {
+                connectionSingle.beforeOnError(cause -> {
                     // Getting the remote-address may involve volatile reads and potentially a syscall, so guard it.
                     if (LOGGER.isDebugEnabled()) {
                         LOGGER.debug("Failed to create a connection for remote address {}", channel.remoteAddress(),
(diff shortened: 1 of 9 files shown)


servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
```diff
@@ -122,7 +122,7 @@ protected void initChannel(Channel channel) {
                                     // subscribeOn is required to offload calls to connectionAcceptor#accept
                                     .subscribeOn(executionContext.executor()));
                 }
-                connectionSingle.whenOnError(cause -> {
+                connectionSingle.afterOnError(cause -> {
```
NiteshKant, line 125: Is `before` not appropriate here as we are sending an error and hence the channel will not be usable?

Scottmitch: I guess it shouldn't matter ... was just trying to keep behavior the same. I will make it explicit `before*` rather than having to deal with this ambiguity in the future.

PR #1006, opened 2020-04-07
`RoundRobinLoadBalancer` should compose connection closure with error response

idelpivnitskiy:
Motivation:

Dangling subscribes is usually a bad idea. As RRLB is not composing the
subscribe, it will still keep closing the connection after it returns a failure.
Worse if the closure does not complete, it will keep this subscribe around
forever (and so the resources associated with the subscribe).

Modification:

- Compose new connection closure with the `Single.failed` returned from
`flatMap`;
- Add a test to verify that new connection is closed if it's rejected by
selector;

Result:

A new connection is closed before RRLB fails the returned `Single` from
`selectConnection`.

### DIFF
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018-2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2018-2020 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -277,18 +277,17 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
                     // Invoke the selector before adding the connection to the pool, otherwise, connection can be used
                     // concurrently and hence a new connection can be rejected by the selector.
                     if (!selector.test(newCnx)) {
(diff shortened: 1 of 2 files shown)


servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -299,18 +298,20 @@ public void onComplete() {
...
-                                    newCnx.closeAsync().subscribe();
+                                    needClose = true;
                                     break;
                                 }
                                 existing = connections;
                             }
 
-                            return failed(new IllegalStateException("LoadBalancer has closed"));
+                            final Single<C> lbClosed = failed(new IllegalStateException("LoadBalancer has closed"));
+                            return needClose ? newCnx.closeAsync().concat(lbClosed) : lbClosed;
```
idelpivnitskiy, line 308: I use `needClose` to keep the existing behavior. But because connection closure is idempotent we may simplify and always close it.

NiteshKant: I think it is better to keep `needClose` since we are anyways doing the heavy-lifting to figure out who should close the connection anyways.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -277,18 +277,17 @@ public void onComplete() {
                     // Invoke the selector before adding the connection to the pool, otherwise, connection can be used
                     // concurrently and hence a new connection can be rejected by the selector.
                     if (!selector.test(newCnx)) {
-                        newCnx.closeAsync().subscribe();
                         // Failure in selection could be temporary, hence add it to the queue and be consistent with the
                         // fact that select failure does not close a connection.
-                        return failed(new ConnectionRejectedException("Newly created connection " + newCnx +
-                                " rejected by the selection filter."));
+                        return newCnx.closeAsync().concat(failed(new ConnectionRejectedException(
```
NiteshKant, line 282: Can you verify if we have a test that checks the connection is closed when we get the error?

idelpivnitskiy:

PR #1008, opened 2020-04-07
Promote Single.TerminalSignalConsumer to top level interface

Scottmitch:
Motivation:
TerminalSignalConsumer for Completable and Publisher is a top level
type, but Single has an internal type for the same purpose. This makes
it more challenging to expose utilities classes in a consistent manner
and also may lead to naming collisions.

Modifications:
- Move Single.TerminalSignalConsumer to a top level interface and rename
to SingleTerminalSignalConsumer to avoid naming collisions.
- Expose RunnableTerminalSignalConsumer and re-use it in
BeforeFinallyOnHttpResponseOperator.

Result:
More consistent type definition scheme for TerminalSignalConsumer and
SingleTerminalSignalConsumer and ability to expose conversion/utility
methods outside the scope of the Source.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RunnableTerminalSignalConsumer.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RunnableTerminalSignalConsumer.java
@@ -17,17 +17,10 @@ package io.servicetalk.concurrent.api;
 
 import static java.util.Objects.requireNonNull;
 
-/**
- * A {@link TerminalSignalConsumer} where each method executes a {@link Runnable#run()}.
- */
-public final class RunnableTerminalSignalConsumer implements TerminalSignalConsumer {
+final class RunnableTerminalSignalConsumer implements TerminalSignalConsumer {
     private final Runnable onFinally;
 
-    /**
-     * Create a new instance.
-     * @param onFinally The {@link Runnable} to run for each method.
-     */
-    public RunnableTerminalSignalConsumer(final Runnable onFinally) {
+    RunnableTerminalSignalConsumer(final Runnable onFinally) {
         this.onFinally = requireNonNull(onFinally);
     }
 
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleTerminalSignalConsumer.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleTerminalSignalConsumer.java
@@ -45,4 +45,14 @@ public interface SingleTerminalSignalConsumer<T> {
      * Callback to indicate termination via {@link Cancellable#cancel()}.
      */
     void cancel();
+
+    /**
+     * Create a {@link SingleTerminalSignalConsumer} where each method executes a {@link Runnable#run()}.
+     * @param runnable The {@link Runnable} which is invoked in each method of the returned
+     * {@link SingleTerminalSignalConsumer}.
+     * @return a {@link SingleTerminalSignalConsumer} where each method executes a {@link Runnable#run()}.
+     */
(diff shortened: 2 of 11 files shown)


servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/BeforeFinallyOnHttpResponseOperator.java
```diff
@@ -254,28 +255,4 @@ public void onComplete() {
             // Ignore.
         }
     }
-
-    private static final class RunnableTerminalSignalConsumer implements TerminalSignalConsumer {
```
Scottmitch, line 258: I realize I requested to make the conversion classes internal when these types were recently introduced, but if we are going to have to duplicate the code lets just expose it so we can share.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleTerminalSignalConsumer.java
```diff
@@ -0,0 +1,48 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * A contract that provides discrete callbacks for various ways in which a {@link SingleSource.Subscriber} can
+ * terminate.
+ *
+ * @param <T> Type of the result of the {@link Single}.
+ */
+public interface SingleTerminalSignalConsumer<T> {
```
Scottmitch, line 29: not strong on internal type vs top level type, but leaning toward external type so it is more consistent with `TerminalSignalConsumer` and we can expose `RunnableSingleTerminalSignalConsumer` in the future if necessary without cluttering `Single`'s public exposure.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/BeforeFinallyOnHttpResponseOperator.java
idelpivnitskiy: Sad that we can not place it in some `-internal` module due to the dependencies between modules to share it with `http-utils`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RunnableTerminalSignalConsumer.java
```diff
@@ -17,11 +17,17 @@
 
 import static java.util.Objects.requireNonNull;
 
-final class RunnableTerminalSignalConsumer implements TerminalSignalConsumer {
-
+/**
+ * A {@link TerminalSignalConsumer} where each method executes a {@link Runnable#run()}.
+ */
+public final class RunnableTerminalSignalConsumer implements TerminalSignalConsumer {
```
idelpivnitskiy, line 23: If we are ok to expose it as public API, alternative could be to add a static method `TerminalSignalConsumer.from(Runnable)`. That will simplify discoverability and be consistent with other interfaces, like `HostAndPort`, factories, initializers, etc.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/BeforeFinallyOnHttpResponseOperator.java
Scottmitch: agreed the API needs to be public in `concurrent-api` so we would need to push the APIs down to the `concurrent` module in order to put the utility classes in an `-internal` module.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RunnableTerminalSignalConsumer.java
Scottmitch: done

PR #1011, opened 2020-04-14
Add Publisher#flatMapMerge operator

Scottmitch:
Motivation:
Publisher#flatMapMerge is a basic building block for each element of a
Publisher to map into another Publisher.

Modifications:
- Add Publisher#flatMapMerge operator

Result:
Each item emitted on a Publisher can be mapped to another Publisher.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -189,8 +189,15 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Each element of this {@link Publisher} is mapped into a {@link Publisher} (potentially of a different type) and
-     * flatten all signals emitted from each mapped {@link Publisher} into the returned {@link Publisher}.
+     * Map each element of this {@link Publisher} into a {@link Publisher}&lt;{@link R}&gt; and flatten all signals
+     * emitted from each mapped {@link Publisher}&lt;{@link R}&gt; into the returned
+     * {@link Publisher}&lt;{@link R}&gt;.
+     * <p>
+     * To control the amount of concurrent processing done by this operator see
+     * {@link #flatMapMerge(Function, int, int)}.
+     * <p>
+     * This method is similar to {@link #map(Function)} but the result is an asynchronous stream, and provides a data
+     * transformation in sequential programming similar to:
      * <pre>{@code
      *     ExecutorService e = ...;
      *     List<Future<List<R>>> futures = ...; // assume this is thread safe
@@ -219,8 +226,12 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Each element of this {@link Publisher} is mapped into a {@link Publisher} (potentially of a different type) and
-     * flatten all signals emitted from each mapped {@link Publisher} into the returned {@link Publisher}.
+     * Map each element of this {@link Publisher} into a {@link Publisher}&lt;{@link R}&gt; and flatten all signals
+     * emitted from each mapped {@link Publisher}&lt;{@link R}&gt; into the returned
+     * {@link Publisher}&lt;{@link R}&gt;.
+     * <p>
+     * This method is similar to {@link #map(Function)} but the result is an asynchronous stream, and provides a data
+     * transformation in sequential programming similar to:
      * <pre>{@code
      *     ExecutorService e = ...;
      *     List<Future<List<R>>> futures = ...; // assume this is thread safe
@@ -252,12 +263,20 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Each element of this {@link Publisher} is mapped into a {@link Publisher} (potentially of a different type) and
-     * flatten all signals emitted from each mapped {@link Publisher} into the returned {@link Publisher}.
+     * Map each element of this {@link Publisher} into a {@link Publisher}&lt;{@link R}&gt; and flatten all signals
+     * emitted from each mapped {@link Publisher}&lt;{@link R}&gt; into the returned
+     * {@link Publisher}&lt;{@link R}&gt;.
+     * <p>
      * This is the same as {@link #flatMapMerge(Function)} just that if any mapped {@link Publisher} returned by
      * {@code mapper}, terminates with an error, the returned {@link Publisher} will not immediately terminate. Instead,
      * it will wait for this {@link Publisher} and all mapped {@link Publisher}s to terminate and then terminate the
      * returned {@link Publisher} with all errors emitted by the mapped {@link Publisher}s.
+     * <p>
+     * To control the amount of concurrent processing done by this operator see
+     * {@link #flatMapMergeDelayError(Function, int, int)}.
+     * <p>
+     * This method is similar to {@link #map(Function)} but the result is an asynchronous stream, and provides a data
+     * transformation in sequential programming similar to:
      * <pre>{@code
      *     Executor e = ...;
      *     List<T> tResults = resultOfThisPublisher();
@@ -294,12 +313,17 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Each element of this {@link Publisher} is mapped into a {@link Publisher} (potentially of a different type) and
-     * flatten all signals emitted from each mapped {@link Publisher} into the returned {@link Publisher}.
-     * This is the same as {@link #flatMapMerge(Function, int, int)} just that if any mapped {@link Publisher}
-     * returned by {@code mapper}, terminates with an error, the returned {@link Publisher} will not immediately
-     * terminate. Instead, it will wait for this {@link Publisher} and all mapped {@link Publisher}s to terminate and
-     * then terminate the returned {@link Publisher} with all errors emitted by the mapped {@link Publisher}s.
+     * Map each element of this {@link Publisher} into a {@link Publisher}&lt;{@link R}&gt; and flatten all signals
+     * emitted from each mapped {@link Publisher}&lt;{@link R}&gt; into the returned
+     * {@link Publisher}&lt;{@link R}&gt;.
+     * <p>
+     * This is the same as {@link #flatMapMerge(Function)} just that if any mapped {@link Publisher} returned by
+     * {@code mapper}, terminates with an error, the returned {@link Publisher} will not immediately terminate. Instead,
+     * it will wait for this {@link Publisher} and all mapped {@link Publisher}s to terminate and then terminate the
+     * returned {@link Publisher} with all errors emitted by the mapped {@link Publisher}s.
+     * <p>
+     * This method is similar to {@link #map(Function)} but the result is an asynchronous stream, and provides a data
+     * transformation in sequential programming similar to:
      * <pre>{@code
      *     Executor e = ...;
      *     List<T> tResults = resultOfThisPublisher();
@@ -339,8 +363,9 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Turns every item emitted by this {@link Publisher} into a {@link Single} and emits the items emitted by each of
-     * those {@link Single}s.
+     * Map each element of this {@link Publisher} into a {@link Single}&lt;{@link R}&gt; and flatten all signals
+     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned
+     * {@link Publisher}&lt;{@link R}&gt;.
      * <p>
      * To control the amount of concurrent processing done by this operator see
      * {@link #flatMapMergeSingle(Function, int)}.
@@ -377,8 +402,9 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Turns every item emitted by this {@link Publisher} into a {@link Single} and emits the items emitted by each of
-     * those {@link Single}s.
+     * Map each element of this {@link Publisher} into a {@link Single}&lt;{@link R}&gt; and flatten all signals
+     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned
+     * {@link Publisher}&lt;{@link R}&gt;.
      * <p>
      * This method is similar to {@link #map(Function)} but the result is asynchronous, and provides a data
      * transformation in sequential programming similar to:
@@ -415,8 +441,11 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Turns every item emitted by this {@link Publisher} into a {@link Single} and emits the items emitted by each of
-     * those {@link Single}s. This is the same as {@link #flatMapMergeSingle(Function, int)} just that if any
+     * Map each element of this {@link Publisher} into a {@link Single}&lt;{@link R}&gt; and flatten all signals
+     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned
+     * {@link Publisher}&lt;{@link R}&gt;.
+     * <p>
+     * The behavior is the same as {@link #flatMapMergeSingle(Function, int)} with the exception that if any
      * {@link Single} returned by {@code mapper}, terminates with an error, the returned {@link Publisher} will not
      * immediately terminate. Instead, it will wait for this {@link Publisher} and all {@link Single}s to terminate and
      * then terminate the returned {@link Publisher} with all errors emitted by the {@link Single}s produced by the
@@ -451,7 +480,7 @@ public abstract class Publisher<T> {
      *         return rResults;
      *     }
      *     createAndThrowACompositeException(errors);
-    * }</pre>
+     * }</pre>
      *
      * @param mapper {@link Function} to convert each item emitted by this {@link Publisher} into a {@link Single}.
      * @param <R> Type of items emitted by the returned {@link Publisher}.
@@ -466,8 +495,11 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Turns every item emitted by this {@link Publisher} into a {@link Single} and emits the items emitted by each of
-     * those {@link Single}s. This is the same as {@link #flatMapMergeSingle(Function, int)} just that if any
+     * Map each element of this {@link Publisher} into a {@link Single}&lt;{@link R}&gt; and flatten all signals
+     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned
+     * {@link Publisher}&lt;{@link R}&gt;.
+     * <p>
+     * The behavior is the same as {@link #flatMapMergeSingle(Function, int)} with the exception that if any
      * {@link Single} returned by {@code mapper}, terminates with an error, the returned {@link Publisher} will not
      * immediately terminate. Instead, it will wait for this {@link Publisher} and all {@link Single}s to terminate and
      * then terminate the returned {@link Publisher} with all errors emitted by the {@link Single}s produced by the
@@ -516,10 +548,11 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Turns every item emitted by this {@link Publisher} into a {@link Completable} and terminate the returned
-     * {@link Completable} when all the intermediate {@link Completable}s have terminated successfully or any one of
-     * them has terminated with a failure.
-     * If the returned {@link Completable} should wait for the termination of all remaining {@link Completable}s when
+     * Map each element of this {@link Publisher} into a {@link Completable} and flatten all signals
+     * such that the returned {@link Completable} terminates when all mapped {@link Completable}s have terminated
+     * successfully or any one of them has terminated with a failure.
+     * <p>
+     * If the returned {@link Completable} should wait for the termination of all mapped {@link Completable}s when
      * any one of them terminates with a failure, {@link #flatMapCompletableDelayError(Function)} should be used.
      * <p>
      * To control the amount of concurrent processing done by this operator see
@@ -555,10 +588,11 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Turns every item emitted by this {@link Publisher} into a {@link Completable} and terminate the returned
-     * {@link Completable} when all the intermediate {@link Completable}s have terminated successfully or any one of
-     * them has terminated with a failure.
-     * If the returned {@link Completable} should wait for the termination of all remaining {@link Completable}s when
+     * Map each element of this {@link Publisher} into a {@link Completable} and flatten all signals
+     * such that the returned {@link Completable} terminates when all mapped {@link Completable}s have terminated
+     * successfully or any one of them has terminated with a failure.
+     * <p>
+     * If the returned {@link Completable} should wait for the termination of all mapped {@link Completable}s when
      * any one of them terminates with a failure, {@link #flatMapCompletableDelayError(Function)} should be used.
      * <p>
      * This method is similar to {@link #map(Function)} but the result is asynchronous, and provides a data
@@ -592,12 +626,12 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Turns every item emitted by this {@link Publisher} into a {@link Completable} and terminate the returned
-     * {@link Completable} when all the intermediate {@link Completable}s have terminated. If any {@link Completable}
-     * returned by {@code mapper}, terminates with an error, the returned {@link Completable} will not immediately
-     * terminate. Instead, it will wait for this {@link Publisher} and all {@link Completable}s to terminate and then
-     * terminate the returned {@link Completable} with all errors emitted by the {@link Completable}s produced by the
-     * {@code mapper}.
+     * Map each element of this {@link Publisher} into a {@link Completable} and flatten all signals
+     * such that the returned {@link Completable} terminates when all mapped {@link Completable}s have terminated
+     * successfully or any one of them has terminated with a failure.
+     * <p>
+     * If any mapped {@link Completable} terminates with an error the returned {@link Completable} will not immediately
+     * terminate. Instead, it will wait for this {@link Publisher} and all mapped {@link Completable}s to terminate.
      * <p>
      * To control the amount of concurrent processing done by this operator see
      * {@link #flatMapCompletableDelayError(Function, int)}.
@@ -638,12 +672,12 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Turns every item emitted by this {@link Publisher} into a {@link Completable} and terminate the returned
-     * {@link Completable} when all the intermediate {@link Completable}s have terminated.If any {@link Completable}
-     * returned by {@code mapper}, terminates with an error, the returned {@link Completable} will not immediately
-     * terminate. Instead, it will wait for this {@link Publisher} and all {@link Completable}s to terminate and then
-     * terminate the returned {@link Completable} with all errors emitted by the {@link Completable}s produced by the
-     * {@code mapper}.
+     * Map each element of this {@link Publisher} into a {@link Completable} and flatten all signals
+     * such that the returned {@link Completable} terminates when all mapped {@link Completable}s have terminated
+     * successfully or any one of them has terminated with a failure.
+     * <p>
+     * If any mapped {@link Completable} terminates with an error the returned {@link Completable} will not immediately
+     * terminate. Instead, it will wait for this {@link Publisher} and all mapped {@link Completable}s to terminate.
      * <p>
      * This method is similar to {@link #map(Function)} but the result is asynchronous, and provides a data
      * transformation in sequential programming similar to:
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -251,15 +251,15 @@ public abstract class Publisher<T> {
      * }</pre>
      * @param mapper Convert each item emitted by this {@link Publisher} into another {@link Publisher}.
      * @param maxConcurrency Maximum amount of outstanding upstream {@link Subscription#request(long) demand}.
-     * @param maxMappedDemandHint Hint for the maximum amount of {@link Subscription#request(long) demand} for
+     * @param minMappedDemandHint Hint for the minimum amount of {@link Subscription#request(long) demand} for
      * each mapped {@link Publisher}.
      * @param <R> The type of mapped {@link Publisher}.
      * @return A new {@link Publisher} which flattens the emissions from all mapped {@link Publisher}s.
      * @see <a href="http://reactivex.io/documentation/operators/flatmap.html">ReactiveX flatMap operator.</a>
      */
     public final <R> Publisher<R> flatMapMerge(Function<? super T, ? extends Publisher<? extends R>> mapper,
-                                               int maxConcurrency, int maxMappedDemandHint) {
-        return new PublisherFlatMapMerge<>(this, mapper, false, maxConcurrency, maxMappedDemandHint, executor);
+                                               int maxConcurrency, int minMappedDemandHint) {
+        return new PublisherFlatMapMerge<>(this, mapper, false, maxConcurrency, minMappedDemandHint, executor);
     }
 
     /**
@@ -351,15 +351,15 @@ public abstract class Publisher<T> {
      * }</pre>
      * @param mapper Convert each item emitted by this {@link Publisher} into another {@link Publisher}.
      * @param maxConcurrency Maximum amount of outstanding upstream {@link Subscription#request(long) demand}.
-     * @param maxMappedDemandHint Hint for the maximum amount of {@link Subscription#request(long) demand} for
+     * @param minMappedDemandHint Hint for the minimum amount of {@link Subscription#request(long) demand} for
      * each mapped {@link Publisher}.
      * @param <R> The type of mapped {@link Publisher}.
      * @return A new {@link Publisher} which flattens the emissions from all mapped {@link Publisher}s.
      * @see <a href="http://reactivex.io/documentation/operators/flatmap.html">ReactiveX flatMap operator.</a>
      */
     public final <R> Publisher<R> flatMapMergeDelayError(Function<? super T, ? extends Publisher<? extends R>> mapper,
-                                                         int maxConcurrency, int maxMappedDemandHint) {
-        return new PublisherFlatMapMerge<>(this, mapper, true, maxConcurrency, maxMappedDemandHint, executor);
+                                                         int maxConcurrency, int minMappedDemandHint) {
+        return new PublisherFlatMapMerge<>(this, mapper, true, maxConcurrency, minMappedDemandHint, executor);
     }
 
     /**
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -193,8 +193,7 @@ public abstract class Publisher<T> {
      * emitted from each mapped {@link Publisher}&lt;{@link R}&gt; into the returned
      * {@link Publisher}&lt;{@link R}&gt;.
      * <p>
-     * To control the amount of concurrent processing done by this operator see
-     * {@link #flatMapMerge(Function, int, int)}.
+     * To control the amount of concurrent processing done by this operator see {@link #flatMapMerge(Function, int)}.
      * <p>
      * This method is similar to {@link #map(Function)} but the result is an asynchronous stream, and provides a data
      * transformation in sequential programming similar to:
@@ -251,15 +250,13 @@ public abstract class Publisher<T> {
      * }</pre>
      * @param mapper Convert each item emitted by this {@link Publisher} into another {@link Publisher}.
      * @param maxConcurrency Maximum amount of outstanding upstream {@link Subscription#request(long) demand}.
-     * @param minMappedDemandHint Hint for the minimum amount of {@link Subscription#request(long) demand} for
-     * each mapped {@link Publisher}.
      * @param <R> The type of mapped {@link Publisher}.
      * @return A new {@link Publisher} which flattens the emissions from all mapped {@link Publisher}s.
      * @see <a href="http://reactivex.io/documentation/operators/flatmap.html">ReactiveX flatMap operator.</a>
      */
     public final <R> Publisher<R> flatMapMerge(Function<? super T, ? extends Publisher<? extends R>> mapper,
-                                               int maxConcurrency, int minMappedDemandHint) {
-        return new PublisherFlatMapMerge<>(this, mapper, false, maxConcurrency, minMappedDemandHint, executor);
+                                               int maxConcurrency) {
+        return new PublisherFlatMapMerge<>(this, mapper, false, maxConcurrency, executor);
     }
 
     /**
@@ -273,7 +270,7 @@ public abstract class Publisher<T> {
      * returned {@link Publisher} with all errors emitted by the mapped {@link Publisher}s.
      * <p>
      * To control the amount of concurrent processing done by this operator see
-     * {@link #flatMapMergeDelayError(Function, int, int)}.
+     * {@link #flatMapMergeDelayError(Function, int)}.
      * <p>
      * This method is similar to {@link #map(Function)} but the result is an asynchronous stream, and provides a data
      * transformation in sequential programming similar to:
@@ -351,15 +348,13 @@ public abstract class Publisher<T> {
      * }</pre>
      * @param mapper Convert each item emitted by this {@link Publisher} into another {@link Publisher}.
      * @param maxConcurrency Maximum amount of outstanding upstream {@link Subscription#request(long) demand}.
-     * @param minMappedDemandHint Hint for the minimum amount of {@link Subscription#request(long) demand} for
-     * each mapped {@link Publisher}.
      * @param <R> The type of mapped {@link Publisher}.
      * @return A new {@link Publisher} which flattens the emissions from all mapped {@link Publisher}s.
      * @see <a href="http://reactivex.io/documentation/operators/flatmap.html">ReactiveX flatMap operator.</a>
      */
     public final <R> Publisher<R> flatMapMergeDelayError(Function<? super T, ? extends Publisher<? extends R>> mapper,
-                                                         int maxConcurrency, int minMappedDemandHint) {
-        return new PublisherFlatMapMerge<>(this, mapper, true, maxConcurrency, minMappedDemandHint, executor);
+                                                         int maxConcurrency) {
+        return new PublisherFlatMapMerge<>(this, mapper, true, maxConcurrency, executor);
     }
 
     /**
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -189,6 +189,156 @@ public abstract class Publisher<T> {
     }
 
     /**
+     * Each element of this {@link Publisher} is mapped into a {@link Publisher} (potentially of a different type) and
+     * flatten all signals emitted from each mapped {@link Publisher} into the returned {@link Publisher}.
+     * <pre>{@code
+     *     ExecutorService e = ...;
+     *     List<Future<List<R>>> futures = ...; // assume this is thread safe
+     *     for (T t : resultOfThisPublisher()) {
+     *         // Note that flatMap process results in parallel.
+     *         futures.add(e.submit(() -> {
+     *             return mapper.apply(t); // Asynchronous result is flatten into a value by this operator.
+     *         }));
+     *     }
+     *     List<R> results = new ArrayList<>(futures.size());
+     *     // This is an approximation, this operator does not provide any ordering guarantees for the results.
+     *     for (Future<List<R>> future : futures) {
+     *         List<R> rList = future.get(); // Throws if the processing for this item failed.
+     *         results.addAll(rList);
+     *     }
+     *     return results;
+     * }</pre>
+     * @param mapper Convert each item emitted by this {@link Publisher} into another {@link Publisher}.
+     * each mapped {@link Publisher}.
+     * @param <R> The type of mapped {@link Publisher}.
+     * @return A new {@link Publisher} which flattens the emissions from all mapped {@link Publisher}s.
+     * @see <a href="http://reactivex.io/documentation/operators/flatmap.html">ReactiveX flatMap operator.</a>
+     */
+    public final <R> Publisher<R> flatMapMerge(Function<? super T, ? extends Publisher<? extends R>> mapper) {
+        return new PublisherFlatMapMerge<>(this, mapper, false, executor);
+    }
+
+    /**
+     * Each element of this {@link Publisher} is mapped into a {@link Publisher} (potentially of a different type) and
+     * flatten all signals emitted from each mapped {@link Publisher} into the returned {@link Publisher}.
+     * <pre>{@code
+     *     ExecutorService e = ...;
+     *     List<Future<List<R>>> futures = ...; // assume this is thread safe
+     *     for (T t : resultOfThisPublisher()) {
+     *         // Note that flatMap process results in parallel.
+     *         futures.add(e.submit(() -> {
+     *             return mapper.apply(t); // Asynchronous result is flatten into a value by this operator.
+     *         }));
+     *     }
+     *     List<R> results = new ArrayList<>(futures.size());
+     *     // This is an approximation, this operator does not provide any ordering guarantees for the results.
+     *     for (Future<List<R>> future : futures) {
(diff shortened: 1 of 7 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -187,6 +187,72 @@ protected Publisher() {
...
+     *         results.addAll(rList);
+     *     }
+     *     return results;
+     * }</pre>
+     * @param mapper Convert each item emitted by this {@link Publisher} into another {@link Publisher}.
+     * @param delayError {@code false} if the returned {@link Publisher} should fail as soon as any of the mapped
+     * {@link Publisher}s fail. {@code true} to wait until this {@link Publisher} terminates to emit any failures from
+     * the mapped {@link Publisher}s.
+     * @param maxConcurrency maximum amount of outstanding upstream {@link Subscription#request(long) demand}.
+     * @param maxMappedDemandHint hint for the maximum amount of {@link Subscription#request(long) demand} for
```
Scottmitch, line 245: the API here declares this as a hint, but the current implementation will respect this value.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,681 @@
...
+
+            // Currently we always request maxConcurrency elements from upstream.
+            subscription.request(source.maxConcurrency);
+        }
+
+        @Override
+        public void onNext(@Nullable final T t) {
+            final Publisher<? extends R> publisher = requireNonNull(source.mapper.apply(t));
+            FlatMapPublisherSubscriber<T, R> subscriber = new FlatMapPublisherSubscriber<>(this);
+            if (subscribers.add(subscriber)) {
```
normanmaurer, line 167: how could this ever return false ? `FlatMapPublisherSubscriber` doesn't have an equals implementation and so this will always return true imho.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,681 @@
...
+                    subscription.cancel();
+                }
+            } finally {
+                for (FlatMapPublisherSubscriber<T, R> subscriber : subscribers) {
+                    try {
+                        subscriber.cancelFromUpstream();
+                    } catch (Throwable c) {
+                        if (delayedCause == null) {
+                            delayedCause = c;
+                        }
```
normanmaurer, line 233: should we use `addSuppressed(...)` if `delayedCause != null` ?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,681 @@
...
+                if (!acquireEmittingLock()) {
+                    break;
+                }
+                Object t;
+                while ((t = pending.poll()) != null) {
+                    if (sendToTarget(t)) {
+                        ++emittedCount;
+                    }
+                }
+                if (releaseEmittingLock()) {
```
normanmaurer, line 280: wouldn't it be better to call `releaseEmittingLock()` in a finally block ? Just to ensure we can not end up in a limbo state.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,681 @@
...
+                        if (remainingRequestN > 0) {
+                            giveBackRequestN(remainingRequestN);
+                        }
+                    }
+                } catch (Throwable cause) {
+                    if (delayedCause == null) {
+                        delayedCause = cause;
+                    }
+                }
+                if (releaseTryRequestMoreLock()) {
```
normanmaurer, line 375: same comment as above... I think we want to call this method in a finally block.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,681 @@
...
+
+                if (requestNGiveBack > 0) {
+                    giveBackRequestN(requestNGiveBack);
+                    tryRequestMore();
+                }
+            }
+            return false;
+        }
+
+        public static final class FlatMapPublisherSubscriber<T, R> implements Subscriber<R> {
```
normanmaurer, line 434: does this need to be public ?

Scottmitch: no, good catch!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: +1 I'll put in a finally

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: +1 I'll put in a finally

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: throwing from methods in this scope isn't supported, so re-throwing the first cause is a compromise to provide context in the event an exception does happen but doesn't attempting to retain all exceptions that occur along the way. `addSuppressed` can also be painful for static/cached exceptions.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: it isn't expected to fail ... but if it did we would be in an inconsistent state if we attempted to modify the other state.

Scottmitch: I'll add a throw in an `else` statement to clarify.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,675 @@
...
+                boolean acquiredLock = false;
+                for (;;) {
+                    Subscription prevSubscription = subscription;
+                    assert prevSubscription != null;
+                    if (prevSubscription == CANCELLED || prevSubscription == CANCEL_PENDING) {
+                        // we have already given our undelivered requestN quota up, or will after we unroll process
+                        // onNext. we are not allowed to deliver more data or else we may violate upstream's requestN.
+                        break;
+                    } else if (prevSubscription == PROCESSING_ONNEXT ||
+                            (acquiredLock = subscriptionUpdater.compareAndSet(this,
```
Scottmitch, line 599: a key part to understanding the approach is as follows:
there is no queuing of signals for the purpose of managing excess demand [1]. instead the demand from downstream is carefully distributed to all mapped publishers. the "careful" part means that we need to be able to reclaim demand from mapped publishers when they terminated or are cancelled, otherwise we may deadlock due to not delivering data that was requested. bcz we need to reclaim demand upon cancellation we need to be sure there is no data delivered after this point (to avoid delivering too much data). So the `subscription` lock provides that coordination and makes sure we only deliver what we are allowed, and give it all back when we terminate/cancel.

[1] there is an internal queue to manage concurrency requirements of

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
normanmaurer: Fair enough... that said just because you not use `addSuppressed` here doesn't mean that a user will not do it later on ;) So in any case special care need to be taken for static / cached exceptions

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
normanmaurer, line 245: nit: in other places you always start with an `uppercase` letter while here you dont.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
normanmaurer: I still don't understand how this ever could return `false` even in an "inconsistent state".

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
normanmaurer: consider adding this as a comment in the code-base.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: as mentioned above it isn't expected to fail. is your suggestion to ignore the return value?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: agreed. I added it to the class level comments.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: I'll just change it to an assert and add a comment. let me know if you had something else in mind.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,694 @@
...
+                            throw cause;
+                        }
+                    }
+
+                    final Subscription afterLockSubscription = subscription;
+                    if (afterLockSubscription == PROCESSING_ONNEXT || afterLockSubscription == CANCELLED) {
+                        // onNext is allowed to interrupt, and will handle any pending cancel/request events
+                        // after it completes (otherwise we may drop signals).
+                        break;
+                    } else if (afterLockSubscription == CANCEL_PENDING) {
```
normanmaurer, line 553: nit: you could remove the `else`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,694 @@
...
+                        subscription = CANCELLED;
+                        cancelAndGiveBack(prevSubscription);
+                        break;
+                    } else if (afterLockSubscription == REQUEST_PENDING) {
```
normanmaurer, line 557: nit: you could remove the `else`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,694 @@
...
+                } finally {
+                    giveBackUnusedRequestN();
+                }
+            }
+
+            private boolean giveBackUnusedRequestN() {
+                // we need to give back the outstanding amount that has been requested, but not emitted.
+                // resetting sourceRequestedUpdater must happen unconditionally first, as this stops
+                // calculateSourceRequested from attempting to request any more.
+                final long prevOutstandingDemand = outstandingDemandUpdater.getAndSet(this, -1);
```
normanmaurer, line 581: nit: define a static field for `-1` to make it more clear why this magic number is used.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,694 @@
...
+                        } finally {
+                            if (acquiredLock) {
+                                for (;;) {
+                                    Subscription afterLockSubscription = subscription;
+                                    assert afterLockSubscription != PROCESSING_REQUEST;
+                                    if (afterLockSubscription == CANCEL_PENDING) {
+                                        subscription = CANCELLED; // this is a terminal state.
+                                        cancelAndGiveBack(prevSubscription);
+                                        break;
+                                    } else if (afterLockSubscription == REQUEST_PENDING) {
```
normanmaurer, line 632: nit: you could remove the `else`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: I will keep the `else` as it clarifies expectations, avoids accidental fall-through issues as code evolves, and keeps code more vertically compact.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: discussed above

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: discussed above

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: done

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
Scottmitch: done

Scottmitch:
CI failure
```
11:51:18 
Creating servicetalk-java8-prb_runtime-setup_1 ... done
time="2020-04-17T18:51:19Z" level=error msg="error waiting for container: context canceled"
11:51:19 Error response from daemon: network servicetalk-java8-prb_test-net not found
```

Scottmitch:
build this please

Scottmitch:
test this please

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,705 @@
...
+
+            // Currently we always request maxConcurrency elements from upstream.
+            subscription.request(source.maxConcurrency);
+        }
+
+        @Override
+        public void onNext(@Nullable final T t) {
+            final Publisher<? extends R> publisher = requireNonNull(source.mapper.apply(t));
+            FlatMapPublisherSubscriber<T, R> subscriber = new FlatMapPublisherSubscriber<>(this);
+            final boolean added = subscribers.add(subscriber);
```
normanmaurer, line 185: Is a `ConcurrentHashMap` good enough as the backing for the `Set` or do we also need to preserve some sort of "ordering" when iterate over all the subscribers later on and so would need some sorted of `LinkedHashSet` ?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,705 @@
...
+            // source completion from incrementing the count to 0 or positive as terminateActiveMappedSources flips
+            // the count to negative to signal to mapped sources we have completed.
+            activeMappedSources = Integer.MIN_VALUE;
+            try {
+                if (cancelSubscription) {
+                    assert subscription != null;
+                    subscription.cancel();
+                }
+            } finally {
+                for (FlatMapPublisherSubscriber<T, R> subscriber : subscribers) {
```
normanmaurer, line 247: is it possible that `cancel()` and `onNext(...)` is called concurrently ? If so just iterating over the subscribers and then call `clear()` is racy as you may miss some subscribers that were added by `onNext(...)`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: Ordering isn't required. Ideally we iterate in a random order to distribute demand uniformly, but a best effort approximation should be "good enough" as there is no guarantee about uniform distribution.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: Yes this is possible. I believe this case is accounted for as follows:

- `cancel()` sets `activeMappedSources = Integer.MIN_VALUE;` before iterating/clearing the map. See comments from `onCancel(..)`:
```
// Prevent future onNext operations from adding to subscribers which otherwise may result in
// not cancelling mapped Subscriptions. This should be Integer.MIN_VALUE to prevent subsequent mapped
// source completion from incrementing the count to 0 or positive as terminateActiveMappedSources flips
// the count to negative to signal to mapped sources we have completed.
```

- `onNext(..)` attempts to atomically increment `activeMappedSources` after adding to the map, and removes the entry if failed to increment because negative value is seen (negative values are also used when the `FlatMapSubscriber` terminates to coordinate with `FlatMapPublisherSubscriber` completion). See comments from `onNext(..)`:
```
// We have been cancelled, or already completed and the active count flipped to negative, either way
// we don't want to Subscribe or retain a reference to this Publisher.
```

Scottmitch:
ci failiure...
```
time="2020-04-28T04:36:32Z" level=error msg="error waiting for container: context canceled"
21:36:32 Error response from daemon: network servicetalk-java11-prb_test-net not found
```

Scottmitch:
@servicetalk-bot - test this please

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
```diff
@@ -47,19 +47,22 @@
      * subsequent calls will result in {@link #cancel()} being called on {@code delayedSubscription}.
      *
      * @param delayedSubscription The delayed {@link Subscription}.
+     * @return {@code true} if {@code delayedSubscription} was accepted as the delayed subscription. {@code false} if
+     * this method had already been called and {@code delayedSubscription} has been {@link #cancel() cancelled}.
      */
-    public void delayedSubscription(Subscription delayedSubscription) {
+    public boolean delayedSubscription(Subscription delayedSubscription) {
```
Scottmitch, line 53: changes extracted out to https://github.com/apple/servicetalk/pull/1046

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: note that the approach has changed and we no longer iterate over all Subscribers to distribute request n.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,547 @@
...
+        @Nullable
+        private volatile TerminalNotification terminalNotification;
+        @Nullable
+        private volatile CompositeException delayedError;
+        @SuppressWarnings("unused")
+        private volatile int emittingLock;
+        private volatile int activeMappedSources;
+        private volatile long pendingDemand;
+
+        // protected by emitting lock, or only accessed in side the Subscriber thread
```
normanmaurer, line 113: nit: s/in side/inside/

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,547 @@
...
+        private final Queue<Object> signals;
+        private final PublisherFlatMapMerge<T, R> source;
+        private final Set<FlatMapPublisherSubscriber<T, R>> subscribers;
+
+        FlatMapSubscriber(PublisherFlatMapMerge<T, R> source, Subscriber<? super R> target) {
+            this.source = source;
+            this.target = target;
+            // Start with a small capacity as maxConcurrency can be large.
+            signals = newUnboundedMpscQueue(min(2, source.maxConcurrency * source.maxMappedDemand));
+            subscribers = newSetFromMap(new ConcurrentHashMap<>());
```
normanmaurer, line 129: nit: should we use some "non default" initial capacity ?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,547 @@
...
+            final boolean added = subscribers.add(subscriber);
+            assert added; // FlatMapPublisherSubscriber relies upon object equality and should never fail.
+            for (;;) {
+                final int prevActiveSources = activeMappedSources;
+                if (prevActiveSources < 0) {
+                    // We have been cancelled, or already completed and the active count flipped to negative, either way
+                    // we don't want to Subscribe or retain a reference to this Publisher.
+                    subscribers.remove(subscriber);
+                    break;
+                } else if (activeMappedSourcesUpdater.compareAndSet(this, prevActiveSources, prevActiveSources + 1)) {
```
normanmaurer, line 176: I wonder if we could just use `incrementAndGet(...)` and decrement again if its not what we expect. This should be cheaper as this will use XADD when on x86

Something like:

```java
            for (;;) {
                 final int prevActiveSources = activeMappedSources;
                 if (prevActiveSources < 0) {
                     // We have been cancelled, or already completed and the active count flipped to negative, either way
                     // we don't want to Subscribe or retain a reference to this Publisher.
                     subscribers.remove(subscriber);
                     break;
                 // try to optimistically increment and if the outcome is not what we expect rollback
                 } else  if (activeMappedSourcesUpdater.incrementAndGet(this) ==  prevActiveSources + 1) {
                     publisher.subscribeInternal(subscriber);
                     break;
                 } else {
                     // rollback
                     activeMappedSourcesUpdater.decrementAndGet(this)
                 }
             }

File: servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java [hunk 18]
```diff
@@ -0,0 +1,547 @@
...
+                throwException(delayedCause);
+            }
+        }
+
+        private boolean tryDecrementDemand() {
+            for (;;) {
+                final long prevPendingDemand = pendingDemand;
+                if (prevPendingDemand <= 0) {
+                    return false;
+                } else if (pendingDemandUpdater.compareAndSet(this, prevPendingDemand, prevPendingDemand - 1)) {
```
normanmaurer, line 247: same comment as above... consider using `decrementAndGet(...)`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,547 @@
...
+                tryRequestMoreFromUpstream();
+                return false;
+            } else if (item instanceof TerminalNotification) {
+                // Load the terminal notification in case an error happened after an onComplete and we override the
+                // terminal value.
+                TerminalNotification terminalNotification = this.terminalNotification;
+                assert terminalNotification != null;
+                sendToTarget(terminalNotification);
+                return false;
+            } else if (item instanceof FlatMapPublisherSubscriber) {
```
normanmaurer, line 388: nit: you could remove all the `else` as in each block you return.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,547 @@
...
+            } else {
+                terminalNotification.terminate(target);
+            }
+        }
+
+        private boolean terminateActiveMappedSources() {
+            for (;;) {
+                final int prevActiveSources = activeMappedSources;
+                assert prevActiveSources >= 0; // otherwise we have seen multiple onComplete signals
+                if (activeMappedSourcesUpdater.compareAndSet(this, prevActiveSources, -prevActiveSources)) {
```
normanmaurer, line 416: same comment as above... consider using `addAndGet(...)` for XADD.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,547 @@
...
+            for (;;) {
+                final int prevActiveSources = activeMappedSources;
+                assert prevActiveSources != 0;
+                if (prevActiveSources > 0) {
+                    if (activeMappedSourcesUpdater.compareAndSet(this, prevActiveSources, prevActiveSources - 1)) {
+                        return false;
+                    }
+                } else if (activeMappedSourcesUpdater.compareAndSet(this, prevActiveSources, prevActiveSources + 1)) {
+                    return prevActiveSources == -1;
+                }
```
normanmaurer, line 432: same comment as above... consider using `incrementAndGet()` / `decrementAndGet()`

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapMergeTest.java
```diff
@@ -0,0 +1,799 @@
...
+    private static final long TERMINAL_POLL_MS = 10;
+    @Rule
+    public final Timeout timeout = new ServiceTalkTestTimeout();
+    @Nullable
+    private static ExecutorService executorService;
+    @Nullable
+    private static Executor executor;
+
+    private final TestCollectingPublisherSubscriber<Integer> subscriber = new TestCollectingPublisherSubscriber<>();
+    private TestPublisher<Integer> publisher = new TestPublisher<>();
```
normanmaurer, line 77: nit: final ?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: sure will put a cap on the initial capacity.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: negative values are meaningful as we use the negative values to signal completion to the mapped sources `terminateActiveMappedSources`, so optimistically incrementing may result in premature completion. lets investigate this optimization as a followup PR.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: I'll just keep them as it clarifies the clauses are expected to be mutually exclusive and reduces the number of lines.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapMergeTest.java
Scottmitch: it is re-assigned in some tests

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: actually we don't expect `onNext` after termination so we may be able to use this approach. However this code is using consistent approach with `PublisherFlatMapSingle` and we should keep them inline ... good candidate for followup PR!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,547 @@
...
+        private final Subscriber<? super R> target;
+        private final Queue<Object> signals;
+        private final PublisherFlatMapMerge<T, R> source;
+        private final Set<FlatMapPublisherSubscriber<T, R>> subscribers;
+
+        FlatMapSubscriber(PublisherFlatMapMerge<T, R> source, Subscriber<? super R> target) {
+            this.source = source;
+            this.target = target;
+            // Start with a small capacity as maxConcurrency can be large.
+            signals = newUnboundedMpscQueue(min(2, source.maxConcurrency * source.maxMappedDemand));
```
NiteshKant, line 128: This approach of child `Publisher`s requesting `maxMappedDemand` irrespective of downstream demand will tend to have larger buffers irrespective of how items are requested. 

An alternative approach could be to take a `minMappedDemand` and then split the downstream demand across `maxConcurrency` number of sources. 

- For `maxConcurrency == 8` and downstream demand `16`; we request `2` from each child `Publisher`.
- Each time a child `Publisher` emits as much as requested we split the remaining demand again in `maxConcurrency`. If `maxConcurrency > downstream demand` then request `minMappedDemand`.

This approach should result in smaller queues since we request depending on actual downstream demand as opposed to static `maxMappedDemand`. Additionally, we can default to lower defaults; `maxConcurrency: 8` and `minMappedDemand: 1`. The negative is that if we do not have as many items as `maxConcurrency` we end up requesting in smaller chunks.

WDYT?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -188,6 +188,156 @@ protected Publisher() {
         return new ResumePublisher<>(this, nextFactory, executor);
     }
 
+    /**
+     * Each element of this {@link Publisher} is mapped into a {@link Publisher} (potentially of a different type) and
```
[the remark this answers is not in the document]
Scottmitch: I will make them consistent. I prefer using the operator method name in the description (e.g. `flatten` and `map`) to make the method/operator naming more clear (otherwise `flatMap` can be a bit esoteric if you are not familiar with these functional concepts).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,547 @@
...
+        // protected by emitting lock, or only accessed inside the Subscriber thread
+        private boolean targetTerminated;
+        private int upstreamDemand;
+        @Nullable
+        private Subscription subscription;
+
+        private final Subscriber<? super R> target;
+        private final Queue<Object> signals;
+        private final PublisherFlatMapMerge<T, R> source;
+        private final Set<FlatMapPublisherSubscriber<T, R>> subscribers;
```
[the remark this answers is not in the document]
Scottmitch: Duplicate `Subscribers` are not supported, mapped `Subscribers` are removed when they terminate (random order), and the cardinality for the general implementation is not strictly limited. I had a specialized collection for smaller concurrency cardinalities but decided to back this out for now for simplicity.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,547 @@
...
+                // pendingOnNext must be updated before tryEmitItem for visibility of signalsQueued. signalsQueued
+                // is updated in replenishDemandAndClearSignalsQueued (potentially from another thread) before
+                // pendingOnNext is reset and we want signalsQueued to be visible to this thread.
+                //
+                // Heuristic which triggers replenishDemandAndClearSignalsQueued when demand is consumed. It must be
+                // when all demand is consumed in order for clearSignalsQueued to work properly.
+                final boolean doReplenish = pendingOnNextUpdater.incrementAndGet(this) == parent.source.maxMappedDemand;
+                parent.tryEmitItem(wrapNull(r), this);
+                if (doReplenish) {
+                    parent.tryEmitItem(this, this);
```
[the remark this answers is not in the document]
Scottmitch: This is done to invoke `replenishDemandAndClearSignalsQueued` after the exhausted demand has been processed to avoid requesting more than max concurrency (the state for demand is maintained in each `Subscriber`). Can you clarify the source of confusion and do you have a suggestion to improve?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,547 @@
...
+    private static final class FlatMapSubscriber<T, R> implements Subscriber<T>, Subscription {
+        private static final Object MAPPED_SOURCE_COMPLETE = new Object();
+        @SuppressWarnings("rawtypes")
+        private static final AtomicReferenceFieldUpdater<FlatMapSubscriber, CompositeException> delayedErrorUpdater =
+                newUpdater(FlatMapSubscriber.class, CompositeException.class, "delayedError");
+        @SuppressWarnings("rawtypes")
+        private static final AtomicIntegerFieldUpdater<FlatMapSubscriber> emittingLockUpdater =
+                AtomicIntegerFieldUpdater.newUpdater(FlatMapSubscriber.class, "emittingLock");
+        @SuppressWarnings("rawtypes")
+        private static final AtomicLongFieldUpdater<FlatMapSubscriber> pendingDemandUpdater =
```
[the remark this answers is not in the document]
Scottmitch: This operator is different from `flatMapMergeSingle` in that the demand isn't tightly coupled between mapped async sources and demand from downstream. The operator's demand on async sources is managed independently from downstream demand and limited by `maxMappedDemand`, at the cost of queuing internally. A previous implementation we discussed offline was distributing downstream demand to mapped sources (to avoid the queuing) but this requires a heuristic to distribute demand evenly and may lead to "dead locks" if some mapped sources don't deliver signals in a timely manner.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,547 @@
...
+                }
+            }
+
+            if (delayedCause != null) {
+                throwException(delayedCause);
+            }
+        }
+
+        private void tryRequestMoreFromUpstream() {
+            if (--upstreamDemand == 0) { // heuristic to replenish demand when it is exhausted.
```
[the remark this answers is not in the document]
Scottmitch: Good question. This is not intentional and the motivation for this heuristic was to batch demand to avoid `request(1)` when each mapped source completes. I'll switch to batching demand within the draining of the signal delivery queue. Unless there is concurrency on completion this will digress to `request(1)` but we can enhance later with something more sophisticated batching (e.g. configurable heuristic).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: discussed offline and adopted this approach with trade-offs documented in javadocs

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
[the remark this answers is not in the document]
Scottmitch: the approach has changed due to offline discussions. ptal.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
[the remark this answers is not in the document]
Scottmitch: the approach has changed due to offline discussions. ptal

Scottmitch:
@servicetalk-bot - test this please

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,661 @@
...
+                } else if (pendingDemandUpdater.compareAndSet(this, prevDemand, prevDemand - 1)) {
+                    return true;
+                }
+            }
+        }
+
+        private void tryEmitItem(Object item, FlatMapPublisherSubscriber<T, R> subscriber) {
+            if (tryAcquireLock(emittingLockUpdater, this)) { // fast path. no concurrency, try to skip the queue.
+                try {
+                    if (subscriber.hasSignalsQueued() || (needsDemand(item) && !tryDecrementPendingDemand())) {
```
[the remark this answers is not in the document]
Scottmitch: No, `signalsQueued` is to avoid out of order delivery of signals relative to each mapped publisher. Either we go through the queue for all signals and ordering is provided by the queue, or if we may skip the queue (like we are doing now) we need to check if there maybe items already queued (in which case we need to go through the queue to preserve ordering). The `signalsQueued` flag is currently cleared after demand is exhausted as we anyways need a reference to the `FlatMapPublisherSubscriber` to request more demand. We could potentially clear this flag earlier but we would need a reference to the `FlatMapPublisherSubscriber` while processing `onNext` which would require object allocation.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,661 @@
...
+                throwException(delayedCause);
+            }
+        }
+
+        private int calculateRequestNQuota(long availableRequestN) {
+            // Get an approximate quota to distribute to each active mapped subscriber. If
+            // activeMappedSources changes (subscriber added or terminated) we will re-distribute.
+            final int prevActiveSources = activeMappedSources;
+            return (int) min(Integer.MAX_VALUE,
+                    max(prevActiveSources > 0 ? availableRequestN / prevActiveSources : availableRequestN,
```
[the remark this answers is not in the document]
Scottmitch: the `mappedDemand` accounting is not quite right and needs some fixing. the intention was to at least distribute `minMappedDemand` to each source so that each source is allowed to produce data while capping the queue size at `minMappedDemand * maxConcurrency`. let me re-work this a bit.

using `availableRequestN` was an optimization geared toward steady state to get a more accurate estimate which should result in larger demand chunks being distributed and less re-distribution events. It would be unfair during non-steady state as you indicated. This optimization is not critical and I will remove it for now.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,661 @@
...
+                        pendingDemandUpdater.accumulateAndGet(this, prevDemand - emittedCount,
+                                FlowControlUtils::addWithOverflowProtectionIfNotNegative);
+                    }
+                } finally {
+                    tryAcquire = !releaseLock(emittingLockUpdater, this);
+                }
+            }
+
+            if (mappedSourcesCompleted != 0) {
+                requestMoreFromUpstream(mappedSourcesCompleted);
```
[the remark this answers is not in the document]
Scottmitch: > The subscription is a ConcurrentSubscription which means we can concurrently request more items. Can we go to a simpler model of not going through the queue ...

`ConcurrentSubscription` is responsible for concurrency control, and in this case the call isn't inside a critical section. I will move the other call to `requestMoreFromUpstream` outside the critical block to make it consistent.

Requesting upstream isn't strictly tied to "going through the queue" but focused on waiting until events are delivered downstream (we skip the queue if no concurrency) to keep the queue bounded (we should be requesting `minMappedDemand` when a new mapped source is delivered).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,661 @@
...
+        private void drainPending() {
+            Throwable delayedCause = null;
+            boolean tryAcquire = true;
+            int mappedSourcesCompleted = 0;
+            while (tryAcquire && tryAcquireLock(emittingLockUpdater, this)) {
+                try {
+                    final long prevDemand = pendingDemandUpdater.getAndSet(this, 0);
+                    long emittedCount = 0;
+                    if (prevDemand < 0) {
+                        // if we have been cancelled then we avoid delivering any signals, but we still deliver error
```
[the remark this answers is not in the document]
Scottmitch: this code isn't quite correct, thx for bringing my attention to this. the comment is meant to reference a cancel that is generated internally (e.g. if delayError=false and a mapped source fails, or failure to enqueue. The goal is to be consistent with `PublisherFlatMapSingle` however we have to manage demand in this case which makes it a bit more involved (which is where the bug is, currently discarding signals and delivering terminal out of order).

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/pull/1116

NiteshKant:
@Scottmitch is this ready for another pass?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,624 @@
...
+            Throwable delayedCause = null;
+            boolean tryAcquire = true;
+            int mappedSourcesCompleted = 0;
+            while (tryAcquire && tryAcquireLock(emittingLockUpdater, this)) {
+                try {
+                    final long prevDemand = pendingDemandUpdater.getAndSet(this, 0);
+                    long emittedCount = 0;
+                    if (prevDemand < 0) {
+                        pendingDemand = prevDemand;
+                        // Don't wait for demand to deliver the terminalNotification if present. The queued signals
```
NiteshKant, line 365: I don't think this is required, i.e. have dual mode whether errors skip the queue or not based on `pendingDemand` being negative.

The context is spread out to call-sites of `doCancel()` and makes the control flow too hard to follow.

NiteshKant (commented):
Few comments then LGTM

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,688 @@
...
+                AtomicLongFieldUpdater.newUpdater(FlatMapSubscriber.class, "pendingDemand");
+        @SuppressWarnings("rawtypes")
+        private static final AtomicIntegerFieldUpdater<FlatMapSubscriber> activeMappedSourcesUpdater =
+                AtomicIntegerFieldUpdater.newUpdater(FlatMapSubscriber.class, "activeMappedSources");
+        @SuppressWarnings("rawtypes")
+        private static final AtomicReferenceFieldUpdater<FlatMapSubscriber, TerminalNotification>
+                terminalNotificationUpdater = newUpdater(FlatMapSubscriber.class, TerminalNotification.class,
+                "terminalNotification");
+
+        @SuppressWarnings("unused")
```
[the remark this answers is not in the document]
Scottmitch: I will update to `UnusedDeclaration` since they are not directly assigned we get a different warning if the suppression is completely removed.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,624 @@
...
+                    }
+                }
+            }
+
+            @Override
+            public void onComplete() {
+                if (parent.removeSubscriber(this, pendingDemandUpdater.getAndSet(this, -1))) {
+                    parent.enqueueAndDrain(complete());
+                } else {
+                    parent.tryEmitItem(MAPPED_SOURCE_COMPLETE, this);
```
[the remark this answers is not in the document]
Scottmitch: `MAPPED_SOURCE_COMPLETE` going through the queue isn't for concurrency control on the `Subscription` but instead to bound how much data we queue. In order to avoid deadlock scenarios we always give some demand to mapped sources which means queued data should be expected in low downstream demand scenarios. Due to this we simplify demand management and request more from the mapped source in its subscriber's `onSubscribe` method (as opposed to trying to avoid distributing demand when downstream demand is `0`). If we don't go through the queue here we will have to adjust when we requestN flow. I'll add a unit test to demonstrate the behavior and we can consider changes to control flow in a followup PR.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -0,0 +1,624 @@
...
+            target.onSubscribe(this);
+
+            subscription.request(source.maxConcurrency);
+        }
+
+        @Override
+        public void onNext(@Nullable final T t) {
+            final Publisher<? extends R> publisher = requireNonNull(source.mapper.apply(t));
+            FlatMapPublisherSubscriber<T, R> subscriber = new FlatMapPublisherSubscriber<>(this);
+            if (!cancellableSubscribers.add(subscriber)) {
```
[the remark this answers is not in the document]
Scottmitch: Good catch! https://github.com/apple/servicetalk/pull/1159

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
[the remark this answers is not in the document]
Scottmitch: discussed offline, fine as is.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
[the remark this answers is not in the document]
Scottmitch: the original question is a duplicate of https://github.com/apple/servicetalk/pull/1011#discussion_r496979928 and discussed/resolved below. lets open a new thread if there are still questions on this topic.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: discussed details offline and clarified the failure scenarios we are accommodating for with this code. fine as is.

Scottmitch:
thanks for review @NiteshKant @normanmaurer !

PR #1012, opened 2020-04-14
Publisher#flatMapMergeSingle terminal condition race

Scottmitch:
Motivation:
Publisher#flatMapMergeSingle has the following race condition:
T1 - FlatMapSubscriber#onComplete (the outer Subscriber)
T2 - FlatMapSingleSubscriber#onComplete (the inner Subscriber)
T1 - terminal.cas(onComplete()) -> true, active == 0 -> false
T2 - active.dec() -> 0, terminal == null -> true
There is no guarantee that T2 will see the terminal.cas() from T1

Publisher#flatMapMerge was recently introduced, shares some control flow
logic with Publisher#flatMapMergeSingle that is now inconsistent.

Modifications:
- use atomic operations on the active count to flip to negative when the
outer source completes to ensure visibility on termination between
differen threads.

Result:
More correct termination of Publisher#flatMapMergeSingle

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
@@ -179,7 +179,8 @@ final class PublisherFlatMapSingle<T, R> extends AbstractAsynchronousPublisherOp
             // If we are cancelled the activeUpdater count is best effort depending upon which sources finish. This best
             // effort behavior mimics the semantics of cancel though so we don't take any special action to try to
             // adjust the count or prematurely terminate.
-            activeMappedSourcesUpdater.incrementAndGet(this);
+            final int activeMappedSources = activeMappedSourcesUpdater.incrementAndGet(this);
+            assert activeMappedSources > 0; // otherwise onComplete was previously invoked, which is invalid.
             next.subscribeInternal(new FlatMapSingleSubscriber());
         }
 
@@ -203,10 +204,8 @@ final class PublisherFlatMapSingle<T, R> extends AbstractAsynchronousPublisherOp
         private boolean terminateActiveMappedSources() {
             for (;;) {
                 final int prevActiveMappedSources = activeMappedSources;
-                // Should always be >= 0, but just in case there is a bug in user code that results in multiple terminal
-                // events we avoid corrupting our internal state.
-                if (prevActiveMappedSources < 0 || activeMappedSourcesUpdater.compareAndSet(this,
-                        prevActiveMappedSources, -prevActiveMappedSources)) {
+                assert prevActiveMappedSources >= 0; // otherwise we have seen multiple onComplete signals
+                if (activeMappedSourcesUpdater.compareAndSet(this, prevActiveMappedSources, -prevActiveMappedSources)) {
                     return prevActiveMappedSources == 0;
                 }
             }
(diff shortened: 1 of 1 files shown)


Scottmitch:
this is a similar approach as used in https://github.com/apple/servicetalk/pull/1011 ... we may be able to share code after one is merged.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
```diff
@@ -193,16 +192,42 @@ public void onError(Throwable t) {
...
+                if (prevActiveMappedSources < 0 || activeMappedSourcesUpdater.compareAndSet(this,
+                        prevActiveMappedSources, -prevActiveMappedSources)) {
+                    return prevActiveMappedSources == 0;
+                }
+            }
+        }
+
+        private boolean decrementActiveMappedSources() {
+            for (;;) {
+                final int prevActivePublishers = activeMappedSources;
```
NiteshKant, line 217: `prevActivePublishers` -> `prevActiveSingles`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
```diff
@@ -193,16 +192,42 @@ public void onError(Throwable t) {
...
+            final boolean allSourcesTerminated = terminateActiveMappedSources();
+            if (setTerminal && allSourcesTerminated) {
                 enqueueAndDrain(complete());
             }
         }
 
+        private boolean terminateActiveMappedSources() {
+            for (;;) {
+                final int prevActiveMappedSources = activeMappedSources;
+                // Should always be >= 0, but just in case there is a bug in user code that results in multiple terminal
```
NiteshKant, line 206: Is this comment saying that `prevActiveMappedSources` should always be >= 0? We are setting it to `-prevActiveMappedSources`, so this is confusing. Maybe just elaborate more like:
```
// prevActiveMappedSources >= 0 if only a single terminal event is received (expected)
// prevActiveMappedSources < 0 if multiple terminal events are received (unexpected)
// We return true only for the first terminal event.

File: servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java [hunk 3]
```diff
@@ -180,7 +179,7 @@ public void onNext(T t) {
             // If we are cancelled the activeUpdater count is best effort depending upon which sources finish. This best
             // effort behavior mimics the semantics of cancel though so we don't take any special action to try to
             // adjust the count or prematurely terminate.
-            activeUpdater.incrementAndGet(this);
+            activeMappedSourcesUpdater.incrementAndGet(this);
```
NiteshKant, line 182: Since we are setting a negative value to this variable; should we instead use:
```java
activeMappedSourcesUpdater.accumulateAndGet(this, 1, (prev, next) -> prev >= 0 ? prev + next : prev);
```

Scottmitch: good question. `activeMappedSources` is set to a negative value in this Subscriber's `onComplete` method, so `onNext` should not see a negative value and I'll add an assert to clarify expectations.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
Scottmitch: good point. I'll simplify to an asset and clarify the comment.

PR #1014, opened 2020-04-14
Publisher#flatMapMergeSingle potential queue drain visibility issue

Scottmitch:
Motivation:
Publisher#flatMapMergeSingle uses the
ConcurrentUtils#drainSingleConsumerQueue to drain the pending signal
queue. However this method relies upon the queue's isEmpty() method
providing visibility as to if an item has been inserted into the queue
since the last poll information. We should avoid leveraging the queue's
visibility guarantees (which may not hold) and instead leverage the
concurrency avoidance for single consumer drain to provide visibility.

Modifications:
- Remove ConcurrentUtils#drainSingleConsumerQueue
- Introduce ConcurrentUtils#acquirePendingLock and releasePendingLock
methods that allow for notification if another thread attempted to
acquire the lock while it was being held. If all threads that publish
and event (e.g. insert into a queue) use this lock it can ensure no
concurrency and visibility.

Result:
No dependency on queue visibility semantics for draining the signal
queue in Publisher#flatMapMergeSingle.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
@@ -33,13 +33,16 @@ import java.util.function.Function;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.api.SubscriberApiUtils.NULL_TOKEN;
-import static io.servicetalk.concurrent.internal.ConcurrentUtils.drainSingleConsumerQueue;
+import static io.servicetalk.concurrent.internal.ConcurrentUtils.releaseLock;
+import static io.servicetalk.concurrent.internal.ConcurrentUtils.tryAcquireLock;
 import static io.servicetalk.concurrent.internal.SubscriberUtils.calculateSourceRequested;
 import static io.servicetalk.concurrent.internal.SubscriberUtils.checkDuplicateSubscription;
 import static io.servicetalk.concurrent.internal.SubscriberUtils.isRequestNValid;
 import static io.servicetalk.concurrent.internal.SubscriberUtils.trySetTerminal;
 import static io.servicetalk.concurrent.internal.TerminalNotification.complete;
+import static io.servicetalk.concurrent.internal.ThrowableUtils.catchUnexpected;
 import static io.servicetalk.utils.internal.PlatformDependent.newUnboundedMpscQueue;
+import static io.servicetalk.utils.internal.PlatformDependent.throwException;
 import static java.lang.Math.min;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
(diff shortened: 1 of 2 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
```diff
@@ -222,23 +223,31 @@ private void enqueueAndDrain(Object item) {
...
+            for (;;) {
+                if (!acquirePendingLock(emittingUpdater, this)) {
+                    break;
+                }
+                Object t;
+                while ((t = pending.poll()) != null) {
+                    ++drainCount;
+                    sendToTarget(t);
+                }
+                if (releasePendingLock(emittingUpdater, this)) {
```
normanmaurer, line 242: shouldn't we release in a finally block ?

Scottmitch: good point ... let me make this more robust.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
```diff
@@ -222,29 +224,58 @@ private void enqueueAndDrain(Object item) {
...
+            while (acquirePendingLock(emittingUpdater, this)) {
+                try {
+                    Object t;
+                    while ((t = pending.poll()) != null) {
+                        ++drainCount;
+                        try {
+                            sendToTarget(t);
+                        } catch (Throwable cause) {
+                            if (delayedCause == null) {
+                                delayedCause = cause;
```
normanmaurer, line 244: nit: should we use `addSuppressed(...)` when `delayedCause != null`

Scottmitch: lets discuss on https://github.com/apple/servicetalk/pull/1011#discussion_r408440046

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/ConcurrentUtils.java
```diff
@@ -31,47 +31,44 @@
 
     public static final int CONCURRENT_IDLE = 0;
     public static final int CONCURRENT_EMITTING = 1;
+    private static final int CONCURRENT_PENDING = 2;
```
normanmaurer, line 34: Just wondering why the others need to be public and this can be private ...

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
```diff
@@ -222,29 +231,62 @@ private void enqueueAndDrain(Object item) {
...
                     subscription.request(actualSourceRequestN);
                 }
             }
+
+            if (delayedCause != null) {
+                throwException(delayedCause);
+            }
+        }
+
+        private void enqueueAndDrainFail(Object item) {
```
normanmaurer, line 282: nit: `...DrainFailed` ?

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/ConcurrentUtils.java
Scottmitch: its a good question. I would like to investigate making them all private and instead expose lock methods, but didn't want to expand the scope of this PR: https://github.com/apple/servicetalk/issues/1023

normanmaurer: sounds good...

PR #1018, opened 2020-04-15
DNS SRV query support

Scottmitch:
Motivation:
Our DNS client currently doesn't support
[SRV Resource Records](https://tools.ietf.org/html/rfc2782) which is a
commonly used type of resolution.

Modifications:
- Rename DnsServiceDiscoverer to DnsClient internally to clarify the
feature set is inteded to cover dns client use cases.
- Remove the ServiceDiscovererFilter related classes as the only
implementations are specific to DNS, can be replicated with simple
Publisher#retryWhen operator, and the general utility of these broad
APIs across different clients/protocols is yet to be demonstrated.
- Remove tracking of all outstanding streams and close() cleanup to
reduce state. Subscribers of these sources (e.g. LoadBalancers) are
responsible for cancelling if they are no longer interested in demand.
There is best effort close cleanup when TTLs timeout fire in the event
the DnsClient is closed but the streams are not cancelled (which is not
an expected use case).
- Introduce DNS SRV record support into DefaultDnsClient. Initial
support is limited to a single A/AAAA record resolution per SRV record
due to operator availability (e.g. flatMapMergeSingle) which is
sufficient to get started and can be enhanced later.
- Expose static factory methods for SRV based lookups on HttpClients and
GrpcClients

Result:
DNS supports SRV resolution.


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,694 @@
...
+            this.serviceName = serviceName;
+        }
+
+        @Override
+        protected AbstractEntriesPublisherSubscription newSubscription(
+                final Subscriber<? super Iterable<ServiceDiscovererEvent<HostAndPort>>> subscriber) {
+            return new AbstractEntriesPublisherSubscription(subscriber) {
+                @Override
+                protected Future<DnsAnswer<HostAndPort>> doDnsQuery() {
+                    Promise<DnsAnswer<HostAndPort>> promise = ImmediateEventExecutor.INSTANCE.newPromise();
```
NiteshKant, line 255: This can just be a follow up but it looks like we have all the context here to actually do the A lookup when the SRV lookup finishes in netty itself. This will simplify the control flow by eliminating the `flatMap` on the SRV publisher and the whole mechanics of representing the DNS A record lookup as an infinite `Publisher`.

I am still not convinced that we need to keep listening to the A lookup on the SRV record result. It feels surprising to have both SRV record and a changing A record for each SRV record result.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -186,14 +184,12 @@ public DefaultDnsServiceDiscovererBuilder noRetriesOnDnsFailures() {
...
-     * @param factory {@link ServiceDiscovererFilterFactory} to decorate a {@link ServiceDiscoverer} for the purpose of
+     * @param factory {@link DnsClientFilterFactory} to decorate a {@link DnsClient} for the purpose of
      * filtering.
      * @return {@code this}
      */
-    public DefaultDnsServiceDiscovererBuilder appendFilter(
-            final ServiceDiscovererFilterFactory<String, InetAddress, ServiceDiscovererEvent<InetAddress>>
-                    factory) {
-        serviceDiscoveryFilterFactory = serviceDiscoveryFilterFactory.append(factory);
+    DefaultDnsServiceDiscovererBuilder appendFilter(final DnsClientFilterFactory factory) {
```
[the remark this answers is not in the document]
Scottmitch: this method is also package private (currently used in tests)

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -209,75 +205,49 @@ public DefaultDnsServiceDiscovererBuilder ioExecutor(final IoExecutor ioExecutor
...
-            }
-        };
+    DnsClient build() {
+        DnsClient rawClient = new DefaultDnsClient(
+                ioExecutor == null ? globalExecutionContext().ioExecutor() : ioExecutor, minTTLSeconds, ndots,
+                invalidateHostsOnDnsFailure, optResourceEnabled, queryTimeout, dnsResolverAddressTypes,
+                dnsServerAddressStreamProvider);
+        DnsClientFilterFactory rawFilterFactory = filterFactory;
+        if (applyRetryFilter) {
+            DnsClientFilterFactory retryFilterFactory = client -> new RetryingDnsClientFilter(client,
```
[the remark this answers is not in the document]
Scottmitch: done

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/GlobalDnsClient.java
```diff
@@ -32,35 +32,55 @@
  * A lazily initialized singleton DNS {@link ServiceDiscoverer} using a default {@link ExecutionContext}, the lifecycle
  * of this instance shouldn't need to be managed by the user. Don't attempt to close the {@link ServiceDiscoverer}.
  */
-final class GlobalDnsServiceDiscoverer {
+final class GlobalDnsClient {
```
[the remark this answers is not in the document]
Scottmitch: This is a package private API, but I can change back since the client is no longer visible here.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,694 @@
...
+                    @Override
+                    protected void onFinallyCleanup() {
+                        removeSrvEntry0(this);
+                    }
+                };
+                addSrvEntry0(entry);
+                return entry.flatMapConcatIterable(identity()).map(inetEvent -> new DefaultServiceDiscovererEvent<>(
+                        new InetSocketAddress(inetEvent.address(), srvEvent.address().port()), inetEvent.isAvailable())
+                       ).firstOrElse(() -> {
+                           throw new IllegalStateException("0 SRV records found. 1 SRV record is required!");
```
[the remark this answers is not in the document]
Scottmitch: I agree we actually don't need to error out here and can just return `null`, which we filter later.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,694 @@
...
+                            activeAddresses = addresses;
+                            --pendingRequests;
+                            if (pendingRequests > 0) {
+                                scheduleQuery0(ttlNanos);
+                            } else {
+                                resolveDoneNoScheduleTime = nanoTime();
+                                cancellableForQuery = null;
+                            }
+                            try {
+                                subscriber.onNext(events);
```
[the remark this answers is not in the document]
Scottmitch: they got lost in the shuffle. let me add some back.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,694 @@
...
+                LOGGER.warn("Exception from subscriber {} while handling error in DNS subscription {}",
+                        subscriber, this, cause);
+            }
+
+            @SuppressWarnings("ForLoopReplaceableByForEach")
+            private void clearAddressesAndPropagateRemovalEvents() {
+                if (!activeAddresses.isEmpty()) {
+                    // cleanup code may invoke this method after pendingRequests is set to negative
+                    assert pendingRequests != 0;
+                    --pendingRequests;
```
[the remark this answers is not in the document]
Scottmitch: the predicate in `handleResolveDone0` doesn't allow this method to be called after `onNext` ... however I agree the control flow is convoluted here let me fix this up

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,694 @@
...
+            private void cancelWithoutRemove0(boolean tryGenerateInactiveEvents) {
+                if (cancellableForQuery != null) {
+                    cancellableForQuery = TERMINATED;
+                    discoverySubscriber = null;
+                    pendingRequests = -1;
+                    cancellableForQuery.cancel();
+                }
+                if (tryGenerateInactiveEvents) {
+                    if (activeAddresses.isEmpty() || pendingRequests > 0) {
+                        terminateOnComplete();
```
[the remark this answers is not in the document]
Scottmitch: the close operation on the publisher is only used when the SRV resolution comes back as "unavailable" we then lookup the corresponding ARecordPublisher and close it. In this case we want the publisher to complete normally so in the case of `Publisher#flatMapMerge` it shouldn't take down the entire stream.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
Scottmitch: lets discuss as followup.

current goals are:
- re-use existing control flow/logic
- represent the DNS control flow as defined in the DNS SRV RFC [1]

[1] https://tools.ietf.org/html/rfc2782
> Target
        The domain name of the target host.  There MUST be one **or more**
        address records for this name,

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,694 @@
...
+    @Override
+    public Publisher<ServiceDiscovererEvent<InetAddress>> dnsQuery(final String address) {
+        return new ARecordPublisher(address).flatMapConcatIterable(identity());
+    }
+
+    @Override
+    public Publisher<ServiceDiscovererEvent<InetSocketAddress>> dnsSrvQuery(final String serviceName) {
+        return new SrvRecordPublisher(serviceName).flatMapConcatIterable(identity())
+                .<ServiceDiscovererEvent<InetSocketAddress>>flatMapMergeSingle(srvEvent -> {
+            assertInEventloop();
```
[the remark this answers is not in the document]
Scottmitch: This is how the IDE lined it up. what would you suggest as an alternative? lets adjust in a followup PR if necessary or consider adjusting the checkstyle rules if there are specific formatting needs for this case.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
NiteshKant: >    The domain name of the target host.  There MUST be one **or more**

To be clear; I am suggesting to do a one time lookup in netty for each SRV record entry which will give all A* records. 
The difference is that with the `ARecordPublisher` we are also listening to changes on that A record over time which seems unnecessary in this case.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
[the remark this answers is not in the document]
NiteshKant: +1 for not holding off the PR; I just reformatted using IDE.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,721 @@
...
+    public Completable closeAsync() {
+        return asyncCloseable.closeAsync();
+    }
+
+    @Override
+    public Completable closeAsyncGracefully() {
+        return asyncCloseable.closeAsyncGracefully();
+    }
+
+    private void closeAsync0() {
```
NiteshKant, line 225: So previously[1] we were terminating all active queries in `close()`. I don't see a test for the close behavior in the current `master` but looks like this something that is missing from this change?

[1] https://github.com/apple/servicetalk/blob/master/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscoverer.java#L209-L223

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
```diff
@@ -123,10 +121,112 @@ public void unknownHostDiscover() throws Exception {
...
+        toSource(publisher).subscribe(subscriber);
+
+        latch.await();
+        assertNull(throwableRef.get());
+        assertThat(subscriber.activeCount(), equalTo(expectedActiveCount));
+        assertThat(subscriber.inactiveCount(), equalTo(expectedInactiveCount));
+    }
+
+    @Test
+    public void multipleSrvSingleADiscover() throws InterruptedException {
```
NiteshKant, line 176: This can be done as a followup if you have verified that changes to SRV records are reflected in the lookup, but it will be good to have a unit test for it.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -209,75 +202,47 @@ public DefaultDnsServiceDiscovererBuilder ioExecutor(final IoExecutor ioExecutor
...
+    DnsClient build() {
+        DnsClient rawClient = new DefaultDnsClient(
+                ioExecutor == null ? globalExecutionContext().ioExecutor() : ioExecutor, minTTLSeconds, ndots,
+                invalidateHostsOnDnsFailure, optResourceEnabled, queryTimeout, dnsResolverAddressTypes,
+                dnsServerAddressStreamProvider);
+        DnsClientFilterFactory rawFilterFactory = filterFactory;
+        if (applyRetryFilter) {
+            DnsClientFilterFactory retryFilterFactory = new RetryingDnsClientFilter();
+            rawFilterFactory = rawFilterFactory == null ? retryFilterFactory :
+                    retryFilterFactory.append(rawFilterFactory);
```
idelpivnitskiy, line 244: Because we do not expose `DnsClient` as a public API, can we keep existing `ServiceDiscovererFilter` and `ServiceDiscovererFilterFactory` and apply them in `buildSrvDiscoverer` and `buildARecordDiscoverer` methods respectively?
Then we do not need to change the pre-existing retrying filter as we can apply it on top of the returned `ServiceDiscoverer` interface.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
Scottmitch: The only constraint from the SRV RFC is target can't be an alias, other than that it is a normal A* resolution (which has its own TTL and may change over time). If the DNS server provides the A* resolution in the `Additional` section of the SRV answer we may be able to avoid the secondary lookup, but depends upon the DNS server behavior (we can explore this optimization later).

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
Scottmitch: > Remove tracking of all outstanding streams and close() cleanup to
reduce state. Subscribers of these sources (e.g. LoadBalancers) are
responsible for cancelling if they are no longer interested in demand.
There is best effort close cleanup when TTLs timeout fire in the event
the DnsClient is closed but the streams are not cancelled (which is not
an expected use case).

This was removed as it isn't guaranteed by the interface and subscribers are currently responsible for managing lifetime/cancellation. Removing this constraint simplified the code but we can add it back if necessary.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
Scottmitch: discussed offline. lets hold off on the `ServiceDiscovererFilterFactory` general APIs for now and re-evaluate after we get more experience with DNS and other implementations if the general APIs are worth it.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
Scottmitch: good call, I added a test. The test class was introduced before we had testing utilities and is difficult to control/analyze. Here is an issue to analyze https://github.com/apple/servicetalk/issues/1022.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,715 @@
+/*
+ * Copyright © 2018 Apple Inc. and the ServiceTalk project authors
```
normanmaurer, line 2: nit: 2020 ?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,715 @@
...
+                                        final List<HostAndPort> hostAndPorts = new ArrayList<>(dnsRecords.size());
+                                        for (DnsRecord dnsRecord : dnsRecords) {
+                                            if (!SRV.equals(dnsRecord.type()) || !(dnsRecord instanceof DnsRawRecord)) {
+                                                throw new IllegalArgumentException(
+                                                        "Unsupported DNS record type for SRV query: " + dnsRecord);
+                                            }
+                                            if (dnsRecord.timeToLive() < minTTLSeconds) {
+                                                minTTLSeconds = dnsRecord.timeToLive();
+                                            }
+                                            ByteBuf content = ((DnsRawRecord) dnsRecord).content();
```
normanmaurer, line 269: @Scottmitch I think you also need to call `ReferenceCounted.release(dnsRecord)`here.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,715 @@
...
+                                } else {
+                                    final DnsAnswer<HostAndPort> dnsAnswer;
+                                    long minTTLSeconds = Long.MAX_VALUE;
+                                    try {
+                                        @SuppressWarnings("unchecked")
+                                        final List<DnsRecord> dnsRecords = (List<DnsRecord>) completedFuture.getNow();
+                                        final List<HostAndPort> hostAndPorts = new ArrayList<>(dnsRecords.size());
+                                        for (DnsRecord dnsRecord : dnsRecords) {
+                                            if (!SRV.equals(dnsRecord.type()) || !(dnsRecord instanceof DnsRawRecord)) {
+                                                throw new IllegalArgumentException(
```
normanmaurer, line 263: ensure to release all records that were not released before throwing here.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,715 @@
...
+                                            ByteBuf content = ((DnsRawRecord) dnsRecord).content();
+                                            // https://tools.ietf.org/html/rfc2782
+                                            // Priority(16u) Weight(16u) Port(16u) Target(varstring)
+                                            content.skipBytes(4); // ignore Priority and Weight for now
+                                            final int port = content.readUnsignedShort();
+                                            hostAndPorts.add(HostAndPort.of(decodeName(content), port));
+                                        }
+                                        dnsAnswer = new DnsAnswer<>(hostAndPorts, SECONDS.toNanos(minTTLSeconds));
+                                    } catch (Throwable cause2) {
+                                        promise.setFailure(cause2);
```
normanmaurer, line 278: ensure you release all `records`.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,715 @@
...
+                cancellableForQuery = nettyIoExecutor.asExecutor().schedule(
+                        this::doQuery0, nanos, NANOSECONDS);
+            }
+
+            private void handleResolveDone0(final Future<DnsAnswer<T>> addressFuture) {
+                assertInEventloop();
+                assert pendingRequests > 0;
+                if (cancellableForQuery == TERMINATED) {
+                    return;
+                } else if (cancellableForQuery == TERMINATE_ON_NEXT_REQUEST_N) {
```
normanmaurer, line 569: nit: you could remove the `else` here

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -0,0 +1,715 @@
...
+                LOGGER.warn("Exception from subscriber {} while handling error in DNS subscription {}",
+                        subscriber, this, cause);
+            }
+
+            private boolean clearAddressesAndPropagateRemovalEvents() {
+                assertInEventloop();
+
+                if (activeAddresses.isEmpty()) {
+                    return true;
+                } else if (pendingRequests > 0) {
```
normanmaurer, line 639: nit: you could remove the `else` here

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
idelpivnitskiy: Fixed in https://github.com/apple/servicetalk/pull/1044

PR #1027, opened 2020-04-18
Add `Publisher#buffer` operator

NiteshKant:
__Motivation__

[`buffer`](http://reactivex.io/documentation/operators/buffer.html) operator is useful for batching.

__Modification__

Add `buffer()` operator for `Publisher` to buffer based on arbitrary buffer boundaries or based on size + time duration.

__Result__

Users can buffer events from a `Publisher`.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BufferStrategies.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BufferStrategies.java
@@ -1,0 +1,298 @@
+/*
+ * Copyright © 2020 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.concurrent.api;
+
+import io.servicetalk.concurrent.CompletableSource.Subscriber;
+import io.servicetalk.concurrent.api.BufferStrategy.Accumulator;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
+
+import static io.servicetalk.concurrent.Cancellable.IGNORE_CANCEL;
+import static io.servicetalk.concurrent.api.ImmediateExecutor.IMMEDIATE_EXECUTOR;
+import static io.servicetalk.concurrent.api.Publisher.defer;
+import static io.servicetalk.concurrent.internal.SubscriberUtils.handleExceptionFromOnSubscribe;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
+
+/**
+ * A static factory of {@link BufferStrategy} instances.
+ */
+public final class BufferStrategies {
+    private BufferStrategies() {
+        // no instances
+    }
+
+    /**
+     * Returns a {@link BufferStrategy} that creates buffer boundaries based on number of items buffered or time elapsed
+     * since the current buffer boundary started. This does not guarantee that the emitted buffers after applying the
+     * returned {@link BufferStrategy} will have exactly {@code count} number of items. The emitted buffers may have
+     * less or more items than the {@code count}.
+     *
+     * @param count Number of items to add before closing the current buffer boundary, if not already closed. The
+     * emitted buffers may have less or more items than this {@code count}.
+     * @param duration {@link Duration} after which the current buffer boundary is closed, if not already closed.
+     * @param <T> Type of items added to the buffer.
+     * @return {@link BufferStrategy} that creates buffer boundaries based on number of items buffered or time elapsed
+     * since the current buffer boundary started.
+     */
+    public static <T> BufferStrategy<T, Accumulator<T, Iterable<T>>, Iterable<T>> forCountOrTime(
+            final int count, final Duration duration) {
+        return forCountOrTime(count, duration, IMMEDIATE_EXECUTOR);
+    }
+
+    /**
+     * Returns a {@link BufferStrategy} that creates buffer boundaries based on number of items buffered or time elapsed
+     * since the current buffer boundary started. This does not guarantee that the emitted buffers after applying the
+     * returned {@link BufferStrategy} will have exactly {@code count} number of items. The emitted buffers may have
+     * less or more items than the {@code count}.
+     *
+     * @param count Number of items to add before closing the current buffer boundary, if not already closed.
+     * @param duration {@link Duration} after which the current buffer boundary is closed, if not already closed.
+     * @param executor {@link Executor} to use for recording the passed {@code duration}.
+     * @param <T> Type of items added to the buffer.
+     * @return {@link BufferStrategy} that creates buffer boundaries based on number of items buffered or time elapsed
+     * since the current buffer boundary started.
+     */
+    public static <T> BufferStrategy<T, Accumulator<T, Iterable<T>>, Iterable<T>> forCountOrTime(
+            final int count, final Duration duration, final Executor executor) {
+        return forCountOrTime(count, duration, () -> new Accumulator<T, Iterable<T>>() {
+            private final List<T> accumulate = new ArrayList<>();
+
+            @Override
+            public void accumulate(final T t) {
+                accumulate.add(t);
+            }
+
+            @Override
+            public Iterable<T> finish() {
+                return accumulate;
+            }
+        }, executor);
+    }
+
+    /**
+     * Returns a {@link BufferStrategy} that creates buffer boundaries based on number of items buffered or time elapsed
+     * since the current buffer boundary started. This does not guarantee that the emitted buffers after applying the
+     * returned {@link BufferStrategy} will have exactly {@code count} number of items. The emitted buffers may have
+     * less or more items than the {@code count}.
+     *
+     * @param count Number of items to add before closing the current buffer boundary, if not already closed.
+     * @param duration {@link Duration} after which the current buffer boundary is closed, if not already closed.
+     * @param accumulatorSupplier A {@link Supplier} of {@link Accumulator} every time a buffer boundary is closed.
+     * Methods on the {@link Accumulator} returned from this {@link Supplier} may or may not be called.
+     * @param <T> Type of items added to the buffer.
+     * @param <BC> Type of {@link Accumulator} used to accumulate items in a buffer.
+     * @param <B> Type of object created after an {@link Accumulator} is {@link Accumulator#finish() finished}.
+     * @return {@link BufferStrategy} that creates buffer boundaries based on number of items buffered or time elapsed
+     * since the current buffer boundary started.
+     */
+    public static <T, BC extends Accumulator<T, B>, B> BufferStrategy<T, Accumulator<T, B>, B> forCountOrTime(
+            final int count, final Duration duration, Supplier<BC> accumulatorSupplier) {
+        return forCountOrTime(count, duration, accumulatorSupplier, IMMEDIATE_EXECUTOR);
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Collectors.java
```diff
@@ -0,0 +1,170 @@
...
+                        .repeat(__ -> true));
+    }
+
+    /**
+     * A {@link Supplier} of {@link Collector}s and a {@link Publisher} of buffer boundaries.
+     *
+     * @param <T> Type of items collected.
+     */
+    interface CollectorFactory<T> extends Supplier<Collector<T, Iterable<T>>> {
+        Publisher<?> boundaries();
```
normanmaurer, line 56: nit: add java doc

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Collectors.java
```diff
@@ -0,0 +1,170 @@
...
+        public void add(@Nullable final T t) {
+            try {
+                buffer.add(t);
+            } finally {
+                sizeBasedSignal.itemAdded();
+            }
+        }
+
+        @Override
+        public Iterable<T> close() {
```
normanmaurer, line 97: nit: I think we should throw in `add` after `close` was called. This should be part of the interface contract imho

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Collectors.java
```diff
@@ -0,0 +1,170 @@
...
+                AtomicReferenceFieldUpdater.newUpdater(SizeBasedSignal.class, Subscriber.class, "currentSub");
+
+        private final int bufferSize;
+
+        private volatile int pending;
+        @Nullable
+        private volatile Subscriber currentSub;
+
+        SizeBasedSignal(final int bufferSize) {
+            this.bufferSize = bufferSize;
```
normanmaurer, line 115: ensure this is positive ?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -0,0 +1,353 @@
...
+         *     <li>A {@link Collector} representing the current buffer.</li>
+         * </ul>
+         */
+        @SuppressWarnings({"FieldCanBeLocal", "FieldMayBeFinal"})
+        private volatile Object maybeCollector;
+
+        BufferSubscriber(final Publisher<Collector<T, B>> boundaries, final Subscriber<? super B> target,
+                         final Collector<T, B> firstCollector, final int bufferSizeHint) {
+            this.boundaries = boundaries;
+            tSubscription = new DelayedSubscription();
```
normanmaurer, line 102: nit: `this.` to make it consistent

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -0,0 +1,353 @@
...
+            this.boundariesSubscriber = new BoundariesSubscriber(target, tSubscription);
+            this.bufferSizeHint = bufferSizeHint;
+            maybeCollector = requireNonNull(firstCollector);
```
normanmaurer, line 105: nit: `this.` to make it consistent

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -0,0 +1,353 @@
...
+                            BufferSubscriber.this, cPendingBuffers, cPendingBuffers - 1)) {
+                        for (;;) {
+                            final Object cMaybeCollector = maybeCollector;
+                            if (cMaybeCollector == TERMINATED) {
+                                return;
+                            }
+                            assert cMaybeCollector == delayedTermination;
+                            if (maybeCollectorUpdater.compareAndSet(
+                                    BufferSubscriber.this, delayedTermination, TERMINATED)) {
+                                delayedTermination.terminate(target);
```
Scottmitch, line 301: IIUC we may concurrently invoke methods on `target`

```
T1 - BoundariesSubscriber#onNext(..)
  target.onNext(..)
T2 - BufferSubscriber#onError(..)
  terminateFromItemsSource(..)
  cMaybeCollector = maybeCollector -> Collector
  maybeCollectorUpdater.CAS(cMaybeCollector, DelayedTermination) -> true
  tryTerminate()
  cPendingBuffers = pendingBuffers -> 100 (>0)
  pendingBuffersUpdater.CAS(100, 99) -> true
  cMaybeCollector = maybeCollector -> DelayedTermination
  target.onError(..)
  // target.onError and target.onNext concurrently invoked
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -0,0 +1,353 @@
...
+                        }
+                        subscription.request(n);
+                    }
+
+                    @Override
+                    public void cancel() {
+                        try {
+                            subscription.cancel();
+                        } finally {
+                            tCancellable.cancel();
```
Scottmitch, line 217: IIUC `tSubscription` may be concurrently invoked.

```
T1 - requestMore()
  tSubscription.request() //  in invoked in outer `Subscriber`s `onNext`/`onSubscribe`

T2 - target.onSubscribe(bs)
  bs.cancel()
  tCancellable.cancel() // tCancellable == tSubscription
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -0,0 +1,353 @@
...
+                            if (oldUpdater != TERMINATED) {
+                                subscription.cancel();
+                                safeOnError(target, newExceptionForInvalidRequestN(n));
+                            }
+                            return;
+                        }
+                        if (pendingBuffersUpdater.getAndAccumulate(BufferSubscriber.this, n,
+                                FlowControlUtils::addWithOverflowProtection) == 0) {
+                            Object cMaybeCollector = maybeCollector;
+                            if (cMaybeCollector instanceof DelayedTermination) {
```
Scottmitch, line 203: `onError` maybe delivered to `BufferSubscriber`, but it isn't clear what provides visibility here such that the event will be delivered.

```
T1 - BufferSubscriber#onSubscribe(..)
  toSource(boundaries).subscribe(boundariesSubscriber)

...
  BufferSubscriber#onError(..)
  boundariesSubscriber.terminateFromItemsSource
  terminate(this, terminalNotification)
  maybeCollectorUpdater.CAS(null, DelayedTermination)
  tryTerminate(..)
  cPendingBuffers = pendingBuffers -> 0
  return;
  
// Some later time...
T2 - BoundariesSubscriber#onSubscribe
  target.onSubscribe(bs)
  bs.request(1)
  pendingBuffersUpdater.GAA() -> 1
  cMaybeCollector = maybeCollector -> null // no-visibility?
  subscription.request(1)

...
  BoundariesSubscriber#onNext(..)
  cMaybeCollector = maybeCollector -> DelayedTermination
  return;
```

the unconditional dependency on `pendingBuffersUpdater` has a nice side effect the way the code is currently structured in that it also prevents `onError` from coming before `onSubscribe`, but has some other un-intended consequences.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Collector.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Closes this {@link Collector} and no further calls are made to this {@link Collector}.
+     * <p>
+     * This method should be called at most once for any instance of a {@link Collector}
+     *
+     * @return The buffer {@link B} which contains all items {@link #add(Object)} added} to this {@link Collector}
+     * till now.
+     */
+    B close();
```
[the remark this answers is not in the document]
normanmaurer: Big +1... This also fits well with my comment about what to do once it was closed in terms of checks

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -0,0 +1,353 @@
...
+                    if (cMaybeCollector == TERMINATED || cMaybeCollector instanceof DelayedTermination) {
+                        cancellable.cancel();
+                        return;
+                    }
+                    DelayedTermination delayedTermination =
+                            new DelayedTermination(cMaybeCollector, terminalNotification, cancellable);
+                    if (maybeCollectorUpdater.compareAndSet(
+                            BufferSubscriber.this, cMaybeCollector, delayedTermination)) {
+                        // If we raced with ADDING then hand-off termination to ADDING
+                        if (cMaybeCollector != ADDING) {
```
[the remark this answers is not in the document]
NiteshKant: good catch; in this case, we can overwrite `ADDING`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
NiteshKant: good catch; lemme fix

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -0,0 +1,353 @@
...
+                target.onSubscribe(bs);
+            }
+
+            @Override
+            public void onNext(@Nullable final Collector<T, B> collector) {
+                for (;;) {
+                    final Object cMaybeCollector = maybeCollector;
+                    if (cMaybeCollector == ADDING || cMaybeCollector == TERMINATED
+                            || cMaybeCollector instanceof DelayedTermination) {
+                        // If we are ADDING then we will try emitting in the next cycle
```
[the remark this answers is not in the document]
Scottmitch: `Publisher#buffer(..)` exposes a `Publisher<?>` for arbitrary signals so this could be driven by any type of event (maybe even an even that fires only upon certain amount of data that is produced which would make dropping `Collector` problematic). Even if the events are periodic the "next cycle" may not come for some time, or as mentioned you may get unlucky and have data be synchronized with timer (e.g. imagine lots of data being delivered). It would provide more predictable behavior to hand-off the latest `Collector` to ensure some data gets delivered as close to the boundary as possible, and it would be difficult avoid/understand these types of concurrency issues.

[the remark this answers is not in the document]
NiteshKant: @Scottmitch here the case is that the item is emitted concurrently with a boundary. 
Since, item is concurrently emitted, users should not have any expectation of which boundary the item should belong. We have the flexibility to determine whether the item should belong to the previous boundary or next. This code is making the decision of adding the item to the next boundary and handing-off the emission to the boundary source, which happens on the next boundary emission.

NiteshKant:
@Scottmitch based on the feedback I have modified the approach in this PR

- Use `BufferStrategy` as the only parameter to the `buffer()` operator.
- Provide `BufferStrategies` static factory to provide various strategies. Today, only size and time based strategies are provided.

and tightened a few contracts:

- Boundaries source is never expected to terminate; if it does, all bets are off and we bail out immediately.
- The eventual subscriber is only invoked from the scope of the boundaries subscriber.

This has simplified the code both for `BufferStrategy` implementation and the operator implementation. In addition I have restructured the code to encapsulate the complex multi-faceted state variable in a separate class. PTAL

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BufferStrategies.java
```diff
@@ -0,0 +1,298 @@
...
+                        subscriber.onComplete();
+                        return;
+                    }
+                } else if (cState instanceof AccumulatorAndSubscriber) {
+                    @SuppressWarnings("unchecked")
+                    final AccumulatorAndSubscriber<T, B> accumulatorAndSubscriber =
+                            (AccumulatorAndSubscriber<T, B>) cState;
+                    if (accumulatorAndSubscriber.accumulator != breachedAccumulator ||
+                            stateUpdater.compareAndSet(this, cState, null)) {
+                        return;
```
Scottmitch, line 224: if we did `stateUpdater.compareAndSet(this, cState, null)` ... shouldn't we also do `subscriber.onComplete()`?

NiteshKant:
Discussed offline with @Scottmitch , I will pull this in. https://github.com/apple/servicetalk/issues/1048 is created to track the hand-off suggested here.

PR #1028, opened 2020-04-19
`RoundRobinLoadBalancer` should close connections gracefully

NiteshKant:
__Motivation__

When a host used by `RoundRobinLoadBalancer` is unregistered from service discovery, we force close the connection. This means any outstanding requests on the connection will get interrupted. Since service discovery status is a hint we should err on the side of caution and let the outstanding requests complete gracefully.
A fallout of this decision is that a connection may actually be dead and an absence of writes may mean that the we will never detect closure hence graceful closure will never complete.
This is a trade-off between interrupting requests which may otherwise be successfully completed vs not force closing requests that may never complete. As requests could be infinitely latent in general, we assume that users have configured timeouts on the requests and hence all in-flight requests will eventually complete.

__Modification__

Use `closeAsyncGracefully()` instead of `closeAsync()` when a host turns inactive.

__Result__

Host inactivation does not interrupt in-flight requests.


NiteshKant:
Jenkins issue with quality build

idelpivnitskiy:
Build timed out

PR #1029, opened 2020-04-22
Introduce HTTP/2 keep alive

NiteshKant:
__Motivation__

HTTP/2 PING frames are useful to keep a connection alive relatively cheaply in presence of long running streams. ServiceTalk should support adding this functionality for both clients and servers.

__Modification__

- Add `KeepAlivePolicy` to `H2ProtocolConfig` that can be configured to enable keep-alive behavior on either clients or servers. Following features are provided:
    -- Specify an idleness threshold, after which a PING frame will be sent on the connection to detect liveness.
    -- Specify an ack timeout, within which we expect an ack for the sent PING. If no ack is received, the connection is closed.
    -- Specify whether PING frames should be sent even when there are no active streams. This defaults to `false`.

__Result__

Keep alive behavior can be enabled for either client or server.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
@@ -20,8 +20,9 @@ import io.servicetalk.http.netty.H2ProtocolConfig.KeepAlivePolicy;
 import java.time.Duration;
 
 import static java.time.Duration.ofSeconds;
+import static java.util.Objects.requireNonNull;
 
-class DefaultKeepAlivePolicy implements KeepAlivePolicy {
+final class DefaultKeepAlivePolicy implements KeepAlivePolicy {
     static final Duration DEFAULT_IDLE_DURATION = ofSeconds(30);
     static final Duration DEFAULT_ACK_TIMEOUT = ofSeconds(30);
     static final boolean DEFAULT_WITHOUT_ACTIVE_STREAMS = false;
@@ -34,8 +35,8 @@ class DefaultKeepAlivePolicy implements KeepAlivePolicy {
     }
 
     DefaultKeepAlivePolicy(final Duration idleDuration, final Duration ackTimeout, final boolean withoutActiveStreams) {
-        this.idleDuration = idleDuration;
-        this.ackTimeout = ackTimeout;
+        this.idleDuration = requireNonNull(idleDuration);
+        this.ackTimeout = requireNonNull(ackTimeout);
         this.withoutActiveStreams = withoutActiveStreams;
     }
 
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright © 2020 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+import io.servicetalk.http.netty.H2ProtocolConfig.KeepAlivePolicy;
+
+import java.time.Duration;
+
+import static java.time.Duration.ofSeconds;
+
+class DefaultKeepAlivePolicy implements KeepAlivePolicy {
+    static final Duration DEFAULT_IDLE_DURATION = ofSeconds(30);
+    static final Duration DEFAULT_ACK_TIMEOUT = ofSeconds(30);
+    static final boolean DEFAULT_WITHOUT_ACTIVE_STREAMS = false;
+    private final Duration idleDuration;
+    private final Duration ackTimeout;
+    private final boolean withoutActiveStreams;
+
+    DefaultKeepAlivePolicy() {
+        this(DEFAULT_IDLE_DURATION, DEFAULT_ACK_TIMEOUT, false);
+    }
+
+    DefaultKeepAlivePolicy(final Duration idleDuration, final Duration ackTimeout, final boolean withoutActiveStreams) {
+        this.idleDuration = idleDuration;
+        this.ackTimeout = ackTimeout;
+        this.withoutActiveStreams = withoutActiveStreams;
+    }
+
+    @Override
+    public Duration idleDuration() {
+        return idleDuration;
+    }
+
+    @Override
+    public Duration ackTimeout() {
+        return ackTimeout;
+    }
+
+    @Override
+    public boolean withoutActiveStreams() {
+        return withoutActiveStreams;
+    }
+}
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
@@ -19,20 +19,12 @@ import io.servicetalk.http.netty.H2ProtocolConfig.KeepAlivePolicy;
 
 import java.time.Duration;
 
-import static java.time.Duration.ofSeconds;
 import static java.util.Objects.requireNonNull;
 
 final class DefaultKeepAlivePolicy implements KeepAlivePolicy {
-    static final Duration DEFAULT_IDLE_DURATION = ofSeconds(30);
-    static final Duration DEFAULT_ACK_TIMEOUT = ofSeconds(30);
-    static final boolean DEFAULT_WITHOUT_ACTIVE_STREAMS = false;
     private final Duration idleDuration;
     private final Duration ackTimeout;
     private final boolean withoutActiveStreams;
-
-    DefaultKeepAlivePolicy() {
-        this(DEFAULT_IDLE_DURATION, DEFAULT_ACK_TIMEOUT, false);
-    }
 
     DefaultKeepAlivePolicy(final Duration idleDuration, final Duration ackTimeout, final boolean withoutActiveStreams) {
         this.idleDuration = requireNonNull(idleDuration);
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
@@ -19,6 +19,7 @@ import io.servicetalk.http.api.HttpProtocolConfig;
 
 import org.slf4j.event.Level;
 
+import java.time.Duration;
 import java.util.function.BiPredicate;
 import javax.annotation.Nullable;
 
@@ -53,4 +54,46 @@ public interface H2ProtocolConfig extends HttpProtocolConfig {
      */
     @Nullable
     String frameLoggerName();
+
+    /**
+     * Configured {@link KeepAlivePolicy}.
+     *
+     * @return configured {@link KeepAlivePolicy} or {@code null} if none is configured.
+     */
+    @Nullable
+    KeepAlivePolicy keepAlivePolicy();
+
+    /**
+     * A policy for sending <a href="https://tools.ietf.org/html/rfc7540#page-42">PING frames</a> to the peer.
+     */
+    interface KeepAlivePolicy {
+        /**
+         * {@link Duration} of time the connection has to be idle before which a
+         * <a href="https://tools.ietf.org/html/rfc7540#page-42">ping</a> is sent.
+         *
+         * @return {@link Duration} of time the connection has to be idle before which a
+         * <a href="https://tools.ietf.org/html/rfc7540#page-42">ping</a> is sent.
+         */
+        Duration idleDuration();
+
+        /**
+         * {@link Duration} to wait for acknowledgment from the peer after a
+         * <a href="https://tools.ietf.org/html/rfc7540#page-42">ping</a> is sent. If there is no acknowledgment
+         * is received, a closure of the connection will be initiated.
+         *
+         * @return {@link Duration} to wait for acknowledgment from the peer after a
+         * <a href="https://tools.ietf.org/html/rfc7540#page-42">ping</a> is sent.
+         */
+        Duration ackTimeout();
+
+        /**
+         * Whether this policy allows to send <a href="https://tools.ietf.org/html/rfc7540#page-42">pings</a> even if
+         * there are no streams active on the connection.
+         *
+         * @return {@code true} if this policy allows to send
+         * <a href="https://tools.ietf.org/html/rfc7540#page-42">pings</a> even if there are no streams active on the
+         * connection.
+         */
+        boolean withoutActiveStreams();
+    }
 }
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
@@ -64,35 +64,35 @@ public interface H2ProtocolConfig extends HttpProtocolConfig {
     KeepAlivePolicy keepAlivePolicy();
 
     /**
-     * A policy for sending <a href="https://tools.ietf.org/html/rfc7540#page-42">PING frames</a> to the peer.
+     * A policy for sending <a href="https://tools.ietf.org/html/rfc7540#section-6.7">PING frames</a> to the peer.
      */
     interface KeepAlivePolicy {
         /**
          * {@link Duration} of time the connection has to be idle before a
-         * <a href="https://tools.ietf.org/html/rfc7540#page-42">ping</a> is sent.
+         * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a> is sent.
          *
          * @return {@link Duration} of time the connection has to be idle before a
-         * <a href="https://tools.ietf.org/html/rfc7540#page-42">ping</a> is sent.
+         * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a> is sent.
          */
         Duration idleDuration();
 
         /**
          * {@link Duration} to wait for acknowledgment from the peer after a
-         * <a href="https://tools.ietf.org/html/rfc7540#page-42">ping</a> is sent. If there is no acknowledgment
+         * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a> is sent. If there is no acknowledgment
          * is received, a closure of the connection will be initiated.
          *
          * @return {@link Duration} to wait for acknowledgment from the peer after a
-         * <a href="https://tools.ietf.org/html/rfc7540#page-42">ping</a> is sent.
+         * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a> is sent.
          */
         Duration ackTimeout();
 
         /**
-         * Whether this policy allows to send <a href="https://tools.ietf.org/html/rfc7540#page-42">pings</a> even if
-         * there are no streams active on the connection.
+         * Whether this policy allows to send <a href="https://tools.ietf.org/html/rfc7540#section-6.7">pings</a>
+         * even if there are no streams active on the connection.
          *
          * @return {@code true} if this policy allows to send
-         * <a href="https://tools.ietf.org/html/rfc7540#page-42">pings</a> even if there are no streams active on the
-         * connection.
+         * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">pings</a> even if there are no streams active on
+         * the connection.
          */
         boolean withoutActiveStreams();
     }
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
@@ -68,10 +68,10 @@ public interface H2ProtocolConfig extends HttpProtocolConfig {
      */
     interface KeepAlivePolicy {
         /**
-         * {@link Duration} of time the connection has to be idle before which a
+         * {@link Duration} of time the connection has to be idle before a
          * <a href="https://tools.ietf.org/html/rfc7540#page-42">ping</a> is sent.
          *
-         * @return {@link Duration} of time the connection has to be idle before which a
+         * @return {@link Duration} of time the connection has to be idle before a
          * <a href="https://tools.ietf.org/html/rfc7540#page-42">ping</a> is sent.
          */
         Duration idleDuration();
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
@@ -78,8 +78,8 @@ public interface H2ProtocolConfig extends HttpProtocolConfig {
 
         /**
          * {@link Duration} to wait for acknowledgment from the peer after a
-         * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a> is sent. If there is no acknowledgment
-         * is received, a closure of the connection will be initiated.
+         * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a> is sent. If no acknowledgment is received,
+         * a closure of the connection will be initiated.
          *
          * @return {@link Duration} to wait for acknowledgment from the peer after a
          * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a> is sent.
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfigBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfigBuilder.java
@@ -17,12 +17,17 @@ package io.servicetalk.http.netty;
 
 import io.servicetalk.http.api.HttpHeaders;
 import io.servicetalk.http.api.HttpHeadersFactory;
+import io.servicetalk.http.netty.H2ProtocolConfig.KeepAlivePolicy;
 
 import org.slf4j.event.Level;
 
+import java.time.Duration;
 import java.util.function.BiPredicate;
 import javax.annotation.Nullable;
 
+import static io.servicetalk.http.netty.DefaultKeepAlivePolicy.DEFAULT_ACK_TIMEOUT;
+import static io.servicetalk.http.netty.DefaultKeepAlivePolicy.DEFAULT_IDLE_DURATION;
+import static io.servicetalk.http.netty.DefaultKeepAlivePolicy.DEFAULT_WITHOUT_ACTIVE_STREAMS;
 import static io.servicetalk.http.netty.H2HeadersFactory.DEFAULT_SENSITIVITY_DETECTOR;
 import static java.util.Objects.requireNonNull;
 
@@ -37,6 +42,8 @@ public final class H2ProtocolConfigBuilder {
     private BiPredicate<CharSequence, CharSequence> headersSensitivityDetector = DEFAULT_SENSITIVITY_DETECTOR;
     @Nullable
     private String frameLoggerName;
+    @Nullable
+    private KeepAlivePolicy keepAlivePolicy;
 
     H2ProtocolConfigBuilder() {
     }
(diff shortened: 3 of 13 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package io.servicetalk.http.netty;
+
+import io.servicetalk.http.netty.H2ProtocolConfig.KeepAlivePolicy;
+
+import java.time.Duration;
+
+import static java.time.Duration.ofSeconds;
+
+class DefaultKeepAlivePolicy implements KeepAlivePolicy {
```
normanmaurer, line 24: nit: final

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
```diff
@@ -0,0 +1,56 @@
...
+    private final Duration ackTimeout;
+    private final boolean withoutActiveStreams;
+
+    DefaultKeepAlivePolicy() {
+        this(DEFAULT_IDLE_DURATION, DEFAULT_ACK_TIMEOUT, false);
+    }
+
+    DefaultKeepAlivePolicy(final Duration idleDuration, final Duration ackTimeout, final boolean withoutActiveStreams) {
+        this.idleDuration = idleDuration;
+        this.ackTimeout = ackTimeout;
```
normanmaurer, line 38: nit: `requireNonNull(...)` for both `Duration`s.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
```diff
@@ -53,4 +54,46 @@ default String alpnId() {
...
+     */
+    @Nullable
+    KeepAlivePolicy keepAlivePolicy();
+
+    /**
+     * A policy for sending <a href="https://tools.ietf.org/html/rfc7540#page-42">PING frames</a> to the peer.
+     */
+    interface KeepAlivePolicy {
+        /**
+         * {@link Duration} of time the connection has to be idle before which a
```
normanmaurer, line 71: nit: s/which//

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,320 @@
...
+     * <ul>
+     *     <li>{@code null} if graceful close has not started.</li>
+     *     <li>{@link #GRACEFUL_CLOSE_START} if graceful close process has been initiated.</li>
+     *     <li>{@link Future} instance to timeout ack of PING sent to measure RTT.</li>
+     *     <li>{@link #GRACEFUL_CLOSE_SECOND_GO_AWAY_SENT} if we have sent the second go away frame.</li>
+     *     <li>{@link #CLOSED} if the channel is closed.</li>
+     * </ul>
+     */
+    @Nullable
+    private Object gracefulCloseState;
```
normanmaurer, line 75: if java would just have better enums like Swift... This makes me sad :/

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    void initiateGracefulClose(final Runnable whenInitiated) {
+        EventLoop eventLoop = channel.eventLoop();
+        if (eventLoop.inEventLoop()) {
+            doCloseAsyncGracefully0(whenInitiated);
+        } else {
+            try {
+                eventLoop.execute(() -> doCloseAsyncGracefully0(whenInitiated));
+            } catch (Throwable cause) {
```
normanmaurer, line 162: I would just catch `RejectExecutionException` and let the rest bubble up.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,320 @@
...
+        }
+    }
+
+    void channelIdle() {
+        assert channel.eventLoop().inEventLoop();
+
+        if (keepAliveState == CLOSED) {
+            return;
+        }
+        if (activeChildChannels == 0 && disallowKeepAliveWithoutActiveStreams) {
```
normanmaurer, line 176: nit: you could merge the `if` blocks

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,320 @@
...
+            return;
+        }
+        // idleness detected for the first time, send a ping to detect closure, if any.
+        keepAliveState = KEEP_ALIVE_ACK_PENDING;
+        channel.writeAndFlush(new DefaultHttp2PingFrame(KEEP_ALIVE_PING_CONTENT, false))
+                .addListener(future -> {
```
normanmaurer, line 182: consider caching the `ChannelFutureListener` to reduce GC pressure if you need to write the next ping again

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,320 @@
...
+        }
+        // Set the pingState before doing the write, because we will reference the state
+        // when we receive the PING(ACK) to determine if action is necessary, and it is conceivable that the
+        // write future may not be executed which sets the timer.
+        gracefulCloseState = GRACEFUL_CLOSE_START;
+
+        try {
+            whenInitiated.run();
+        } catch (Throwable t) {
+            LOGGER.error("Failed to invoke callback for graceful closure initiation, ignoring.", t);
```
normanmaurer, line 251: consider rethrow if `Error`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,322 @@
...
+                        channel, ree);
+            }
+        }
+    }
+
+    void channelIdle() {
+        assert channel.eventLoop().inEventLoop();
+        assert pingWriteCompletionListener != null;
+
+        if (keepAliveState == CLOSED || activeChildChannels == 0 && disallowKeepAliveWithoutActiveStreams) {
```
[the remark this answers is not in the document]
NiteshKant: Good question. 

Keeping a connection alive when there are no requests has limited usage for connection warm up anticipating a traffic spike. This change is predominantly enabling checking connection liveness in presence of long-lived streams where activity is sparse.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfigBuilder.java
```diff
@@ -81,13 +88,124 @@ public H2ProtocolConfigBuilder enableFrameLogging(final String loggerName) {
...
+            }
+
+            @Override
+            public KeepAliveConfigurator withoutActiveStreams(final boolean withoutActiveStreams) {
+                this.withoutActiveStreams = withoutActiveStreams;
+                return this;
+            }
+
+            @Override
+            public KeepAliveConfigurator disable() {
```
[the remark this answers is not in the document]
NiteshKant: skipping keep-alive != disable keep-alive if we enable keep-alive by default which I think we should in the long term. I will be changing to a separate `KeepAlivePolicyBuilder` as per your other suggestion so this `disable()` method may just become a constant, but we would still want a way to explicitly disable keep-alives.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
```diff
@@ -0,0 +1,57 @@
...
+import io.servicetalk.http.netty.H2ProtocolConfig.KeepAlivePolicy;
+
+import java.time.Duration;
+
+import static java.time.Duration.ofSeconds;
+import static java.util.Objects.requireNonNull;
+
+final class DefaultKeepAlivePolicy implements KeepAlivePolicy {
+    static final Duration DEFAULT_IDLE_DURATION = ofSeconds(30);
+    static final Duration DEFAULT_ACK_TIMEOUT = ofSeconds(30);
```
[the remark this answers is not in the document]
NiteshKant: Hmm .. actually we should not send another ping if the ack is pending; that is a bug, I will fix.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,322 @@
...
+    @Nullable
+    private Object keepAliveState;
+    @Nullable
+    private final GenericFutureListener<Future<? super Void>> pingWriteCompletionListener;
+
+    KeepAliveManager(final Channel channel, @Nullable final KeepAlivePolicy keepAlivePolicy) {
+        this(channel, keepAlivePolicy, (task, delayInMillis) ->
+                channel.eventLoop().schedule(task, delayInMillis, MILLISECONDS),
+                (ch, idlenessThresholdMillis, onIdle) -> ch.pipeline().addLast(
+                        new IdleStateHandler(idlenessThresholdMillis, idlenessThresholdMillis, 0) {
```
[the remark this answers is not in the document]
NiteshKant: I will rename the parameter in `IdlenessDetector` to reflect seconds.

`KeepAliveConfigurator#idleDuration()`  forces the duration to be `>= 10 seconds`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfigBuilder.java
```diff
@@ -81,13 +88,124 @@ public H2ProtocolConfigBuilder enableFrameLogging(final String loggerName) {
         return this;
     }
 
+    /**
+     * Configure keep alive behavior using <a href="https://tools.ietf.org/html/rfc7540#page-42">PING frames</a>.
+     *
+     * @return A {@link KeepAliveConfigurator} which adds keep-alive configuration to this {@link H2ProtocolConfig} when
+     * {@link KeepAliveConfigurator#commit() committed}.
```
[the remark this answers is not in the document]
NiteshKant: 👍 I have changed the approach, PTAL

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Returns a {@link KeepAlivePolicy} that sends a <a href="https://tools.ietf.org/html/rfc7540#section-6.7">
+     * ping</a> if the channel is idle for the passed {@code idleDuration}. Default values are used for other parameters
+     * of the returned {@link KeepAlivePolicy}.
+     *
+     * @param idleDuration {@link Duration} of idleness on a connection after which a
+     * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a> is sent.
+     * @return A {@link KeepAlivePolicy} that disables all keep alive behaviors.
```
idelpivnitskiy, line 54: Incorrect comment for `@return` tag.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
```diff
@@ -0,0 +1,143 @@
...
+     * Returns a {@link KeepAlivePolicy} that sends a <a href="https://tools.ietf.org/html/rfc7540#section-6.7">
+     * ping</a> if the channel is idle for the passed {@code idleDuration} and waits for {@code ackTimeout} for an ack
+     * for that <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a>. Default values are used for other
+     * parameters of the returned {@link KeepAlivePolicy}.
+     *
+     * @param idleDuration {@link Duration} of idleness on a connection after which a
+     * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a> is sent.
+     * @param ackTimeout {@link Duration} to wait for an acknowledgment of a previously sent
+     * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a>.
+     * @return A {@link KeepAlivePolicy} that disables all keep alive behaviors.
```
idelpivnitskiy, line 70: Incorrect comment for `@return` tag.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
```diff
@@ -0,0 +1,143 @@
...
+    public static final class KeepAlivePolicyBuilder {
+        private Duration idleDuration = DEFAULT_IDLE_DURATION;
+        private Duration ackTimeout = DEFAULT_ACK_TIMEOUT;
+        private boolean withoutActiveStreams;
+
+        /**
+         * Set the {@link Duration} of idleness on a connection after which a
+         * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a> is sent.
+         * <p>
+         * <strong>Too short ping durations may cause high network traffic, implementations may enforce a minimum
```
idelpivnitskiy, line 88: Since this is a final class now instead of an interface, "implementation_s_ may enforce" is not correct anymore. Consider something like:
> Too short ping durations may cause high network traffic. Too small duration is not allowed.

Or be specific about 10 seconds.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
```diff
@@ -0,0 +1,143 @@
...
+         * @return {@code this}.
+         * @see KeepAlivePolicy#ackTimeout()
+         */
+        public KeepAlivePolicyBuilder ackTimeout(final Duration ackTimeout) {
+            this.ackTimeout = requireNonNull(ackTimeout);
+            return this;
+        }
+
+        /**
+         * Allow/disallow sending or receiving <a href="https://tools.ietf.org/html/rfc7540#section-6.7">pings</a> even
```
idelpivnitskiy, line 120: "disallow receiving" currently is not true. `KeepAliveManager#pingReceived` always acks received ping frames.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
```diff
@@ -0,0 +1,143 @@
...
+         * when no streams are <a href="https://tools.ietf.org/html/rfc7540#section-5.1">active</a>.
+         *
+         * @param withoutActiveStreams {@code true} if
+         * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">pings</a> are expected when no streams are
+         * <a href="https://tools.ietf.org/html/rfc7540#section-5.1">active</a>.
+         * @return {@code this}.
+         * @see KeepAlivePolicy#withoutActiveStreams()
+         */
+        public KeepAlivePolicyBuilder withoutActiveStreams(final boolean withoutActiveStreams) {
```
idelpivnitskiy, line 129: WDYT about `whenNoActiveStreams` name here and for `KeepAlivePolicy`? "without" does not match with the javadoc language.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
```diff
@@ -0,0 +1,57 @@
...
+    static final boolean DEFAULT_WITHOUT_ACTIVE_STREAMS = false;
```
idelpivnitskiy, line 28: This is an unused variable now.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
```diff
@@ -0,0 +1,57 @@
...
+    private final Duration idleDuration;
+    private final Duration ackTimeout;
+    private final boolean withoutActiveStreams;
+
+    DefaultKeepAlivePolicy() {
```
idelpivnitskiy, line 33: This ctor is unused now.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultKeepAlivePolicy.java
idelpivnitskiy, line 27: If you will remove the default ctor, you can move these constants to `KeepAlivePolicies`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
```diff
@@ -53,4 +54,46 @@ default String alpnId() {
      */
     @Nullable
     String frameLoggerName();
+
+    /**
+     * Configured {@link KeepAlivePolicy}.
+     *
+     * @return configured {@link KeepAlivePolicy} or {@code null} if none is configured.
+     */
+    @Nullable
```
idelpivnitskiy, line 63: Since now we have a special `DISABLE_KEEP_ALIVE` constant, consider making this field not-nullable. We may start with `disabled()` as a default value and change it later if necessary.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
```diff
@@ -0,0 +1,143 @@
...
+         * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a>. If no acknowledgment is received, the
+         * connection will be closed.
+         *
+         * @param ackTimeout {@link Duration} to wait for an acknowledgment of a previously sent
+         * <a href="https://tools.ietf.org/html/rfc7540#section-6.7">ping</a>.
+         * @return {@code this}.
+         * @see KeepAlivePolicy#ackTimeout()
+         */
+        public KeepAlivePolicyBuilder ackTimeout(final Duration ackTimeout) {
+            this.ackTimeout = requireNonNull(ackTimeout);
```
idelpivnitskiy, line 115: Should we check that this is at least 1 ms?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,321 @@
...
+        if (pingFrame.ack()) {
+            long pingAckContent = pingFrame.content();
+            if (pingAckContent == GRACEFUL_CLOSE_PING_CONTENT) {
+                LOGGER.debug("channel={}, graceful close ping ack received.", channel);
+                cancelIfStateIsAFuture(gracefulCloseState);
+                gracefulCloseWriteSecondGoAway();
+            } else if (pingAckContent == KEEP_ALIVE_PING_CONTENT) {
+                cancelIfStateIsAFuture(keepAliveState);
+                keepAliveState = null;
+            }
```
idelpivnitskiy, line 154: RFC says:

> Receivers of a PING frame that does not include an ACK flag MUST send a PING frame with the ACK flag set in response, with an identical payload.

Should we error ack frames with payload we do not expect?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,321 @@
...
+        } else {
+            // Send an ack for the received ping
+            channel.writeAndFlush(new DefaultHttp2PingFrame(pingFrame.content(), true));
+        }
+    }
+
+    void trackActiveStream(final Channel streamChannel) {
+        activeChildChannelsUpdater.incrementAndGet(this);
```
idelpivnitskiy, line 162: If we change this code to always use `addListener` (doesn't look it's practical in real network application that promise will complete so fast):
https://github.com/apple/servicetalk/blob/b22e5ea082e3fba2bd0e97d7ef416941301f835f/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentConnectionContext.java#L229-L234
Then this method will always be on the event loop. `closeFuture` also invokes listeners on the event loop. We will be able to make this variable non-volatile then. WDYT? Can be a follow-up as it's not related to this PR.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
idelpivnitskiy: We can also check at the build time that the `ackTimeout` is less than `idleDuration`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
normanmaurer: +1... I think this should be a connection error

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,321 @@
...
+    void initiateGracefulClose(final Runnable whenInitiated) {
+        EventLoop eventLoop = channel.eventLoop();
+        if (eventLoop.inEventLoop()) {
+            doCloseAsyncGracefully0(whenInitiated);
+        } else {
+            try {
+                eventLoop.execute(() -> doCloseAsyncGracefully0(whenInitiated));
+            } catch (RejectedExecutionException ree) {
+                close0();
+                LOGGER.warn("channel={} EventLoop rejected a task for graceful shutdown, force closing connection",
```
normanmaurer, line 189: first log then call `close0()` just in case that close0 will throw

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,321 @@
...
+            return;
+        }
+        gracefulCloseState = CLOSED;
+        keepAliveState = CLOSED;
+
+        // The way netty H2 stream state machine works, we may trigger stream closures during writes with flushes
+        // pending behind the writes. In such cases, we may close too early ignoring the writes. Hence we flush before
+        // closure, if there is no write pending then flush is a noop.
+        channel.flush();
+        channel.close();
```
normanmaurer, line 308: should the close be done once all previous written data is really written out on the network ? At the moment we may loose some data if the channel was not writable when flush() is called.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    private void cancelIfStateIsAFuture(@Nullable final Object state) {
+        if (state instanceof Future) {
+            try {
+                ((Future<?>) state).cancel(true);
```
normanmaurer, line 314: cancel(...) should never throw. So I would also not catch it.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
```diff
@@ -0,0 +1,143 @@
...
+
+import static io.servicetalk.http.netty.DefaultKeepAlivePolicy.DEFAULT_ACK_TIMEOUT;
+import static io.servicetalk.http.netty.DefaultKeepAlivePolicy.DEFAULT_IDLE_DURATION;
+import static java.time.Duration.ofDays;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A factory to create {@link KeepAlivePolicy} instances.
+ */
+public final class KeepAlivePolicies {
```
Scottmitch, line 30: This utility class is in the `http` namespace but is specific to `h2` functionality.

We should consider the following options:
1. move `KeepAlivePolicy` outside the scope of h2 specific scope (e.g. `H2ProtocolConfig`) and make it less h2 specific. I don't think the API would have to change in a material way (e.g. methods all can stay, comments may need to be adjusted). If/when we were to use it in h1 we would need some additional input from the user on sending the "ping" request, but I think that can be done independently of the `KeepAlivePolicy` API.
OR
2. clarify that this is h2 specific (e.g. add H2 in the name).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
```diff
@@ -0,0 +1,143 @@
...
+     * @return A {@link KeepAlivePolicy} that disables all keep alive behaviors.
+     */
+    public static KeepAlivePolicy whenIdleFor(final Duration idleDuration, final Duration ackTimeout) {
+        return new KeepAlivePolicyBuilder().idleDuration(idleDuration).ackTimeout(ackTimeout).build();
+    }
+
+    /**
+     * A builder of {@link KeepAlivePolicy}.
+     */
+    public static final class KeepAlivePolicyBuilder {
```
Scottmitch, line 79: do we need both static factories and a builder?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
NiteshKant: Good point. I will avoid mentioning a value in the doc though.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
NiteshKant: I guess they are equivalent:

`keepAlive.withoutActiveStreams()` or `keepAlive.withNoActiveStreams()`

`without` is a single negative whereas `withNo` is a positive (with) and a negative (no), so I feel `without` is easier to understand. 

Javadocs language has more context than a name so I feel it is OK to be different.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
NiteshKant: The constant equating to disable is just an implementation detail IMHO. This returning `null` better represents that there is no policy set.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
NiteshKant: I would like to avoid making assumptions here as this does not have a network traffic impact. Although a good point that smaller values will cause issues with the scheduler. Let me also pass the unit to the internal `Scheduler` interface.

> We can also check at the build time that the ackTimeout is less than idleDuration.

There isn't really a relation b/w these two values.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
NiteshKant: hmm I would be hesitant to assume that we are the only entity sending PINGs when we do not have the visibility of which all frames are sent on the wire.
If at all we do feel this is necessary, I think this should be part of an enforcement policy as opposed to the keep-alive policy as it exists in this PR.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
NiteshKant: I am not sure about the API inter-connects here and would like to defer.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
NiteshKant: (this code is pre-existing, just refactored).

We enter here when `active streams == 0`, there is a fatal error or missing keep-alive. In all these cases we do not expect further graceful closure. 
Practically, this would require we track all writes on the channel from here, ROI of which is arguable here.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
NiteshKant: Whenever there is a state cleanup required (here sending a second go away for example) we just are pedantic about unexpected exceptions. So, this is inline with other places.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,321 @@
...
+        } else {
+            // Send an ack for the received ping
+            channel.writeAndFlush(new DefaultHttp2PingFrame(pingFrame.content(), true));
+        }
+    }
+
+    void trackActiveStream(final Channel streamChannel) {
+        activeChildChannelsUpdater.incrementAndGet(this);
+        streamChannel.closeFuture().addListener(f -> {
+            activeChildChannelsUpdater.decrementAndGet(this);
```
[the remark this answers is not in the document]
NiteshKant: I will defer the `inEventLoop` assumption as this is pre-existing code.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    void initiateGracefulClose(final Runnable whenInitiated) {
+        EventLoop eventLoop = channel.eventLoop();
+        if (eventLoop.inEventLoop()) {
+            doCloseAsyncGracefully0(whenInitiated);
+        } else {
+            try {
+                eventLoop.execute(() -> doCloseAsyncGracefully0(whenInitiated));
+            } catch (RejectedExecutionException ree) {
```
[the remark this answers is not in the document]
NiteshKant: Lemme just remove the error handling here as this has been changed twice based on comments. All callers handle errors from here and have more context how to handle exceptions.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,321 @@
...
+                            LOGGER.debug(
+                                    "channel={}, timeout {}ms waiting for keep-alive PING(ACK), writing go_away.",
+                                    this.channel, pingAckTimeoutMillis);
+                            channel.writeAndFlush(new DefaultHttp2GoAwayFrame(NO_ERROR))
+                                    .addListener(goAwayListener);
+                        }
+                    }, pingAckTimeoutMillis);
+                }
+            };
+            int idleInSeconds = (int) keepAlivePolicy.idleDuration().getSeconds();
```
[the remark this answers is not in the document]
NiteshKant: Good point, lemme put an upper bound that resembles infinity for the time here.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
NiteshKant: Good point, lemme add the `H2` prefix as this is creating a `H2ProtocolConfig.KeepAlivePolicy`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
NiteshKant: We don't need it but the static methods are convenient to express common cases, eg: `disabled()`, `whenIdleFor()` as opposed to `new KeepAlivePolicyBuilder().idleDuration().ackTimeout().build()`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    KeepAliveManager(final Channel channel, @Nullable final KeepAlivePolicy keepAlivePolicy,
+                     final Scheduler scheduler, final IdlenessDetector idlenessDetector) {
+        this.channel = channel;
+        this.scheduler = scheduler;
+        if (keepAlivePolicy != null) {
+            disallowKeepAliveWithoutActiveStreams = !keepAlivePolicy.withoutActiveStreams();
+            pingAckTimeoutMillis = keepAlivePolicy.ackTimeout().toMillis();
+            final GenericFutureListener<Future<? super Void>> goAwayListener = f -> {
```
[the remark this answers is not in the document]
idelpivnitskiy: There was another comment that suggested not to inline but cache the value instead :)
https://github.com/apple/servicetalk/pull/1029#discussion_r413046611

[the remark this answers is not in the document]
Scottmitch: go_away is once (or twice) per connection on shutdown so I don't think GC would be the dominating factor here (vs readability).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAlivePolicies.java
Scottmitch: fair enough. 🚢

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,321 @@
...
+
+    void initiateGracefulClose(final Runnable whenInitiated) {
+        EventLoop eventLoop = channel.eventLoop();
+        if (eventLoop.inEventLoop()) {
+            doCloseAsyncGracefully0(whenInitiated);
+        } else {
+            try {
+                eventLoop.execute(() -> doCloseAsyncGracefully0(whenInitiated));
+            } catch (RejectedExecutionException ree) {
+                close0();
```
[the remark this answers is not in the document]
NiteshKant: I removed the error handling here.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -0,0 +1,315 @@
...
+                                        if (f.isSuccess()) {
+                                            LOGGER.debug("Closing channel={}, after keep-alive timeout.", this.channel);
+                                            KeepAliveManager.this.close0();
+                                        }
+                                    });
+                        }
+                    }, pingAckTimeoutNanos, NANOSECONDS);
+                }
+            };
+            int idleInSeconds = (int) keepAlivePolicy.idleDuration().getSeconds();
```
Scottmitch, line 132: KeepAlivePolicyBuilder applies a limit when setting the Idle duration, although I think it is safer to avoid overflow here and then we don't have to worry about applying a limit upfront or missing applying the limit if someone implement their own `KeepAlivePolicy` etc..

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelListenableAsyncCloseable.java
```diff
@@ -82,7 +82,13 @@ public final Completable closeAsyncGracefully() {
             @Override
             protected void handleSubscribe(final Subscriber subscriber) {
                 if (stateUpdater.compareAndSet(NettyChannelListenableAsyncCloseable.this, OPEN, GRACEFULLY_CLOSING)) {
-                    doCloseAsyncGracefully();
+                    try {
+                        doCloseAsyncGracefully();
+                    } catch (Throwable t) {
+                        subscriber.onSubscribe(IGNORE_CANCEL);
```
Scottmitch, line 88: use `SubscriberUtils#deliverTerminalFromSource` instead?

PR #1034, opened 2020-04-29
More robust exception handling from Subscribers

Scottmitch:
Motivation:
The ReactiveStreams specification states that Publisher#subscribe must
not throw [1] and also that Subscriber methods must not throw [2].
However in practice these methods may throw and we need to make a best
effort to propagate exceptions and preserve the asynchronous control
flow. SubscriberUtils has utility methods deliverTerminalFromSource and
handleExceptionFromOnSubscribe to help with this task. However the
latter method may propagate exceptions which will result in our error
recovery code delivering multiple onSubscribe and onError calls which
further violates the ReactiveStreams specification [1][3]. Here is an
example:

```
NeverPublisher#subscribe()
subscriber.onSubscribe(..)
catch(cause)
handleExceptionFromOnSubscribe(..)
subscriber.onError(cause)
catch(cause2)  (in Publisher#handleSubscribe(..))
deliverTerminalFromSource(..)
subscriber.onSubscribe(..) // duplicate call by our error recovery
catch(cause3)
handleExceptionFromOnSubscribe(..)
subscriber.onError(cause3) // duplicate call by our error recovery
```

[1] https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#1.9
[2] https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#2.13
[3] https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#1.7

Modifications:
- SubscriberUtils#handleExceptionFromOnSubscribe should not re-throw
- Usage of IGNORE_CANCEL and EMPTY_SUBSCRIPTION should be investigated
to utilize the SubscriberUtils methods to minimize impact of invalid
Subscribers and preserve asynchronous control flow

Result:
More robust best-effort error recovery in the event of invalid
Subscribers throwing exceptions.

### DIFF
--- a/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/SubscriberUtils.java
+++ b/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/SubscriberUtils.java
@@ -159,7 +159,7 @@ public final class SubscriberUtils {
      * @param subscriber The {@link Subscriber} to terminate.
      * @param <T> The type of {@link Subscriber}.
      */
-    public static <T> void deliverTerminalFromSource(Subscriber<T> subscriber) {
+    public static <T> void deliverCompleteFromSource(Subscriber<T> subscriber) {
         try {
             subscriber.onSubscribe(EMPTY_SUBSCRIPTION);
         } catch (Throwable t) {
@@ -176,14 +176,14 @@ public final class SubscriberUtils {
      * @param value The value to pass to {@link SingleSource.Subscriber#onSuccess(Object)}.
      * @param <T> The type of {@link SingleSource.Subscriber}.
      */
-    public static <T> void deliverTerminalFromSource(SingleSource.Subscriber<T> subscriber, @Nullable T value) {
+    public static <T> void deliverSuccessFromSource(SingleSource.Subscriber<T> subscriber, @Nullable T value) {
         try {
             subscriber.onSubscribe(IGNORE_CANCEL);
         } catch (Throwable t) {
             handleExceptionFromOnSubscribe(subscriber, t);
             return;
         }
-        safeOnComplete(subscriber, value);
+        safeOnSuccess(subscriber, value);
     }
 
     /**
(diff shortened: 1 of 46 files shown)


servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/SubscriberUtils.java
```diff
@@ -335,4 +349,39 @@ public static void handleExceptionFromOnSubscribe(CompletableSource.Subscriber s
...
     }
+
+    /**
+     * Invokes {@link SingleSource.Subscriber#onSuccess(Object)} ignoring an occurred exception if any.
+     * @param subscriber The {@link SingleSource.Subscriber} that may throw an exception from
+     * {@link SingleSource.Subscriber#onSuccess(Object)}.
+     * @param value The value to pass to {@link SingleSource.Subscriber#onSuccess(Object)}.
+     * @param <T> The type of {@link SingleSource.Subscriber}.
+     */
+    public static <T> void safeOnComplete(SingleSource.Subscriber<T> subscriber, @Nullable T value) {
```
NiteshKant, line 360: The name seems misleading, perhaps you wanted to name this `deliverTerminalFromSource`?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/Flush.java
```diff
@@ -74,24 +73,22 @@ private Flush() {
...
-            }
-            subscriber.onSubscribe(new Subscription() {
-                @Override
-                public void request(long n) {
-                    subscription.request(n);
-                }
+            } finally {
+                // As long as we call onSubscribe we can let exceptions propagate and the source should terminate
+                // this Subscriber for cleanup.
+                subscriber.onSubscribe(new Subscription() {
```
NiteshKant, line 79: This seems to be different than what was happening before as we are not cancelling the `subscription` any more. I think the assumption here is that this exception will bubble up to the source which effectively cancels the `subscription`?

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/SubscriberUtils.java
Scottmitch: good catch ... this method should be broken up into `deliverTerminalFromSource` and `safeOnComplete`

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/Flush.java
Scottmitch: Correct. I attempted to articulate that in the comment above but let me expand on this to clarify more.

PR #1035, opened 2020-04-29
Adding TerminalSignalConsumer for Single.whenFinally et al

colestanfield:
Motivation:

Not every consumer needs the onSuccess value.

Modifications:

- Added whenFinally/beforeFinally/afterFinally that takes a TerminalSignalConsumer
- Extended the type bounds for SingleTerminalSignalConsumer

Result:

- Can now call Single.whenFinally(TerminalSignalConsumer) if users don't need the onSuccess value
- Compiler will allow you to pass a SingleTerminalSignalConsumer typed with a super type to whenFinally/etc


NiteshKant:

NiteshKant:
Thanks @colestanfield

Issue #1033, opened 2020-04-27
CompletableProcessor, SingleProcess, and DynamicCompositeCancellable cancellation ordering


Scottmitch:
CompletableProcessor, SingleProcess, and DynamicCompositeCancellable current cancel and pending futures in the order they are inserted in (ignoring concurrency non-determinism). Recently we switched to using MPMC data structures which may allow us to remove specific elements without going through the drain loop, with the side effect of no longer invoking elements out of order. We should consider if ordering is necessary and if not simplify to only remove the element in question.

original discussions:
- https://github.com/apple/servicetalk/pull/1026#discussion_r415977824
- https://github.com/apple/servicetalk/pull/1026#discussion_r416073415

PR #1036, opened 2020-04-30
Remove in-order termination of listeners

NiteshKant:
__Motivation__

`CompletableProcessor`, `SingleProcessor`, and `DynamicCompositeCancellable` currently terminates listeners in the order they are inserted. This ordering isn't expected from the API and it enforces code complexity.

__Modification__

Remove the in-order termination semantics from these classes.

__Result__

Optional in-order termination semantics is removed and code is simplified.

Fixes https://github.com/apple/servicetalk/issues/1033

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
@@ -29,6 +29,7 @@ import javax.annotation.Nullable;
 import static io.servicetalk.concurrent.internal.TerminalNotification.complete;
 import static io.servicetalk.concurrent.internal.ThrowableUtils.catchUnexpected;
 import static io.servicetalk.utils.internal.PlatformDependent.throwException;
+import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
 
 /**
  * A {@link Completable} which is also a {@link Subscriber}. State of this {@link Completable} can be modified by using
@@ -36,7 +37,7 @@ import static io.servicetalk.utils.internal.PlatformDependent.throwException;
  */
 final class CompletableProcessor extends Completable implements Processor {
     private static final AtomicReferenceFieldUpdater<CompletableProcessor, TerminalNotification> terminalSignalUpdater =
-            AtomicReferenceFieldUpdater.newUpdater(CompletableProcessor.class, TerminalNotification.class,
+            newUpdater(CompletableProcessor.class, TerminalNotification.class,
                     "terminalSignal");
 
     private final Queue<Subscriber> subscribers = new ConcurrentLinkedQueue<>();
@@ -53,9 +54,8 @@ final class CompletableProcessor extends Completable implements Processor {
         subscriber.onSubscribe(delayedCancellable);
         if (subscribers.offer(subscriber)) {
             TerminalNotification terminalSignal = this.terminalSignal;
-            if (terminalSignal != null) {
-                // To ensure subscribers are notified in order we go through the queue to notify subscribers.
-                notifyListeners(terminalSignal);
+            if (terminalSignal != null && subscribers.remove(subscriber)) {
+                terminalSignal.terminate(subscriber);
             } else {
                 delayedCancellable.delayedCancellable(() -> {
                     // Cancel in this case will just cleanup references from the queue to ensure we don't prevent GC of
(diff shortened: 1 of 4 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
```diff
@@ -29,14 +29,15 @@
...
+import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
 
 /**
  * A {@link Completable} which is also a {@link Subscriber}. State of this {@link Completable} can be modified by using
  * the {@link Subscriber} methods which is forwarded to all existing or subsequent {@link Subscriber}s.
  */
 final class CompletableProcessor extends Completable implements Processor {
     private static final AtomicReferenceFieldUpdater<CompletableProcessor, TerminalNotification> terminalSignalUpdater =
-            AtomicReferenceFieldUpdater.newUpdater(CompletableProcessor.class, TerminalNotification.class,
+            newUpdater(CompletableProcessor.class, TerminalNotification.class,
```
normanmaurer, line 40: nit: I think usually we dont use imports for this one...

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleProcessor.java
```diff
@@ -58,9 +58,16 @@ protected void handleSubscribe(final Subscriber<? super T> subscriber) {
...
         if (subscribers.offer(subscriber)) {
             Object terminalSignal = this.terminalSignal;
-            if (terminalSignal != TERMINAL_NULL) {
-                // To ensure subscribers are notified in order we go through the queue to notify subscribers.
-                notifyListeners(terminalSignal);
+            if (terminalSignal != TERMINAL_NULL && subscribers.remove(subscriber)) {
+                if (terminalSignal instanceof TerminalNotification) {
+                    final Throwable error = ((TerminalNotification) terminalSignal).cause();
+                    assert error != null : "Cause can't be null from TerminalNotification.error(..)";
+                    subscriber.onError(error);
```
Scottmitch, line 65: should we use `safeOnError` and `safeOnSuccess` here? Otherwise the Single may catch the throwable and propagate another duplicate terminal event.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
```diff
@@ -53,9 +54,8 @@ protected void handleSubscribe(Subscriber subscriber) {
         subscriber.onSubscribe(delayedCancellable);
         if (subscribers.offer(subscriber)) {
             TerminalNotification terminalSignal = this.terminalSignal;
-            if (terminalSignal != null) {
-                // To ensure subscribers are notified in order we go through the queue to notify subscribers.
-                notifyListeners(terminalSignal);
+            if (terminalSignal != null && subscribers.remove(subscriber)) {
+                terminalSignal.terminate(subscriber);
```
Scottmitch, line 58: should we have a `TerminalNotification#terminate(..)` variant which doesn't throw? Otherwise the `Completable#handleSubscribe` would catch the exception and propagate a duplicate terminal event.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleProcessor.java
Scottmitch: also while we are here perhaps consider using `try { subscriber.onSubscribe(..); } catch(e) { handleExceptionFromOnSubscribe(..) } ` for more graceful error handling.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SetDynamicCompositeCancellable.java
```diff
@@ -73,7 +80,9 @@ public boolean add(Cancellable toAdd) {
         if (!cancellables.add(toAdd)) {
             toAdd.cancel(); // out of memory, or user has implemented equals/hashCode so there is overlap.
         } else if (isCancelled()) {
-            cancelAll(); // avoid concurrency on toAdd if another thread invokes cancel(), just cancel all.
```
Scottmitch, line 76: note this comment. I don't think we can safely remove/cancel here because `Set` doesn't provide an API to "atomically" get/remove items during iteration (like `Queue#poll()`). This change may lead to concurrent invocation of `toAdd.cancel()`.

Scottmitch: `NavigableSet` provides [pollFirst](https://docs.oracle.com/javase/7/docs/api/java/util/NavigableSet.html#pollFirst()) and [ConcurrentSkipListSet](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentSkipListSet.html) is an implementation we could use, but this requires item ordering (e.g. `Comparable`, or `Comparator`) which doesn't seem practical in this case. We can always punt on the `SetDynamicCompositeCancellable` specific part of this change.

PR #1039, opened 2020-05-04
Improve `Processors`

NiteshKant:
__Motivation__

`Processors` is a static factory to create different `Processor` implementations but it is missing a few things:

- No way to create a `Processor` for `PublisherSource`
- `BlockingIterable.Processor` does not provide a way to create custom buffering strategies.

__Modification__

- Introduce `PublisherSource.Processor`.
- Add new methods to `Processors` to create `PublisherSource.Processor` instances.
- Since buffering for a streaming `Processor` is involved and may differ for different use cases (eg: fail on overflow, drop on overflow, etc), provide a public API for `PublisherProcessorBuffer` that can help use different strategies for buffering.
- Use similar buffering API for `BlockingIterable.Processor`.
- Removes the custom `SpscPublisherProcessor` implementation.
- Replace all `SpscPublisherProcessor` usages with the new implementation.

__Result__

Enhanced usefulness of `Processor` static factory.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AbstractProcessorBuffer.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+        if (nextItem == NULL_ITEM) {
+            consumer.consumeItem(null);
+        } else {
+            @SuppressWarnings("unchecked")
+            T t = (T) nextItem;
+            consumer.consumeItem(t);
+        }
+        return true;
+    }
```
normanmaurer, line 90: nit: please add javadocs for all protected methods.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AbstractProcessorBuffer.java
```diff
@@ -0,0 +1,91 @@
...
+    protected abstract void addTerminal(TerminalNotification terminalNotification);
+
+    protected boolean consumeIfTerminal(final BufferConsumer<T> consumer, @Nullable final Object signal) {
+        if (signal instanceof TerminalNotification) {
+            TerminalNotification terminalNotification = (TerminalNotification) signal;
+            if (terminalNotification.cause() != null) {
+                consumer.consumeTerminal(terminalNotification.cause());
+            } else {
+                consumer.consumeTerminal();
+            }
```
normanmaurer, line 72: nit: consider this:

```
Throwable cause = ((TerminalNotification) signal).cause();
if (cause != null) {
    consumer.consumeTerminal(cause));
} else {
    consumer.consumeTerminal();
}
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AbstractProcessorBuffer.java
```diff
@@ -0,0 +1,91 @@
...
+        final TerminalNotification terminalNotification = terminal;
+        if (terminalNotification != null) {
+            throw new IllegalStateException("Buffer " + this + " is already terminated: " + terminal);
+        }
+        addItem(item == null ? NULL_ITEM : item);
+    }
+
+    @Override
+    public final void terminate() {
+        if (terminalUpdater.compareAndSet(this, null, complete())) {
```
normanmaurer, line 48: Consider storing the return value of `complete()` and use the same value for `addTerminal(...)`. Just to be sure its the same instance at the end...

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AbstractPublisherProcessorBuffer.java
```diff
@@ -0,0 +1,88 @@
...
+            throw new IllegalArgumentException("maxBuffer: " + maxBuffer + " (expected > 0)");
+        }
+        this.maxBuffer = maxBuffer;
+        this.signals = requireNonNull(signals);
+    }
+
+    @Override
+    protected void addItem(final Object item) {
+        if (bufferedUpdater.getAndAccumulate(this, 1,
+                (prev, next) -> prev == maxBuffer ? maxBuffer :
```
normanmaurer, line 50: imho this is quite hard to read...

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherProcessorBuffers.java
```diff
@@ -0,0 +1,88 @@
...
+     * @param <T> Type of items added to the returned {@link PublisherProcessorBuffer}.
+     * @return A new {@link PublisherProcessorBuffer}.
+     */
+    public static <T> PublisherProcessorBuffer<T> fixedSizeDropOldest(final int maxBuffer) {
+        return new AbstractPublisherProcessorBuffer<T, ConcurrentLinkedQueue<Object>>(maxBuffer,
+                new ConcurrentLinkedQueue<>()) {
+            @Override
+            void offerPastBufferSize(final Object signal, final ConcurrentLinkedQueue<Object> queue) {
+                queue.poll(); // drop oldest
+                queue.offer(signal);
```
normanmaurer, line 84: I think this is racy as you may have `queue.offer` return false if this method is invoked by multiple threads.

You should do something like:
```java
void offerPastBufferSize(final Object signal, final ConcurrentLinkedQueue<Object> queue) {
    while (!queue.offer(signal)) {
        queue.poll();
    }
}
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultBlockingIterableProcessor.java
```diff
@@ -144,43 +120,33 @@ public boolean hasNext() {
...
-            if (next == NULL_MASK) {
-                return null;
-            } else {
-                @SuppressWarnings("unchecked")
-                T t = (T) next;
-                return t;
-            }
+        private boolean hasNextWhenTerminated() {
+            assert terminal != null;
+            return terminal.cause() != null && (boolean) throwException(terminal.cause());
```
Scottmitch, line 134: `throwException` casting to a `boolean` doesn't look right, despite the return not being expected. Consider refactoring for clarity:

```java
Throwable cause = terminal.cause();
if (cause != null) {
  throwException(cause);
}
return false;
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherProcessorBuffers.java
normanmaurer: still I think it should be done to guard against problems later when the code is changed. At least I would add a comment with a warning that the queue must be unbound.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AbstractPublisherProcessorBuffer.java
normanmaurer: thats fine

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BlockingProcessorBuffer.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Adds an item to this buffer.
+     *
+     * @param item to add.
+     * @throws InterruptedException If the add was interrupted.
+     */
+    void add(@Nullable T item) throws InterruptedException;
+
+    /**
+     * Terminates this buffer, such that no further modifications of this buffer are allowed. Subsequent
```
normanmaurer, line 40: maybe clarify what will happen if you try to call `add(...)` after it (maybe throwing).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BlockingProcessorBuffer.java
```diff
@@ -0,0 +1,90 @@
...
+     * {@link BufferConsumer#consumeTerminal() successful} or {@link BufferConsumer#consumeTerminal(Throwable) failed}
+     * termination.
+     * <p>
+     * This method will block till an item or a terminal event is available in the buffer or the passed {@code waitFor}
+     * duration has elapsed.
+     *
+     * @param consumer {@link BufferConsumer} to consume the next item or termination in this buffer
+     * @param waitFor Duration to wait for an item or termination to be available.
+     * @param waitForUnit {@link TimeUnit} for {@code waitFor}.
+     * @return {@code true} if any method was called on the passed {@link BufferConsumer}.
```
normanmaurer, line 83: is this really true ? Like will it also return true if `consume` was called ?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultBlockingIterableProcessor.java
```diff
@@ -127,7 +100,9 @@ public T next() {
             if (!hasNext()) {
                 throw new NoSuchElementException();
             }
-            return processNext();
+            T next = this.next;
+            this.next = null;
+            return next;
```
normanmaurer, line 105: these 3 lines of code are the same as the change of line 92... Consider sharing the code via a private method:

```java
private T consumeNext() {
    T next = this.next;
    this.next = null;
    return null;
}
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultBlockingIterableProcessor.java
```diff
@@ -144,43 +119,37 @@ public boolean hasNext() {
...
-            final Object next;
+            final boolean consumed;
             try {
-                next = buffer.take();
+                consumed = buffer.consume(this);
             } catch (InterruptedException e) {
                 return throwException(e);
             }
-            return processHasNext(next);
+            return terminal == null ? consumed : hasNextWhenTerminated();
```
normanmaurer, line 128: nit: this is the same as line 83. Maybe share ?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultBlockingIterableProcessor.java
```diff
@@ -144,43 +119,37 @@ public boolean hasNext() {
...
-        private T processNext() {
-            Object next = this.next;
-            this.next = null;
-            if (next == NULL_MASK) {
-                return null;
-            } else {
-                @SuppressWarnings("unchecked")
-                T t = (T) next;
-                return t;
+        private boolean hasNextWhenTerminated() {
```
normanmaurer, line 131: how is the `boolean` useful here ? This always returns false....

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultBlockingProcessorBuffer.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public boolean consume(final BufferConsumer<T> consumer, final long waitFor, final TimeUnit waitForUnit)
+            throws TimeoutException, InterruptedException {
+        if (consumeIfTerminal(consumer, signals.peek())) {
+            return true;
+        }
+
+        Object nextItem = signals.poll(waitFor, waitForUnit);
```
normanmaurer, line 70: isn't this racy? Two threads could pass the `peek()` above as the first item at the time of the call is not a terminal but after the first calls `poll()` the next would be one

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherProcessor.java
```diff
@@ -0,0 +1,259 @@
...
+                } catch (Throwable t) {
+                    // Assume that we did not deliver terminal to the consumer.
+                    earlyTerminateConsumerHoldingLock(target, t);
+                    return;
+                }
+                if (consumed) {
+                    pending = Long.MIN_VALUE;
+                    return;
+                }
+                break;
```
normanmaurer, line 197: nit: use `return` to be consistent.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherProcessor.java
```diff
@@ -0,0 +1,259 @@
...
+        try {
+            delayedSubscription.cancel();
+        } finally {
+            consumer.consumeTerminal(cause);
+        }
+    }
+
+    private static final class SubscriberBufferConsumer<T> implements BufferConsumer<T> {
+        private final Subscriber<? super T> subscriber;
+        private boolean terminated;
```
normanmaurer, line 213: does this need to be volatile ?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AbstractPublisherProcessorBuffer.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+        this.maxBuffer = maxBuffer;
+        this.signals = requireNonNull(signals);
+    }
+
+    @Override
+    public void add(@Nullable final T item) {
+        if (bufferedUpdater.getAndAccumulate(this, 1,
+                (prev, next) -> prev == maxBuffer ? maxBuffer : (prev + next)) == maxBuffer) {
+            offerPastBufferSize(maskNull(item), signals);
```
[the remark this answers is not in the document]
NiteshKant: We have been preferring inlining such calls as opposed to storing it in a local field.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BlockingProcessorBuffer.java
NiteshKant: actually we are not enforcing this now, so lemme change `allowed` => `expected`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BlockingProcessorBuffer.java
NiteshKant: Yes that is correct. Return value is indicating whether an item or terminal was consumed.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultBlockingIterableProcessor.java
NiteshKant: mmm .. its a one-liner which uses a method anyways. I will avoid creating a method for it

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultBlockingIterableProcessor.java
NiteshKant: It just make it easy to use in `hasNext()`:

```java
return hasNextWhenTerminated()
```
instead of 
```java
hasNextWhenTerminated();
return false;
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultBlockingProcessorBuffer.java
NiteshKant: yikes, good catch. Lemme make this consistent with the above `consume` method.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherProcessor.java
NiteshKant: nope, this class ins't used concurrently by different threads.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherProcessorBuffer.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @param <T>  Type of items stored in this buffer.
+ */
+public interface PublisherProcessorBuffer<T> extends ProcessorBuffer<T> {
+
+    /**
+     * Try to consume the next item stored in this buffer. If there are no items stored in the buffer and the buffer has
+     * terminated {@link #terminate() successfully} or with an {@link #terminate(Throwable) error} then consume that
+     * {@link BufferConsumer#consumeTerminal() successful} or
+     * {@link BufferConsumer#consumeTerminal(Throwable) failed} termination.
```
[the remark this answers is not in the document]
Scottmitch: `BufferConsumer` implementations that exist today do not support concurrent access, and it would be surprising to have concurrency consumption, is this something we expect when the consumer is typically a `Subscriber`? The connection to the `Processor` comments is nuanced and I think it is worth clarifying here to avoid confusion (see https://github.com/apple/servicetalk/pull/1039#discussion_r421359876, https://github.com/apple/servicetalk/pull/1039#discussion_r421365637).

NiteshKant:
quality build failed due to jenkins setup issues.

NiteshKant:
@normanmaurer @Scottmitch thanks for the reviews! I have addressed/answered all your comments and will pull this in now.

PR #1042, opened 2020-05-06
`SingleConcatWithPublisher` and invalid request-n

NiteshKant:
__Motivation__

If invalid calls to request-n are made for `Single.concat(Publisher)` before the `Single` result is propagated, we first emit the result and then propagate the invalid demand.
For cases, when the concatenated `Publisher` completes inline with `subscribe()`` (eg: `Publisher.empty()`), this means we will never deliver an error for the invalid demand.

__Modification__

Deliver error when invalid demand is made in the above case instead of `Single`'s result and do not subscribe to the concatenated `Publisher`

__Result__

Invalid demand always terminates the `Subscriber` with an error.

### DIFF
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleConcatWithPublisherTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleConcatWithPublisherTest.java
@@ -15,6 +15,7 @@
  */
 package io.servicetalk.concurrent.api.single;
 
+import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.concurrent.api.TestCancellable;
 import io.servicetalk.concurrent.api.TestPublisher;
 import io.servicetalk.concurrent.api.TestPublisherSubscriber;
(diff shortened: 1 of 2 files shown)


Scottmitch:
> `subscribe()``

@NiteshKant - The commit message has an extra back-tick which is confusing the markdown parsing.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleConcatWithPublisherTest.java
```diff
@@ -140,6 +152,7 @@ private void invalidThenValidRequest(long invalidN) {
     @Test
     public void request0PropagatedAfterSuccess() {
         source.onSuccess(1);
+        subscriber.request(1); // get the success from the Single
```
Scottmitch, line 155: consider breaking into 2 tests:

- `request(0)` delivers on `onError`
- `request(0)` after `request(>0)` delivers the item and propagates the `0` to the subscription (what is done now in this test)

NiteshKant: The new test that I added `invalidRequestNWithInlineSourceCompletion()` is equivalent to the first test you mention, that is why I did not add it separately.

PR #1043, opened 2020-05-08
Disable caching of CNAMEs to workaround resolving issues

normanmaurer:
Motivation:

In netty there is a seperate CNAME cache that is used to cache CNAMEs that are resolved during resolving an A / AAAA record. While this sounded like a great idea when I first implemented this in netty it turned out this is problematic and may produce errors while resolving hostnames later on.
The problem with the CNAME cache is that we may end up with entries in the CNAME cache wile the original cached resolution was already removed from the cache (due TTL). In this case we may find an entry in teh CNAME cache that we will then try to resolve via the specified nameserver. This resolution may fail as the nameserver may not return an entry for the previous cached CNAME. This usually happens when we receive the full "resolution chain" during query for an A / AAAA record.

Modifications:

Introduce NoopDnsCnameCache and explicit set it when building the used DnsNameResolver. This is a workaround as netty itself will do the same in the next release and deprecate DnsCnameCache

Result:

No more resolution failures caused by cached CNAME entries


normanmaurer:
This is a workaround for https://github.com/netty/netty/pull/10263

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/NoopDnsCnameCache.java
```diff
@@ -0,0 +1,52 @@
...
+
+import io.netty.channel.EventLoop;
+import io.netty.resolver.dns.DnsCnameCache;
+
+import javax.annotation.Nullable;
+
+/**
+ * {@link DnsCnameCache} implementation that does not cache CNAME entries.
+ */
+final class NoopDnsCnameCache implements DnsCnameCache {
```
NiteshKant, line 26: Can we use this implementation provided by netty itself?

normanmaurer: @NiteshKant we can... I just didn't do it because the one of netty will be `@Deprecated` in the next release... Let me know if you prefer to still use it tho.

NiteshKant: We can use the netty one and then remove this whole explicit setting of noop cache once netty is released.

normanmaurer: done

NiteshKant:
@normanmaurer this is good to merge once the conflict is resolved!

normanmaurer:
@NiteshKant this fix is sub-optimal... I working on proper fix in netty.

PR #1044, opened 2020-05-08
Release `DnsRawRecord`s after SRV response

idelpivnitskiy:
Motivation:

SRV response returns a list of `DnsRawRecord`s that are
`ByteBufHolder`s and contain reference counted content.

Modifications:

- Release `DnsRawRecord`s after processing content;

Result:

No memory leak in DNS client.

---
Follow-up PR for https://github.com/apple/servicetalk/pull/1018#discussion_r410165996

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -255,10 +255,11 @@ final class DefaultDnsClient implements DnsClient {
                                 } else {
                                     final DnsAnswer<HostAndPort> dnsAnswer;
                                     long minTTLSeconds = Long.MAX_VALUE;
-                                    List<DnsRecord> dnsRecords = null;
+                                    List<DnsRecord> toRelease = null;
                                     try {
-                                        //noinspection unchecked
-                                        dnsRecords = (List<DnsRecord>) completedFuture.getNow();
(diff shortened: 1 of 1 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -254,9 +255,10 @@ protected AbstractDnsSubscription newSubscription(
                                 } else {
                                     final DnsAnswer<HostAndPort> dnsAnswer;
                                     long minTTLSeconds = Long.MAX_VALUE;
+                                    List<DnsRecord> dnsRecords = null;
                                     try {
-                                        @SuppressWarnings("unchecked")
-                                        final List<DnsRecord> dnsRecords = (List<DnsRecord>) completedFuture.getNow();
+                                        //noinspection unchecked
```
NiteshKant, line 260: We have been favoring annotations for suppression instead of comments.

idelpivnitskiy: Annotation is allowed only when you define a new local variable. In this case, when I reassign a value, it's not allowed.

Scottmitch: Is this comment IDE specific? if so consider keeping the local variable to preserve the previous `@ SuppressWarnings` and re-assign to outer scope variable.

PR #1045, opened 2020-05-09
Consolidate concurrent.api NULL_TOKEN usage

Scottmitch:
Motivation:
SubscriberApiUtils exposes a NULL_TOKEN to the concurrent.api package
and usages share common code to wrap/unwrap the null token. This could
could be shared into methods instead of duplicated.

Modifications:
- Make SubscriberApiUtils#NULL_TOKEN private
- Add SubscriberApiUtils wrapNull and unwrapNull methods

Result:
More shared code in concurrent.api for wrapping/unwrapping NULL_TOKEN.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PubFirstOrError.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PubFirstOrError.java
@@ -20,6 +20,8 @@ import io.servicetalk.concurrent.PublisherSource;
 import java.util.NoSuchElementException;
 import javax.annotation.Nullable;
 
+import static io.servicetalk.concurrent.api.SubscriberApiUtils.wrapNull;
+
 final class PubFirstOrError<T> extends AbstractPubToSingle<T> {
 
     PubFirstOrError(Publisher<T> source) {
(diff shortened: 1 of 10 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PubFirstOrError.java
```diff
@@ -20,6 +20,8 @@
 import java.util.NoSuchElementException;
 import javax.annotation.Nullable;
 
+import static io.servicetalk.concurrent.api.SubscriberApiUtils.wrapNull;
```
idelpivnitskiy, line 23: Unused import.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SubscriberApiUtils.java
```diff
@@ -15,13 +15,25 @@
...
         // no instances
     }
+
+    static Object wrapNull(@Nullable Object o) {
+        return o == null ? NULL_TOKEN : o;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Nullable
+    static <T> T unwrapNull(Object o) {
```
NiteshKant, line 36: nit: rename to `unwrapNullUnchecked` to indicate that no attempt to verify the type is done.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PubFirstOrError.java
NiteshKant: You may have been confused because the new method name is the same as before in the parent class. This import is used.

idelpivnitskiy:
docker build issue

PR #1046, opened 2020-05-09
DelayedSubscription no longer uses ConcurrentSubscription internally

Scottmitch:
Motivation:
DelayedSubscription allocates a ConcurrentSubscription internally to
avoid concurrent access on the subscrption when setting the delayed
subscription. This introdcues additional object allocations that can be
avoided.

Modifications:
- DelayedSubscription no longer uses ConcurrentSubscription

Result:
Less object allocation overhead introduced by DelayedSubscription.

### DIFF
--- a/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
+++ b/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
@@ -18,28 +18,26 @@ package io.servicetalk.concurrent.internal;
 import io.servicetalk.concurrent.PublisherSource.Subscription;
 
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
-import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import javax.annotation.Nullable;
 
-import static io.servicetalk.concurrent.internal.ConcurrentSubscription.wrap;
-import static io.servicetalk.concurrent.internal.EmptySubscription.EMPTY_SUBSCRIPTION;
-import static io.servicetalk.concurrent.internal.SubscriberUtils.isRequestNValid;
+import static io.servicetalk.concurrent.internal.FlowControlUtils.addWithOverflowProtection;
 import static java.lang.Long.MIN_VALUE;
-import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.atomic.AtomicLongFieldUpdater.newUpdater;
 
 /**
  * A {@link Subscription} which serves as a placeholder until the "real" {@link Subscription} is available.
  */
 public final class DelayedSubscription implements Subscription {
-    private static final AtomicReferenceFieldUpdater<DelayedSubscription, Subscription> currentUpdater =
-            newUpdater(DelayedSubscription.class, Subscription.class, "current");
+    private static final long SUBSCRIPTION_SETTING = MIN_VALUE;
+    private static final long SUBSCRIPTION_SET = MIN_VALUE + 1;
+    private static final long SUBSCRIPTION_CANCEL_PENDING = MIN_VALUE + 2;
+    private static final long INVALID_GREATEST_VALUE = SUBSCRIPTION_CANCEL_PENDING;
     private static final AtomicLongFieldUpdater<DelayedSubscription> requestedUpdater =
-            AtomicLongFieldUpdater.newUpdater(DelayedSubscription.class, "requested");
+            newUpdater(DelayedSubscription.class, "requested");
 
-    @SuppressWarnings("unused")
     @Nullable
-    private volatile Subscription current;
-    @SuppressWarnings({"unused", "FieldCanBeLocal"})
+    private Subscription subscription;
     private volatile long requested;
 
     /**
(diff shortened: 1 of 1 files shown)


servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
```diff
@@ -18,84 +18,104 @@
...
 
 /**
  * A {@link Subscription} which serves as a placeholder until the "real" {@link Subscription} is available.
  */
 public final class DelayedSubscription implements Subscription {
-    private static final AtomicReferenceFieldUpdater<DelayedSubscription, Subscription> currentUpdater =
-            newUpdater(DelayedSubscription.class, Subscription.class, "current");
+    private static final long SUBSCRIPTION_SETTING = MIN_VALUE;
+    private static final long SUBSCRIPTION_SET = MIN_VALUE + 1;
+    private static final long SUBSCRIPTION_CANCEL_PENDING = MIN_VALUE + 2;
```
normanmaurer, line 34: nit: I like to use `<< 1` etc

Scottmitch: these defines are the 3 most negative numbers in succession, and not used in bit masking. there is also an optimization since they are sequential and negative most values which uses `<=` to avoid explicit checks of each value when all of them take the same code path.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
```diff
@@ -18,84 +18,104 @@
...
-            if (current != null) {
-                tryDrainRequested(current);
+        for (;;) {
+            final long prevRequested = requested;
+            if (prevRequested == SUBSCRIPTION_SET) {
+                assert subscription != null;
+                subscription.request(n);
+                break;
+            } else if (prevRequested == SUBSCRIPTION_SETTING) {
+                if (requestedUpdater.compareAndSet(this, SUBSCRIPTION_SETTING, addRequestN(0, n))) {
```
idelpivnitskiy, line 82: Considering the case:
1. Initial `requested == 0`
2. Thread 1 invokes `delayedSubscription`, does `requested = SUBSCRIPTION_SETTING`.
3. Thread 2 does `request(0)`, does `requested = -1`.
3. Thread 1 goes to the next for loop iteration and does `delayedSubscription.request(-1)` because `prevRequested == -1`.
While both values `0` and `-1` are illegal according to rule 3.9 it doesn't sound right that we change the original `n` value from `0` to `-1`. I'm not sure if this is acceptable or not.
If not, consider having another state that will represent `request(0)`: `REQUESTED_ZERO = SUBSCRIPTION_CANCEL_PENDING + 1`.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
```diff
@@ -18,84 +18,104 @@
...
-        } else {
-            tryDrainRequested(concurrentSubscription);
-
-            // Unwrap the concurrent subscription because there will be no more concurrency.
-            currentUpdater.compareAndSet(this, concurrentSubscription, delayedSubscription);
+    public boolean delayedSubscription(Subscription delayedSubscription) {
+        requireNonNull(delayedSubscription);
+        for (;;) {
+            final long prevRequested = requested;
+            if (prevRequested <= INVALID_GREATEST_VALUE) {
```
NiteshKant, line 55: I did a quick check on usages of this method and we always call it from `onSubscribe()`. Can we simply disallow calling this method twice? This will do two things:
- Fail fast for unexpected usage (we do not expect this to be called more than once)
- Simplify contract (returned `boolean` should always be `true`)

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
Scottmitch: we are overloading the `long` domain to serve multiple purposes:

1. concurrency control
2. accumulation of demand
3. representing invalid demand
4. representing cancel

The trade-off is we lose some fidelity (e.g. minimum of 3 invalid `request(n)` values) on the "representing invalid demand" in order to represent these state transition. Ideally we preserve the original `n` for debuggability but we use a similar approach in other places to reduce state and consolidate atomic operations on a single field.

We can preserve `0` via some additional mapping/unmapping if we think that is more likely to be used than some other negative number (doesn't seem unreasonable), but also don't think it is a hard requirement based upon the spec.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
Scottmitch: discussed offline. for now I'll just stick with `void` return type as `boolean` isn't strictly required. future considerations for throwing:
- more aggressive breaking change that could potentially break downstream `onSubscribe` propagation such that a subsequent `onError` is called without a prior `onSubscribe`:

```
onSubscribe(s) {
  mySubscription.delayedSubscription(s);
...
  downstreamSubscriber.onSubscribe(mySubscription);
}
onError(cause) {
  downstreamSubscriber.onError(cause); // onSubscribe wasn't called
}
```
- investigation required if `DelayedCancellable` (which is used in similar capacity for Single/Completable) should follow the same behavior (currently they are both `void`)

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
```diff
@@ -18,28 +18,26 @@
...
     private static final AtomicLongFieldUpdater<DelayedSubscription> requestedUpdater =
-            AtomicLongFieldUpdater.newUpdater(DelayedSubscription.class, "requested");
+            newUpdater(DelayedSubscription.class, "requested");
 
-    @SuppressWarnings("unused")
     @Nullable
-    private volatile Subscription current;
-    @SuppressWarnings({"unused", "FieldCanBeLocal"})
+    private Subscription subscription;
     private volatile long requested;
```
NiteshKant, line 41: If you merge `requested` and `Subscription` fields in a single `Object state` then you can simplify the state-machine and reduce fields here.

(`requested` handling may be simplified, I haven't thought about it, but this gives the idea of what I am suggesting)

```java
public final class DelayedSubscription implements Subscription {
    private static final AtomicReferenceFieldUpdater<DelayedSubscription, Object> stateUpdater =
            AtomicReferenceFieldUpdater.newUpdater(DelayedSubscription.class, Object.class, "state");
    private static final Object STATE_CANCELLED = new Object();
    private static final Object STATE_TERMINATED = new Object();
    private volatile Object state = 0L;

    /**
     * Set the delayed {@link Subscription}. This method can only be called a single time and
     * subsequent calls will result in {@link #cancel()} being called on {@code delayedSubscription}.
     *
     * @param delayedSubscription The delayed {@link Subscription}.
     */
    public void delayedSubscription(Subscription delayedSubscription) {
        requireNonNull(delayedSubscription);
        for (;;) {
            final Object cState = state;
            if (cState == STATE_TERMINATED || cState instanceof Subscription) {
                delayedSubscription.cancel();
                return;
            }
            if (cState == STATE_CANCELLED && stateUpdater.compareAndSet(this, cState, STATE_TERMINATED)) {
                delayedSubscription.cancel();
                return;
            }

            if (cState instanceof Long && stateUpdater.compareAndSet(this, cState, 0L)) {
                long requested = (Long) cState;
                if (requested > 0 || requested == MIN_VALUE) {
                    delayedSubscription.request(requested);
                }
                if (stateUpdater.compareAndSet(this, 0L, delayedSubscription)) {
                    break;
                }
            }
        }
    }

    @Override
    public void request(long n) {
        for (;;) {
            final Object cState = state;
            if (cState == STATE_CANCELLED || cState == STATE_TERMINATED) {
                return;
            }
            if (cState instanceof Subscription) {
                Subscription s = (Subscription) cState;
                s.request(n);
                return;
            }
            if (cState instanceof Long) {
                Long prevReq = (Long) cState;
                if (n > 0 &&
                        stateUpdater.compareAndSet(this, cState, addWithOverflowProtectionIfNotNegative(prevReq, n))
                        || n <= 0 && stateUpdater.compareAndSet(this, cState, MIN_VALUE)) {
                    break;
                }
            }
        }
    }

    @Override
    public void cancel() {
        for (;;) {
            final Object cState = state;
            if (cState == STATE_CANCELLED || cState == STATE_TERMINATED) {
                return;
            }
            if (cState instanceof Subscription) {
                Subscription s = (Subscription) cState;
                s.cancel();
                return;
            }
            if (cState instanceof Long && stateUpdater.compareAndSet(this, cState, STATE_CANCELLED)) {
                return;
            }
        }
    }
}
```

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
```diff
@@ -18,28 +18,26 @@
...
 /**
  * A {@link Subscription} which serves as a placeholder until the "real" {@link Subscription} is available.
  */
 public final class DelayedSubscription implements Subscription {
-    private static final AtomicReferenceFieldUpdater<DelayedSubscription, Subscription> currentUpdater =
-            newUpdater(DelayedSubscription.class, Subscription.class, "current");
+    private static final long SUBSCRIPTION_SETTING = MIN_VALUE;
+    private static final long SUBSCRIPTION_SET = MIN_VALUE + 1;
+    private static final long SUBSCRIPTION_CANCEL_PENDING = MIN_VALUE + 2;
+    private static final long INVALID_GREATEST_VALUE = SUBSCRIPTION_CANCEL_PENDING;
```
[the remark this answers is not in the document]
Scottmitch: it is the greatest value out of the invalid numbers (all control values are more negative than this).

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
Scottmitch: lets consider this alternative approach as a followup PR. the impacts of boxing/unboxing and additional object allocation need to be considered/evaluated. the motivation for this PR was avoiding intermediate allocations.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedSubscription.java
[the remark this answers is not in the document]
Scottmitch: I'll rename to `GREATEST_CONTROL_VALUE` to clarify.

PR #1049, opened 2020-05-13
Introduce HTTP Zipkin reporter

NiteshKant:
__Motivation__

We do not currently provide a zipkin reporter implementation for HTTP collectors. Providing an HTTP reporter using ServiceTalk HTTP client will be useful when ServiceTalk HTTP client is used otherwise in applications.

__Modification__

Add an `HttpReporter` with configurable batching strategy.

__Result__

Users can use ServiceTalk HTTP client to publish spans to a Zipkin collector.

### DIFF
--- a/servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/Codec.java
+++ b/servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/Codec.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright © 2020 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.opentracing.zipkin.publisher.reporter;
+
+import zipkin2.Span;
+import zipkin2.codec.SpanBytesEncoder;
+
+/**
+ * Zipkin data formats for reporting of {@link Span}s.
+ */
+public enum Codec {
+    /**
+     * Zipkin V1 JSON format.
+     */
+    JSON_V1(SpanBytesEncoder.JSON_V1),
+    /**
+     * Zipkin V2 JSON format.
+     */
+    JSON_V2(SpanBytesEncoder.JSON_V2),
+    /**
+     * Zipkin V2 THRIFT format.
+     */
+    THRIFT(SpanBytesEncoder.THRIFT),
+    /**
+     * Zipkin V2 protocol buffers V3 format.
+     */
+    PROTO3(SpanBytesEncoder.PROTO3);
+
+    private final SpanBytesEncoder encoder;
+
+    Codec(SpanBytesEncoder encoder) {
+        this.encoder = encoder;
+    }
+
+    SpanBytesEncoder spanBytesEncoder() {
+        return encoder;
+    }
+}
--- a/servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
+++ b/servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
@@ -1,0 +1,252 @@
+/*
+ * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.opentracing.zipkin.publisher.reporter;
+
+import io.servicetalk.buffer.api.Buffer;
+import io.servicetalk.buffer.api.BufferAllocator;
+import io.servicetalk.concurrent.CompletableSource;
+import io.servicetalk.concurrent.PublisherSource;
+import io.servicetalk.concurrent.api.AsyncCloseable;
+import io.servicetalk.concurrent.api.BufferStrategy.Accumulator;
+import io.servicetalk.concurrent.api.Completable;
+import io.servicetalk.concurrent.api.CompositeCloseable;
+import io.servicetalk.concurrent.api.Publisher;
+import io.servicetalk.http.api.HttpClient;
+import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import zipkin2.Component;
+import zipkin2.Span;
+import zipkin2.codec.SpanBytesEncoder;
+import zipkin2.reporter.Reporter;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nonnull;
+
+import static io.servicetalk.concurrent.api.AsyncCloseables.newCompositeCloseable;
+import static io.servicetalk.concurrent.api.AsyncCloseables.toAsyncCloseable;
+import static io.servicetalk.concurrent.api.BufferStrategies.forCountOrTime;
+import static io.servicetalk.concurrent.api.Completable.completed;
+import static io.servicetalk.concurrent.api.Processors.newCompletableProcessor;
+import static io.servicetalk.concurrent.api.Processors.newPublisherProcessorDropHeadOnOverflow;
+import static io.servicetalk.concurrent.api.SourceAdapters.fromSource;
+import static io.servicetalk.concurrent.api.SourceAdapters.toSource;
+import static io.servicetalk.concurrent.internal.FutureUtils.awaitTermination;
+import static io.servicetalk.http.api.CharSequences.newAsciiString;
+import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_TYPE;
+import static io.servicetalk.http.api.HttpHeaderValues.APPLICATION_JSON;
+import static java.time.Duration.ofSeconds;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A {@link Span} {@link Reporter} that will publish to an HTTP endpoint with a configurable encoding {@link Codec}.
+ */
+public final class HttpReporter extends Component implements Reporter<Span>, AsyncCloseable {
+    private static final Logger LOGGER = LoggerFactory.getLogger(HttpReporter.class);
+    static final String V1_PATH = "/api/v1/spans";
+    static final String V2_PATH = "/api/v2/spans";
+    static final CharSequence THRIFT_CONTENT_TYPE = newAsciiString("application/x-thrift");
+    static final CharSequence PROTO_CONTENT_TYPE = newAsciiString("application/protobuf");
+
+    private final PublisherSource.Processor<Span, Span> buffer;
+    private final AsyncCloseable closeable;
+
+    private volatile boolean closeInitiated;
+
+    private HttpReporter(final Builder builder) {
+        final HttpClient client = builder.clientBuilder.build();
+        SpanBytesEncoder spanEncoder = builder.codec.spanBytesEncoder();
+        final String path;
+        final CharSequence contentType;
+        switch (builder.codec) {
+            case JSON_V1:
+                path = V1_PATH;
+                contentType = APPLICATION_JSON;
+                break;
+            case JSON_V2:
+                path = V2_PATH;
+                contentType = APPLICATION_JSON;
+                break;
+            case THRIFT:
+                path = V2_PATH;
+                contentType = THRIFT_CONTENT_TYPE;
+                break;
+            case PROTO3:
+                path = V2_PATH;
+                contentType = PROTO_CONTENT_TYPE;
+                break;
+            default:
+                throw new IllegalArgumentException("Unknown codec: " + builder.codec);
+        }
+        final BufferAllocator allocator = client.executionContext().bufferAllocator();
+        Publisher<Buffer> spans;
+        if (builder.disableBatching) {
+            buffer = newPublisherProcessorDropHeadOnOverflow(builder.maxConcurrentReports);
+            spans = fromSource(buffer).map(span -> allocator.wrap(spanEncoder.encode(span)));
+        } else {
+            buffer = newPublisherProcessorDropHeadOnOverflow(builder.batchSizeHint * builder.maxConcurrentReports);
+            spans = fromSource(buffer)
+                    .buffer(forCountOrTime(builder.batchSizeHint, builder.maxBatchDuration,
+                            () -> new ListAccumulator(builder.batchSizeHint), client.executionContext().executor()))
+                    .filter(accumulate -> !accumulate.isEmpty())
+                    .map(bufferedSpans -> allocator.wrap(spanEncoder.encodeList(bufferedSpans)));
+        }
+
+        final CompletableSource.Processor spansTerminated = newCompletableProcessor();
+        toSource(spans.flatMapCompletable(encodedSpans -> reportSpans(client, encodedSpans, path, contentType),
+                builder.maxConcurrentReports)).subscribe(spansTerminated);
+
+        CompositeCloseable closeable = newCompositeCloseable();
+        closeable.append(toAsyncCloseable(graceful -> {
+            closeInitiated = true;
+            try {
+                buffer.onComplete();
+            } catch (Throwable t) {
+                LOGGER.error("Failed to dispose request buffer. Ignoring.", t);
+            }
+            return graceful ? fromSource(spansTerminated) : completed();
+        }));
+        closeable.append(client);
+        this.closeable = closeable;
+    }
+
+    private Completable reportSpans(final HttpClient client, final Buffer encodedSpans, final String path,
+                                    final CharSequence contentType) {
+        return client.request(client.post(path).addHeader(CONTENT_TYPE, contentType).payloadBody(encodedSpans))
+                .ignoreElement()
+                .onErrorResume(cause -> {
+                    LOGGER.error("Failed to send a span, ignoring.", cause);
+                    return completed();
+                });
+    }
+
+    /**
+     * A builder to create a new {@link HttpReporter}.
+     */
+    public static final class Builder {
+        private Codec codec = Codec.JSON_V2;
+        private final SingleAddressHttpClientBuilder<?, ?> clientBuilder;
+        private boolean disableBatching;
+        private int batchSizeHint = 16;
+        private int maxConcurrentReports = 32;
+        private Duration maxBatchDuration = ofSeconds(30);
+
+        /**
+         * Create a new {@link Builder} using the passed {@link SingleAddressHttpClientBuilder}.
+         *
+         * @param clientBuilder the collector SocketAddress
+         */
+        public Builder(final SingleAddressHttpClientBuilder<?, ?> clientBuilder) {
+            this.clientBuilder = clientBuilder;
+        }
+
+        /**
+         * Sets the {@link Codec} to encode the Spans with.
+         *
+         * @param codec the codec to use for this span.
+         * @return {@code this}
+         */
+        public Builder codec(Codec codec) {
+            this.codec = requireNonNull(codec);
+            return this;
+        }
+
+        public Builder maxConcurrentReports(final int maxConcurrentReports) {
+            this.maxConcurrentReports = maxConcurrentReports;
+            return this;
+        }
+
+        /**
+         * Configure batching of spans before sending it to the zipkin collector.
+         *
+         * @param batchSizeHint Hint of how many spans should be batched together.
+         * @param maxBatchDuration {@link Duration} of time to wait for {@code batchSizeHint} spans in a batch.
+         * @return {@code this}.
+         */
+        public Builder batchSpans(final int batchSizeHint, final Duration maxBatchDuration) {
+            if (batchSizeHint <= 0) {
+                throw new IllegalArgumentException("batchSizeHint: " + batchSizeHint + " (expected > 0)");
+            }
+            disableBatching = false;
+            this.batchSizeHint = batchSizeHint;
+            this.maxBatchDuration = requireNonNull(maxBatchDuration);
+            return this;
+        }
+
+        /**
+         * Disable batching of spans before sending them to the zipkin collector.
+         *
+         * @return {@code this}.
+         */
+        public Builder disableSpanBatching() {
+            disableBatching = true;
+            return this;
+        }
+
+        /**
+         * Builds a new {@link HttpReporter} instance with this builder's options.
+         *
+         * @return a new {@link HttpReporter}
+         */
+        public HttpReporter build() {
+            return new HttpReporter(this);
+        }
+    }
+
+    @Override
+    public void report(final Span span) {
+        if (closeInitiated) {
+            throw new IllegalStateException("Span: " + span + " reported after reporter " + this + " is closed.");
+        }
+        buffer.onNext(span);
+    }
+
+    @Override
+    public void close() {
+        awaitTermination(closeable.closeAsync().toFuture());
+    }
+
+    @Override
+    public Completable closeAsync() {
+        return closeable.closeAsync();
+    }
+
+    @Override
+    public Completable closeAsyncGracefully() {
+        return closeable.closeAsyncGracefully();
+    }
+
+    private static final class ListAccumulator implements Accumulator<Span, List<Span>> {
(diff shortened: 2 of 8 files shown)


servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/Codec.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    THRIFT(SpanBytesEncoder.THRIFT),
+    /**
+     * Zipkin V2 protocol buffers V3 format.
+     */
+    PROTO3(SpanBytesEncoder.PROTO3);
+
+    private final SpanBytesEncoder encoder;
+
+    Codec(SpanBytesEncoder encoder) {
```
normanmaurer, line 44: private ?

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,252 @@
...
+                break;
+            case PROTO3:
+                path = V2_PATH;
+                contentType = PROTO_CONTENT_TYPE;
+                break;
+            default:
+                throw new IllegalArgumentException("Unknown codec: " + builder.codec);
+        }
+        final BufferAllocator allocator = client.executionContext().bufferAllocator();
+        Publisher<Buffer> spans;
```
normanmaurer, line 98: nit: consider marking as final so we ensure it is really set at some point.

servicetalk-opentracing-zipkin-publisher/src/test/java/io/servicetalk/opentracing/zipkin/publisher/reporter/SpanUtils.java
```diff
@@ -0,0 +1,65 @@
...
+                .putTag("doubleKey", String.valueOf(Double.MAX_VALUE))
+                .addAnnotation(System.currentTimeMillis() * 1000, "some event happened")
+                .build();
+    }
+
+    static void verifySpan(final Span span) {
+        assertEquals("test operation", span.name());
+        assertEquals("0000000000001234", span.traceId());
+        assertEquals("0000000000000002", span.id());
+        assertEquals(123456789L, (long) span.timestamp());
```
normanmaurer, line 53: consider storing the value into a static final field and use it in `newSpan(...)` and `verifySpan(...)`. This way it is less error-prone once we refactor

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/Codec.java
NiteshKant: As this is an `enum` constructors are by default `private`

servicetalk-opentracing-zipkin-publisher/src/test/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporterTest.java
```diff
@@ -0,0 +1,162 @@
...
+    @Test
+    public void disableBatching() throws Exception {
+        HttpReporter reporter = initReporter(Builder::disableSpanBatching);
+        reporter.report(newSpan());
+        List<Span> spans = verifyRequest(receivedRequests.take(), false);
+        assertThat("Unexpected spans received.", spans, hasSize(1));
+        verifySpan(spans.get(0));
+        reporter.report(newSpan());
+        assertThat("Unexpected spans received.", spans, hasSize(1));
+        verifySpan(spans.get(0));
```
idelpivnitskiy, line 103: This test misses another `verifyRequest` after the second `reporter.report(newSpan())`. To prevent that, consider enhancing `newSpan` and `verifySpan` utils to compare a provided name or id. e.g.:
```
Span s = newSnap("somename");
verifySpan(s, "somename");
```

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,252 @@
...
+    public void report(final Span span) {
+        if (closeInitiated) {
+            throw new IllegalStateException("Span: " + span + " reported after reporter " + this + " is closed.");
+        }
+        buffer.onNext(span);
+    }
+
+    @Override
+    public void close() {
+        awaitTermination(closeable.closeAsync().toFuture());
```
idelpivnitskiy, line 222: Consider changing `closeable` type to `CompositeCloseable`. Then you can use `closeable.close()` which does the same.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,252 @@
...
+import static io.servicetalk.http.api.CharSequences.newAsciiString;
+import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_TYPE;
+import static io.servicetalk.http.api.HttpHeaderValues.APPLICATION_JSON;
+import static java.time.Duration.ofSeconds;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A {@link Span} {@link Reporter} that will publish to an HTTP endpoint with a configurable encoding {@link Codec}.
+ */
+public final class HttpReporter extends Component implements Reporter<Span>, AsyncCloseable {
```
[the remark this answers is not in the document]
idelpivnitskiy: The `Component` abstract class also defines `check()` method that always returns `OK`. I think we should override this method and return `CheckResult.failed(Throwable)` after this reporter is closed.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,252 @@
...
+
+    private HttpReporter(final Builder builder) {
+        final HttpClient client = builder.clientBuilder.build();
+        SpanBytesEncoder spanEncoder = builder.codec.spanBytesEncoder();
+        final String path;
+        final CharSequence contentType;
+        switch (builder.codec) {
+            case JSON_V1:
+                path = V1_PATH;
+                contentType = APPLICATION_JSON;
```
idelpivnitskiy, line 80: Consider providing path and content-type as pkg-private methods on `Codec` enum.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,252 @@
...
+    static final CharSequence PROTO_CONTENT_TYPE = newAsciiString("application/protobuf");
+
+    private final PublisherSource.Processor<Span, Span> buffer;
+    private final AsyncCloseable closeable;
+
+    private volatile boolean closeInitiated;
+
+    private HttpReporter(final Builder builder) {
+        final HttpClient client = builder.clientBuilder.build();
+        SpanBytesEncoder spanEncoder = builder.codec.spanBytesEncoder();
```
idelpivnitskiy, line 74: Consider saving reference for `builder.codec` in a final local variable to prevent concurrent state modifications between `spanEncoder` and `switch`.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,252 @@
...
+    static final CharSequence THRIFT_CONTENT_TYPE = newAsciiString("application/x-thrift");
+    static final CharSequence PROTO_CONTENT_TYPE = newAsciiString("application/protobuf");
+
+    private final PublisherSource.Processor<Span, Span> buffer;
+    private final AsyncCloseable closeable;
+
+    private volatile boolean closeInitiated;
+
+    private HttpReporter(final Builder builder) {
+        final HttpClient client = builder.clientBuilder.build();
```
idelpivnitskiy, line 73: Consider creating `CompositeCloseable` upfront and appending the client asap. Then try-catch the whole ctor and close the closeable in case of any errors.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,252 @@
...
+         *
+         * @param codec the codec to use for this span.
+         * @return {@code this}
+         */
+        public Builder codec(Codec codec) {
+            this.codec = requireNonNull(codec);
+            return this;
+        }
+
+        public Builder maxConcurrentReports(final int maxConcurrentReports) {
```
idelpivnitskiy, line 170: No javadoc on this method.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,252 @@
...
+                path = V2_PATH;
+                contentType = PROTO_CONTENT_TYPE;
+                break;
+            default:
+                throw new IllegalArgumentException("Unknown codec: " + builder.codec);
+        }
+        final BufferAllocator allocator = client.executionContext().bufferAllocator();
+        final Publisher<Buffer> spans;
+        if (builder.disableBatching) {
+            buffer = newPublisherProcessorDropHeadOnOverflow(builder.maxConcurrentReports);
```
idelpivnitskiy, line 100: When batching is disabled, does it matter what is the maximum number of items to buffer for the `PublisherProcessor`? We always subscribe and deliver items asap. Looks like in this case we can avoid having a queue or minimize its size.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,252 @@
...
+            spans = fromSource(buffer).map(span -> allocator.wrap(spanEncoder.encode(span)));
+        } else {
+            buffer = newPublisherProcessorDropHeadOnOverflow(builder.batchSizeHint * builder.maxConcurrentReports);
```
idelpivnitskiy, line 103: Can you please left a comment why we need to multiply `batchSizeHint` and `maxConcurrentReports`?

servicetalk-opentracing-zipkin-publisher/src/test/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporterTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+    @Test
+    public void batching() throws Exception {
+        HttpReporter reporter = initReporter(builder -> builder.batchSpans(2, ofMillis(200)));
+        reporter.report(newSpan());
+        reporter.report(newSpan());
+        List<Span> spans = verifyRequest(receivedRequests.take(), true);
+        assertThat("Unexpected spans received.", spans, hasSize(2));
+        verifySpan(spans.get(0));
+        verifySpan(spans.get(0));
```
idelpivnitskiy, line 114: This test verifies the same span twice.

idelpivnitskiy: Consider verifying that the second batching request also works as expected and we do not send the old values again.

servicetalk-opentracing-zipkin-publisher/src/test/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporterTest.java
NiteshKant: good catch

servicetalk-opentracing-zipkin-publisher/src/test/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporterTest.java
NiteshKant: good catch

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    @Override
+    public CheckResult check() {
+        return closeInitiated ? OK : failed(new IllegalStateException("Reporter is closed."));
+    }
+
+    @Override
+    public void report(final Span span) {
+        if (closeInitiated) {
```
Scottmitch, line 98: since this is best effort, and in practice may not be triggered, does it make sense to drop this this check to avoid providing a false sense of security (e.g. it triggers in tests, but not in real world)?

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,287 @@
...
+                });
+    }
+
+    /**
+     * A builder to create a new {@link HttpReporter}.
+     */
+    public static final class Builder {
+        private Codec codec = Codec.JSON_V2;
+        private final SingleAddressHttpClientBuilder<?, ?> clientBuilder;
+        private boolean disableBatching;
```
Scottmitch, line 193: optional: consider removing this additional state and set `batchSizeHint < 0` to indicate disabled

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,287 @@
...
+         * Builds a new {@link HttpReporter} instance with this builder's options.
+         *
+         * @return a new {@link HttpReporter}
+         */
+        public HttpReporter build() {
+            return new HttpReporter(this);
+        }
+    }
+
+    private static final class ListAccumulator implements Accumulator<Span, List<Span>> {
```
Scottmitch, line 270: `ListAccumulator` seems like a very common use case, does it make sense to make this more accessible to avoid future duplication (possibly followup PR)?

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
idelpivnitskiy: I see, thanks for explaining.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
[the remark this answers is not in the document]
idelpivnitskiy: Let's use `channel.isActive()` for `UdpReporter#check()` as we do for `UdpReporter#report(Span)`?

servicetalk-opentracing-zipkin-publisher/src/test/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporterTest.java
```diff
@@ -0,0 +1,165 @@
...
+        verifySpan(spans.get(0), "1");
+        verifySpan(spans.get(1), "2");
+    }
+
+    @Test
+    public void reportAfterClose() {
+        HttpReporter reporter = initReporter(Builder::disableSpanBatching);
+        reporter.close();
+        assertThrows("Report post close accepted.", IllegalStateException.class,
+                () -> reporter.report(newSpan("1")));
```
idelpivnitskiy, line 124: Consider also an assert for `reporter.check()` before this one.

servicetalk-opentracing-zipkin-publisher/src/test/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporterTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+    @Test
+    public void batching() throws Exception {
+        HttpReporter reporter = initReporter(builder -> builder.batchSpans(2, ofMillis(200)));
+        reporter.report(newSpan("1"));
+        reporter.report(newSpan("2"));
+        List<Span> spans = verifyRequest(receivedRequests.take(), true);
+        assertThat("Unexpected spans received.", spans, hasSize(2));
+        verifySpan(spans.get(0), "1");
+        verifySpan(spans.get(1), "2");
```
idelpivnitskiy, line 116: Let's also add a second batching report:
```java
reporter.report(newSpan("3"));
reporter.report(newSpan("4"));
List<Span> spans2 = verifyRequest(receivedRequests.take(), true);
assertThat("Unexpected spans received.", spans2, hasSize(2));
verifySpan(spans2.get(0), "3");
verifySpan(spans2.get(1), "4");
```
It will confirm that we discard the spans that are already sent.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,287 @@
...
+            case PROTO3:
+                path = V2_PATH;
+                contentType = PROTO_CONTENT_TYPE;
+                break;
+            default:
+                throw new IllegalArgumentException("Unknown codec: " + codec);
+        }
+        return encodedSpans -> client.request(
+                client.post(path).addHeader(CONTENT_TYPE, contentType).payloadBody(encodedSpans))
+                .ignoreElement()
```
idelpivnitskiy, line 180: Before we `ignoreElement`, should we log (at warn or error level) any response statuses that are not `SUCCESSFUL_2XX`?

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -0,0 +1,287 @@
...
+                .onErrorResume(cause -> {
+                    LOGGER.error("Failed to send a span, ignoring.", cause);
```
idelpivnitskiy, line 182: Consider adding a test that verifies we recover from errors.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
NiteshKant: > and in practice may not be triggered,

I think you are saying that in practice it won't be triggered when the usage is as expected, i.e. no spans are reported after the reporter is closed.  I don't think we can assume expected usage. This is a trade-off between being extremely defensive (making sure no spans can be reported after closure, even when they are concurrent) at the cost of complexity and performance or a best-effort check to provide visibility when an inadvertent misuse has happened, like the same `HttpReporter` instance is shared across multiple `ZipkinPublisher`s.

This trade-off seems reasonable to me.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
NiteshKant: In this case, the span encoder APIs are unnecessarily restrictive that they need a `List`. Our general methods return an `Iterable` which should be generally useful apart from speicific semantics of lets say a `Set`.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
NiteshKant: Since this is a builder I would prefer being explicit as opposed to overloading a field.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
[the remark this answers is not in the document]
NiteshKant: As I said I will prefer to do the `UdpReporter` in a separate PR; feel free to send a follow up.

servicetalk-opentracing-zipkin-publisher/src/test/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporterTest.java
NiteshKant: Verifying items sent to a `Processor` is consumed once or not is a `Processor` test. I do not see value in doing it here as those spans not emitted in the next request sent does not mean they will never be emitted.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
NiteshKant: Good point, lemme add a  log.

servicetalk-opentracing-zipkin-publisher/src/test/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporterTest.java
NiteshKant: Good call, there was a bug 🤕

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
Scottmitch: To clarify "best effort" was meant to highlight even if it isn't used as expected this code may not provide the desired result (e.g. throwing an error). It depends upon a happens-before relationship between closing and reporting which may or may-not exist. I'm suggesting just removing the check because it isn't required from the API contract, there is no guarantee it will throw (e.g. silently dropping data may happen anyways and should be tolerated, users can't rely upon this class detecting use-after-close, which may not be clear from code inspection or unit tests that have different lifetimes/barriers), and it introduces code into the hot path.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -177,6 +179,13 @@ public Completable closeAsyncGracefully() {
         }
         return encodedSpans -> client.request(
                 client.post(path).addHeader(CONTENT_TYPE, contentType).payloadBody(encodedSpans))
+                .beforeOnSuccess(response -> {
+                    HttpResponseStatus status = response.status();
+                    if (status.statusClass() != SUCCESSFUL_2XX) {
+                        LOGGER.info("Unexpected response from the collector. Response headers: {}",
+                                response.toString((__, headerValue) -> headerValue));
```
idelpivnitskiy, line 186: That may log any secret header values (auth tokens). Consider not logging all the headers.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
NiteshKant: I understood your suggestion. 

I think you are overlooking that this field is `volatile`?

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
NiteshKant: This is for response and there is a trade-off b/w this log being of limited usefulness and may  disclose more than necessary information. Sending secrets in a response header is somewhat less likely as compared to providing more information about the error in a header. So, I would like to keep this log here to aid debuggability.

servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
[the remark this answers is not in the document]
idelpivnitskiy: https://github.com/apple/servicetalk/pull/1054

PR #1051, opened 2020-05-13
`RoundRobinLoadBalancer` should not sort the addresses

NiteshKant:
__Motivation__

For quicker lookups on receiving Service discovery events,  `RoundRobinLoadBalancer` keeps a sorted list of addresses. This results in the same order of hosts for all client instances hence potentially creating hot spots. As, the sorting is not required functionally, we should avoid doing that.

__Modification__

Use unsorted address list and let service discoverer maintain order of addresses.

__Result__

`RoundRobinLoadBalancer` does not store sorted addresses.

### DIFF
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -424,25 +424,6 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
         }
 
         @Override
-        public boolean equals(final Object o) {
-            if (this == o) {
-                return true;
-            }
-            if (o == null || getClass() != o.getClass()) {
-                return false;
-            }
-
-            final Host<?, ?> host = (Host<?, ?>) o;
-
-            return address.equals(host.address);
-        }
-
-        @Override
-        public int hashCode() {
-            return address.hashCode();
-        }
-
-        @Override
         public String toString() {
             return "Host{" +
                     "address=" + address +
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -40,14 +40,12 @@ import org.slf4j.LoggerFactory;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.function.Predicate;
-import javax.annotation.Nullable;
 
 import static io.servicetalk.client.api.LoadBalancerReadyEvent.LOAD_BALANCER_NOT_READY_EVENT;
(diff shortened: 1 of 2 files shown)


servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -440,6 +423,25 @@ public Completable closeAsyncGracefully() {
                     .map(AsyncCloseable::closeAsyncGracefully)::iterator);
         }
 
+        @Override
+        public boolean equals(final Object o) {
```
idelpivnitskiy, line 427: Doesn't look like we started to use equals and hashCode. Can we defer adding them for now?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -161,26 +150,29 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
-                            if (event.isAvailable()) {
-                                if (i < 0) {
-                                    refreshedAddresses.add(-i - 1, new Host<>(event.address()));
-                                }
-                            } else if (i >= 0) {
-                                Host<ResolvedAddress, C> removed = refreshedAddresses.remove(i);
-                                if (removed != null) {
-                                    removed.markInactive();
+                            final List<Host<ResolvedAddress, C>> refreshedAddresses =
+                                    new ArrayList<Host<ResolvedAddress, C>>(currentAddresses);
```
Scottmitch, line 154: It is also expected that the ServiceDiscoverer will filter duplicate events, and there is no concurrent modification outside of `Subscriber#onNext`(other than emptying the list). I think we can assume we won't find any duplicates on addition:

This will also avoid ArrayList copy/move of array elements after it has already been copied in the constructor.
```java
                            final ResolvedAddress addr = requireNonNull(event.address());
                            final List<Host<ResolvedAddress, C>> refreshedAddresses;
                            if (event.isAvailable()) {
                                refreshedAddresses = new ArrayList<>(currentAddresses.size() + 1);
                                refreshedAddresses.addAll(currentAddresses);
                                refreshedAddresses.add(new Host<>(addr));
                            } else if (currentAddresses.isEmpty()) {
                                refreshedAddresses = currentAddresses;
                            } else {
                                refreshedAddresses = new ArrayList<>(currentAddresses.size() - 1);
                                for (Host<ResolvedAddress, C> host : (List<Host<ResolvedAddress, C>>) currentAddresses) {
                                        host.markInactive();
                                    } else {
                                        refreshedAddresses.add(host);
                                    }
                                }
                            }
                            return refreshedAddresses;
```

idelpivnitskiy: I was thinking about double copying when we add an element. Unfortunately, `refreshedAddresses.addAll(currentAddresses)` does 2 copies internally. So, it will be the same as using `new ArrayList(anotherList)` + `add`.

> It is expected that the ServiceDiscoverer will filter duplicate events

DNS SD currently does not filter duplicates. It takes a List from netty's `resolver.resolveAll` and passes it as-is. I didn't check netty's code but assume it just parses each DnsRecord from the response. If DNS response contains duplicated addresses, we will see them here.

NiteshKant: > I think we can assume we won't find any duplicates on addition

Actually I think we should allow duplicates which enables folks to effectively add weights to certain hosts. I would prefer not to do these two changes (unsorted and allow dups) in the same PR though. So, ya feel free to send a followup

Scottmitch: https://github.com/apple/servicetalk/pull/1052

PR #1052, opened 2020-05-13
RoundRobinLoadBalancer to allow duplicate addresses

Scottmitch:
Motivation:
RoundRobinLoadBalancer currently prevents duplicates from being
inserted. However the ServiceDiscoverer is responsible for filtering
duplicates in the typical use case. Allowing duplicates at the LB layer
allows to give more weight to specific addresses if desired.

Modifications:
- RoundRobinLoadBalancer simplify addition/removal to forgo duplicate
detection

Result:
RoundRobinLoadBalancer code simplified to ignore duplicate detection,
addition doesn't require linear search.

### DIFF
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -150,29 +150,26 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
                         event);
                 final List<Host<ResolvedAddress, C>> activeAddresses =
                         activeHostsUpdater.updateAndGet(RoundRobinLoadBalancer.this, currentAddresses -> {
-                            final List<Host<ResolvedAddress, C>> refreshedAddresses =
-                                    new ArrayList<Host<ResolvedAddress, C>>(currentAddresses);
                             final ResolvedAddress addr = requireNonNull(event.address());
-                            for (int i = 0; i < refreshedAddresses.size(); i++) {
-                                Host<ResolvedAddress, C> host = refreshedAddresses.get(i);
-                                if (host.address.equals(addr)) {
-                                    if (event.isAvailable()) {
-                                        LOGGER.debug("Address {} added but it already exists.", addr);
-                                        return currentAddresses;
-                                    } else {
-                                        refreshedAddresses.remove(i);
+                            final List<Host<ResolvedAddress, C>> refreshedAddresses;
+                            if (event.isAvailable()) {
+                                refreshedAddresses = new ArrayList<>(currentAddresses.size() + 1);
+                                refreshedAddresses.addAll(currentAddresses);
+                                refreshedAddresses.add(new Host<>(addr));
+                            } else if (currentAddresses.isEmpty()) {
+                                refreshedAddresses = currentAddresses;
+                            } else {
+                                refreshedAddresses = new ArrayList<>(currentAddresses.size() - 1);
+                                for (Host<ResolvedAddress, C> host :
+                                        (List<Host<ResolvedAddress, C>>) currentAddresses) {
+                                    if (host.address.equals(addr)) {
                                         host.markInactive();
-                                        return refreshedAddresses;
+                                    } else {
+                                        refreshedAddresses.add(host);
                                     }
                                 }
                             }
-                            if (event.isAvailable()) {
-                                refreshedAddresses.add(new Host<>(addr));
-                                return refreshedAddresses;
-                            } else {
-                                LOGGER.debug("Address {} removed but it does not exist.", addr);
-                                return currentAddresses;
-                            }
+                            return refreshedAddresses;
                         });
 
                 LOGGER.debug("Load balancer {} now using {} addresses: {}", RoundRobinLoadBalancer.this,
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -173,12 +173,12 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
                                     for (int x = i + 1; x < oldHostsTyped.size(); ++x) {
                                         newHosts.add(oldHostsTyped.get(x));
                                     }
-                                    break;
+                                    return newHosts.isEmpty() ? emptyList() : newHosts;
                                 } else {
                                     newHosts.add(host);
                                 }
                             }
-                            return newHosts.isEmpty() ? emptyList() : newHosts;
+                            return newHosts;
                         }
                     });
 
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -59,6 +59,7 @@ import static io.servicetalk.concurrent.api.Single.succeeded;
 import static io.servicetalk.concurrent.api.SourceAdapters.fromSource;
 import static io.servicetalk.concurrent.api.SourceAdapters.toSource;
 import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.atomic.AtomicIntegerFieldUpdater.newUpdater;
 import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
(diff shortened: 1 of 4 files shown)


servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -150,29 +150,26 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
-                                        refreshedAddresses.remove(i);
-                                        host.markInactive();
-                                        return refreshedAddresses;
-                                    }
-                                }
-                            }
+                            final List<Host<ResolvedAddress, C>> refreshedAddresses;
                             if (event.isAvailable()) {
+                                refreshedAddresses = new ArrayList<>(currentAddresses.size() + 1);
+                                refreshedAddresses.addAll(currentAddresses);
```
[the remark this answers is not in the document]
Scottmitch: Good observation that `addAll` does a `toArray()`. However this is an implementation detail of `addAll` which isn't strictly required. This approach avoids iteration for duplicate detection and an upfront copy and additional copy/shift for removal (at the expense of iterative additions instead of bulk copy operations). I'll add benchmarks for more insight.

Scottmitch:
```
## this pr
Benchmark                                          (ops)   Mode  Cnt        Score       Error  Units
RoundRobinLoadBalancerSDEventsBenchmark.available      5  thrpt    5  2856870.483 ± 17853.380  ops/s
RoundRobinLoadBalancerSDEventsBenchmark.available     10  thrpt    5  1644923.209 ± 26508.869  ops/s
RoundRobinLoadBalancerSDEventsBenchmark.available    100  thrpt    5   143425.550 ± 10880.633  ops/s
RoundRobinLoadBalancerSDEventsBenchmark.mixed          5  thrpt    5  2151777.842 ± 51603.588  ops/s
RoundRobinLoadBalancerSDEventsBenchmark.mixed         10  thrpt    5  1249466.523 ± 27359.906  ops/s
RoundRobinLoadBalancerSDEventsBenchmark.mixed        100  thrpt    5    55072.715 ±   530.522  ops/s 

## master
Benchmark                                          (ops)   Mode  Cnt        Score       Error  Units
RoundRobinLoadBalancerSDEventsBenchmark.available      5  thrpt    5  1809302.630 ± 47720.945  ops/s
RoundRobinLoadBalancerSDEventsBenchmark.available     10  thrpt    5   768544.812 ± 23195.555  ops/s
RoundRobinLoadBalancerSDEventsBenchmark.available    100  thrpt    5    11612.291 ±   262.869  ops/s
RoundRobinLoadBalancerSDEventsBenchmark.mixed          5  thrpt    5  1911995.403 ± 35511.624  ops/s
RoundRobinLoadBalancerSDEventsBenchmark.mixed         10  thrpt    5   810828.434 ± 10277.926  ops/s
RoundRobinLoadBalancerSDEventsBenchmark.mixed        100  thrpt    5    17890.800 ±   452.113  ops/s
```

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -150,29 +150,26 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
-                                        return currentAddresses;
-                                    } else {
-                                        refreshedAddresses.remove(i);
-                                        host.markInactive();
-                                        return refreshedAddresses;
-                                    }
-                                }
-                            }
+                            final List<Host<ResolvedAddress, C>> refreshedAddresses;
                             if (event.isAvailable()) {
```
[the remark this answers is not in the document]
Scottmitch: thx for pointing this out. I don't think we need todo duplicate detection in the LoadBalancer. the previous `binarySearch` approach was done for faster lookup and duplicate detection was a side effect.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -150,29 +150,26 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
+                                refreshedAddresses = new ArrayList<>(currentAddresses.size() + 1);
+                                refreshedAddresses.addAll(currentAddresses);
                                 refreshedAddresses.add(new Host<>(addr));
-                                return refreshedAddresses;
+                            } else if (currentAddresses.isEmpty()) {
+                                refreshedAddresses = emptyList();
                             } else {
-                                LOGGER.debug("Address {} removed but it does not exist.", addr);
```
NiteshKant, line 173: An equivalent log may still be useful to detect descrepancies of events from SD or an issue with our matching.

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
```diff
@@ -242,15 +242,6 @@ public void unknownAddressIsRemoved() {
         assertThat(lb.activeAddresses(), is(empty()));
     }
 
-    @Test
-    public void addressIsAddedTwice() {
```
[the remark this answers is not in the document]
Scottmitch: sounds good

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: I would prefer to avoid modifying control flow just for debug logs if possible (e.g. add `boolean` check it after loop). there is a debug statement before this operation for the event and after to prints all addresses which should be sufficient to detect the no-change scenario. I understand not having this log is dropping a targeted log statement, but this was the state for a while so trying to understand more context (e.g. too difficult to detect in practice, etc..).

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -150,29 +150,26 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
-                                return currentAddresses;
+                                refreshedAddresses = new ArrayList<>(currentAddresses.size() - 1);
```
idelpivnitskiy, line 162: `currentAddresses.size() - 1` may give a zero as the result. Consider rewriting in this way:
```java
if (event.isAvailable()) {
    ...
} else if (currentAddresses.size() <= 1) {
    refreshedAddresses = emptyList();
} else {
    refreshedAddresses = new ArrayList<>(currentAddresses.size() - 1);
    ...
}
```

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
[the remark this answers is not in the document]
idelpivnitskiy: sgtm!

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
idelpivnitskiy: hm, may be worth checking that addresses are equal when `currentAddresses.size() == 0`.

idelpivnitskiy: Not sure if we need to be too paranoid, but here is the version 😄 
```java
final ResolvedAddress addr = requireNonNull(event.address());
if (event.isAvailable()) {
    if (currentAddresses.isEmpty()) {
        return singletonList(new Host<>(addr));
    }
    final List<Host<ResolvedAddress, C>> refreshedAddresses =
            new ArrayList<>(currentAddresses.size() + 1);
    refreshedAddresses.addAll(currentAddresses);
    refreshedAddresses.add(new Host<>(addr));
    return refreshedAddresses;
} else {
    Host<ResolvedAddress, C> host;
    switch (currentAddresses.size()) {
        case 0:
            return emptyList();
        case 1:
            host = (Host<ResolvedAddress, C>) currentAddresses.get(0);
            return addr.equals(host.address) ? emptyList() : singletonList(host);
        case 2:
            host = (Host<ResolvedAddress, C>) currentAddresses.get(0);
            if (addr.equals(host.address)) {
                return singletonList(currentAddresses.get(1));
            }
            host = (Host<ResolvedAddress, C>) currentAddresses.get(1);
            if (addr.equals(host.address)) {
                return singletonList(currentAddresses.get(0));
            }
            return currentAddresses;
        default:
            final List<Host<ResolvedAddress, C>> refreshedAddresses =
                    new ArrayList<>(currentAddresses.size() - 1);
            for (Host<ResolvedAddress, C> hostItem :
                    (List<Host<ResolvedAddress, C>>) currentAddresses) {
                if (hostItem.address.equals(addr)) {
                    hostItem.markInactive();
                } else {
                    refreshedAddresses.add(hostItem);
                }
            }
            return refreshedAddresses;
    }
}
```

Scottmitch: this code would be a bit more complex in practice because we need to call `markInactive` when removing, and also for the removal of case 2 we need to check if both elements match bcz there maybe duplicates. I will take some of the suggestions (singleton list for adding when empty, return empty list if after removal the array is empty) but keep it simple as the list being empty isn't generally expected (e.g. we wouldn't be able to communicate with servers in that case).

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: I changed the control flow so that we now only remove the first match ... so the control flow change would be minimal to add this log statement back. however we would need need to add another log statement if the `oldHosts.isEmpty()` to cover both cases ... still not sure the additional log(s) are necessary. wdyt?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -143,37 +144,43 @@ public void onSubscribe(final Subscription s) {
...
                                 if (host.address.equals(addr)) {
-                                    if (event.isAvailable()) {
-                                        LOGGER.debug("Address {} added but it already exists.", addr);
-                                        return currentAddresses;
-                                    } else {
-                                        refreshedAddresses.remove(i);
-                                        host.markInactive();
-                                        return refreshedAddresses;
+                                    host.markInactive();
+                                    for (int x = i + 1; x < oldHostsTyped.size(); ++x) {
```
idelpivnitskiy, line 173: We said that it's ok to have duplicated addresses from SD. Neither SD, nor RRLB removes them. What is expected behavior when we receive inactive event for an address that appears multiple times in the list? Should we remove all entries or only the first one?

Scottmitch: initially this code was removing all items that matched, however I updated it last night to update just the first match. for each available event there should be a corresponding unavailable event. if duplicates are used for weighting (e.g. duplicate entries make it more likely to select) then adjusting the weighting by removing an entry shouldn't remove all the entries.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
NiteshKant: I didn't see that we already have logs at start and end of this method, so adding this log makes less sense. Disregard my comment here.

PR #1056, opened 2020-05-15
CompositeException rename finishAndThrow to transferPendingToSuppressed

Scottmitch:
Motivation:
CompositeException's finishAndThrow method implies that it directly
throws an exception. While this is possible if the internal transfer
encounters an exception, it is not the expected use case. The method
name causes confusion on first glance and requires more investigation to
understand throwing is not expected.

Modifications:
- CompositeException rename finishAndThrow to
transferPendingToSuppressed

Result:
CompositeException method name is clarified and more readable.


Scottmitch:
discussed offline with @NiteshKant

PR #1057, opened 2020-05-15
CancellableThenSubscription avoiding infinite loop

Scottmitch:
Motivation:
CancellableThenSubscription may get into an infinite loop if
setSubscription method is called twice. This is invalid usage as
described by the API but there are issues that suggest this may be
occurring [1][2][3] and if it does we should avoid infinite looping.

[1] https://github.com/apple/servicetalk/issues/744
[2] https://github.com/apple/servicetalk/issues/1040
[3] https://github.com/apple/servicetalk/issues/1053

Modifications:
- Rename CancellableThenSubscription to
DelayedCancellableThenSubscription
- Leverage DelayedCancellable and DelayedSubscription existing code to
construct DelayedCancellableThenSubscription to share code and lower
complexity

Result:
DelayedCancellableThenSubscription prevents infinite loops. This may
impact the following operators:
- Single#flatMapPublisher(..)
- Completable#concat(Publisher)
- Single#concat(Publisher)


Scottmitch:
I wasn't able to reproduce the duplicate `setSubscribe` behavior described locally for the tests in question, and created PR https://github.com/apple/servicetalk/pull/1058 to get more visibility into a potential root cause.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedCancellable.java
```diff
@@ -52,4 +52,12 @@ public void cancel() {
             oldCancellable.cancel();
         }
     }
+
+    /**
+     * Dereference the internal {@link Cancellable} and prevent future calls to
+     * {@link #delayedCancellable(Cancellable)}.
+     */
+    protected final void disableCancellable() {
```
NiteshKant, line 60: Is this useful anywhere else but for the concat case? 

Can the switch between `cancellable` and `subscription` be local to `DelayedCancellableThenSubscription`?

Scottmitch: `DelayedCancellableThenSubscription` is the only consumer of this API.

Options to accomplish dereference:
- `DelayedCancellable#cancel()` - this may involve additional offloading and complicates the control flow a bit.
- `DelayedCancellable#disableCancellable` - adds extra protected method to the API. The API
- `DelayedCancellableThenSubscription` to duplicate `DelayedCancellable` logic - logic isn't overly complicated so may not be a deal breaker, but more code duplication.

I'll update to the last option to avoid API change for review.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelayedCancellableThenSubscription.java
```diff
@@ -0,0 +1,75 @@
...
+ * An implementation of {@link Subscription} that starts as a {@link Cancellable} but then is replaced with an actual
+ * {@link Subscription}. The expected sequence of events is:
+ * <ul>
+ *     <li>{@link #delayedCancellable(Cancellable)}</li>
+ *     <li>{@link #delayedSubscription(Subscription)}</li>
+ * </ul>
+ * The {@link Subscription} methods can be invoked at any time.
+ */
+class DelayedCancellableThenSubscription implements Subscription {
+    private final DelayedSubscription delayedSubscription = new DelayedSubscription();
```
Scottmitch, line 38: I considered making `DelayedSubscription` non-`final` and extending it, but lets defer this for a followup PR if necessary.

NiteshKant: This LGTM

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelayedCancellableThenSubscription.java
```diff
@@ -0,0 +1,75 @@
...
+    private volatile Cancellable current;
+
+    @Override
+    public void request(final long n) {
+        delayedSubscription.request(n);
+    }
+
+    @Override
+    public void cancel() {
+        try {
```
NiteshKant, line 51: small followup https://github.com/apple/servicetalk/pull/1061 to just use `current` instead of `current` and `delayedSubscription`

PR #1060, opened 2020-05-19
Allow chunked responses without payload body

idelpivnitskiy:
Motivation:

Some older implementations may use connection closures as a marker of the
end of HTTP/1.1 messages if the receiver did not receive any part of the payload
body before the connection closure.

Modifications:

- Introduce `H1SpecExceptions` that allow configuring special exceptions
for HTTP/1.1 specification on `H1ProtocolConfig`;
- Add a `H1SpecExceptions#allowPrematureClosureBeforePayloadBody()` option;
- Add tests to verify that the new exception works as expected on the client side;

Result:

`HttpObjectDecoder` marks the chunked response as complete if server closes
the connection right after the headers.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -448,12 +452,16 @@ protected final void decodeLast(final ChannelHandlerContext ctx, final ByteBuf i
...
             boolean chunked = isTransferEncodingChunked(message.headers());
-            if (currentState == State.READ_VARIABLE_LENGTH_CONTENT && !in.isReadable() && !chunked) {
-                // End of connection.
-                ctx.fireChannelRead(EmptyHttpHeaders.INSTANCE);
-                closeHandler.protocolPayloadEndInbound(ctx);
-                resetNow();
-                return;
+            if (!in.isReadable()) {
+                if ((currentState == State.READ_VARIABLE_LENGTH_CONTENT && !chunked) ||
+                        (currentState == State.READ_CHUNK_SIZE && chunked && !isDecodingRequest() &&
```
NiteshKant, line 457: Why is `isDecodingRequest()` a special case here?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
```diff
@@ -130,14 +131,39 @@ public H1ProtocolConfigBuilder trailersEncodedSizeEstimate(final int trailersEnc
         return this;
     }
 
+    /**
+     * Defines if an HTTP/1.1 response with <a href="https://tools.ietf.org/html/rfc7230#section-6.1">
+     * Connection: close</a> and <a href="https://tools.ietf.org/html/rfc7230#section-3.3.1">
```
NiteshKant, line 136: Is it necessary to special case this only when `Connection: close` header is present?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
```diff
@@ -130,14 +131,39 @@ public H1ProtocolConfigBuilder trailersEncodedSizeEstimate(final int trailersEnc
...
+     * <a href="https://tools.ietf.org/html/rfc7230#section-4.1">chunked-body</a> before server closes the connection
+     * should be considered as a legit response.
+     * <p>
+     * While this use-case is not supported by <a href="https://tools.ietf.org/html/rfc7230#section-3.3.3">RFC 7230</a>,
+     * some older server implementations may use connection closure as an indicator of message completion even if
+     * {@code Transfer-Encoding: chunked} header is present:
+     * <pre>{@code
+     *     HTTP/1.1 200 OK
+     *     Content-Type: text/plain
+     *     Transfer-Encoding: chunked
```
NiteshKant, line 147: I am thinking if we need to make a special mention of `Transfer-Encoding: chunked` here? 

Can we have this option generally as `allowPrematureClosureWithNoPayload()`?

We can then generally say that if we see connection closure and no payload was received we assume there was no payload sent, unless `contentLength > 0`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
```diff
@@ -130,14 +131,39 @@ public H1ProtocolConfigBuilder trailersEncodedSizeEstimate(final int trailersEnc
...
+     *     Connection: close
+     * }</pre>
+     *
+     * @return {@code this}
+     */
+    public H1ProtocolConfigBuilder allowChunkedResponseWithoutBody() {
```
NiteshKant, line 153: It will be good if we can create a bag of spec relaxation options (expecting for more options in the future), which has the following benefits:

- Single place to enable/disable all spec relaxations (better documentation around potential dangers)
- More emphasis that these options are potentially dangerous.
- Better visibility for users when looking for unexpected behavior, eg: check if I use one method as opposed to n.

There are a few ways to approach this:

1. Use a hierarchical DSL like `clientBuilder.secure()`
2. Use an independent class like `H1SpecExceptions` which sets the flags and then provide a method here as `specExceptions(H1SpecExceptions)`
3. Use a single method like `h1SpecExceptions(boolean allowPrematureClosureWithNoPayload)` and add overloads later.

Option (2) seems better to me.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionClosedAfterResponseTest.java
```diff
@@ -0,0 +1,268 @@
...
+import static io.servicetalk.transport.netty.internal.EventLoopAwareNettyIoExecutors.toEventLoopAwareNettyIoExecutor;
+import static io.servicetalk.transport.netty.internal.GlobalExecutionContext.globalExecutionContext;
+import static java.lang.Integer.MAX_VALUE;
+import static java.nio.charset.StandardCharsets.US_ASCII;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThrows;
+
+public class ConnectionClosedAfterResponseTest {
```
NiteshKant, line 70: Add a few negative cases? 

- if headers are not complete then we do not infer empty payload.
- if content-length is present and we did not receive the entire content.
- if chunk-header is not complete then we do not infer empty payload.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionClosedAfterResponseTest.java
```diff
@@ -0,0 +1,268 @@
...
+                ch.pipeline().addLast(new HttpRequestDecoder());
+                ch.pipeline().addLast(new HttpObjectAggregator(MAX_VALUE));
+                ch.pipeline().addLast(new ChannelInboundHandlerAdapter() {
+                    @Override
+                    public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
+                        if (msg instanceof FullHttpRequest) {
+                            ctx.writeAndFlush(ByteBufUtil.writeAscii(ctx.alloc(), encodedResponse.get()))
+                                    .addListener(ChannelFutureListener.CLOSE);
+                        }
+                        ctx.fireChannelRead(msg);
```
normanmaurer, line 100: IMHO it would be more correct to not call `fireChannelRead(...)` but `ReferenceCountUtil.release(msg)` and so consume it.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionClosedAfterResponseTest.java
```diff
@@ -0,0 +1,268 @@
...
+
+        client = HttpClients.forSingleAddress(HostAndPort.of(server.localAddress()))
+                .protocols(h1().allowChunkedResponseWithoutBody().build())
+                .buildBlocking();
+    }
+
+    @After
+    public void turnDown() throws Exception {
+        try {
+            client.closeGracefully();
```
normanmaurer, line 119: Is this also tearing down the `eventLoopAwareNettyIoExecutor` ? If not we need to do this as well. That said if that's the case we should close the `server` first as otherwise the `EventLoop` will be shutdown already

idelpivnitskiy: Both client and server use `GlobalExecutionContext` that never shuts down.
Client does not take ownership of `IoExecutor`, neither for provided one nor for global.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
NiteshKant: > On the server-side, this is not expected because if a client closes the connection server can not write the response:

Response may have already been sent.

If we follow spec for spec exceptions then isn't that ironic?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
NiteshKant: We are adding an exception to spec rules, it seems we can make this more generic than to follow the spec.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
NiteshKant: > If a response has neither Content-Length nor Transfer-Encoding: chunked this is a legit use-case

Yes I do understand that it is a valid use-case and we support it today. My suggestion is to be less specific when describing the nature of this flag (either in name or javadoc).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfig.java
```diff
@@ -87,4 +87,13 @@ default String alpnId() {
      * <a href="https://tools.ietf.org/html/rfc7230#section-4.1.2trailers">trailer fields</a>
      */
     int trailersEncodedSizeEstimate();
+
+    /**
+     * Additional extensions for <a href="https://tools.ietf.org/html/rfc7230">HTTP/1.1</a> specification that help to
```
NiteshKant, line 92:
```suggestion
     * Additional exceptions for <a href="https://tools.ietf.org/html/rfc7230">HTTP/1.1</a> specification that help to
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfig.java
```diff
@@ -87,4 +87,13 @@ default String alpnId() {
...
+     * relax constrains for backward compatibility with older systems.
```
NiteshKant, line 93: I would suggest to remove the words after `specification` above. "exceptions for specifications" is sufficient.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
```diff
@@ -29,12 +29,15 @@
  */
 public final class H1ProtocolConfigBuilder {
 
+    private static final H1SpecExceptions DEFAULT_H1_SPEC_EXTENSIONS = new H1SpecExceptionsBuilder().build();
```
NiteshKant, line 32: `EXTENSIONS` -> `EXCEPTIONS`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1SpecExceptions.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+/**
+ * Additional extensions for <a href="https://tools.ietf.org/html/rfc7230">HTTP/1.1</a> specification that help to
+ * relax constrains for backward compatibility with older systems.
+ */
+public interface H1SpecExceptions {
```
NiteshKant, line 22: Since this is just a collection of flags and no behavior, I would suggest to make this a concrete class with a private constructor and have the builder as an inner class. It helps to reduce the level of indirections to create an instance.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
idelpivnitskiy: > A client that sends a "close" connection option
   MUST close the connection after reading the final response message
   corresponding to this request.

> A server that sends a "close" connection option MUST initiate a close
   of the connection (see below) after it sends the response containing
   "close".

Server-side is a bit more involved. Because closure initiated by the server may be concurrent with closure initiated by the client, decoder can not correctly understand which side closes the connection. It requires more work and additional testing for this PR to make sure we do not behave incorrectly.

Let's limit the scope of this PR and defer server-side changes until we know the use-case.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
idelpivnitskiy: The question: is it worth making it more generic before we know the use-case? I would prefer to keep the scope of each exception minimal as broader scope of not following RFC may introduce undesired behavior.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
idelpivnitskiy: As in the comment above, I will prefer to see exceptions to be specific.
Since other use-cases are legit and do not require this exception, it will confuse users and some users will enable this exception without need for that.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
idelpivnitskiy: Discussed offline, we will make it work for client and server decoders. Tests for the server side will be added later.
Done in https://github.com/apple/servicetalk/pull/1060/commits/64fb4cc1ab4a8c3316a094560bd531dd236615f3

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
idelpivnitskiy: Discussed offline and decided to remove requirement for the presence of `Connection: close` header.
Fixed in https://github.com/apple/servicetalk/pull/1060/commits/64fb4cc1ab4a8c3316a094560bd531dd236615f3

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
idelpivnitskiy: Discussed offline and decided to avoid mentioning chunked encoding in the API and javadoc to simplify understanding of the exception.
Done in https://github.com/apple/servicetalk/pull/1060/commits/64fb4cc1ab4a8c3316a094560bd531dd236615f3

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1SpecExceptions.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private final boolean allowPrematureClosureBeforePayloadBody;
+
+    H1SpecExceptions(final boolean allowPrematureClosureBeforePayloadBody) {
+        this.allowPrematureClosureBeforePayloadBody = allowPrematureClosureBeforePayloadBody;
+    }
+
+    /**
+     * Allows interpreting a premature connection closures as the end of HTTP/1.1 messages if a receiver has not started
+     * to read the payload body yet.
```
NiteshKant, line 31:
```suggestion
     * Allows interpreting connection closures as the end of HTTP/1.1 messages if the receiver did not receive any part of the payload body before the connection closure.
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1SpecExceptions.java
```diff
@@ -0,0 +1,64 @@
...
+package io.servicetalk.http.netty;
+
+/**
+ * Additional exceptions for <a href="https://tools.ietf.org/html/rfc7230">HTTP/1.1</a> specification.
+ */
+public final class H1SpecExceptions {
+
+    private final boolean allowPrematureClosureBeforePayloadBody;
+
+    H1SpecExceptions(final boolean allowPrematureClosureBeforePayloadBody) {
```
NiteshKant, line 25: Make this `private`?

PR #1061, opened 2020-05-20
Minor improvement for `DelayedCancellableThenSubscription`

NiteshKant:
__Motivation__

`DelayedCancellableThenSubscription` can just use `current` inside `cancel()` and reduce best-effort set in `delayedSubscription()`

__Modification__

Reduce `cancel()` implementation to only cancel `current`

__Result__

More typical implementation when handling sequential entities.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelayedCancellableThenSubscription.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelayedCancellableThenSubscription.java
@@ -35,9 +35,11 @@ import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater
  * The {@link Subscription} methods can be invoked at any time.
  */
 class DelayedCancellableThenSubscription implements Subscription {
-    private final DelayedSubscription delayedSubscription = new DelayedSubscription();
+    private static final Cancellable CANCELLED = () -> { };
     private static final AtomicReferenceFieldUpdater<DelayedCancellableThenSubscription, Cancellable> currentUpdater =
             newUpdater(DelayedCancellableThenSubscription.class, Cancellable.class, "current");
+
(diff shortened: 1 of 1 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelayedCancellableThenSubscription.java
```diff
@@ -65,11 +63,11 @@ final void delayedCancellable(Cancellable delayedCancellable) {
     }
 
     final void delayedSubscription(Subscription subscription) {
-        // The operation corresponding to the first cancellable is considered done, so we dereference the Cancellable.
-        // Best effort is OK as subsequent calls to cancel should be NOOPs [1][2].
-        // [1] https://github.com/reactive-streams/reactive-streams-jvm#2.4
-        // [2] https://github.com/reactive-streams/reactive-streams-jvm#3.7
-        current = IGNORE_CANCEL;
-        delayedSubscription.delayedSubscription(subscription);
+        if (currentUpdater.getAndAccumulate(this, delayedSubscription,
```
Scottmitch, line 66: I'm not convinced we need the additional atomics on top of `DelayedSubscription`. IIUC the best effort semantics should be good enough, or does this cause a problem? This approach adds additional atomics on the non-cancel path which I assume would be more common (if we are completing work).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelayedCancellableThenSubscription.java
```diff
@@ -35,9 +35,11 @@
  * The {@link Subscription} methods can be invoked at any time.
  */
 class DelayedCancellableThenSubscription implements Subscription {
```
normanmaurer, line 37: did you consider extending `DelayedSubscription ` ?

PR #1063, opened 2020-05-21
AbstractInMemorySpanBuilder to respect FOLLOWS_FROM

Scottmitch:
Motivation:
OpenTracing semantics defines a FOLLOWS_FROM relationship which is an
alternative means of representing a parent relationship. However
AbstractInMemorySpanBuilder doesn't respect this Reference type and only
recognizes CHILD_OF.

Modifications:
- AbstractInMemorySpanBuilder#parent() should also match References
against FOLLOWS_FROM

Result:
AbstractInMemorySpanBuilder#parent() respects the FOLLOWS_FROM
relationship.

### DIFF
--- a/servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/AbstractInMemorySpanBuilder.java
+++ b/servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/AbstractInMemorySpanBuilder.java
@@ -20,7 +20,6 @@ import io.servicetalk.opentracing.inmemory.api.InMemorySpan;
 import io.servicetalk.opentracing.inmemory.api.InMemorySpanBuilder;
 import io.servicetalk.opentracing.inmemory.api.InMemorySpanContext;
 
-import io.opentracing.References;
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.tag.Tags;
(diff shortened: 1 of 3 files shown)


servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/AbstractInMemorySpanBuilder.java
```diff
@@ -52,13 +53,13 @@ protected AbstractInMemorySpanBuilder(String operationName, int maxTagSize) {
...
     public final InMemorySpanBuilder asChildOf(SpanContext parent) {
-        addReference(References.CHILD_OF, parent);
+        addReference(CHILD_OF, parent);
         return this;
     }
 
     @Override
     public final InMemorySpanBuilder asChildOf(Span parent) {
-        addReference(References.CHILD_OF, parent.context());
+        addReference(CHILD_OF, parent.context());
```
idelpivnitskiy, line 62: Do we need similar public methods to simplify adding `FOLLOWS_FROM` references?

idelpivnitskiy:
Pulling this in for the release

servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/AbstractInMemorySpanBuilder.java
Scottmitch: This method is overriden from `io.opentracing.SpanBuilder` to refine/specialize the return type. I don't think it is necessary to add additional methods for `followsFrom` but we can do this in a followup PR if necessary.

PR #1069, opened 2020-05-22
Update jackson-databind 2.10.3 -> 2.10.4

Scottmitch:


gradle.properties
```diff
@@ -42,7 +42,7 @@ mockitoCoreVersion=2.28.2
 
 reactiveStreamsVersion=1.0.3
 jcToolsVersion=3.0.0
-jacksonVersion=2.10.3
+jacksonVersion=2.10.4
```
idelpivnitskiy, line 45: The latest is 2.11.0, do you prefer to hold upgrading to it?

Scottmitch: for now lets stay on 2.10, but we should move to 2.11 soon (https://github.com/FasterXML/jackson/wiki/Jackson-Releases)

PR #1070, opened 2020-05-27
Fail requests if ServiceDiscoverer emits an error

idelpivnitskiy:
Motivation:

By default, DNS lookup failures are retried by `DefaultDnsServiceDiscoverer`.
This has a downside that if a DNS name is wrong, requests will wait forever
with no indication that the DNS name is wrong. Although, treating DNS failures
as transient is mostly appropriate, lack of visibility of these failures in
the request path is confusing.

Modifications:

- Disable retry on `UnknownHostException` for `DefaultDnsServiceDiscoverer`;
- Retry all SD failures on the HTTP client side by default;
- Provide `HttpClientBuilder` API to configure a custom retry strategy for SD;
- Default `AutoRetryStrategy` stops waiting for LB ready event if SD emits
an error;
- Allow users to configure `DefaultAutoRetryStrategyProvider` to ignore
errors from SD while it waits for LB ready event;
- Add tests for new behavior;

Result:

Request fails if `ServiceDiscoverer` emits an error. Users can configure
`AutoRetryStrategyProvider` to ignore those errors.
HTTP client retries `UnknownHostException`s. Users can apply a filter for
`ServiceDiscoverer` to alter the retry strategy.


servicetalk-client-api/src/main/java/io/servicetalk/client/api/DefaultAutoRetryStrategyProvider.java
```diff
@@ -70,6 +77,19 @@ public Builder disableWaitForLoadBalancer() {
...
 
+        /**
+         * By default, automatic retries waits for the associated {@link LoadBalancer} to be ready or
+         * {@link ServiceDiscoverer} to emit an error before triggering a retry for requests. This method allows a retry
+         * strategy to ignore errors from {@link ServiceDiscoverer} and wait for {@link LoadBalancer} forever.
+         *
+         * @return {@code this}.
+         * @see #disableWaitForLoadBalancer()
+         */
+        public Builder ignoreSdErrors() {
```
NiteshKant, line 88: Better not to use short-forms in method names => `ignoreServiceDiscovererErrors()`

servicetalk-client-api/src/main/java/io/servicetalk/client/api/AutoRetryStrategyProvider.java
```diff
@@ -32,15 +33,16 @@
...
     /**
      * Create a new {@link AutoRetryStrategy} instance using the passed {@link LoadBalancer}.
      *
-     * @param loadBalancer {@link LoadBalancer} to use.
+     * @param lbEventStream a stream of events from {@link LoadBalancer#eventStream() LoadBalancer}
+     * @param sdErrorStream a stream of errors from {@link ServiceDiscoverer#discover(Object) ServiceDiscoverer}
      * @return New {@link AutoRetryStrategy} instance.
      */
-    AutoRetryStrategy forLoadbalancer(LoadBalancer<?> loadBalancer);
+    AutoRetryStrategy forClient(Publisher<Object> lbEventStream, Publisher<Throwable> sdErrorStream);
```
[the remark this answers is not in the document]
idelpivnitskiy: I was thinking between `Publisher` and `Completable` during the implementation and decided to go with a `Publisher` to make it work in a similar way as LB events: each request waits for either an LB-ready event or a _new_ SD error.
If you think it's enough to wait only for the first error from SD and subsequent requests should fail fast, then I can switch to `Completable`. Both ways work for me.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -156,12 +155,13 @@ public DefaultDnsServiceDiscovererBuilder dnsResolverAddressTypes(
...
     /**
-     * Do not perform retries if DNS lookup fails. Instead, terminate the {@link Publisher} with the error.
+     * Perform retries if DNS lookup fails instead of terminating the
+     * {@link ServiceDiscoverer#discover(Object) discovery} with an error.
      *
      * @return {@code this}.
      */
-    public DefaultDnsServiceDiscovererBuilder noRetriesOnDnsFailures() {
-        this.applyRetryFilter = false;
+    public DefaultDnsServiceDiscovererBuilder retryOnDnsFailures() {
```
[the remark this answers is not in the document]
idelpivnitskiy: Was going to ask if we need this at all. If not, we can remove this method, `DnsClientFilterFactory, and `RetryingDnsClientFilter`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -219,6 +228,33 @@ private DefaultSingleAddressHttpClientBuilder(@Nullable final U address,
                     proxyAddress != null ? proxyAddress : builder.address);
         }
 
+        Publisher<? extends ServiceDiscovererEvent<R>> discover(
+                PublisherSource.Processor<Throwable, Throwable> processor) {
+            assert builder.address != null : "Attempted to buildStreaming with an unknown address";
+            return builder.serviceDiscoverer.discover(
+                    proxyAddress != null ? proxyAddress : builder.address)
+                    .retryWhen(retryWithConstantBackoffAndJitter(MAX_VALUE, t -> {
```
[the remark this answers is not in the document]
idelpivnitskiy: How about starting without providing configurations for this and deferring this API until we have a request? I didn't find a good place for this config, not sure if a client builder or auto-retry strategy provider builder are good for this.

It looks ok to me to always retry `java.net.UnknownHostException`, because this is a well-known JDK exception and not-retrying it means that the discovery publisher will terminate. Therefore, the entire client becomes useless. This is a bit separate discussion, but maybe we need to close the client if SD terminates to clear resources.

For now, users can wrap SD to apply a different retry strategy or map `java.net.UnknownHostException` into something else if their intent is to terminate the discovery publisher.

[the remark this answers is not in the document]
NiteshKant: > How about starting without providing configurations for this and deferring this API until we have a request from users? 

If we are using some default strategy that depends on an externally provided implementation (service discoverer in this case), we should always assume there is a case which needs customization. As an example; what if the retry duration has to be changed?

> I didn't find a good place for this config, not sure if a client builder or auto-retry strategy provider builder are good for this.

ServiceDiscoverer retry is not really tied to auto-retry so both of them having some configuration around SD errors. How about adding the method `retryServiceDiscoveryErrors(BiIntFunction<Throwable, ? extends Completable>)`  on the client builder?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -219,6 +228,33 @@ private DefaultSingleAddressHttpClientBuilder(@Nullable final U address,
...
+                        processor.onNext(t);
+                        return t instanceof UnknownHostException;
```
NiteshKant, line 238: Why not retry all SD errors by default?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -233,16 +221,11 @@ public DefaultDnsServiceDiscovererBuilder ioExecutor(final IoExecutor ioExecutor
      * @return a new instance of {@link DnsClient}.
      */
     DnsClient build() {
-        DnsClient rawClient = new DefaultDnsClient(
+        final DnsClient rawClient = new DefaultDnsClient(
                 ioExecutor == null ? globalExecutionContext().ioExecutor() : ioExecutor, minTTLSeconds, ndots,
                 invalidateHostsOnDnsFailure, optResourceEnabled, queryTimeout, dnsResolverAddressTypes,
                 dnsServerAddressStreamProvider);
-        DnsClientFilterFactory rawFilterFactory = filterFactory;
-        if (applyRetryFilter) {
```
NiteshKant, line 241: As we do not use `RetryingDNSClientFilter` anymore, we can delete the class and remove references from the tests

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
idelpivnitskiy: Just to keep existing behavior as-is. But if we will have `retryServiceDiscoveryErrors` on the client builder +1 for retry all by default. Will update.

servicetalk-client-api/src/test/java/io/servicetalk/client/api/DefaultAutoRetryStrategyProviderTest.java
```diff
@@ -119,6 +128,46 @@ public void defaultForNoAvailableHost() {
...
+
+    @Test
+    public void defaultForNoAvailableHostWhenServiceDiscovererTerminated() {
+        AutoRetryStrategy strategy = newStrategy(identity());
+        sdErrors.onComplete();
+        Completable retry = strategy.apply(1, NO_AVAILABLE_HOST);
+        toSource(retry).subscribe(retrySubscriber);
+        TerminalNotification terminal = retrySubscriber.takeTerminal();
+        assertThat("Unexpected terminal.", terminal, is(notNullValue()));
+        assertThat("Unexpected terminal.", terminal.cause(), is(instanceOf(IllegalStateException.class)));
```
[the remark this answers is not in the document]
idelpivnitskiy: Not for this PR, but for the potential follow-up:
Agreed that if SD terminates we should rely on LB events to handle further retries. But it looks like we should return a `Completable.failed(new IllegalStateException("LB events stream is terminated"))` here [1] if we still see `NoAvailableHostException` after termination. WDYT?

[1] https://github.com/apple/servicetalk/blob/5e83430adf69174d66dc06fc306522970161b5cd/servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerReadySubscriber.java#L47

servicetalk-client-api/src/main/java/io/servicetalk/client/api/AutoRetryStrategyProvider.java
```diff
@@ -32,15 +33,18 @@
...
+    AutoRetryStrategyProvider DISABLE_AUTO_RETRIES = (lbEventStream, sdErrorStream) -> (___, cause) -> failed(cause);
 
     /**
-     * Create a new {@link AutoRetryStrategy} instance using the passed {@link LoadBalancer}.
+     * Creates a new {@link AutoRetryStrategy} instance.
      *
-     * @param loadBalancer {@link LoadBalancer} to use.
+     * @param lbEventStream a stream of events from {@link LoadBalancer#eventStream() LoadBalancer}
+     * @param sdStatus a {@link Completable} that reports current {@link ServiceDiscoverer} status. It fails when
+     * {@link ServiceDiscoverer#discover(Object) discovery publisher} emits an error or completes when the publisher
```
NiteshKant, line 43: As we do not expect SD streams to complete, can we not mention the completion contract here?

If we complete the `Completable` when SD completes, any request waiting may prematurely be retried before LB being ready.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DefaultAutoRetryStrategyProvider.java
```diff
@@ -70,6 +76,19 @@ public Builder disableWaitForLoadBalancer() {
             return this;
         }
 
+        /**
+         * By default, automatic retries wait for the associated {@link LoadBalancer} to be ready or
```
NiteshKant, line 80: It would be useful to not cross-reference methods here as it is correct but more complex to understand. 

Simply, you can say that auto-retry strategies fail a request if the last signal from the associated `ServiceDiscoverer` was an error. This method disables that behavior.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuilder.java
```diff
@@ -181,6 +184,16 @@
     public abstract HttpClientBuilder<U, R, SDE> serviceDiscoverer(
             ServiceDiscoverer<U, R, ? extends SDE> serviceDiscoverer);
 
+    /**
+     * Sets a retry strategy for errors emitted by {@link ServiceDiscoverer#discover(Object) discovery publisher}.
```
NiteshKant, line 188: `for errors ...` => `to retry errors`

`discovery publisher` => `service discoverer`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -552,4 +602,52 @@ public Completable closeAsyncGracefully() {
             return closeable.closeAsyncGracefully();
         }
     }
+
+    private static final class SdStatusCompletable extends Completable implements CompletableSource.Processor {
```
NiteshKant, line 606: Do we need the complete `Processor` contract here? 

If we are just using `onComplet()` and `onError()` then not using the `Processor` contract helps remove the extra `subscribe()` method and delegation to `subscribeInternal()`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -552,4 +602,52 @@ public Completable closeAsyncGracefully() {
...
+        private static final AtomicReferenceFieldUpdater<SdStatusCompletable, CompletableSource.Processor>
+                currentProcessorUpdater = AtomicReferenceFieldUpdater.newUpdater(SdStatusCompletable.class,
+                CompletableSource.Processor.class, "currentProcessor");
+        @Nullable
+        private volatile CompletableSource.Processor currentProcessor;
+
+        @Override
+        protected void handleSubscribe(final CompletableSource.Subscriber subscriber) {
+            final CompletableSource.Processor currentProcessor = currentProcessorUpdater.updateAndGet(this,
```
NiteshKant, line 615: This approach is incorrect as it misses errors when `nextError()` happens first and then never happens again.

Instead what you need is to always have a `Processor` which queues the error till a subscriber arrives.

Once the error is queued, it needs a callback to clear the queued error once a successful event happens on the SD

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -219,6 +234,32 @@ private DefaultSingleAddressHttpClientBuilder(@Nullable final U address,
...
+            final BiIntFunction<Throwable, ? extends Completable> retryWhen = builder.serviceDiscovererRetryStrategy;
+            return builder.serviceDiscoverer.discover(proxyAddress != null ? proxyAddress : builder.address)
+                    .retryWhen((i, t) -> {
+                        sdStatus.nextError(t);
+                        return retryWhen.apply(i, t);
+                    })
+                    .whenFinally(new TerminalSignalConsumer() {
+                        @Override
+                        public void onComplete() {
+                            sdStatus.onComplete();
```
NiteshKant, line 248: As mentioned in the `AutoRetryStrategy` comment, we should not send an `onComplete()` here as it may prematurely retry a request. Instead let the LB decide when to retry if SD completes.

normanmaurer:
@idelpivnitskiy just as a side note you may want to retry for an `UnknownHostException` after some "ttl" again. That is also how we handle caching of such errors on the netty level.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -552,4 +587,32 @@ public Completable closeAsyncGracefully() {
...
+        @Override
+        protected void handleSubscribe(final Subscriber subscriber) {
+            processor.subscribe(subscriber);
+        }
+
+        void nextError(final Throwable t) {
+            final CompletableSource.Processor newProcessor = newCompletableProcessor();
+            newProcessor.onError(t);
+            final CompletableSource.Processor oldProcessor = processorUpdater.getAndSet(this, newProcessor);
+            oldProcessor.onError(t);
```
NiteshKant, line 607: nit: inline `oldProcessor` 
```suggestion
            processorUpdater.getAndSet(this, newProcessor).onError(t);
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -552,4 +587,32 @@ public Completable closeAsyncGracefully() {
...
+            seenError = true;
```
NiteshKant, line 608: set this state before setting errors on the processors for weird re-entrancy back to `resetError()`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -552,4 +587,32 @@ public Completable closeAsyncGracefully() {
             return closeable.closeAsyncGracefully();
         }
     }
+
+    private static final class SdStatusCompletable extends Completable {
+        private static final AtomicReferenceFieldUpdater<SdStatusCompletable, CompletableSource.Processor>
```
NiteshKant, line 592: As the updates to `processor ` (`nextError` and `resetError`) aren't concurrent, we do not need an atomic updater.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -552,4 +587,32 @@ public Completable closeAsyncGracefully() {
...
+                processorUpdater = newUpdater(SdStatusCompletable.class, CompletableSource.Processor.class,
+                "processor");
+        private volatile CompletableSource.Processor processor = newCompletableProcessor();
+        private boolean seenError;
```
NiteshKant, line 596: Since this field isn't `volatile`, just add a comment that this is only accessed from `nextError` and `resetError` which are not concurrent.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -552,4 +587,32 @@ public Completable closeAsyncGracefully() {
...
+
+        @Override
+        protected void handleSubscribe(final Subscriber subscriber) {
+            processor.subscribe(subscriber);
+        }
+
+        void nextError(final Throwable t) {
+            final CompletableSource.Processor newProcessor = newCompletableProcessor();
+            newProcessor.onError(t);
```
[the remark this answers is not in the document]
idelpivnitskiy: Because we expect any `Throwable` here, we use it as a way to update the state and always deliver the **latest** observed `t` to subscribers.
Since this is not on the hot path, it's ok to use this approach vs duplicating support for multiple subscribers from `CompletableProcessor`.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/AutoRetryStrategyProvider.java
```diff
@@ -32,15 +33,16 @@
...
-    AutoRetryStrategyProvider DISABLE_AUTO_RETRIES = __ -> (___, cause) -> failed(cause);
+    AutoRetryStrategyProvider DISABLE_AUTO_RETRIES = (lbEventStream, sdErrorStream) -> (___, cause) -> failed(cause);
 
     /**
-     * Create a new {@link AutoRetryStrategy} instance using the passed {@link LoadBalancer}.
+     * Creates a new {@link AutoRetryStrategy} instance.
      *
-     * @param loadBalancer {@link LoadBalancer} to use.
+     * @param lbEventStream a stream of events from {@link LoadBalancer#eventStream() LoadBalancer}
+     * @param sdStatus a {@link Completable} that reports {@link ServiceDiscoverer} status
```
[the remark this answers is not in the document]
idelpivnitskiy: Good question,
Let's discuss this offline and consider as a follow-up if necessary.
`AutoRetryStrategyProvider` is currently used in `http-api` and `grpc-api`. Not easy to move it to `http-netty`.

PR #1073, opened 2020-05-29
Clarify exception when payload body can not be aggregated

idelpivnitskiy:
Motivation:

Aggregation of the payload body may fail due to some reasons.
For example, when data size does not fit into `Integer.MAX_VALUE`
netty throws `IllegalArgumentException` from the `CompositeByteBuf`.
To simplify understanding of the issue we need to wrap the cause
into a more meaningful exception type for this context.

Motivation:
- Throw `BufferOverflowException` with the original cause;

Result:

Better debuggabillity experience for users of aggregated API.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpDataSourceTransformations.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpDataSourceTransformations.java
@@ -412,14 +412,18 @@ final class HttpDataSourceTransformations {
                                                                   BufferAllocator allocator) {
         return payloadAndTrailers.collect(PayloadAndTrailers::new, (pair, nextItem) -> {
             if (nextItem instanceof Buffer) {
-                Buffer buffer = (Buffer) nextItem;
-                if (pair.payload == null) {
-                    pair.payload = buffer;
-                } else if (pair.payload instanceof CompositeBuffer) {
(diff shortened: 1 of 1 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpDataSourceTransformations.java
```diff
@@ -433,4 +437,12 @@ static RuntimeException newTrailersExpectedButNotSeenException() {
             }
         });
     }
+
+    private static final class AggregationException extends IllegalStateException {
```
NiteshKant, line 441: I question the value of this change after https://github.com/netty/netty/pull/10330 is available. The only benefit I see is that the name of the exception is different in the logs. As the class is private, users can not use the exception to decide further actions. Is that what your intent is? 

If we create a publicly accessible exception class like `TooLongHttpPayloadException` then the change makes more sense to me.

NiteshKant: > The confusion was that IllegalArgumentException is thrown when users don't pass any argument. A more clear instance of IllegalStateException makes more sense in this case.

The benefit is trivial IMO.

> I'm ok with making the exception type public.

My suggestion is to narrow the scope of the exception (only when there is a buffer overflow) and then make it public

PR #1074, opened 2020-05-29
Avoid blocking JDK DNS resolutions

idelpivnitskiy:
Motivation:

Users of `HttpClients.forResolvedAddress(HostAndPort)` API could end up
using blocking JDK DNS resolution if they pass unresolved hostname.

Modification:

- Add `BuilderUtils.toResolvedInetSocketAddress(HostAndPort)` that creates
a resolved `InetSocketAddress` without DNS resolution;
- Add tests for new utility method;
- Use this new utility for `HttpClients. forResolvedAddress` instead of
`new java.net.InetSocketAddress`;

Result:

Users who pass unresolved hostname will see `IllegalArgumentException`
with `UnknownHostException` cause and should use `HttpClients.forAddress`
instead that performs async DNS resolutions using netty-dns-resolver.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
@@ -34,7 +34,7 @@ import java.net.InetSocketAddress;
 import java.util.function.Function;
 
 import static io.servicetalk.http.netty.DefaultSingleAddressHttpClientBuilder.forUnknownHostAndPort;
-import static java.net.InetSocketAddress.createUnresolved;
+import static io.servicetalk.transport.netty.internal.BuilderUtils.toInetSocketAddress;
 
 /**
  * Factory methods for building {@link HttpClient} (and other API variations) instances.
@@ -204,8 +204,7 @@ public final class HttpClients {
      */
     public static SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> forResolvedAddress(
             final HostAndPort address) {
-        return DefaultSingleAddressHttpClientBuilder.forResolvedAddress(address,
-                createUnresolved(address.hostName(), address.port()));
+        return DefaultSingleAddressHttpClientBuilder.forResolvedAddress(address, toInetSocketAddress(address));
     }
 
     /**
@@ -222,7 +221,7 @@ public final class HttpClients {
     public static SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> forResolvedAddressViaProxy(
             final HostAndPort address, final HostAndPort proxyAddress) {
         return DefaultSingleAddressHttpClientBuilder.forResolvedAddressViaProxy(address,
-                createUnresolved(address.hostName(), address.port()), proxyAddress);
+                toInetSocketAddress(address), proxyAddress);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
```diff
@@ -221,7 +221,7 @@ private HttpClients() {
     public static SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> forResolvedAddressViaProxy(
             final HostAndPort address, final HostAndPort proxyAddress) {
         return DefaultSingleAddressHttpClientBuilder.forResolvedAddressViaProxy(address,
-                new InetSocketAddress(address.hostName(), address.port()), proxyAddress);
+                toInetSocketAddress(address), proxyAddress);
```
idelpivnitskiy, line 224: @Scottmitch `createUnresolved` doesn't work when you pass an IP-address. I had to create a utility for this conversion.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/BuilderUtils.java
```diff
@@ -158,12 +163,27 @@ public static SocketAddress toNettyAddress(Object address) {
...
         throw new IllegalArgumentException("Unsupported address: " + address);
     }
 
+    /**
+     * Converts {@link HostAndPort} that contains a resolved address into {@link InetSocketAddress}.
+     *
+     * @param resolvedAddress the resolved address to convert.
+     * @return {@link InetSocketAddress} from the passed resolved {@link HostAndPort}.
+     */
+    public static InetSocketAddress toInetSocketAddress(final HostAndPort resolvedAddress) {
```
NiteshKant, line 177: nit: Rename to `toUnresolvedAddress()` to indicate it is unresolved?

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/BuilderUtilsTest.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+public class BuilderUtilsTest {
+
+    @Test
+    @SuppressWarnings("PMD.AvoidUsingHardCodedIP")
+    public void toInetSocketAddressFromIPv4() {
+        InetSocketAddress address = toInetSocketAddress(HostAndPort.of("192.168.1.1", 8080));
```
NiteshKant, line 35: Use loopback IPs instead of some specific IPs here and below?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/BuilderUtils.java
idelpivnitskiy: Actually, it's a resolved address, we expect an IP instead of the host. `toUnresolvedAddress` name is misleading in the same way as `InetSocketAddress.createUnresolved` that doesn't work in this case, because it initializes `InetSocketAddress.getHostName()` instead of `InetSocketAddress.getAddress()`.

NiteshKant: aah rite .. can we use a name which indicates that it is resolved then?

PR #1075, opened 2020-05-30
Unwrap instances of `SSLException`

idelpivnitskiy:
Motivation:

`SslHandler` extends `ByteToMessageDecoder` that wraps all thrown
exceptions with `DecoderException`. This complicates understanding of
the problem when users expect an instance of `SSLException`.

Modifications:

- Intercept all caught exceptions and unwrap `SSLException` if possible;

Result:

Users directly receive `SSLException`.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
@@ -43,6 +43,7 @@ import io.netty.channel.ChannelPipeline;
 import io.netty.channel.EventLoop;
 import io.netty.channel.socket.ChannelInputShutdownReadComplete;
 import io.netty.channel.socket.ChannelOutputShutdownEvent;
+import io.netty.handler.codec.DecoderException;
 import io.netty.handler.ssl.SslHandshakeCompletionEvent;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -528,7 +530,18 @@ public void handlerRemoved(ChannelHandlerContext ctx) {
...
-            connection.nettyChannelPublisher.exceptionCaught(cause);
+            connection.nettyChannelPublisher.exceptionCaught(unwrapThrowable(cause));
+        }
+
+        /**
+         * Unwraps certain types of netty exceptions to directly expose its cause to improve debuggability.
+         */
+        private static Throwable unwrapThrowable(final Throwable t) {
+            final Throwable cause = t.getCause();
+            if (t instanceof DecoderException && cause instanceof SSLException) {
```
normanmaurer, line 541: nit: you could only do the actual unwrap if needed as `getCause()` may be expensive.

```java
final Throwable cause;

if (t instanceof DecoderException && ((cause = t.getCause()) instanceof SSLException) {
     return cause;
}
return t;
```

idelpivnitskiy: Follow-up: https://github.com/apple/servicetalk/pull/1077

PR #1076, opened 2020-06-02
Replace MaxSizeBasedWriteDemandEstimator with EWMAWriteDemandEstimator

Scottmitch:
Motivation:
MaxSizeBasedWriteDemandEstimator allocates an array of 32 elements by
default to keep historical size information. This can introduce a lot of
allocation and has been shown in benchmarks to consist of >3% of
allocations inside a TLAB and >%90 of allocations outside of a TLAB. The
estimation fidelity can be reduced with an Exponential Weighted Moving
Average given the loose estimation requirements which can also reduce
the allocation pressure.

Modification:
- Remove MaxSizeBasedWriteDemandEstimator
- Add EWMAWriteDemandEstimator which uses Exponential Weighted Moving
Average to estimate requestN

Result:
Less memory pressure introduces while estimating requestN for writing
data.


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/EWMAWriteDemandEstimator.java
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.transport.netty.internal;
+
+import static io.servicetalk.transport.netty.internal.OverlappingCapacityAwareEstimator.SizeEstimator.defaultEstimator;
+import static java.lang.Math.max;
+
+final class EWMAWriteDemandEstimator extends OverlappingCapacityAwareEstimator {
+    private static final double WEIGHT_NEW = 1 / 5d;
```
Scottmitch, line 22: we can make this more configurable in the future (e.g. 5 correlates to decay factor) ... but these classes aren't very exposed at the moment and I think we can defer until we get some more demand to expose.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/OverlappingCapacityAwareEstimator.java
```diff
@@ -17,6 +17,8 @@
 
 import io.netty.channel.FileRegion;
 
+import java.nio.ByteBuffer;
```
normanmaurer, line 20: What is this import needed for ?

Scottmitch: good question, it isn't :) removed.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/EWMAWriteDemandEstimator.java
```diff
@@ -0,0 +1,51 @@
...
+
+    EWMAWriteDemandEstimator() {
+        this(1024);
+    }
+
+    EWMAWriteDemandEstimator(long sizeAccumulator) {
+        this(sizeAccumulator, defaultEstimator());
+    }
+
+    private EWMAWriteDemandEstimator(long sizeAccumulator, SizeEstimator sizeEstimator) {
```
normanmaurer, line 34: btw why is this constructor needed at all ? This is private so it will never be used outside of this class which means we could just only keep the `EWMAWriteDemandEstimator(long sizeAccumulator)` constructor.

Scottmitch: let me kill this for now

PR #1081, opened 2020-06-05
Observer for DNS ServiceDiscoverer

idelpivnitskiy:
Motivation:

Provides visibility into DNS resolutions behind DNS `ServiceDiscoverer`.

Modifications:

- Add `DnsServiceDiscovererObserver` interface that users can use to
be notified about internal DNS resolutions;
- Add `DefaultDnsServiceDiscovererBuilder#oberser(...)` method to
configure `DnsServiceDiscovererObserver`;
- Improve javadoc;
- Add tests;

Result:

User can get visibility into DNS resolutions.


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -146,27 +155,37 @@
...
+    @Nullable
+    private DnsDiscoveryObserver newDiscoveryObserver(final String address) {
+        return observer != null ? observer.newDiscovery(address) : null;
+    }
+
     @Override
     public Publisher<ServiceDiscovererEvent<InetAddress>> dnsQuery(final String address) {
-        return new ARecordPublisher(true, address).flatMapConcatIterable(identity());
+        requireNonNull(address);
+        return defer(() -> new ARecordPublisher(true, address, newDiscoveryObserver(address))
```
idelpivnitskiy, line 166: This will create a new `DnsDiscoveryObserver` every time a retry strategy resubscribes. This is not ideal, but will work in consistent way with `dnsSrvQuery`.
Alternative approaches I see:
1. Create a `DnsDiscoveryObserver` when users invoke `ServiceDiscoverer#discover(...)`. Because `ARecordPublisher` does not support multiple subscribers anyway, that will work fine. However, the behavior will be different from `dnsSrvQuery`.
2. Create a `DnsDiscoveryObserver` inside `handleSubscribe0`. It will require rework of `dnsSrvQuery` to not trigger `newResolution` for subsequent A queries.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -506,13 +548,25 @@ private void doQuery0() {
                     handleTerminalError0(new ClosedServiceDiscovererException(DefaultDnsClient.this +
                             " has been closed!"));
                 } else {
+                    final DnsResolutionObserver resolutionObserver;
+                    if (discoveryObserver != null) {
+                        try {
+                            resolutionObserver = discoveryObserver.newResolution(name());
+                        } catch (Throwable t) {
+                            handleTerminalError0(t);
```
idelpivnitskiy, line 556: How should we react to exceptions from any method of the observer?

Currently, I fail the discovery publisher if `newDiscovery` or `newResolution` throws an exception. But log & ignore exceptions from `resolutionFailed` and `resolutionCompleted` methods of observer, because there we have additional logic that decides if we need to deliver a terminal now or not, and if we need to generate inactive events or not.

I think we can say that an observer should not affect the behavior of the discovery publisher. Therefore, we can always log & ignore exceptions from the observer methods. WDYT?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
idelpivnitskiy: For now, I trigger `newResolution` for subsequent A queries of SRV discoverer to keep the scope of this PR minimal and avoid changing the SRV discoverer code.
Since SRV is not the most popular use-case it should be ok to defer enhancements there.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -229,25 +248,25 @@ private void assertInEventloop() {
...
-            return serviceName;
+            return "SRV lookups for " + name();
         }
 
         @Override
         protected AbstractDnsSubscription newSubscription(
-                final Subscriber<? super Iterable<ServiceDiscovererEvent<HostAndPort>>> subscriber) {
-            return new AbstractDnsSubscription(true, subscriber) {
+                final Subscriber<? super Iterable<ServiceDiscovererEvent<HostAndPort>>> subscriber,
+                @Nullable final DnsDiscoveryObserver discoveryObserver) {
```
NiteshKant, line 264: Just make the `discoveryObserver` `protected` in the base class instead of passing back again in this method.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
NiteshKant: Since we can not distinguish between a retry vs a new subscribe, so creating a new observer per subscribe makes sense to me.
Since, users control the scope of `DnsServiceDiscovererObserver` and association with a client, if the same observer instance needs to be shared across retries/subscribes, they can do so by returning the same observer instance from `newDiscovery()`?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
NiteshKant: Logging and ignoring keeps the control-flow simple. Since, we do not expect the observer methods to throw, this sounds reasonable.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -625,6 +682,42 @@ private void handleResolveDone0(final Future<DnsAnswer<T>> addressFuture) {
...
+                                                @Nullable final List<ServiceDiscovererEvent<T>> events) {
+                if (resolutionObserver == null) {
+                    return;
+                }
+                final ResolutionResult result;
+                final int ttl = (int) NANOSECONDS.toSeconds(dnsAnswer.ttlNanos());
+                if (events == null) {
+                    result = new DefaultResolutionResult(dnsAnswer.answer().size(), ttl, 0, 0);
+                } else {
+                    final int becameActive = (int) events.stream().filter(ServiceDiscovererEvent::isAvailable).count();
```
NiteshKant, line 709: Instead of reiterating the events list can we have `calculateDifference()` method accept a `BiFunction` to report the numbers?

normanmaurer:
@idelpivnitskiy nit: there is a typo in the commit message in the method name.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -377,11 +397,25 @@ long ttlNanos() {
...
+        AbstractDnsPublisher(final String name, @Nullable final DnsDiscoveryObserver discoveryObserver) {
+            this.name = name;
+            this.discoveryObserver = discoveryObserver;
+        }
+
+        protected abstract AbstractDnsSubscription newSubscription(
+                Subscriber<? super Iterable<ServiceDiscovererEvent<T>>> subscriber,
+                @Nullable DnsDiscoveryObserver discoveryObserver);
+
+        protected final String name() {
```
normanmaurer, line 416: add javadocs to the protected methods

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -625,6 +682,42 @@ private void handleResolveDone0(final Future<DnsAnswer<T>> addressFuture) {
...
+            private void reportResolutionFailed(@Nullable final DnsResolutionObserver resolutionObserver,
+                                                final Throwable cause) {
+                if (resolutionObserver == null) {
+                    return;
+                }
+                try {
+                    resolutionObserver.resolutionFailed(cause);
+                } catch (Throwable unexpected) {
+                    LOGGER.warn("Unexpected exception from {} while reporting DNS resolution failure",
+                            resolutionObserver, unexpected);
```
normanmaurer, line 694: should we also add the `cause` as `addSuppressed` ?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserver.java
```diff
@@ -0,0 +1,104 @@
...
+             * Notifies that the current DNS resolution completed successfully.
+             *
+             * @param result the {@link ResolutionResult}
+             */
+            void resolutionCompleted(ResolutionResult result);
+
+            /**
+             * Results of the current DNS resolution.
+             */
+            interface ResolutionResult {
```
normanmaurer, line 72: nit: this may be just a matter of style but this is quite some odd hierarchy with all these nested interfaces (3 levels).

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserver.java
```diff
@@ -0,0 +1,104 @@
...
+
+                /**
+                 * Number of resolved DNS records.
+                 *
+                 * @return the number of resolved DNS records
+                 */
+                int resolvedRecords();
```
normanmaurer, line 79: consider: `numResolvedRecords`

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserver.java
```diff
@@ -0,0 +1,104 @@
...
+                 * @return the minimum Time To Live (TTL) of the resolved DNS records in seconds
+                 */
+                int ttl();
+
+                /**
+                 * Number of records that became {@link ServiceDiscovererEvent#isAvailable() active}.
+                 *
+                 * @return the number of records that became {@link ServiceDiscovererEvent#isAvailable() active}
+                 */
+                int becameActive();
```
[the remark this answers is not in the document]
normanmaurer: Also consider prefix with `num`

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserverTest.java
```diff
@@ -0,0 +1,266 @@
...
+
+        @Override
+        public DnsResolutionObserver newResolution(final String name) {
+            return NoopDnsResolutionObserver.INSTANCE;
+        }
+    }
+
+    private static class NoopDnsResolutionObserver implements DnsResolutionObserver {
+        static final DnsResolutionObserver INSTANCE = new NoopDnsResolutionObserver();
+
```
normanmaurer, line 255: nit: add private constructor.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
idelpivnitskiy: >  if the same observer instance needs to be shared across retries/subscribes, they can do so by returning the same observer instance from `newDiscovery()`?

Correct, this is the way for users if necessary.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserver.java
```diff
@@ -0,0 +1,104 @@
...
+    interface DnsDiscoveryObserver {
+
+        /**
+         * Notifies that a new DNS resolution started.
+         *
+         * @param name the name for the <a href="https://tools.ietf.org/html/rfc1035#section-4.1.2">DNS question</a> to
+         * be queried
+         * @return {@link DnsResolutionObserver} that provides visibility into results of the current DNS resolution
+         */
+        DnsResolutionObserver newResolution(String name);
```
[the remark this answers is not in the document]
idelpivnitskiy: I will "reserve" `onNewQuery()` for later, if we will need to provide more visibility into individual DNS queries during resolution. Renamed `newResolution` -> `onNewResolution`.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserver.java
```diff
@@ -0,0 +1,104 @@
...
+         * An observer that provides visibility into DNS resolution results.
+         */
+        interface DnsResolutionObserver {
+
+            /**
+             * Notifies that the current DNS resolution failed.
+             *
+             * @param cause {@link Throwable} as a cause for the failure
+             */
+            void resolutionFailed(Throwable cause);
```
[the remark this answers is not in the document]
idelpivnitskiy: Prefixes in the name help when users what a "flatten" version of the interfaces: 
```java
class MyObserver implements DnsServiceDiscovererObserver, DnsDiscoveryObserver,
        DnsResolutionObserver {
...
}
```
If later we will have `DnsQueryObserver` inside `DnsResolutionObserver`, without prefixes users won't be able to differentiate `onError` from one interface from another.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserver.java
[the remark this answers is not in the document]
idelpivnitskiy: Renamed to `nAvailable/nUnavailable` to make them consistent with `ServiceDiscovererEvent#isAvailable()`.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserverTest.java
idelpivnitskiy: This is not a final class, I extend it in a few tests.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserver.java
idelpivnitskiy: Added a prefix for `available/unavailable`. `int resolvedRecords()` look ok to me.

PR #1084, opened 2020-06-10
ServiceTalkThreadContextMap to support null values

Scottmitch:
Motivation:
The MDC API permits for not supporting null [1] but the log4j2
ThreadContextMap API [2] doesn't specify the behavior for null keys. It
is possible direct usage of ThreadContextMap API may assume null values
are permitted as the default implemetnations support null values.

[1] https://www.javadoc.io/doc/org.slf4j/slf4j-api/1.7.30/org/slf4j/MDC.html#put-java.lang.String-java.lang.String-
[2] https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/spi/ThreadContextMap.html#put-java.lang.String-java.lang.String-

Modifications:
- Modify ServiceTalkThreadContextMap and
ServiceTalkTracingThreadContextMap to support null values

Result:
ServiceTalkThreadContextMap permits null values

### DIFF
--- a/servicetalk-log4j2-mdc-utils/src/main/java/io/servicetalk/log4j2/mdc/utils/ServiceTalkThreadContextMap.java
+++ b/servicetalk-log4j2-mdc-utils/src/main/java/io/servicetalk/log4j2/mdc/utils/ServiceTalkThreadContextMap.java
@@ -22,7 +22,6 @@ import io.servicetalk.concurrent.api.AsyncContextMap.Key;
 import org.apache.logging.log4j.ThreadContext;
 import org.apache.logging.log4j.spi.CleanableThreadContextMap;
 import org.apache.logging.log4j.spi.ReadOnlyThreadContextMap;
-import org.apache.logging.log4j.spi.ThreadContextMap;
 import org.apache.logging.log4j.util.BiConsumer;
 import org.apache.logging.log4j.util.ReadOnlyStringMap;
 import org.apache.logging.log4j.util.StringMap;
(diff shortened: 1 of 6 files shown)


servicetalk-opentracing-log4j2/src/main/java/io/servicetalk/opentracing/log4j2/ServiceTalkTracingThreadContextMap.java
```diff
@@ -132,116 +114,34 @@ public boolean isEmpty() {
...
-                        }
-                        break;
-                    }
-                    default:
-                        break;
-                }
-                return (V) storage.get(key);
-            }
+    public StringMap getReadOnlyContextData() {
+        StringMap map = new JdkMapAdapterStringMap(getCopy());
```
Scottmitch, line 118: @NiteshKant - note this is an API breaking change. The previous APIs leaked the underlying `Map` to avoid copy operations as an optimization. Now that the `Map` contains "token" elements that need to be unwrapped it leaks more knowledge and would require exposing the wrap/unwrap of `null`. I don't expect folks will be extending/calling these APIs so the risk is low, and it reduces our API footprint, at the cost of some breakage and less optimal implementation.

servicetalk-log4j2-mdc-utils/src/main/java/io/servicetalk/log4j2/mdc/utils/ServiceTalkThreadContextMap.java
```diff
@@ -40,16 +39,18 @@
  */
 public class ServiceTalkThreadContextMap implements ReadOnlyThreadContextMap, CleanableThreadContextMap {
     private static final Key<Map<String, String>> key = Key.newKey("log4j2Mdc");
+    @SuppressWarnings("RedundantStringConstructorCall")
+    private static final String NULL_STRING = new String("");
 
     @Override
     public final void put(String key, String value) {
-        getStorage().put(key, value);
+        getStorage().put(key, wrapNull(value));
```
colestanfield, line 47: Since MDC doesn't have a contains operator, why not just call remove(key) here? Is the concern that getCopyOfContextMap wouldn't show the null valued key?

Issue #999, opened 2020-04-03
Flaky test: io.servicetalk.http.router.jersey.CancellationTest.cancelRsStreams


idelpivnitskiy:
https://ci.servicetalk.io/job/servicetalk-java11-prb/1138/testReport/junit/io.servicetalk.http.router.jersey/CancellationTest/cancelRsStreams/

```
Error Message
java.lang.AssertionError: 
Expected: is null
     but: was <java.lang.IllegalStateException: Entity input stream has already been closed.>
Stacktrace
java.lang.AssertionError: 
Expected: is null
     but: was <java.lang.IllegalStateException: Entity input stream has already been closed.>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:964)
	at org.junit.Assert.assertThat(Assert.java:930)
	at io.servicetalk.http.router.jersey.CancellationTest.testCancelResponseSingle(CancellationTest.java:262)
	at io.servicetalk.http.router.jersey.CancellationTest.testCancelResponseSingle(CancellationTest.java:213)
	at io.servicetalk.http.router.jersey.CancellationTest.cancelRsStreams(CancellationTest.java:169)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.mockito.internal.junit.JUnitRule$1.evaluateSafely(JUnitRule.java:57)
	at org.mockito.internal.junit.JUnitRule$1.evaluate(JUnitRule.java:48)
[...]

PR #1090, opened 2020-06-26
Do not shutdown output channel until client receives the full response

idelpivnitskiy:
Motivation:

When client observes "Connection: close" header it shutdowns the output
channel as soon as request is written. Some servers interpret FIN from
the client as an indicator that it lost interest in their data and
therefore, server just closes the second half of the connection asap.
As the result, connection may be closed before client receives the full
response from the server.

Modifications:

- Reproduce this scenario using a simple proxy tunnel that is not aware
of HTTP protocol semantics;
- Defer connection closure on the client side until it completes the
full request-response iteration;
- Fail all subsequent or pipelined requests on the connection that moves
to the "closing" state;
- Adjust `RequestResponseCloseHandlerTest` to not expecting outbound
half-closure on the client side;
- Add more tests to verify that client handles "Connection: close" header
correctly;

Result:

Response is not aborted when client observes "Connection: close" header.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
@@ -31,6 +31,7 @@ import io.servicetalk.transport.api.DefaultExecutionContext;
 import io.servicetalk.transport.api.ExecutionContext;
 import io.servicetalk.transport.api.ExecutionStrategy;
 import io.servicetalk.transport.api.ServiceTalkSocketOptions;
+import io.servicetalk.transport.netty.internal.CloseHandler.ChannelOutputClosingEvent;
 import io.servicetalk.transport.netty.internal.CloseHandler.CloseEvent;
 import io.servicetalk.transport.netty.internal.CloseHandler.CloseEventObservedException;
 import io.servicetalk.transport.netty.internal.WriteStreamSubscriber.AbortedFirstWrite;
@@ -408,7 +409,8 @@ public final class DefaultNettyConnection<Read, Write> extends NettyChannelListe
             // It is possible that we have set the writeSubscriber, then the channel becomes inactive, and we will
             // never notify the write writeSubscriber of the inactive event. So if the channel is inactive we notify
             // the writeSubscriber.
-            if (!channel().isActive()) {
+            // It is also possible that Channel is in closing state, so we need to prevent sending more requests.
+            if (!channel().isActive() || closeReason != null) {
                 newChannelOutboundListener.channelClosed(StacklessClosedChannelException.newInstance(
                         DefaultNettyConnection.class, "failIfWriteActive(...)"));
                 return false;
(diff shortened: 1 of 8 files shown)


servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandlerTest.java
```diff
@@ -175,6 +178,7 @@ public Scenarios(final Mode mode, final List<Events> events, final ExpectEvent e
             SR,         // validate Socket TCP RST -> SO_LINGER=0
             UC,         // emit User Closing
             IH, OH, FC, // validate Input/Output Half-close, Full-Close
+            CE,         // emit ChannelOutputClosingEvent
```
NiteshKant, line 181: Till now these events represented things that happened on the socket or what is exposed on the Close handler APIs. `ChannelOutputClosingEvent` is an event generated by the `CloseHandler` itself. Can we avoid expanding scope of events here? Instead validate/simulate things that will happen on the socket, like aborting writes.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -408,7 +409,8 @@ private boolean failIfWriteActive(ChannelOutboundListener newChannelOutboundList
             // It is possible that we have set the writeSubscriber, then the channel becomes inactive, and we will
             // never notify the write writeSubscriber of the inactive event. So if the channel is inactive we notify
             // the writeSubscriber.
-            if (!channel().isActive()) {
+            // It is also possible that Channel is in closing state, so we need to prevent sending more requests.
+            if (!channel().isActive() || closeReason != null) {
```
NiteshKant, line 413: `closeReason` isn't `volatile` so this may never see the `closeReason`.

Instead of using `closeReason` here use the new pipeline event to replace `writableListener` with a "poison value" that indicates abort writes and also aborts the current write if present.

idelpivnitskiy: Good catch!
I have to introduce another volatile boolean to prevent new writes because:
1. Can not reuse `AbortWritesEvent` because it's emitted only when there are additional requests in the pipeline. If there is no pipelining and all requests are sequential, it's not emitted => does not abort new writes.
2. Tried to reuse `ProtocolPayloadEndEvent.OUTBOUND` but it has a much broader scope (also used for h2) and fails a lot of tests.
3. Handling of "poison value" for `channelOutboundListener` complicates internal state management. It adds at least 2 volatile reads (1st in `cleanupOnWriteTerminated, 2nd in `failIfWriteActive`) and complicates `isWriteActive` implementation.

idelpivnitskiy:
quality check failed because of some glitch in network and failed `validateLocalHtmlLinks` task

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
idelpivnitskiy: Discussed offline. Because `closeReason` is not on the hot path it's better just to make `closeReason` volatile instead of expanding internal state with another boolean.

idelpivnitskiy:
JDK11 check failed due to https://github.com/apple/servicetalk/issues/999#issuecomment-652564751

PR #1091, opened 2020-07-07
Add example for gRPC using Maven

Scottmitch:


servicetalk-examples/grpc/helloworld-maven/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>io.servicetalk.examples.grpc.helloworld</groupId>
+  <artifactId>helloworld-maven</artifactId>
+  <packaging>jar</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <name>st-examples</name>
+  <url>https://servicetalk.io</url>
+
+  <properties>
+    <servicetalk.version>0.29.0</servicetalk.version>
```
Scottmitch, line 12: keeping versions up to date maybe painful down the line, but since this is meant to be an example from the user's perspective it is representative of actual use.

servicetalk-examples/grpc/helloworld-maven/src/main/java/io/servicetalk/examples/grpc/helloworld/BlockingHelloWorldClient.java
```diff
@@ -0,0 +1,34 @@
...
+package io.servicetalk.examples.grpc.helloworld.blocking;
+
+import io.servicetalk.grpc.netty.GrpcClients;
+
+import io.grpc.examples.helloworld.Greeter.BlockingGreeterClient;
+import io.grpc.examples.helloworld.Greeter.ClientFactory;
+import io.grpc.examples.helloworld.HelloReply;
+import io.grpc.examples.helloworld.HelloRequest;
+
+public final class BlockingHelloWorldClient {
```
NiteshKant, line 25: Also add a server so the example is self sufficient?

servicetalk-examples/grpc/helloworld-maven/pom.xml
NiteshKant: Any way we can source `gradle.properties` and get the version?

Scottmitch: we maybe able to but because gradle and maven treat build dependencies differently using the version verbatim would lead to unresolved artifacts. I'll update the release script to bump this version.

servicetalk-examples/grpc/helloworld-maven/src/main/java/io/servicetalk/examples/grpc/helloworld/BlockingHelloWorldClient.java
Scottmitch: I just moved the `pom.xml` into the existing `helloworld` example to avoid duplication

PR #1095, opened 2020-07-13
Observability for connection level events

idelpivnitskiy:
Motivation:

Users should be able to get visibility into connection-level events.

Modifications:

- Introduce new `TransportObserver` and `ConnectionObserver` interfaces;
- Support `TransportObserver` for client and server side;
- Tests to verify new functionality;

Result:

Visibility into connection-level events.


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/BaseGrpcClientBuilder.java
```diff
@@ -73,6 +74,14 @@
      */
     BaseGrpcClientBuilder<U, R> enableWireLogging(String loggerName);
 
+    /**
+     * Sets a {@link TransportObserver} that provides visibility into transport events.
+     *
+     * @param transportObserver A {@link TransportObserver} that provides visibility into transport events.
+     * @return {@code this}.
+     */
+    BaseGrpcClientBuilder<U, R> transportObserver(TransportObserver transportObserver);
```
idelpivnitskiy, line 83: Started with the client builder approach for now. Will investigate impact of moving to `ConnectionFactory` in a follow-up PR.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
```diff
@@ -0,0 +1,96 @@
...
+     * Assigns a {@link ConnectionObserver} to the passed {@link Channel}.
+     *
+     * @param channel a {@link Channel} to assign a {@link ConnectionObserver} to
+     * @param observer a {@link ConnectionObserver}
+     */
+    public static void assignConnectionObserver(final Channel channel, @Nullable final ConnectionObserver observer) {
+        if (observer == null) {
+            return;
+        }
+        channel.attr(CONNECTION_OBSERVER).set(observer);
```
idelpivnitskiy, line 51: I will start using channel attributes for now and reconsider approach when all events will be implemented. Don't want to change the internal API to propagate it correctly until I see the whole picture. For now, this PR focuses on behavior correctness for notified events.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
```diff
@@ -0,0 +1,96 @@
...
+import io.netty.util.AttributeKey;
+
+import javax.annotation.Nullable;
+
+import static io.netty.util.AttributeKey.newInstance;
+
+/**
+ * Utilities for {@link TransportObserver}.
+ */
+public final class TransportObserverUtils {
```
NiteshKant, line 32: I am generally concerned about creating such publicly accessible, broadly scoped utilities prematurely.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ChannelSet.java
```diff
@@ -87,14 +97,17 @@ public ChannelSet(Executor offloadingExecutor) {
      */
     public boolean addIfAbsent(final Channel channel) {
         final boolean added = channelMap.putIfAbsent(channel.id(), channel) == null;
+        if (added) {
+            final ConnectionObserver observer = this.connectionObserverFactory != null ?
+                    this.connectionObserverFactory.get() : null;
+            if (observer != null) {
+                assignConnectionObserver(channel, observer);
```
NiteshKant, line 104: Can we do this assignment in `TcpServerBinder`? `ChannelSet` is just a means to have a collection of channels for a server; adding this functionality here seems incorrect. Additionally it increases the scope of `TransportObserverUtils` outside the tcp-netty-internal module

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
```diff
@@ -0,0 +1,96 @@
...
+        if (observer == null) {
+            return;
+        }
+        channel.attr(CONNECTION_OBSERVER).set(observer);
+        channel.closeFuture().addListener((ChannelFutureListener) future -> {
+            Throwable t = connectionError(channel);
+            if (t == null) {
+                observer.connectionClosed();
+            } else {
+                observer.connectionClosed(t);
```
NiteshKant, line 57: In which cases does the `closeFuture()` complete with an error? 
I think this usage does not map to your intent of `connectionClosed(Throwable)` contract, does it?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public static final ChannelInitializer TRANSPORT_OBSERVER_INITIALIZER = new TransportObserverInitializer();
+
+    private TransportObserverInitializer() {
+        // Singleton
+    }
+
+    @Override
+    public void init(final Channel channel) {
+        final ConnectionObserver observer = connectionObserver(channel);
```
NiteshKant, line 42: You are decoupling observer creation from channel initialization. Is the intent here to account for time delta between start of connect and call to `init()` here? 
The init sequence of creating an observer and adding the transport handler seems convoluted. Have you thought about creating the observer in this method?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+        if (observer != null) {
+            channel.pipeline().addLast(new TransportObserverChannelHandler(observer));
+        }
+    }
+
+    private static final class TransportObserverChannelHandler extends ChannelDuplexHandler {
```
NiteshKant, line 48: Can we also hook in close events to the observer using this handler instead of the close future?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -0,0 +1,54 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.transport.api;
+
+/**
+ * An observer interface that provides visibility into events associated with a single network connection.
```
NiteshKant, line 19: `a single network` => `a network`

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+public interface ConnectionObserver {
+
+    /**
+     * Notifies when the connection reads a chunk of data.
+     *
+     * @param size size of the data chunk read
+     */
+    void dataRead(int size);
```
NiteshKant, line 28: The method names here are in past tense "dataWritten", "flushed", etc. Is this intentional that they are associated with timing? While invoking these callbacks we are not consistent; `dataRead()` is after we have read from socket, `dataWritten()` is before we have written to the socket. 

An alternate is to use `onDataRead`, `onDataWrite`, `onFlush` ... etc or `whenDataRead`, `whenFlushed` which hides the timing (before/after) and are also similar to the callbacks on our async primitives.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package io.servicetalk.transport.api;
+
+/**
+ * An observer interface that provides visibility into events associated with a single network connection.
+ */
+public interface ConnectionObserver {
+
+    /**
+     * Notifies when the connection reads a chunk of data.
```
NiteshKant, line 24: Suggestion: A callback when {@code size} bytes are read from the connection.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @param size size of the data chunk read
+     */
+    void dataRead(int size);
+
+    /**
+     * Notifies when the connection writes a chunk of data.
```
NiteshKant, line 31: Suggestion: A callback when {@code size} bytes are written to the connection.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @param size size of the data chunk written
+     */
+    void dataWritten(int size);
+
+    /**
+     * Notifies when flush operation is made on the connection. The flush operation will try to flush out all previous
```
NiteshKant, line 38: Suggestion: Callback when previously written data is flushed to the connection.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Assigns a {@link Throwable} to the passed {@link Channel}.
+     *
+     * @param channel a {@link Channel} to assign a {@link Throwable} to
+     * @param error a {@link Throwable}
+     */
+    public static void assignConnectionError(final Channel channel, final Throwable error) {
+        if (connectionObserver(channel) != null) {
+            channel.attr(CONNECTION_ERROR).setIfAbsent(error);
```
NiteshKant, line 82: IIUC, the current code paths registering an error is a subset of things you would want to cover eventually.
Looks like we need to have some association with the close-handler to determine which error should be the cause of closure instead of this "first-writer-wins" semantics?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ChannelSet.java
Scottmitch, line 104: its somewhat awkward to handle the `ConnectionObserver` in the `ChannelSet` on the server, and directly in the transport code for the client. did you consider moving this into the transport as well?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/BaseGrpcClientBuilder.java
idelpivnitskiy: That will give more flexibility for users to correlate events/state before and after `ConnectionFactory#newConnection` invocation. Otherwise, it's hard to connect with events on the builder. Similar limitations that we discussed for `DnsQueryLifecycleObserver` in netty.
In addition, it gives users control when they need to create a new observer and for which resolved addresses.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
```diff
@@ -62,6 +63,9 @@
     @Override
     public abstract MultiAddressHttpClientBuilder<U, R> enableWireLogging(String loggerName);
 
+    @Override
+    public abstract MultiAddressHttpClientBuilder<U, R> transportObserver(TransportObserver transportObserver);
```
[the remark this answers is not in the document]
idelpivnitskiy: `TransportObserver` works as a factory for `ConnectionObserver`s. So, I don't think we need a supplier here.
For multi-address builder we have a general problem that most of the options are not per-client (except security config and filters). Taking into account that `TransportObserver` may move to the `ConnectionFactory` I will defer this enhancement.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
idelpivnitskiy: This is an internal utility.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ChannelSet.java
idelpivnitskiy: Good call, let me move it from `ChannelSet` to a better place.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
idelpivnitskiy: The cases are represented in `TcpServerTransportObserverErrorsTest`. The error may happen in the pipeline or in the `ConnectionAcceptor`.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
idelpivnitskiy: Correct, the intend was to create a `ConnectionObserver` as soon as we have access to the `Channel`. But you are right, there is not much between the netty's `initChannel` and the current `init` method. Moved it to `TransportObserverInitializer` and it looks much better. Thanks!

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
idelpivnitskiy: That will lose errors outside of the pipeline.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
idelpivnitskiy: Not sure about `CloseHandler` because a `Channel` may be closed from different places. After investigating different closure cases it looks acceptable to use "first-writer-wins" semantics because channels are always closed right after observing an error.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ChannelSet.java
```diff
@@ -68,14 +72,20 @@ public void operationComplete(final ChannelFuture future) {
...
      *
      * @param offloadingExecutor {@link Executor} to use for offloading close signals.
+     * @param connectionObserverFactory a factory that creates a {@link ConnectionObserver} when a new {@link Channel}
+     * is added
      */
-    public ChannelSet(Executor offloadingExecutor) {
+    public ChannelSet(final Executor offloadingExecutor,
+                      @Nullable final Supplier<ConnectionObserver> connectionObserverFactory) {
         onClose = fromSource(onCloseProcessor).publishOn(offloadingExecutor);
+        this.connectionObserverFactory = connectionObserverFactory;
```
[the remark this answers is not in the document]
idelpivnitskiy: Thanks for the suggestion! I moved the code from `ChannelSet` to another place.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
NiteshKant: > Not sure about CloseHandler because a Channel may be closed from different places.

Sure close can be initiated from multiple places but close-handler holds the close state machine (for H1). 
Having the determination here that the first error is important and the rest are silently ignored essentially is ignoring the state represented by the close state machine (close handler for h1 and netty for h2).

This change only contains 2 calls to this method, how about we determine the mechanism to handle connection errors as a follow-up when we have a better idea about how to handle the reason for close holistically.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
NiteshKant: Sure but it is a public class. Do we really want a whole utility to get and set from a channel attribute?

In any case, now it can be package private I think?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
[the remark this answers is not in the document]
NiteshKant: `Supplier` is needed as many clients can be built from the same builder. It is completely fair to ask folks to set a new `TransportObserver` every time they create a client, is that what you intend? If so, then this is what makes the multi-address client case harder. Can we avoid adding this to multi-address client or at least follow the `secure()` method model where someone can provide a `TransportObserver` for a `HostAndPort`?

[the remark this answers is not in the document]
idelpivnitskiy: Updated multi-address client builder to build a `TransportObserver` per `HostAndPort`.

idelpivnitskiy:
Discussed offline. Merging this in to continue work on other events and will do follow up for outstanding comments after that.

PR #1099, opened 2020-07-19
Security handshake observability

idelpivnitskiy:
Motivation:

Provide visibility into security handshake events: when it's started and
when it's complete or failed.

Modifications:

- Add `ConnectionObserver.SecurityHandshakeObserver`;
- Enhance `TransportObserverInitializer` to capture when security
handshake starts;
- Report the result of SSL handshake from
`NettyPipelineSslUtils.extractSslSession`;
- Test observability of successful security handshakes;
- Test observability of failed security handshakes;

Result:

Users can observe when security handshake starts and completes.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
@@ -42,27 +42,19 @@ import static java.util.Objects.requireNonNull;
  */
 public final class TransportObserverInitializer implements ChannelInitializer {
 
-    /**
-     * Tells which side is using secure connection.
-     */
-    public enum SecureSide {
-        CLIENT,
-        SERVER,
-        NONE
-    }
-
     private final TransportObserver transportObserver;
-    private final SecureSide secureSide;
+    private final boolean secure;
 
     /**
      * Creates a new instance.
      *
      * @param transportObserver {@link TransportObserver} to initialize for the channel
-     * @param secureSide tells which side is using secure connection
+     * @param secure {@code true} if the observed connection is secure
      */
-    public TransportObserverInitializer(final TransportObserver transportObserver, final SecureSide secureSide) {
+    public TransportObserverInitializer(final TransportObserver transportObserver,
+                                        final boolean secure) {
         this.transportObserver = requireNonNull(transportObserver);
-        this.secureSide = secureSide;
+        this.secure = secure;
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
```diff
@@ -16,46 +16,93 @@
...
-        channel.pipeline().addLast(new TransportObserverChannelHandler(observer));
+        final ChannelPipeline pipeline = channel.pipeline();
+        pipeline.addLast(new TransportObserverHandler(observer, secureSide));
     }
 
-    private static final class TransportObserverChannelHandler extends ChannelDuplexHandler {
+    private static final class TransportObserverHandler extends ChannelDuplexHandler {
         private final ConnectionObserver observer;
+        private final SecureSide secure;
+        private boolean handshakeStarted;
```
NiteshKant, line 79: rename to `handshakeStartNotified` to indicate that this does not reflect handshake start state but the notification to the observer.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
```diff
@@ -16,46 +16,93 @@
...
+            if (secure == SecureSide.CLIENT) {
+                reportSecurityHandshakeStarting(ctx.channel());
+            }
+            ctx.fireChannelActive();
+        }
+
+        void reportSecurityHandshakeStarting(final Channel channel) {
+            if (!handshakeStarted) {
+                TransportObserverUtils.reportSecurityHandshakeStarting(channel);
+                handshakeStarted = true;
```
NiteshKant, line 104: Set this before sending the callback to account for reentrancy?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
```diff
@@ -16,46 +16,93 @@
...
+        @Override
+        public void handlerAdded(final ChannelHandlerContext ctx) {
+            if (secure != SecureSide.NONE && ctx.channel().isActive()) {
+                reportSecurityHandshakeStarting(ctx.channel());
+            }
+        }
+
+        @Override
+        public void channelActive(final ChannelHandlerContext ctx) {
+            if (secure == SecureSide.CLIENT) {
```
NiteshKant, line 95: Why is the condition in `handlerAdded` and `channelActive` different?
If this is valid, then add a comment to explain the same.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
```diff
@@ -92,4 +139,32 @@ public void flush(final ChannelHandlerContext ctx) {
...
+            // Singleton
+        }
+
+        @Override
+        public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) {
+            if (evt instanceof SslHandshakeCompletionEvent) {
+                final Channel channel = ctx.channel();
+                final SecurityHandshakeObserver observer = securityHandshakeObserver(channel);
+                assert observer != null;
+                final SSLSession sslSession = extractSslSession(ctx.pipeline(), (SslHandshakeCompletionEvent) evt,
```
NiteshKant, line 158: We are already intercepting this event and extract SSL session in `DefaultNettyConnection`, can we do these callbacks there instead of adding another handler?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
```diff
@@ -16,46 +16,93 @@
...
 
 /**
  * A {@link ChannelInitializer} that registers a {@link ConnectionObserver} for all channels.
  */
 public final class TransportObserverInitializer implements ChannelInitializer {
 
+    /**
+     * Tells which side is using secure connection.
+     */
+    public enum SecureSide {
```
NiteshKant, line 48: The name is confusing since TLS can not be on one side.
I think this enum is overloaded to give two piece of information:
- Whether TLS is enabled.
- Whether this initializer is used by client/server.

To avoid confusion caused by this overload, I would suggest splitting this into two fields:
- `boolean isSecure`
- `boolean forClient`

This is similar to what we do for security config.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
idelpivnitskiy: That was my overthinking :) Channel is already active when server adds a handler. Therefore, there is no need to expect `channelActive` on the server-side. But we can make conditions the same with no side-effects. Thanks!

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverInitializer.java
idelpivnitskiy: There is `H2ParentConnectionContext` that also performs a similar logic, but this class is from `servicetalk-http-netty`.
The goal to have `SecurityHandshakeObserverHandler` is to encapsulate everything related to transport observer in `servicetalk-transport-netty-internal`. For users who are interested in observability, it looks like small and acceptable overhead.

idelpivnitskiy: Moved this logic to `NettyPipelineSslUtils.extractSslSession` instead: https://github.com/apple/servicetalk/pull/1099/commits/e118cd1c3dbf862f5d4a4e8fe1467bffdfdeb7db

Issue #1102, opened 2020-07-23
Flaky test: DefaultDnsClientTest.repeatDiscoverNxDomain


idelpivnitskiy:
https://ci.servicetalk.io/job/servicetalk-java8-prb/1450/testReport/junit/io.servicetalk.dns.discovery.netty/DefaultDnsClientTest/repeatDiscoverNxDomain/

```
Regression
io.servicetalk.dns.discovery.netty.DefaultDnsClientTest.repeatDiscoverNxDomain

Error Message
java.lang.AssertionError: Unexpected exception during DNS lookup.
Expected: an instance of java.net.UnknownHostException
     but: null
Stacktrace
java.lang.AssertionError: Unexpected exception during DNS lookup.
Expected: an instance of java.net.UnknownHostException
     but: null
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at io.servicetalk.dns.discovery.netty.DefaultDnsClientTest.repeatDiscoverNxDomain(DefaultDnsClientTest.java:478)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at io.servicetalk.concurrent.internal.ServiceTalkTestTimeout$TimeoutStatement$CallableStatement.call(ServiceTalkTestTimeout.java:171)
	at io.servicetalk.concurrent.internal.ServiceTalkTestTimeout$TimeoutStatement$CallableStatement.call(ServiceTalkTestTimeout.java:163)
[...]

PR #1105, opened 2020-07-27
Observability for connection establishment and read-write events

idelpivnitskiy:
Motivation:

Provide users ways to know when connection is established and ready for
usage, what was the type of connection (multiplexed or non-multiplexed),
read-write events associated with the connection.

Modifications:

- Add `NonMultiplexedObserver`, `MultiplexedObserver`, `StreamObserver`,
`ReadObserver`, and `WriteObserver` interfaces;
- Implement trigger points for new observability interfaces;
- Extract immutable methods of `ConnectionContext` into `ConnectionInfo`
interface;
- Test new functionality;

Result:

Observability for connection establishment and read-write events on the
connection.


idelpivnitskiy:
build failed due to #1102 

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelPublisher.java
```diff
@@ -47,17 +49,21 @@
...
+    private NonMultiplexedObserver readWriteObserver;
 
     private final Channel channel;
     private final CloseHandler closeHandler;
     private final EventLoop eventLoop;
     private final Predicate<T> terminalSignalPredicate;
 
-    NettyChannelPublisher(Channel channel, Predicate<T> terminalSignalPredicate, CloseHandler closeHandler) {
+    NettyChannelPublisher(Channel channel, Predicate<T> terminalSignalPredicate, CloseHandler closeHandler,
+                          @Nullable NonMultiplexedObserver readWriteObserver) {
```
NiteshKant, line 61: The `@Nullable` semantics are not ideal here (and in other places) specially here when we can set this later through `readWriteObserver()` method below.

It seems there are two scenarios here:
1. There is no observer configured
2. Observer is configured but is not available because of how our APIs are structured (observer is returned as part of `established()` call)

Instead of wiring in `null` and late set behavior everywhere can we scope it to an observer implementation:
- For case (1) have a NOOP observer
- For case (2) have a place holder observer at the creation time of connection which delegates to an actual observer, when available.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelPublisher.java
```diff
@@ -69,6 +75,10 @@ protected void handleSubscribe(Subscriber<? super T> nextSubscriber) {
         }
     }
 
+    void readWriteObserver(@Nullable final NonMultiplexedObserver readWriteObserver) {
+        this.readWriteObserver = readWriteObserver;
```
NiteshKant, line 79: If you do end up doing this, clarify and verify whether this should only be accessed/assigned from the eventloop

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelPublisher.java
```diff
@@ -318,14 +333,44 @@ private void assertInEventloop() {
...
+        @Nullable
+        private final ReadObserver observer;
         final Subscriber<? super T> associatedSub;
 
-        private SubscriptionImpl(Subscriber<? super T> associatedSub) {
-            this.associatedSub = associatedSub;
+        private SubscriptionImpl(final Subscriber<? super T> subscriber,
+                                 @Nullable final NonMultiplexedObserver readWriteObserver) {
+            observer = readWriteObserver == null ? null : requireNonNull(readWriteObserver.onNewRead());
+            this.associatedSub = observer == null ? subscriber : new Subscriber<T>() {
```
NiteshKant, line 343: this seems backward to wrap the `Subscriber` inside a `Subscription`. Also, we are not handling errors thrown from the observer.
Can we start with using `when*` methods on the `NettyChannelPublisher` for this instrumentation instead of baking it in here?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
```diff
@@ -27,7 +28,10 @@
 import static io.netty.util.AttributeKey.newInstance;
 import static java.util.Objects.requireNonNull;
 
-final class TransportObserverUtils {
+/**
+ * Utilities for {@link TransportObserver}.
+ */
+public final class TransportObserverUtils {
```
NiteshKant, line 34: I am getting increasingly concerned about the number of utilities that we are creating as part of this work. It smells like we are missing an abstraction but I can't put my finger on it yet.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
```diff
@@ -128,13 +134,17 @@ protected void handleSubscribe(final Subscriber<? super H2ServerParentConnection
...
+                            connectionObserver(channel));
 
                     new H2ServerParentChannelInitializer(h2ServerConfig,
-                        new io.netty.channel.ChannelInitializer<Channel>() {
+                        new io.netty.channel.ChannelInitializer<Http2StreamChannel>() {
                             @Override
-                            protected void initChannel(final Channel streamChannel) {
+                            protected void initChannel(final Http2StreamChannel streamChannel) {
                                 connection.trackActiveStream(streamChannel);
+                                final StreamObserver streamObserver = registerStreamObserver(streamChannel,
```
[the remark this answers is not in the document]
Scottmitch: If this throws, do we cleanup and propagate async control flow?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionContext.java
```diff
@@ -17,79 +17,8 @@
...
-    interface Protocol {
-
-        /**
-         * Returns name of the protocol.
-         *
-         * @return name of the protocol
-         */
-        String name();
-    }
+public interface ConnectionContext extends ConnectionInfo, ListenableAsyncCloseable {
```
Scottmitch, line 23: can you clarify the motivation for this change? is this to prevent folks from closing in the observer APIs?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/Flush.java
```diff
@@ -80,11 +92,17 @@ public void onSubscribe(Subscription subscription) {
...
+                        if (observer != null) {
+                            observer.requestedToWrite(n);
+                        }
                         subscription.request(n);
                     }
 
                     @Override
                     public void cancel() {
+                        if (observer != null) {
+                            observer.writeCancelled();
```
Scottmitch, line 104: should we be more defensive here and account for exceptions?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -626,6 +646,9 @@ private void completeSubscriber() {
             assert subscriber != null;
             SingleSource.Subscriber<? super DefaultNettyConnection<Read, Write>> subscriberCopy = subscriber;
             subscriber = null;
+            if (observer != null) {
+                connection.readWriteObserver(requireNonNull(observer.established(connection)));
```
Scottmitch, line 650: should we be more defensive here and account for exceptions? we are invoking user code that may throw and break async control flow notification to `subscriberCopy` IIUC.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -79,4 +95,135 @@
          */
         void handshakeComplete(SSLSession sslSession);
     }
+
+    /**
+     * An observer interface that provides visibility into read and write events of a non-multiplexed connection.
+     */
+    interface NonMultiplexedObserver {
```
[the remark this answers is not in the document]
idelpivnitskiy: I would prefer to avoid the word `Factory` and keep just an `Observer` suffix everywhere for consistency. Observer provides callbacks for events, we don't need to be too specific for events that return another observer.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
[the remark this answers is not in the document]
idelpivnitskiy: @NiteshKant this scenario should be visible via `ConnectionObserver.connectionClosed(Throwable)` because netty will propagate this exception through the pipeline.

@Scottmitch added try-catch for an observer invocation.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
idelpivnitskiy: The current goal is to build the API and provide test coverage to ensure that callbacks are triggered at the correct time. Implementation intentionally uses a simplified approach to minimize internal changes until the full picture is there. The follow-up PR will reconsider the internal implementation approach.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelPublisher.java
idelpivnitskiy: There is also a 3rd case when an observer is available at the ctor time for HTTP/2 stream.

Your suggestion to use `when*` operators on `NettyChannelPublisher` helped to simplify it a bit. I will defer further improvements of internal implementation for the follow-up PR.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -48,6 +48,22 @@
      */
     SecurityHandshakeObserver onSecurityHandshake();
```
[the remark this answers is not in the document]
idelpivnitskiy: The motivation was to avoid having a specific protocol name, like SSL or TLS, and be consistent with other ST API: `secure()` method on the builders and `SecurityConfigurator` interfaces. `SSLSession` is a java class that we can not change, therefore decided not to stick to its name.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionContext.java
idelpivnitskiy: Correct, `ConnectionInfo` is a read-only view on the connection to prevent modifications of its state, like closing.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -79,4 +95,135 @@
...
+         * Callback when new items are requested to read.
+         *
+         * @param n number of requested items to read
+         */
+        void requestedToRead(long n);
+
+        /**
+         * Invokes when a new item is read.
+         */
+        void itemRead();
```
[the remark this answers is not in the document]
idelpivnitskiy: Improved javadoc.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -79,4 +95,135 @@
...
+         *
+         * @return {@link StreamObserver} that provides visibility into stream events
+         */
+        StreamObserver onNewStream();
+    }
+
+    /**
+     * An observer interface that provides visibility into stream events.
+     */
+    interface StreamObserver extends NonMultiplexedObserver {
```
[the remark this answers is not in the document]
idelpivnitskiy: Good idea! Added this to the interface javadoc, ptal: https://github.com/apple/servicetalk/pull/1105/commits/c336f398494e708d78cddc30351a4b07ce73d4bf

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverTest.java
```diff
@@ -0,0 +1,409 @@
...
+        // Failure of the read triggers cancellation of the write.
+        verify(clientWriteObserver, await()).writeCancelled();
+
+        verify(serverReadObserver, atLeastOnce()).requestedToRead(anyLong());
+        verify(serverReadObserver, atLeastOnce()).itemRead();
+        if (serverReadCompletes) {
+            verify(serverReadObserver).readComplete();
+        } else {
+            // FIXME: because nobody subscribes to the request payload publisher, the terminal signal is not delivered
+            // verify(serverReadObserver).readFailed(any(ClosedChannelException.class));
```
idelpivnitskiy, line 310: @NiteshKant @Scottmitch wdyt about this?

The problem here is that connection is closed, but the `connection.read()` did not terminate and was not canceled. This happens because service did not subscribe to the payload body publisher and auto-draining does not happen if response body publisher fails because the connection (or stream) will be closed anyway. Therefore, `NettyChannelPublisher#pending` queue has 2 elements:
1. empty trailers
2. `ClosedChannelException`

Because there are trailers, the terminal event is not delivered. IIUC the control flow the terminal signal is delayed intentionally to let the subscriber consume a successful request before terminating it.

Is it ok that we do not terminate `connection.read()` in this case when connection (or stream) is closed?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -268,6 +275,49 @@ protected void handleSubscribe(
         };
     }
 
+    private Publisher<Read> registerReadObserver(final Publisher<Read> readPublisher) {
+        return defer(() -> {
+            final DataObserver dataObserver = this.dataObserver;
+            final ReadObserver observer;
+            if (dataObserver == null ||
+                    (observer = safeReport(dataObserver::onNewRead, dataObserver, "new read")) == null) {
+                return readPublisher.subscribeShareContext();
```
NiteshKant, line 284: Prefer operators over new source creation and context propagation handling complexity.

```java
        return readPublisher.liftSync(target -> {
            final DataObserver dataObserver = this.dataObserver;
            final ReadObserver observer;
            if (dataObserver == null ||
                    (observer = safeReport(dataObserver::onNewRead, dataObserver, "new read")) == null) {
                return target;
            }
            return new PublisherSource.Subscriber<Read>() {
                @Override
                public void onSubscribe(final Subscription subscription) {
                    target.onSubscribe(new Subscription() {
                        @Override
                        public void request(final long n) {
                            safeReport(() -> observer.requestedToRead(n), observer, "requested to read");
                            subscription.request(n);
                        }

                        @Override
                        public void cancel() {
                            safeReport(observer::readCancelled, observer, "read cancelled");
                            subscription.cancel();
                        }
                    });
                }

                @Override
                public void onNext(@Nullable final Read read) {
                    safeReport(observer::itemRead, observer, "item read");
                    target.onNext(read);
                }

                @Override
                    safeReport(() -> observer.readFailed(t), observer, "read failed", t);
                    target.onError(t);
                }

                @Override
                public void onComplete() {
                    safeReport(observer::readComplete, observer, "read complete");
                    target.onComplete();
                }
            };
        });

```

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
```diff
@@ -68,25 +83,84 @@ static ConnectionObserver connectionObserver(final Channel channel) {
...
     }
+
+    /**
+     * Safely executes the passed {@link Runnable} and logs unexpected exception if any.
+     *
+     * @param runnable {@link Runnable} to run
+     * @param observer {@link Object} that is invoked
+     * @param event event that is invoked
+     */
+    public static void safeReport(final Runnable runnable, final Object observer, final String event) {
```
NiteshKant, line 122: An alternate to having call sites worry about unexpected error handling is to wrap the `Observer` with a catch-all implementation. The positives are:
- Less pollution on every call-site and danger of missing the error handling at some call-sites.
- Central place where certain observer implementations can avoid allocation if they are known to be safe.

The cons are arguably higher allocation which is debatable when compared to this approach which allocates a new `Runnable` (capturing lambdas mostly) on each call.

Did you consider this alternative?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverTest.java
NiteshKant: > the connection.read() did not terminate and was not canceled. 

The server side subscribes to `connection.read()` to read the request so it should be cancelled if the response failed which in turn shuts down the connection processing. Looks like something is missing in this analysis/test.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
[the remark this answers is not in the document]
idelpivnitskiy: For the connection, time delta between `onNewConnection()` and `established*()` is significant because it involves a TCP syn-ack and optionally an SSL handshake. 

Stream creation happens locally without network interaction. Therefore, the time between a request for a new stream and `childChannelActive` is negligible. Can not imagine a use-case when having 2 events will provide value for users.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
idelpivnitskiy: Yes, I considered this alternative and do not worry much about the allocation overhead of catch-all implementation.
The case that I couldn't solve with that approach is when the original observer unexpectedly returns `null`, instead of another observer layer. If I do
```java
final class CatchAllTransportObserver implements TransportObserver {
    //...
    @Override
    public ConnectionObserver onNewConnection() {
        final ConnectionObserver connectionObserver = safeReport(delegate::onNewConnection, delegate, "new connection");
        if (connectionObserver == null) {
            return null;
        }
        return new CatchAllConnectionObserver(connectionObserver);
    }
```
Then intellij idea and spotbugs both complain:
```
io.servicetalk.transport.netty.internal.CatchAllTransportObserver.onNewConnection() may return null, but is declared @Nonnull
```

Also, the caller of an observer API should always do something like:
```java
        final ConnectionObserver observer = transportObserver.onNewConnection();
        if (observer == null) {
            return;
        }
```
where intellij idea and spotbugs again complain:
```
Redundant nullcheck of observer, which is known to be non-null in io.servicetalk.transport.netty.internal.TransportObserverInitializer.init(Channel)
```

I decided that it will be annoying to suppress all these warnings twice: in spotbugs exclusion XML and an intellij idea via comments/annotations. And don't think that we need to annotate all methods as `@Nullable`.

LMK what you think and I can address it in a follow-up.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverTest.java
idelpivnitskiy: Agreed, `readCancel` is expected here instead, but does not happen:
```
Wanted but not invoked:
serverReadObserver.readCancelled();
-> at io.servicetalk.http.netty.HttpTransportObserverTest.testServerFailsResponsePayloadBody(HttpTransportObserverTest.java:310)

However, there were exactly 2 interactions with this mock:
serverReadObserver.requestedToRead(1L);
-> at io.servicetalk.transport.netty.internal.DefaultNettyConnection$2$1.lambda$request$0(DefaultNettyConnection.java:291)

serverReadObserver.itemRead();
-> at io.servicetalk.transport.netty.internal.TransportObserverUtils.safeReport(TransportObserverUtils.java:124)
```

Opened #1112

idelpivnitskiy:
Merging this PR to start working on follow-ups.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/TransportObserverUtils.java
NiteshKant:
```
if (connectionObserver == null) {
            return null;
}
```

return a `NoopConnectionObserver` instead of `null`?

PR #1111, opened 2020-07-31
Do not half-close outbound channel on graceful client closure

idelpivnitskiy:
Motivation:

When client observes users request to close the connection gracefully it
shutdowns the output channel if the in-flight request is already written,
then awaits for the incoming response to close the full channel.
Some servers interpret FIN from the half-closed client as an indicator that
it lost interest in their data and therefore, server may closes the second
half of the connection asap. As the result, connection may be closed before
client receives the full response from the server.

Modifications:

- Reproduce this scenario using a simple proxy tunnel that does not support
half-closed connections;
- Close the proxy tunnel when it sees EOF or an error to propagate closure
to the other side of the tunnel;
- Defer connection closure on the client-side until it completes the full
request-response iteration and connection becomes idle;
- Add more tests to verify that client handles graceful closure correctly
at any state of in-flight request-response processing;

Result:

In-flight responses are not aborted when client closes gracefully.


idelpivnitskiy:
@NiteshKant yes, we should. Server-side change is coming soon

PR #1114, opened 2020-08-01
Configure `TransportObserver` via `ConnectionFactory` on the client-side

idelpivnitskiy:
Motivation:

Having `TransportObserver` on `ConnectionFactory#newConnection` method gives
users more flexibility when they need to correlate prior events and a state
with observability events of new connections. It also allows them to choose
for which `ResolvedAddress`es they need an observer.

Modification:

- Add `TransportObserver` as a second argument for
`ConnectionFactory#newConnection`;
- Remove `#transportObserver(...)` configuration method from all client builders;
- Introduce `TransportObserverConnectionFactoryFilter` to simplify `TransportObserver`
configuration for the client;
- Update code to use new `ConnectionFactory` API;

Results:

More flexible approach for users to configure `TransportObserver` on the
client-side.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/BiTransportObserver.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/BiTransportObserver.java
@@ -63,29 +63,20 @@ final class BiTransportObserver implements TransportObserver {
 
         @Override
         public void onDataRead(final int size) {
-            try {
-                first.onDataRead(size);
-            } finally {
-                second.onDataRead(size);
-            }
+            first.onDataRead(size);
+            second.onDataRead(size);
         }
 
         @Override
         public void onDataWrite(final int size) {
-            try {
-                first.onDataWrite(size);
-            } finally {
-                second.onDataWrite(size);
-            }
+            first.onDataWrite(size);
+            second.onDataWrite(size);
         }
 
         @Override
         public void onFlush() {
-            try {
-                first.onFlush();
-            } finally {
-                second.onFlush();
-            }
+            first.onFlush();
+            second.onFlush();
         }
 
         @Override
@@ -105,20 +96,14 @@ final class BiTransportObserver implements TransportObserver {
 
         @Override
         public void connectionClosed(final Throwable error) {
-            try {
-                first.connectionClosed(error);
-            } finally {
-                second.connectionClosed(error);
-            }
+            first.connectionClosed(error);
+            second.connectionClosed(error);
         }
 
         @Override
         public void connectionClosed() {
-            try {
-                first.connectionClosed();
-            } finally {
-                second.connectionClosed();
-            }
+            first.connectionClosed();
+            second.connectionClosed();
         }
     }
 
@@ -135,20 +120,14 @@ final class BiTransportObserver implements TransportObserver {
 
         @Override
         public void handshakeFailed(final Throwable cause) {
-            try {
-                first.handshakeFailed(cause);
-            } finally {
-                second.handshakeFailed(cause);
-            }
+            first.handshakeFailed(cause);
+            second.handshakeFailed(cause);
         }
 
         @Override
         public void handshakeComplete(final SSLSession sslSession) {
-            try {
-                first.handshakeComplete(sslSession);
-            } finally {
-                second.handshakeComplete(sslSession);
-            }
+            first.handshakeComplete(sslSession);
+            second.handshakeComplete(sslSession);
         }
     }
 
@@ -202,20 +181,14 @@ final class BiTransportObserver implements TransportObserver {
 
         @Override
         public void streamClosed(final Throwable error) {
-            try {
-                first.streamClosed(error);
-            } finally {
-                second.streamClosed(error);
-            }
+            first.streamClosed(error);
+            second.streamClosed(error);
         }
 
         @Override
         public void streamClosed() {
-            try {
-                first.streamClosed();
-            } finally {
-                second.streamClosed();
-            }
+            first.streamClosed();
+            second.streamClosed();
         }
     }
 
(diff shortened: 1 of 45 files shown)


servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionFactory.java
```diff
@@ -31,7 +34,9 @@
      * Creates and asynchronously returns a connection.
      *
      * @param address to connect.
+     * @param observer {@link TransportObserver} that provides visibility into transport events associated with a new
+     * connection.
      * @return {@link Single} that emits the created connection.
      */
-    Single<C> newConnection(ResolvedAddress address);
+    Single<C> newConnection(ResolvedAddress address, @Nullable TransportObserver observer);
```
NiteshKant, line 41: IIUC, expected usage is to add a `ConnectionFactory` for a client that introduces a `TransportObserver` for the call.
What is the guidance for such factories if there is already an observer passed to them?

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionFactory.java
```diff
@@ -31,7 +34,9 @@
      * Creates and asynchronously returns a connection.
      *
      * @param address to connect.
+     * @param observer {@link TransportObserver} that provides visibility into transport events associated with a new
```
NiteshKant, line 37: Instead of repeating the `TransportObserver` definition, just say `{@link TransportObserver} for the newly created connections`?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -273,7 +273,7 @@ public void onComplete() {
         }
 
         // No connection was selected: create a new one
-        return connectionFactory.newConnection(host.address)
+        return connectionFactory.newConnection(host.address, null)
```
NiteshKant, line 276: Add a comment why we are passing `null` here?

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionFactory.java
idelpivnitskiy: Good question!
I plan to add a `BiTransportObserver` implementation later that will merge two observers into one, when necessary. Something similar to [BiDnsQueryLifecycleObserverFactory](https://github.com/netty/netty/blob/4.1/resolver-dns/src/main/java/io/netty/resolver/dns/BiDnsQueryLifecycleObserverFactory.java) in netty.

idelpivnitskiy: Added `TransportObserverConnectionFactoryFilter` that uses `BiTransportObserver` internally, ptal.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/BiTransportObserver.java
```diff
@@ -0,0 +1,351 @@
...
+    private final TransportObserver second;
+
+    /**
+     * Creates a new instance.
+     *
+     * @param first the {@link TransportObserver} that will receive events first
+     * @param second the {@link TransportObserver} that will receive events second
+     */
+    BiTransportObserver(final TransportObserver first, final TransportObserver second) {
+        this.first = first;
```
NiteshKant, line 45: `requireNotNull`?

servicetalk-client-api/src/main/java/io/servicetalk/client/api/BiTransportObserver.java
```diff
@@ -0,0 +1,351 @@
...
+            this.first = first;
+            this.second = second;
+        }
+
+        @Override
+        public void onDataRead(final int size) {
+            try {
+                first.onDataRead(size);
+            } finally {
+                second.onDataRead(size);
```
NiteshKant, line 69: If `first` and `second` both throw then because `second` is called from a `finally` block, the first exception is lost. An alternative is to add separate `try-catch` blocks for each observer which also extends generically to `n` observers.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/TransportObserverConnectionFactoryFilter.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public ConnectionFactory<ResolvedAddress, C> create(final ConnectionFactory<ResolvedAddress, C> original) {
+        return new DelegatingConnectionFactory<ResolvedAddress, C>(original) {
+            @Override
+            public Single<C> newConnection(final ResolvedAddress resolvedAddress,
+                                           @Nullable final TransportObserver originalObserver) {
+                return defer(() -> {
+                    final TransportObserver newObserver = requireNonNull(observerFactory.apply(resolvedAddress));
```
NiteshKant, line 66: Is creating a `TransportObserver` per subscribe what we want semantically? It seems to be conflicting with the fact that the argument to `newConnection()` is a `TransportObserver` and not a factory.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/BiTransportObserver.java
idelpivnitskiy: I wanted to avoid safety guarantees for this implementation. Do you mind if I remove all try-catch blocks and will assume that first/second observers are never throw?
They can be wrapped by `CatchAllTransportObserver` impl as we discussed before.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/TransportObserverConnectionFactoryFilter.java
idelpivnitskiy: `TransportObserver` technically is a factory for `ConnectionObserver`s. Therefore, I don't see a strong need to provide another level of abstraction for that internally in the transport.
Here we also have access to `ResolvedAddress` and it's an opportunity for us to let users decide if they need a `TransportObserver` for this address or not and what instance of `TransportObserver` they need.
Btw, let me remote the `requireNonNull` from here and allow `null` values.

idelpivnitskiy: Actually, we don't need to create a new `TransportObserver` per subscribe. We can remove `defer` operator.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/BiTransportObserver.java
idelpivnitskiy: This is an internal class and we do null-check upfront. At this time caller code does not expect any exception.

NiteshKant: I think you wanted to make it `public` at some point? It is your call but standalone pig-private classes are good to be coded as if one day they will be `public`. If they are inner-class then the assumptions are valid.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/BiTransportObserver.java
NiteshKant: ok that sounds fair

PR #1115, opened 2020-08-02
Retain addresses on service discovery failure

NiteshKant:
__Motivation__

Service discovery errors are retried by clients to avoid transient failures cause request errors. DNS service discoverer eagerly clears addresses for errors of type `UnknownHostException` but such errors can also be transient. This DNS discoverer behavior effectively defeats the retries done by the client as it results in request errors.
We should avoid service discovery implementations to react pessimistically to service discovery errors as service discovery results are just a hint and do not effectively represent the current state of the servers. Server availability may change asynchronously and may not accurately be reflected in service discovery result due to delays at multiple levels. Instead, we should assume that all service discovery errors are transient and retain the addresses till a subsequent success from service discovery.

__Modification__

Introduced a new contract for `ServiceDiscoveryRetryStrategy` and added a default implementation that retains addresses between an SD error and success. Old addresses are retained till a configured percentage of new addresses are removed post retry. This is done because service discovery events are emitted one at a time and hence does not provide a natural boundary which indicates completion of a successful resolution. Users can tune this percentage; the default is 75%.

Removed DNS service discoverer behavior of clearing addresses on certain errors.

__Result__

Clients better tolerate transient service discoverer errors irrespective of individual service discoverer implementation.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultServiceDiscoveryRetryStrategy.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultServiceDiscoveryRetryStrategy.java
@@ -36,6 +36,7 @@ import java.util.function.UnaryOperator;
 import static io.servicetalk.concurrent.api.Publisher.defer;
 import static io.servicetalk.concurrent.api.RetryStrategies.retryWithExponentialBackoffAndJitter;
 import static java.lang.Math.ceil;
+import static java.lang.Math.max;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonList;
@@ -112,6 +113,15 @@ public final class DefaultServiceDiscoveryRetryStrategy<ResolvedAddress,
             return this;
         }
 
+        /**
+         * Specifies a {@link BiFunction} which is applied as-is using {@link Publisher#retryWhen(BiIntFunction)}
+         * on the {@link Publisher} passed to {@link DefaultServiceDiscoveryRetryStrategy#apply(Publisher)}.
+         *
+         * @param retryStrategy A {@link BiFunction} which is applied as-is using
+         * {@link Publisher#retryWhen(BiIntFunction)} on the {@link Publisher} passed to
+         * {@link DefaultServiceDiscoveryRetryStrategy#apply(Publisher)}.
+         * @return {@code this}.
+         */
         public Builder<ResolvedAddress, E> retryStrategy(
                 final BiIntFunction<Throwable, ? extends Completable> retryStrategy) {
             this.retryStrategy = requireNonNull(retryStrategy);
@@ -119,9 +129,7 @@ public final class DefaultServiceDiscoveryRetryStrategy<ResolvedAddress,
         }
 
         /**
-         * Creates a new {@link ServiceDiscoveryRetryStrategy} using the passed {@link BiFunction} which is applied
-         * as-is using {@link Publisher#retryWhen(BiIntFunction)} on the {@link Publisher} passed to
-         * {@link DefaultServiceDiscoveryRetryStrategy#apply(Publisher)}.
+         * Creates a new {@link ServiceDiscoveryRetryStrategy}.
          *
          * @return A new {@link ServiceDiscoveryRetryStrategy}.
          */
@@ -232,7 +240,7 @@ public final class DefaultServiceDiscoveryRetryStrategy<ResolvedAddress,
                 // new address after a retry
                 activeAddresses.put(address, event);
                 final boolean removed = retainedAddresses.remove(address) != null;
-                if (activeAddresses.size() == targetSize) {
+                if (activeAddresses.size() >= targetSize) {
                     final List<E> allEvents = new ArrayList<>(retainedAddresses.size() + (removed ? 0 : 1));
                     if (!removed) {
                         allEvents.add(event);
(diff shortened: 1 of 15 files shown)


servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultServiceDiscoveryRetryStrategyTest.java
```diff
@@ -0,0 +1,207 @@
...
+    public final Timeout timeout = new ServiceTalkTestTimeout();
+
+    private final TestExecutor executor;
+    private final LinkedBlockingQueue<TestPublisher<ServiceDiscovererEvent<String>>> pubs;
+    private final TestPublisherSubscriber<ServiceDiscovererEvent<String>> subscriber;
+
+    public DefaultServiceDiscoveryRetryStrategyTest() {
+        executor = new TestExecutor();
+        ServiceDiscoveryRetryStrategy<String, ServiceDiscovererEvent<String>> strategy =
+                Builder.<String>withDefaults(executor, ofSeconds(1)).retainAddressesTillSuccess(75).build();
```
idelpivnitskiy, line 56: Consider testing a use-case for `retainAddressesTillSuccess(0)`.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultServiceDiscoveryRetryStrategyTest.java
```diff
@@ -0,0 +1,207 @@
...
+        sdEvents.onNext(evt1); // previously existing, should not be emitted
+        verifyNoEventsReceived();
+
+        sdEvents = triggerRetry(sdEvents); // error while retaining
+
+        final DefaultServiceDiscovererEvent<String> evt3 = new DefaultServiceDiscovererEvent<>("addr3", true);
+        sdEvents.onNext(evt3); // threshold breach, should evict addr2
+
+        assertThat("Unexpected event received", subscriber.takeItems(),
+                containsInAnyOrder(flipAvailable(evt1), flipAvailable(evt2), evt3));
```
idelpivnitskiy, line 132: The comment above says "should evict addr2", but `evt1` is also evicted. Can you please clarify why this is expected?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultServiceDiscoveryRetryStrategy.java
```diff
@@ -261,18 +269,23 @@ void errorSeen() {
...
+            delegate = retryWithExponentialBackoffAndJitter(maxRetries, __ -> true, initialDelay, executor);
+            this.maxRetries = maxRetries;
         }
 
         @Override
         public Completable apply(final int count, final Throwable cause) {
             // As we are retrying indefinitely (unless closed), cap the backoff on MAX_RETRIES retries to avoid
             // impractical backoffs
-            return delegate.apply(count % (MAX_RETRIES - 1), cause);
+            return delegate.apply(max(1, count % (maxRetries)), cause);
```
idelpivnitskiy, line 288: With `max` approach it will return `1` two times. Not a big deal. But if you don't like it, alternative could be `(count % MAX_RETRIES) + 1`.
Also, IIRC, the power of 2 value will be a bit more optimal for %. How about 8 or 16 instead of 10? Or just a mask of `0x0F`?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultServiceDiscoveryRetryStrategy.java
```diff
@@ -0,0 +1,291 @@
...
+         * observed till the total number of received addresses after an error is {@code retainTillReceivePercentage}
+         * percent of the addresses that were available before the error was received.
+         *
+         * @param retainTillReceivePercentage A percentage of addresses that were received before an error that should
+         * be received after a success, after which unreceived addresses are removed by sending appropriate
+         * {@link ServiceDiscovererEvent}s. When set to {@code 0}, addresses are not retained.
+         * @return {@code this}.
+         */
+        public Builder<ResolvedAddress, E> retainAddressesTillSuccess(final int retainTillReceivePercentage) {
+            if (retainTillReceivePercentage < 0) {
```
connect2gaurav, line 108: Minor ask : Requires upper cap as well retainTillReceivePercentage should not be greater than 100

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultServiceDiscoveryRetryStrategy.java
```diff
@@ -0,0 +1,291 @@
...
+            this.flipAvailability = flipAvailability;
+        }
+
+        void errorSeen() {
+            if (retainedAddresses == NONE_RETAINED) {
+                retainedAddresses = new HashMap<>(activeAddresses);
+            } else {
+                retainedAddresses.putAll(activeAddresses);
+            }
+            targetSize = (int) (ceil(retainTillReceivePercentage / 100d) * activeAddresses.size());
```
connect2gaurav, line 223: `System.out.println("Output for 0	:" +(int)Math.ceil(0/100d));
		System.out.println("Output for 10	:" +(int)Math.ceil(10/100d));
		System.out.println("Output for 75	:" +(int)Math.ceil(75/100d));
		System.out.println("Output for 100	:" +(int)Math.ceil(100/100d));`

`Output for 0	:0
Output for 10	:1
Output for 75	:1
Output for 100	:1`

Seems like Percentage computation is bit off. Irrespective of percentage targetSize = 1 * activeAddresses.size()

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultServiceDiscoveryRetryStrategy.java
NiteshKant: Not adding an upper cap is intentional here as it enables folks to retain older addresses for a longish time.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultServiceDiscoveryRetryStrategy.java
NiteshKant: Great catch!

Wrong placement of brackets 😢 

I will fix and add tests, thanks!

PR #1117, opened 2020-08-05
Add missed overrides for the client builders

idelpivnitskiy:
Motivation:

Some methods are not overridden in `BaseSingleAddressHttpClientBuilder`
and `MultiAddressHttpClientBuilder`. Users have to cast if they use
those missed methods on this builders.

Modifications:

- Add overrides for `BaseSingleAddressHttpClientBuilder`: `executionStrategy`
and `appendClientFilter`;
- Add overrides for `MultiAddressHttpClientBuilder.java`: `protocols`;

Result:

Users can use fluent builder API without need to cast builder types.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
@@ -63,6 +63,9 @@ public abstract class MultiAddressHttpClientBuilder<U, R>
     public abstract MultiAddressHttpClientBuilder<U, R> enableWireLogging(String loggerName);
 
     @Override
+    public abstract MultiAddressHttpClientBuilder<U, R> protocols(HttpProtocolConfig... protocols);
+
+    @Override
     public abstract MultiAddressHttpClientBuilder<U, R> disableHostHeaderFallback();
 
     /**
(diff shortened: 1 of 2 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
```diff
@@ -114,8 +117,7 @@
     @Override
     public MultiAddressHttpClientBuilder<U, R> appendClientFilter(final Predicate<StreamingHttpRequest> predicate,
                                                                   final StreamingHttpClientFilterFactory factory) {
-        super.appendClientFilter(predicate, factory);
-        return this;
```
Scottmitch, line 118: can you clarify why the `cast` approach is used here?

idelpivnitskiy: Just wanted to make it consistent to the approach we use in other places:
1. https://github.com/apple/servicetalk/blob/05c28f8339441494f12defece5770dee700d1702/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java#L84
2. https://github.com/apple/servicetalk/blob/05c28f8339441494f12defece5770dee700d1702/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BaseSingleAddressHttpClientBuilder.java#L57
3. https://github.com/apple/servicetalk/blob/05c28f8339441494f12defece5770dee700d1702/servicetalk-http-api/src/main/java/io/servicetalk/http/api/PartitionedHttpClientBuilder.java#L74-L75

PR #1118, opened 2020-08-07
Introduce `CatchAllTransportObserver` that catches and logs unexpected exceptions

idelpivnitskiy:
Motivation:

We have to make sure that all invocations of observer methods are safe and do not
impact normal ST operations. Currently, we do that via `safeReport` utilities that
are used everywhere. It's easy to forget to make an invocation safe in all places.

Modifications:

- Add `CatchAllTransportObserver` implementation that makes sure all unexpected
exceptions are logged;
- Add public `TransportObservers` factory that can convert any observer into its safe
version;
- Wrap the passed by users `TransportObserver` with
`TransportObservers#asSafeObserver` instead of using `safeReport` utilities for all
invocations of observer API;

Result:

A single place that makes sure all calls to `TransportObserver` are safe and do not
throw an exception.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CatchAllTransportObserver.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CatchAllTransportObserver.java
@@ -1,0 +1,296 @@
+/*
+ * Copyright © 2020 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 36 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CatchAllTransportObserver.java
```diff
@@ -0,0 +1,296 @@
...
+import java.util.function.Supplier;
+import java.util.function.UnaryOperator;
+import javax.net.ssl.SSLSession;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * {@link TransportObserver} wrapper that catches and logs all exceptions.
+ */
+final class CatchAllTransportObserver implements TransportObserver {
```
idelpivnitskiy, line 47: Should I make it public to be able to use in `TransportObserverConnectionFactoryFilter` with `BiTransportObserver`?

PR #1123, opened 2020-08-12
Set the correct log level for wire logging and h2 frame logging

Scottmitch:
Motivation:
When enableWireLogging and h2 frameLogger are set on the builder we
always log at TRACE level regardless of what level the logging
configuration specifies. The h2 frameLogger has conditional logic which
limits the amount of data logged for levels less than TRACE, and it is
generally unexpected to log at a different level than configured by the
user.

Modifications:
- When Http2FrameLogger and LoggingHandler are created we should get the
log level from SL4J and pass the translated Netty LogLevel upon
construction

Result:
Debug logging uses the same log level as configured by the user.

### DIFF
--- a/servicetalk-examples/grpc/helloworld/src/main/resources/log4j2.xml
+++ b/servicetalk-examples/grpc/helloworld/src/main/resources/log4j2.xml
@@ -21,14 +21,6 @@
     </Console>
   </Appenders>
   <Loggers>
-    <!--
-        This is an example of how to configure logging of wire events:
-          1. Add a Logger with `TRACE` level;
-          2. Use this new Logger name as an argument for `enableWireLogging(name)` method with desired builder/starter.
-    -->
-    <Logger name="servicetalk-client-wire-logger" level="TRACE"/>
-    <Logger name="servicetalk-server-wire-logger" level="TRACE"/>
-
     <!-- Prints server start and shutdown -->
     <Logger name="io.servicetalk.http.netty.H2ServerParentConnectionContext" level="DEBUG"/>
 
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyLoggerUtils.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyLoggerUtils.java
@@ -20,6 +20,8 @@ import io.netty.util.internal.logging.InternalLogger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nullable;
+
 import static io.netty.handler.logging.LogLevel.DEBUG;
 import static io.netty.handler.logging.LogLevel.ERROR;
 import static io.netty.handler.logging.LogLevel.INFO;
@@ -36,8 +38,10 @@ public final class NettyLoggerUtils {
     /**
      * Translate the log level as determined by SL4J for the {@code loggerName} to Netty's {@link LogLevel} type.
      * @param loggerName The logger name to lookup.
-     * @return Netty's {@link LogLevel} corresponding to the SL4J log level for {@code loggerName}.
+     * @return Netty's {@link LogLevel} corresponding to the SL4J log level for {@code loggerName}, or {@code null} if
+     * the log level cannot be mapped to a {@link LogLevel}.
      */
+    @Nullable
     public static LogLevel getNettyLogLevel(String loggerName) {
         final Logger logger = LoggerFactory.getLogger(loggerName);
         if (logger.isTraceEnabled()) {
@@ -51,6 +55,6 @@ public final class NettyLoggerUtils {
         } else if (logger.isErrorEnabled()) {
             return ERROR;
         }
-        throw new IllegalArgumentException("unknown log level: " + logger);
+        return null;
     }
 }
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyLoggerUtils.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyLoggerUtils.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright © 2020 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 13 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyLoggerUtils.java
```diff
@@ -0,0 +1,56 @@
...
+public final class NettyLoggerUtils {
+    private NettyLoggerUtils() {
+    }
+
+    /**
+     * Translate the log level as determined by SL4J for the {@code loggerName} to Netty's {@link LogLevel} type.
+     * @param loggerName The logger name to lookup.
+     * @return Netty's {@link LogLevel} corresponding to the SL4J log level for {@code loggerName}.
+     */
+    public static LogLevel getNettyLogLevel(String loggerName) {
```
Scottmitch, line 41: this API was designed to avoid exposing SL4J types in our public API, at the cost of insuring multiple logger lookups (one in this method, one done by Netty via `InternalLoggerFactory`).

idelpivnitskiy: sounds good to me, should not harm at the build time

NiteshKant, line 41: A behavior that is useful is to enable wire logging in code but disable it using logger levels. I *think* this will always enable wire logging if it is enabled in code, is that correct?

idelpivnitskiy: Good question! Most of the logger implementations also support `OFF` that disables logging [1, 2] from the logger config. But we need to adjust `getNettyLogLevel` method to support it.

1. https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Level.html#OFF
2. http://logback.qos.ch/apidocs/ch/qos/logback/classic/Level.html#OFF

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyLoggerUtils.java
```diff
@@ -0,0 +1,56 @@
...
+        } else if (logger.isDebugEnabled()) {
+            return DEBUG;
+        } else if (logger.isInfoEnabled()) {
+            return INFO;
+        } else if (logger.isWarnEnabled()) {
+            return WARN;
+        } else if (logger.isErrorEnabled()) {
+            return ERROR;
+        }
+        throw new IllegalArgumentException("unknown log level: " + logger);
```
idelpivnitskiy, line 54: As discussed in https://github.com/apple/servicetalk/pull/1123#discussion_r478584141 the logger config can disable logging. We should support it instead of throwing an exception.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyLoggerUtils.java
Scottmitch: The configuration can be modified to accommodate this, an appender can include a threshold level. I will enhance our test `log4j2-test.xml` to accommodate this.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyLoggerUtils.java
Scottmitch: good point, done.

servicetalk-examples/grpc/helloworld/src/main/resources/log4j2.xml
```diff
@@ -21,14 +21,6 @@
     </Console>
   </Appenders>
   <Loggers>
-    <!--
-        This is an example of how to configure logging of wire events:
-          1. Add a Logger with `TRACE` level;
```
Scottmitch, line 26: since there is no more strict requirement on the log level, we have an opportunity to simplify the examples. Log4j can be configured in many ways (filters, multiple appenders, using different loggers, etc.) and I don't think this is central to understanding servicetalk.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyLoggerUtils.java
NiteshKant: I am not sure I am following the motivation now. Threshold filter is specific to log4j, so not ideal that we have to depend on that for the expected behavior.

Why not just let the user specify the level at which they want to log?

Scottmitch: Threshold filter was one example of how this can be handled in configuration, setting the logging value to OFF is another example. Specifying the log level programmatically requires additional API surface area (e.g. log level API, propagation of log level) which seems redundant if the configuration has to be setup to enable the logger.

idelpivnitskiy (approved):
👍 
Like new approach. It gives more flexibility. For example, in tests we don't need to worry about removing `enableWireLogging` before committing changes (always was the pain) because the wire logger is off by default and controlled by system property.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WireLoggingInitializer.java
```diff
@@ -34,11 +37,14 @@
      * @param loggerName The name of the logger to log wire events.
      */
     public WireLoggingInitializer(final String loggerName) {
-        loggingHandler = new LoggingHandler(loggerName, TRACE);
+        LogLevel logLevel = getNettyLogLevel(loggerName);
+        loggingHandler = logLevel != null ? new LoggingHandler(loggerName, logLevel) : null;
```
NiteshKant, line 41: Just a suggestion; feel free to ignore.

This approach precludes dynamic log level manipulation for existing channels. I think we should always add the handler if user opts-in to it using the builder setting.

NiteshKant (approved):

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WireLoggingInitializer.java
Scottmitch: `logLevel  == null` will only if the logging level can't be mapped to a Netty `LogLevel` (e.g. "OFF"). The user's logging config can set the logging level set to a valid level (e.g. not "OFF") and use a filter to achieve the dynamic enable/disable of logs on pre-existing channels.

PR #1124, opened 2020-08-12
Propagate ConnectionObserver without channel attribute

idelpivnitskiy:
Motivation:

Usage of `Channel`'s attributes increase a shared state of the `Channel`.
`ConnectionObserver` can be propagated without using it.

Modifications:

- Report `onNewConnection()` inside `connectionFactory` functions;
- Rename `TransportObserverInitializer` -> `ConnectionObserverInitializer`;
- Pass the created `ConnectionObserver` to the first and last netty handlers;
- Use `ChannelCloseUtils` to register a reason exception for `Channel` closure;
- Improve tests related to stream and connection closure;

Result:

Less state on the `Channel` instance, better handling of the `Channel` close reason.


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ObservabilityProvider.java
```diff
@@ -0,0 +1,79 @@
...
+import io.servicetalk.transport.api.ConnectionObserver;
+import io.servicetalk.transport.api.ConnectionObserver.SecurityHandshakeObserver;
+import io.servicetalk.transport.api.TransportObserver;
+
+import javax.annotation.Nullable;
+
+/**
+ * A helper class which allows sharing observers between different layers and handlers.
+ */
+public final class ObservabilityProvider {
```
NiteshKant, line 27: I do not quite understand the reasoning behind this class. `connectionObserver` and `handshakeObserver` references are temporary, i.e. they can only be retrieved once which makes the contract very error prone and not very different than the connection attribute approach. My hope was that we can be explicit in contract while passing these objects between classes, can we try doing that?

idelpivnitskiy: Updated according to the approach we discussed offline. Looks a lot better now!
I will kill `SECURITY_HANDSHAKE_OBSERVER` in a follow-up PR.
Will keep `ChannelCloseUtils` here, because it's related to the `ConnectionObserver` and there was a bug in tests that is fixed now.

idelpivnitskiy:

PR #1128, opened 2020-08-18
Improve tests for ALPN

idelpivnitskiy:
Motivation:

Assertions from non-main thread (e.g. inside the service handle method)
do not fail the test.
Also, existing tests do not verify that `SSLSession` is set on the client
side.

Modifications:

- Assert server-side state in the main thread;
- Assert that `ConnectionContext` is set correctly on the client-side;
- Avoid using deprecated jUnit API;

Result:

Better test coverage for ALPN use cases.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/AlpnClientAndServerTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/AlpnClientAndServerTest.java
@@ -38,7 +38,8 @@ import org.junit.runners.Parameterized.Parameters;
 import java.nio.channels.ClosedChannelException;
 import java.util.Collection;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingDeque;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.http.api.HttpResponseStatus.OK;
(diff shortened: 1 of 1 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/AlpnClientAndServerTest.java
```diff
@@ -110,19 +112,16 @@ public AlpnClientAndServerTest(List<String> serverSideProtocols,
...
                 .secure()
                 .provider(OPENSSL)
                 .commit(DefaultTestCerts::loadServerPem, DefaultTestCerts::loadServerKey)
                 .listenBlocking((ctx, request, responseFactory) -> {
-                    if ("PRI".equals(request.method().name()) && expectedProtocol == null) {
-                        return responseFactory.badRequest();
-                    }
-                    assertThat(request.version(), is(expectedProtocol));
-                    assertThat(ctx.sslSession(), is(notNullValue()));
+                    serviceContext.set(ctx);
```
NiteshKant, line 123: Overwrites like this causes debugging pain if inadvertently multiple requests are sent to the service. 
I find it useful to instead use a `BlockingQueue` to track requests which gives an additional benefit of being able to wait for a request to be received by the server. However, YMMV

idelpivnitskiy: Thanks for reminder, `BlockingQueue` is better!

PR #1129, opened 2020-08-19
Propagate `SecurityHandshakeObserver` without channel attribute

idelpivnitskiy:
Motivation:

Usage of Channel's attributes increase a shared state of the `Channel`.
`SecurityHandshakeObserver` can be propagated without using it.

Modifications:

- When `SslHandshakeCompletionEvent` arrives, look up for
`ConnectionObserverHandler` to access previously initialized
`SecurityHandshakeObserver`;
- Add `SecurityHandshakeObserverTest` to verify that
`SecurityHandshakeObserver` is working with different configurations:
HTTP/1.1, HTTP/2, ALPN, secure proxy tunnel;
- Improve `TestServiceStreaming` to echo content-type header;

Result:

Less state on the `Channel` instance.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
@@ -32,7 +32,6 @@ import io.servicetalk.concurrent.internal.DelayedCancellable;
 import io.servicetalk.transport.api.ConnectionObserver;
 import io.servicetalk.transport.api.ConnectionObserver.DataObserver;
 import io.servicetalk.transport.api.ConnectionObserver.ReadObserver;
-import io.servicetalk.transport.api.ConnectionObserver.SecurityHandshakeObserver;
 import io.servicetalk.transport.api.ConnectionObserver.StreamObserver;
 import io.servicetalk.transport.api.ConnectionObserver.WriteObserver;
 import io.servicetalk.transport.api.DefaultExecutionContext;
@@ -42,7 +41,6 @@ import io.servicetalk.transport.api.ServiceTalkSocketOptions;
 import io.servicetalk.transport.netty.internal.CloseHandler.AbortWritesEvent;
 import io.servicetalk.transport.netty.internal.CloseHandler.CloseEvent;
 import io.servicetalk.transport.netty.internal.CloseHandler.CloseEventObservedException;
-import io.servicetalk.transport.netty.internal.ConnectionObserverInitializer.WaitingForHandshakeCompletionEvent;
 import io.servicetalk.transport.netty.internal.WriteStreamSubscriber.AbortedFirstWrite;
 
 import io.netty.channel.Channel;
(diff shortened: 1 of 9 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -672,7 +680,16 @@ public void channelInactive(ChannelHandlerContext ctx) {
...
         private void doChannelActive(ChannelHandlerContext ctx) {
+            if (active) {
+                return;
+            }
+            active = true;
+
             if (waitForSslHandshake) {
+                if (observer != null) {
+                    // Notify ConnectionObserverHandler that we are ready to receive SecurityHandshakeObserver instance
+                    ctx.pipeline().fireUserEventTriggered(WaitingForHandshakeCompletionEvent.INSTANCE);
```
NiteshKant, line 691: A negative of sending an event is that there is no acknowledgment of receipt which makes debugging harder.
An alternative is to follow what we do for `DeferSslHandler` => lookup the handler and register interest for receiving the observer. The negative is more public API (internal so arguably low) but gives a more determinsitic behavior.

idelpivnitskiy: Good suggestion! Everything is pkg-private without public exposure 🎉

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
```diff
@@ -54,28 +52,29 @@ public static boolean isSslEnabled(ChannelPipeline pipeline) {
...
+        final SecurityHandshakeObserver observer = shouldReport ? handshakeObserver(pipeline) : null;
         if (sslEvent.isSuccess()) {
             final SslHandler sslHandler = pipeline.get(SslHandler.class);
             if (sslHandler != null) {
                 final SSLSession session = sslHandler.engine().getSession();
-                if (securityObserver != null) {
-                    securityObserver.handshakeComplete(session);
+                if (observer != null) {
+                    observer.handshakeComplete(session);
                 }
```
NiteshKant, line 70: Log at `warn` if `shouldReport` is `true`?

idelpivnitskiy: The `warn` log is easy to miss. This is not expected to be caught by users, we should provide a guarantee that it never happens. I'm thinking about an assertion, but as we discussed offline we will avoid using `null` values. Therefore, `shouldReport` will go away in a follow-up.

idelpivnitskiy:

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
NiteshKant: Ok what I meant was that if we expect the observer to be present but it is not; is there something we need to do to bubble it up to the user?

Yes good to guarantee but bugs will make it hard to debug. Perhaps a good item to think about when using NOOP observers.

Issue #1119, opened 2020-08-10
gRPC more informative exceptions for plain h2/h1 responses not following gRPC protocol


Scottmitch:
Description:
Currently we give a general error message if a gRPC client recieves a general h2 response (e.g. from a proxy) which makes it more difficult for users to understand what the problem is (e.g. invalid request, connecting to invalid server, etc…).

Actual Result:
```
io.servicetalk.grpc.api.GrpcStatusException: INTERNAL: Response does not contain grpc-status header or trailer
```

Expected Result:
```
io.servicetalk.grpc.api.GrpcStatusException: INTERNAL: HTTP status code 400
	invalid content-type: text/html
	headers: Metadata(:status=400,server=dlb/1.0.2,date=Fri, 14 Feb 2020 19:50:14 GMT,content-type=text/html,content-length=250,strict-transport-security=max-age=31536000; includeSubDomains;,via=10.38.39.238)
	non-grpc error!
```

Reproducer:
```java
    @Test
    public void serverRespondsWithBadRequestStatus() throws Exception {
        final String responsePayload = "non-grpc error!";
        HttpService streamingService = (ctx, request, responseFactory) -> succeeded(responseFactory.badRequest()
                .payloadBody(responsePayload, HttpSerializationProviders.textSerializer()));
        HttpServerBuilder serverBuilder = HttpServers.forAddress(localAddress(0))
                .protocols(h2Default());
        try (ServerContext serverContext = serverBuilder.listenAndAwait(streamingService)) {
            try (TesterProto.Tester.BlockingTesterClient client =
                         GrpcClients.forAddress(serverHostAndPort(serverContext))
                    .executionStrategy(noOffloadsStrategy())
                    .buildBlocking(new TesterProto.Tester.ClientFactory())) {
                client.test(TesterProto.TestRequest.newBuilder().build());
            }
        }
    }
```

PR #1132, opened 2020-08-21
Provide informative exceptions upon gRPC calls when plain h2/h1 resp

tkountis:
**Motivation:**

When an h1/h2 response comes back, not following the gRPC protocol (eg. due to proxy error), the exception thrown to the client is not informative. It only mentions that a grpc-status wasn't found.
We can improve this to include the HTTP status code along side the header names to help identify what went wrong.

**Modifications:**

Introduced a new function that validates the content type. If an HTML type is found (text/plain or text/html) we throw the new more informative error. The check happens both for headers & trailers.

**Result:**

Before
```
io.servicetalk.grpc.api.GrpcStatusException: INTERNAL: Response does not contain grpc-status header or trailer
```

After
```
io.servicetalk.grpc.api.GrpcStatusException: INTERNAL: HTTP  status code: 400 Bad Request
	invalid content-type: text/plain; charset=UTF-8
	headers: NettyH2HeadersToHttpHeaders[content-type: <filtered>
content-length: <filtered>
transfer-encoding: <filtered>]
```

Fixes: https://github.com/apple/servicetalk/issues/1119


idelpivnitskiy (approved):
Looks good!

NiteshKant (approved):

PR #1134, opened 2020-08-22
Avoid using nullable fields for observers

idelpivnitskiy:
Motivation:

`@Nullable` fields make the internal control flow more complex,
as we always have to checks for `null` before using the field.

Modifications:

- Use NOOP implementation of observers instead of handling `null`
everywhere;

Result:

Simplified internal control flow.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
@@ -640,7 +640,7 @@ public final class DefaultNettyConnection<Read, Write> extends NettyChannelListe
                 connection.nettyChannelPublisher.channelInboundClosed();
             } else if (evt instanceof SslHandshakeCompletionEvent) {
                 connection.sslSession = extractSslSessionAndReport(ctx.pipeline(), (SslHandshakeCompletionEvent) evt,
-                        this::tryFailSubscriber);
+                        this::tryFailSubscriber, observer != NoopConnectionObserver.INSTANCE);
                 if (subscriber != null) {
                     assert waitForSslHandshake;
                     completeSubscriber();
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
@@ -41,6 +41,8 @@ import io.servicetalk.transport.api.ServiceTalkSocketOptions;
 import io.servicetalk.transport.netty.internal.CloseHandler.AbortWritesEvent;
 import io.servicetalk.transport.netty.internal.CloseHandler.CloseEvent;
 import io.servicetalk.transport.netty.internal.CloseHandler.CloseEventObservedException;
+import io.servicetalk.transport.netty.internal.NoopTransportObserver.NoopConnectionObserver;
+import io.servicetalk.transport.netty.internal.NoopTransportObserver.NoopDataObserver;
 import io.servicetalk.transport.netty.internal.WriteStreamSubscriber.AbortedFirstWrite;
 
 import io.netty.channel.Channel;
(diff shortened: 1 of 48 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
```diff
@@ -60,7 +60,21 @@ public static SSLSession extractSslSessionAndReport(ChannelPipeline pipeline,
...
-        final SecurityHandshakeObserver observer = shouldReport ? handshakeObserver(pipeline) : null;
+        final SecurityHandshakeObserver observer;
+        if (shouldReport) {
+            try {
+                observer = handshakeObserver(pipeline);
+            } catch (Exception e) {
+                if (!sslEvent.isSuccess()) {
+                    e.addSuppressed(sslEvent.cause());
+                }
+                deliverFailureCause(failureConsumer, e, null);
```
NiteshKant, line 71: We should avoid failing the normal operations due to issues with observability. Doing this make the observability failures non-recoverable which isn't desirable as opposed to continuing with degraded observability.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -641,7 +640,7 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
                 connection.nettyChannelPublisher.channelInboundClosed();
             } else if (evt instanceof SslHandshakeCompletionEvent) {
                 connection.sslSession = extractSslSessionAndReport(ctx.pipeline(), (SslHandshakeCompletionEvent) evt,
-                        this::tryFailSubscriber, observer != null);
+                        this::tryFailSubscriber, observer != NoopConnectionObserver.INSTANCE);
```
NiteshKant, line 643: I was hoping `shouldReport` to go away. AFAIU, it is required because we are adding the `ConnectionObserverHandler` conditionally. Can we add the handler unconditionally?

My motivation here is to remove conditional code based on instance checks which assumes that an observer is homogeneous => an observer can not be partially NOOP.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
idelpivnitskiy: There is nothing that users can do to misconfigure and hit this error. If they pass a custom transport observer, it won't happen. Also, users don't have ways to modify the pipeline.

This is only for us, as developers, to make sure that new protocols that we add do not have misconfiguration around observability. Any test that uses SSL for a new protocol will fail if `shouldReport` is true, but pipeline does not have a handler. It helps to avoid writing an additional test to verify that new protocol reports handshake status to `SecurityHandshakeObserver`.

I don't have a strong opinion here, can remove it if it's not necessary.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
idelpivnitskiy: Because we anyway need to have some NOOP impl, I use an internal `NoopTransportObserver` as a default when no observer is configured by users. For this case we have an opportunity to optimize and skip unnecessary handlers and operators.
For the case when users add a custom observer (even partially NOOP), we don't need to optimize those use-cases. Therefore, a reference check only for out impl is enough.

Instead of adding `ConnectionObserverHandler` unconditionally, we can simplify `extractSslSessionAndReport` and look up for the handler unconditionally. But it will require us to write a test for `SecurityHandshakeObserver` every time we add a new protocol, as described in my previous comment.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
NiteshKant: known errors are avoidable, unknown errors are the problem with such things. Being pessimistic is better, so will suggest removing.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
NiteshKant: > But it will require us to write a test for SecurityHandshakeObserver every time we add a new protocol

As adding new protocols is less frequent, I think it is fair to trade-off code clarity in this case. 

> look up for the handler unconditionally.

sgtm

PR #1135, opened 2020-08-25
Provide a public factory for `BiTransportObserver`

idelpivnitskiy:
Motivation:

Users who need to report events to two `TransportObserver`s simultaneously
on the server-side or who need their own variant of
`TransportObserverConnectionFactoryFilter` will find it useful.

Modifications:

- Move `BiTransportObserver` from `client-api` to `transport-api`;
- Convert passes first and second observers into safe observers;
- Add public `TransportObservers.biTransportObserver` factory;

Result:

Users can easily report events to two `TransportObserver`s.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/TransportObservers.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/TransportObservers.java
@@ -38,4 +38,16 @@ public final class TransportObservers {
         }
         return new CatchAllTransportObserver(observer);
     }
+
+    /**
+     * Combines two {@link TransportObserver}s into a single {@link TransportObserver}.
+     *
+     * @param first the {@link TransportObserver} that will receive events first
+     * @param second the {@link TransportObserver} that will receive events second
+     * @return a {@link TransportObserver} that delegates all invocations to the {@code first} and {@code second}
+     * {@link TransportObserver}s
+     */
+    public static TransportObserver biTransportObserver(final TransportObserver first, final TransportObserver second) {
+        return new BiTransportObserver(first, second);
+    }
 }
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/TransportObservers.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/TransportObservers.java
@@ -36,18 +36,27 @@ public final class TransportObservers {
         if (observer instanceof CatchAllTransportObserver) {
             return observer;
         }
+        if (observer instanceof BiTransportObserver) {
+            // BiTransportObserver is always safe
+            return observer;
+        }
         return new CatchAllTransportObserver(observer);
     }
 
     /**
-     * Combines two {@link TransportObserver}s into a single {@link TransportObserver}.
+     * Combines multiple {@link TransportObserver}s into a single {@link TransportObserver}.
      *
      * @param first the {@link TransportObserver} that will receive events first
      * @param second the {@link TransportObserver} that will receive events second
-     * @return a {@link TransportObserver} that delegates all invocations to the {@code first} and {@code second}
-     * {@link TransportObserver}s
+     * @param others additional {@link TransportObserver}s that will receive events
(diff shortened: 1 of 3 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/TransportObservers.java
```diff
@@ -38,4 +38,16 @@ public static TransportObserver asSafeObserver(final TransportObserver observer)
...
+
+    /**
+     * Combines two {@link TransportObserver}s into a single {@link TransportObserver}.
+     *
+     * @param first the {@link TransportObserver} that will receive events first
+     * @param second the {@link TransportObserver} that will receive events second
+     * @return a {@link TransportObserver} that delegates all invocations to the {@code first} and {@code second}
+     * {@link TransportObserver}s
+     */
+    public static TransportObserver biTransportObserver(final TransportObserver first, final TransportObserver second) {
```
NiteshKant, line 50: For a public API, I would prefer a `combine()` method with a var-arg and if we want, we can optimize for 2 vs many cases.

idelpivnitskiy: Good idea! Added support for multiple

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/TransportObservers.java
```diff
@@ -36,18 +36,27 @@ public static TransportObserver asSafeObserver(final TransportObserver observer)
...
      * @param first the {@link TransportObserver} that will receive events first
      * @param second the {@link TransportObserver} that will receive events second
-     * @return a {@link TransportObserver} that delegates all invocations to the {@code first} and {@code second}
-     * {@link TransportObserver}s
+     * @param others additional {@link TransportObserver}s that will receive events
+     * @return a {@link TransportObserver} that delegates all invocations to the provided {@link TransportObserver}s
      */
-    public static TransportObserver biTransportObserver(final TransportObserver first, final TransportObserver second) {
-        return new BiTransportObserver(first, second);
+    public static TransportObserver combine(final TransportObserver first, final TransportObserver second,
```
NiteshKant, line 54: Why not just have `combine(final TransportObserver... observers)` ?

Also, may be modify the `BiTransportObserver` to accept an array instead of two observers.

idelpivnitskiy: At least two observer are required to combine. Good to highlight that in the API instead of verifying at runtime.

Having an array inside `BiTransportObserver` will require an array inside every observer level, producing boilerplate at every level to traverse the array. Let's start with simplified approach and we can reconsider if necessary.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/TransportObservers.java
```diff
@@ -36,6 +38,31 @@ public static TransportObserver asSafeObserver(final TransportObserver observer)
...
+    /**
+     * Combines multiple {@link TransportObserver}s into a single {@link TransportObserver}.
+     *
+     * @param other {@link TransportObserver}s to combine
+     * @return a {@link TransportObserver} that delegates all invocations to the provided {@link TransportObserver}s
+     */
+    public static TransportObserver combine(final TransportObserver... other) {
+        switch (other.length) {
+            case 0:
+                throw new IllegalArgumentException("At least one TransportObserver is required to combine");
```
idelpivnitskiy, line 57: This is what I was trying to avoid by having `first, second, others` arguments: http://jtechies.blogspot.com/2012/07/item-42-use-varargs-judiciously.html

NiteshKant: I think this is OK and inline with what we do elsewhere:

https://github.com/apple/servicetalk/blob/main/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MergeCompletable.java#L41

PR #1136, opened 2020-08-25
Transport observability

idelpivnitskiy:
Motivation:

Provide visibility into internal transport-level events.

Modifications:

- Introduce new `TransportObserver` interfaces to consume events from the
transport layer;
- Configure `TransportObserver` via builder method on the server-side;
- Configure `TransportObserver` via connection factory filter the client-side;
- Tests to verify new functionality;

Result:

Visibility into transport-level events.


idelpivnitskiy:
Rebased version of this branch: https://github.com/apple/servicetalk/pull/1138

PR #1142, opened 2020-09-04
Improve wrapping with the close reason

idelpivnitskiy:
Motivation:

We have two places in `DefaultNettyConnection` when we may need to wrap
the error with known close reason. We can share this logic in both places
consistently.

Modifications:

- Use `wrapIfReasonIsKnown` for `RetryableClosureException`;
- Improve `wrapIfReasonIsKnown` to skip wrapping if the error is already
wrapped with the same reason;

Result:

Consistent wrapping.


NiteshKant (approved):

idelpivnitskiy:

PR #1144, opened 2020-09-05
Simplify ServiceDiscoverer generics in client builders

NiteshKant:
__Motivation__

`SingleAddressHttpClientBuilder` is used as a template for partitioned and multi-address builders for which SD event types is different. Although, users are not expected to use a ServiceDiscoverer with different event type, we accommodate this usage in our public APIs. Accomodating this usage further trickles down to builder implementation and the external facing `ServiceDiscoveryRetryStrategy`. It also complicates generic handling whenever we change `ServiceDiscoverer` APIs.
Further the partitioned builder uncheck casts the service discovery stream while using which hides type issues.

__Modification__

 - Remove the `? extends` in the `ServiceDiscoverer` event type while accepting a `ServiceDiscoverer` in the builders. When a different event type is required (partitioned builder), store the `ServiceDiscoverer` locally.
 - Simplify `ServiceDiscoveryRetryStrategy` generics.

 __Result__

 Less generics clutter in public APIs.

### DIFF
--- a/servicetalk-client-api-internal/src/main/java/io/servicetalk/client/api/internal/DefaultPartitionedClientGroup.java
+++ b/servicetalk-client-api-internal/src/main/java/io/servicetalk/client/api/internal/DefaultPartitionedClientGroup.java
@@ -48,6 +48,7 @@ import static io.servicetalk.concurrent.api.AsyncCloseables.emptyAsyncCloseable;
 import static io.servicetalk.concurrent.api.SourceAdapters.toSource;
 import static io.servicetalk.concurrent.internal.SubscriberUtils.deliverCompleteFromSource;
 import static java.util.Objects.requireNonNull;
+import static java.util.function.Function.identity;
 
 /**
  * An implementation of {@link ClientGroup} that can be used for partitioned client use-cases where {@link
@@ -182,7 +183,7 @@ public final class DefaultPartitionedClientGroup<U, R, Client extends Listenable
                     }
                     return acceptEvent;
                 }
-            }).beforeFinally(partition::closeNow).map(psde -> psde);
+            }).beforeFinally(partition::closeNow).map(identity());
         }
 
         @Override
--- a/servicetalk-client-api-internal/src/main/java/io/servicetalk/client/api/internal/DefaultPartitionedClientGroup.java
+++ b/servicetalk-client-api-internal/src/main/java/io/servicetalk/client/api/internal/DefaultPartitionedClientGroup.java
@@ -17,6 +17,7 @@ package io.servicetalk.client.api.internal;
 
 import io.servicetalk.client.api.ClientGroup;
 import io.servicetalk.client.api.ServiceDiscoverer;
+import io.servicetalk.client.api.ServiceDiscovererEvent;
 import io.servicetalk.client.api.partition.PartitionAttributes;
 import io.servicetalk.client.api.partition.PartitionMap;
 import io.servicetalk.client.api.partition.PartitionMapFactory;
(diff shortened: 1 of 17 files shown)


servicetalk-client-api-internal/src/main/java/io/servicetalk/client/api/internal/DefaultPartitionedClientGroup.java
```diff
@@ -74,8 +75,7 @@
          * @param psd the partitioned {@link ServiceDiscoverer}
          * @return new client for the given arguments
          */
-        Client apply(PartitionAttributes pa,
-                     ServiceDiscoverer<U, R, ? extends PartitionedServiceDiscovererEvent<R>> psd);
+        Client apply(PartitionAttributes pa, ServiceDiscoverer<U, R, ServiceDiscovererEvent<R>> psd);
```
idelpivnitskiy, line 78: We can also kill `? extends ` in ctor argument (line 101) and in `GroupedByPartitionSubscriber` (lines 258, 278).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -263,8 +252,9 @@ public StreamingHttpClient buildStreaming() {
         // Track resources that potentially need to be closed when an exception is thrown during buildStreaming
         final CompositeCloseable closeOnException = newCompositeCloseable();
         try {
-            final SdStatusCompletable sdStatus = new SdStatusCompletable();
-            final Publisher<? extends ServiceDiscovererEvent<R>> sdEvents = ctx.discover(sdStatus);
+
+            final Publisher<? extends ServiceDiscovererEvent<R>> sdEvents =
```
idelpivnitskiy, line 256: Kill `? extends` here as well?

idelpivnitskiy (approved):
Overall LGTM!
There are still a few places that use `? extends ServiceDiscovererEvent` pattern: LB-related classes and `DefaultDnsClient`. Do you plan to update those places as well?

servicetalk-client-api-internal/src/main/java/io/servicetalk/client/api/internal/DefaultPartitionedClientGroup.java
```diff
@@ -181,7 +182,7 @@ public boolean test(PSDE evt) {
                     }
                     return acceptEvent;
                 }
-            }).beforeFinally(partition::closeNow);
+            }).beforeFinally(partition::closeNow).map(psde -> psde);
```
idelpivnitskiy, line 185: Consider using `Function.identity()`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
```diff
@@ -292,11 +297,33 @@ private HttpClients() {
...
+                            private final ListenableAsyncCloseable closeable = emptyAsyncCloseable();
+
+                            @Override
+                            public Publisher<ServiceDiscovererEvent<R>> discover(final U u) {
+                                return failed(new IllegalStateException("Invalid service discoverer."));
+                            }
+
+                            @Override
+                            public Completable onClose() {
+                                return closeable.onClose();
```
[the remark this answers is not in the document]
NiteshKant: `emptyAsyncCloseable()` makes sure that `onClose()` does not complete till the closeable is closed. Using `completed()` will violate that constraint.

NiteshKant:
> There are still a few places that use ? extends ServiceDiscovererEvent pattern: LB-related classes and DefaultDnsClient. Do you plan to update those places as well?

I do not think we should change that for cases where we take the `Publisher<SDE>`, the builder usage is for `ServiceDiscoverer` itself.

NiteshKant:
Jenkins setup failure.

PR #1146, opened 2020-09-09
Support batched service discoverer updates

NiteshKant:
__Motivation__

`ServiceDiscoverer#discover()` returns a `Publisher<ServiceDiscovererEvent>` where each `ServiceDiscovererEvent` contains the state of a single address. Clients always retry service discovery failures assuming transient failures but try to hold on to the last known addresses till a subsequent successful `discover()` call.  Upon a subsequent successful call; the client then has to decide after how many addresses should the old cache be disposed to reduce churn associated by deactivating an address and then subsequently activating again. Today this is done with heuristics around percentage address received between failed and successful discover call.

If we implement batch service discoverer events where a ServiceDiscoverer implementation can send all known addresses in one batch then the above mentioned heuristics can be removed.

__Modification__

- Modify `ServiceDiscoverer#discover()` to return a `Publisher<List<ServiceDiscovererEvent<ResolvedAddress>>>`
- Modify `DefaultServiceDiscoveryRetryStrategy` to enable/disable retention of addresses between retries as opposed to the percent based heuristics.

__Result__

Deterministic caching of addresses on service discoverer failures.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
@@ -19,6 +19,8 @@ import io.servicetalk.concurrent.PublisherSource.Subscription;
 import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
 import io.servicetalk.concurrent.api.Publisher;
 
+import java.util.List;
+
 /**
  * Represents the interaction pattern with a service discovery system. It is assumed that once {@link #discover(Object)}
  * is called that the service discovery system will push data updates or implementations of this interface will poll for
@@ -47,5 +49,5 @@ public interface ServiceDiscoverer<UnresolvedAddress, ResolvedAddress,
      * </ul>
      * @return a {@link Publisher} that represents a stream of events from the service discovery system.
      */
-    Publisher<E> discover(UnresolvedAddress address);
+    Publisher<List<E>> discover(UnresolvedAddress address);
 }
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
@@ -19,7 +19,7 @@ import io.servicetalk.concurrent.PublisherSource.Subscription;
 import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
 import io.servicetalk.concurrent.api.Publisher;
 
-import java.util.List;
+import java.util.Collection;
 
 /**
  * Represents the interaction pattern with a service discovery system. It is assumed that once {@link #discover(Object)}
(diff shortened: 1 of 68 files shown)


NiteshKant:
Depends on #1144

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
```diff
@@ -47,5 +49,5 @@
      * </ul>
      * @return a {@link Publisher} that represents a stream of events from the service discovery system.
      */
-    Publisher<E> discover(UnresolvedAddress address);
+    Publisher<List<E>> discover(UnresolvedAddress address);
```
idelpivnitskiy, line 52: Do we need a `List`? Will `Collection` be enough?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -263,8 +255,9 @@ public StreamingHttpClient buildStreaming() {
         // Track resources that potentially need to be closed when an exception is thrown during buildStreaming
         final CompositeCloseable closeOnException = newCompositeCloseable();
         try {
-            final SdStatusCompletable sdStatus = new SdStatusCompletable();
-            final Publisher<? extends ServiceDiscovererEvent<R>> sdEvents = ctx.discover(sdStatus);
+
+            final Publisher<ServiceDiscovererEvent<R>> sdEvents =
+                    ctx.serviceDiscoverer.discover(ctx.address()).flatMapConcatIterable(identity());
```
idelpivnitskiy, line 260: If we change public API of SD to emit `List` of addresses, should we also change LB factory API to use the same API for `eventPublisher`? Looks a bit weird do have these API different.

idelpivnitskiy (commented):
LGTM, 2 questions:

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
NiteshKant: Good call, `Collection` is enough

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
NiteshKant: I think sending the `List` to load balancer does not have any value as LBs are not expected to cache, batch SD updates. Passing the `List` makes the API more complex, so I decided not to propagate the changes to the LB layer. If we see a need in the future, it should be straight forward to change.

NiteshKant:
Failure attributed to: https://github.com/apple/servicetalk/issues/1148

idelpivnitskiy (approved):
🚢

PR #1150, opened 2020-09-11
Clear AsyncContext before invoking HttpService

Scottmitch:
Motivation:
Each HttpService should get an empty AsyncContext. This will happen if
the AsyncContext at the root level is empty, which it typically is.
However it is possible that the AsyncContext maybe modified before this
time (e.g. MDC) and if this happens that may result in shared state
between requests (if values are modifiable).

Modification:
- Wrap each service in a filter that explicitly clears the AsyncContext
before executing the user's service filter chain

Result:
AsyncContext is empty when the user service/filter chain is invoked.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamingHttpServiceAsyncContextTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamingHttpServiceAsyncContextTest.java
@@ -87,20 +87,12 @@ public class StreamingHttpServiceAsyncContextTest extends AbstractHttpServiceAsy
     }
 
     private static StreamingHttpService newEmptyAsyncContextService() {
-        // The service should get an empty AsyncContext regardless of what is done outside the service.
-        // There are utilities that may be accessed in a static context or before service initialization that
-        // shouldn't pollute the service's AsyncContext.
-        AsyncContext.current().put(K1, "value");
-
         return (ctx, request, factory) -> {
             request.payloadBodyAndTrailers().ignoreElements().subscribe();
 
             AsyncContextMap current = AsyncContext.current();
             if (!current.isEmpty()) {
-                StreamingHttpResponse response = factory.internalServerError();
-                return succeeded(response.payloadBody(textSerializer()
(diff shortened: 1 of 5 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamingHttpServiceAsyncContextTest.java
```diff
@@ -84,15 +87,24 @@ protected ServerContext serverWithEmptyAsyncContextService(HttpServerBuilder ser
...
             request.payloadBodyAndTrailers().ignoreElements().subscribe();
 
-            if (!AsyncContext.isEmpty()) {
-                return succeeded(factory.internalServerError());
+            AsyncContextMap current = AsyncContext.current();
+            if (!current.isEmpty()) {
+                StreamingHttpResponse response = factory.internalServerError();
+                return succeeded(response.payloadBody(textSerializer()
+                        .serialize(response.headers(), from(current.toString()),
+                                ctx.executionContext().bufferAllocator())));
```
idelpivnitskiy, line 103: This can be simplified to:
```java
return succeeded(factory.internalServerError().payloadBody(from(current.toString()), textSerializer()));
```

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamingHttpServiceAsyncContextTest.java
```diff
@@ -84,15 +87,24 @@ protected ServerContext serverWithEmptyAsyncContextService(HttpServerBuilder ser
     }
 
     private static StreamingHttpService newEmptyAsyncContextService() {
+        // The service should get an empty AsyncContext regardless of what is done outside the service.
+        // There are utilities that may be accessed in a static context or before service initialization that
+        // shouldn't pollute the service's AsyncContext.
+        AsyncContext.current().put(K1, "value");
```
idelpivnitskiy, line 93: Consider testing that `<K1, "value">` is still present for the main thread after server started and the request complete.

idelpivnitskiy (approved):

PR #1153, opened 2020-09-16
Unix Domain Sockets HttpClientBuilder fixes

Scottmitch:
Motivation:
UDS is supported by ServicetTalk by passing in a DomainSocketAddress
address. However the HttpClientBuilder had some restrictions/assumptions
about the address type which prevented its use.

Modifications:
- Relax assumptions/restrictions in the HttpClientBuilder to allow for
DomainSocketAddress
- Add unit test to demonstrate UDS
- Add example for UDS

Result:
UDS client is now working and tested.

### DIFF
--- a/servicetalk-examples/http/uds/build.gradle
+++ b/servicetalk-examples/http/uds/build.gradle
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2020 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/servicetalk-examples/http/uds/build.gradle
+++ b/servicetalk-examples/http/uds/build.gradle
@@ -1,0 +1,25 @@
+/*
+ * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: "java"
+apply from: "../../gradle/idea.gradle"
+
+dependencies {
+  implementation project(":servicetalk-annotations")
+  implementation project(":servicetalk-http-netty")
+
+  runtimeOnly "org.apache.logging.log4j:log4j-slf4j-impl:$log4jVersion"
+}
--- a/servicetalk-examples/http/uds/src/main/java/io/servicetalk/examples/http/uds/blocking/BlockingUdsServer.java
+++ b/servicetalk-examples/http/uds/src/main/java/io/servicetalk/examples/http/uds/blocking/BlockingUdsServer.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright © 2018 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.examples.http.uds.blocking;
+
+import io.servicetalk.http.netty.HttpServers;
+import io.servicetalk.transport.api.DomainSocketAddress;
+
+import java.io.File;
+
+import static io.servicetalk.examples.http.uds.blocking.UdsUtils.udsAddress;
+import static io.servicetalk.http.api.HttpSerializationProviders.textSerializer;
+
+/**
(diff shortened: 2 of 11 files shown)


servicetalk-examples/http/uds/build.gradle
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
```
idelpivnitskiy, line 2: Please update the copyright year in all files to 2020.

servicetalk-examples/http/uds/src/main/java/io/servicetalk/examples/http/uds/blocking/BlockingUdsServer.java
```diff
@@ -0,0 +1,41 @@
...
+public final class BlockingUdsServer {
+    public static void main(String[] args) throws Exception {
+        DomainSocketAddress udsAddress = udsAddress();
+        try {
+            HttpServers.forAddress(udsAddress)
+                    .listenBlockingAndAwait((ctx, request, responseFactory) ->
+                            responseFactory.ok().payloadBody("Hello World!", textSerializer()))
+                    .awaitShutdown();
+        } finally {
+            new File(udsAddress.getPath()).delete(); // After the server is done, clean up the file.
```
idelpivnitskiy, line 38: May be in a separate PR: can you please remove `get` prefix from `DomainSocketAddress#getPath()` method for consistency with other API?

servicetalk-examples/http/uds/src/main/java/io/servicetalk/examples/http/uds/blocking/UdsUtils.java
```diff
@@ -0,0 +1,38 @@
...
+     * Create a {@link DomainSocketAddress} that is backed by a fixed file such that different JVMs
+     * can use it to communicate via UDS.
+     * @return a {@link DomainSocketAddress} that is backed by a fixed file such that different JVMs
+     * can use it to communicate via UDS.
+     */
+    public static DomainSocketAddress udsAddress() {
+        final String tempDirProp = "java.io.tmpdir";
+        final String tempDir = System.getProperty(tempDirProp);
+        if (tempDir == null) {
+            throw new IllegalStateException("unable to find " + tempDirProp + " in System properties");
```
idelpivnitskiy, line 34: Consider using `File#createTempFile` instead of requiring a system property to simplify the example.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -495,6 +495,7 @@ HttpExecutionStrategyInfluencer buildStrategyInfluencerForClient(HttpExecutionSt
         return influencerChainBuilder.buildForClient(strategy);
     }
 
+    @Nullable
```
idelpivnitskiy, line 498: Why the `@Nullable` required now?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -179,9 +179,9 @@ private DefaultSingleAddressHttpClientBuilder(@Nullable final U address,
                 globalDnsServiceDiscoverer());
     }
 
-    static <U> DefaultSingleAddressHttpClientBuilder<U, InetSocketAddress> forResolvedAddress(
-            final U u, final InetSocketAddress address) {
-        ServiceDiscoverer<U, InetSocketAddress, ServiceDiscovererEvent<InetSocketAddress>> sd =
+    static <U, R> DefaultSingleAddressHttpClientBuilder<U, R> forResolvedAddress(
```
idelpivnitskiy, line 182: Should `R` extend `SocketAddress`?

idelpivnitskiy (commented):
Good catch!

servicetalk-examples/http/uds/src/main/java/io/servicetalk/examples/http/uds/blocking/UdsUtils.java
tkountis: I believe a fixed name is needed, not a random one, the `File#createTempFile` wouldn't work in that case

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
Scottmitch: done. somewhat on the fence if we need to take a dependency on the marker base class in our APIs, but we should be able to relax the generic constraints later if necessary.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
Scottmitch: good call, not needed. was returning `null` at one point during development.

servicetalk-examples/http/uds/src/main/java/io/servicetalk/examples/http/uds/blocking/UdsUtils.java
Scottmitch: +1 to what @tkountis said. Random file name is not suitable for this use case. The system property is set by the JVM [1] and this code is just a bit defensive.

[1] https://docs.oracle.com/javase/8/docs/api/java/lang/System.html

servicetalk-examples/http/uds/src/main/java/io/servicetalk/examples/http/uds/blocking/BlockingUdsServer.java
Scottmitch: Separate PR, I'm not sure this is worth breaking APIs over.

idelpivnitskiy: Sounds good, looks like `DomainSocketAddress` was not useful before this PR, should be ok to change API.

idelpivnitskiy (approved):
🎖️

PR #1155, opened 2020-09-18
Do not complete server write if there are still pending requests

idelpivnitskiy:
Motivation:

`RequestResponseCloseHandler.protocolPayloadEndOutbound` callback triggers
`ProtocolPayloadEndEvent` when server is in closing state without accounting
for pending requests. As the result, server will not send a response for the
second pipelined request, will not transition to the idle state, and will never
complete close the connection.

Modifications:

- Account for `pending` value before emitting `ProtocolPayloadEndEvent`;
- Renamve `ProtocolPayloadEndEvent` -> `OutboundDataEndEvent`;
- Add a test to verify server does not trigger `OutboundDataEndEvent`
while requests are pending;
- Add more tests to verify that `PROTOCOL_CLOSING_INBOUND`,
`PROTOCOL_CLOSING_OUTBOUND`, and `USER_CLOSING` events are
correctly handled for pipelined server connection;

Result:

Server responds to pending requests and closes the connection if it's already
in closing state while 2+ pipelined requests are in process.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
@@ -165,7 +165,7 @@ class RequestResponseCloseHandler extends CloseHandler {
     @Override
     public void protocolPayloadEndOutbound(final ChannelHandlerContext ctx) {
         if (isClient || (has(state, CLOSING) && pending == 0)) {
-            ctx.pipeline().fireUserEventTriggered(ProtocolPayloadEndEvent.OUTBOUND);
+            ctx.pipeline().fireUserEventTriggered(OutboundDataEndEvent.INSTANCE);
         }
     }
 
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
@@ -164,7 +164,7 @@ class RequestResponseCloseHandler extends CloseHandler {
 
     @Override
     public void protocolPayloadEndOutbound(final ChannelHandlerContext ctx) {
-        if (isClient || has(state, CLOSING)) {
+        if (isClient || (has(state, CLOSING) && pending == 0)) {
             ctx.pipeline().fireUserEventTriggered(ProtocolPayloadEndEvent.OUTBOUND);
         }
     }
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionCloseHeaderHandlingTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionCloseHeaderHandlingTest.java
@@ -17,6 +17,7 @@ package io.servicetalk.http.netty;
 
 import io.servicetalk.buffer.api.Buffer;
 import io.servicetalk.concurrent.BlockingIterator;
+import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.internal.ServiceTalkTestTimeout;
 import io.servicetalk.http.api.HttpPayloadWriter;
 import io.servicetalk.http.api.HttpServerBuilder;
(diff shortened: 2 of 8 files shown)


normanmaurer (approved):

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
```diff
@@ -164,7 +164,7 @@ public void protocolPayloadBeginOutbound(final ChannelHandlerContext ctx) {
 
     @Override
     public void protocolPayloadEndOutbound(final ChannelHandlerContext ctx) {
-        if (isClient || has(state, CLOSING)) {
+        if (isClient || (has(state, CLOSING) && pending == 0)) {
```
NiteshKant, line 167: I think the real problem is that we are not cancelling the `Subscription` when we are prematurely completing the write source in `WriteStreamSubscriber`.  

This event is meant to indicate payload end so I don't think we should further override it to mean "last" payload end.

NiteshKant (changes requested):

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
idelpivnitskiy: You had a similar discussion with Jo a while ago: https://github.com/apple/servicetalk/pull/919#discussion_r366532762
As was said:
> this event is only valuable to determine if we expect any other write on this connection

Taking into account how how the read/write is used on H1 server:
> we have to suppress this event before it reaches WSS in the case of H1 server

The only thing that was not accounted in this statement is that server may already receive 2+ requests, while graceful closure is triggered. Therefore, `pending == 0` is added. Only internal state of `RequestResponseCloseHandler` knows if there are some requests pending or not.

I'm not sure we need to cancel the `Subscription` as we want to continue processing until `pending == 0`.

NiteshKant: This is the write method we are discussing (reducing signature for brevity)

`Completable write(Publisher)`

For this method `WriteStreamSubscriber` is terminating the `Subscriber` of the returned `Completable` but not cancelling the `Subscription` of the `Publisher`. So we have a ` Subscription` which has no way to be cancelled and no more items can be requested. That is the problem why the server does not ever complete as it does not read any more and does not receive any error. 

Re: event semantics

I think that statement is overloaded. I do not think the statement was made w.r.t multiple different `Publisher` (requests) but multiple items from a single `Publisher`.

With this change would we wait for pending responses to be written if the user adds a `Connection: Close` header in one of the responses when more requests are pipelined?

idelpivnitskiy: After further investigation, I can conclude that the same issue happens not only for `USER_CLOSING` case, but also for `PROTOCOL_CLOSING_INBOUND`. If a pipelined server connection already read 2 requests, then users initiated a graceful closure or the 2nd request had `Connection: close` header, server will send a response only for the 1st request. Then `ProtocolPayloadEndEvent` will be emitted that is closing WSS => 2nd response will never send back => connection won't become idle and won't be closed.

For the `PROTOCOL_CLOSING_OUTBOUND` this PR does not change the behavior, because we reset `pending` to 0 in this case:
https://github.com/apple/servicetalk/blob/1aac864ede98705fb62a375193a554855c0b8f81/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java#L267-L274

idelpivnitskiy: I'm agreed that there is also an issue that we do not cancel `Subscription` in WSS. We will need to fix that one as well in a separate PR.

idelpivnitskiy: As we discussed offline, I've remaned the event to `OutboundDataEndEvent`.
Also added tests to verify that all use-cases that we discussed in this thread work correctly.

idelpivnitskiy: > For this method `WriteStreamSubscriber` is terminating the `Subscriber` of the returned `Completable` but not cancelling the `Subscription` of the `Publisher`. So we have a ` Subscription` which has no way to be cancelled and no more items can be requested.

@NiteshKant submitted a follow-up for the WSS issue you described: https://github.com/apple/servicetalk/pull/1160

NiteshKant (approved):

PR #1156, opened 2020-09-18
Do not abort server write if PROTOCOL_CLOSING_INBOUND was already observed

idelpivnitskiy:
Motivation:

If a server receives a request with `Connection: close` header, it registers
`PROTOCOL_CLOSING_INBOUND` event and transitions to the `CLOSING` state.
This event can be observed before the server connection started processing
requests. As the result, server's write stream fails and does not let the server
respond.
This use-case happens frequently when `TLSv1.3` is used, because a client
starts sending data before the server completes TLS handshake. Therefore,
the first request can be pending before the server connection is established.
A similar scenario (request received before processing starts, closing state
waits for the response) may occur with graceful closure or
`CHANNEL_CLOSED_INBOUND` event (client half-closes the connection
after a request is sent).

Modifications:

- Abort new writes only on the client side in `DefaultNettyConnection`;
- Add tests that reproduce described behavior;
- Improve `FlushStrategyOnServerTest` to let other tests reuse its code;

Result:

Server initializes request-response processing and responds to the already
received request.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
@@ -134,13 +134,14 @@ public final class DefaultNettyConnection<Read, Write> extends NettyChannelListe
     @Nullable
     private final ChannelConfig parentChannelConfig;
     private volatile DataObserver dataObserver;
+    private final boolean isClient;
 
     private DefaultNettyConnection(Channel channel, BufferAllocator allocator, Executor executor,
                                    Predicate<Read> terminalPredicate, CloseHandler closeHandler,
                                    FlushStrategy flushStrategy, @Nullable Long idleTimeoutMs,
                                    ExecutionStrategy executionStrategy, Protocol protocol,
                                    @Nullable SSLSession sslSession, @Nullable ChannelConfig parentChannelConfig,
-                                   DataObserver dataObserver) {
+                                   DataObserver dataObserver, boolean isClient) {
         super(channel, executor);
         nettyChannelPublisher = new NettyChannelPublisher<>(channel, terminalPredicate, closeHandler);
         this.readPublisher = registerReadObserver(nettyChannelPublisher.recoverWith(this::enrichErrorPublisher));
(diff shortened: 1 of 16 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -470,10 +474,10 @@ private boolean failIfWriteActive(ChannelOutboundListener newChannelOutboundList
             // It is possible that we have set the writeSubscriber, then the channel becomes inactive, and we will
             // never notify the write writeSubscriber of the inactive event. So if the channel is inactive we notify
             // the writeSubscriber.
-            // It is also possible that Channel is in closing state, we should abort new writes if a closeReason was
-            // observed:
+            // It is also possible that Channel is in closing state, we should abort new writes from the client-side
+            // if a closeReason was observed:
             CloseEvent closeReason = null;
-            if (!channel().isActive() || (closeReason = this.closeReason) != null) {
+            if (!channel().isActive() || (isClient && (closeReason = this.closeReason) != null)) {
```
NiteshKant, line 480: Did you consider delaying the event from `CloseHandler` as that has most context of when it is correct to close. This propagates the close state machine understanding to the connection.

idelpivnitskiy: Interesting idea, let me try

idelpivnitskiy: There is a possibility to handle this scenario in `CloseHandler`, but it requires to remember the original close event and complicates the internal state machine, conflicting with #1141.
I would propose to go with this approach and rework it in a follow-up after #1141 is merged. Or work around this issue in #1141 by using TLSv1.2. It will simplify life a lot if we merge #1141 first, as it changes the state machine and introduces new important tests.

idelpivnitskiy:
Closing this for now.
Will rework and reopen after #1141 is merged.

idelpivnitskiy:
Reopened #1177.

PR #1160, opened 2020-09-29
Cancel Subscription when channel outbound has closed

idelpivnitskiy:
Motivation:

`DefaultNettyConnection` may signal its `ChannelOutboundListener` that the
outbound has closed. `WriteStreamSubscriber` implements
`ChannelOutboundListener`. It terminates the `Subscriber` but does not
cancel `Subscription`.

Modifications:

- Cancel current `Subscription` on
`WriteStreamSubscriber#channelOutboundClosed()`;
- Add tests to verify the behavior;

Result:

`Subscription` receives cancel signal when channel outbound is closed.

### DIFF
--- a/servicetalk-test-resources/src/main/resources/log4j2.xml
+++ b/servicetalk-test-resources/src/main/resources/log4j2.xml
@@ -16,8 +16,8 @@
   -->
 <Configuration status="info">
   <Properties>
-    <Property name="rootLevel">${sys:servicetalk.logger.rootLevel:-INFO}</Property>
-    <Property name="wireLogLevel">${sys:servicetalk.logger.wireLogLevel:-OFF}</Property>
+    <Property name="rootLevel">${sys:servicetalk.logger.rootLevel:-DEBUG}</Property>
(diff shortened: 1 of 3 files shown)


idelpivnitskiy:

idelpivnitskiy:
After investigating the failing `io.servicetalk.http.router.predicate.HttpServerOverrideOffloadingTest.offloadDifferentRoutes` test: we do not need to cancel at `WSS#channelOutboundClosed()` time. This event helps to avoid concurrency between output shutdown event from IO-thread and write publisher completion from an offloaded thread.

On the client-side, `WSS#channelOutboundClosed()` is invoked before encoder writes trailers. There is no need to cancel, because we expect the source to terminate naturally after trailers. If we cancel the subscription, we may cancel the `FLUSH` signal => client won't write the request to the wire.

On the server-side, `WSS#channelOutboundClosed()` is invoked when the connection is in the closing state, after the last required response was received by the encoder. There is also no need to cancel, because after closure is complete, we invoke `WSS#channelClosed(Throwable)` method, which will cancel the subscription.

PR #1163, opened 2020-09-30
servicetalk-grpc-protoc remove plugin.proto

Scottmitch:
Motivation:
At one point we copied the plugin.proto interface which defines the
proto interface for protoc plugins into our project and manually
generated the code. However the PluginProtos is now included in the
protobuf-java dependency and is no longer necessary. Also packaging the
proto file with the servicetalk-grpc-protoc jar file is unnecessary.

Modifications:
- Remove the plugin.proto file from servicetalk-grpc-protoc

Result:
Less unnecessary files in the servicetalk-grpc-protoc.jar file and less
chance for conflict with users who consume this artifact.


idelpivnitskiy (approved):
🚀

PR #1165, opened 2020-10-02
Support half-closure for UnixDomainSockets

idelpivnitskiy:
Motivation:

`CloseHandler.forPipelinedRequestResponse` enables half-closure only for
`SocketChannelConfig`. `DomainSocketChannel` does not use this config.
As the result, half-closure does not work for UnixDomainSockets.

Modifications:

- Use `ALLOW_HALF_CLOSURE` socket option to enable half-closure for the
`Channel`;
- Assert in `RequestResponseCloseHandler` that `Channel` implements
`DuplexChannel` and also has `ALLOW_HALF_CLOSURE` socket option enabled;
- Enhance `ConnectionCloseHeaderHandlingTest` to test the same scenarios
for `DomainSocketAddress`;
- Move `newSocketAddress()` utility to `AddressUtils`;

Result:

Half-closure is enabled for `DomainSocketChannel`.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionCloseHeaderHandlingTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionCloseHeaderHandlingTest.java
@@ -68,6 +68,7 @@ import static io.servicetalk.http.api.HttpSerializationProviders.textSerializer;
 import static io.servicetalk.http.api.Matchers.contentEqualTo;
 import static io.servicetalk.http.netty.HttpsProxyTest.safeClose;
 import static io.servicetalk.transport.netty.internal.AddressUtils.localAddress;
+import static io.servicetalk.transport.netty.internal.AddressUtils.newSocketAddress;
 import static io.servicetalk.transport.netty.internal.AddressUtils.serverHostAndPort;
 import static io.servicetalk.transport.netty.internal.ExecutionContextRule.cached;
 import static io.servicetalk.utils.internal.PlatformDependent.throwException;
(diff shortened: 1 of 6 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionCloseHeaderHandlingTest.java
```diff
@@ -338,14 +359,13 @@ public void serverCloseTwoPipelinedRequestsSentBeforeFirstResponse() throws Exce
             assertResponsePayloadBody(response);
 
             awaitConnectionClosed();
-            // FIXME: temporary disable check for /second until https://github.com/apple/servicetalk/pull/1141
```
Scottmitch, line 341: can we revert this portion now that we have the `@Ignore`?

Scottmitch: (same comment below)

Scottmitch (approved):

idelpivnitskiy:
Build failed due to `validateLocalHtmlLinks` check.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionCloseHeaderHandlingTest.java
idelpivnitskiy: Originally, I thought only the last check is causing the issue. But actually, the response payload body for the first request may also fail.
Decided to uncomment specific assertions and ignore the whole test instead until we have #1141.

PR #1166, opened 2020-10-03
Add an `EmbeddedChannel` that supports `DuplexChannel` interface

idelpivnitskiy:
Motivation:

`RequestResponseCloseHandler` requires a `Channel` to implement `DuplexChannel`
interface. To reliably reproduce some racy behavior on the network we use
`EmbeddedChannel` to control all the network events manually, but it does
not implement `DuplexChannel`. Therefore, we can not use it to verify
the behavior of `RequestResponseCloseHandler`.

Modifications:

- Implement `EmbeddedDuplexChannel` test fixture that extends
`EmbeddedChannel` and implements `DuplexChannel` interface;
- Use a new variant in all tests that require an `EmbeddedChannel` with
`RequestResponseCloseHandler`;
- Update `RequestResponseCloseHandler` to support any `DuplexChannel`
implementations that has `ALLOW_HALF_CLOSURE` socket option enabled;

Result:

Better test coverage for `RequestResponseCloseHandler`, no unexpected
`ClassCastException`(s) on the pipeline. No test-related code in
`RequestResponseCloseHandler`.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
@@ -129,7 +129,7 @@ public class FlushStrategyOnServerTest {
     @After
     public void tearDown() throws Exception {
         try {
-            serverConnection.closeAsyncGracefully().toFuture().get();
+            serverConnection.closeAsync().toFuture().get();
         } finally {
             channel.close().syncUninterruptibly();
         }
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
@@ -29,11 +29,11 @@ import io.servicetalk.http.api.StreamingHttpService;
 import io.servicetalk.http.netty.NettyHttpServer.NettyHttpServerConnection;
 import io.servicetalk.tcp.netty.internal.TcpServerChannelInitializer;
 import io.servicetalk.transport.api.ConnectionObserver;
+import io.servicetalk.transport.netty.internal.EmbeddedDuplexChannel;
 
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelOutboundHandlerAdapter;
 import io.netty.channel.ChannelPromise;
-import io.netty.channel.embedded.EmbeddedChannel;
 import org.junit.After;
 import org.junit.ClassRule;
 import org.junit.Rule;
@@ -78,7 +78,7 @@ public class FlushStrategyOnServerTest {
     public static final ExecutorRule<Executor> EXECUTOR_RULE = newRule();
 
     private final OutboundWriteEventsInterceptor interceptor;
-    private final EmbeddedChannel channel;
+    private final EmbeddedDuplexChannel channel;
     private final AtomicBoolean useAggregatedResponse;
     private final NettyHttpServerConnection serverConnection;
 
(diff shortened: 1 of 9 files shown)


servicetalk-transport-netty-internal/src/testFixtures/java/io/servicetalk/transport/netty/internal/EmbeddedDuplexChannel.java
```diff
@@ -0,0 +1,471 @@
...
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicBoolean;
+import javax.annotation.Nullable;
+
+import static io.netty.channel.ChannelOption.ALLOW_HALF_CLOSURE;
+
+/**
+ * {@link EmbeddedChannel} that implements {@link DuplexChannel}.
+ */
+public final class EmbeddedDuplexChannel extends EmbeddedChannel implements DuplexChannel {
```
idelpivnitskiy, line 50: @normanmaurer is it something you will be interested to see in netty? Implementation can be a bit simplified if we implement `DuplexChannel` for the existing `EmbeddedChannel`, or provide `EmbeddedDuplexChannel` on the same package.

idelpivnitskiy:
This PR is required to test the new behavior of `RequestResponseCloseHandler` in #1141.

servicetalk-transport-netty-internal/src/testFixtures/java/io/servicetalk/transport/netty/internal/EmbeddedDuplexChannel.java
```diff
@@ -0,0 +1,476 @@
...
+        } else if (shutdownInputCause != null) {
+            promise.setFailure(shutdownInputCause);
+        } else {
+            promise.setSuccess();
+        }
+    }
+
+    @Override
+    public Queue<Object> inboundMessages() {
+        if (isInputShutdown.get()) {
```
Scottmitch, line 215: if the input is shutdown real transports will stop producing data (eventually), did you consider force clear/release when shutdown inbound happens to emulate this behavior?

Scottmitch (approved):
few questions comments but nothing that needs to block the PR

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
```diff
@@ -129,7 +129,7 @@ public FlushStrategyOnServerTest(final Param param) throws Exception {
     @After
     public void tearDown() throws Exception {
         try {
-            serverConnection.closeAsyncGracefully().toFuture().get();
+            serverConnection.closeAsync().toFuture().get();
```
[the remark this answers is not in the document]
idelpivnitskiy: I was thinking that in `tearDown()` it's not expected that something is still in process and we should just close. But then I realized that graceful closure will show us if there is some unexpected outstanding data on the pipeline that do not allow us to close. Let me revert it.

servicetalk-transport-netty-internal/src/testFixtures/java/io/servicetalk/transport/netty/internal/EmbeddedDuplexChannel.java
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    @Override
+    public ChannelFuture shutdownInput() {
+        return shutdownInput(newPromise());
+    }
+
+    @Override
+    public ChannelFuture shutdownInput(final ChannelPromise promise) {
+        if (!isOpen()) {
```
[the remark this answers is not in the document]
idelpivnitskiy: `EmbeddedChannel` is not expected to be used from different threads. But this is what we do to verify offloading. If we port code in netty, it should not have atomics. Instead, we should enhance `EmbeddedEventLoop` in netty to account for the thread and use a concurrent queue for tasks.

servicetalk-transport-netty-internal/src/testFixtures/java/io/servicetalk/transport/netty/internal/EmbeddedDuplexChannel.java
idelpivnitskiy: It will stop producing new messages, but it should read all pending messages. This is what I emulate using:
```java
pipeline().fireUserEventTriggered(ChannelInputShutdownEvent.INSTANCE);
super.flushInbound();
pipeline().fireUserEventTriggered(ChannelInputShutdownReadComplete.INSTANCE);
```
After that, I make sure that it can not produce more messages discarding any following attempts to write inbound.

idelpivnitskiy:
Quality build failed due to #1169 

PR #1171, opened 2020-10-09
Support write of read-only `Buffer`(s) for HTTP/2

idelpivnitskiy:
Motivation:

ServiceTalk provides a default `BufferAllocator` backed by netty's
`ByteBufAllocator`, but also provides a `ReadOnlyBufferAllocator`
implementation that does not depend on netty. `AbstractH2DuplexHandler`
currently supports only `ServiceTalkBufferAllocator`.

Modifications:
- Share `encodeAndRetain` logic between HTTP/1.x and HTTP/2 encoders;
- Add a test that verified both HTTP versions support all `BufferAllocator`s;
- Share `HttpProtocol` enum between different tests;
- Implement `ReadOnlyBufferAllocator.toString()` for better output for test
parameters;

Result:

HTTP/2 encoder supports `Buffer`s produced by `ReadOnlyBufferAllocator`.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SupportedBufferAllocatorsTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SupportedBufferAllocatorsTest.java
@@ -17,9 +17,9 @@ package io.servicetalk.http.netty;
 
 import io.servicetalk.buffer.api.BufferAllocator;
 import io.servicetalk.http.api.BlockingHttpService;
-import io.servicetalk.http.api.HttpResponseStatus;
 import io.servicetalk.http.api.StreamingHttpRequest;
 import io.servicetalk.http.api.StreamingHttpResponse;
+import io.servicetalk.http.api.StreamingHttpService;
 
 import org.junit.Test;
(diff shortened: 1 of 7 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SupportedBufferAllocatorsTest.java
```diff
@@ -0,0 +1,86 @@
...
+                strategy -> strategy)).adaptor());
+    }
+
+    @Test
+    public void test() throws Exception {
+        String payload = "Hello ServiceTalk";
+        StreamingHttpRequest request = streamingHttpConnection().post("/")
+                .payloadBody(from(allocator.fromAscii(payload)));
+        StreamingHttpResponse response = makeRequest(request);
+        assertResponse(response, protocol.version, OK, singletonList(payload));
```
Scottmitch, line 84: not introduced by this PR but the validation of the `getBodyAsListOfStrings` doesn't seem right. The way data is written may not necessarily be the way data is read. fixed in https://github.com/apple/servicetalk/pull/1173

Scottmitch (approved):

PR #1175, opened 2020-10-12
BlockingStreamingHttp[Response|Response] to support InputStream setter

Scottmitch:
Motivation:
BlockingStreamingHttp[Response|Response] allow a user to get the payload
body as an InputStream. However the user is not able to set the payload
body as an InputStream. This requires manual conversion from the user
perspective and raises the bar to entry.

Modifications:
- BlockingStreamingHttp[Response|Response] supports an additional method
to set the payloadBody as an InputStream

Result:
BlockingStreamingHttp[Response|Response] APIs are more InputStream
friendly.


idelpivnitskiy (approved):
❤️

PR #1179, opened 2020-10-13
Add tests and examples for mutual TLS

Scottmitch:
Motivation:
We currently do not have unit tests and examples dedicated to mutual TLS.

Modifications:
- Add client/server example for HTTP which demonstrates mutual TLS
- Add HTTP unit test which verifies expected behavior for mutual tls
- Update scripts to generate client/server variants of certificates and
generate the new certificates

Result:
Unit test and examples exist for mutual TLS


idelpivnitskiy (approved):
Good that you added separate certs for the client. Can you please also use new client certs here? https://github.com/apple/servicetalk/blob/0a60ec959f0c858252bd3891a1c8bada1de862af/servicetalk-tcp-netty-internal/src/test/java/io/servicetalk/tcp/netty/internal/SecureTcpTransportObserverErrorsTest.java#L115

PR #1180, opened 2020-10-14
Verify that all graceful closure use-cases also work with HTTP/2

idelpivnitskiy:
Motivation:

`GracefulConnectionClosureHandlingTest` provides a comprehensive testing
for graceful closure with HTTP/1. We need to make sure, the same cases
work correctly when users switch to HTTP/2.

Modifications:

- Add `HttpProtocol` parameter for `GracefulConnectionClosureHandlingTest`;

Result:

All use-cases of `GracefulConnectionClosureHandlingTest` are tested with
HTTP/1.x and HTTP/2.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/GracefulConnectionClosureHandlingTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/GracefulConnectionClosureHandlingTest.java
@@ -53,6 +53,7 @@ import org.junit.runners.Parameterized.Parameters;
 import java.io.IOException;
 import java.net.SocketAddress;
 import java.nio.channels.ClosedChannelException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.concurrent.ArrayBlockingQueue;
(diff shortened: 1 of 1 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/GracefulConnectionClosureHandlingTest.java
```diff
@@ -479,6 +497,10 @@ private void assertClosedChannelException(ThrowingRunnable runnable, CloseEvent
         Exception e = assertThrows(ExecutionException.class, runnable);
         Throwable cause = e.getCause();
         assertThat(cause, instanceOf(ClosedChannelException.class));
+        if (protocol == HTTP_2) {
+            // HTTP/2 does not enhance ClosedChannelException with CloseEvent
```
Scottmitch, line 501: this is an unfortunate side effect of error handling being coupled to the close handler. we should consider a followup effort to preserve the exception.

Scottmitch (approved):

PR #1184, opened 2020-10-15
grpc-protoc `typeNameSuffix` option

Scottmitch:
Motivation:
If a project uses multiple protoc plugins it is possible there are
naming collisions for the resulting generated code. The protoc plugin
interface doesn't give visibility into types generated by other plugins,
and the plugins may not control the package name (specified in the
`.proto` file, maybe used by builtins and plugins if generating into a
single file).

Modifications:
- servicetalk-grpc-protoc now supports a typeNamePostfix option that
appends a postfix to service type names to avoid collisions in the
generated code.

Result:
Users can configure grpc-protoc to generate unique type names to avoid
collisions with other gRPC protoc plugins.


idelpivnitskiy (approved):
Should not block this PR, but will be great to add some sort of verification that this option is working as expected. We may use it in one of the grpc examples we have. We don't have tests for examples, but it will fail at build time if postfix is not working.

tkountis (approved):

PR #1193, opened 2020-10-22
Option to configure capacity of the datagram packet buffer for DNS SD

idelpivnitskiy:
Motivation:

UDP packet may be truncated if it exceeds the maximum payload size of 4096
bytes. In this case, DNS ServiceDiscoverer will retry with TCP. However,
users should be able to configure higher capacity of the datagram packet
buffer.

Modifications:

- Add `DefaultDnsServiceDiscovererBuilder#maxUdpPayloadSize` method;
- Fix `IllegalArgumentException` message for `minTTL` option;

Result:

Users can control the capacity of the datagram packet buffer.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
@@ -39,7 +39,7 @@ public final class DefaultDnsServiceDiscovererBuilder {
     @Nullable
     private DnsResolverAddressTypes dnsResolverAddressTypes;
     @Nullable
-    private Integer datagramBufferCapacity;
+    private Integer maxUdpPayloadSize;
     @Nullable
     private Integer ndots;
     @Nullable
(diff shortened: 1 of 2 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -92,6 +94,20 @@ public DefaultDnsServiceDiscovererBuilder optResourceEnabled(final boolean optRe
         return this;
     }
 
+    /**
+     * Set the capacity of the datagram packet buffer (in bytes).
+     *
+     * @param datagramBufferCapacity the capacity of the datagram packet buffer
+     * @return {@code this}.
+     */
+    public DefaultDnsServiceDiscovererBuilder datagramBufferCapacity(final int datagramBufferCapacity) {
```
Scottmitch, line 103: `datagramBufferCapacity` -> `maxUdpPayloadSize`?

also consider clarifying in the comments (e.g. if DNS response exceeds this amount the UDP request will)

Scottmitch (commented):
generally lgtm, consider clarifying the name

PR #1195, opened 2020-11-02
Enhance null header value exception

Scottmitch:
Motivation:
For HTTP null header values are not allowed. An exception is thrown but
doesn't include the key which makes it more difficult for users to
understand the root cause.

Modifications:
- Enrich the null value exception to include the key

Result:
More informative exceptions.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeaders.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeaders.java
@@ -59,17 +59,6 @@ final class DefaultHttpHeaders extends MultiMap<CharSequence, CharSequence> impl
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeaders.java
```diff
@@ -58,17 +58,6 @@
         this.validateCookies = validateCookies;
     }
 
-    @Override
-    MultiMapEntry<CharSequence, CharSequence> newEntry(final CharSequence key,
-                                                       final CharSequence value, final int keyHash) {
-        return new MultiMapEntry<CharSequence, CharSequence>(value, keyHash) {
```
Scottmitch, line 64: I think at one point this class may have been used in multiple places, however this is no longer the case so I consolidated the key into the base class.

idelpivnitskiy (approved):

PR #1199, opened 2020-11-06
Introduce primitives to control internal log level

Scottmitch:
Motivation:
ServiceTalk can generate logs internally. These logs are either at a
fixed level hard coded by ServiceTalk devs, or taken dynamically at the
time of creation based upon the logging configuration. The first
approach is too ridged and may not be appropriate for the application's
logging level filtering (e.g. trace logs are never logged, or go against
logging configuration policies), or take the log level at a snapshot in
time which doesn't allow for practical filtering (e.g. log level may
start at OFF, and be switched to non-OFF value later when debugging, and
ThresholdFilters are not commonly used and introduce performance
overhead because isEnabled() pre-checks will succeed resulting in
wasted string building).

Modifications:
- Introduce a way to explicitly configure the logging level for
interesting cases (e.g. wire logging, h2 frame logging, zipkin logging
reporter)
- Introduce a way to avoid logging user data for wire logging and h2
frame logging so it can be enabled in production.

Result:
Primitives exits such that Applications can more explicitly control the
log level used internally by ServiceTalk.


Scottmitch:

test this please

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentChannelInitializer.java
```diff
@@ -54,12 +52,10 @@ public void init(final Channel channel) {
...
-        final String frameLoggerName = config.frameLoggerName();
-        if (frameLoggerName != null) {
-            LogLevel logLevel = getNettyLogLevel(frameLoggerName);
-            if (logLevel != null) {
-                multiplexCodecBuilder.frameLogger(new Http2FrameLogger(logLevel, frameLoggerName));
-            }
+        final FixedLevelLogger frameLogger = config.frameLogger();
+        if (frameLogger != null) {
+            multiplexCodecBuilder.frameLogger(
+                    new ServiceTalkHttp2FrameLogger(frameLogger, config.frameLoggerUserData()));
```
idelpivnitskiy, line 58: This code duplicates what we do in `H2ClientParentChannelInitializer`. Consider moving it to a utility method and reusing. It can be in one of these h2-initializers or in `ServiceTalkHttp2FrameLogger`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ServiceTalkHttp2FrameLogger.java
```diff
@@ -0,0 +1,185 @@
...
+import io.netty.handler.codec.http2.Http2Headers;
+import io.netty.handler.codec.http2.Http2Settings;
+import io.netty.handler.logging.LogLevel;
+
+final class ServiceTalkHttp2FrameLogger extends Http2FrameLogger {
+    private static final int BUFFER_LENGTH_THRESHOLD = 64;
+    private final FixedLevelLogger logger;
+    private final boolean logUserData;
+
+    ServiceTalkHttp2FrameLogger(final FixedLevelLogger logger, final boolean logUserData) {
```
idelpivnitskiy, line 34: Will be great to enhance `Http2FrameLogger` in netty to support "logUserData" option in future. Looks like a useful feature for all netty users.

servicetalk-logging-slf4j-internal/build.gradle
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright © 2018-2019 Apple Inc. and the ServiceTalk project authors
```
idelpivnitskiy, line 2: 2020

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BaseHttpBuilder.java
```diff
@@ -73,14 +72,25 @@
 
     /**
      * Enables wire-logging for connections created by this builder.
-     * <p>
-     * All wire events will be logged at {@link Level#TRACE TRACE} level.
```
idelpivnitskiy, line 77: Comment for all occurances of `enableWireLogging` method:

Should we mark the existing `enableWireLogging` methods as deprecated? If not and you prefer to keep both, it's better to keep this info in javadoc, add desciption that this method logs all user data, and reference another overload if users need more customization.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BaseHttpBuilder.java
```diff
@@ -73,14 +72,25 @@
...
      */
     public abstract BaseHttpBuilder<ResolvedAddress> enableWireLogging(String loggerName);
 
+    /**
+     * Enables wire-logging for connections created by this builder.
+     *
+     * @param loggerName The name of the logger to log wire events.
+     * @param logLevel The level to log at.
+     * @param logUserData {@code true} to include user data (e.g. data, headers, etc.). {@code false} to exclude this
+     * data.
```
idelpivnitskiy, line 87: "to exclude this data" is a bit confusing because you do not know what to expect in this case. Consider clarifying: `{@code false} to exclude this data and log only network events`

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WireLoggingInitializer.java
```diff
@@ -15,36 +15,30 @@
...
-     * @param loggerName The name of the logger to log wire events.
+     * @param logger The logger to use for log wire events.
+     * @param logUserData {@code true} to log user data. {@code false} to not log user data.
      */
-    public WireLoggingInitializer(final String loggerName) {
-        LogLevel logLevel = getNettyLogLevel(loggerName);
-        loggingHandler = logLevel != null ? new LoggingHandler(loggerName, logLevel) : null;
+    public WireLoggingInitializer(final FixedLevelLogger logger,
+                                  final boolean logUserData) {
+        loggingHandler = new ServiceTalkWireLogger(logger, logUserData);
```
idelpivnitskiy, line 37: Instead of introducing a custom `ServiceTalkWireLogger` can we use `LoggingHandler(name, level, logUserData ? ByteBufFormat.HEX_DUMP : ByteBufFormat.SIMPLE)` from netty?

Scottmitch: Netty's `LoggingHandler` will use `toString()` on messages, so depending upon what types are in the pipeline this may log user data (e.g. headers, full http message, etc.).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ServiceTalkHttp2FrameLogger.java
Scottmitch: agreed this is a candidate to upstream

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WireLoggingInitializer.java
idelpivnitskiy: We insert it as the first handler at tcp level where only buffers are expected. If something will change for our tcp initializer in future, we can reconsider approach.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
```diff
@@ -46,14 +45,20 @@ default String alpnId() {
...
-     * <p>
-     * All frames will be logged at {@link Level#TRACE TRACE} level.
+     * Logger for HTTP/2 frames.
      *
-     * @return the name of the logger to log HTTP/2 frames or {@code null} to disable it
+     * @return the logger to use for HTTP/2 frames or {@code null} to disable it
      */
     @Nullable
-    String frameLoggerName();
+    FixedLevelLogger frameLogger();
```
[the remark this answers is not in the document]
Scottmitch: I agree the visibility of `FixedLevelLogger` should be reduced out of API.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BaseHttpBuilder.java
```diff
@@ -73,14 +72,25 @@
...
+     * @return {@code this}.
+     */
+    public abstract BaseHttpBuilder<ResolvedAddress> enableWireLogging(String loggerName,
+                                                                       LogLevel logLevel,
+                                                                       boolean logUserData);
```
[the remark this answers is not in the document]
Scottmitch: I'm not sure this level of dynamic switching is necessary but it can be accomplished with minimal overhead by using a `BooleanSupplier`.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BaseHttpBuilder.java
Scottmitch: I'll just deprecate the methods and remove the logs for now, we can remove them in the next release.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WireLoggingInitializer.java
Scottmitch: latest round of review changes introduces `BoolenSupplier` to dynamically include/exclude data. there is some duplication but I think its simpler just to keep this for now. it is internal so we can consolidate later if we enhance upstream or find a better way.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/AbstractReadOnlyTcpConfig.java
```diff
@@ -51,8 +52,10 @@ protected AbstractReadOnlyTcpConfig(final AbstractTcpConfig<SecurityConfig, Read
         options = from.options() == null ? emptyMap() : unmodifiableMap(new HashMap<>(from.options()));
         idleTimeoutMs = from.idleTimeoutMs();
         flushStrategy = from.flushStrategy();
-        final String wireLoggerName = from.wireLoggerName();
-        wireLoggingInitializer = wireLoggerName != null ? new WireLoggingInitializer(wireLoggerName) : null;
+        final UserDataLoggerConfig wireLoggerConfig = from.wireLoggerConfig();
+        wireLoggingInitializer = wireLoggerConfig != null ?
+                new WireLoggingInitializer(wireLoggerConfig.loggerName(),
+                        wireLoggerConfig.logLevel(), wireLoggerConfig.logUserData()) : null;
```
idelpivnitskiy, line 58: No idea why it was pre-existing, but it looks like we need just a `UserDataLoggerConfig` on the `AbstractReadOnlyTcpConfig` API. We should create `WireLoggingInitializer` inside `Tcp[Client|Server]ChannelInitializer` to make it consistent with all the other initializers we create there. `AbstractReadOnlyTcpConfig` should be just a config object without knowledge of initializers.

servicetalk-logging-slf4j-internal/src/main/java/io/servicetalk/logging/slf4j/internal/Slf4jFixedLevelLoggers.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * Create a {@link FixedLevelLogger}.
+     * @param loggerName the name of the logger to use.
+     * @param level the level to log at.
+     * @return a {@link FixedLevelLogger}.
+     */
+    public static FixedLevelLogger newLogger(final String loggerName, final LogLevel level) {
+        final Logger logger = LoggerFactory.getLogger(loggerName);
```
idelpivnitskiy, line 37: `requireNonNull(loggerName)`?

idelpivnitskiy (approved):
Thanks!
A few more suggestions, then LGTM:

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BaseHttpBuilder.java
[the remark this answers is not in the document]
idelpivnitskiy: Good idea with `BooleanSupplier` 🎖️

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/AbstractTcpConfig.java
```diff
@@ -135,22 +130,21 @@ public final void flushStrategy(final FlushStrategy flushStrategy) {
      * @param loggerName The name of the logger to log wire events
      */
     public final void enableWireLogging(final String loggerName) {
-        enableWireLogging(loggerName, TRACE, true);
+        enableWireLogging(loggerName, TRACE, () -> false);
```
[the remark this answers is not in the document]
Scottmitch: I think it is worth breaking behavior here now that we have the option. by default it is safer to not log user data, and if users opt-in then they accept the risk.

servicetalk-logging-slf4j-internal/src/main/java/io/servicetalk/logging/slf4j/internal/DefaultUserDataLoggerConfig.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public LogLevel logLevel() {
+        return logLevel;
+    }
+
+    @Override
+    public BooleanSupplier logUserData() {
+        return logUserData;
+    }
```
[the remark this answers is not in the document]
Scottmitch: done

servicetalk-logging-slf4j-internal/src/main/java/io/servicetalk/logging/slf4j/internal/Slf4jFixedLevelLoggers.java
Scottmitch: I'll let `LoggerFactory.getLogger` enforce this.

Scottmitch:

test this please

PR #1203, opened 2020-11-11
Update jaxb-impl 2.3.2 -> 2.3.3

Scottmitch:


gradle.properties
```diff
@@ -28,7 +28,7 @@ javaxAnnotationsApiVersion=1.3.5
 javaxActivationVersion=1.2.1
 javaxJaxbApiVersion=2.3.2
 javaxJaxbCoreVersion=2.3.0.1
-javaxJaxbImplVersion=2.3.2
+javaxJaxbImplVersion=2.3.3
```
idelpivnitskiy, line 31: `javaxJaxbApiVersion` should also be updated to 2.3.3: https://mvnrepository.com/artifact/jakarta.xml.bind/jakarta.xml.bind-api/2.3.3

It looks like `bind-api` is pulled transitively for `jaxb-impl`, can we remove the explicit dependency on `bind-api`?
//cc @ddossot

Scottmitch: I tried removing the dependency locally and got a bunch of errors, lets leave it for now. I'll update both versions in this PR to keep them in sync.

PR #1218, opened 2020-11-18
Publisher#flatMapMerge recover from mapped Publisher exception

Scottmitch:
Motivation:
Publisher#flatMapMerge does not attempt to recover from throwing in
onNext from a mapped publisher. In this case we should attempt to
replenish demand incase there is a recoverWith or similar operator
applied to the mapped publisher, or else we may deadlock.

Modifications:
- FlatMapSubscriber should try to replenish demand even if onNext throws
an exception

Result:
Publisher#flatMapMerge is compatible with recoverWith style operators on
the mapped publishers.

### DIFF
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapMergeTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapMergeTest.java
@@ -15,6 +15,7 @@
  */
 package io.servicetalk.concurrent.api;
 
+import io.servicetalk.concurrent.PublisherSource.Processor;
 import io.servicetalk.concurrent.PublisherSource.Subscriber;
 import io.servicetalk.concurrent.internal.DeliberateException;
 import io.servicetalk.concurrent.internal.ServiceTalkTestTimeout;
@@ -30,6 +31,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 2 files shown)


Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1192 and https://github.com/apple/servicetalk/issues/1219

test this please

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapMergeTest.java
```diff
@@ -96,6 +105,43 @@ public static void afterClass() throws Exception {
...
+
+        Processor<Integer, Integer> processor = newPublisherProcessor();
+        toSource(fromSource(processor).flatMapMergeDelayError(i -> from(i + 10).recoverWith(cause ->
+                from(i + 20).concat(failed(cause))))).subscribe(mockSubscriber);
+
+        latchOnSubscribe.await();
+        processor.onNext(1);
+        assertThat(results.take(), is(11));
+        processor.onError(DELIBERATE_EXCEPTION);
+        assertThat(results.take(), is(21));
```
idelpivnitskiy, line 140: The ordering of events in this test is a bit confusing. `results` queue already has `21` before `processor.onError(DELIBERATE_EXCEPTION)` is invoked. Consider moving this check up if this is expected.
Why the concatenated `cause` is not delivered to `mockSubscriber`?

idelpivnitskiy: oh, bcz we delay error.

idelpivnitskiy: WDYT about this ordering of events?
```java
        latchOnSubscribe.await();
        processor.onNext(1);
        assertThat(results.take(), is(11));
        assertThat(results.take(), is(21));
        assertThat(causeRef.get(), is(nullValue()));
        processor.onComplete();
        latchOnError.await();
        final Throwable t = causeRef.get();
        assertThat(t, is(instanceOf(CompositeException.class)));
        assertThat(t.getCause(), is(DELIBERATE_EXCEPTION));
```

Scottmitch: I adjusted slightly but sgtm!

idelpivnitskiy (approved):
Good catch 🎖️

PR #1227, opened 2020-11-20
Make `HttpObjectDecoder` more RFC7230 compatible and other improvements

idelpivnitskiy:
Motivation:

RFC7230 defines `obs-text` as `%x80-FF` (any negative byte value), and
reason-phrase as `*( HTAB / SP / VCHAR / obs-text )`.

Modifications:

- Adjust `isObsText` and reason-phrase validation according to RFC7230;
- Verify that the passed `closeHandler` is non-null;

Result:

`HttpObjectDecoder` is more robust and RFC7230 compatible.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
@@ -110,6 +110,7 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
     };
     private static final ByteProcessor FIND_COLON = value -> value != COLON;
     private static final ByteProcessor FIND_FIELD_VALUE = value -> {
+        // Skip preceded and/or followed OWS
         if (isWS(value)) {
             return true;
         }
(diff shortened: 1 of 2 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -225,6 +226,9 @@ protected final void decode(final ChannelHandlerContext ctx, final ByteBuf buffe
                 currentState = State.READ_INITIAL;
             }
             case READ_INITIAL: {
+                if (!buffer.isReadable()) {
```
Scottmitch, line 229: IIUC this is just an optimization to avoid attempting to search for CRLF? is this thought to be common?

Scottmitch (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
idelpivnitskiy: Can not say it's common, because usually TCP packet includes enough data for the start-line. This is to prevent going inside `handlePartialInitialLine` when CRLF was not found. Let me move it inside the `if` statement.

idelpivnitskiy: Actually, I don't think we need it. Let me kill it :)

PR #1228, opened 2020-11-20
Improve exception messages from `HttpObjectDecoder`

idelpivnitskiy:
Motivation:

`HttpObjectDecoder` and its subclasses may throw an exception if the received
HTTP message violates the RFC and can not be parsed. However, exception messages
are short and hard to understand, without any pointers to what part of the
message caused an exception. Those cases are hard to debug and usually require
knowledge of RFC.

Modifications:

- Introduce `IllegalCharacterException` type that shows an illegal character as
a `char` and in hex format, as well as expected values;
- `try-catch` an `IllegalCharacterException` and wrap it with `DecoderException`
that clarifies where the illegal character was found;
- Adjust tests for new exception messages;

Result:

More understandable exception messages from `HttpObjectDecoder`.

Depends on #1227.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
@@ -41,6 +41,7 @@ import io.servicetalk.http.api.HttpResponseMetaData;
 import io.servicetalk.transport.netty.internal.ByteToMessageDecoder;
 import io.servicetalk.transport.netty.internal.CloseHandler;
 import io.servicetalk.transport.netty.internal.CloseHandler.DiscardFurtherInboundEvent;
+import io.servicetalk.utils.internal.IllegalCharacterException;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
@@ -82,7 +83,6 @@ import static java.lang.Character.isISOControl;
 import static java.lang.Character.isWhitespace;
 import static java.lang.Long.parseUnsignedLong;
 import static java.lang.Math.min;
-import static java.lang.String.format;
 import static java.nio.charset.StandardCharsets.US_ASCII;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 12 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpResponseDecoder.java
```diff
@@ -41,17 +43,19 @@
...
-    private static final ByteProcessor ENSURE_NO_CONTROL_CHARS = value -> {
-        if (value != HT && isISOControl(value)) {   // allow HTAB
-            throw newIllegalCharacter(value);
+    private static final byte[] FIRST_BYTES = "HTTP/".getBytes(US_ASCII);
+    private static final byte DEL = 127;
+
+    private static final ByteProcessor ENSURE_REASON_PHRASE = value -> {
+        // Any control character (0x00-0x1F) except HT
+        if (((value & 0xE0) == 0 && value != HT) || value == DEL) {
+            throw new IllegalCharacterException(value, "HTAB / SP / VCHAR / obs-text");
```
Scottmitch, line 58: since we will wrap this immediately with a `DecoderException` can we use `ThrowableUtils.unknownStackTrace()` here?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpResponseDecoder.java
```diff
@@ -140,12 +150,18 @@ protected boolean isContentAlwaysEmpty(final HttpResponseMetaData msg) {
...
         final int medium = buffer.getUnsignedMedium(start);
-        return toDecimal((medium & 0xff0000) >> 16) * 100 +
-                toDecimal((medium & 0xff00) >> 8) * 10 +
-                toDecimal(medium & 0xff);
+        try {
+            return toDecimal((medium & 0xff0000) >> 16) * 100 +
+                    toDecimal((medium & 0xff00) >> 8) * 10 +
+                    toDecimal(medium & 0xff);
+        } catch (IllegalCharacterException cause) {
+            throw new DecoderException("Invalid start-line: HTTP status-code must contain only 3 digits, found: " +
```
Scottmitch, line 163: we should consider moving the exceptional cases that have to create exceptions and build strings separate methods to allow the JVM to inline more easily for less frequent code paths.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -96,7 +96,8 @@
         if (value == CR || value == LF) {
             return true;
         }
-        throw newIllegalCharacter(value);
+        throw new DecoderException("Invalid preface character before the start-line of the HTTP message",
+                new IllegalCharacterException(value, "CR (0x0D), LF (0x0A)"));
```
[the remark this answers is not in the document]
idelpivnitskiy: I've created `StacklessDecoderException` for cases when its stack trace is not relevant.

tkountis (approved):
LGTM

PR #1229, opened 2020-11-21
Consolidate Test*Subscriber classes

Scottmitch:
Motivation:
There are 3 variants of test subscriber classes that have organically
been added over time:
- LegacyMocked*ListenerRule
- concurrent.api.Test*Subscriber
- concurrent.api.TestCollecting*Subscriber

There is no technical justification for the duplication, and some older
test utilities lend themselves toward violating the reactive streams
specification.

Modifications:
- Remove all the of test utilities mentioned above
- Create a servicetalk-concurrent-test-internal module with
TestSubscribers build from concurrent.api.TestCollecting*Subscriber

Result:
There is a single set of low-level, currently internal, Test*Subscriber
classes.

PR #1231, opened 2020-11-24
Publisher#flatMapConcatIterable error recovery from Subscriber#onNext…

Scottmitch:
… throwing

Motivation:
If Subscriber#onNext throws in Publisher#flatMapConcatIterable we don't
reset the currentIterator state if there is no pending requestN demand.
This may result in not propagating the error to the downstream
subscriber.

Modifications:
- PublisherConcatMapIterable should reset the currentIterator state when
catching an exception from Subscriber#onNext
- PublisherConcatMapIterable should also attempt to close the Iterator
if an exception is caught from Subscriber methods

Result:
Publisher#flatMapConcatIterable has more robust exception handling and
error propagation.

### DIFF
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherConcatMapIterableTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherConcatMapIterableTest.java
@@ -21,6 +21,7 @@ import io.servicetalk.concurrent.PublisherSource.Subscriber;
 import io.servicetalk.concurrent.PublisherSource.Subscription;
 import io.servicetalk.concurrent.internal.DeliberateException;
 
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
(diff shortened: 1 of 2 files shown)


servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherConcatMapIterableTest.java
```diff
@@ -315,6 +316,7 @@ public void exceptionFromBufferedOnNextThenOnErrorIsPropagated() {
         testExceptionFromBufferedOnNextThenTerminalIsPropagated(publisher::onError);
     }
 
+    @Ignore("fixed in https://github.com/apple/servicetalk/pull/1229")
```
Scottmitch, line 319: the updated test are in https://github.com/apple/servicetalk/pull/1229 and depend upon a few core changes (e.g. `TestPublisher`, `TestPublisherSubscriber`) so I will defer the testing to that PR with the larger changes.

idelpivnitskiy (approved):
Good catch!

tkountis (approved):

Scottmitch:
jdk11 test failure due to https://github.com/apple/servicetalk/issues/1192

PR #1235, opened 2020-12-02
Add interruptable methods to concurrent-test-internal

Scottmitch:
Motivation:
Some test utilities would like to allow interrupts to be propagated, but
for simplicity the concurrent-test-internal default methods don't
propagate the interrupts.

Modifications:
- Add interrupt variants of methods which may block to
concurrent-test-internal utilities
- TestPublisherSubscriber#pollTerminal* methods should return a Supplier
so if there is a terminal it is returned as a result of this method
- TestPublisherSubscriber#pollOnNext* methods should return a Supplier
to avoid ambigiutity with a null Optional and consistency with the
TestPublisherSubscriber#pollTerminal* methods.
- TestPublisherSubscriber should insert terminal events into the event
queue. This will make sure if a test is waiting for an onNext and a
terminal arrives the test will be failed fast with a helpful error
message.

Result:
concurrent-test-internal support interruptable methods.


Scottmitch:
The PR that dependent upon this PR (https://github.com/apple/servicetalk/pull/1236) is closed. I will also close this PR and extract the useful pieces.

Scottmitch:
the relevant parts have been moved to https://github.com/apple/servicetalk/pull/1241

PR #1236, opened 2020-12-02
Add concurrent-api-test module

Scottmitch:
Motivation:
There is currently a servicetalk-concurrent-test-internal module for low
level testing of concurrent utilities. However there is no user facing
testing utility for dealing with concurrent-api primitives.

Modifications:
- Introduce the concurrent-api-test module which has the StepVerifiers
entry point. There are 3 entry points one for each async source type:
  - PublisherFirstStep
  - SingleFirstStep
  - CompletableFirstStep

Each of these APIs exposes a sequence of further APIs which represents
the expected signal flow of each respective source (e.g. onSubscribe,
onNext*, [onComplete | onError]?). This approach is inspired by
[reactor-test](https://projectreactor.io/docs/core/release/reference/index.html#testing).

Result:
User facing test utilities exist for concurrent-api.


Scottmitch:
depends upon PR https://github.com/apple/servicetalk/pull/1235

Scottmitch:
closed in favor of https://github.com/apple/servicetalk/pull/1240. This PR leverages the existing `concurrent-test-internal` modules, but this means verification is done outside the `Subscriber` thread. This would not be able to run verifications which must be in the subscriber thread (e.g. `AsyncContext`, thread locals, re-entrant subscription requesting/cancellation, throwing from the subscriber, etc.).

PR #1241, opened 2020-12-05
TestPublisherSubscriber API updates

Scottmitch:
Motivation:
TestPublisherSubscriber supports some consumers which can be tested via
other means. TestPublisherSubscriber#pollTerminal doesn't inform the
caller as to which terminal event has arrived, and there are only
blocking alternatives to get the terminal event, so if you call the
wrong method you may block indefinitely.

Modifications:
- Remove onSubscribe and onNext Consumers
- Change the return type of TestPublisherSubscriber#pollTerminal so the
caller knows which type of terminal was processed
- Change the pollOnNext to avoid a null Optional value which is
generally discouraged, and make the API more inline with pollTerminal

Result:
TestPublisherSubscriber has smaller API surface and poll methods provide
results in a more consistent fashion.

### DIFF
--- a/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/TestPublisherSubscriber.java
+++ b/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/TestPublisherSubscriber.java
@@ -217,8 +217,8 @@ public final class TestPublisherSubscriber<T> implements Subscriber<T> {
      *
      * @param timeout The amount of time to wait.
      * @param unit The units of {@code timeout}.
-     * @return item delivered to {@link #onNext(Object)}, or {@code null} if a timeout occurred or a terminal signal
-     * has been received.
+     * @return A {@link Supplier} that returns the signal delivered to {@link #onNext(Object)},
+     * or {@code null} if a timeout occurred or a terminal signal has been received.
      */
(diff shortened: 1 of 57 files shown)


servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/TestPublisherSubscriber.java
```diff
@@ -246,12 +217,13 @@ public T takeOnNext() {
      *
      * @param timeout The amount of time to wait.
      * @param unit The units of {@code timeout}.
-     * @return item delivered to {@link #onNext(Object)}, or {@code null} if a timeout occurred.
+     * @return item delivered to {@link #onNext(Object)}, or {@code null} if a timeout occurred or a terminal signal
```
idelpivnitskiy, line 220: It returns `Supplier`, not an item

idelpivnitskiy (approved):

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/TestPublisherSubscriber.java
Scottmitch: fixed

PR #1242, opened 2020-12-07
Jersey to return 400 response for invalid URI

Scottmitch:
Motivation:
DefaultJerseyStreamingHttpRouter must create two URI objects in order to
initialize Jersey's ContainerRequest object. URI creation will also do
validation, and may throw. In this case we should return a 400 response
instead of a 500 internal server error.

Modifications:
- DefaultJerseyStreamingHttpRouter to try/catch creation of URIs and
return a 400 if an exception is thrown.

Result:
More correct error returned in the event of invalid URI when using
Jersey.

### DIFF
--- a/servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/DefaultJerseyStreamingHttpRouter.java
+++ b/servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/DefaultJerseyStreamingHttpRouter.java
@@ -212,7 +212,7 @@ final class DefaultJerseyStreamingHttpRouter implements StreamingHttpService {
         try {
             baseURI = URI.create(baseUri.toString());
             requestURI = URI.create(requestUriBuilder.toString());
-        } catch (Throwable cause) {
+        } catch (IllegalArgumentException cause) {
             Buffer message = serviceCtx.executionContext().bufferAllocator().fromAscii(cause.getMessage());
             StreamingHttpResponse response = factory.badRequest().payloadBody(from(message));
             response.headers().add(CONTENT_LENGTH, Integer.toString(message.readableBytes()));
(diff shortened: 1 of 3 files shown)


servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/DefaultJerseyStreamingHttpRouter.java
```diff
@@ -202,9 +207,23 @@ private void handle0(final HttpServiceContext serviceCtx, final StreamingHttpReq
             requestUriBuilder.append(requestTarget);
         }
 
+        final URI baseURI;
+        final URI requestURI;
+        try {
+            baseURI = URI.create(baseUri.toString());
+            requestURI = URI.create(requestUriBuilder.toString());
+        } catch (Throwable cause) {
+            Buffer message = serviceCtx.executionContext().bufferAllocator().fromAscii(cause.getMessage());
```
idelpivnitskiy, line 216: Exception msg may leak some server details, which is hard to control if this `try` block changes in future. Consider logging the exception and generating a response without payload body. Alternatively, consider catching only specific exception types relevant to `URI`.

idelpivnitskiy (approved):

Scottmitch:
ci failure...
```
13:38:56 
Creating servicetalk-quality-prb_runtime-setup_1 ... error

13:38:56 ERROR: for servicetalk-quality-prb_runtime-setup_1  Cannot start service runtime-setup: network servicetalk-quality-prb_default not found
```

test this please

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/DefaultJerseyStreamingHttpRouter.java
Scottmitch: I'll refine/narrow the exceptions type

PR #1243, opened 2020-12-07
ConcurrentSubscriptionTest#multiThreadCancelNotDeliveredIfRequestThro…

Scottmitch:
…ws buggy test

Motivation:
ConcurrentSubscriptionTest#multiThreadCancelNotDeliveredIfRequestThrows
execute cancel() and request(n) concurrently. The assertion condition
should permit for either executing first, but currently there is a bug
that if cancel happens the test may fail.

Modifications:
- The assertion should check either for a failure, or that cancel
happened

Result:
More correct ConcurrentSubscriptionTest, less false positive test
failures.


idelpivnitskiy (approved):
💯

PR #1256, opened 2020-12-08
Add Publisher#completableOrError operator

Scottmitch:
Motivation:
We have Publisher#ignoreElements which ignores any onNext signals
emitted from the Publisher. However there are use cases where the
conversion should be more strict and fail if there are any onNext
signals.

Modifications:
- Add Publisher#completableOrError method to do a strict conversion.
Share code with Publisher#ignoreElements operator.

Result:
Operator exists which does a strict conversion from Publisher to
Completable.


idelpivnitskiy (approved):
🚀

PR #1258, opened 2020-12-08
Github Actions Pull Request Builder

Scottmitch:


Scottmitch:
test failure observed https://github.com/apple/servicetalk/issues/1259

.github/workflows/prb.yml
```diff
@@ -0,0 +1,86 @@
...
+      run: ./gradlew --parallel --max-workers=4 clean test
+    - name: Publish Test Results
+      if: ${{ always() }}
+      uses: scacap/action-surefire-report@v1.0.5
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        report_paths: '**/build/test-results/test/TEST-*.xml'
+        check_name: JDK11 Test Report
+
+  quality-jdk11:
```
idelpivnitskiy, line 67: Because we release using JDK8 we should run `quality` task on JDK8 as well. Otherwise, there is a risk to fail the release. We observed behavior different for spotbugs on different JDK versions.

.github/workflows/prb.yml
```diff
@@ -0,0 +1,86 @@
...
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build-jdk8:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
idelpivnitskiy, line 14: You use `name` for all other steps, consider using a `name` here as well

.github/workflows/prb.yml
```diff
@@ -0,0 +1,86 @@
...
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        report_paths: '**/build/test-results/test/TEST-*.xml'
+        check_name: JDK8 Test Report
+
+  build-jdk11:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK11
```
idelpivnitskiy, line 43: Will be nice to try matrix testing: https://github.com/actions/setup-java#matrix-testing
If that works in parallel, we can reduce duplication in config file. Will be also great to include the latest JDK15.

.github/workflows/prb.yml
```diff
@@ -0,0 +1,86 @@
...
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build-jdk8:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK8
+      uses: actions/setup-java@v1.4.3
```
idelpivnitskiy, line 16: Consider using `v1` to always get the latest version

.github/workflows/prb.yml
```diff
@@ -0,0 +1,86 @@
...
+      with:
+        java-version: 1.8
```
idelpivnitskiy, line 18: Sad that it doesn't print the installed version build number. I believe, 1.8 has the latest version of 1.8.212. Then they switched to 8.0.x pattern. Consider using the major version format (8 and 11) as described here: https://github.com/actions/setup-java#basic

.github/workflows/prb.yml
```diff
@@ -0,0 +1,86 @@
...
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+        restore-keys: ${{ runner.os }}-gradle
+    - name: Build with Gradle
+      run: ./gradlew --parallel --max-workers=4 clean test
+    - name: Publish Test Results
+      if: ${{ always() }}
+      uses: scacap/action-surefire-report@v1.0.5
```
idelpivnitskiy, line 33: Consider using `v1` to always get the latest minor version

idelpivnitskiy (commented):
Looks exiting 🎉

.github/workflows/prb.yml
idelpivnitskiy: We can also add a step that executes `java -version`

.github/workflows/prb.yml
Scottmitch: sounds good, let me try this.

.github/workflows/prb.yml
Scottmitch: adding a print java version step is simple enough

idelpivnitskiy (approved):
Looks great!

Scottmitch:
test failure caused by https://github.com/apple/servicetalk/issues/1192, unrelated to this PR

PR #1260, opened 2020-12-09
Relax can-add-content-length check

tkountis:
Motivation:

The additional content-type check in`canAddContentLength`, was added due to a bug, addressed [here](**url**)  

Modifications:

Removed the additional check

Result:

Behavior restored as pre http compression feature.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
```diff
@@ -113,9 +113,7 @@ static boolean canAddResponseTransferEncodingProtocol(final int statusCode,
     }
 
     private static boolean canAddContentLength(final HttpMetaData metadata) {
-        // TODO once this bug is addressed (https://github.com/apple/servicetalk/pull/1213)
-        // we should relax the check here, remove content-encoding clause
-        return !hasContentHeaders(metadata.headers()) && !hasContentEncoding(metadata.headers()) &&
```
idelpivnitskiy, line 118: Consider removing private `hasContentEncoding` method now

idelpivnitskiy (approved):

PR #1268, opened 2020-12-10
Github Actions PRB clarify report task names

Scottmitch:
Motivation:
Task names for generated reports don't indicate they are just the
published reports, and don't represent the actual analysis execution.

Modifications:
- Include "report" in the name for these tasks to clarify

Result:
Task names for reports are more clear that they just represent the
report/visualization.


Scottmitch:
`pull_request_target` means merge is required for change to take effect.

PR #1269, opened 2020-12-10
Github Actions CI Release Builds

Scottmitch:


.github/workflows/ci-release.yml
```diff
@@ -0,0 +1,43 @@
...
+        env:
+          ORG_GRADLE_PROJECT_signingKey: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGKEY }}
+          ORG_GRADLE_PROJECT_signingPassword: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGPASSWORD }}
+          SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
+          SONATYPE_TOKEN: ${{ secrets.SONATYPE_TOKEN }}
+          CI: true
+        run: ./gradlew --parallel --max-workers=4 -PreleaseBuild=true clean check && ./gradlew -PreleaseBuild=true publish
+      - name: Publish Test Results
+        if: ${{ always() }}
+        uses: scacap/action-surefire-report@v1
```
jerrinot, line 39: for 3rd party actions is better to use a commit hash instead of a tag. 
tags are not immutable and a bad actor could change Action code, harvest secrets, etc..

Scottmitch:
failure caused by https://github.com/apple/servicetalk/issues/1224, unrelated to this PR.

.github/workflows/ci-release.yml
Scottmitch: thank @jerrinot! I will submit a followup PR to address this and make it consistent for other actions as well.

Scottmitch: https://github.com/apple/servicetalk/pull/1271

PR #1274, opened 2020-12-11
release.sh to also publish docs

Scottmitch:
Motivation:
The release process has changed and now involved Github actions.
Previously the CI ran the release.sh script an a user manually ran the
publish-docs.sh script. This split came from a time when publishing docs
was more complicated, but it no longer needs to be manually orchistrated
in two different steps.

Modifications:
- release.sh should invoke publish-docs.sh
- publish-docs.sh should force docs to be published with JDK11 to get
newer javadocs (including search).

Results:
Single script to prepare release and generate artifacts.

### DIFF
--- a/scripts/publish-docs.sh
+++ b/scripts/publish-docs.sh
@@ -21,17 +21,24 @@ cd "$(dirname "$0")"
 cd ..
 
 version=""
+cleanRemoteSite=true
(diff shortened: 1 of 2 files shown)


idelpivnitskiy (approved):
Good improvements!

PR #1281, opened 2020-12-16
DynamicCompositeCancellable race condition fix

Scottmitch:
Motivation:
DynamicCompositeCancellable are used in operators that keep a collection
of Cancellable objects (e.g. flatMap, merge, etc.). The implementations
are intended to be thread safe between adding Cancellables and
cancelling Cancellables that have been previously add (and not removed).
The DynamicCompositeCancellable implementations rely upon
reading volatile state after an element is added to a concurrent
collection to cancel the new element if there is a race. However the add
thread is not gaurenteed to observe a write from the cancel thread, and
if this is the case an element maybe added and never cancelled.

Modifications:
- The add thread should attempt to read/modify state in coordination
with the cancel thread to gaurentee that element will be cancelled.

Result:
More correct concurrency code in DynamicCompositeCancellable.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DynamicCompositeCancellable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DynamicCompositeCancellable.java
@@ -34,8 +34,9 @@ interface DynamicCompositeCancellable extends Cancellable {
      * Add a {@link Cancellable} that will be cancelled when this object's {@link #cancel()} method is called,
      * or be cancelled immediately if this object's {@link #cancel()} method has already been called.
      * @param toAdd The {@link Cancellable} to add.
-     * @return {@code true} if the {@code toAdd} was added, and {@code false} if {@code toAdd} was not added and
-     * {@link Cancellable#cancel()} was called on {@code toAdd}.
+     * @return {@code true} if the {@code toAdd} was added, and {@code false} if {@code toAdd} was not added because
+     * it already exists. If {@code false} then {@link Cancellable#cancel()} will be called unless the reason is this
+     * implementation does not supporting duplicates and {@code toAdd} has already been added.
      */
     boolean add(Cancellable toAdd);
 
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/QueueDynamicCompositeCancellable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/QueueDynamicCompositeCancellable.java
@@ -20,17 +20,18 @@ import io.servicetalk.concurrent.internal.FlowControlUtils;
 
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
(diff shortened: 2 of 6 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/QueueDynamicCompositeCancellable.java
```diff
@@ -16,63 +16,67 @@
...
 
 final class QueueDynamicCompositeCancellable implements DynamicCompositeCancellable {
-    private static final AtomicIntegerFieldUpdater<QueueDynamicCompositeCancellable> cancelledUpdater =
-            AtomicIntegerFieldUpdater.newUpdater(QueueDynamicCompositeCancellable.class, "cancelled");
+    private static final AtomicLongFieldUpdater<QueueDynamicCompositeCancellable> sizeUpdater =
+            newUpdater(QueueDynamicCompositeCancellable.class, "size");
     @SuppressWarnings("unused")
-    private volatile int cancelled;
-
+    private volatile long size;
```
idelpivnitskiy, line 34: Queue size is `int` users won't be able to add more elements, why do we need `long` here?

idelpivnitskiy (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/QueueDynamicCompositeCancellable.java
Scottmitch: `size()` is int, however the implementations which are not `array` backed could conceivably hold more than `Integer.MAX_VALUE` elements. These implementations just cap the value of `size()` to `Integer.MAX_VALUE`. I refactored so instead of tracking size I am now just using an atomic operation on the collection.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DynamicCompositeCancellable.java
```diff
@@ -34,8 +34,9 @@
      * Add a {@link Cancellable} that will be cancelled when this object's {@link #cancel()} method is called,
      * or be cancelled immediately if this object's {@link #cancel()} method has already been called.
      * @param toAdd The {@link Cancellable} to add.
-     * @return {@code true} if the {@code toAdd} was added, and {@code false} if {@code toAdd} was not added and
-     * {@link Cancellable#cancel()} was called on {@code toAdd}.
+     * @return {@code true} if the {@code toAdd} was added, and {@code false} if {@code toAdd} was not added because
+     * it already exists. If {@code false} then {@link Cancellable#cancel()} will be called unless the reason is this
```
Scottmitch, line 38: note this interface is internal, but the semantics around it didn't really make sense for `set` so I clarified here and added a unit test to verify.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/QueueDynamicCompositeCancellable.java
idelpivnitskiy: 🤯 looked at `ConcurrentLinkedQueue` impl, TIL

Issue #1277, opened 2020-12-11
ConnectionCloseHeaderHandlingTest$PipelinedRequestsTest.serverCloseSecondPipelinedRequestWriteAborted test failure


Scottmitch:
maybe related https://github.com/apple/servicetalk/issues/1154 ?

https://github.com/apple/servicetalk/runs/1540120456
```
[3: useUds=false, viaProxy=true, awaitRequestPayload=true]

java.util.concurrent.ExecutionException: io.servicetalk.transport.netty.internal.CloseHandler$CloseEventObservedException: PROTOCOL_CLOSING_INBOUND(The application protocol closed the read side of this connection. This maybe the result of sending an HTTP header such as Connection: close.) [id: 0x10e1f286, L:/127.0.0.1:45892 ! R:localhost/127.0.0.1:43377]
	at io.servicetalk.concurrent.api.SourceToFuture.reportGet(SourceToFuture.java:121)
	at io.servicetalk.concurrent.api.SourceToFuture.get(SourceToFuture.java:92)
	at io.servicetalk.http.netty.ConnectionCloseHeaderHandlingTest$ConnectionSetup.assertResponsePayloadBody(ConnectionCloseHeaderHandlingTest.java:226)
	at io.servicetalk.http.netty.ConnectionCloseHeaderHandlingTest$PipelinedRequestsTest.serverCloseSecondPipelinedRequestWriteAborted(ConnectionCloseHeaderHandlingTest.java:386)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
[...]

PR #1283, opened 2020-12-17
CompletableProcessor and SingleProcessor race condition fix

Scottmitch:
Motivation:
CompletableProcessor and SingleProcessor maintain a queue and an
independent state for termination. It is possible that a Subscriber
maybe added on different thread than is terminating the Processor, and
the terminal state may not be visible yet such that the Subscriber never
gets terminated.

Modifications:
- The thread adding the the queue should attempt to read/modify state in
coordination with the thread terminating to gaurentee the new Subscriber
will be terminated.
- Allow for out of order termination in the event that the processor
completes before the subscribe operation happens.

Result:
Fixes https://github.com/apple/servicetalk/issues/1033

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
@@ -18,27 +18,18 @@ package io.servicetalk.concurrent.api;
 import io.servicetalk.concurrent.Cancellable;
 import io.servicetalk.concurrent.CompletableSource.Processor;
 import io.servicetalk.concurrent.internal.DelayedCancellable;
-import io.servicetalk.concurrent.internal.QueueFullAndRejectedSubscribeException;
 import io.servicetalk.concurrent.internal.TerminalNotification;
 
-import java.util.Queue;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.internal.TerminalNotification.complete;
-import static io.servicetalk.concurrent.internal.ThrowableUtils.catchUnexpected;
-import static io.servicetalk.utils.internal.PlatformDependent.throwException;
 
 /**
  * A {@link Completable} which is also a {@link Subscriber}. State of this {@link Completable} can be modified by using
  * the {@link Subscriber} methods which is forwarded to all existing or subsequent {@link Subscriber}s.
  */
 final class CompletableProcessor extends Completable implements Processor {
-    private static final AtomicLongFieldUpdater<CompletableProcessor> counterUpdater =
-            AtomicLongFieldUpdater.newUpdater(CompletableProcessor.class, "counter");
-    private final Queue<Subscriber> queue = new ConcurrentLinkedQueue<>();
-    private volatile long counter;
+    private final ConcurrentStack<Subscriber> stack = new ConcurrentStack<>();
     @Nullable
     private TerminalNotification terminalSignal;
 
@@ -50,24 +41,14 @@ final class CompletableProcessor extends Completable implements Processor {
         // we would add the subscriber to the queue and possibly never (until termination) dereference the subscriber.
         DelayedCancellable delayedCancellable = new DelayedCancellable();
         subscriber.onSubscribe(delayedCancellable);
-        if (counter < 0) { // best effort to short circuit and avoid the queue if terminated already
-            terminateLateSubscriber(subscriber);
-        } else if (queue.offer(subscriber)) {
-            // We must check the state after we insert into the queue as the terminal event could have raced with this
-            // method and we need to ensure the subscriber is terminated.
-            // We don't check for overflow as we assume we will never exceed Long number of elements.
-            if (counterUpdater.incrementAndGet(this) > 0) {
-                delayedCancellable.delayedCancellable(() -> {
-                    // Cancel in this case will just cleanup references from the queue to ensure we don't prevent GC of
-                    // these references.
-                    queue.remove(subscriber);
-                });
-            } else if (queue.remove(subscriber)) {
-                // We don't decrement counter, we assume there will be no overflow from Long.MIN_VALUE.
-                terminateLateSubscriber(subscriber);
-            }
+        if (stack.push(subscriber)) {
+            delayedCancellable.delayedCancellable(() -> {
+                // Cancel in this case will just cleanup references from the queue to ensure we don't prevent GC of
+                // these references.
+                stack.relaxedRemove(subscriber);
+            });
         } else {
-            subscriber.onError(new QueueFullAndRejectedSubscribeException("queue"));
+            terminateLateSubscriber(subscriber);
         }
     }
 
@@ -94,23 +75,10 @@ final class CompletableProcessor extends Completable implements Processor {
 
     private void terminate(TerminalNotification terminalSignal) {
         if (this.terminalSignal == null) {
-            // We must set terminalSignal before counter as we depend upon happens-before relationship for this value
+            // We must set terminalSignal before close as we depend upon happens-before relationship for this value
             // to be visible for any future late subscribers.
             this.terminalSignal = terminalSignal;
-            if (counterUpdater.getAndSet(this, Long.MIN_VALUE) >= 0) {
-                Throwable delayedCause = null;
-                Subscriber subscriber;
-                while ((subscriber = queue.poll()) != null) {
-                    try {
-                        terminalSignal.terminate(subscriber);
-                    } catch (Throwable cause) {
-                        delayedCause = catchUnexpected(delayedCause, cause);
-                    }
-                }
-                if (delayedCause != null) {
-                    throwException(delayedCause);
-                }
-            }
+            stack.close(terminalSignal::terminate);
         }
     }
 
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
@@ -23,24 +23,22 @@ import io.servicetalk.concurrent.internal.TerminalNotification;
 
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.internal.TerminalNotification.complete;
 import static io.servicetalk.concurrent.internal.ThrowableUtils.catchUnexpected;
 import static io.servicetalk.utils.internal.PlatformDependent.throwException;
-import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
 
 /**
  * A {@link Completable} which is also a {@link Subscriber}. State of this {@link Completable} can be modified by using
  * the {@link Subscriber} methods which is forwarded to all existing or subsequent {@link Subscriber}s.
  */
 final class CompletableProcessor extends Completable implements Processor {
-    @SuppressWarnings("rawtypes")
-    private static final AtomicReferenceFieldUpdater<CompletableProcessor, Queue> queueUpdater =
-            newUpdater(CompletableProcessor.class, Queue.class, "queue");
-    @Nullable
-    private volatile Queue<Subscriber> queue = new ConcurrentLinkedQueue<>();
+    private static final AtomicLongFieldUpdater<CompletableProcessor> counterUpdater =
+            AtomicLongFieldUpdater.newUpdater(CompletableProcessor.class, "counter");
+    private final Queue<Subscriber> queue = new ConcurrentLinkedQueue<>();
+    private volatile long counter;
     @Nullable
     private TerminalNotification terminalSignal;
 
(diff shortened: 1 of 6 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
```diff
@@ -84,25 +87,32 @@ public void onError(Throwable t) {
...
-    private void notifyListeners(TerminalNotification terminalSignal) {
-        Throwable delayedCause = null;
-        Subscriber subscriber;
-        while ((subscriber = subscribers.poll()) != null) {
-            try {
-                terminalSignal.terminate(subscriber);
-            } catch (Throwable cause) {
-                delayedCause = catchUnexpected(delayedCause, cause);
+    private void terminate(TerminalNotification terminalSignal) {
+        if (this.terminalSignal == null) {
```
Scottmitch, line 97: Note that `this.terminalSignal` is no longer atomic and the visibility is provided by the `queueUpdater` atomic operations. The `Subscriber` API doesn't allow for concurrent access so this shouldn't violate the API but the value maybe overridden if no external synchronization is provided.

Scottmitch:
the extra CAS has an impact on single threaded performance. I tried a different approach which uses cheaper atomic operations (GAS, IAG) and observed a smaller degradation for the complete-after-subscribe cases and gains for the subscribe-after-complete cases.

```
# queue reference compareAndSet
Benchmark                                       Mode  Cnt        Score         Error  Units
CompletableProcessorBenchmark.mtAdd20Complete  thrpt    5   827397.729 ±  145198.667  ops/s
CompletableProcessorBenchmark.mtAdd3Cancel     thrpt    5   112720.286 ±   14772.002  ops/s
CompletableProcessorBenchmark.mtAdd3Complete   thrpt    5   149970.493 ±   22144.766  ops/s
CompletableProcessorBenchmark.stAdd20Complete  thrpt    5  1075128.397 ±  228974.142  ops/s
CompletableProcessorBenchmark.stAdd3Cancel     thrpt    5  4975179.331 ± 1324293.843  ops/s
CompletableProcessorBenchmark.stAdd3Complete   thrpt    5  5920659.540 ±  382412.937  ops/s

# long incrementAndGet (no decrement)
Benchmark                                       Mode  Cnt        Score        Error  Units
CompletableProcessorBenchmark.mtAdd20Complete  thrpt    5   850356.235 ± 100126.357  ops/s
CompletableProcessorBenchmark.mtAdd3Cancel     thrpt    5   114501.563 ±  10265.678  ops/s
CompletableProcessorBenchmark.mtAdd3Complete   thrpt    5   124274.604 ±   4143.882  ops/s
CompletableProcessorBenchmark.stAdd20Complete  thrpt    5  1166903.549 ±  26511.557  ops/s
CompletableProcessorBenchmark.stAdd3Cancel     thrpt    5  5414642.294 ±  93128.348  ops/s
CompletableProcessorBenchmark.stAdd3Complete   thrpt    5  6340309.749 ± 388974.856  ops/s

# main
Benchmark                                       Mode  Cnt        Score        Error  Units
CompletableProcessorBenchmark.mtAdd20Complete  thrpt    5   568637.321 ±  20293.199  ops/s
CompletableProcessorBenchmark.mtAdd3Cancel     thrpt    5   114377.073 ±  11147.559  ops/s
CompletableProcessorBenchmark.mtAdd3Complete   thrpt    5   145511.365 ±  15154.028  ops/s
CompletableProcessorBenchmark.stAdd20Complete  thrpt    5  1304582.520 ±  19717.008  ops/s
CompletableProcessorBenchmark.stAdd3Cancel     thrpt    5  5428589.825 ±  93058.215  ops/s
CompletableProcessorBenchmark.stAdd3Complete   thrpt    5  6983714.635 ± 119887.406  ops/s
```

Scottmitch:

Scottmitch:
I made similar changes for `DynamicCompositeCancellable` in https://github.com/apple/servicetalk/pull/1284

Scottmitch:
discussed offline with @idelpivnitskiy. to summarize: I experimented with a few different approaches, and the approach that had the best results was the `ConcurrentStack`. We sacrifice strict ordering (which isn't a requirement), and may accumulate nodes because the current stack implementation doesn't take nodes out of the collection on removal. However for our needs this is a good trade-off as the number of subscribers is expected to be small and cancellation (which triggers removal) is expected to be rare. If folks run into increased memory footprint due to high removal and long lived Processor we can provide alternative implementations.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ConcurrentStack.java
```diff
@@ -0,0 +1,128 @@
...
+     *     <li>invoked multiple times with the same {@code item} from different threads</li>
+     *     <li>{@link #pop()} removes this item from another thread</li>
+     * </ul>
+     */
+    boolean relaxedRemove(T item) {
+        requireNonNull(item);
+        Node<T> currTop = top;
+        while (currTop != null) {
+            if (item.equals(currTop.item)) {
+                currTop.item = null; // best effort null out the item. pop/close will discard the Node later.
```
tkountis, line 78: This is not safely published to other threads. (eg. volatile). Intended due to method contract?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleProcessor.java
```diff
@@ -88,38 +67,34 @@ public void onError(final Throwable t) {
...
-            }
+            subscriber.onSuccess(value);
         }
-        if (delayedCause != null) {
-            throwException(delayedCause);
+    }
+
+    private void terminate(@Nullable Object terminalSignal) {
+        if (this.terminalSignal == TERMINAL_UNSET) {
+            // We must set terminalSignal before counter as we depend upon happens-before relationship for this value
```
tkountis, line 86: counter?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ConcurrentStack.java
```diff
@@ -0,0 +1,128 @@
...
+            newTop.next = oldTop;
+        } while (!topUpdater.compareAndSet(this, oldTop, newTop));
+        return true;
+    }
+
+    @Nullable
+    T pop() {
+        for (;;) {
+            final Node<T> oldTop = top;
+            if (oldTop == null) {
```
tkountis, line 52: consider fast-exit when CLOSED

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ConcurrentStack.java
Scottmitch: correct we don't need strong guarantees of removal, mostly just for GC. if we needed stronger guarantees we would have to make volatile and use atomics here and in `pop`, but the relaxed semantics are OK.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleProcessor.java
Scottmitch: whoops, old comment let me update.

tkountis (approved):
LGTM

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ConcurrentStack.java
```diff
@@ -0,0 +1,128 @@
...
+                currTop.item = null; // best effort null out the item. pop/close will discard the Node later.
+                return true;
+            } else {
+                currTop = currTop.next;
+            }
+        }
+        return false;
+    }
+
+    void close(Consumer<T> closer) {
```
idelpivnitskiy, line 87: `requireNonNull(closer)`?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ConcurrentStack.java
```diff
@@ -0,0 +1,128 @@
...
+        @SuppressWarnings("unchecked")
+        Node<T> oldTop = topUpdater.getAndSet(this, closedNode());
+        Throwable delayedCause = null;
+        while (oldTop != null && oldTop != CLOSED) {
```
idelpivnitskiy, line 91: `CLOSED` can be only at the top. Consider removing this check from the loop and fast exit if `oldTop == CLOSED`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ConcurrentStack.java
```diff
@@ -0,0 +1,128 @@
...
+            final Node<T> next = oldTop.next;
+            final T item = oldTop.item;
+            oldTop = next;
```
idelpivnitskiy, line 94: Doesn't look like `next` variable is useful. Can we remove it and just do `oldTop = `oldTop.next`?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
idelpivnitskiy: Consider clarifying this in the comment in code

Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/1277

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
idelpivnitskiy: Actually, it makes me feel a bit uncomfortable that late subscribers may terminate differently. WDYT if instead of a `CLOSED` node inside `ConcurrentStack` we will intoduce a subtype `TerminalNode` which will keep `terminalSignal`? `ConcurrentStack` will have 2 generic types: `ConcurrentStack<Subscriber, TerminalNotification>.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleProcessor.java
```diff
@@ -88,38 +67,34 @@ public void onError(final Throwable t) {
...
+            // to be visible for any future late subscribers.
+            this.terminalSignal = terminalSignal;
+            if (terminalSignal instanceof TerminalNotification) {
+                final Throwable error = ((TerminalNotification) terminalSignal).cause();
```
idelpivnitskiy, line 90: Instead of going back and forth between types, consider using `Consumer<Subscriber<...>>` type for `terminalSignal`. Then you can use `null` instead of `TERMINAL_UNSET` and simply pass the consumer into the `stack.close(...)`.

idelpivnitskiy (commented):
You know how to have fun :)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ConcurrentStack.java
```diff
@@ -0,0 +1,128 @@
...
+    @Nullable
+    T pop() {
+        for (;;) {
+            final Node<T> oldTop = top;
+            if (oldTop == null) {
+                return null;
+            } else if (topUpdater.compareAndSet(this, oldTop, oldTop.next)) {
+                final T item = oldTop.item;
+                if (item != null) {
+                    return item;
```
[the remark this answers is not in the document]
Scottmitch: I tired to clarify the "relaxed" constraints in the java docs for `relaxedRemove`. if you don't use `relaxedRemove` then there are no special cases to consider. any suggestions on how to improve?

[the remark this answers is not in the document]
idelpivnitskiy: Should we name it `relaxedPop()` then and reference `#relaxedRemove(T)`?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleProcessor.java
Scottmitch: interesting suggestion. this makes the code cleaner but may add allocation on the success path.

idelpivnitskiy: `onSuccess` invokes this method and creates a lambda anyway, what will cause additional allocation? Assuming terminal methods are invoked only one time

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ConcurrentStack.java
```diff
@@ -0,0 +1,128 @@
...
+
+    boolean push(T item) {
+        final Node<T> newTop = new Node<>(item);
+        Node<T> oldTop;
+        do {
+            oldTop = top;
+            if (oldTop == CLOSED) {
+                return false;
+            }
+            newTop.next = oldTop;
```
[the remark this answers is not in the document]
Scottmitch: I don't think it is worth the complexity as this will only help in specific scenarios and we are assuming small sets. I think if we run into issues with the "relaxed" reclamation policy we can provide an alternative implementation which is more strict.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleProcessor.java
Scottmitch: good point, you are right let me make this change!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ConcurrentStack.java
```diff
@@ -0,0 +1,128 @@
...
+     *     <li>{@link #pop()} removes this item from another thread</li>
+     * </ul>
+     */
+    boolean relaxedRemove(T item) {
+        requireNonNull(item);
+        Node<T> currTop = top;
+        while (currTop != null) {
+            if (item.equals(currTop.item)) {
+                currTop.item = null; // best effort null out the item. pop/close will discard the Node later.
+                return true;
```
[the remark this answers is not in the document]
Scottmitch: similar comment to https://github.com/apple/servicetalk/pull/1283#discussion_r546178075 ... lets punt on trying to patch removals in this implementation.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ConcurrentStack.java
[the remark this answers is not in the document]
idelpivnitskiy: sgtm!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableProcessor.java
Scottmitch: yes I can do this. the extra memory barriers adds overhead in multithreaded cases, however it still is an improvement over what we had originally.

```
# non-atomic terminal
CompletableProcessorBenchmark.mtAdd20Complete  thrpt    5    951658.114 ±  15566.060  ops/s

# atomic terminal (with volatile coordination between threads)
CompletableProcessorBenchmark.mtAdd20Complete  thrpt    5    715757.301 ±  22471.221  ops/s 
```

Scottmitch: addressed in https://github.com/apple/servicetalk/pull/1298

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ConcurrentStack.java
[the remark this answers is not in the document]
Scottmitch: addressed in https://github.com/apple/servicetalk/pull/1298

Scottmitch:
thanks for the review @idelpivnitskiy - comments address in https://github.com/apple/servicetalk/pull/1298

PR #1284, opened 2020-12-17
CancellableSet use less expensive atomic operations

Scottmitch:
Motivation:
Recent benchmarking in CompletableProcessor has shown that a long with
GAS and IAG operations can be less costly than a reference with CAS on
every operation.

Modifications:
- CancellableSet to use a volatile long with GAS to cancel and IAG to
ensure not cancelled after insertion.

Result:
Improved performance in single threaded benchmarks, according to
similar changes in CompletableProcessor demonstrated by
CompletableProcessorBechmark for usages of CancellableSet such as
Publisher#flatMapMerge* operators.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/QueueDynamicCompositeCancellable.java
```diff
@@ -55,12 +50,14 @@ public void cancel() {
 
     @Override
     public boolean add(Cancellable toAdd) {
-        final Queue<Cancellable> currentQueue = queue;
-        if (currentQueue == null || !currentQueue.offer(toAdd)) {
+        if (counter < 0 || !queue.offer(toAdd)) { // make best effort to avoid the queue if cancelled already
```
tkountis, line 53: nit: The CLQ is unbounded so the check for false, is not needed. I would remove it for logic simplification.
Maybe with a comment in the queue declaration in case we ever change that.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/QueueDynamicCompositeCancellable.java
```diff
@@ -55,12 +50,14 @@ public void cancel() {
...
             toAdd.cancel();
             return false;
-        } else if (!queueUpdater.compareAndSet(this, currentQueue, currentQueue)) {
-            if (currentQueue.remove(toAdd)) {
+        } else if (counterUpdater.incrementAndGet(this) <= 0) {
```
tkountis, line 56: consider plain `.get()`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SetDynamicCompositeCancellable.java
```diff
@@ -81,14 +78,16 @@ public void cancel() {
...
-        if (currentSet == null) {
+        if (counter < 0) { // make best effort to avoid the set if cancelled already
             toAdd.cancel();
             return false;
-        } else if (!currentSet.add(toAdd)) {
+        } else if (!set.add(toAdd)) {
             return false; // user has implemented equals/hashCode so there is overlap?
-        } else if (!setUpdater.compareAndSet(this, currentSet, currentSet)) {
-            if (currentSet.remove(toAdd)) {
+        } else if (counterUpdater.incrementAndGet(this) <= 0) {
```
tkountis, line 86: consider plain `.get()`

tkountis (approved):

Scottmitch:
Depends upon https://github.com/apple/servicetalk/pull/1289

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SetDynamicCompositeCancellable.java
Scottmitch: The IAG (or some atomic update) is necessary to ensure this thread observes a cancellation that raced with our above queue addition.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/QueueDynamicCompositeCancellable.java
Scottmitch: this has been updated to `ConcurrentStack` and the logic has been simplified accordingly.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CancellableSet.java
```diff
@@ -0,0 +1,122 @@
...
+     * it already exists. If {@code false} then {@link Cancellable#cancel()} will be called unless the reason is
+     * {@code toAdd} has already been added.
+     */
+    boolean add(Cancellable toAdd) {
+        if (counter < 0) { // make best effort to avoid the set if cancelled already
+            toAdd.cancel();
+            return false;
+        } else if (!set.add(toAdd)) {
+            return false; // user has implemented equals/hashCode so there is overlap?
+        } else if (counterUpdater.incrementAndGet(this) <= 0) {
```
jerrinot, line 93: What is the reason for the increment? 

This `counter` feels a lot like a boolean flag to me: A negative value indicates the whole set has been cancelled (or is being cancelled) Why there cannot be a regular volatile read?

Scottmitch: The increment is an alternative approach to a `CAS` to ensure we observe any state change made by the `cancel` thread. Each thread atomically updating the variable means we can use the result here to guarantee either we have been cancelled and we should clean up, or we have not yet been cancelled (and if cancel happens in the future that thread will observe our previously added element bcz of [JMM happens-before](https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.4)). I don't see where the JMM provides guarantees around a plain volatile read (e.g. cancel thread sets the value, when is the add thread guaranteed to observe that read). For example this is a sequence where the visibility guarantees aren't clear to me...

```
T1
add(x)
cancellables.offer(x) // true

T2
cancel()
cancelledUpdater.CAS // true
cancelAll() // drains the queue, calls .cancel() on each element, but doesn't observe offer(x)

T1
isCancelled() // false -> memory not yet visible to T1
// x never has .cancel() invoked :(
```

Note the independent volatile read approach was implemented, but [recently changed](https://github.com/apple/servicetalk/commit/f4447d5fc5048edbcb620d4abc0e224525533029#diff-cecd984825c587898fcc3b8f2f68631d1a37702cb53f074e1ee9e0143b2c698c).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CancellableSet.java
```diff
@@ -0,0 +1,122 @@
...
+    @Override
+    public void cancel() {
+        if (counterUpdater.getAndSet(this, Long.MIN_VALUE) >= 0) {
+            Throwable delayedCause = null;
+            for (Cancellable c : set) {
+                try {
+                    // Removal while iterating typically results in ConcurrentModificationException, but not for
+                    // ConcurrentHashMap. We use this approach to avoid concurrent invocation of cancel() between
+                    // this method and add (if they race).
+                    if (set.remove(c)) {
```
idelpivnitskiy, line 65: IIRC it will be cheaper to create an `Iterator` and use `Iterator#remove()` here instead. We anyway cancel from a single thread, we don't need `CHM#remove(...)` guarantees here. WDYT?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CancellableSet.java
```diff
@@ -0,0 +1,122 @@
...
+import static io.servicetalk.concurrent.internal.ThrowableUtils.catchUnexpected;
+import static io.servicetalk.utils.internal.PlatformDependent.throwException;
+import static java.util.Collections.newSetFromMap;
+import static java.util.concurrent.atomic.AtomicLongFieldUpdater.newUpdater;
+
+final class CancellableSet implements Cancellable {
+    private static final AtomicLongFieldUpdater<CancellableSet> counterUpdater =
+            newUpdater(CancellableSet.class, "counter");
+    private final Set<Cancellable> set;
+    private volatile long counter;
```
idelpivnitskiy, line 33: Consider clarifying in a comment the purpose of the `counter` in this context.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CancellableSet.java
```diff
@@ -0,0 +1,122 @@
...
+    boolean add(Cancellable toAdd) {
+        if (counter < 0) { // make best effort to avoid the set if cancelled already
+            toAdd.cancel();
+            return false;
+        } else if (!set.add(toAdd)) {
+            return false; // user has implemented equals/hashCode so there is overlap?
+        } else if (counterUpdater.incrementAndGet(this) <= 0) {
+            // We must check the state after we insert into the set as the terminal event could have raced with this
+            // method and we need to ensure the subscriber is terminated.
+            // We don't check for overflow as we assume we will never exceed Long number of elements.
```
idelpivnitskiy, line 96: Hm, risky assumption. We can improve it by starting a counter from `Long.MIN_VALUE + 1` and use `Long.MIN_VALUE` as a special value instead of any negative value. It will increase the range.
In addition, consider using `accumulateAndGet` which will help to avoid setting a new value to `Long.MIN_VALUE`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CancellableSet.java
Scottmitch: `remove()` doesn't let you know if you actually removed the element or not. this change is necessary if we are doing direct removal from `add` (as opposed to going through a guarded iterate all and remove method).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CancellableSet.java
jerrinot: @Scottmitch: Thank you for taking time to provide the explanation. It made me to refresh my JMM knowledge and there is never enough JMM! I believe there is no difference between volatile read and atomic increment. They are both Synchronization Actions and provide the same visibility guarantees.   

Longer version: 
JMM provides no Wall-Clock guarantee when T1 will read the updated value, but that's irrelevant. Volatile reads and volatile writes are in totally ordered: Synchronization Order. 

Now your example has 3 valid outcomes:
1. T1 will read a positive counter value. This is only possible when T2 has not updated the counter yet. Technically it might have already been updated, but the program must behave as if it was not. Sequential Consistency and all that. This means once T2 starts iterating over the cancellable set it will find the item inserted by T1 and cancel it. 
2. T2 will see a negative value and will try to cancel it on its own
3. Obviously it's still possible both T1 and T2 will try to cancel the item, but you have this case covered. 
4. Other result would be inconsistent with JMM.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CancellableSet.java
jerrinot: overflowing 63 bit counter will take a while :-D

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CancellableSet.java
Scottmitch: > Thank you for taking time to provide the explanation. It made me to refresh my JMM knowledge and there is never enough JMM!

No problem. Thanks for your feedback on this, JMM is always a tricky topic and the more eyes the better :)

I haven't forgotten about this but will likely put this on the back burner for a while since it is an optimization. I would like to consult with some additional folks internally and think about it more, I also created a jcstress test to experiment too: https://github.com/Scottmitch/servicetalk-1/tree/jcstress_cancellable_set

Scottmitch:
This PR has become stale. I will close for now and re-visit later if necessary.

PR #1285, opened 2020-12-18
Handle `close_notify` TLS alert

idelpivnitskiy:
Motivation:

TLS protocol defined `close_notify` alert message [1], which notifies that the
sender will not send any more messages on this connection. `SslHandler`
generates `SslCloseCompletionEvent` in this case, then transport emits
`ChannelInputShutdownReadComplete` event when the FIN is received. There is a
race between these 2 user events and sending a new request on the client-side.
We MUST close down the connection immediately, discarding any pending writes,
and preventing new requests. We should also mark that the connection is
closing asap to prevent LB from selecting this connection.

1. https://tools.ietf.org/html/rfc5246#section-7.2.1

Modifications:

- Handle `SslCloseCompletionEvent` in `DefaultNettyConnection`;
- Add tests to verify client and server behavior when `SslCloseCompletionEvent`
is received;

Result:

The connection transitions to the "closing" state as soon as the
`SslCloseCompletionEvent` is observed, preventing new writes on this connection
and failing the first write, making failures retryable.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
@@ -54,6 +54,7 @@ import io.netty.channel.EventLoop;
 import io.netty.channel.socket.ChannelInputShutdownReadComplete;
 import io.netty.channel.socket.ChannelOutputShutdownEvent;
 import io.netty.handler.codec.DecoderException;
+import io.netty.handler.ssl.SslCloseCompletionEvent;
 import io.netty.handler.ssl.SslHandshakeCompletionEvent;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -641,8 +642,17 @@ public final class DefaultNettyConnection<Read, Write> extends NettyChannelListe
                 connection.closeHandler.channelClosedOutbound(ctx);
                 connection.channelOutboundListener.channelClosed(StacklessClosedChannelException.newInstance(
                         DefaultNettyConnection.class, "userEventTriggered(ChannelOutputShutdownEvent)"));
+            } else if (evt == SslCloseCompletionEvent.SUCCESS) {
+                // Received "close_notify" alert from the peer: https://tools.ietf.org/html/rfc5246#section-7.2.1.
+                // This message notifies that the sender will not send any more messages on this connection.
+
+                // Notify close handler first to enhance error reporting and prevent LB from selecting this connection
+                connection.closeHandler.channelClosedInbound(ctx);
+                // We MUST respond with a "close_notify" alert and close down the connection immediately,
+                // discarding any pending writes.
+                connection.closeHandler.closeChannelOutbound(ctx.channel());
             } else if (evt == ChannelInputShutdownReadComplete.INSTANCE) {
-                // Notify close handler first to enhance error reporting
+                // Notify close handler first to enhance error reporting and prevent LB from selecting this connection
                 connection.closeHandler.channelClosedInbound(ctx);
                 // ChannelInputShutdownEvent is not always triggered and can get triggered before we tried to read
                 // all the available data. ChannelInputShutdownReadComplete is the one that seems to (at least in
--- a/servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/AbstractSslCloseNotifyAlertHandlingTest.java
+++ b/servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/AbstractSslCloseNotifyAlertHandlingTest.java
@@ -16,7 +16,7 @@
 package io.servicetalk.transport.netty.internal;
 
 import io.servicetalk.concurrent.PublisherSource;
-import io.servicetalk.concurrent.api.test.FirstSteps;
+import io.servicetalk.concurrent.api.test.StepVerifiers;
 import io.servicetalk.concurrent.internal.ServiceTalkTestTimeout;
 import io.servicetalk.logging.api.LogLevel;
 import io.servicetalk.transport.api.ConnectionInfo.Protocol;
@@ -98,7 +98,7 @@ abstract class AbstractSslCloseNotifyAlertHandlingTest {
             assertThat("Underlying Channel is not closed", channel.isOpen(), is(false));
             assertThat("Unexpected inbound messages", channel.inboundMessages(), hasSize(0));
             assertThat("Unexpected outbound messages", channel.outboundMessages(), hasSize(0));
-            FirstSteps.create(conn.onClose()).expectComplete().verify();
+            StepVerifiers.create(conn.onClose()).expectComplete().verify();
         } finally {
             // In case of test errors, do the clean up:
             try {
@@ -117,7 +117,7 @@ abstract class AbstractSslCloseNotifyAlertHandlingTest {
 
     protected final void closeNotifyAndVerifyClosing() {
         channel.pipeline().fireUserEventTriggered(SslCloseCompletionEvent.SUCCESS);
-        FirstSteps.create(conn.onClosing()).expectComplete().verify();
+        StepVerifiers.create(conn.onClosing()).expectComplete().verify();
     }
 
     protected final void writeMsg(PublisherSource.Processor<String, String> writeSource, String msg) {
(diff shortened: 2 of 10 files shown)


servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/SslCloseNotifyAlertClientHandlingTest.java
```diff
@@ -0,0 +1,123 @@
...
+public class SslCloseNotifyAlertClientHandlingTest extends AbstractSslCloseNotifyAlertHandlingTest {
+
+    public SslCloseNotifyAlertClientHandlingTest() throws Exception {
+        super(true);
+    }
+
+    @Test
+    public void afterExchangeIdleConnection() {
+        sendRequest();
+        FirstSteps.create(conn.read())
```
idelpivnitskiy, line 37: @Scottmitch as we discussed [here](https://github.com/apple/servicetalk/pull/1240#discussion_r543650104), I will submit a PR to rename it to `StepVerifiers`.
I will keep it without static import. In the context of `PublisherStepVerifierTest`, `create(from("foo", "bar"))` looks ok. But when the test is related to some business logic, things like `create(conn.read())` are a bit confusing. Not clear what are we creating. Something like `Verifiers.stepVerifier(...)` might be a better option.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -641,8 +642,17 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
...
                         DefaultNettyConnection.class, "userEventTriggered(ChannelOutputShutdownEvent)"));
+            } else if (evt == SslCloseCompletionEvent.SUCCESS) {
+                // Received "close_notify" alert from the peer: https://tools.ietf.org/html/rfc5246#section-7.2.1.
+                // This message notifies that the sender will not send any more messages on this connection.
+
+                // Notify close handler first to enhance error reporting and prevent LB from selecting this connection
+                connection.closeHandler.channelClosedInbound(ctx);
+                // We MUST respond with a "close_notify" alert and close down the connection immediately,
+                // discarding any pending writes.
+                connection.closeHandler.closeChannelOutbound(ctx.channel());
```
idelpivnitskiy, line 653: It looks enough for me to call `closeHandler.closeChannelOutbound` here. It will generate `ChannelOutputShutdownEvent`, triggering `channelOutboundListener.channelClosed`. WDYT?

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/SslCloseNotifyAlertClientHandlingTest.java
```diff
@@ -0,0 +1,123 @@
...
+                    closeNotifyAndVerifyClosing();
+                })
+                .expectError(ClosedChannelException.class)
+                .verify();
+    }
+
+    @Test
+    public void whileWritingRequestAndReadingResponse() {
+        PublisherSource.Processor<String, String> writeSource = newPublisherProcessor();
+        FirstSteps.create(conn.write(fromSource(writeSource)).merge(conn.read()))
```
idelpivnitskiy, line 83: An example of when `FirstSteps.create` overloads for sources are useful. It's a bit annoying to use `fromSource` everywhere.

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/SslCloseNotifyAlertClientHandlingTest.java
idelpivnitskiy: Renaming: #1286

idelpivnitskiy:
Test failure due to #1124.

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/SslCloseNotifyAlertClientHandlingTest.java
```diff
@@ -0,0 +1,123 @@
...
+import io.servicetalk.concurrent.api.test.StepVerifiers;
+
+import org.junit.Test;
+
+import java.nio.channels.ClosedChannelException;
+
+import static io.servicetalk.concurrent.api.Processors.newPublisherProcessor;
+import static io.servicetalk.concurrent.api.SourceAdapters.fromSource;
+
+public class SslCloseNotifyAlertClientHandlingTest extends AbstractSslCloseNotifyAlertHandlingTest {
```
Scottmitch, line 28: can you add a test case that verifies a Retryable exception is delivered if the first write fails due to prior `close_notify` being received?

Scottmitch (commented):
change generally looks good. few questions/comments.

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/SslCloseNotifyAlertClientHandlingTest.java
Scottmitch: gotcha, we can add method overloads for the `Source` types if we will make use of them.

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/SslCloseNotifyAlertClientHandlingTest.java
Scottmitch: I'm fine with a more descriptive method name (e.g. `Verifiers.stepVerifier`). now is the time before we release 👍  feel free to open a PR.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
idelpivnitskiy: I think `channel.close()` for h2 is fine. This will happen for the stream only, not for the parent connection. For the parent channel, I think it's netty responsibility to handle this event and fail all subsequent attempts to create new child channels.

Why should we try to flush any data if the spec says that we must discard them and close immediately? IIUC, the remote peer that generated `close_nofity` alert won't be able to ready any new data on that connection because the `SSLEngine` is already closed.

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/AbstractSslCloseNotifyAlertHandlingTest.java
```diff
@@ -0,0 +1,127 @@
...
+                            closeHandler.protocolPayloadEndOutbound(ctx);
+                            promise.addListener(f -> {
+                                if (f.isSuccess()) {
+                                    closeHandler.protocolPayloadEndOutboundSuccess(ctx);
+                                }
+                            });
+                        }
+                        ctx.write(msg, promise);
+                    }
+                })), OFFLOAD_ALL_STRATEGY, mock(Protocol.class), NoopConnectionObserver.INSTANCE, isClient)
```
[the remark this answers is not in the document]
idelpivnitskiy: We previously agreed not to use offloading with embedded channels because it does not have a proper event loop and offloading leads to flaky tests. Actually, the `OFFLOAD_ALL_STRATEGY` in this case is noop, because I pass `immediate()` executor for offloading. I can create a real noop `ExecutionStrategy` to avoid confusion.

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/SslCloseNotifyAlertClientHandlingTest.java
idelpivnitskiy: I've enhanced existing tests to expect a correct subtype of `ChannelClosedException` where it's appropriate. It covers the suggested use-case.

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/SslCloseNotifyAlertClientHandlingTest.java
idelpivnitskiy: Verifiers: #1290

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/AbstractSslCloseNotifyAlertHandlingTest.java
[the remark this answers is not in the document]
Scottmitch: not suggesting we need to enhance `EmbeddedChannel` to be thread safe. the reason I suggested using different executors was to ensure expectations around dropping data are consistently validated.

Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/1294

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
Scottmitch: closing the outbound makes it so that netty itself cannot followup with any fatal alerts or `close_notify` which makes debugging harder.

idelpivnitskiy: `SslHandler` intercepts `close()` and emits `close_notify` alert. It was not intercepting half-closure (shoutdown output), but we handle it manually in `RequestResponseCloseHandler`.

Scottmitch (approved):

PR #1290, opened 2020-12-18
Improve factories for `StepVerifier`s

idelpivnitskiy:
Motivation:

1. `StepVerifiers.create` is a confusing name for users who prefer static
imports. Not obvious what the method is creating when you read the source
code of the test.
2. `Processor`s is a most common way to create a source for which users
can control emitted signals. However, users always have to use
`SourceAdapters.fromSource`, which is not convenient.

Modifications:

- Rename `StepVerifiers.create` -> `Verifiers.stepVerifiers`;
- Add `Verifiers.stepVerifiersForSource` methods for async sources;
- Use `Assert.assertThrows` instead of `try-catch + Assert.fail()`;

Result:

More convenient factory API for `StepVerifier`.


servicetalk-concurrent-api-test/src/main/java/io/servicetalk/concurrent/api/test/Verifiers.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Create a new {@link PublisherFirstStep}.
+     *
+     * @param source The {@link PublisherSource} to verify.
+     * @param <T> The type of {@link PublisherSource}.
+     * @return A {@link PublisherFirstStep} that can be used to verify {@code source}'s signal emission(s).
+     */
+    public static <T> PublisherFirstStep<T> stepVerifierForSource(PublisherSource<T> source) {
```
idelpivnitskiy, line 57: If I use the same `stepVerifier` for `PublisherSource`, java is confused when an async source implements both types `Publisher and `PublisherSource` (example: `TestPublisher`). Which approach will you prefer: different factory names or force users to use a concrete type?

servicetalk-concurrent-api-test/src/test/java/io/servicetalk/concurrent/api/test/PublisherStepVerifierTest.java
```diff
@@ -729,19 +728,14 @@ private static void verifyException(Supplier<Duration> verifier, String failedTe
...
-            assertThat("first stacktrace element expected <class: " + classNamePrefix +
-                    "> actual: " + stackTraceElements[0] + " error: " + error,
-                    stackTraceElements[0].getClassName(), startsWith(classNamePrefix));
-            StackTraceElement testMethodStackTrace = error.testMethodStackTrace();
-            assertEquals("unexpected test method failure: " + testMethodStackTrace, failedTestMethod,
-                    testMethodStackTrace.getMethodName());
-            return;
-        }
-        fail();
+        StepAssertionError error = assertThrows(StepAssertionError.class, verifier::get);
```
Scottmitch, line 731: ah this is what you were asking for ... sorry I missed this and nice improvement 👍

servicetalk-concurrent-api-test/src/test/java/io/servicetalk/concurrent/api/test/PublisherStepVerifierTest.java
```diff
@@ -729,19 +728,14 @@ private static void verifyException(Supplier<Duration> verifier, String failedTe
...
+        StackTraceElement[] stackTraceElements = error.getStackTrace();
+        assertThat(stackTraceElements.length, greaterThanOrEqualTo(1));
+        assertThat("first stacktrace element expected <class: " + classNamePrefix +
+                        "> actual: " + stackTraceElements[0] + " error: " + error,
```
Scottmitch, line 735: nit: consider putting more content on the previous line now that indentation has changed.

servicetalk-concurrent-api-test/src/main/java/io/servicetalk/concurrent/api/test/Verifiers.java
Scottmitch: I'm fine with different factory method name especially if it will come up in common cases and require additional casting/effort.

fwiw `ReactiveStreamsAdapters` uses the same method name for both, but I guess this is less commonly used...

Scottmitch (approved):

PR #1291, opened 2020-12-18
Completable#merge concurrency fix, new Completable#merge single arg o…

Scottmitch:
…verride

Motivation:
Completable#merge(Iterable) must iterate before it knows how many
Completables to wait for. The implementation uses two independent
volatile variables for the expected value and current count, which may
result in visibility issues.

Modifications:
- Completable#merge(Iterable) should use the same volatile variable to
manage the count, and avoid data races/visibility issues.
- Add Compleable#merge(Completable) and
Completable#mergeDelayError(Completable) overrides to avoid array
allocation.
- Clean up Compleable#merge implementations to remove unused methods,
and create the optimized implementation depending upon array
cardinality.

Result:
Less data races in Completable#merge(Iterable) and operator overload for
single Completable merge operations.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
@@ -413,9 +413,33 @@ public abstract class Completable {
     }
 
     /**
+     * Merges this {@link Completable} with the {@code other} {@link Completable} so that the resulting
+     * {@link Completable} terminates successfully when both of these complete or either terminates with an error.
+     * <p>
+     * This method provides a means to merge multiple asynchronous sources, fails-fast in the presence of any errors,
+     * and in sequential programming is similar to:
+     * <pre>{@code
+     *     ExecutorService e = ...;
+     *     List<Future<Void>> futures = ...;
+     *     futures.add(e.submit(() -> resultOfThisCompletable()));
+     *     futures.add(e.submit(() -> resultOfCompletable(other));
+     *     // This is an approximation, this operator does not provide any ordering guarantees for the results.
+     *     for (Future<Void> future : futures) {
+     *         future.get(); // Throws if the processing for this item failed.
+     *     }
+     * }</pre>
+     *
+     * @param other {@link Completable}s to merge.
+     * @return {@link Completable} that terminates successfully when this and {@code other} {@link Completable}s
+     * complete or terminates with an error when either terminates with an error.
+     */
+    public final Completable merge(Completable other) {
+        return new MergeOneCompletable(false, this, executor, other);
+    }
+
+    /**
      * Merges this {@link Completable} with the {@code other} {@link Completable}s so that the resulting
-     * {@link Completable} terminates successfully when all of these complete or terminates with an error when any one
-     * terminates with an error.
+     * {@link Completable} terminates successfully when all of these complete or any one terminates with an error.
      * <p>
      * This method provides a means to merge multiple asynchronous sources, fails-fast in the presence of any errors,
      * and in sequential programming is similar to:
@@ -437,7 +461,7 @@ public abstract class Completable {
      * complete or terminates with an error when any one terminates with an error.
      */
     public final Completable merge(Completable... other) {
-        return new MergeCompletable(false, this, executor, other);
+        return MergeCompletable.newInstance(false, this, executor, other);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -437,7 +461,7 @@ public final Completable concat(Completable next) {
      * complete or terminates with an error when any one terminates with an error.
      */
     public final Completable merge(Completable... other) {
-        return new MergeCompletable(false, this, executor, other);
+        return MergeCompletable.newInstance(false, this, executor, other);
```
idelpivnitskiy, line 464: If `other` is empty we can return `this` here and in `mergeDelayError`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MergeCompletable.java
```diff
@@ -15,77 +15,54 @@
...
-     * @param original {@link Completable} to merge with {@code others}.
-     * @param others {@link Completable}s to merge with {@code original}.
-     */
-    MergeCompletable(boolean delayError, Completable original, Executor executor, Completable... others) {
+    private MergeCompletable(boolean delayError, Completable original, Executor executor, Completable... others) {
         super(original, executor);
         this.delayError = delayError;
-        switch (others.length) {
-            case 0:
-                throw new IllegalArgumentException("At least one Completable required to merge");
```
idelpivnitskiy, line 46: Consider reverting this check until we fix API as described above.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MergeCompletable.java
```diff
@@ -15,77 +15,54 @@
...
+        FixedCountMergeSubscriber(Subscriber subscriber, int expectedCount, boolean delayError) {
+            super(subscriber, delayError);
+            this.expectedCount = expectedCount;
         }
 
         @Override
-        boolean isDone() {
-            return completedCount == 0;
+        boolean onTerminate() {
+            return remainingCountUpdater.incrementAndGet(this) == expectedCount;
```
idelpivnitskiy, line 65: It's actually not a counter of remaining items, it's a counter of terminated items. Consider renaming to something like `terminatedCount` or `doneCount`.

Alternatively, you can initialize `remainingCount = expectedCount` and decrement until 0. This way we will save a few bytes in this object by removing `expectedCount`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/IterableMergeCompletable.java
```diff
@@ -37,49 +35,37 @@
 
     @Override
     public MergeSubscriber apply(final Subscriber subscriber) {
-        if (others instanceof Collection) {
-            return new FixedCountMergeSubscriber(subscriber, 1 + ((Collection) others).size(), delayError);
-        } else {
-            return new DynamicCountSubscriber(subscriber, delayError);
-        }
+        return new DynamicCountSubscriber(subscriber, delayError);
```
idelpivnitskiy, line 38: Why not to keep pre-existing check for `Collection`?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/IterableMergeCompletable.java
```diff
@@ -37,49 +35,37 @@
...
-        } else {
-            for (Completable itr : others) {
-                itr.subscribeInternal(subscriber);
-            }
+        long count = 1;
+        for (Completable itr : others) {
+            ++count;
+            itr.subscribeInternal(subscriber);
         }
+        ((DynamicCountSubscriber) subscriber).setExpectedCount(count);
```
idelpivnitskiy, line 48: If you prefer to keep only `DynamicCountSubscriber`, consider adding a generic type for `AbstractMergeCompletableOperator` to avoid cast here.

idelpivnitskiy (approved):
Good catch!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -437,7 +461,7 @@ public final Completable concat(Completable next) {
      * complete or terminates with an error when any one terminates with an error.
      */
     public final Completable merge(Completable... other) {
```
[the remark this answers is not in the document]
Scottmitch: I think it makes sense for `merge(Completable)` to exist either way as you said to avoid allocations when not necessary.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MergeCompletable.java
Scottmitch: I don't think we need to prevent the empty use case you described. There is nothing incorrect about passing an empty array or empty iterator, and I think we would want consistency between the two. If folks are passing empty collections they will just incur some overhead. If we have an API which prevents it that is fine, but until then lets just allow it consistently.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/IterableMergeCompletable.java
Scottmitch: there is not longer a meaningful different between the two implementations, and some collections `size()` is costly and requires iteration.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/IterableMergeCompletable.java
Scottmitch: I experimented with this but you can't have a generic sub-type with an inner class. I'll move the inner classes so I can avoid the cast.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/IterableMergeCompletable.java
idelpivnitskiy: Agreed, difference is negligible. Saying that, do we need 2 impls or can we always use dynamic one?

idelpivnitskiy: Sorry for back and forth. Existing code lgtm as well.

Scottmitch: np. thanks for review!

PR #1292, opened 2020-12-18
DefaultHttpCookiePair#parseCookiePair more strict overflow detection

Scottmitch:
Motivation:
DefaultHttpCookiePair#parseCookiePair doesn't check for overflow when
calculating the value starting index.

Modifications:
- Check for overflow when calculating overflow index.

Result:
More robust overflow detection in DefaultHttpCookiePair#parseCookiePair.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
@@ -85,10 +85,12 @@ public final class DefaultHttpCookiePair implements HttpCookiePair {
         return parseCookiePair0(sequence, nameStart, nameLength, valueEnd < 0 ? sequence.length() : valueEnd);
     }
 
-    static HttpCookiePair parseCookiePair0(final CharSequence sequence, int nameStart, int nameLength, int valueEnd) {
+    private static HttpCookiePair parseCookiePair0(final CharSequence sequence, int nameStart, int nameLength,
+                                                   int valueEnd) {
         final int valueStart = nameStart + nameLength + 1;
         if (valueEnd <= valueStart || valueStart < 0) {
-            throw new IllegalArgumentException("unexpected format of cookie pair, empty value");
+            throw new IllegalArgumentException("value indexes are invalid. valueStart: " + valueStart
+                    + " valueEnd: " + valueEnd);
         }
         if (sequence.charAt(valueStart) == '"' && sequence.charAt(valueEnd - 1) == '"') {
             if (valueEnd - 2 <= valueStart) {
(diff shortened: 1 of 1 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
```diff
@@ -87,7 +87,7 @@ public static HttpCookiePair parseCookiePair(final CharSequence sequence, int na
 
     static HttpCookiePair parseCookiePair0(final CharSequence sequence, int nameStart, int nameLength, int valueEnd) {
```
idelpivnitskiy, line 88: While you are here, can you please make this method private?

Btw, are the previous two `parseCookiePair` method have to be public? They currently used only by `HeaderUtils`, where `HeaderUtils#parseCookiePair` is also public. If any of these methods have to be public, can we consolidate them all in one place?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
```diff
@@ -87,7 +87,7 @@ public static HttpCookiePair parseCookiePair(final CharSequence sequence, int na
...
         final int valueStart = nameStart + nameLength + 1;
-        if (valueEnd - 1 < valueStart) {
+        if (valueEnd <= valueStart || valueStart < 0) {
```
idelpivnitskiy, line 90: `valueStart` is always >= 0

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
```diff
@@ -87,7 +87,7 @@ public static HttpCookiePair parseCookiePair(final CharSequence sequence, int na
...
             throw new IllegalArgumentException("unexpected format of cookie pair, empty value");
```
idelpivnitskiy, line 91: This exception is a bit confusing and doesn't tell what was wrong with the arguments. Consider validating some passed params independently and thow if any of them are incorrect. For example, if `nameStart < 0` or `nameLength` is not positive.

idelpivnitskiy (approved):
Change LGTM. I left a few other comments for the areas nearby, can be in a follow-up PR

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
idelpivnitskiy: Ignore the previous comment for now. Only if we validate each input param independently, it will make sense.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
Scottmitch: > While you are here, can you please make this method private?

done.

I can make both of the `parseCookiePair` methods package private, they may have been used by the `netty` packages at some point during development. i'll do this as a followup PR.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
Scottmitch: I'll adjust the exception message to include more context.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
Scottmitch: package private PR: https://github.com/apple/servicetalk/pull/1296

PR #1300, opened 2020-12-21
Avoid incorrectly throwing `QueueFullException` from `PublisherAsBlockingIterable`

idelpivnitskiy:
Motivation:

`PublisherAsBlockingIterable` uses an intermediate queue to keep the received
items from the `Publisher` until users consume them. However, `onNextQueued`
counter is updated when users consume `next()` item, but it's actually removed
from the queue inside `hasNext()`. As the result, `QueueFullException` is
generated when the queue is still able to fit one more element.

Modifications:

- Removed unnecessary `maxBufferedItems` internal variable;
- Update `onNextQueued` counter together with `itemsToNextRequest` counter
inside `hasNext()`;
- Enhance
`PublisherAsBlockingIterableTest#replenishingRequestedShouldHonourQueueContents()`
test to account for this use-case;

Result:

`PublisherAsBlockingIterable` correctly updates `onNextQueued` counter.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherAsBlockingIterable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherAsBlockingIterable.java
@@ -81,6 +81,7 @@ final class PublisherAsBlockingIterable<T> implements BlockingIterable<T> {
         private final BlockingQueue<Object> data;
         private final DelayedSubscription subscription = new DelayedSubscription();
         private final int queueCapacity;
+        private final int requestN;
         private volatile int onNextQueued;
         /**
          * Number of items to emit from {@link #next()} till we request more.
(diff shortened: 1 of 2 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherAsBlockingIterable.java
```diff
@@ -103,8 +104,10 @@
...
 
         SubscriberAndIterator(int queueCapacity) {
-            maxBufferedItems = queueCapacity;
-            itemsToNextRequest = max(1, maxBufferedItems / 2);
+            itemsToNextRequest = max(1, queueCapacity / 2);
+            // We set this value ahead of time to avoid modifications from different threads. This will be true as
+            // soon as onSubscribe is called.
+            itemsRequested = itemsToNextRequest;
             this.queueCapacity = queueCapacity;
             data = new LinkedBlockingQueue<>();
```
idelpivnitskiy, line 112: @Scottmitch while I'm here, does it make sense to switch to `ArrayBlockingQueue` (because the queue capacity is known in advance and is has an upper bound of 128), account for one more terminal event, and adjust `itemsToNextRequest` to max(1, queueCapacity - 1)`?

tkountis (approved):
LGTM

idelpivnitskiy:
Failure attributed to #1305

PR #1303, opened 2021-01-07
Make `AsyncContext` visible from `TransportObserver#onNewConnection()`

idelpivnitskiy:
Motivation:

Users who want to correlate `TransportObserver#onNewConnection()` events with
the request object may use `AsyncContext` to propagate correlation key.
However, `AsyncContext` is not visible from `TransportObserver#onNewConnection()`
because it runs from IO-thread. It's also important to invoke
`TransportObserver#onNewConnection()` callback earlier on the client-side.

Modifications:

- Trigger `TransportObserver#onNewConnection()` as soon as someone subscribes to
`TcpConnector#connect` single on the client-side to get access to the caller's
thread `AsyncContext`;
- Trigger `TransportObserver#onNewConnection()` at the beginning of `initChannel`
on the server-side for consistency with the client-side;
- Update internal classes and tests to use new API;
- Add tests to verify `AsyncContext` is visible for expected callbacks;

Result:

`AsyncContext` is visible from `TransportObserver#onNewConnection()` on the
client-side.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverAsyncContextTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverAsyncContextTest.java
@@ -41,7 +41,7 @@ import org.junit.runners.Parameterized.Parameters;
 
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 
 import static io.servicetalk.concurrent.api.AsyncContextMap.Key.newKey;
@@ -154,7 +154,7 @@ public class HttpTransportObserverAsyncContextTest extends AbstractNettyHttpServ
 
     private static final class AsyncContextCaptureTransportObserver implements TransportObserver {
 
-        private final Map<String, String> storageMap = new ConcurrentSkipListMap<>();
+        private final Map<String, String> storageMap = new ConcurrentHashMap<>();
         private final AsyncContextMap.Key<String> key;
 
         AsyncContextCaptureTransportObserver(AsyncContextMap.Key<String> key) {
@@ -167,6 +167,8 @@ public class HttpTransportObserverAsyncContextTest extends AbstractNettyHttpServ
 
         @Override
         public ConnectionObserver onNewConnection() {
+            // Use String.valueOf(...) here and in all other callbacks to prevent passing `null` value to the
+            // ConcurrentHashMap which does not allow `null` values:
             storageMap.put("onNewConnection", valueOf(AsyncContext.get(key)));
             return new AsyncContextCaptureConnectionObserver();
         }
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverAsyncContextTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverAsyncContextTest.java
@@ -59,9 +59,9 @@ import static io.servicetalk.http.netty.TestServiceStreaming.SVC_THROW_ERROR;
 import static java.lang.String.valueOf;
 import static java.util.Collections.unmodifiableMap;
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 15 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverAsyncContextTest.java
```diff
@@ -0,0 +1,328 @@
...
+            }
+
+            @Override
+            public void onFlush() {
+                // AsyncContext is unknown at this point because protocols can flush multiple requests concurrently
+            }
+
+            @Override
+            public SecurityHandshakeObserver onSecurityHandshake() {
+                // AsyncContext is unknown at this point because this event is triggered by network
```
idelpivnitskiy, line 193: In some cases, `ConnectionObserver` events are generated by network events when the request-related `AsyncContext` is unknown. For those cases, users should capture the `AsyncContext` state from other callbacks, keep it in local variable(s), and propagate when necessary. Examples:
1. `ConnectionObserver#onSecurityHandshake()` and all `SecurityHandshakeObserver` callbacks should use the state captured in `TransportObserver#onNewConnection()`.
2. `ReadObserver`/`WriteObserver` events should use the state captured in `DataObserver` callbacks.
3. `StreamObserver#streamClosed` events should use the state captured in `StreamObserver#streamEstablished()` or `MultiplexedObserver#onNewStream()`.
4. `connectionClosed` events can be correlated with request(s) using `requestId<->connectionId` correlation at the higher levels.

@Scottmitch we can cover more cases (for example, `onSecurityHandshake `) using another technique if necessary. Because we have access to `subscriber` at the time we invoke this callback, we can check if it's an instance of `ContextPreservingSubscriber` and extract the `AsyncContextMap` from it. However, it will require some movements for `ContextPreserving*` classes. I will defer that until we have a use-case.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverAsyncContextTest.java
```diff
@@ -0,0 +1,328 @@
...
+        StreamingHttpResponse response = streamingHttpClient().request(newRequest(path))
+                .toFuture().get();
+        assertResponse(response, protocol.version, OK);
+        assertThrows(ExecutionException.class, () -> response.payloadBody().ignoreElements().toFuture().get());
+        assertMap(clientObserver.storageMap(), CLIENT_VALUE);
+    }
+
+    private static void assertMap(Map<String, String> map, String expected) {
+        Set<Map.Entry<String, String>> entries = map.entrySet();
+        assertThat("No entries in the map", entries, hasSize(greaterThan(0)));
```
Scottmitch, line 148: nit `hasSize(greaterThan(0))` -> `not(Matchers.empty())` (prefer `isEmpty()` instead of `size()` if you don't need the actual size)

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverAsyncContextTest.java
Scottmitch: agreed lets defer until necessary.

Scottmitch (approved):

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverAsyncContextTest.java
```diff
@@ -0,0 +1,328 @@
...
+            this.key = key;
+        }
+
+        Map<String, String> storageMap() {
+            return unmodifiableMap(storageMap);
+        }
+
+        @Override
+        public ConnectionObserver onNewConnection() {
+            storageMap.put("onNewConnection", valueOf(AsyncContext.get(key)));
```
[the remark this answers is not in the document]
idelpivnitskiy: `ConcurrentHashMap` does not allow `null` values, `String.valueOf(nullableValue)` helps to convert `null` into "null" string.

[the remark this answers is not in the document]
Scottmitch: gotcha. consider adding comment to clarify.

PR #1304, opened 2021-01-07
Wrap all netty's HTTP/2 related exceptions with `IOException`

idelpivnitskiy:
Motivation:

Netty's `Http2Exception` and `Http2FrameStreamException` extend `Exception`
instead of `IOException`. `IOException` provide better user-experience for
interception of HTTP/2 errors and more compatible with exception handling
logic in other libraries.

Modifications:

- Create a new `io.servicetalk.http.netty.Http2Exception` class that extends
`IOException`;
- Wrap all netty's HTTP/2 related exceptions with this new type;
- Change pre-existing `H2StreamRefusedException` and `H2StreamResetException`
to also extend the new type;
- Include streamId in `H2StreamRefusedException` and `H2StreamResetException`
exception messages;
- Update tests that expect netty exception types;

Result:

All HTTP/2 related exceptions extend `IOException`.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/Http2Exception.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/Http2Exception.java
@@ -20,9 +20,9 @@ import io.servicetalk.transport.api.RetryableException;
 import java.io.IOException;
 
 /**
- * An exception that indicates <a href="https://tools.ietf.org/html/rfc7540">HTTP/2</a> protocol error.
+ * An exception that indicates <a href="https://tools.ietf.org/html/rfc7540#section-5.4">HTTP/2 error</a>.
  */
-public class Http2Exception extends IOException {
+class Http2Exception extends IOException {
     private static final long serialVersionUID = 745695232431963628L;
 
     Http2Exception(final String message) {
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/Http2Exception.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/Http2Exception.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+import io.servicetalk.transport.api.RetryableException;
+
+import java.io.IOException;
+
+/**
+ * An exception that indicates <a href="https://tools.ietf.org/html/rfc7540">HTTP/2</a> protocol error.
+ */
+public class Http2Exception extends IOException {
+    private static final long serialVersionUID = 745695232431963628L;
+
+    Http2Exception(final String message) {
+        super(message);
+    }
+
+    Http2Exception(final Throwable cause) {
+        super(cause);
+    }
+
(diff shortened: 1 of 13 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamObserverTest.java
```diff
@@ -162,15 +162,16 @@ public void maxActiveStreamsViolationError() throws Exception {
...
 
             verify(clientMultiplexedObserver, times(2)).onNewStream();
             verify(clientStreamObserver, times(2)).streamEstablished();
             verify(clientDataObserver, times(2)).onNewRead();
             verify(clientDataObserver, times(2)).onNewWrite();
             verify(clientReadObserver).readCancelled();
-            verify(clientWriteObserver).writeFailed(e.getCause());
-            verify(clientStreamObserver, await()).streamClosed(e.getCause());
+            verify(clientWriteObserver).writeFailed(e.getCause().getCause());
+            verify(clientStreamObserver, await()).streamClosed(e.getCause().getCause());
```
idelpivnitskiy, line 174: `WriteObserver` and `StreamObserver` report the unwrapped netty's exception because the reporting happens in `WriteStreamSubscriber`:
https://github.com/apple/servicetalk/blob/c3ff7bf27f461d6a138f48c784e3ceab8678bccb/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java#L478-L480

The `Single<StreamingHttpResponse>` sees wrapped `IOException`.
@Scottmitch wdyt, should we enrich the error in `WriteStreamSubscriber`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/Http2Exception.java
```diff
@@ -0,0 +1,82 @@
...
+package io.servicetalk.http.netty;
+
+import io.servicetalk.transport.api.RetryableException;
+
+import java.io.IOException;
+
+/**
+ * An exception that indicates <a href="https://tools.ietf.org/html/rfc7540">HTTP/2</a> protocol error.
+ */
+public class Http2Exception extends IOException {
```
Scottmitch, line 25: does this type need to be public?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/Http2Exception.java
```diff
@@ -0,0 +1,82 @@
...
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+import io.servicetalk.transport.api.RetryableException;
+
+import java.io.IOException;
+
+/**
+ * An exception that indicates <a href="https://tools.ietf.org/html/rfc7540">HTTP/2</a> protocol error.
```
Scottmitch, line 23: [protocol error](https://tools.ietf.org/html/rfc7540#section-11.4) is a distinct concept in the h2 RFCs ... consider changing to `<a href="https://tools.ietf.org/html/rfc7540#section-5.4">HTTP/2 error</a>`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamObserverTest.java
Scottmitch: if we are going to wrap the exception seems like we should wrap it everywhere it is exposed.

Scottmitch (commented):
generally lgtm, few comments/questions

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/Http2Exception.java
idelpivnitskiy: I made it public with a private ctor to allow users try-catch it (or check instanceof) without allowing them to create instances of this exception type.

Scottmitch: do we expect users to react differently to `Http2Exception` vs `IOException`? Currently `Http2Exception` doesn't expose any additional methods so it isn't clear when we would expect folks to know about the type.

idelpivnitskiy: sounds reasonable, let me hide it for now

Scottmitch (approved):

PR #1307, opened 2021-01-08
Don't close entire HTTP/2 connection when request is cancelled

idelpivnitskiy:
Motivation:

`LoadBalancedStreamingHttpClient` pessimistically assumes the connection
that serves a request should be closed on request cancellation. However,
in case of HTTP/2 it closes the parent TCP connection instead of a stream
(because stream is not accessible from the connection API). It also fails
all other in-flight requests on the same connection.

Modifications:

- In case of HTTP/2 and above mark request as finished instead of closing
the connection when cancellation happens;
- Test that request cancellation doesn't close parent TCP connection for
HTTP/2 protocol;
- Change `HttpConnectionContext#protocol()` to return `HttpProtocolVersion`;
- Change `GrpcConnectionContext.GrpcProtocol` to return `HttpProtocolVersion`;

Result:

HTTP/2 request cancellation doesn't close the TCP connection.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
@@ -1,0 +1,188 @@
+/*
+ * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+import io.servicetalk.client.api.DelegatingConnectionFactory;
+import io.servicetalk.concurrent.Cancellable;
+import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.http.api.FilterableStreamingHttpConnection;
+import io.servicetalk.http.api.HttpServiceContext;
+import io.servicetalk.http.api.ReservedStreamingHttpConnection;
+import io.servicetalk.http.api.StreamingHttpConnection;
+import io.servicetalk.http.api.StreamingHttpRequest;
+import io.servicetalk.http.api.StreamingHttpRequestFactory;
+import io.servicetalk.http.api.StreamingHttpRequester;
+import io.servicetalk.http.api.StreamingHttpResponse;
+import io.servicetalk.http.api.StreamingHttpResponseFactory;
+import io.servicetalk.http.api.StreamingHttpServiceFilter;
+import io.servicetalk.http.netty.H2ToStH1Utils.H2StreamResetException;
+import io.servicetalk.transport.api.TransportObserver;
+
+import org.junit.Test;
+
+import java.net.InetSocketAddress;
+import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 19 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
```diff
@@ -0,0 +1,188 @@
...
+        firstRequestReceivedLatch.await();
+
+        AtomicReference<Cancellable> cancellable = new AtomicReference<>();
+        stepVerifier(requester.request(defaultStrategy(), newRequest(requester, "second"))
+                        .whenOnSuccess(responses::add)) // Add response to the queue to verify that we never receive it
+                .expectCancellableConsumed(cancellable::set)
+                .then(() -> {
+                    try {
+                        secondRequestReceivedLatch.await();
+                        cancellable.get().cancel(); // If I use thenCancel() the current then(Runnable) does not run
```
idelpivnitskiy, line 166: @Scottmitch if I do this way:
```java
stepVerifier(requester.request(defaultStrategy(), newRequest(requester, "second"))
                .whenOnSuccess(responses::add)) // Add response to the queue to verify that we never receive it
        .expectCancellable()
        .then(() -> {
            try {
                secondRequestReceivedLatch.await();
                // ignore
            }
        })
        .thenCancel()
        .verify();
```
`then(Runnable)` is not invoked, after `expectCancellable()` the next step will be `thenCancel()`. Is this expected behavior?

Scottmitch: discussed offline, we will investigate clarifying/improving as part of followup issues. tldr; the "then" runnable operations are run on the verifier thread, the expectations are run on the subscriber thread (including `thenCancel`), so these aren't sequential.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
```diff
@@ -0,0 +1,188 @@
...
+                    } catch (InterruptedException e) {
+                        // ignore
+                    }
+                })
+                .expectError()
+                // .thenCancel()
+                .verify();
+
+        assertThat("Unexpected responses", responses, hasSize(0));
```
Scottmitch, line 175: nit: `hasSize(0)` -> `is(empty())`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
```diff
@@ -0,0 +1,188 @@
...
+        firstResponseLatch.countDown();
+
+        assertResponse(responses.take(), HTTP_2_0, OK, "first");
+        // Mak sure we can use the same connection for future requests:
```
Scottmitch, line 179: nit: `Mak` -> `Make`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
```diff
@@ -0,0 +1,188 @@
...
+    public void testClient() throws Exception {
+        // Release existing connection to avoid creating unexpected number of new h2 connections because of randomness
+        // in RRLB (it may randomly pick the reserved connection until all attempts are taken).
+        ((ReservedStreamingHttpConnection) streamingHttpConnection()).releaseAsync().toFuture().get();
+
+        int connectionsBefore = newConnectionsCounter.get();
+        requestCancellationResetsStreamButNotParentConnection(streamingHttpClient());
+        int connectionsAfter = newConnectionsCounter.get();
+        assertThat("Client unexpectedly created more connections instead of reusing the existing one",
+                connectionsAfter - connectionsBefore, is(0));
```
Scottmitch, line 124: nit: `assertThat(connectionsAfter, is(connectionsBefore));` or `assertEquals(connectionsBefore, connectionsAfter)` will give a more informative error message if a failure occurs.

Scottmitch (approved):

PR #1308, opened 2021-01-08
Share `AsyncContext` when `RequestTarget*Filter`(s) are used

idelpivnitskiy:
Motivation:

`RequestTarget*Filter`(s) use `Single.defer` operator to modify the
request-target on each subscribe, however it doesn't share the
`AsyncContext`. As the result, context changes made during request
processing won't be visible during processing response payload body.

Modifications:

- Add `subscribeShareContext()` operator for all `RequestTarget*Filter`(s);

Result:

Correct `AsyncContext` propagation in `RequestTarget*Filter`(s).


tkountis (approved):
LGTM consider adding a test

idelpivnitskiy:
@tkountis good point! To avoid duplication in tests for each filter I've created #1312.

PR #1314, opened 2021-01-13
Identify retryable `Http2Exception`(s) generated by netty

idelpivnitskiy:
Motivation:

HTTP/2 codec may throw an exception when it creates a stream in some cases:
- When there is a concurrency between closing the parent h2 connection and
creating a new stream;
- When stream IDs are exhausted for this endpoint;
- When maximum active streams violated for this endpoint;
In all this cases it's safe to retry the request because we do not start
writing it on the wire. When parent connection is closing concurrently, we
won't hit this error again because it will be marked as "closing" and LB
won't select this connection again.

Modification:

- Wrap `Http2Exception`(s) with `Http2Error.REFUSED_STREAM` as retryable;
- Wrap `Http2NoMoreStreamIdsException` as retryable;

Result:

`HttpClient` will automatically retry recoverable HTTP/2 exceptions.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/Http2Exception.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/Http2Exception.java
@@ -17,7 +17,11 @@ package io.servicetalk.http.netty;
 
 import io.servicetalk.transport.api.RetryableException;
 
+import io.netty.handler.codec.http2.Http2NoMoreStreamIdsException;
+
 import java.io.IOException;
+
+import static io.netty.handler.codec.http2.Http2Error.REFUSED_STREAM;
 
 /**
  * An exception that indicates <a href="https://tools.ietf.org/html/rfc7540#section-5.4">HTTP/2 error</a>.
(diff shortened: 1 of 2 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/Http2Exception.java
```diff
@@ -35,14 +39,36 @@
...
+     * @return {@code true} if {@link io.netty.handler.codec.http2.Http2Exception} is retryable on a different h2
+     * parent connection.
+     */
+    private static boolean isRetryable(final io.netty.handler.codec.http2.Http2Exception cause) {
+        // The first check captures cases like:
+        //  - Cannot create stream %d greater than Last-Stream-ID %d from GOAWAY.
+        //  - Stream IDs are exhausted for this endpoint.
+        //  - Maximum active streams violated for this endpoint.
+        //  - Http2ChannelClosedException
+        return cause.error() == REFUSED_STREAM
```
Scottmitch, line 67: There is code in `AbstractH2DuplexHandler` that also checks for `REFUSED_STREAM` and throws a `Retryable` exception. is there a way to have these two areas share code so the relationship is more clear and less likely to miss cases in the future (e.g. if we want to add `ENHANCE_YOUR_CALM` or others)?

Scottmitch (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/Http2Exception.java
idelpivnitskiy: I thought about it but did not found a good way to do that because in `AbstractH2DuplexHandler` it handles a user-event and created a new exception based on that event when here we intercept write failures. May give another round of thoughts after the release.

idelpivnitskiy: Follow-up with the fresh head :)
https://github.com/apple/servicetalk/pull/1318

PR #1319, opened 2021-01-14
EmptySubscription propagate error to Subscriber

Scottmitch:
Motivation:
EmptySubscription throws an exception into the calling thread if illegal
demand is detected. However throwing into the calling thread may not be
desirable and may not correctly propagate errors to the Subscriber as
required by the RS spec.

Modifications:
- EmptySubscription is now a utility class EmptySubscriptions that
  supports creating different types of empty subscriptions (without a
  Subscriber and with a Subscriber).
- Update EmptySubscription usage so that cases which are more than just
  references use the correct variant (e.g. SequentialSubscription).

Result:
More correct error propagation for EmptySubscription use cases.


Scottmitch:

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/EmptySubscriptions.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param subscriber The subscriber to propagate
+     * @param <T> The type of {@link Subscriber}.
+     * @return An empty {@link Subscription} that will propagate an error to a {@link Subscriber} upon invalid demand.
+     */
+    public static <T> Subscription newEmptySubscription(Subscriber<T> subscriber) {
+        return new Subscription() {
+            private boolean terminated;
+            @Override
+            public void request(final long n) {
+                if (!terminated && !isRequestNValid(n)) {
```
idelpivnitskiy, line 45: The [rule 3.9](https://github.com/reactive-streams/reactive-streams-jvm#3.9) says:

> While the `Subscription` is not cancelled, `Subscription.request(long n)` MUST signal `onError` ...

We need to mark it as `terminated` inside `cancel()`. That also helps to satisfy [rule 3.6](https://github.com/reactive-streams/reactive-streams-jvm#3.6):

> After the `Subscription` is cancelled, additional `Subscription.request(long n)` MUST be NOPs.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/EmptySubscriptions.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+public final class EmptySubscriptions {
+    public static final Subscription EMPTY_SUBSCRIPTION = newEmptySubscription(true);
+    public static final Subscription EMPTY_SUBSCRIPTION_NO_THROW = newEmptySubscription(false);
+
+    private EmptySubscriptions() {
+    }
+
+    /**
+     * Create an empty {@link Subscription} that will propagate an error to a {@link Subscriber} upon invalid demand.
```
idelpivnitskiy, line 35: Consider adding a link to the https://github.com/reactive-streams/reactive-streams-jvm#3.9

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/EmptySubscriptions.java
```diff
@@ -0,0 +1,89 @@
...
+     * {@code false} will not validate demand.
+     * @return A {@link Subscription} that will optionally validate invalid demand.
+     */
+    public static Subscription newEmptySubscription(boolean throwOnInvalidDemand) {
+        if (throwOnInvalidDemand) {
+            return new Subscription() {
+                @Override
+                public void request(final long n) {
+                    if (!isRequestNValid(n)) {
+                        throw newExceptionForInvalidRequestN(n);
```
idelpivnitskiy, line 69: Why do we still need the variant that throws if the spec says that `onError` should be used in this case?

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/EmptySubscriptions.java
```diff
@@ -0,0 +1,89 @@
...
+                    }
+                }
+
+                @Override
+                public void cancel() {
+                }
```
idelpivnitskiy, line 75: If the subscription was canceled we should not throw from `request(n)`.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/EmptySubscriptions.java
```diff
@@ -0,0 +1,89 @@
...
+        };
+    }
+
+    /**
+     * Create an empty {@link Subscription} that will optionally validate invalid demand.
+     * @param throwOnInvalidDemand {@code true} to throw on invalid {@link Subscription#request(long)} demand.
+     * {@code false} will not validate demand.
+     * @return A {@link Subscription} that will optionally validate invalid demand.
+     */
+    public static Subscription newEmptySubscription(boolean throwOnInvalidDemand) {
```
idelpivnitskiy, line 63: If we need to use this method in the future for new use-cases what is the rationale for choosing between `true` and `false`? Consider clarifying that in the javadoc, it will help new team members and OSS folks.

idelpivnitskiy (changes requested):

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/EmptySubscriptions.java
Scottmitch: humm good point. I think the important part is we don't terminate the subscriber more than once, seems strange to "suppress" exceptions just because cancel had been previously called, but it can be done simple enough.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/EmptySubscriptions.java
Scottmitch: In some cases we use a reference to a `Subscription` for type safety reasons, and use it internally in our state machines. We don't expect to invoke `request(n)` on it, but will throw if we do and the `n` value is invliad as a failsafe to point out the issue.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/EmptySubscriptions.java
Scottmitch: I'll make this internal since all usages are through the `static final` variable.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/EmptySubscriptions.java
Scottmitch: the reference from this method is currently stored in a static context, and would have to be made non-static to facilitate this requirement. I don't think we need to strictly enforce this rule because it is effectively invalid use of the API anyways.

idelpivnitskiy (approved):
🚀

PR #1320, opened 2021-01-14
Add PayloadWriter#close(Throwable)

Scottmitch:
Motivation:
In some scenarios the PayloadWriter wraps a Subscriber (e.g.
ConnectablePayloadWriter), and the close may occur as a result of an
error that needs to be propagated to the Subscriber.

Modifications:
- Add close(Throwable) to the PayloadWriter API
- ConnectablePayloadWriter calls subscriber.onError(cause) for the
close(Throwable) method
- BlockingStreamingToStreamingService to use close(Throwable) for more
robust cleanup

Result:
PayloadWriter use cases can close while indicating the error reason.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
@@ -16,9 +16,8 @@
 package io.servicetalk.http.api;
 
 import io.servicetalk.buffer.api.Buffer;
-import io.servicetalk.concurrent.CompletableSource;
 import io.servicetalk.concurrent.CompletableSource.Processor;
-import io.servicetalk.concurrent.PublisherSource;
+import io.servicetalk.concurrent.PublisherSource.Subscription;
 import io.servicetalk.concurrent.SingleSource.Subscriber;
 import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.Publisher;
@@ -42,7 +41,6 @@ import static io.servicetalk.http.api.HeaderUtils.isTransferEncodingChunked;
 import static io.servicetalk.http.api.HttpExecutionStrategies.OFFLOAD_RECEIVE_META_STRATEGY;
 import static io.servicetalk.http.api.HttpHeaderNames.TRANSFER_ENCODING;
 import static io.servicetalk.http.api.HttpHeaderValues.CHUNKED;
-import static io.servicetalk.http.api.HttpProtocolVersion.HTTP_1_0;
 import static io.servicetalk.http.api.HttpRequestMethod.HEAD;
 import static io.servicetalk.http.api.HttpResponseStatus.OK;
 import static io.servicetalk.http.api.StreamingHttpResponses.newTransportResponse;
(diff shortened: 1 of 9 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -142,10 +136,16 @@ public void cancel() {
...
                         safeOnError(subscriber, cause);
-                    } else if (payloadWriterOuter.markSubscriberComplete()) {
-                        payloadProcessor.onError(cause);
                     } else {
-                        LOGGER.error("An exception occurred after the response was sent", cause);
+                        try {
+                            payloadProcessor.onError(cause);
+                        } finally {
+                            try {
+                                payloadWriter.close(cause);
```
idelpivnitskiy, line 144: Discussed offline: we should adjust `payloadProcessor` to correctly propagate the error instead of using `payloadWriter.close(cause)` in this case.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -165,25 +165,29 @@ public Completable closeAsyncGracefully() {
...
+            this.subscriber = threadSafeSubscriber;
         }
 
         @Override
         public void write(final Buffer object) throws IOException {
             payloadWriter.write(object);
         }
 
+        @Override
+        public void close(final Throwable cause) throws IOException {
```
idelpivnitskiy, line 183: Consider moving this `close(Throwable)` method next to its brother `close()`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
Scottmitch: I think we can use both, to make sure the `payloadWriter` resources are reclaimed.

Scottmitch: The larger changes we discussed related to the `CompletableProcessor` will have to be delayed until my followup PR which fixes payload body and trailers propagation (otherwise gRPC `ErrorHandlerTest` fails for blocking streaming use cases).

Scottmitch:
build failure due to https://github.com/apple/servicetalk/issues/999

idelpivnitskiy (approved):

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -97,36 +90,33 @@ protected void handleSubscribe(final Subscriber<? super StreamingHttpResponse> s
                             HttpHeaders headers = metaData.headers();
                             boolean addTrailers = isTransferEncodingChunked(headers);
                             if (!addTrailers && !HTTP_1_0.equals(metaData.version()) && !hasContentLength(headers) &&
+                                    !hasContentLength(headers) &&
```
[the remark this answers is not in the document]
Scottmitch: good catch! copy/pasta from my other future PR :)

Issue #1287, opened 2020-12-18
SingleRequestOrResponseApiTest.serverResponseStreamingRouteFailsOnSecondRequestItem[streamingService=false, streamingClient=false] test failure


Scottmitch:
failed locally. gradle, jdk11.0.9, macOS.

```
java.lang.AssertionError: unexpected exception type thrown; expected:<io.servicetalk.grpc.api.GrpcStatusException> but was:<io.servicetalk.http.netty.H2ToStH1Utils.H2StreamResetException>
	at org.junit.Assert.assertThrows(Assert.java:1020)
	at org.junit.Assert.assertThrows(Assert.java:981)
	at io.servicetalk.grpc.netty.SingleRequestOrResponseApiTest.serverResponseStreamingRouteFailsWithInvalidArgument(SingleRequestOrResponseApiTest.java:132)
	at io.servicetalk.grpc.netty.SingleRequestOrResponseApiTest.serverResponseStreamingRouteFailsOnSecondRequestItem(SingleRequestOrResponseApiTest.java:124)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at io.servicetalk.concurrent.internal.ServiceTalkTestTimeout$TimeoutStatement$CallableStatement.call(ServiceTalkTestTimeout.java:171)
	at io.servicetalk.concurrent.internal.ServiceTalkTestTimeout$TimeoutStatement$CallableStatement.call(ServiceTalkTestTimeout.java:163)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[...]

PR #1321, opened 2021-01-14
Infer HttpVersion in requests/responses based upon protocol selection

Scottmitch:
Motivation:
HTTP/1.x protocol requires the version is included in the request and
response. When generating requests/response objects in the application
we tend to default to HTTP_1_1 despite the protocol and user
preferences. This may result in additional conversions and confusion
when using version accessors.

Modifications:
- For request/response factories that can derive the real version from
  the transport they should use the real version
- For use cases that don't know which protocol will be used (e.g. Client
  where h1, h2, etc. is requested) use the version most preferred
locally. Worst case additional conversions will be required if the
preferred protocol isn't used.

Result:
HttpVersion in request/response objects will be more accurate and less
conversions will be required.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultPartitionedHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultPartitionedHttpClientBuilder.java
@@ -68,8 +68,10 @@ import static io.servicetalk.buffer.netty.BufferAllocators.DEFAULT_ALLOCATOR;
 import static io.servicetalk.concurrent.api.Completable.completed;
 import static io.servicetalk.concurrent.api.Single.defer;
 import static io.servicetalk.concurrent.api.Single.failed;
+import static io.servicetalk.http.api.HttpProtocolVersion.HTTP_1_1;
 import static io.servicetalk.http.netty.DefaultSingleAddressHttpClientBuilder.SD_RETRY_STRATEGY_INIT_DURATION;
 import static io.servicetalk.http.netty.DefaultSingleAddressHttpClientBuilder.SD_RETRY_STRATEGY_JITTER;
+import static io.servicetalk.http.netty.DefaultSingleAddressHttpClientBuilder.defaultReqRespFactory;
 import static java.util.Objects.requireNonNull;
 import static java.util.function.Function.identity;
 
@@ -117,7 +119,9 @@ class DefaultPartitionedHttpClientBuilder<U, R> extends PartitionedHttpClientBui
 
         DefaultPartitionedStreamingHttpClientFilter<U, R> partitionedClient =
                 new DefaultPartitionedStreamingHttpClientFilter<>(psdEvents, serviceDiscoveryMaxQueueSize,
-                        clientFactory, partitionAttributesBuilderFactory, buildContext.reqRespFactory,
+                        clientFactory, partitionAttributesBuilderFactory,
+                        defaultReqRespFactory(buildContext.httpConfig().asReadOnly(),
+                                buildContext.executionContext.bufferAllocator()),
                         buildContext.executionContext, partitionMapFactory);
         return new FilterableClientToClient(partitionedClient, buildContext.executionContext.executionStrategy(),
                 buildContext.builder.buildStrategyInfluencerForClient(
@@ -221,7 +225,8 @@ class DefaultPartitionedHttpClientBuilder<U, R> extends PartitionedHttpClientBui
 
         @Override
         public StreamingHttpResponseFactory httpResponseFactory() {
-            return new DefaultStreamingHttpResponseFactory(DefaultHttpHeadersFactory.INSTANCE, DEFAULT_ALLOCATOR);
+            return new DefaultStreamingHttpResponseFactory(DefaultHttpHeadersFactory.INSTANCE, DEFAULT_ALLOCATOR,
+                    HTTP_1_1);
         }
 
         @Override
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultPartitionedHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultPartitionedHttpClientBuilder.java
@@ -32,7 +32,6 @@ import io.servicetalk.client.api.partition.UnknownPartitionException;
 import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.concurrent.api.Single;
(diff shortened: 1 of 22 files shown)


Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/1287

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -326,6 +362,37 @@ public StreamingHttpClient buildStreaming() {
         }
     }
 
+    static StreamingHttpRequestResponseFactory defaultReqRespFactory(ReadOnlyHttpClientConfig roConfig,
```
idelpivnitskiy, line 365: Can we reuse this method inside `buildStreaming` for `reqRespFactory ` instead of duplicating the logic?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -278,19 +281,52 @@ public StreamingHttpClient buildStreaming() {
...
+                        ctx.builder.connectionFilterFactory,
+                        new AlpnReqRespFactoryFunc(ctx.executionContext.bufferAllocator(),
+                                h1Config == null ? null : h1Config.headersFactory(),
+                                h2Config == null ? null : h2Config.headersFactory()),
                         ctx.builder.influencerChainBuilder.buildForConnectionFactory(executionStrategy),
                         connectionFactoryFilter, ctx.builder.loadBalancerFactory::toLoadBalancedConnection);
+                // The client can't know which protocol will be negotiated, so just default to the preferred protocol
+                // and pay additional conversion costs if we don't get our preference.
+                String preferredAlpnProtocol = roConfig.tcpConfig().preferredAlpnProtocol();
+                if (H2_ALPN_ID.equals(preferredAlpnProtocol)) {
```
idelpivnitskiy, line 308: Can you please add a test when a client has h2 as `preferredAlpnProtocol` but the server can negotiate only h1? Worry that h2 headers may not fly with h1 encoder.

idelpivnitskiy (approved):
Great improvement! Few comments then LGTM:

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultPartitionedHttpClientBuilder.java
```diff
@@ -221,7 +225,8 @@ public HttpExecutionContext executionContext() {
 
         @Override
         public StreamingHttpResponseFactory httpResponseFactory() {
-            return new DefaultStreamingHttpResponseFactory(DefaultHttpHeadersFactory.INSTANCE, DEFAULT_ALLOCATOR);
+            return new DefaultStreamingHttpResponseFactory(DefaultHttpHeadersFactory.INSTANCE, DEFAULT_ALLOCATOR,
```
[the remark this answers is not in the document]
Scottmitch: Humm good point this seems reasonable... let me try this.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
idelpivnitskiy: Actually, it should be already covered by `AlpnClientAndServerTest`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -278,19 +281,52 @@ public StreamingHttpClient buildStreaming() {
...
+                    if (h2Config == null) {
+                        throw new IllegalStateException("h2 is preferred alpn protocol, but the h2 config is null");
+                    }
+                    reqRespFactory = new DefaultStreamingHttpRequestResponseFactory(
+                            ctx.executionContext.bufferAllocator(), h2Config.headersFactory(), HTTP_2_0);
+                } else { // just fall back to h1, all other protocols should convert to/from it.
+                    if (h1Config == null) {
+                        throw new IllegalStateException(preferredAlpnProtocol +
+                                " is preferred alpn protocol, but the h1 config is null");
```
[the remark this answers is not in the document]
Scottmitch: reworded. I left `preferredAlpnProtocol` in the statement because this should help indicate what was requested vs what we support.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -278,19 +281,52 @@ public StreamingHttpClient buildStreaming() {
...
+                                h1Config == null ? null : h1Config.headersFactory(),
+                                h2Config == null ? null : h2Config.headersFactory()),
                         ctx.builder.influencerChainBuilder.buildForConnectionFactory(executionStrategy),
                         connectionFactoryFilter, ctx.builder.loadBalancerFactory::toLoadBalancedConnection);
+                // The client can't know which protocol will be negotiated, so just default to the preferred protocol
+                // and pay additional conversion costs if we don't get our preference.
+                String preferredAlpnProtocol = roConfig.tcpConfig().preferredAlpnProtocol();
+                if (H2_ALPN_ID.equals(preferredAlpnProtocol)) {
+                    if (h2Config == null) {
+                        throw new IllegalStateException("h2 is preferred alpn protocol, but the h2 config is null");
```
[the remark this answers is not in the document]
Scottmitch: I can change to assert.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
Scottmitch: good point, some duplication of conditionals but shouldn't be a big issue

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
Scottmitch: discussed offline, `AlpnClientAndServerTest` uses `reserveConnection` so your original request is valid and I'll enhance this test to issue requests at the client level too (where we don't know what protocol will be negotiated).

PR #1322, opened 2021-01-14
ServiceTalkTestTimeout remove default min timeout limit

Scottmitch:
Motivation:
During testing/development it is often useful to limit the test value
below the default value. This allows test failures to occur more quickly
when running a sequence of tests, and the developer can then narrow in
on the failing tests and identify patterns. However
ServiceTalkTestTimeout currently enforces a lower bound on the timeout
to the default value (currently 10 seconds per test).

Modifications:
- Remove lower bound in ServiceTalkTestTimeout

Result:
ServiceTalkTestTimeout can set lower timeout values during development.


Scottmitch:
test failure due to https://github.com/apple/servicetalk/issues/1224

Scottmitch:
test failure due to https://github.com/apple/servicetalk/issues/1323

idelpivnitskiy (approved):
🚢

PR #1324, opened 2021-01-14
Publisher#flatMapConcatIterable(..) to not cancel if onNext throws

Scottmitch:
Motivation:
Publisher#flatMapConcatIterable(..) will throw exceptions when
encountered in the context of onNext(..), but it will first cancel. This
may result in cleanup without knowledge of the root cause and even
result in incorrect status to be propagated (e.g. error is recovered and
translated into a response).

Modifications:
- Publisher#flatMapConcatIterable(..) should not call cancel() in the
  event an exception is thrown. It should throw the exception and rely
upon the upstream source to send an error signal via onError(Throwable).

Result:
More informative errors when Publisher#flatMapConcatIterable(..) is
used.


Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1224

Scottmitch:

Scottmitch:
Build failure attributed to https://github.com/apple/servicetalk/issues/999

Scottmitch:
Build failure attributed to https://github.com/apple/servicetalk/issues/1192

idelpivnitskiy (approved):
🎖️

PR #1325, opened 2021-01-15
Publisher#flatMapMerge* preserve error

Scottmitch:
Motivation:
The Publisher#flatMapMerge* operators call cancel() on the upstream
subscription before calling onError(..) on the downstream subscriber in
the event of errors. This may result in cleanup without knowledge of the
root cause and even result in incorrect status to be propagated (e.g.
error is recovered and translated into a response).

Modifications:
- Publisher#flatMapMerge* implementations attempt to propagate
  onError(..) before cancel upstream subscription
- Remove the atomic terminalNotification from Publisher#flatMapMerge*
  implementaitons as it is no longer necessary becuase we keep a fine
  grained count of how many mapped sources are outstanding which lets us
  know exactly when we have terminate.
- Remove CompositeException which adds additional layer of queue and
  transfer to suppressed exception list in THrowable. This also
  throws an exception type which must be unwrapped in order to
  understand the root cause. Change default number of suppressed
  exceptions to 0 by default to avoid potential memory issues if static
  exceptions are thrown.

Result:
More informative errors when Publisher#flatMapMerge*(..) is used.


Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1224

Scottmitch:

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -357,13 +332,8 @@ private void enqueueAndDrain(Object item) {
...
         private void enqueueFailed(Object item) {
-            QueueFullException exception = new QueueFullException("signals");
-            if (item instanceof TerminalNotification) {
-                LOGGER.error("Queue should be unbounded, but an offer failed!", exception);
-                throw exception;
-            } else {
-                onError0(exception, false);
-            }
+            LOGGER.error("Queue should be unbounded, but an offer failed for item {}!", item);
+            throw new QueueFullException("pending");
```
idelpivnitskiy, line 336: We discussed [here](https://github.com/apple/servicetalk/pull/1326#discussion_r560579593) that we can't throw inside terminal methods. But this one may potentially throw (if the queue changes). Consider using try-catch where this method is invoked.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -608,27 +569,32 @@ private void throwInvalidDemand(int pendingDemand) {
...
             @Override
             public void onError(final Throwable t) {
+                Throwable currPendingError = parent.pendingError;
                 if (parent.source.maxDelayedErrors == 0) {
-                    // Make sure errors aren't delivered out of order relative to onNext signals which maybe queued.
-                    try {
-                        parent.doCancel(true, false);
-                    } finally {
-                        parent.tryEmitItem(TerminalNotification.error(t), this);
+                    if (currPendingError == null && pendingErrorUpdater.compareAndSet(parent, null, t)) {
```
idelpivnitskiy, line 574: Inside the above `if` branch (line 573) `currPendingError` is never used. Consider simplifying this one to
```
if (pendingErrorUpdater.compareAndSet(parent, null, t)) {
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -194,53 +197,30 @@ public void onSubscribe(final Subscription s) {
...
 
         @Override
         public void onError(final Throwable t) {
-            if (!onError0(t, true)) {
-                LOGGER.debug("Already terminated/cancelled, ignoring error notification.", t);
+            try {
+                upstreamError = t;
+                enqueueAndDrain(error(t));
+            } finally {
+                doCancel(false, true);
```
idelpivnitskiy, line 212: We should cancel before draining signals. Consider adding a comment explaining why "cancel-before-onError" is important in cases when we use this pattern:
```java
try {
    // cancel
} finally {
    // onError
}
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
```diff
@@ -320,14 +300,9 @@ private void doDrainPostProcessing(final long drainCount) {
...
-            QueueFullException exception = new QueueFullException("pending");
-            if (item instanceof TerminalNotification) {
-                LOGGER.error("Queue should be unbounded, but an offer failed!", exception);
-                throw exception;
-            } else {
-                onError0(exception, true, true);
-            }
+        private void enqueueFailed(Object item) {
+            LOGGER.error("Queue should be unbounded, but an offer failed for item {}!", item);
+            throw new QueueFullException("pending");
```
idelpivnitskiy, line 305: We discussed [here](https://github.com/apple/servicetalk/pull/1326#discussion_r560579593) that we can't throw inside terminal methods. But this one may potentially throw (if the queue changes). Consider using try-catch where this method is invoked.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
```diff
@@ -399,27 +364,28 @@ public void onSuccess(@Nullable R result) {
 
             @Override
             public void onError(Throwable t) {
+                Throwable currPendingError = pendingError;
                 if (source.maxDelayedErrors == 0) {
-                    onError0(t, true, true);
+                    if (currPendingError == null &&
+                            pendingErrorUpdater.compareAndSet(FlatMapSubscriber.this, null, t)) {
```
idelpivnitskiy, line 370: Inside the above `if` branch (line 368) `currPendingError` is never used. Consider simplifying this one to
```
if (pendingErrorUpdater.compareAndSet(FlatMapSubscriber.this, null, t)) {
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
```diff
@@ -399,27 +364,28 @@ public void onSuccess(@Nullable R result) {
...
+                        onError0(t, true);
```
idelpivnitskiy, line 371: Looking at `FlatMapPublisherSubscriber#onError`, it uses `tryEmitItem` in this case instead of `enqueueAndDrain`. Should we align them because the rest of the logic is similar?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: If we fail to enqueue a terminal event the async control flow won't be completed and the user won't be notified. This is a relatively extreme failure condition where I think it is justified to fail loudly as we are already violating the spec in more severe ways (e.g. not delivering all signals) and the behavior at this point is undefined (e.g. likely a hang). Note we were previously throwing in this scenario as well.

Scottmitch: I'll add a comment

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: if the current thread sees a non-null we can avoid an atomic operation.

idelpivnitskiy (approved):
LGTM after the last 2 unresolved discussions above ☝️ 
Thank you!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
Scottmitch: in this case we aren't cancelling the upstream subscription because we have been terminated by upstream. However I'll make it consistent with flatMapMergeSingle.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
Scottmitch: `tryEmitItem` from the publisher variant attempts to emit and skip the queue if possible (preserves ordering of events relative to that mapped publisher), and enqueues the error if this isn't possible (e.g. lack of demand or elements for that mapped publisher are already queued). since the publisher variant we request demand from each mapped publisher optimistically (to ensure we don't deadlock) this approach avoids queueing events which may require more demand before delivering any errors.

idelpivnitskiy: Thanks for clarifying

PR #1330, opened 2021-01-20
OnSubscribeIgnoringSubscriberForOffloading empty subscription usage

Scottmitch:
Motivation:
93165f04398085600a628e5af9d5de625b37aa2f made a change to ensure invalid
requestN demand is propagated to the associated Subscriber. However
OnSubscribeIgnoringSubscriberForOffloading#offloadWithDummyOnSubscribe
is used in such a way that a terminal event will already be delivered
and demand is validated externally. This usage is so the offloader has
the expected onSubscribe method invoked, but the Subscription will not
actually be used. The usage can therefore be simplified to use
EMPTY_SUBSCRIPTION and avoid potential for duplicate terminal delivery
to the Subscriber.

Modifications:
- OnSubscribeIgnoringSubscriberForOffloading#offloadWithDummyOnSubscribe
  to use EMPTY_SUBSCRIPTION.

Result:
Avoid potential for duplicate terminal event delivery in
OnSubscribeIgnoringSubscriberForOffloading.


idelpivnitskiy (approved):
🚀

PR #1332, opened 2021-01-21
HTTP Trailers Propagation Fix

Scottmitch:
Motivation:
The HTTP protocol represents data as a single flat stream of data of the
following format {MetaData+Headers, Data*, Trailers*}. Trailers may or
may not be present depending when reading from the transport, and users
don't always add trailers. Because of this we attempt to avoid including
trailers in the stream unless we read them from the transport or the
user explicitly does a transformation to add trailers. This helps reduce
the cost/complexity of intermediate transformations which may need to
expose the Buffers and still preserve the trailers. However this
conditional logic doesn't reliably work on the client side because the
control flow merges the read and write stream and we only concatenate
the trailers after the response stream completes successfully, which
maybe too late if the write stream never completes (or only completes in
error when the transport is closed). The end result is trailers may be
dropped even when they are explicitly added (e.g. in gRPC).

Modifications:
- Rework StreamingHttpPayloadHolder to use a merge operation so the
  trailers are processed inline with the flat stream, and don't depend
  upon the stream completion.
- StreamingHttpPayloadHolder to avoid storing state outside the
  Publisher transformations, and instead state should be coupled with
  each subscribe operation to be more ReactiveStreams compliant and
  supporting use cases like re-subscribe.
- Deprecate "payloadAndTrailers()" and "raw" transformation methods. The
  raw payload type was intended to be used for FileRegion but this is
  work is currently unplanned and memory mapped files as Buffers is the
  current preferred alternative (which also works with TLS is enabled).
  These methods will be removed in a future commit and this will also
  simplify the internal implementation of StreamingHttpPayloadHolder and
  other areas that deal with the underlying payload.

Result:
Trailers are preserved on the client and server side regardless of the
payload stream completion status.


Scottmitch:
Build failures expected to be resolved when https://github.com/apple/servicetalk/pull/1336 is merged and this PR is rebased.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouteConversions.java
```diff
@@ -143,15 +138,16 @@ public void flush() throws IOException {
...
+
+                            // The user code has returned successfully, complete the processor so the response stream
+                            // can complete. If the user handles the request asynchronously (e.g. on another thread)
+                            // they are responsible for closing the payloadWriter.
+                            exceptionProcessor.onComplete();
                         } catch (Throwable t) {
-                            concurrentTerminalSubscriber.onError(t);
-                        } finally {
                             try {
-                                grpcPayloadWriter.close();
```
Scottmitch, line 150: @idelpivnitskiy - note this change. we don't auto-close in `BlockingStreamingToStreamingService` so I think we should be consistent here. we can consider adding an option to auto-close for both http & grpc, but lets do this as a followup PR if necessary.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -88,8 +86,9 @@ protected void handleSubscribe(final Subscriber<? super StreamingHttpResponse> s
                             // Content-Length header field can provide the anticipated size.
                             // https://tools.ietf.org/html/rfc7230#section-3.3.2
                             HttpHeaders headers = metaData.headers();
-                            boolean addTrailers = isTransferEncodingChunked(headers);
-                            if (!addTrailers && !HTTP_1_0.equals(metaData.version()) && !hasContentLength(headers) &&
+                            boolean addTrailers = metaData.version().major() > 1 || isTransferEncodingChunked(headers);
+                            if (!addTrailers && metaData.version().major() == 1 && metaData.version().minor() > 0 &&
+                                    !hasContentLength(headers) &&
```
idelpivnitskiy, line 91: After #1311 we will be able to share `netty/HeaderUtils` and reuse these utilities

servicetalk-http-api/src/test/java/io/servicetalk/http/api/BlockingStreamingToStreamingServiceTest.java
```diff
@@ -439,10 +439,12 @@ public void throwAfterPayloadWriterClosed() throws Exception {
...
 
-        List<Object> response = invokeService(syncService, reqRespFactory.get("/"));
-        assertMetaData(OK, response);
-        assertPayloadBody("", response);
-        assertEmptyTrailers(response);
+        try {
+            invokeService(syncService, reqRespFactory.get("/"));
+            fail("Payload body should complete with an error");
+        } catch (ExecutionException e) {
+            assertThat(e.getCause(), is(DELIBERATE_EXCEPTION));
```
idelpivnitskiy, line 446: Consider using `assertThrows`

servicetalk-oio-api-internal/src/main/java/io/servicetalk/oio/api/internal/PayloadWriterUtils.java
```diff
@@ -0,0 +1,44 @@
...
+    private PayloadWriterUtils() {
+    }
+
+    /**
+     * Invokes {@link PayloadWriter#close(Throwable)} ignoring an occurred exception if any.
+     * @param writer The {@link PayloadWriter} to close.
+     * @param cause The cause to pass to {@link PayloadWriter#close(Throwable)}.
+     * @param <T> The type of {@link PayloadWriter}.
+     */
+    public static <T> void safeClose(PayloadWriter<T> writer, Throwable cause) {
```
idelpivnitskiy, line 37: What do you think about making it public in `servicetalk-oio-api`? Looks like useful utility for users as well, and will help us to avoid creating a new module.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouteConversions.java
```diff
@@ -319,10 +315,11 @@ protected void handleSubscribe(final CompletableSource.Subscriber subscriber) {
                     } else {
                         original.close();
                     }
-                    subscriber.onComplete();
                 } catch (Throwable t) {
                     subscriber.onError(t);
+                    return;
                 }
+                subscriber.onComplete();
```
idelpivnitskiy, line 322: 🎖️

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -666,7 +666,8 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
                 // ChannelInputShutdownEvent is not always triggered and can get triggered before we tried to read
                 // all the available data. ChannelInputShutdownReadComplete is the one that seems to (at least in
                 // the current netty version) gets triggered reliably at the appropriate time.
-                connection.nettyChannelPublisher.channelInboundClosed();
+                connection.nettyChannelPublisher.channelInboundClosed(StacklessClosedChannelException.newInstance(
+                        DefaultNettyConnection.class, "userEventTriggered(ChannelInputShutdownReadComplete)"));
```
idelpivnitskiy, line 670: Thanks!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
```diff
@@ -171,7 +201,8 @@ private static void updateResponseContentLength(final int contentLength, final H
                 List<Object> itemsUnchecked = (List<Object>) reduction;
                 items = itemsUnchecked;
             } else {
-                items = new ArrayList<>();
+                // this method is called if the payload has been aggregated, we expect {meta, buffer, trailers}.
+                items = new ArrayList<>(3);
```
idelpivnitskiy, line 205: We do not add metadata to this list, should the size be equal to 2?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2PriorKnowledgeFeatureParityTest.java
```diff
@@ -849,6 +850,31 @@ public void serverThrowsFromHandler() throws Exception {
         }
     }
 
+    @Test
+    public void trailersWithContentLength() throws Exception {
```
idelpivnitskiy, line 854: ❤️

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpRequest.java
```diff
@@ -140,10 +168,12 @@
     /**
      * Returns a {@link StreamingHttpRequest} with its underlying payload transformed to {@link Object}s,
      * with access to the trailers.
+     * @deprecated use {@link #transform(TrailersTransformer)}.
      * @param trailersTransformer {@link TrailersTransformer} to use for this transform.
      * @param <T> The type of state used during the transformation.
      * @return {@code this}
      */
+    @Deprecated
     <T> StreamingHttpRequest transformRaw(TrailersTransformer<T, Object> trailersTransformer);
```
idelpivnitskiy, line 177: This method is still used in `ContentHeadersTest`, please update the test.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpResponse.java
```diff
@@ -46,11 +46,24 @@
 
     /**
      * Gets a {@link Publisher} that combines the raw payload body concatenated with the {@link HttpHeaders trailers}.
+     * @deprecated Use {@link #messageBody()}.
      * @return a {@link Publisher} that combines the raw payload body concatenated with the
      * {@link HttpHeaders trailers}.
      */
+    @Deprecated
     Publisher<Object> payloadBodyAndTrailers();
```
idelpivnitskiy, line 54: This method is still used in `StreamingHttpServiceToBlockingStreamingHttpService#handleBlockingRequest`, please update it to use `messageBody()`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpResponse.java
```diff
@@ -122,11 +135,26 @@
     /**
      * Returns a {@link StreamingHttpResponse} with its underlying payload transformed. Note that the raw objects of the
      * underlying {@link Publisher} may be exposed. The object types are not guaranteed to be homogeneous.
+     * @deprecated Use {@link #transformPayloadBody(UnaryOperator)}.
      * @param transformer Responsible for transforming the payload body.
      * @return {@code this}
      */
+    @Deprecated
     StreamingHttpResponse transformRawPayloadBody(UnaryOperator<Publisher<?>> transformer);
```
idelpivnitskiy, line 143: This method is still used in `ContentLengthTest`, please update the test. Also, referenced in javadoc of both tracing filters.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpResponse.java
```diff
@@ -139,10 +167,12 @@
     /**
      * Returns a {@link StreamingHttpResponse} with its underlying payload transformed to {@link Object}s,
      * with access to the trailers.
+     * @deprecated use {@link #transform(TrailersTransformer)}.
      * @param trailersTransformer {@link TrailersTransformer} to use for this transform.
      * @param <T> The type of state used during the transformation.
      * @return {@code this}
      */
+    @Deprecated
     <T> StreamingHttpResponse transformRaw(TrailersTransformer<T, Object> trailersTransformer);
```
idelpivnitskiy, line 176: This method is still used in `ContentHeadersTest`, please update the test.

idelpivnitskiy (commented):
Awesome work! Thanks for making it in a good shape ❤️

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultPayloadInfo.java
```diff
@@ -57,8 +73,8 @@ DefaultPayloadInfo setSafeToAggregate(boolean safeToAggregate) {
         return set(SAFE_TO_AGGREGATE, safeToAggregate);
     }
 
-    DefaultPayloadInfo setMayHaveTrailers(boolean mayHaveTrailers) {
-        return set(MAY_HAVE_TRAILERS, mayHaveTrailers);
+    DefaultPayloadInfo setMayHaveTrailers(boolean trailersPresent) {
+        return set(MAY_HAVE_TRAILERS, trailersPresent);
```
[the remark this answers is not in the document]
Scottmitch: +1! this was from an intermediate change that was abandoned :)

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
```diff
@@ -17,127 +17,210 @@
...
+            final Publisher<?> oldMessageBody = messageBody;
+            messageBody = defer(() -> {
+                Processor<HttpHeaders, HttpHeaders> trailersProcessor = newSingleProcessor();
+                return merge(payloadBody.liftSync(new BridgeFlowControlAndDiscardOperator(
+                                oldMessageBody.liftSync(new PreserveTrailersBufferOperator(trailersProcessor)))),
+                        fromSource(trailersProcessor)).subscribeShareContext();
+            });
+        } else {
+            messageBody = payloadBody.liftSync(new BridgeFlowControlAndDiscardOperator(messageBody));
+            payloadInfo.setOnlyEmitsBuffer(true);
```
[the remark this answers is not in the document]
Scottmitch: If the publisher has trailers it doesn't always emit buffers (e.g. its type isn't `Publisher<Buffer>`). Note that I plan on removing `onlyEmitsBuffer()` (and related `raw` methods) in a followup commit which should simplify the logic in this class.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
```diff
@@ -17,127 +17,210 @@
...
+    void payloadBody(final Publisher<Buffer> payloadBody) {
+        if (messageBody == null) {
+            messageBody = requireNonNull(payloadBody);
+            payloadInfo.setOnlyEmitsBuffer(true);
+        } else if (payloadInfo.mayHaveTrailers()) {
+            final Publisher<?> oldMessageBody = messageBody;
+            messageBody = defer(() -> {
+                Processor<HttpHeaders, HttpHeaders> trailersProcessor = newSingleProcessor();
+                return merge(payloadBody.liftSync(new BridgeFlowControlAndDiscardOperator(
+                                oldMessageBody.liftSync(new PreserveTrailersBufferOperator(trailersProcessor)))),
```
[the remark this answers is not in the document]
Scottmitch: Right we never really supported non-buffer types in our impls and the `PreserveTrailersBufferOperator` supports buffer/trailers. This class is the equivalent of `HttpDataSourceTransformations$HttpBufferTrailersSpliceOperator` (which was removed in this PR).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
```diff
@@ -17,127 +17,210 @@
...
 
-    public void transformPayloadBody(UnaryOperator<Publisher<Buffer>> transformer) {
-        updatePayloadBody(transformer.apply(payloadBody()), true);
+    void transformPayloadBody(UnaryOperator<Publisher<Buffer>> transformer) {
+        if (messageBody != null && payloadInfo.mayHaveTrailers()) {
+            final Publisher<?> oldMessageBody = messageBody;
+            messageBody = defer(() -> {
+                Processor<HttpHeaders, HttpHeaders> trailersProcessor = newSingleProcessor();
+                return merge(transformer.apply(oldMessageBody.liftSync(
+                        new PreserveTrailersBufferOperator(trailersProcessor))), fromSource(trailersProcessor))
```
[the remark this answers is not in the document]
Scottmitch: I think this is a duplicate of https://github.com/apple/servicetalk/pull/1332#discussion_r565695510 above? lets discuss there.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
```diff
@@ -17,127 +17,210 @@
...
 
     Single<PayloadAndTrailers> aggregate() {
         payloadInfo.setSafeToAggregate(true);
-        return aggregatePayloadAndTrailers(payloadBodyAndTrailers(), allocator);
+        return aggregatePayloadAndTrailers(messageBody(), allocator);
+    }
+
+    private <T, P> void transformWithTrailersUnchecked(final TrailersTransformer<T, P> trailersTransformer,
+                                                       final Function<Object, P> caster) {
+        if (majorHttpVersion == 1) {
```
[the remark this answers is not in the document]
Scottmitch: let me update, the goal was to minimize "protocol conditionals" outside our encoder/decoder as much as possible.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
```diff
@@ -198,209 +277,153 @@ public int hashCode() {
...
-                            subscriber.onError(t);
-                            return;
-                        }
-                    }
-                    subscriber.onSuccess(result);
-                }
+    /**
+     * Merge a {@link Publisher} and a {@link Single} into a single {@link Publisher}.
+     * <p>
+     * The transport for HTTP client connections does {@code connection.write(requestStream).merge(connection.read())}
```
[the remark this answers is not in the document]
Scottmitch: I can update, but for the purpose of this comment the impacts are the same.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -163,6 +163,15 @@ public final HttpServerBuilder disableDrainingRequestPayloadBody() {
...
     }
 
+    /**
+     * Set if the <a href="https://tools.ietf.org/html/rfc7230#section-4.4">Trailer</a> header is required to process
+     * trailers.
+     * @param requireTrailerHeader {@code true} if <a href="https://tools.ietf.org/html/rfc7230#section-4.4">Trailer</a>
+     * header is required to accept trailers. {@code false} assumes trailers may be present if other criteria allows.
+     * @return {@code this}
+     */
+    public abstract HttpServerBuilder requireTrailerHeader(boolean requireTrailerHeader);
```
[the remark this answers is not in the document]
Scottmitch: The default value respects `Trailer` header as optional. This option offers an optimization to avoid more costly intermediate trailers preservation operations if they aren't explicitly specified as being used (basically OK to drop trailers if this is true and the request/response doesn't indicate trailers are present).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
Scottmitch: it would be nice to consolidate these trailers checks.

servicetalk-oio-api-internal/src/main/java/io/servicetalk/oio/api/internal/PayloadWriterUtils.java
Scottmitch: I prefer to start with it in internal as it isn't strictly required for the API, and this gives us freedom to move it later if required.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpRequest.java
Scottmitch: My thought process here -> while the deprecated method still exists if there are tests dedicated to the deprecated methods we should leave them, so the behavior is still verified. I plan to remove all usage when the deprecated methods are removed.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpResponse.java
Scottmitch: Lets discuss on https://github.com/apple/servicetalk/pull/1332#discussion_r565782546 due to overlapping discussion.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpResponse.java
Scottmitch: Lets discuss on https://github.com/apple/servicetalk/pull/1332#discussion_r565782546 due to overlapping discussion.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
```diff
@@ -17,127 +17,210 @@
...
 
-    public void payloadBody(final Publisher<Buffer> payloadBody) {
-        updatePayloadBody(payloadBody, false);
+    void payloadBody(final Publisher<Buffer> payloadBody) {
+        if (messageBody == null) {
+            messageBody = requireNonNull(payloadBody);
+            payloadInfo.setOnlyEmitsBuffer(true);
+        } else if (payloadInfo.mayHaveTrailers()) {
+            final Publisher<?> oldMessageBody = messageBody;
+            messageBody = defer(() -> {
```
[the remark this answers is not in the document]
Scottmitch: I'll add a unit test, I agree we only need to drain 1 time. FWIW even as is the defer operation isn't expected to fail, and it shouldn't impact the consumption of the new `messageBody`, but instead the `subscriber` will be terminated with an error (which is OK bcz we just ignore it).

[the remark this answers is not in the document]
idelpivnitskiy: oh, right, `BridgeFlowControlAndDiscardOperator` ignores errors. sounds good

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
```diff
@@ -17,127 +17,211 @@
...
 
     Single<PayloadAndTrailers> aggregate() {
         payloadInfo.setSafeToAggregate(true);
-        return aggregatePayloadAndTrailers(payloadBodyAndTrailers(), allocator);
+        return aggregatePayloadAndTrailers(messageBody(), allocator);
+    }
+
+    private <T, P> void transformWithTrailersUnchecked(final TrailersTransformer<T, P> trailersTransformer,
+                                                       final Function<Object, P> caster) {
+        if (version == HTTP_1_1) {
```
idelpivnitskiy, line 180: Consider using a check similar to what you have in `BlockingStreamingToStreamingService`:
```
metaData.version().major() == 1 && metaData.version().minor() > 0
```

Every time I see this, thinking about implementing `Comparable` for `HttpProtocolVersion` :)

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpRequest.java
idelpivnitskiy: Makes sense, thanks for clarifying!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpResponse.java
idelpivnitskiy: Makes sense for tests, let's just update javadoc in `TracingHttpRequesterFilter` and `TracingHttpServiceFilter`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
Scottmitch: good point. I will update `BlockingStreamingToStreamingService` to use `== HTTP_1_1 ` which is simpler and 1.2+ isn't a thing.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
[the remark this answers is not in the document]
idelpivnitskiy: The most frequent use-case for trailers is grpc, which doesn't use `Trailer` header. HTTP/2 supports it but doesn't suggest or require using it if the trailing `HEADERS` frame is expected.

Other use-cases for trailers are rare and it's not always the case that users use `Trailer` header. I'm just not sure if anyone will find this method useful.
If the performance is the main concern, we can try to run the benchmarks with and without this flag to see the impact. If it's negligible for the use-case when there are no trailers (most of HTTP/1.1 users with streaming API), I would propose to defer adding it to the public API (remove or make pkg-private).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -87,8 +88,7 @@ protected void handleSubscribe(final Subscriber<? super StreamingHttpResponse> s
                             // https://tools.ietf.org/html/rfc7230#section-3.3.2
                             HttpHeaders headers = metaData.headers();
                             boolean addTrailers = metaData.version().major() > 1 || isTransferEncodingChunked(headers);
-                            if (!addTrailers && metaData.version().major() == 1 && metaData.version().minor() > 0 &&
-                                    !hasContentLength(headers) &&
+                            if (!addTrailers && metaData.version() == HTTP_1_1 && !hasContentLength(headers) &&
```
idelpivnitskiy, line 91: I expected the opposite change :)
It seems more reliable to check for `metaData.version().major() == 1 && metaData.version().minor() > 0` in case HTTP/1.2 will ever happen. All minor versions have to be backward compatible with previous versions.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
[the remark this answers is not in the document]
idelpivnitskiy: Discussed offline, this is only for receiving side (`newTransport[Request|Response]`). Let's give it a try and perf-test with gRPC (server can use it because [gRPC requests](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests) never have trailers).

Agreed with the offline proposal to clarify that in javadoc.

Also, consider not using `boolean` param because we don't need to flip it back and forth. The proposed `allowIgnoreTrailersIfNoHeader()` or just `ignoreTrailersIfNoHeader()` sgtm.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
idelpivnitskiy: Ok, both ways are fine. HTTP/1.2 is not official, I was just thinking that someone can define it (we allow custom `HttpProtocolVersion`) with some custom subset of headers or something like that.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpResponse.java
idelpivnitskiy: Just a reminder about this thread

idelpivnitskiy (approved):
Thank you!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpResponse.java
Scottmitch: +1 good catch!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
Scottmitch: I think `== HTTP_1_1` is fine in practice, but I'll switch back to the other conditional as it is more general. I'll also move this into a common utility method so it will be easier to detect where this clause is checked in the future.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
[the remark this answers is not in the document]
Scottmitch: I updated the method name and clarified the javadocs.

Scottmitch:
build failure due to https://github.com/apple/servicetalk/issues/1192

Issue #1216, opened 2020-11-17
Improve performance of CharSequences split


tkountis:
Improve split to return trimmed CharSequence values instead of doing a toString() conversion and an additional trim here that makes 2 copies of data when a space char is present.

PR #1335, opened 2021-01-26
Improve CharSequences#split to support trim and min copies

tkountis:
**Motivation**:

CharSequences#split needed some improvements to avoid extra allocations for trimming (externally) and conversion to String due to calls to subsequence.

**Modifications**:

This PR improves the helper to support trimming individual tokens without additional allocations, and also better utilizes AsciiBuffer internals to avoid unnecessary conversions.

**Result**:

API can now handle representations that contain whitespaces in between delimiters.

Fixes: https://github.com/apple/servicetalk/issues/1216

### DIFF
--- a/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
+++ b/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
@@ -239,12 +239,10 @@ public final class CharSequences {
             }
         }
 
-        if (startIndex != -1) {
-            if ((input.length() - startIndex) > 0) {
-                result.add(subsequence(isAscii, input, startIndex, input.length()));
-            } else {
-                result.add(isAscii ? newAsciiString("") : "");
-            }
+        if ((input.length() - startIndex) > 0) {
+            result.add(subsequence(isAscii, input, startIndex, input.length()));
+        } else {
+            result.add(isAscii ? EMPTY_ASCII_BUFFER : "");
         }
 
         return result;
@@ -271,7 +269,7 @@ public final class CharSequences {
                 if (endIndex > startIndex) {
                     result.add(subsequence(isAscii, input, startIndex, endIndex));
                 } else {
-                    result.add(isAscii ? newAsciiString("") : "");
+                    result.add(isAscii ? EMPTY_ASCII_BUFFER : "");
                 }
 
                 startIndex = i + 1;
@@ -284,7 +282,7 @@ public final class CharSequences {
             if ((input.length() - startIndex) > 0) {
                 result.add(subsequence(isAscii, input, startIndex, endIndex));
             } else {
-                result.add(isAscii ? newAsciiString("") : "");
+                result.add(isAscii ? EMPTY_ASCII_BUFFER : "");
             }
         }
 
--- a/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
+++ b/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
@@ -206,37 +206,67 @@ public final class CharSequences {
     }
 
     /**
-     * Split a given {@link CharSequence} to separate ones on the given {@code delimiter}.
-     * The returned {@link CharSequence}s are created by invoking the {@link CharSequence#subSequence(int, int)} method
-     * on the main one.
+     * Split a given {@link #newAsciiString(Buffer) AsciiString} to separate ones on the given {@code delimiter}.
      *
+     * Trimming white-space before and after each token can be controlled by the {@code trim} flag
      * This method has no support for regex.
      *
-     * @param input The initial {@link CharSequence} to split, this experiences no side effects
-     * @param delimiter The delimiter character
+     * @param input The initial {@link CharSequence} to split, this experiences no side effects.
+     * @param delimiter The delimiter character.
+     * @param trim Flag to control whether the individual items must be trimmed.
      * @return a {@link List} of {@link CharSequence} subsequences of the input with the separated values
      */
-    public static List<CharSequence> split(final CharSequence input, final char delimiter) {
+    public static List<CharSequence> split(final CharSequence input, final char delimiter, final boolean trim) {
         if (input.length() == 0) {
             return emptyList();
         }
 
-        int startIndex = 0;
         List<CharSequence> result = new ArrayList<>();
+
+        int startIndex = trim ? -1 : 0;
+        int endIndex = -1;
+
         for (int i = 0; i < input.length(); i++) {
-            if (input.charAt(i) == delimiter) {
-                if ((i - startIndex) > 0) {
-                    result.add(input.subSequence(startIndex, i));
+            char c = input.charAt(i);
+            if (!trim) {
+                endIndex = i;
+            } else if (c != ' ' && c != delimiter) {
+                endIndex = i + 1;
+            }
+
+            if (startIndex == -1 && c != ' ' && c != delimiter) {
+                startIndex = i;
+            } else if (c == delimiter) {
+                if (startIndex != -1 && endIndex != -1 && (i - startIndex) > 0) {
+                    result.add(subsequence(input, startIndex, endIndex));
                 }
 
-                startIndex = i + 1;
+                if (!trim) {
+                    startIndex = i + 1;
+                } else {
+                    startIndex = -1;
+                    endIndex = -1;
+                }
             }
         }
 
-        if ((input.length() - startIndex) > 0) {
-            result.add(input.subSequence(startIndex, input.length()));
+        if (!trim) {
+            endIndex = input.length();
         }
+
+        if (startIndex != -1 && endIndex != -1 && (input.length() - startIndex) > 0) {
+            result.add(subsequence(input, startIndex, endIndex));
+        }
+
         return result;
+    }
+
+    private static CharSequence subsequence(final CharSequence input, final int start, final int end) {
+        if (isAsciiString(input)) {
+            return newAsciiString(((AsciiBuffer) input).unwrap().copy(start, end - start));
+        } else {
+            return input.subSequence(start, end);
+        }
     }
 
     private static boolean equalsIgnoreCase(final char a, final char b) {
--- a/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
+++ b/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
@@ -221,48 +221,60 @@ public final class CharSequences {
             return emptyList();
         }
 
-        List<CharSequence> result = new ArrayList<>();
+        boolean isAscii = isAsciiString(input);
+        List<CharSequence> result = new ArrayList<>(4);
 
-        int startIndex = trim ? -1 : 0;
+        int startIndex = 0;
         int endIndex = -1;
 
-        for (int i = 0; i < input.length(); i++) {
-            char c = input.charAt(i);
-            if (!trim) {
-                endIndex = i;
-            } else if (c != ' ' && c != delimiter) {
-                endIndex = i + 1;
-            }
-
-            if (startIndex == -1 && c != ' ' && c != delimiter) {
-                startIndex = i;
-            } else if (c == delimiter) {
-                if (startIndex != -1 && endIndex != -1 && (i - startIndex) > 0) {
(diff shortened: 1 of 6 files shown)


Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/1224

Scottmitch:
another test failure attributed to #1224

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
```diff
@@ -206,39 +206,69 @@ public static int caseInsensitiveHashCode(CharSequence seq) {
...
+        int endIndex = -1;
+
         for (int i = 0; i < input.length(); i++) {
-            if (input.charAt(i) == delimiter) {
-                if ((i - startIndex) > 0) {
-                    result.add(input.subSequence(startIndex, i));
+            char c = input.charAt(i);
+            if (!trim) {
+                endIndex = i;
+            } else if (c != ' ' && c != delimiter) {
```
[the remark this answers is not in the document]
tkountis: Added tests to show behavior.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcMessageEncodingTest.java
```diff
@@ -227,89 +224,91 @@ public StreamingHttpServiceFilter create(final StreamingHttpService service) {
...
+                                    .collect(toList());
+
+                    if (!expectedRespAcceptedEncodings.isEmpty() && !actualRespAcceptedEncodings.isEmpty()) {
+                        assertEquals(expectedRespAcceptedEncodings, actualRespAcceptedEncodings);
+                    }
 
-                            if (!expectedRespAcceptedEncodings.isEmpty() && !actualRespAcceptedEncodings.isEmpty()) {
-                                assertEquals(expectedRespAcceptedEncodings, actualRespAcceptedEncodings);
+                    final String respEncName = response.headers()
+                            .get(MESSAGE_ENCODING, "identity").toString();
```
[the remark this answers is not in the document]
tkountis: Will follow up on that with the related change on the gRPC headers.

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
```diff
@@ -206,39 +206,81 @@ public static int caseInsensitiveHashCode(CharSequence seq) {
...
         int startIndex = 0;
-        List<CharSequence> result = new ArrayList<>();
-        for (int i = 0; i < input.length(); i++) {
-            if (input.charAt(i) == delimiter) {
-                if ((i - startIndex) > 0) {
-                    result.add(input.subSequence(startIndex, i));
+        int endIndex = -1;
+
+        if (trim) {
+            startIndex = -1;
```
Scottmitch, line 231: consider putting each variant in its own method. which has the following benefits:
- reduces scope so that only the necessary state is defined
- easier inlining for JVM -> reduces the overall byte code of the containing method and makes it easier for the JVM to inline this method if the `trim` value is mostly the same.

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
```diff
@@ -206,39 +206,81 @@ public static int caseInsensitiveHashCode(CharSequence seq) {
...
+                    }
+
+                    startIndex = -1;
+                    endIndex = -1;
+                }
+            }
+        } else {
+            for (int i = 0; i < input.length(); i++) {
+                char c = input.charAt(i);
+                endIndex = i;
```
Scottmitch, line 254: I think we have an issue if there are consecutive delimiters. Here is a test case to add:

```java
assertThat(split(",,,", ',', false), empty());
```

IIUC we should only need a the following state:
- `i` -> current index
- `startIndex` -> last non-delimiter `char` (reset after each subsequence)

Scottmitch: I think the same issue applies for the `trim` path too:
```java
assertThat(split(",,,", ',', true), empty());
```

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
```diff
@@ -206,39 +206,81 @@ public static int caseInsensitiveHashCode(CharSequence seq) {
...
                 }
 
-                startIndex = i + 1;
+                if (startIndex == -1 && c != ' ' && c != delimiter) {
+                    startIndex = i;
+                } else if (c == delimiter) {
+                    if (endIndex > startIndex) {
+                        result.add(subsequence(isAscii, input, startIndex, endIndex));
+                    } else {
+                        result.add(isAscii ? newAsciiString("") : "");
```
Scottmitch, line 244: Can you comment if the following unit test is correct? If so this code needs to be adjusted.
```java
// there are no non-delimiter chars present -> content should be empty
assertThat(split("  ", ' ', true), empty());
```

we may need more state (e.g. differentiate between last index of non-space and delimiter)?

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
```diff
@@ -206,39 +206,81 @@ public static int caseInsensitiveHashCode(CharSequence seq) {
...
+        if (startIndex != -1 && endIndex != -1 && (input.length() - startIndex) > 0) {
+            result.add(subsequence(isAscii, input, startIndex, endIndex));
         }
+
         return result;
     }
 
+    private static CharSequence subsequence(final boolean isAscii, final CharSequence input,
+                                            final int start, final int end) {
+        if (isAscii) {
```
Scottmitch, line 277: nit: consider using a tertiary statement

servicetalk-buffer-api/src/test/java/io/servicetalk/buffer/api/CharSequencesTest.java
```diff
@@ -0,0 +1,164 @@
...
+                contains("", "gzip", "", "", "", "", "compress"));
+        assertThat(split("gzip,,,,,compress", ',', true),
+                contains("gzip", "", "", "", "", "compress"));
+        assertThat(split(",gzip,,,,,compress,", ',', true),
+                contains("", "gzip", "", "", "", "", "compress"));
+    }
+
+    @Test
+    public void splitAsciiString() {
+        assertThat(split(newAsciiString(" ,      "), ',', false),
```
Scottmitch, line 101: lets share code between the `String` and `AsciiString` test methods:

```java
    private static void splitNoTrim(Function<String, ? extends CharSequence> f) {
        assertThat(split(f.apply(" ,      "), ',', false),
                contains(f.apply(" "), f.apply("      ")));
        // ...
    }

    @Test
    public void splitString2() {
        splitNoTrim(identity());
    }

    @Test
    public void splitAsciiString2() {
        splitNoTrim(CharSequences::newAsciiString);
    }
```

Scottmitch (changes requested):

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
tkountis: This is confusing. Even `String#split` has different outcomes which I find surprising.

- ` " ,,,, ".split(",") => [ , , , ,  ]`
- ` ",,,, ".split(",") => [, , , ,  ]`
- `" ,,,,".split(",") => [ ]`
- `" ,,, ,".split(",") => [ , , ,  ]`
- `" , , , , ".split(",") => [ ,  ,  ,  ,  ]`

Python on the other hand, is more predictable ie. WYSIWYG
eg.

- `" ,,, ,".split(",") => [' ', '', '', ' ', '']`

My expectation is the same; that if you have N delimiters in a string, you want N+1 elements in the result set, regardless if they are empty. Otherwise its surprising to the user for this `split(",,,", ',', _)` to be equal to `split("", ',', _)`.

With the trim=true & del=' ', that is a special case, where I chose to give priority to the del, thus these two should behave similarly

- ` ",,,,".split(",") => [ , , , , ]`
- ` "    ".split("") => [ , , , , ]`

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
```diff
@@ -206,39 +206,97 @@ public static int caseInsensitiveHashCode(CharSequence seq) {
...
+        for (int i = 0; i < input.length(); i++) {
+            char c = input.charAt(i);
+
+            if (c == delimiter) {
+                result.add(subsequence(isAscii, input, startIndex, i));
+                startIndex = i + 1;
+            }
+        }
+
+        if (startIndex != -1) {
```
idelpivnitskiy, line 242: How the `startIndex` can be equal to `-1`? we start with 0 and can only have positive numbers later: `startIndex = i + 1`

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
```diff
@@ -206,39 +206,97 @@ public static int caseInsensitiveHashCode(CharSequence seq) {
...
+            if ((input.length() - startIndex) > 0) {
+                result.add(subsequence(isAscii, input, startIndex, input.length()));
+            } else {
+                result.add(isAscii ? newAsciiString("") : "");
```
idelpivnitskiy, line 246: Consider using a constant for `newAsciiString("")`. There is `AsciiBuffer.EMPTY_ASCII_BUFFER` constant

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcMessageEncodingTest.java
```diff
@@ -213,103 +213,95 @@ public StreamingHttpServiceFilter create(final StreamingHttpService service) {
...
-                                            .map((CharSequence::toString))
-                                            .collect(toList());
+                        return response;
+                    }));
+                }
 
-                            if (!expectedRespAcceptedEncodings.isEmpty() && !actualRespAcceptedEncodings.isEmpty()) {
-                                assertEquals(expectedRespAcceptedEncodings, actualRespAcceptedEncodings);
-                            }
+                private void _handle(final List<ContentCodec> clientSupportedEncodings,
```
idelpivnitskiy, line 240: We usually use `handle0` pattern for internal methods instead of the leading underscore

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcMessageEncodingTest.java
```diff
@@ -213,9 +212,7 @@ public StreamingHttpServiceFilter create(final StreamingHttpService service) {
                                 .get(MESSAGE_ACCEPT_ENCODING, "NOT_PRESENT").toString().split(","))
                                     .map((String::trim)).collect(toList());
```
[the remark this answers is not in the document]
idelpivnitskiy: reminder: still not updated

idelpivnitskiy (approved):
LG!

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
idelpivnitskiy: 🤯 interesting discussion.
Ideally, we should be consistent with JDK behavior of `String#split`, but I'm agreed that it behaves with lots of confusion. 
Let's merge the current version, because it fixes the bug. And we can follow-up here later. Anyway, usage of the `split` method is limited atm to compression headers only.

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
tkountis: Not the case anymore behavior is as described in comment above.

- `split("  ", ' ', true) -> ["", "", ""]`

similar to
- `split(",,", ' ', true) -> ["", "", ""]`

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
tkountis: leftover

PR #1336, opened 2021-01-26
Completable#mergeDelayError(Publisher) and HTTP client control flow

Scottmitch:
Motivation:
The HTTP client control flow merges the resule of the write Publisher
(a Completable) and the read Publisher via Completable#merge(Publisher).
However when offloading is enabled it is possible events will be
delivered out of order which may result in dropped data/trailers.

Modifications:
- Introduce the Completable#mergeDelayError(Publisher) operator
- Use new mergeDelayError(..) operator instead of merge(..) so that both
streams must complete which ensures all the read data is delivered to
the application before any errors. It is assumed that if the write
Publisher fails that the read publisher will also fail due to transport
failure.

Result:
No more dropped data/trailres on HTTP client due to write failure (e.g.
connection closure) being delivered before read data due to offloading
(e.g. not enough demand, subscription is offloaded, demand arrives after
write error is already delivered).

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableMergeWithPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableMergeWithPublisher.java
@@ -24,7 +24,10 @@ import io.servicetalk.concurrent.internal.DelayedSubscription;
 import io.servicetalk.concurrent.internal.SignalOffloader;
 
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import javax.annotation.Nullable;
+
+import static java.util.Objects.requireNonNull;
 
 /**
  * {@link Publisher} as returned by {@link Completable#merge(Publisher)}.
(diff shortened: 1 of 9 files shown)


Scottmitch:

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableMergeWithPublisher.java
```diff
@@ -34,22 +37,152 @@
...
+                        throw duplicateTerminalException(currState);
+                    }
+                    if (currState.cause == null) {
+                        if (terminalSignal.cause == null) {
+                            offloadedSubscriber.onComplete();
+                        } else {
+                            offloadedSubscriber.onError(terminalSignal.cause);
+                        }
+                    } else {
+                        offloadedSubscriber.onError(currState.cause);
```
idelpivnitskiy, line 134: Looking at `CompletableMergeSubscriber`, there we add the 2nd error as suppressed for the 1st one, but here we propagate only the first failure. The 2nd may also be useful, consider propagating both.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionCloseHeaderHandlingTest.java
```diff
@@ -417,12 +423,13 @@ public void clientCloseTwoPipelinedRequestsSentFirstInitiatesClosure() throws Ex
                 responseReceived.countDown();
             });
             // Send another request before connection receives a response for the first request:
-            assertClosedChannelException("/second");
+            Future<StreamingHttpResponse> secondFuture = sendZeroLengthRequest("/second");
```
idelpivnitskiy, line 426: +1,  the change is good enough for these tests

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableMergeWithPublisher.java
```diff
@@ -67,7 +200,6 @@ void handleSubscribe(final Subscriber<? super T> subscriber, final SignalOffload
             // invoke it from the mergeWith Publisher Executor thread.
             this.offloadedSubscriber = new ConcurrentTerminalSubscriber<>(signalOffloader.offloadSubscriber(
```
idelpivnitskiy, line 201: We use `ConcurrentSubscription` and `ConcurrentTerminalSubscriber` to handle concurrency in this class. The `ConcurrentSubscription` has a comment clarifying why it's required. Consider adding a similar comment for `ConcurrentTerminalSubscriber`.

idelpivnitskiy (approved):
🎉  for more operators!

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/CompletableMergeWithPublisherTest.java
```diff
@@ -69,55 +116,141 @@ public void testDelayedPublisherSubscriptionForCancelBuffering() {
...
-        LegacyTestCompletable completable = new LegacyTestCompletable();
+        TestCompletable completable = new TestCompletable.Builder().disableAutoOnSubscribe().build();
+        TestCancellable cancellable = new TestCancellable();
         toSource(completable.merge(publisher)).subscribe(subscriber);
         publisher.onSubscribe(subscription);
+        completable.onSubscribe(cancellable);
         assertFalse(subscription.isCancelled());
         publisher.onError(DELIBERATE_EXCEPTION);
-        completable.verifyCancelled();
+        assertTrue(cancellable.isCancelled());
```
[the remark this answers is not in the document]
Scottmitch: https://github.com/apple/servicetalk/pull/1334 to provides finer control here but there is still a race condition (two subscribers may be on different threads) that I don't think we can avoid. In practice I think it is OK as the subscription is cancelled after termination which means any interactions should be "NOPs".

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableMergeWithPublisher.java
Scottmitch: `ConcurrentTerminalSubscriber` usage isn't changed in this PR and it is removed all together in #1334 so lets defer on this for now.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableMergeWithPublisher.java
Scottmitch: I'll add it but I'm hesitant to use `addSuppressed` in general as it may result in memory leak with static exceptions, difficult to "turn off" and limit the queue size ... but we can look at this issue more generally in a followup.

Scottmitch: actually unit tests remind me of another issue with `addSuppressed`, self-suppression throws an exception so we would have to guard against this everywhere (which we don't). I'll skip `addSuppressed` for now...

PR #1338, opened 2021-01-27
URL fix for JAX-RS API documentation

bondolo:
Motivation:
The public URL for the JAX-RS 2.1.X API documentation has moved to the Eclipse Jakarta EE
site. This has broken links to the documentation in the published ServiceTalk.io site and
prevents successful validation of generated site documentation for local builds.

Modifications:
The broken URL is updated to match the current published location for JAX-RS API
documentation.

Result:
Validation succeeds for local documentation builds and the link is no longer broken.


idelpivnitskiy (approved):
Good catch, thank you!

PR #1341, opened 2021-01-29
gRPC Server allow drop of request trailers

Scottmitch:
Motivation:
The
[gRPC protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests)
does not use request trailers. This means the server is allowed to drop
trailers if any are present. This may provide a performance benefit as
request transformation operations do not have to worry about preserving
trailers.

Modifications:
- DefaultGrpcServerBuilder to use the allowDropRequestTrailers(true)
  method.

Result:
gRPC Server is allowed to drop the request trailers, requiring less work
to preserve trailers (which don't exist) through request transform
operations.

### DIFF
--- a/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
+++ b/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
@@ -54,7 +54,7 @@ final class DefaultGrpcServerBuilder extends GrpcServerBuilder implements Server
             .executionStrategy(defaultStrategy());
 
     DefaultGrpcServerBuilder(final HttpServerBuilder httpServerBuilder) {
-        this.httpServerBuilder = httpServerBuilder.protocols(h2Default());
+        this.httpServerBuilder = httpServerBuilder.protocols(h2Default()).allowDropRequestTrailers(true);
     }
 
     @Override


servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
```diff
@@ -54,7 +54,7 @@
             .executionStrategy(defaultStrategy());
 
     DefaultGrpcServerBuilder(final HttpServerBuilder httpServerBuilder) {
-        this.httpServerBuilder = httpServerBuilder.protocols(h2Default());
+        this.httpServerBuilder = httpServerBuilder.protocols(h2Default()).allowDropRequestTrailers(true);
```
idelpivnitskiy, line 57: We do not use methods with `boolean` argument on the builders. Instead, we provide methods that override the default. Examples:
1. `HttpServerBuilder#disableDrainingRequestPayloadBody()`
2. `H1SpecExceptions.Builder#allowPrematureClosureBeforePayloadBody()`
3. `DefaultAutoRetryStrategyProvider.Builder#disableWaitForLoadBalancer()`
Consider removing a boolean argument from `allowDropRequestTrailers` method for consistency, before we released it.

idelpivnitskiy (approved):
The change for `DefaultGrpcServerBuilder` looks reasonable. Let's see how it improves grpc-server perf 🤞 
One suggestion for `allowDropRequestTrailers` API:

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
Scottmitch: Lets consider API changes as a followup PR. In general I'm not sure why we are averse to `boolean` argument methods, which seems like a canonical approach for builders. Passing the state as an arguments may remove additional conditional logic from user code:

```java
Builder builder = ...;
if(shouldAllow) {
  builder.shouldAllow();
} else {
  builder.noAllow();
} 
builder.build();
```

vs
```java
new Builder().shouldAllow(shouldAllow).build();
```

idelpivnitskiy: We do not provide methods that flip the default back either. Only the one that changes the default:

```java
Builder builder = ...;
if(shouldAllow) {
  builder.shouldAllow();
}
builder.build();
```

PR #1342, opened 2021-01-29
Examples for HTTP and gRPC compression

bondolo:
Motivation:
Provide an example demonstrating use of the content encoding compression features.

Modifications:
- A new HTTP example, Compression
- Additional documentation

Result:
No further need to reference test code to understand how to use compression features.

### DIFF
--- a/servicetalk-examples/grpc/compression/build.gradle
+++ b/servicetalk-examples/grpc/compression/build.gradle
@@ -1,0 +1,68 @@
+/*
+ * Copyright © 2019, 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+buildscript {
+  dependencies {
+    classpath "com.google.protobuf:protobuf-gradle-plugin:$protobufGradlePluginVersion"
+  }
+}
+
+apply plugin: "java"
+apply plugin: "com.google.protobuf"
+apply from: "../../gradle/idea.gradle"
+
+dependencies {
+  implementation project(":servicetalk-annotations")
+  implementation project(":servicetalk-grpc-netty")
+  implementation project(":servicetalk-grpc-protoc")
+  implementation project(":servicetalk-grpc-protobuf")
+
+  implementation "org.slf4j:slf4j-api:$slf4jVersion"
+  runtimeOnly "org.apache.logging.log4j:log4j-slf4j-impl:$log4jVersion"
+}
+
+protobuf {
+  protoc {
+    artifact = "com.google.protobuf:protoc:$protobufVersion"
+  }
+  plugins {
+    servicetalk_grpc {
+      //// Users are expected to use "artifact" instead of "path". we use "path"
+      //// only because we want to use the gradle project local version of the plugin
+      // artifact = "io.servicetalk:servicetalk-grpc-protoc:$serviceTalkVersion:all@jar"
+      path = file("${project.rootProject.rootDir}/servicetalk-grpc-protoc/build" +
+                  "/buildExecutable/servicetalk-grpc-protoc-${project.version}-all.jar").path
+    }
+  }
+  generateProtoTasks {
+    all().each { task ->
+      task.plugins {
+        servicetalk_grpc {
+          // Need to tell protobuf-gradle-plugin to output in the correct directory if all generated
+          // code for a single proto goes to a single file (e.g. "java_multiple_files = false" in the .proto).
+          outputSubDir = "java"
+        }
+      }
+    }
+  }
+  generatedFilesBaseDir = "$buildDir/generated/sources/proto"
+}
+
+// The following setting must be omitted in users projects and is necessary here
+// only because we want to use the locally built version of the plugin
+afterEvaluate {
+  generateProto.dependsOn(":servicetalk-grpc-protoc:buildExecutable")
+}
--- a/servicetalk-examples/grpc/compression/src/main/java/io/servicetalk/examples/grpc/compression/CompressionExampleClient.java
+++ b/servicetalk-examples/grpc/compression/src/main/java/io/servicetalk/examples/grpc/compression/CompressionExampleClient.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright © 2018, 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.examples.grpc.compression;
+
+import io.servicetalk.encoding.api.ContentCodec;
+import io.servicetalk.encoding.api.ContentCodings;
+import io.servicetalk.grpc.netty.GrpcClients;
+
+import io.grpc.examples.compression.Greeter;
+import io.grpc.examples.compression.Greeter.ClientFactory;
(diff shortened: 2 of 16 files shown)


bondolo:
Hmmm, I had intended to propose these as separate PRs. Oh well, they are fine together.

servicetalk-examples/grpc/compression/build.gradle
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright © 2019, 2021 Apple Inc. and the ServiceTalk project authors
```
idelpivnitskiy, line 2: New files should have only the current year

servicetalk-examples/grpc/compression/src/main/java/io/servicetalk/examples/grpc/compression/CompressionExampleServer.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright © 2021, 2021 Apple Inc. and the ServiceTalk project authors
```
idelpivnitskiy, line 2: Duplicated year

servicetalk-examples/grpc/compression/src/main/java/io/servicetalk/examples/grpc/compression/package-info.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright © 2021 2021 Apple Inc. and the ServiceTalk project authors
```
idelpivnitskiy, line 2: Duplicated year

servicetalk-examples/http/compression/src/main/java/io/servicetalk/examples/http/compression/CompressionFilterExampleServer.java
```diff
@@ -0,0 +1,73 @@
...
+                // Adds a content coding service filter that includes the encodings supported for requests and the
+                // preferred encodings for responses. Responses will automatically be compressed if the request includes
+                // a mutually agreeable compression encoding that the client indicates they will accept and that the
+                // server supports.
+                .appendServiceFilter(new ContentCodingHttpServiceFilter(SUPPORTED_REQ_ENCODINGS, SUPPORTED_RESP_ENCODINGS))
+                .listenAndAwait((ctx, request, responseFactory) -> {
+                        String who = request.payloadBody(textDeserializer());
+                        return succeeded(responseFactory.ok()
+                                .payloadBody("Hello " + who +  "!", textSerializer())
+                                );
```
idelpivnitskiy, line 69: nit: it can be on the previous line

idelpivnitskiy (commented):
Great work @bondolo!
Can you please update PR title and description to account for gRPC examples as well?

servicetalk-examples/grpc/compression/src/main/java/io/servicetalk/examples/grpc/compression/CompressionExampleClient.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    private static final List<ContentCodec> PREFERRED_ENCODINGS = List.of(
+            ContentCodings.gzipDefault(),
+            ContentCodings.deflateDefault(),
+            ContentCodings.identity()
+    );
+
+    public static void main(String... args) throws Exception {
+        try (GreeterClient client = GrpcClients.forAddress("localhost", 8080).build(new ClientFactory()
+                // Requests will include 'Accept-Encoding' HTTP header to inform the server which encodings we support.
```
tkountis, line 45: It's 'Message-Accept-Encoding' for gRPC

servicetalk-examples/grpc/compression/src/main/java/io/servicetalk/examples/grpc/compression/CompressionExampleServer.java
```diff
@@ -0,0 +1,80 @@
...
+            ContentCodings.identity()
+    );
+
+    public static void main(String... args) throws Exception {
+        GrpcServers.forPort(8080)
+                // Adds a content coding service filter that includes the encodings supported for requests and the
+                // preferred encodings for responses. Responses will automatically be compressed if the request includes
+                // a mutually agreeable compression encoding that the client indicates they will accept and that the
+                // server supports.
+                .appendHttpServiceFilter(new ContentCodingHttpServiceFilter(SUPPORTED_REQ_ENCODINGS, SUPPORTED_RESP_ENCODINGS))
```
tkountis, line 68: The `ContentCodingHttpServiceFilter` only applies to HTTP compression, not gRPC.
Look at https://github.com/apple/servicetalk/blob/main/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/api/GrpcMessageEncodingTest.java#L420 for how to enable for a gRPC service. This API is code-generated.

tkountis: Also please update the comment in relation to the content-coding reference.

servicetalk-examples/grpc/compression/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright © 2018 Apple Inc. and the ServiceTalk project authors
```
tkountis, line 3: New file, Copyright year

tkountis (commented):
Good work @bondolo, I left some comments. I would consider also printing out the response header of the server, to demonstrate compression selection.

servicetalk-examples/grpc/compression/build.gradle
bondolo: Since it is a copy of the hello world file it is a derived work and probably should retain the original copyright. Consider that if it had someone else's copyright on it and we copied and altered it for the compression it would not be different enough. If there is a standard for "new work" vs "derived work" I will be happy to follow that. It matters less in this case since there is unlikely to be a claim of copyright infringement by Apple against itself, but something to keep in mind for files with others copyrights, even if the license allows us to copy and modify.

servicetalk-examples/grpc/compression/src/main/java/io/servicetalk/examples/grpc/compression/CompressionExampleServer.java
bondolo: Are both needed or just the change to use the other ServiceFactory overload?

tkountis: Only the service factory change. The filter is only for HTTP. You can verify behaviour by logging the frames, see. '.enableWireLogging()'

idelpivnitskiy: Compression for gRPC and HTTP works at different levels:
- HTTP compresses the entire payload body and uses the `content-encoding` header to signal which compression was applied.
- gRPC compresses each proto-message independently and uses `Compressed-Flag` to mark which are compressed and `grpc-encoding` header to tell which algorithm was used.
 
Therefore, filters that are used for HTTP compression can not be used with gRPC because they break protocol: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

bondolo: OK, thank you for the explanation. I will remove the HTTP filter.

servicetalk-examples/grpc/compression/build.gradle
idelpivnitskiy: Good point! @Scottmitch ptal in case we need a new policy for copyright headers when the file is copied

servicetalk-examples/grpc/compression/src/main/java/io/servicetalk/examples/grpc/compression/CompressionExampleServer.java
```diff
@@ -0,0 +1,83 @@
...
+ * A simple extension of the gRPC "Hello World" example which demonstrates
+ * compression of the request and response bodies.
+ */
+public class CompressionExampleServer {
+
+    /**
+     * Supported encodings for the request. Requests using unsupported encodings receive
+     * an error response in the "grpc-status".
+     */
+    private static final List<ContentCodec> SUPPORTED_REQ_ENCODINGS =
```
idelpivnitskiy, line 46: This list is not used. IIUC, because gRPC protocol (unlike HTTP) always expects compressed request messages from the client, there is no need to have two lists. Consider keeping only one list and removing `_REQ_/_RESP_` from the middle of the name: `SUPPORTED_ENCODINGS`.

idelpivnitskiy (approved):
Great! Hope it will help users to start with using compression in ST 🎉 
LGTM after the unused constant is removed:

servicetalk-examples/grpc/compression/src/main/java/io/servicetalk/examples/grpc/compression/CompressionExampleServer.java
bondolo: Correct. Now removed with suggested name change

PR #1343, opened 2021-01-30
NettyChannelPublisher duplicate onError

Scottmitch:
Motivation:
NettyChannelPublisher may deliver multiple onError terminal signals to
the same subscriber if the Subscriber#onSubscribe(..) method requests
and there is an error queued up. The reactive streams specification only
allows for a single terminal signal to be delivered.

Modifications:
- NettyChannelPublisher handleSubscribe method should verify the
  subscriber has not already been terminated to check if a queued error
  should be delivered.

Result:
NettyChannelPublisher handleSubscribe will deliver at most 1 terminal
signal to the Subscriber.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelPublisher.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelPublisher.java
@@ -166,50 +166,42 @@ final class NettyChannelPublisher<T> extends SubscribablePublisher<T> {
 
     private boolean processPending(SubscriptionImpl target) {
         // Should always be called from eventloop. (assert done before calling)
-        if (inProcessPending || pending == null) {
-            // Guard against re-entrance.
-            return false;
-        }
-
-        inProcessPending = true;
-        try {
-            while (requestCount > 0) {
-                Object p = pending.poll();
-                if (p == null) {
-                    break;
-                }
-                if (p instanceof TerminalNotification) {
-                    Throwable throwable = ((TerminalNotification) p).cause();
-                    assert throwable != null : "onComplete notification can not be enqueued.";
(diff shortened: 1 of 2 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelPublisher.java
```diff
@@ -166,50 +166,42 @@ private void requestN(long n, SubscriptionImpl forSubscription) {
...
+        if (!inProcessPending && pending != null && !pending.isEmpty()) {
+            inProcessPending = true;
+            try {
+                while (requestCount > 0) {
+                    Object p = pending.poll();
+                    if (p == null) {
+                        break;
+                    }
+                    if (p instanceof TerminalNotification) {
+                        sendErrorToTarget(target, (TerminalNotification) p);
```
Scottmitch, line 178: the code in this method wasn't meaningfully changed, just refactored to share code and remove the initial return branch.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelPublisher.java
```diff
@@ -305,7 +303,8 @@ private void subscribe0(Subscriber<? super T> subscriber) {
             subscriber.onSubscribe(subscription);
             // Fatal error is removed from the queue once it is drained for a Subscriber.
             // In absence of the below, any subsequent Subscriber will not get any fatal error.
-            if (!processPending(subscription) && (fatalError != null && (pending == null || pending.isEmpty()))) {
+            if (subscription == this.subscription && !processPending(subscription) &&
```
idelpivnitskiy, line 306: 🎉

idelpivnitskiy (approved):
Great catch!

PR #1344, opened 2021-02-02
Add MDC ThreadContextMap conflict check on ST default provider

tkountis:
**Motivation**:

When multiple MDC providers are available in the cp, then unless the user specifies one through a system property, we can't know which one will be used. 

**Modifications**:

Add conflict check on ST's default provider, to detect presence of the other provider, and throw an exception.

**Result**:

Fast fail for common conflicts in the class path.

**Example output**:
```
Detected multiple ServiceTalk MDC adapters in classpath. The [io.servicetalk.log4j2.mdc.DefaultServiceTalkThreadContextMap, io.servicetalk.opentracing.log4j2.ServiceTalkTracingThreadContextMap] MDC adapters should not be loaded at the same time. Please exclude one from your dependencies.
ERROR StatusLogger Unable to locate or load configured ThreadContextMap io.servicetalk.opentracing.log4j2.ServiceTalkTracingThreadContextMap
 java.lang.IllegalStateException: log4j2.threadContextMap is not set to io.servicetalk.opentracing.log4j2.ServiceTalkTracingThreadContextMap. There is no way to determine which ThreadContextMap has precedence for MDC storage and behavior is therefore undefined.
	at io.servicetalk.log4j2.mdc.utils.ServiceTalkThreadContextMap.detectPossibleConflicts(ServiceTalkThreadContextMap.java:82)
	at io.servicetalk.log4j2.mdc.utils.ServiceTalkThreadContextMap.<init>(ServiceTalkThreadContextMap.java:58)
	at io.servicetalk.opentracing.log4j2.ServiceTalkTracingThreadContextMap.<init>(ServiceTalkTracingThreadContextMap.java:44)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at java.base/java.lang.Class.newInstance(Class.java:584)
	at org.apache.logging.log4j.spi.ThreadContextMapFactory.createThreadContextMap(ThreadContextMapFactory.java:106)
	at org.apache.logging.log4j.ThreadContext.init(ThreadContext.java:224)
	at org.apache.logging.log4j.ThreadContext.<clinit>(ThreadContext.java:202)
	at org.apache.logging.log4j.core.impl.ContextDataInjectorFactory.createDefaultInjector(ContextDataInjectorFactory.java:87)
	at org.apache.logging.log4j.core.impl.ContextDataInjectorFactory.createInjector(ContextDataInjectorFactory.java:71)
	at org.apache.logging.log4j.core.lookup.ContextMapLookup.<init>(ContextMapLookup.java:34)
	at org.apache.logging.log4j.core.lookup.Interpolator.<init>(Interpolator.java:126)
	at org.apache.logging.log4j.core.config.AbstractConfiguration.<init>(AbstractConfiguration.java:129)
	at org.apache.logging.log4j.core.config.NullConfiguration.<init>(NullConfiguration.java:32)
	at org.apache.logging.log4j.core.LoggerContext.<clinit>(LoggerContext.java:88)
	at org.apache.logging.log4j.core.selector.ClassLoaderContextSelector.createContext(ClassLoaderContextSelector.java:249)
	at org.apache.logging.log4j.core.selector.ClassLoaderContextSelector.locateContext(ClassLoaderContextSelector.java:213)
	at org.apache.logging.log4j.core.selector.ClassLoaderContextSelector.getContext(ClassLoaderContextSelector.java:137)
	at org.apache.logging.log4j.core.selector.ClassLoaderContextSelector.getContext(ClassLoaderContextSelector.java:124)
	at org.apache.logging.log4j.core.selector.ClassLoaderContextSelector.getContext(ClassLoaderContextSelector.java:118)
	at org.apache.logging.log4j.core.impl.Log4jContextFactory.getContext(Log4jContextFactory.java:150)
	at org.apache.logging.log4j.core.impl.Log4jContextFactory.getContext(Log4jContextFactory.java:47)
	at org.apache.logging.log4j.LogManager.getContext(LogManager.java:194)
	at org.apache.logging.log4j.LogManager.getLogger(LogManager.java:597)
	at Test.<clinit>(Test.java:12)
```

### DIFF
--- a/servicetalk-log4j2-mdc-utils/src/main/java/io/servicetalk/log4j2/mdc/utils/ServiceTalkThreadContextMap.java
+++ b/servicetalk-log4j2-mdc-utils/src/main/java/io/servicetalk/log4j2/mdc/utils/ServiceTalkThreadContextMap.java
@@ -27,7 +27,10 @@ import org.apache.logging.log4j.util.ReadOnlyStringMap;
 import org.apache.logging.log4j.util.StringMap;
 import org.apache.logging.log4j.util.TriConsumer;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import javax.annotation.Nullable;
@@ -41,6 +44,47 @@ public class ServiceTalkThreadContextMap implements ReadOnlyThreadContextMap, Cl
     private static final Key<Map<String, String>> key = Key.newKey("log4j2Mdc");
     @SuppressWarnings("IllegalInstantiationOfString")
     private static final String NULL_STRING = "";
+
+    private static final String[] KNOWN_CONFLICTS = {
+            "io.servicetalk.log4j2.mdc.DefaultServiceTalkThreadContextMap",
+            "io.servicetalk.opentracing.log4j2.ServiceTalkTracingThreadContextMap",
+    };
+
+    private static final String ADAPTER_CONFLICT_MESSAGE = "Detected multiple ServiceTalk MDC adapters in classpath." +
+            " The %s MDC adapters should not be" +
+            " loaded at the same time. Please exclude one from your dependencies.%n";
+
+    public ServiceTalkThreadContextMap() {
+        detectPossibleConflicts();
+    }
+
+    private void detectPossibleConflicts() {
+        List<String> found = new ArrayList<>(KNOWN_CONFLICTS.length);
+        for (String cls : KNOWN_CONFLICTS) {
+            try {
+                Class.forName(cls);
+                found.add(cls);
+            } catch (ClassNotFoundException ex) {
+                // Not found - continue
+            }
+        }
+
+        if (found.size() > 1) {
+            System.err.printf((ADAPTER_CONFLICT_MESSAGE), Arrays.toString(found.toArray()));
+            // Log4j's provider interface provides no formal way to declare ordering or precedence if there are multiple
+            // provides which may overlap. Currently log4j2's ThreadContextMapFactory gives precedent to the logger
+            // specified via system property [1] so we check for this as a "best effort" to detect that this class has
+            // precedence and will be used for MDC storage.
+            // [1] https://logging.apache.org/log4j/2.0/log4j-api/apidocs/org/apache/logging/log4j/spi/ThreadContextMapFactory.html
+            if (!System.getProperty("log4j2.threadContextMap", "")
+                    .equals(this.getClass().getCanonicalName())) {
+                throw new IllegalStateException("log4j2.threadContextMap is not set to " +
+                        this.getClass().getCanonicalName() + ". There is no way to determine" +
+                        " which ThreadContextMap has precedence for MDC storage and behavior is therefore undefined.");
+            }
+            // Do NOT use log4j2 to log here as it can cause some reentrancy problem
+        }
+    }
 
     @Override
     public final void put(String key, String value) {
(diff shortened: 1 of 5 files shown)


servicetalk-log4j2-mdc-utils/src/main/java/io/servicetalk/log4j2/mdc/utils/ServiceTalkThreadContextMap.java
```diff
@@ -42,6 +45,47 @@
     @SuppressWarnings("IllegalInstantiationOfString")
     private static final String NULL_STRING = "";
 
+    private static final String[] KNOWN_CONFLICTS = {
+            "io.servicetalk.log4j2.mdc.DefaultServiceTalkThreadContextMap",
```
Scottmitch, line 49: is this class used to differentiate between classes because `ServiceTalkTracingThreadContextMap` extends the class? Since the only conflict we are interested in involves one class extending another an alternative to adding additional classes and checking loaded classes is as follows:

```java
    private static final AtomicReference<String> loadedClassName = new AtomicReference<>();

    /**
     * Create a new instance.
     */
    public ServiceTalkThreadContextMap() {
        detectPossibleConflicts(getClass().getCanonicalName());
    }

    private static void detectPossibleConflicts(String thisClassName) {
        for (;;) {
            String existingClassName = loadedClassName.get();
            if (existingClassName == null) {
                if (loadedClassName.compareAndSet(null, thisClassName)) {
                    break;
                }
            } else if (existingClassName.equals(thisClassName)) {
                break;
            } else {
                System.err.println("Found conflicting MDC adapter " + existingClassName + " when attempting to load " +
                        thisClassName);
                // Log4j's provider interface provides no formal way to declare ordering or precedence if there are multiple
                // provides which may overlap. Currently log4j2's ThreadContextMapFactory gives precedent to the logger
                // specified via system property [1] so we check for this as a "best effort" to detect that this class has
                // precedence and will be used for MDC storage.
                // [1] https://logging.apache.org/log4j/2.0/log4j-api/apidocs/org/apache/logging/log4j/spi/ThreadContextMapFactory.html
                if (!System.getProperty("log4j2.threadContextMap", "").equals(thisClassName)) {
                    throw new IllegalStateException("log4j2.threadContextMap is not set to " +
                            thisClassName + ". There is no way to determine" +
                            " which ThreadContextMap has precedence for MDC storage and behavior is therefore undefined.");
                }
                // Do NOT use log4j2 to log here as it can cause some reentrancy problem
            }
        }
    }
```

tkountis: These classes won't be loaded together, so there is no need for the atomic operation. Actually, this solution won't report the issue. The problem lies to the order of the dependencies in your class path. If open tracing comes first, then that one will be loaded, otherwise the default mdc will. So there is no form of racing here. We just need to detect the presence of multiple behaviors in the CP and report to the user to avoid confusion; that's why we have to look for the classes specifically by name.

example:

- this will load the default solution
```
    implementation("io.servicetalk:servicetalk-log4j2-mdc:0.37.0-SNAPSHOT")
    implementation("io.servicetalk:servicetalk-opentracing-log4j2:0.37.0-SNAPSHOT")
```

- this will load the open tracing solution
```
    implementation("io.servicetalk:servicetalk-opentracing-log4j2:0.37.0-SNAPSHOT")
    implementation("io.servicetalk:servicetalk-log4j2-mdc:0.37.0-SNAPSHOT")
```

But not both.

Scottmitch: good point, implicit classpath loading too magical 😭 it feels wrong putting info on children classes in the parent class but I don't have a better idea (other than not doing anything)

Scottmitch (approved):

PR #1345, opened 2021-02-02
Fix broken encoding when publisher has no content

tkountis:
**Motivation**:

When a payload publisher on the response flow is not emitting any items, the encoder will not emit the header, thus a client will not be able to decode the response.

**Modifications**:

Make sure the header is emitted even for empty publishers.
Extra checks where added to skip encoding & headers when the request doesn't support body-messages.
 
**Result**:

Clients will be able to decode correctly encoded response payload, even when empty.

### DIFF
--- a/servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
+++ b/servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2020 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2020, 2021 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -45,6 +45,7 @@ abstract class AbstractZipContentCodec extends AbstractContentCodec {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AbstractZipContentCodec.class);
     private static final Buffer END_OF_STREAM = DEFAULT_RO_ALLOCATOR.fromAscii(" ");
+    private static final int FOOTER_LEN = 10;
 
     protected final int chunkSize;
     private final int maxPayloadSize;
@@ -103,8 +104,8 @@ abstract class AbstractZipContentCodec extends AbstractContentCodec {
         return from
                 .concat(succeeded(END_OF_STREAM))
                 .liftSync(subscriber -> new PublisherSource.Subscriber<Buffer>() {
-                    @Nullable
-                    SwappableBufferOutputStream stream;
+
+                    final SwappableBufferOutputStream stream = new SwappableBufferOutputStream();
 
                     @Nullable
                     DeflaterOutputStream output;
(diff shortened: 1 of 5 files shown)


Scottmitch:

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ContentCodingHttpServiceFilter.java
```diff
@@ -126,20 +129,50 @@ public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strat
     }
 
     private static void encodePayloadContentIfAvailable(final HttpHeaders requestHeaders,
+                                                        final HttpRequestMethod method,
+                                                        final HttpProtocolVersion version,
                                                         final List<ContentCodec> supportedEncodings,
                                                         final StreamingHttpResponse response,
                                                         final BufferAllocator allocator) {
         if (supportedEncodings.isEmpty() || hasContentEncoding(response.headers())) {
```
Scottmitch, line 137: nit: add `|| isPassThrough(method, version, response)` to this conditional instead of creating a new `if` block

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ContentCodingHttpServiceFilter.java
```diff
@@ -126,20 +129,50 @@ public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strat
...
+        //
+        // 2.  Any 2xx (Successful) response to a CONNECT request implies that
+        // the connection will become a tunnel immediately after the empty
+        // line that concludes the header fields.  A client MUST ignore any
+        // Content-Length or Transfer-Encoding header fields received in
+        // such a message.
+        // ...
+
+        int code = response.status().code();
+        return code < 200 || code == 204 || code == 304 ||
```
Scottmitch, line 172: `code < 200` -> `INFORMATIONAL_1XX.equals(response.status().statusClass())`

`code == 204 || code == 304` -> `NO_CONTENT.code() || code == NOT_MODIFIED.code()`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ContentCodingHttpServiceFilter.java
```diff
@@ -126,20 +129,50 @@ public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strat
...
+                (method == HEAD || (method == CONNECT && code == 200));
```
Scottmitch, line 173: `method == CONNECT && code == 200` -> `method == CONNECT && SUCCESSFUL_2XX.equals(response.status().statusClass())`

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
```diff
@@ -138,11 +138,18 @@ public void onNext(Buffer next) {
...
-                                stream.swap(dst);
-                                output.finish();
+                                Buffer dst;
+                                if (!headerWritten) {
+                                    // Empty publisher will not produce any items other than the END_OF_STREAM
+                                    // so we need to keep the header & footer in the same chunk and deliver downstream.
+                                    dst = stream.buffer;
+                                } else {
+                                    // ZIP footer is 10 bytes
+                                    dst = allocator.newBuffer(10);
```
Scottmitch, line 148: nit: consider creating a `public static final int FOOTER_BYTES = 10`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ServiceTalkEmptyContentCodingTest.java
```diff
@@ -0,0 +1,45 @@
...
+    public ServiceTalkEmptyContentCodingTest(final HttpProtocol protocol, final Codings serverCodings,
+                                             final Codings clientCodings, final Compression compression,
+                                             final boolean valid) {
+        super(protocol, serverCodings, clientCodings, compression, valid);
+    }
+
+    protected void assertResponse(final StreamingHttpResponse response) throws Throwable {
+        verifyNoErrors();
+
+        assertResponseHeaders(response.headers().get(CONTENT_ENCODING, "identity").toString());
```
Scottmitch, line 38: nit : `"identity"` -> `HttpHeaderValues.IDENTITY`

same comment in base class

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
```diff
@@ -138,11 +138,18 @@ public void onNext(Buffer next) {
...
                             if (next == END_OF_STREAM) {
-                                // ZIP footer is 10 bytes
-                                Buffer dst = allocator.newBuffer(10);
-                                stream.swap(dst);
-                                output.finish();
+                                Buffer dst;
+                                if (!headerWritten) {
+                                    // Empty publisher will not produce any items other than the END_OF_STREAM
+                                    // so we need to keep the header & footer in the same chunk and deliver downstream.
+                                    dst = stream.buffer;
```
Scottmitch, line 145: is there any value to pre-allocating the `Buffer` now that each `onNext` checks if `headerWritten` and potentially does an allocation? consider just allocating the buffer on demand in each onNext and swapping to reduce conditional blocks.

Scottmitch (commented):
Few comments/questions, can you update the commit message too: `... has not content` -> `... has no content`?

Scottmitch:
another test failure attributed to #999

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
```diff
@@ -113,67 +114,31 @@ public final Buffer encode(final Buffer src, final int offset, final int length,
...
-                                    // ZIP footer is 10 bytes
-                                    dst = allocator.newBuffer(10);
-                                    stream.swap(dst);
-                                }
+                            Buffer dst = allocator.newBuffer(next == END_OF_STREAM ? FOOTER_LEN : chunkSize);
+                            stream.swap(dst);
 
+                            if (!headerWritten) {
+                                // This will produce header bytes
+                                output = newDeflaterOutputStream(stream);
```
Scottmitch, line 130: can `output` be created up front too? I know this is `Closable`, but the existing approach doesn't account for `Subscription.cancel()` anyways and I would propose we revisit lifetime in a followup PR (for better isolation of changes).

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
```diff
@@ -589,10 +571,10 @@ private void verifyCrc(Buffer in) throws IOException {
     }
 
     static class SwappableBufferOutputStream extends OutputStream {
+        @Nullable
```
Scottmitch, line 574: if you don't wan't to worry about null you can initialize to `EmptyBuffer.EMPTY_BUFFER`

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
tkountis: If output is created up front, then we need to allocate up-front too because it starts writing the header bytes immediately. This way we only allocate when we start consuming. I am re-working the overall lifetime + enhancements in a follow PR, let's keep further changes out of this one. I can revert if you prefer the old way, but I think this follows the 'cold' semantics of ST.

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
tkountis: I did consider that, but spot-buts starts complaining with ` RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT: Return value of method without side effect is ignored` and I didn't won't to suppress it.

Scottmitch: hum this seems like a general problem that we may want to address, but independent issue

Scottmitch (approved):

PR #1346, opened 2021-02-02
HTTP/2 CloseHandler avoid writing after channel closed

Scottmitch:
Motivation:
HTTP/2 child channels currently use a static CloseHandler instance which
will immediately close the connection upon detecting inbound channel
closing. This event may occur as a result of an exception being thrown
in error code, which may then write data (e.g. error code in response
trailers). Using the static CloseHandler may result in writing a stream
reset frame followed by subsequent data/headers frames which violates
the HTTP/2 protocol.

Modifications:
- Introduce a CloseHandler implementation that can be used in scenarios
  that don't require half-closure or DuplexChannel types. This
  CloseHandler can be used for HTTP/2 to delay any pending close
  operations until after any pending writes complete.

Result:
No more HTTP/2 protocol violations due to writing RST_STREAM frame
before data/headers frames due to error recovery.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
@@ -65,6 +65,7 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
     @Override
     public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
         if (msg instanceof HttpRequestMetaData) {
+            closeHandler.protocolPayloadBeginOutbound(ctx);
             HttpRequestMetaData metaData = (HttpRequestMetaData) msg;
             HttpHeaders h1Headers = metaData.headers();
             CharSequence host = h1Headers.getAndRemove(HOST);
@@ -102,6 +103,7 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
             Http2Headers h2Headers = headersFrame.headers();
             final HttpResponseStatus httpStatus;
             if (!readHeaders) {
+                closeHandler.protocolPayloadBeginInbound(ctx);
                 CharSequence status = h2Headers.getAndRemove(STATUS.value());
                 if (status == null) {
                     throw new IllegalArgumentException("a response must have " + STATUS + " header");
@@ -135,6 +137,7 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
                 } else {
                     ctx.fireChannelRead(h2HeadersToH1HeadersClient(h2Headers, null));
                 }
+                closeHandler.protocolPayloadEndInbound(ctx);
             } else if (httpStatus == null) {
                 throw new IllegalArgumentException("a response must have " + STATUS + " header");
             } else {
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NonPipelinedCloseHandler.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NonPipelinedCloseHandler.java
@@ -1,0 +1,183 @@
+/*
+ * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.transport.netty.internal;
+
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPromise;
+
+import java.util.function.Consumer;
+import javax.annotation.Nullable;
+
+import static io.servicetalk.transport.netty.internal.CloseHandler.CloseEvent.CHANNEL_CLOSED_INBOUND;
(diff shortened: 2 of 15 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NonPipelinedCloseHandler.java
```diff
@@ -0,0 +1,183 @@
...
+    private static final byte OUT_CLOSED = 0x8;
+    private static final byte CLOSED = 0x10;
+    private static final byte GRACEFUL_CLOSE = 0x20;
+    private static final byte IS_CLIENT = 0x40;
+    private static final byte ALL_CLOSED = IN_CLOSED | OUT_CLOSED | CLOSED;
+    private static final byte READ_WRITE = READ | WRITE;
+    private static final byte CLIENT_IN_WRITE = IS_CLIENT | WRITE | IN_CLOSED;
+    private static final byte GRACEFUL_IN_CLOSED = GRACEFUL_CLOSE | IN_CLOSED;
+    private static final byte GRACEFUL_OUT_CLOSED = GRACEFUL_CLOSE | OUT_CLOSED;
+    private byte state;
```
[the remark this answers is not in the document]
Scottmitch: Yes, the closeHandler methods are all invoked from a single `EventLoop` thread.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
```diff
@@ -65,6 +65,7 @@
     @Override
     public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
         if (msg instanceof HttpRequestMetaData) {
+            closeHandler.protocolPayloadBeginOutbound(ctx);
```
idelpivnitskiy, line 68: `HttpObjectEncoder` emits this event right before `ctx.write(serializedData)`. Consider making them consistent. We can move it in `HttpObjectEncoder` to the top, before `T metaData = castMetaData(msg);`

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NonPipelinedCloseHandler.java
```diff
@@ -0,0 +1,183 @@
...
+    @Override
+    public void protocolPayloadBeginOutbound(final ChannelHandlerContext ctx) {
+        state = set(state, WRITE);
+    }
+
+    @Override
+    public void protocolPayloadEndOutbound(final ChannelHandlerContext ctx, final ChannelPromise promise) {
+        ctx.pipeline().fireUserEventTriggered(OutboundDataEndEvent.INSTANCE);
+        promise.addListener(f -> {
+            state = unset(state, WRITE);
```
idelpivnitskiy, line 76: `RequestResponseCloseHandler` does this only if `f.isSuccess()` returns `true`. Should we make them consistent and either check the future or ignore its result?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NonPipelinedCloseHandler.java
```diff
@@ -0,0 +1,183 @@
...
+        state = set(state, OUT_CLOSED);
+        // todo storeCloseRequestAndEmit ?
+        outboundEventCheckClose(channel);
+    }
+
+    @Override
+    void gracefulUserClosing(final Channel channel) {
+        state = set(state, GRACEFUL_CLOSE);
+        storeCloseRequestAndEmit(GRACEFUL_USER_CLOSING);
+        if (!isAllSet(state, READ_WRITE)) {
```
idelpivnitskiy, line 132: IIUC, this should use `!isAnySet(state, READ_WRITE)`

idelpivnitskiy (approved):
LG!

Do you plan adding tests for `NonPipelinedCloseHandler` in isolation, similar to `RequestResponseCloseHandlerTest` to verify internal state transitions?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
Scottmitch: I'll move it after the case bcz the `closeHandler.protocolClosingOutbound(ctx);` needs the `metadata`

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NonPipelinedCloseHandler.java
Scottmitch: this class doesn't need to differentiate between success/failure. I don't understand why we need to check for `success` in `RequestResponseCloseHandler`. The original commit doesn't provide much context https://github.com/apple/servicetalk/commit/7b49eadae705fafadc2304220ec5c392b53ef2ed, and the tests (including the test modified by the commit) pass without this check, I think it should be fine to remove the `success` check and just do it unconditionally as things have changed a lot since the initial change.

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/NonPipelinedCloseHandlerTest.java
```diff
@@ -0,0 +1,273 @@
...
+        });
+    }
+
+    @Parameterized.Parameters(name = "{index}. {3} - {0} {1} = {2}")
+    public static Collection<Object[]> data() {
+        StackTraceElement se = Thread.currentThread().getStackTrace()[1];
+        List<Object[]> params = asList(new Object[][] {
+                {true, e(OB, OE, IB, IE), NIL, "sequential, no close"},
+                {true, e(OB, IB, OE, IE), NIL, "interleaved, no close"},
+                {true, e(OB, CI, OE, FC), NIL, "inbound close while writing"},
```
Scottmitch, line 147: @idelpivnitskiy - I duplicated the tests for now bcz not all states/verification are applicable and `RequestResponseCloseHandlerTest`'s usage of `spy` means the class cannot be final.

Note that there are a bunch of test cases where the inbound/outbound close event occurs that we have an opportunity to enhance. lets chat offline.

Scottmitch:

PR #1349, opened 2021-02-04
Remove `content-length` when `transfer-encoding: chunked` is present

idelpivnitskiy:
Motivation:

If a message is received with both a Transfer-Encoding and a
Content-Length header field, the Transfer-Encoding overrides the
Content-Length.  Such a message might indicate an attempt to
perform request smuggling (Section 9.5) or response splitting
(Section 9.4) and ought to be handled as an error.  A sender MUST
remove the received Content-Length field prior to forwarding such
a message downstream.

See https://tools.ietf.org/html/rfc7230#section-3.3.3

Modifications:

- Remove `content-length` header when `transfer-encoding: chunked`
is present in incoming (`HttpObjectDecoder`) and outgoing
(`HttpObjectEncoder`) messages;
- Add tests to verify the behavior;

Result:

HTTP/1.1 codec follow RFC7230 3.3.3 item 3: removes Content-Length
header when "Transfer-Encoding: chunked" is present.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
@@ -671,12 +671,19 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
             return null;
         }
 
+        final long contentLength = contentLength();
         if (isContentAlwaysEmpty(message)) {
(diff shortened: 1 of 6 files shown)


idelpivnitskiy:
Depends on #1348, review only the 2nd commit: https://github.com/apple/servicetalk/pull/1349/commits/af1b6eb194be894e30becbd22aff5c3ace6f6e5a

tkountis (approved):

bondolo (approved):

PR #1350, opened 2021-02-04
Reject HTTP messages with malformed `content-length` value(s)

idelpivnitskiy:
Motivation:

HTTP/1.1 tries to prevent messages with more than one `content-length`
header values. The recipient must either reject the message or drop
duplicate values. See https://tools.ietf.org/html/rfc7230#section-3.3.2

Because we translate HTTP/2 messages into HTTP/1.1 API, we should follow
the same principles. In addition, HTTP/2 RFC7540 says: A request or response
is also malformed if the value of a content-length header field does not
equal the sum of the DATA frame payload lengths that form the body.
See https://tools.ietf.org/html/rfc7540#section-8.1.2.6

Modifications:

- Add `HeaderUtils#contentLength` utility which guard against multiple
`Content-Length` headers;
- Use the new utility in `HttpObjectDecoder` and `AbstractH2DuplexHandler`;
- Calculate and verify actual DATA with `content-length` header value in
`AbstractH2DuplexHandler`;
- Add tests to verify new behavior;

Result:

ST rejects HTTP messages with malformed `content-length` value(s).

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019-2020 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019-2021 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


idelpivnitskiy:
Should be merged only after #1348, review only the 2nd commit: https://github.com/apple/servicetalk/pull/1350/commits/7ded438eb1294935d7403f178c6eff81c429af13

Scottmitch (approved):

tkountis (approved):

bondolo (approved):

PR #1351, opened 2021-02-04
Debug-log when `RoundRobinLoadBalancer` closes connections to inactive addresses

idelpivnitskiy:
Motivation:

When RRLB receives an inactive event from `ServiceDiscoverer`, it initiates
graceful closure of all connections to the host that became inactive. Give
some visibility to this logic.

Modifications:

- Add `DEBUG` log statement when RRLB closes connections to inactive host;

Result:

Users can debug when RRLB initiates connection closure.


bondolo (approved):

idelpivnitskiy:
Ok, PRB fails because of the flaky tests. This is just a log statement that I verified locally. Merging as-is.

PR #1353, opened 2021-02-05
Provide opt-out mechansim for HTTP/2 content-length enforcement

Scottmitch:
Motivation:
We recently started strictly enforcing the content-length requirements
as specified by https://tools.ietf.org/html/rfc7540#section-8.1.2.6.
However not all h2/gRPC implementations have been updated and may
violate this clause which results in rejected requests/responses.

Modifications:
- Add a system property io.servicetalk.http2.allowInvalidContentLength
  which allows opting out of enforcement of the content-length
  enforcement requirements.

Result:
System property exists to opt-out of content-length enforcement
requirements. This option is temporary and will be removed in a future
release!


idelpivnitskiy (approved):
Thanks!

PR #1359, opened 2021-02-08
Improve `H2PriorKnowledgeFeatureParityTest#trailersWithContentLength()`

idelpivnitskiy:
Motivation:

HTTP/1.x doesn't allow trailers with content-length, but it switches to
"chunked" encoding when trailers are present and removes content-length
header. We should make sure it behaves as expected.

Modifications:

- Enhance the test to verify headers in case of http/1.1 and http/2
protocols;
- Do the same verification for received request message on the
server-side;

Result:

More robust test.


Scottmitch (approved):
thanks!

PR #1365, opened 2021-02-10
Address various reactive-streams issues with content-coding codec

tkountis:
**Motivation:**
Various issues with reactive-streams spec and content-coding codec.
- Decoder onNext requests more data from the Subscription even if an element was delivered downstream. 
- Subscription is owned by the Subscriber of the operator chain. Any operators that pass the same subscription downstream and interact with the same subscription must apply concurrency controls around the Subscription
- Error propagation from onSubscribe should be allowed to reach the source which will act accordingly.
- Same for error propagation from onNext
- Cleanup of lifetime sensitive objects on subscription cancellation
- Improve encodings demand control to deliver compression footer and avoid relying on concat.

**Modifications:**
Issues listed in the motivation section were improved.

**Result:**
Improved solution that better adheres to reactive stream specification.

### DIFF
--- a/servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
+++ b/servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
@@ -18,7 +18,10 @@ package io.servicetalk.encoding.api;
 import io.servicetalk.buffer.api.Buffer;
 import io.servicetalk.buffer.api.BufferAllocator;
 import io.servicetalk.concurrent.PublisherSource;
+import io.servicetalk.concurrent.PublisherSource.Subscription;
 import io.servicetalk.concurrent.api.Publisher;
+import io.servicetalk.concurrent.internal.ConcurrentSubscription;
+import io.servicetalk.concurrent.internal.FlowControlUtils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
```diff
@@ -204,87 +128,343 @@ public final Buffer decode(final Buffer src, final int offset, final int length,
...
+                        subscription.request(n);
+                    } else if (demandUpdater.accumulateAndGet(EncodingOperator.this, n,
+                            FlowControlUtils::addWithOverflowProtectionIfNotNegative) > 0) {
+                        subscription.request(n);
+                    }
+                }
+
+                @Override
+                public void cancel() {
+                    subscription.cancel();
```
Scottmitch, line 187: consider using `try/finally` to ensure cleanup code runs (same comment for DecodingOperator)

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java
```diff
@@ -204,87 +128,343 @@ public final Buffer decode(final Buffer src, final int offset, final int length,
...
-                subscriber.onError(t);
+        @Override
+        public void onComplete() {
+            runSeriallyWithTermHandoff(() -> TERMINATED);
+            subscriber.onComplete();
+        }
+
+        // Execute a signal while maintaining ability to receive a cancellation handoff that will trigger the clean-up
+        // phase of the resources.
+        private void runSeriallyWithTermHandoff(final IntSupplier work) {
```
Scottmitch, line 434: `DecodingOperator` and `EncodingOperator` seem to be doing the same thing (e.g. manage state that has a lifetime and must be closed), with the exception `EncodingOperator` also needs to insert an `onNext` (which is done in `scanWith`). Can we share this code somehow?

Scottmitch (commented):
we should also add tests around edge conditions (e.g. no demand when onComplete arrives, throwing exceptions from subscriber, etc.)

tkountis:
Will rebase once we finalize work with https://github.com/apple/servicetalk/pull/1450

Scottmitch:
@tkountis - lets close this for now. some of this was addressed as part of the serialization/composite work, and any remaining fixes likely warrant a new PR at this point.

PR #1369, opened 2021-02-11
Add additional tests for checking correct validation of HTTP headers

bondolo:
Motiviation:
The existing tests for HTTP header validation do not cover some important header cases.
While adding the missing tests does not reveal any faults in the current header validation
it is important to include the tests for regression checking purposes.

Modification:
Additional tests for more invalid HTTP header cases are added as well as additional
documentation.

Result:
Regressions in HTTP header validation would be more likely detected.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -139,6 +139,9 @@
     private long chunkSize;
     private int cumulationIndex = -1;
     private long contentLength = Long.MIN_VALUE;
+    /**
+     * Line being being parsed. Used for reporting location in exceptions
```
idelpivnitskiy, line 143: doubled "being being"

idelpivnitskiy (approved):
Thank you!

PR #1370, opened 2021-02-11
`HttpCookiePair` impls should treat name and value case-sensitively

idelpivnitskiy:
Motivation:

RFC6265 implicitly defines that cookie names are case-sensitive. All
browsers treat cookie names as case-sensitive, see [1, 2].
Cookie values are always case-sensitive.

1. https://stackoverflow.com/questions/11311893/is-the-name-of-a-cookie-case-sensitive
2. https://github.com/playframework/playframework/issues/2241

Modifications:

- Change `equals` and `hashCode` impl of `DefaultHttpCookiePair` to compare
cookie name and value case-sensitively;
- Change `equals` and `hashCode` impl of `DefaultHttpSetCookie` to compare
cookie name case-sensitively, also include cookie value;
- Add tests to verify changes;

Result:

`HttpCookiePair` implementations correctly implement `equals` and `hashCode`.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
@@ -15,8 +15,6 @@
  */
 package io.servicetalk.http.api;
 
-import static io.servicetalk.buffer.api.CharSequences.caseInsensitiveHashCode;
-import static io.servicetalk.buffer.api.CharSequences.contentEqualsIgnoreCase;
 import static io.servicetalk.buffer.api.CharSequences.indexOf;
 import static io.servicetalk.http.api.HeaderUtils.validateCookieNameAndValue;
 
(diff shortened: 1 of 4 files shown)


tkountis:
Changes look good from a technical perspective, but I am not entirely sure about the expected behavior

Are these cookie values supposed to be equal or not?
`foo=bar;path=/;max-age=1`
`foo=bar;path=/;Max-Age=1`

According to the RFC (which doesn't specific casing for names), the attributes are case-insensitive.

> 5.2.2.  The Max-Age Attribute
> 
>    If the attribute-name case-insensitively matches the string "Max-
>    Age", the user agent MUST process the cookie-av as follows.

idelpivnitskiy:
@tkountis all attributes are case-insensitive, but `name` and `value` are case-sensitive. In your example, those two are equal.

bondolo (approved):

bondolo (approved):
No test changes needed?

tkountis (approved):

PR #1372, opened 2021-02-12
Add Netty epoll dependency for linux-aarch64

Scottmitch:
Motivation:
ServiceTalk includes implementation dependendencies on Netty's native
modules so users will get runtime dependencies for free. We don't yet
include the linux-aarch64 dependency for epoll.

Modifications:
- add the epoll linux-aarch64 dependency
- remove redundant native dependencies from packages that will bring in
  the dependencies transitively

Result:
linux-aarch64 linux use cases get native epoll transport for free.


Scottmitch:
test failure due to https://github.com/apple/servicetalk/issues/1192

idelpivnitskiy (approved):

PR #1373, opened 2021-02-12
Avoid "Maximum active streams violated for this endpoint" at the client level

idelpivnitskiy:
Motivation:

In #1307 we changed `LoadBalancedStreamingHttpClient` to not close the
connection on cancel for HTTP/2 protocol, because our existing API does
not expose methods for closing h2 stream instead of h2 connection.
Instead, we had to _prematurely_ mark the request as finished, because
in case of cancellation, there is no guarantee we will receive a terminal
signal on the response. Also, the stream is not aware of
`LoadBalancedStreamingHttpConnection` API and therefore it can not mark
the request as finished when it closes the stream.
That change significantly increased the probability users see
"Maximum active streams violated for this endpoint." Even though it's a
`RetryableException`, it creates pain and still may pop up after
auto-retry strategy used all attempts.

Modification:

- Introduce `OwnedRunnable` that can be passed to the transport layer
with the request object;
- Add pkg-private `StreamingHttpRequestWithContext` wrapper which
delivers `OwnedRunnable` to the transport;
- Generate that `OwnedRunnable` in `LoadBalancedStreamingHttpClient`;
- `LoadBalancedStreamingHttpClient` marks the request as finished only
if it owns `OwnedRunnable`;
- If h2-transport takes ownership of the `OwnedRunnable`,
`LoadBalancedStreamingHttpClient` is not responsible for marking the
request as finished anymore;
- Add tests to verify the fix;

Result:

HTTP/2 requests marked as finished only before they reach transport or
only after the stream closes.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentConnectionContext.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentConnectionContext.java
@@ -269,10 +269,7 @@ final class H2ClientParentConnectionContext extends H2ParentConnectionContext {
                         bs.open(promise);
                         sequentialCancellable = new SequentialCancellable(() -> promise.cancel(true));
                     } catch (Throwable cause) {
-                        if (ownedRunnable != null) {
-                            ownedRunnable.run();
-                        }
-                        observer.streamClosed(cause);
+                        cleanupWhenError(cause, observer, ownedRunnable);
                         deliverErrorFromSource(subscriber, cause);
                         return;
                     }
@@ -280,9 +277,7 @@ final class H2ClientParentConnectionContext extends H2ParentConnectionContext {
                     try {
                         subscriber.onSubscribe(sequentialCancellable);
                     } catch (Throwable cause) {
-                        if (ownedRunnable != null) {
-                            ownedRunnable.run();
-                        }
+                        cleanupErrorBeforeOpen(cause, promise, observer, ownedRunnable);
                         handleExceptionFromOnSubscribe(subscriber, cause);
                         return;
                     }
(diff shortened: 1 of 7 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentConnectionContext.java
```diff
@@ -310,6 +352,11 @@ private void childChannelActive(Future<Http2StreamChannel> future,
                             unexpected.addSuppressed(cause);
                             LOGGER.warn("Unexpected exception while handling the original cause", unexpected);
                         }
+                    } else {
+                        streamObserver.streamClosed(cause);
```
Scottmitch, line 356: nit: consider putting the `streamObserver` and `onCloseRunnable` error recovery code in a shared method (unlikely to be called, reduce byte code of containing method).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentConnectionContext.java
```diff
@@ -447,4 +498,21 @@ public void eventConsumed() {
...
+        private StacklessCancellationException() { }
+
+        // Override fillInStackTrace() so we not populate the backtrace via a native call and so leak the
+        // Classloader.
+        @Override
+        public Throwable fillInStackTrace() {
+            return this;
+        }
+
+        static StacklessCancellationException newInstance(Class<?> clazz, String method) {
```
Scottmitch, line 514: do we need to create the new exception type, or can we just use `CancellationException` and `ThrowableUtils.unknownStackTrace`?

Scottmitch (commented):
few comments then lgtm for near term workaround.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentConnectionContext.java
idelpivnitskiy: if we don't override `fillInStackTrace()`, the `Throwable` ctor will invoke it and fill in the stacktrace which `ThrowableUtils.unknownStackTrace` will override later => useless expensive operation.

PR #1375, opened 2021-02-17
Add TCP Fast Open support

Scottmitch:
Motivation:
Netty recently added support for
[TCP fast open](https://tools.ietf.org/html/rfc7413) on the client, and
has had server support for a while. We currently don't expose the
necessary configuration knobs to enable this feature, but it may be able
to reduce round trips and reduce connection setup latency.

Modifications:
- Add socket options to configure client and server TCP fast open.
- Add mechanism to configure the listen/acceptor socket options, and use
  this for the backlog option which was an outlier.

Result:
TCP Fast Open can be enabled for client and server. Note that it is only
supported by Netty's native EPOLL transport on linux, and only
initiated on the client when TLS is enabled.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
@@ -67,10 +67,12 @@ public abstract class GrpcServerBuilder {
     /**
      * The maximum queue length for incoming connection indications (a request to connect) is set to the backlog
      * parameter. If a connection indication arrives when the queue is full, the connection may time out.
-     *
+     * @deprecated Use {@link #listenSocketOption(SocketOption, Object)} with key
+     * {@link ServiceTalkSocketOptions#SO_BACKLOG}.
      * @param backlog the backlog to use when accepting connections.
      * @return {@code this}.
      */
+    @Deprecated
     public abstract GrpcServerBuilder backlog(int backlog);
 
     /**
@@ -101,6 +103,17 @@ public abstract class GrpcServerBuilder {
      * @see ServiceTalkSocketOptions
      */
     public abstract <T> GrpcServerBuilder socketOption(SocketOption<T> option, T value);
+
+    /**
+     * Adds a {@link SocketOption} that is applied to the server socket channel which listens/accepts socket channels.
+     * @param <T> the type of the value.
+     * @param option the option to apply.
+     * @param value the value.
+     * @return this.
+     * @see StandardSocketOptions
+     * @see ServiceTalkSocketOptions
+     */
+    public abstract <T> GrpcServerBuilder listenSocketOption(SocketOption<T> option, T value);
 
     /**
      * Enable wire-logging for this server.
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
@@ -73,7 +73,10 @@ public abstract class GrpcServerBuilder {
      * @return {@code this}.
      */
     @Deprecated
-    public abstract GrpcServerBuilder backlog(int backlog);
+    public GrpcServerBuilder backlog(int backlog) {
+        listenSocketOption(ServiceTalkSocketOptions.SO_BACKLOG, backlog);
+        return this;
+    }
 
     /**
      * Set the SSL/TLS configuration.
--- a/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
+++ b/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
@@ -63,13 +63,6 @@ final class DefaultGrpcServerBuilder extends GrpcServerBuilder implements Server
         return this;
     }
 
-    @Deprecated
-    @Override
-    public GrpcServerBuilder backlog(final int backlog) {
-        httpServerBuilder.backlog(backlog);
-        return this;
-    }
-
     @Override
     public GrpcServerBuilder sslConfig(final ServerSslConfig config) {
         httpServerBuilder.sslConfig(config);
--- a/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
+++ b/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
@@ -63,6 +63,7 @@ final class DefaultGrpcServerBuilder extends GrpcServerBuilder implements Server
         return this;
     }
 
+    @Deprecated
     @Override
     public GrpcServerBuilder backlog(final int backlog) {
         httpServerBuilder.backlog(backlog);
@@ -87,6 +88,13 @@ final class DefaultGrpcServerBuilder extends GrpcServerBuilder implements Server
         return this;
     }
 
+    @Override
+    public <T> GrpcServerBuilder listenSocketOption(final SocketOption<T> option, final T value) {
+        httpServerBuilder.listenSocketOption(option, value);
+        return this;
+    }
+
+    @Deprecated
     @Override
     public GrpcServerBuilder enableWireLogging(final String loggerName) {
         httpServerBuilder.enableWireLogging(loggerName);
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
@@ -51,6 +51,7 @@ final class DefaultHttpServerBuilder extends HttpServerBuilder {
         return this;
     }
 
+    @Deprecated
     @Override
     public HttpServerBuilder backlog(final int backlog) {
         config.tcpConfig().backlog(backlog);
@@ -72,6 +73,12 @@ final class DefaultHttpServerBuilder extends HttpServerBuilder {
     @Override
     public <T> HttpServerBuilder socketOption(final SocketOption<T> option, final T value) {
         config.tcpConfig().socketOption(option, value);
+        return this;
+    }
+
+    @Override
+    public <T> HttpServerBuilder listenSocketOption(final SocketOption<T> option, final T value) {
+        config.tcpConfig().listenSocketOption(option, value);
         return this;
     }
 
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
@@ -27,6 +27,7 @@ import io.servicetalk.transport.api.ConnectionAcceptor;
 import io.servicetalk.transport.api.IoExecutor;
 import io.servicetalk.transport.api.ServerContext;
 import io.servicetalk.transport.api.ServerSslConfig;
+import io.servicetalk.transport.api.ServiceTalkSocketOptions;
 import io.servicetalk.transport.api.TransportObserver;
 
 import java.net.SocketAddress;
@@ -51,10 +52,18 @@ final class DefaultHttpServerBuilder extends HttpServerBuilder {
         return this;
     }
 
+    /**
+     * Sets the maximum queue length for incoming connection indications (a request to connect) is set to the backlog
+     * parameter. If a connection indication arrives when the queue is full, the connection may time out.
+     *
+     * @deprecated Use {@link #listenSocketOption(SocketOption, Object)} with key
+     * {@link ServiceTalkSocketOptions#SO_BACKLOG}.
+     * @param backlog the backlog to use when accepting connections.
+     * @return {@code this}.
+     */
     @Deprecated
-    @Override
-    public HttpServerBuilder backlog(final int backlog) {
-        config.tcpConfig().backlog(backlog);
+    public HttpServerBuilder backlog(int backlog) {
+        listenSocketOption(ServiceTalkSocketOptions.SO_BACKLOG, backlog);
         return this;
     }
 
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MutualSslTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MutualSslTest.java
@@ -58,16 +58,16 @@ public class MutualSslTest {
     @Rule
     public final Timeout timeout = new ServiceTalkTestTimeout();
     @DataPoints("serverSslProvider")
-    public static final SslProvider[] serverSslProvider = new SslProvider[] {JDK, OPENSSL};
+    public static final SslProvider[] SERVER_PROVIDERS = {JDK, OPENSSL};
     @DataPoints("clientSslProvider")
-    public static final SslProvider[] clientSslProvider = new SslProvider[] {JDK, OPENSSL};
+    public static final SslProvider[] CLIENT_PROVIDERS = {JDK, OPENSSL};
     @DataPoints("serverListenOptions")
     @SuppressWarnings("rawtypes")
-    public static final List<Map<SocketOption, Object>> serverListenOptions =
+    public static final List<Map<SocketOption, Object>> SERVER_LISTEN_OPTIONS =
             asList(emptyMap(), serverTcpFastOpenOptions());
     @DataPoints("clientOptions")
     @SuppressWarnings("rawtypes")
-    public static final List<Map<SocketOption, Object>> clientOptions = asList(emptyMap(), clientTcpFastOpenOptions());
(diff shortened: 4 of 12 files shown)


gradle.properties
```diff
@@ -17,7 +17,7 @@
 group=io.servicetalk
 version=0.38.0-SNAPSHOT
 
-nettyVersion=4.1.59.Final
+nettyVersion=4.1.60.Final-SNAPSHOT
```
Scottmitch, line 20: build is expected to fail until Netty release with the new option is available.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
```diff
@@ -67,10 +67,12 @@
...
      * The maximum queue length for incoming connection indications (a request to connect) is set to the backlog
      * parameter. If a connection indication arrives when the queue is full, the connection may time out.
-     *
+     * @deprecated Use {@link #listenSocketOption(SocketOption, Object)} with key
+     * {@link ServiceTalkSocketOptions#SO_BACKLOG}.
      * @param backlog the backlog to use when accepting connections.
      * @return {@code this}.
      */
+    @Deprecated
     public abstract GrpcServerBuilder backlog(int backlog);
```
bondolo, line 76: perhaps implement the method by implementing the suggested replacement, 

```
listenSocketOption(ServiceTalkSocketOptions.SO_BACKLOG, backlog);
return this;
```

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
```diff
@@ -63,6 +63,7 @@ public GrpcServerBuilder protocols(final HttpProtocolConfig... protocols) {
         return this;
     }
 
+    @Deprecated
     @Override
     public GrpcServerBuilder backlog(final int backlog) {
         httpServerBuilder.backlog(backlog);
```
bondolo, line 69: Why is this not using the approach suggested in the abstract class?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -50,6 +50,7 @@ public HttpServerBuilder protocols(final HttpProtocolConfig... protocols) {
         return this;
     }
 
+    @Deprecated
     @Override
     public HttpServerBuilder backlog(final int backlog) {
         config.tcpConfig().backlog(backlog);
```
bondolo, line 56: same comment as outer layers, why not use listenSocketOption?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MutualSslTest.java
```diff
@@ -44,32 +54,58 @@
...
 
-    @Parameterized.Parameters(name = "server={0} client={1}")
+    @Parameterized.Parameters(name = "server={0} client={1} server opts={2} client opts={3}")
     public static Collection<Object[]> sslProviders() {
         return asList(
-                new Object[]{JDK, JDK},
-                new Object[]{JDK, OPENSSL},
-                new Object[]{OPENSSL, JDK},
-                new Object[]{OPENSSL, OPENSSL}
+                new Object[]{JDK, JDK, emptyMap(), emptyMap()},
```
bondolo, line 74: Does this assume that fast open will always be available in the test environment?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MutualSslTest.java
```diff
@@ -44,32 +54,58 @@
...
+        HttpServerBuilder serverBuilder = HttpServers.forAddress(localAddress(0))
                 .secure()
                 .provider(serverSslProvider)
                 .clientAuth(REQUIRE)
                 .trustManager(DefaultTestCerts::loadClientCAPem)
-                .commit(DefaultTestCerts::loadServerPem, DefaultTestCerts::loadServerKey)
-                .listenBlockingAndAwait((ctx, request, responseFactory) -> responseFactory.ok());
-             BlockingHttpClient client = HttpClients.forSingleAddress(serverHostAndPort(serverContext))
+                .commit(DefaultTestCerts::loadServerPem, DefaultTestCerts::loadServerKey);
+        for (@SuppressWarnings("rawtypes") Entry<SocketOption, Object> entry : serverListenOptions.entrySet()) {
```
bondolo, line 101: SocketOption<?> perhaps? 
Sadly you can't introduce a type variable here to say <T> Entry<SocketOption<T>, ? extends T>

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
Scottmitch: I will use your suggestion above todo this in the base class. the base method will be removed in future releases.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MutualSslTest.java
Scottmitch: no it won't be but the motivation for testing all the combinations is that even if fast open isn't supported it asking for it shouldn't break anything (provides an opportunity for optimization).

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MutualSslTest.java
Scottmitch: Then the generic types don't line up for `builder.socketOption(option, entry.getValue());` which causes issues.

> reason: no instance(s) of type variable(s) exist so that Object conforms to capture of ? inference variable T has incompatible bounds: equality constraints: capture of ? lower bounds: Object

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MutualSslTest.java
bondolo: OK, as long as it isn't required for successful testing.

bondolo:
Should include enhancements to this docs section: https://apple.github.io/servicetalk//servicetalk/SNAPSHOT/performance.html#socket-and-transport-options

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/TcpFastOpenTest.java
```diff
@@ -0,0 +1,109 @@
...
+        return singletonMap(TCP_FASTOPEN_CONNECT, true);
+    }
+
+    @SuppressWarnings("rawtypes")
+    static Map<SocketOption, Object> serverTcpFastOpenOptions() {
+        return singletonMap(TCP_FASTOPEN_BACKLOG, 1);
+    }
+
+    @Test
+    public void mutualSsl() throws Exception {
```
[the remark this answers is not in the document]
Scottmitch: good catch. cpy/pasta.

> Just to verify that setting fast-open options doesn't break anything? 
yes. I've updated the test name to clarify.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServiceTalkSocketOptions.java
```diff
@@ -41,6 +42,41 @@
      */
     public static final SocketOption<Long> IDLE_TIMEOUT = new ServiceTalkSocketOption<>("IDLE_TIMEOUT", Long.class);
 
+    /**
+     * The number of pending accepted connections for server sockets. For example this value is used for methods like
+     * {@link java.nio.channels.ServerSocketChannel#bind(SocketAddress, int)} and
+     * <a href="https://man7.org/linux/man-pages/man2/listen.2.html">listen(int sockfd, int backlog)</a>.
+     */
+    public static final SocketOption<Integer> SO_BACKLOG = new ServiceTalkSocketOption<>("SO_BACKLOG", Integer.class);
```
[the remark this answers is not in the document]
Scottmitch: I'm hesitant to fragment the options at this point or try to create a strongly-type hierarchy for server options as there is some potential overlap (sndbuf, recbuf, etc.). The JDK's `StandardSocketOption` also has "listen" specific options (e.g.`SO_REUSEADDR` and `SO_REUSEPORT`) and it is difficult to build a type safe hierarchy when options are also dependent upon the underlying transport/protocol being used. Moving the options into another location (e.g. `ServiceTalkListenSocketOptions`) may trigger folks to think-twice on first use (e.g. hunt for other options) but wouldn't provide much/any additional safety (due to type overlap). I'll start by grouping them in the same file, and we can discuss if more should be done.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServiceTalkSocketOptions.java
```diff
@@ -42,31 +44,15 @@
...
      * Enable TCP fast open connect on the client side as described in
      * <a href="https://tools.ietf.org/html/rfc7413#appendix-A.1">RFC 7413 Active Open</a>.
      * <p>
      * Note the following caveats of this option:
      * <ul>
-     *     <li>it may not be supported by the underlying transport (e.g. currently only supported by Netty's native
-     *     EPOLL transport)</li>
+     *     <li>it may not be supported by the underlying transport (e.g. supported by Netty's
+     *     <a href="https://netty.io/wiki/native-transports.html#using-the-linux-native-transport">linux EPOLL
+     *     transport</a></li>
```
idelpivnitskiy, line 55: Forgot the closing `)`

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServiceTalkSocketOptions.java
```diff
@@ -77,6 +63,26 @@
...
+     */
+    public static final SocketOption<Integer> SO_BACKLOG = new ServiceTalkSocketOption<>("SO_BACKLOG", Integer.class);
+
+    /**
+     * The number of pending SYNs with data payload for server sockets as described in
+     * <a href="https://tools.ietf.org/html/rfc7413#appendix-A.2">RFC 7413 Passive Open</a>.
+     * <p>
+     * Note this option may not be supported by the underlying transport (e.g. supported by Netty's
+     * <a href="https://netty.io/wiki/native-transports.html#using-the-linux-native-transport">linux EPOLL
+     * transport</a>.
```
idelpivnitskiy, line 81: Forgot the closing `)`

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SocketOptionUtils.java
```diff
@@ -162,4 +112,32 @@ private SocketOptionUtils() {
...
+
+    @Nullable
+    private static Boolean boolNot(@Nullable Boolean v) {
+        return v == null ? null : !v;
+    }
+
+    private static final class OptConverter<T> {
+        private final ChannelOption<T> option;
+        private final Function<T, Object> chanToSock;
+        private final Function<Object, T> sockToChan;
```
idelpivnitskiy, line 133: Consider `toSocketVal`, `toChannelVal` names

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SocketOptionUtils.java
```diff
@@ -24,13 +24,39 @@
...
+
 /**
  * Utilities to convert {@link SocketOption}s.
  */
 public final class SocketOptionUtils {
+    private static final Map<SocketOption<?>, OptConverter<?>> SOCKET_OPT_MAP = new HashMap<>();
+    static {
+        putOpt(ChannelOption.IP_MULTICAST_IF, StandardSocketOptions.IP_MULTICAST_IF);
+        putOpt(ChannelOption.IP_MULTICAST_LOOP_DISABLED, StandardSocketOptions.IP_MULTICAST_LOOP,
+                SocketOptionUtils::boolNot, SocketOptionUtils::boolNot);
```
idelpivnitskiy, line 42: socket to channel conversion should have a null check for the passed value to fail at the builder vs when a channel is created.

idelpivnitskiy (approved):
LGTM after the build passes, thanks!

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/TcpFastOpenTest.java
```diff
@@ -0,0 +1,109 @@
...
+        return singletonMap(TCP_FASTOPEN_CONNECT, true);
+    }
+
+    @SuppressWarnings("rawtypes")
+    static Map<SocketOption, Object> serverTcpFastOpenOptions() {
+        return singletonMap(TCP_FASTOPEN_BACKLOG, 1);
+    }
+
+    @Test
+    public void requestSucceedsEvenIfTcpFastOpenNotEnabledOrSupported() throws Exception {
```
idelpivnitskiy, line 84: Btw, this test proves that it's safe to have this option always enabled. WDYT about making it on by default (opt-out)?

Scottmitch: Netty logs warnings if you enable an option that isn't supported, and also if it is supported users may not want to have it always enabled (e.g. idempotent, extra state tracking, etc.).

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SocketOptionUtils.java
Scottmitch: I don't think we need to be restrictive/opinionated here. Netty treats `null` as "remove option" and if we utilize that in the future this code wouldn't need to change, and the overhead for null check is minimal vs cognitive overhead to understand why they need to be different.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/TcpFastOpenTest.java
idelpivnitskiy: Thanks for the explanation. What if we move it down to the transport, where we have access to `Epoll.isAvailable()` check (or `Channel` type) and know that the `Channel` is secure? For example, `SslClientChannelInitializer` might be a good place, if we pass the user-defined options there as well (to check that user did not explicitly set this option).

Scottmitch: Interesting idea. I propose we proceed by adding the feature as opt-in first, and after we get some data we can consider enabling by default in a followup PR.

idelpivnitskiy: Sounds good to me!

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MutualSslTest.java
```diff
@@ -44,32 +56,57 @@
...
-                new Object[]{OPENSSL, OPENSSL}
-        );
+        final SslProvider[] providers = new SslProvider[] {JDK, OPENSSL};
+        @SuppressWarnings("rawtypes")
+        final List<Map<SocketOption, Object>> serverOpts = asList(emptyMap(), serverTcpFastOpenOptions());
+        @SuppressWarnings("rawtypes")
+        final List<Map<SocketOption, Object>> clientOpts = asList(emptyMap(), clientTcpFastOpenOptions());
+        final List<Object[]> results = new ArrayList<>(
+                providers.length * 2 * serverOpts.size() * clientOpts.size());
+        for (SslProvider serverProvider : providers) {
```
tkountis, line 82: This is pretty ugly, you could probably replace this with junit's `@Theory` & `@Datapoints`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/TcpFastOpenTest.java
tkountis: Reading around privacy concerns with TFO [1], makes me think that it's probably better left as opt-in. 
1. https://arxiv.org/pdf/1905.03518.pdf

tkountis (approved):
LGTM

bondolo (commented):
LGTM, but I don't feel confident to approve.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MutualSslTest.java
Scottmitch: thx for suggesting! done.

Scottmitch:
depends upon https://github.com/netty/netty/pull/11077

gradle.properties
```diff
@@ -17,7 +17,7 @@
 group=io.servicetalk
 version=0.39.0-SNAPSHOT
 
-nettyVersion=4.1.60.Final
+nettyVersion=4.1.61.Final-SNAPSHOT
```
Scottmitch, line 20: reminder: revert change when 4.1.61 is released

PR #1376, opened 2021-02-18
Introduce compression based on Netty's codecs

tkountis:
Motivation:
Netty has compression codecs for the majority of support codings (deflate, gzip, snappy). Instead of duplicating the effort we could rely on those, to benefit from single code base and improvements.

Modifications:
A new module was introduced servicetalk-encoding-netty, which now offers implementations based on Netty codecs.
All previous solutions are now deprecated, pointing to the alternative builders.

Result:
Single codebase.


tkountis:
Build failure due to https://github.com/apple/servicetalk/issues/1224

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/ContentCodecBuilder.java
```diff
@@ -29,7 +29,7 @@
      * @return {@code this}
      * @see <a href="https://en.wikipedia.org/wiki/Zip_bomb">Zip Bomb</a>
      */
-    ContentCodecBuilder setMaxAllowedPayloadSize(int maxAllowedPayloadSize);
+    ContentCodecBuilder maxAllowedPayloadSize(int maxAllowedPayloadSize);
```
tkountis, line 32: This is an API breaking change, but I am not anticipating anyone using this. If any objections, we can gracefully deprecate it.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/DefaultContentCodecBuilder.java
```diff
@@ -0,0 +1,114 @@
...
+         */
+        public SnappyContentCodecBuilder validateChecksums(boolean validateChecksums) {
+            this.validateChecksums = validateChecksums;
+            return this;
+        }
+
+        @Override
+        protected int maxAllowedPayloadSize() {
+            throw new UnsupportedOperationException("Max allowed payload size is not supported for snappy. " +
+                    "The maximum chunk size of snappy is " + Short.MAX_VALUE);
```
bondolo, line 105: Snappy 1.1 allows 64kb. I don't see it mentioned which version Netty supports.

tkountis: Good catch. I was referring to Netty's internal max encoding chunk size (https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/compression/SnappyFrameEncoder.java#L70) but its irrelevant to decompression. I will remove that sentence, we shouldn't rely on internals anyway.

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/ContentCodecBuilder.java
tkountis: On a similar note, this API has some mismatches with the Netty's meaning.
For Netty max translates to `max-allocation`, and its tracked as the internal buffer used to decode _one_ message. In our current servicetalk-encoding-api based solution, that was used to actually track the max payload size (ie. accumulated https://github.com/apple/servicetalk/blob/main/servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractZipContentCodec.java#L382).
Regardless of the API mismatch, which we can address by deprecating the old flavor, we should decide if Netty's approach is enough for our needs. From my perspective, I think we are fine, because a single chunk can't be larger than `max-allocation`, and the number of chunks can be controlled by a subscription's demand.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/DefaultContentCodecBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public abstract static class ZipContentCodecBuilder extends DefaultContentCodecBuilder {
+
+        private int compressionLevel = 6;
+
+        /**
+         * Sets the compression level for this codec's encoder.
+         * @param compressionLevel 1 yields the fastest compression and 9 yields the best compression.
+         * 0 means no compression. The default compression level is 6
```
idelpivnitskiy, line 54: We usually do not document default values to give us some flexibility for changing them in future, consider removing this detail until we reach 1.0.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/DefaultContentCodecBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+         * @return <code>this</code>
```
idelpivnitskiy, line 55: Consider using `{@code this}` instead of html tags

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/ContentCodings.java
```diff
@@ -43,7 +43,9 @@ public static ContentCodec identity() {
     /**
      * Returns the default GZIP {@link ContentCodec}.
      * @return default GZIP based {@link ContentCodec}
+     * @deprecated API replaced by <code>io.servicetalk.encoding.netty.ContentCodings#gzipDefault()</code>
```
idelpivnitskiy, line 46: Consider using `{@code ...}` instead of html tags

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/DefaultContentCodecBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+         */
+        public ZipContentCodecBuilder withCompressionLevel(final int compressionLevel) {
+            this.compressionLevel = compressionLevel;
+            return this;
+        }
+
+        protected int compressionLevel() {
```
idelpivnitskiy, line 62: All `public` and `protected` methods should have javadoc. Looks like checkstyle rules require some attention

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/DefaultContentCodecBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+                    () -> new JdkZlibEncoder(ZlibWrapper.GZIP, compressionLevel()),
+                    () -> new JdkZlibDecoder(ZlibWrapper.GZIP, maxAllowedPayloadSize())
+            );
+        }
+    }
+
+    static final class DeflateContentCodecBuilder extends ZipContentCodecBuilder {
+        @Override
+        public ContentCodec build() {
+            return new NettyChannelContentCodec("deflate",
```
idelpivnitskiy, line 80: Consider using `newAsciiString` as described in #1385.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+
+    @Override
+    public Buffer encode(final Buffer src, final int offset, final int length,
+                         final BufferAllocator allocator) {
+        requireNonNull(src);
+        if (offset < 0) {
+            throw new IllegalArgumentException("Invalid offset: " + offset + " (expected >= 0)");
+        }
+
+        if (length <= 0) {
```
idelpivnitskiy, line 66: Should we validate that the `length` is not greater than `src.readableBytes()`? Here and in `decode` method.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+
+    private final Supplier<MessageToByteEncoder<ByteBuf>> encoderFn;
+    private final Supplier<ByteToMessageDecoder> decoderFn;
+
+    NettyChannelContentCodec(final CharSequence name,
+                             final Supplier<MessageToByteEncoder<ByteBuf>> encoderFn,
+                             final Supplier<ByteToMessageDecoder> decoderFn) {
+        super(name);
+        this.encoderFn = encoderFn;
+        this.decoderFn = decoderFn;
```
idelpivnitskiy, line 55: Both are `Supplier`s, not funtions. Consider naming them as factories or suppliers.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+    private static ByteBuf toByteBufRetained(final Buffer buffer, final int offset, final int length) {
+        return toByteBuf(
+                buffer.readerIndex(buffer.readerIndex() + offset)
+                      .readSlice(length))
+                      .retain();
+    }
+
+    private static void safeClose(final EmbeddedChannel channel) {
+        try {
+            channel.finish();
```
idelpivnitskiy, line 252: Consider adding assertions before `finish()` to validate that inbound and outbound messages are empty.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+                compositeBuffer.addBuffer(newBufferFromRetained(chunk));
+            }
+
+            return compositeBuffer;
+        } catch (Exception e) {
+            LOGGER.error("Error while encoding with {}", name(), e);
+            throwException(e);
+            return EMPTY_BUFFER;
+        } finally {
+            safeClose(channel);
```
idelpivnitskiy, line 92: Do we need `safeClose` in `finally` if we already closed inside `try`? Can we do this only inside `catch` block? If you will move it to the `catch` block only, do we need `channel.checkException()` after closing the channel inside `try` block? Or maybe just invoke `finish` there?

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+            channel.close();
+
+            CompositeBuffer compositeBuffer = allocator.newCompositeBuffer();
+            ByteBuf chunk;
+            while ((chunk = channel.readOutbound()) != null) {
+                compositeBuffer.addBuffer(newBufferFromRetained(chunk));
+            }
+
+            return compositeBuffer;
+        } catch (Exception e) {
```
idelpivnitskiy, line 87: Should we catch `Throwable`?

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+        if (offset < 0) {
+            throw new IllegalArgumentException("Invalid offset: " + offset + " (expected >= 0)");
+        }
+
+        if (length <= 0) {
+            throw new IllegalArgumentException("Invalid length: " + length + " (expected > 0)");
+        }
+
+        final MessageToByteEncoder<ByteBuf> encoder = encoderFn.get();
+        final EmbeddedChannel channel = new EmbeddedChannel(encoder);
```
idelpivnitskiy, line 71: By default, the `EmbeddedChannel` is created with netty's `ByteBufAllocator` and `RecvByteBufAllocator`. Both will use pooled memory that we may never release. We should override both to use  the passed allocator which have to be a `ServiceTalkBufferAllocator`. You can do that using `channel.config()`.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+
+        try {
+            ByteBuf origin = toByteBufRetained(src, offset, length);
+            channel.writeOutbound(origin);
+
+            // May produce footer
+            channel.close();
```
idelpivnitskiy, line 78: `close()` returns `ChannelFuture`, should we sync it and wait until it completes? I do not expect any real blocking code here.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+
+            CompositeBuffer compositeBuffer = allocator.newCompositeBuffer();
```
idelpivnitskiy, line 80: Do these codecs always produce 2+ chunks? Does it make sense to check the size of `channel.outboundMessages()` and if it's equalt to 1, skip the `CompositeBuffer`?

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+                            if (part != null) {
+                                subscriber.onNext(newBufferFromRetained(part));
+                            }
+
+                            return;
+                        }
+
+                        channel.writeOutbound(toByteBufRetained(next));
+                        ByteBuf part = channel.readOutbound();
+                        if (part != null) {
```
idelpivnitskiy, line 133: Here is a risk that some codecs may produce more than one outbound `ByteBuf`. We should draint it until the outbound messages queue is empty here and for `END_OF_STREAM`.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+                    @Override
+                    public void onError(Throwable t) {
+                        safeClose(channel);
+                        subscriber.onError(t);
+                    }
+
+                    @Override
+                    public void onComplete() {
+                        safeClose(channel);
+                        subscriber.onComplete();
```
idelpivnitskiy, line 147: Should we produce `subscriber.onError` signal if the channel closes with an exception instead of ignoring it?

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+                    }
+                });
+    }
+
+    @Override
+    public Buffer decode(final Buffer src, final int offset, final int length, final BufferAllocator __) {
```
idelpivnitskiy, line 153: allocator should be used for any memory we allocate here, including things allocated by `EmbeddedChannel`

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    @Override
+    public Publisher<Buffer> encode(final Publisher<Buffer> from,
+                                          final BufferAllocator __) {
+        requireNonNull(from);
+        return from
+                .concat(succeeded(END_OF_STREAM))
+                .liftSync(subscriber -> new PublisherSource.Subscriber<Buffer>() {
```
idelpivnitskiy, line 102: Can we use a new `scanWith` operator? Does it miss anything?

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+            }
+        });
+    }
+
+    private static Buffer newBufferFromRetained(final ByteBuf buf) {
+        return newBufferFrom(buf.readRetainedSlice(buf.readableBytes()));
+    }
+
+    private static ByteBuf toByteBufRetained(final Buffer buffer) {
+        return toByteBuf(buffer).retain();
```
idelpivnitskiy, line 240: ST Buffers are always retained, unreleasable, and `retain()` is noop. Consider just using `toByteBuf` directly.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+            @Override
+            public void onComplete() {
+                safeClose(channel);
+                subscriber.onComplete();
+            }
+        });
+    }
+
+    private static Buffer newBufferFromRetained(final ByteBuf buf) {
+        return newBufferFrom(buf.readRetainedSlice(buf.readableBytes()));
```
idelpivnitskiy, line 236: If you configure `EmbeddedChannel` to use ST allocator, you don't need to retain `ByteBuf` it returns, just use `newBufferFrom`.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+        try {
+            ByteBuf origin = toByteBufRetained(src, offset, length);
+            channel.writeOutbound(origin);
+
+            // May produce footer
+            channel.close();
+
+            CompositeBuffer compositeBuffer = allocator.newCompositeBuffer();
+            ByteBuf chunk;
+            while ((chunk = channel.readOutbound()) != null) {
```
idelpivnitskiy, line 82: Sad that `EmbeddedChannel` always allocates a default `ArrayDequeu` with 16 elements for inbound/outbound messages :(

idelpivnitskiy (changes requested):
Simplified code looks amazing!

I have some comments :) Some comments for `encode` methods are also relevant for `decode`, I did not repeat them.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/AbstractContentCodec.java
```diff
@@ -0,0 +1,57 @@
...
+    public final CharSequence name() {
+        return name;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
```
idelpivnitskiy, line 39: To capture @bondolo comment here, see https://github.com/apple/servicetalk/pull/1383/files#r583080915:

> comparing the exact class and instanceof is not quite the same. The existing implementation would declare different sub-classes to be non-equal where this implementation might declare them equivalent. There is also the danger of non-reflexive equals() if one of the sub-classes declares an equals() method that matches specifically against their own class.
I would encourage you to leave it getClass() != o.getClass() unless you can make the equals() method final.

@tkountis IIUC, for the purpose of content codecs it doesn't matter if 2 codecs have the same class or not. the only `name` matters. the `instanceof` probably makes more sense here, but should we make this `equals` and `hashCode` impl `final`?

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
tkountis: Lets keep this PR related to EmbeddedChannel work, I have another ongoing item which will rely on a new operator `scanWithLifetime` 
for reference. https://github.com/apple/servicetalk/pull/1365

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/AbstractContentCodec.java
tkountis: Correct. Made them final and changed it to instanceof

servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyChannelContentCodecTest.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Test
+    public void testEncode_checkAccessibility() {
+        Buffer source = DEFAULT_ALLOCATOR.fromAscii(INPUT);
+        Buffer encoded = codec.encode(source, DEFAULT_ALLOCATOR);
+
+        // We own this, reset the index and try to decode again.
+        // If the reference wasn't retained this could fail
+        source.readerIndex(0);
```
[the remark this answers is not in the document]
tkountis: Removed since we now use ST allocators the test isn't offering any benefit.

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/ContentCodecBuilder.java
```diff
@@ -29,7 +29,7 @@
      * @return {@code this}
      * @see <a href="https://en.wikipedia.org/wiki/Zip_bomb">Zip Bomb</a>
      */
-    ContentCodecBuilder setMaxAllowedPayloadSize(int maxAllowedPayloadSize);
```
Scottmitch, line 32: (not necessarily to fix in the PR, just general comment)
having a builder interface in the `api` package isn't unique to this case, but I think we can simplify in the future by following this pattern:
- interface in the API package (e.g. `ContentCodec`)
- builder implementation closes to implementation as possible (e.g. if the builder impl is generic and needs to be used across multiple package `api` maybe good `*SslConfigBuilder`, but if it is tied to `netty` it can live in the corresponding `netty` package)

Having a builder interface doesn't really provide much value, builders are typically tied to an implementation and don't need to share an interface (where as the thing they build does need to share a common interface).

https://github.com/apple/servicetalk/pull/1376/files#r579151766
> this API has some mismatches with Netty's meaning.

this type of mis-match would be easier to correct the closer the builder is to the implementation.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
Scottmitch: +1, maybe better to just copy into a single buffer if there is an optional small footer. `CompositeBuffer` can requires some additional work to write and may limit how much data we can pass to `writev` (max of 1024 pointers), so if there are small chunks it can be better to combine them.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,257 @@
...
+            throwException(e);
+            return EMPTY_BUFFER;
+        } finally {
+            safeClose(channel);
+        }
+    }
+
+    @Override
+    public Publisher<Buffer> encode(final Publisher<Buffer> from,
+                                          final BufferAllocator __) {
```
Scottmitch, line 98: nit: just move to previous line

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/ContentCodecBuilder.java
bondolo: Agreed that if the interface exposes Netty semantics or is a veneer on Netty APIs then it should be in the Netty package. If it is in the api package then there is a temptation to over-abstract and create these kind of mismatches where there no actual intent to ever have more than the Netty implementation.

tkountis: Yup agreed. We started with the impl being in the API pkg but now this becomes more visible. I am deprecating the old API so we can keep them colocated.

idelpivnitskiy: agreed, +1

servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyChannelContentCodecTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Test
+    public void testEncode() {
+        Buffer source = DEFAULT_ALLOCATOR.fromAscii(INPUT);
+        Buffer encoded = codec.encode(source, DEFAULT_ALLOCATOR);
+        // All codecs produce deflated buffers of less than 100 bytes
+        // ContentCodec{name=gzip} 25
+        // ContentCodec{name=deflate} 17
+        // ContentCodec{name=snappy} 70
+        assertThat(encoded.readableBytes(), lessThan(100));
```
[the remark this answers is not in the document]
tkountis: We have integration tests, that will verify behavior.

servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyChannelContentCodecTest.java
```diff
@@ -0,0 +1,121 @@
...
+        Buffer decoded = codec.decode(encoded, DEFAULT_ALLOCATOR);
+        assertThat(decoded.toString(US_ASCII), equalTo(INPUT));
```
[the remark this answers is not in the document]
tkountis: We have integration tests, that will verify behavior

servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyChannelContentCodecTest.java
```diff
@@ -0,0 +1,121 @@
...
+        Buffer encoded = codec.encode(Publisher.from(source), DEFAULT_ALLOCATOR)
+                .collect(DEFAULT_ALLOCATOR::newCompositeBuffer, CompositeBuffer::addBuffer).toFuture().get();
+        // All codecs produce deflated buffers of less than 100 bytes
+        // ContentCodec{name=gzip} 25
+        // ContentCodec{name=deflate} 17
+        // ContentCodec{name=snappy} 70
+        assertThat(encoded.readableBytes(), lessThan(100));
+
+        Buffer decoded = codec.decode(Publisher.from(encoded), DEFAULT_ALLOCATOR).firstOrError().toFuture().get();
+        assertThat(decoded.toString(US_ASCII), equalTo(INPUT));
```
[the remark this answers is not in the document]
tkountis: We have integration tests, that will verify behavior

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
tkountis: Interesting info. We can certainly optimize later. For now I followed your suggestion for cases where the accumulated size is less than 1KiB. The typical case will be just a few bytes for footer. Let me know if you thing we should increase that threshold, but I would rather benchmark it and follow up with optimizations if needed.

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/ContentCodings.java
```diff
@@ -53,15 +55,19 @@ public static ContentCodec gzipDefault() {
      * a customizable {@link ContentCodec}.
      * @return a GZIP based {@link ContentCodecBuilder} that allows building
      *          a customizable GZIP {@link ContentCodec}
+     * @deprecated API replaced by <code>io.servicetalk.encoding.netty.ContentCodings#gzip()</code>
```
bondolo, line 58: There are few more case of <code> which could be converted to {@link …} or {@code …}

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/ContentCodings.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private ContentCodings() {
+    }
+
+    /**
+     * Returns the default, always supported 'identity' {@link ContentCodec}.
+     * @return the default, always supported 'identity' {@link ContentCodec}
+     */
+    public static ContentCodec identity() {
+        return io.servicetalk.encoding.api.ContentCodings.identity();
```
bondolo, line 38: If io.servicetalk.encoding.api.ContentCodings is to be deprecated perhaps the indentity constant there should reference this one rather than the other way around. That way when it is removed no changes are needed here.

Scottmitch: We typically don't have `api` depend upon `netty` modules (to avoid leaking Netty into API), but is the intention to move the `IdentityContentCodec` into the `netty` package so we can completely remove `io.servicetalk.encoding.api.ContentCodings`? If so lets mark this class as deprecated too.

tkountis: The intention was to keep this one around because it's not bound to an implementation (ie. Netty) and its used in a few places as ref-check to skip coding headers and/or skip encoding/decoding see. `io.servicetalk.encoding.api.internal.HeaderUtils`
So the goal was to only deprecate the actual implementation of ` io.servicetalk.encoding.api.ContentCodings` but not `identity`.

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/ContentCodings.java
tkountis: Fixed. I kept everything as `{@code ...}` because we are linking classes that aren't visible in the api module.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
Scottmitch: we can optimize this in a followup, and even create a customized channel that writes directly to our desired destination instead of intermediate queue.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/ContentCodings.java
Scottmitch: Instead of exposing a new `API` and specialized code (e.g. single element in collection, etc.) can we just use `null` for this?

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/AbstractContentCodec.java
```diff
@@ -0,0 +1,59 @@
...
+import io.servicetalk.encoding.api.ContentCodec;
+
+import static io.servicetalk.buffer.api.CharSequences.contentEquals;
+
+abstract class AbstractContentCodec implements ContentCodec {
+
+    private final CharSequence name;
+
+    AbstractContentCodec(final CharSequence name) {
+        this.name = name;
```
idelpivnitskiy, line 27: Consider validating it's not empty

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/AbstractContentCodec.java
```diff
@@ -0,0 +1,59 @@
...
+            return false;
+        }
+
+        final AbstractContentCodec that = (AbstractContentCodec) o;
+        return contentEquals(name, that.name);
+    }
+
+    @Override
+    public final int hashCode() {
+        return name.hashCode();
```
idelpivnitskiy, line 50: IIUC what was done in #1383, it should use `caseInsensitiveHashCode(name)` to keep pass all requirements for `equals` and `hashCode`.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/ContentCodecBuilder.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package io.servicetalk.encoding.netty;
+
+import io.servicetalk.encoding.api.ContentCodec;
+
+/**
+ * Builder for {@link ContentCodec}.
+ */
+public interface ContentCodecBuilder {
```
idelpivnitskiy, line 23: This interface doesn't provide any methods, consider removing it.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/DefaultContentCodecBuilder.java
```diff
@@ -0,0 +1,96 @@
...
+
+import io.servicetalk.encoding.api.ContentCodec;
+
+import io.netty.handler.codec.compression.JdkZlibDecoder;
+import io.netty.handler.codec.compression.JdkZlibEncoder;
+import io.netty.handler.codec.compression.ZlibWrapper;
+
+import static io.servicetalk.buffer.api.CharSequences.newAsciiString;
+
+abstract class DefaultContentCodecBuilder implements ContentCodecBuilder {
```
idelpivnitskiy, line 26: After removing snappy, there is no value in this `DefaultContentCodecBuilder` top level class. Consider moving `ZipContentCodecBuilder` to the top level

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/DefaultContentCodecBuilder.java
```diff
@@ -0,0 +1,96 @@
...
+
+        /**
+         * Returns the compression level for this codec.
+         * @return return the compression level set for this codec.
+         */
+        protected int compressionLevel() {
+            return compressionLevel;
+        }
+
+        protected int maxChunkSize() {
```
idelpivnitskiy, line 57: Please, add javadoc and make both methods `final`

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,335 @@
...
+    @Override
+    public Buffer encode(final Buffer src, final int offset, final int length, final BufferAllocator allocator) {
+        requireNonNull(src);
+        final int availableBytes = src.readableBytes() - offset;
+        if (offset < 0) {
+            throw new IllegalArgumentException("Invalid offset: " + offset + " (expected >= 0)");
+        }
+
+        if (length <= 0) {
+            throw new IllegalArgumentException("Invalid length: " + length + " (expected > 0");
```
idelpivnitskiy, line 72: It should be ok to attempt to encode an empty buffer. We can return it as-is. Payload body stream can contain empty buffers, it will be annoying to check the `src` size before invoking this method.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,335 @@
...
+        final MessageToByteEncoder<ByteBuf> encoder = encoderSupplier.get();
+        final EmbeddedChannel channel = newEmbeddedChannel(encoder, allocator);
+
+        try {
+            ByteBuf origin = toByteBuf(src, offset, length);
+            channel.writeOutbound(origin);
+
+            // May produce footer
+            preparePendingData(channel);
+            return requireNonNull(drainChannelOutToSingleBuffer(channel, allocator));
```
idelpivnitskiy, line 88: The NPE here will be hard to understand. Consider using a special exception type for encoding and decoding (one or separate) to signal any problems associated with compression. We will use this exception type in server filter to return 413.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,335 @@
...
+
+    @Override
+    public Buffer decode(final Buffer src, final int offset, final int length, final BufferAllocator allocator) {
+        requireNonNull(src);
+        final int availableBytes = src.readableBytes() - offset;
+        if (offset < 0) {
+            throw new IllegalArgumentException("Invalid offset: " + offset + " (expected >= 0)");
+        }
+
+        if (length <= 0) {
```
idelpivnitskiy, line 160: Consider allowing decode of empty buffers. The caller will do `decode(src, 0, src.readableBytes(), alloc)`

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,335 @@
...
+        channel.config().setAllocator(getByteBufAllocator(allocator));
+        return channel;
+    }
+
+    private static ByteBuf toByteBuf(final Buffer buffer) {
+        return BufferUtils.toByteBuf(buffer);
+    }
+
+    private static ByteBuf toByteBuf(final Buffer buffer, final int offset, final int length) {
+        return BufferUtils.toByteBuf(buffer.readerIndex(buffer.readerIndex() + offset).readSlice(length));
```
idelpivnitskiy, line 250: This method will throw if it's not a `NettyBuffer`. However, users can use read-only allocator for the payload body (sending a file as a `MappedByteBuffer` required to use read-only allocator). Consider using `extractByteBufOrCreate`

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,335 @@
...
+            throw new IllegalStateException("Invalid length: " + length + " (expected <= " + availableBytes);
+        }
+
+        final ByteToMessageDecoder decoder = decoderSupplier.get();
+        final EmbeddedChannel channel = newEmbeddedChannel(decoder, allocator);
+
+        try {
+            ByteBuf origin = toByteBuf(src, offset, length);
+            channel.writeInbound(origin);
+            return newBufferFrom(channel.readInbound());
```
idelpivnitskiy, line 174: It's possible that netty's decoder will produce multiple output buffers, we need to read all of them

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,335 @@
...
+            while ((part = channel.readInbound()) != null) {
+                compositeBuffer.addBuffer(newBufferFrom(part));
+            }
+
+            return compositeBuffer;
+        }
+    }
+
+    @Nullable
+    private static Buffer drainChannelOutToSingleBuffer(
```
idelpivnitskiy, line 283: Can we share code between `drainChannelInToSingleBuffer` and `drainChannelOutToSingleBuffer`? A common method can just take a `Queue<Object>`

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,335 @@
...
+                                subscriber.onNext(buffer);
+                            }
+
+                            return;
+                        }
+
+                        channel.writeOutbound(toByteBuf(next));
+                        Buffer buffer = drainChannelOutToSingleBuffer(channel, allocator);
+                        if (buffer != null) {
+                            subscriber.onNext(buffer);
```
idelpivnitskiy, line 134: 1. We probably can not emit two 2+ items from a single `onNext` invocation for the `liftSync` operator. @Scottmitch wdyt?
2. We should request more data if the `buffer` is null.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,335 @@
...
+                this.subscription = subscription;
+                subscriber.onSubscribe(subscription);
+            }
+
+            @Override
+            public void onNext(@Nullable final Buffer src) {
+                assert subscription != null;
+
+                if (src == null) {
+                    return;
```
idelpivnitskiy, line 206: Should we request more data if we ignore one item? Otherwise we will never receive new demand.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,335 @@
...
+                }
+
+                // onNext will produce AT-MOST N items (as received)
+                if (!channel.isOpen()) {
+                    throw new IllegalStateException("Stream encoder previously closed but more input arrived ");
```
idelpivnitskiy, line 211: Consider moving this check to the top. We should not request more items if the channel is already closed

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,335 @@
...
+
+                    @Override
+                    public void onSubscribe(PublisherSource.Subscription subscription) {
+                        subscriber.onSubscribe(subscription);
+                    }
+
+                    @Override
+                    public void onNext(Buffer next) {
+                        if (next == null) {
+                            return;
```
idelpivnitskiy, line 115: Should we request more?

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,335 @@
...
+
+                        // onNext will produce AT-MOST N items (from upstream)
+                        // +1 for the encoding footer (ie. END_OF_STREAM)
+                        if (next == END_OF_STREAM) {
+                            // May produce footer
+                            preparePendingData(channel);
+                            Buffer buffer = drainChannelOutToSingleBuffer(channel, allocator);
+                            if (buffer != null) {
+                                subscriber.onNext(buffer);
+                            }
```
idelpivnitskiy, line 126: Some encoders (like Bzip2) use blocks and do not produce any output until the block is full. We should request more data in this case.

idelpivnitskiy (changes requested):
Welcome to the netty world! It's a complex task when you just start using netty.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
tkountis: Similarly for decoding.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
tkountis: 🦅 👁️

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
tkountis: 🦅 👁️

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
tkountis: Good catch, letting the NPE propagate.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
tkountis: Moved it further up, but it was already guarding further demand.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
tkountis: Propagating the NPE, this is not an expected item.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
tkountis: Added two new Exceptions in the api module. We can further enhance them later as needed to have more info as to what codec was in use etc.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,342 @@
...
+
+            List<Buffer> parts = new ArrayList<>(queue.size());
+            ByteBuf part;
+            while ((part = (ByteBuf) queue.poll()) != null) {
+                parts.add(newBufferFrom(part));
+                accumulateSize += part.readableBytes();
+            }
+
+            // Try to merge everything together if total size is less than 1KiB (small chunks, ie. footer).
+            // CompositeBuffer can requires some additional work to write and may limit how much data we can
```
idelpivnitskiy, line 308: requires -> require

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,342 @@
...
+            // pass to writev (max of 1024 pointers), so if there are small chunks it can be better to combine them.
+            if (accumulateSize <= 1 << 10) {
```
idelpivnitskiy, line 310: Consider using a constant and just use 1024 to improve readability of the code. 1024 is a simple value, it takes longer to compute `1 << 10` in the head :)

idelpivnitskiy (commented):
Hope these are the last comments :)

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,342 @@
...
+        channel.config().setAllocator(getByteBufAllocator(allocator));
+        return channel;
+    }
+
+    private static ByteBuf toByteBuf(final Buffer buffer) {
+        return BufferUtils.extractByteBufOrCreate(buffer);
+    }
+
+    private static ByteBuf toByteBuf(final Buffer buffer, final int offset, final int length) {
+        return BufferUtils.extractByteBufOrCreate(buffer.readerIndex(buffer.readerIndex() + offset).readSlice(length));
```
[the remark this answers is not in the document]
tkountis: We agreed to keep a uniform API across the board with offsets too, I think we should stick to it. The slice here is only an exceptional case and not the norm, most of the times we expect the buffers to be Netty buffers.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    private EmbeddedChannel newEmbeddedChannel(final ChannelHandler handler, final BufferAllocator allocator) {
+        final EmbeddedChannel channel = new EmbeddedChannel(handler);
+        channel.config().setAllocator(getByteBufAllocator(allocator));
+        return channel;
+    }
+
+    private static ByteBuf toByteBuf(final Buffer buffer) {
+        return BufferUtils.extractByteBufOrCreate(buffer);
```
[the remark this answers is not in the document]
tkountis: There is the value of having one place to change it and choose the name. I think its a good fit here and offers no overhead.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,342 @@
...
+                        this.subscription = subscription;
+                        subscriber.onSubscribe(subscription);
+                    }
+
+                    @Override
+                    public void onNext(Buffer next) {
+                        assert subscription != null;
+
+                        if (next == null) {
+                            throw new NullPointerException();
```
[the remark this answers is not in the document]
tkountis: Added it to the encode side too, that shouldn't occur unless RS spec is not respected somewhere in the chain.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,342 @@
...
+                        // onNext will produce AT-MOST N items (from upstream)
+                        // +1 for the encoding footer (ie. END_OF_STREAM)
+                        if (next == END_OF_STREAM) {
+                            // May produce footer
+                            preparePendingData(channel);
+                            Buffer buffer = drainChannelQueueToSingleBuffer(channel.outboundMessages(), allocator);
+                            if (buffer != null) {
+                                subscriber.onNext(buffer);
+                            } else {
+                                subscriber.onNext(EMPTY_BUFFER);
```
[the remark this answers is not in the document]
tkountis: Thanks for the info, I didn't consider that.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,342 @@
...
+        try {
+            ByteBuf origin = toByteBuf(src, offset, length);
+            channel.writeOutbound(origin);
+
+            // May produce footer
+            preparePendingData(channel);
+
+            final Buffer buffer = drainChannelQueueToSingleBuffer(channel.outboundMessages(), allocator);
+            if (buffer == null) {
+                throw new CodecEncodingException("Encoder didn't return any output.");
```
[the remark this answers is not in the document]
tkountis: I was trying to keep this more generic because I am not familiar with compression details of other algorithms. And on top of the algorithm this may be a result of Netty's codec behavior. Anyhow your suggested phrasing is reasonable so using this.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,342 @@
...
+            if (buffer == null) {
+                throw new CodecDecodingException("Decoder didn't return any output.");
+            }
+            return buffer;
+        } catch (Throwable e) {
+            LOGGER.error("Error while decoding with {}", name(), e);
+            throwException(e);
+            return EMPTY_BUFFER;
+        } finally {
+            safeCleanup(channel);
```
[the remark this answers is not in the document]
tkountis: I believe you mean in case of successful completion then exception in safe-cleanup will hide it.
I added the cleanup in the try body, and left the safe one in finally. Any exception (including ones from cleanup) will surface unless something throws before cleanup, in which case that stays silent.

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/IdentityContentCodec.java
```diff
@@ -20,16 +20,15 @@
 import io.servicetalk.concurrent.api.Publisher;
 
 /**
- * NOOP Message encoding codec.
+ * Default, always supported NOOP 'identity' {@link ContentCodec}.
  */
-final class IdentityContentCodec extends AbstractContentCodec {
+public final class IdentityContentCodec implements ContentCodec {
 
     private static final CharSequence NAME = "identity";
```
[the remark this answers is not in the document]
tkountis: :(

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,342 @@
...
+
+    @Override
+    public Buffer encode(final Buffer src, final int offset, final int length, final BufferAllocator allocator) {
+        requireNonNull(src);
+        final int availableBytes = src.readableBytes() - offset;
+        if (offset < 0) {
+            throw new IllegalArgumentException("Invalid offset: " + offset + " (expected >= 0)");
+        }
+
+        if (length > availableBytes) {
```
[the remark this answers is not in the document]
tkountis: Fixed the checks

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,342 @@
...
+
+    @Override
+    public Buffer decode(final Buffer src, final int offset, final int length, final BufferAllocator allocator) {
+        requireNonNull(src);
+        final int availableBytes = src.readableBytes() - offset;
+        if (offset < 0) {
+            throw new IllegalArgumentException("Invalid offset: " + offset + " (expected >= 0)");
+        }
+
+        if (length > availableBytes) {
```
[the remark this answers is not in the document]
tkountis: Fixed the checks

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/ContentCodec.java
```diff
@@ -26,6 +26,11 @@
  */
 public interface ContentCodec {
 
+    /**
+     * Returns the default, always supported NOOP 'identity' {@link ContentCodec}.
+     */
+    ContentCodec IDENTITY = new IdentityContentCodec();
```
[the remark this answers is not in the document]
tkountis: I think I prefer keeping the codec itself hidden. So I will go with the `Identity` class.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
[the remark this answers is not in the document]
idelpivnitskiy: Even if it's a netty buffer, we need a slice because we can not silently skip data between `readerIndex` and `readerIndex + offset`:
```java
Bufffer b = fromAscii("Hi Thomas");
Buffer compressed = codec.encode(b, 3, 6);
```
The "Hi " part is lost now. I don't think any other API that takes an offset behaves in the same way.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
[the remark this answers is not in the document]
idelpivnitskiy: If the `try` throws before `cleanup`, the catch block should do `safeCleanup` and the `finally` is not necessary. Trying to have a consistent approach with `encode` impl.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,372 @@
...
+        return channel;
+    }
+
+    private static ByteBuf toByteBuf(final Buffer buffer) {
+        return BufferUtils.extractByteBufOrCreate(buffer);
+    }
+
+    private static ByteBuf toByteBuf(final Buffer buffer, final int offset, final int length) {
+        final Buffer slice = buffer.slice(buffer.readerIndex() + offset, length);
+        return BufferUtils.extractByteBufOrCreate(slice);
```
idelpivnitskiy, line 308: If you take the slice instead of doing checks by yourself, these methods are not necessary and `extractByteBufOrCreate` can be used.

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/CodecDecodingException.java
```diff
@@ -0,0 +1,56 @@
...
+        this.codec = codec;
+    }
+
+    /**
+     * New instance.
+     *
+     * @param codec the codec in use.
+     * @param cause the cause of the exception.
+     */
+    public CodecDecodingException(final ContentCodec codec, final Throwable cause) {
```
idelpivnitskiy, line 44: Thanks!
It will be good to require a reason message in this case as well in the API. When it's not required, users or we can pass an empty string.

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/CodecEncodingException.java
```diff
@@ -0,0 +1,58 @@
...
+        this.codec = requireNonNull(codec);
+    }
+
+    /**
+     * New instance.
+     *
+     * @param codec the codec in use.
+     * @param cause the cause of the exception.
+     */
+    public CodecEncodingException(final ContentCodec codec, final Throwable cause) {
```
idelpivnitskiy, line 46: Please, add a reason param here

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,372 @@
...
+            }
+
+            cleanup(channel);
+            return buffer;
+        } catch (CodecEncodingException e) {
+            throw e;
+        } catch (Throwable e) {
+            throw new CodecEncodingException(this, e);
+        } finally {
+            safeCleanup(channel);
```
idelpivnitskiy, line 107: Ok, I see the picture now. Let's keep the `finally` block. Thanks!
Ignore my previous https://github.com/apple/servicetalk/pull/1376#discussion_r592504997

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,372 @@
...
+
+            @Override
+            public void onNext(@Nullable final Buffer src) {
+                assert subscription != null;
+                if (!channel.isOpen()) {
+                    throw new IllegalStateException("Stream decoder previously closed but more input arrived");
+                }
+
+                if (src == null) {
+                    throw new NullPointerException("Cannot decode null values");
```
idelpivnitskiy, line 257: Should we throw `CodecDecodingException` in this two cases?

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,372 @@
...
+
+                    @Override
+                    public void onNext(Buffer next) {
+                        assert subscription != null;
+                        if (!channel.isOpen()) {
+                            throw new IllegalStateException("Stream encoder previously closed but more input arrived");
+                        }
+
+                        if (next == null) {
+                            throw new NullPointerException("Cannot encode null values");
```
idelpivnitskiy, line 139: Should we throw `CodecEncodingException` in this two cases?

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -0,0 +1,375 @@
...
+            throw new CodecEncodingException(this, "No data to encode.");
+        }
+
+        if (offset < 0 || offset >= src.readableBytes()) {
+            throw new IllegalArgumentException("Invalid offset: " + offset +
+                    " (expected 0 - " + (src.readableBytes() - 1) + ")");
+        }
+
+        final int availableBytes = src.readableBytes() - offset;
+        if (length > availableBytes || length < 1) {
```
idelpivnitskiy, line 80: You are taking the slice later, it will perform all the same checks for you. Why do we need to duplicate them here instead of taking the slice in the first line as proposed? Do the slice outside of the `try` block.

If I miss something and these checks are still necessary, consider moving them to a private method and share with `decode` method.

idelpivnitskiy (approved):
LGTM after the last minor comments addressed.
Thanks for the pation and all the hard work ❤️

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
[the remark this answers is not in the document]
idelpivnitskiy: Ignore this comment, saw the full picture after review: https://github.com/apple/servicetalk/pull/1376#discussion_r592547778

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
tkountis: Ok, agreed it makes less complicated.

PR #1377, opened 2021-02-19
Fix broken SSL APIs related to SNI and hostname

Scottmitch:
Motivation:
The server SSL config APIs for SNI didn't allow configuring SSL
associated with each individual SNI host. The client APIs coupled the
host name verification algorithm with the non-authoritative peer
host/port. This doesn't correctly represent the semantics of SSL.

Modifications:
- Server side SSL config object supports a new method which allows
  creating a new SSL config object for each SNI host. A map of hostname
  to SSL config is kept internally and used to create the SniHandler.
- Server side SSL config object now uses a consistent way to configure
  the KeyManager material as the client. The protocol specific
  commit(key*) methods were removed and the implementations were
  simplified as a result.
- Client side SSL config object removes methods which couple the host
  name verification algorithm to the non-authoritative peer host/port,
  and adds new methods which allow setting them independently.
- Client implementation now always uses the non-authoritative peer
  host/port (regardless of hostname verification) which is required for
  client session resumption.

Result:
SSL config APIs now correctly represent SNI semantics for client and
server. The client SSL config decouples the non-authoritative peer
host/port from the verification algorithm.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerSecurityConfigurator.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerSecurityConfigurator.java
@@ -33,6 +33,17 @@ public interface HttpServerSecurityConfigurator extends ServerSecurityConfigurat
     HttpServerSecurityConfigurator trustManager(TrustManagerFactory trustManagerFactory);
 
     @Override
+    HttpServerSecurityConfigurator keyManager(KeyManagerFactory keyManagerFactory);
+
+    @Override
+    HttpServerSecurityConfigurator keyManager(Supplier<InputStream> keyCertChainSupplier,
+                                              Supplier<InputStream> keySupplier);
+
+    @Override
+    HttpServerSecurityConfigurator keyManager(Supplier<InputStream> keyCertChainSupplier,
+                                              Supplier<InputStream> keySupplier, String keyPassword);
+
+    @Override
     HttpServerSecurityConfigurator protocols(String... protocols);
 
     @Override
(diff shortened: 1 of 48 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SecurityConfigurator.java
```diff
@@ -65,6 +66,51 @@
...
+    /**
+     * Identifying certificate for this host. {@code keyCertChainInputStream} and {@code keyInputStream} may
+     * be {@code null}, which disables mutual authentication.
+     *
+     * @param keyCertChainSupplier a {@link Supplier} that will provide an input stream for a {@code X.509} certificate
+     * chain in {@code PEM} format.
+     * <p>
+     * The responsibility to call {@link InputStream#close()} is transferred to callers of the {@link Supplier}.
+     * If this is not the desired behavior then wrap the {@link InputStream} and override {@link InputStream#close()}.
+     * @param keySupplier an {@link Supplier} that will provide an input stream for a KCS#8 private key in PEM format.
```
bondolo, line 87: PKCS

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ReadOnlyClientSecurityConfig.java
```diff
@@ -23,27 +23,26 @@
...
-    @Nullable
-    protected String hostNameVerificationHost;
     /**
-     * Only valid if {@link #hostNameVerificationHost} is valid.
+     * See <a href="https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#jssenames">
+     * Endpoint Identification Algorithm Name</a>.
      */
-    protected int hostNameVerificationPort;
+    @Nullable
+    protected String hostnameVerificationAlgorithm = "HTTPS";
```
bondolo, line 31: Should this constant be defined somewhere?

Scottmitch: ideally yes. Ideally this would come from the JDK but [X509TrustManagerImpl](https://github.com/openjdk/jdk11u-dev/blob/master/src/java.base/share/classes/sun/security/ssl/X509TrustManagerImpl.java#L454-L464) just uses the raw string value as well. We could introduce our own constants but we don't really determine what algorithms are and are not supported (we just pass through).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerSecurityConfigurator.java
```diff
@@ -32,6 +32,17 @@
     @Override
     HttpServerSecurityConfigurator trustManager(TrustManagerFactory trustManagerFactory);
 
+    @Override
+    HttpServerSecurityConfigurator keyManager(KeyManagerFactory keyManagerFactory);
+
+    @Override
+    HttpServerSecurityConfigurator keyManager(Supplier<InputStream> keyCertChainSupplier,
```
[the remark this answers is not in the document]
Scottmitch: We created a `InputStream` -> `KeyManagerFactory` utility in Netty. The code is relatively involved and not something I would expect users who just want to use a cert chain from disk to manually code. If there is a simpler way we could always utilize that as the default implementation in the future and remove some code in Netty.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ClientSecurityConfigurator.java
```diff
@@ -51,64 +62,32 @@
...
-     * <a href="https://tools.ietf.org/search/rfc2818#section-3.1">server identity</a>.
-     * @return {@code this}.
-     * @see SSLParameters#setEndpointIdentificationAlgorithm(String)
-     */
-    ClientSecurityConfigurator hostnameVerification(String hostNameVerificationAlgorithm,
-                                                    String hostNameVerificationHost);
-
-    /**
-     * Determines what algorithm to use for hostname verification.
+     * Disable verification of the <a href="https://tools.ietf.org/search/rfc2818#section-3.1">server identity</a>.
```
idelpivnitskiy, line 72: While you are here, can you please clarify in the javadoc the side-effect of this method (i.e. `hostnameVerificationAlgorithm/peerHost/peerPort` will be ignored and it makes your config vulnerable). 

Btw, wdyt about removing this method? For example, grpc-java does not allow users to disable hostname verification. Our generated certificates validate `localhost` hostname, which we can use in tests and examples instead of disabling verification. Users can use our certs for tests or generate their certs with the required hostname.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ClientSecurityConfigurator.java
```diff
@@ -51,64 +62,32 @@
...
-     *
-     * @param hostNameVerificationAlgorithm The algorithm to use when verifying the host name.
-     * See <a href="https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#jssenames">
-     * Supported algorithm names</a>.
-     * @param hostNameVerificationHost the host name used to verify the
-     * <a href="https://tools.ietf.org/search/rfc2818#section-3.1">server identity</a>.
-     * @return {@code this}.
-     * @see SSLParameters#setEndpointIdentificationAlgorithm(String)
-     */
-    ClientSecurityConfigurator hostnameVerification(String hostNameVerificationAlgorithm,
```
idelpivnitskiy, line 71: We should mark these methods as deprecated and keep them around for a few releases, because security configuration is used by many users.

idelpivnitskiy (commented):
Let's discuss API and then will do a second review pass

Scottmitch:
discussed with @idelpivnitskiy offline. closing this PR in favor of https://github.com/apple/servicetalk/pull/1387

PR #1379, opened 2021-02-19
BufferOutputStream.write(int b) should write as byte and not int

muthupalaniappan:
Motivation: BufferOutputStream.write(final int b) writes a byte as int which is wrong.

Modifications:
* in write(final int b){..} method body buffer.writeInt(b) is changed to buffer.writeByte(b)

Result:
Fixes the above bug. Only internal changes and no API changes.


Scottmitch (approved):
good catch! bonus points for unit test.

muthupalaniappan:
Sure let me add the unit test in a new pr.

PR #1380, opened 2021-02-21
`HeaderUtils#pathMatches` incorrectly compares two different `CharSequence` implementations

danfaer:
Motivation:
  This code uses `CharSequence.equals` that does not have well-defined equals behavior. Such a call to equals may return false in cases where equality was expected.
https://errorprone.info/bugpattern/UndefinedEquals

Modifications:
  replaced call to `CharSequence.equals` with `CharSequences.contentEquals`
  
Result:  
 unit test comparing 2 different implementation of `CharSequences` is now passing

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
@@ -318,7 +318,7 @@ public final class HeaderUtils {
         }
 
         if (requestPath.length() == cookiePath.length()) {
-            return requestPath.equals(cookiePath);
+            return contentEquals(requestPath, cookiePath);
         }
         final boolean actualStartsWithSlash = cookiePath.charAt(0) == '/';
         final int length = min(actualStartsWithSlash ? cookiePath.length() - 1 :
--- a/servicetalk-http-api/src/test/java/io/servicetalk/http/api/HeaderUtilsTest.java
+++ b/servicetalk-http-api/src/test/java/io/servicetalk/http/api/HeaderUtilsTest.java
@@ -30,6 +30,7 @@ import static io.netty.util.AsciiString.of;
 import static io.servicetalk.http.api.HeaderUtils.checkContentType;
 import static io.servicetalk.http.api.HeaderUtils.isTchar;
 import static io.servicetalk.http.api.HeaderUtils.isTransferEncodingChunked;
+import static io.servicetalk.http.api.HeaderUtils.pathMatches;
 import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_LENGTH;
 import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_TYPE;
 import static io.servicetalk.http.api.HttpHeaderNames.ORIGIN;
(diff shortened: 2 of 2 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
```diff
@@ -32,7 +36,7 @@
      * @param cookieName The <a href="https://tools.ietf.org/html/rfc6265#section-4.1.1">cookie-name</a>.
      * @param cookieValue The <a href="https://tools.ietf.org/html/rfc6265#section-4.1.1">cookie-value</a>.
      */
-    public DefaultHttpCookiePair(final CharSequence cookieName, final CharSequence cookieValue) {
+    public DefaultHttpCookiePair(final @Nonnull CharSequence cookieName, final @Nonnull CharSequence cookieValue) {
```
tkountis, line 39: The annotations are not needed here, at package level we keep all elements Nonnull by default (see. package-info.java)

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
```diff
@@ -138,9 +142,7 @@ public boolean equals(final Object o) {
 
     @Override
     public int hashCode() {
-        int hash = 31 + name.hashCode();
-        hash = 31 * hash + value.hashCode();
-        return hash;
+        return Objects.hash(name, value);
```
tkountis, line 145: We tend to prefer the direct calculation vs the helper methods for perf/allocations reasons (ie. array creation for args).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
```diff
@@ -318,7 +318,7 @@ public static boolean pathMatches(final CharSequence requestPath, @Nullable fina
         }
 
         if (requestPath.length() == cookiePath.length()) {
-            return requestPath.equals(cookiePath);
+            return contentEquals(requestPath, cookiePath);
```
tkountis, line 321: Good catch 👍

servicetalk-http-api/src/test/java/io/servicetalk/http/api/HeaderUtilsTest.java
```diff
@@ -273,6 +273,17 @@ public void validateToken() {
         }
     }
 
+    @Test
+    public void pathMatches() {
+        assertTrue(HeaderUtils.pathMatches("/a/b/c", "/a/b/c"));
+        assertTrue(HeaderUtils.pathMatches("/a/b/cxxxx", "/a/b/c"));
+        assertTrue(HeaderUtils.pathMatches(new StringBuilder("/a/b/c"), new StringBuilder("/a/b/c")));
```
tkountis, line 280: I was surprised that this implements the `CharSequence` interface. It relies on Object.hashCode & Object.equals because its mutable, but then it acts as a String too!

servicetalk-http-api/src/test/java/io/servicetalk/http/api/HeaderUtilsTest.java
```diff
@@ -273,6 +273,17 @@ public void validateToken() {
         }
     }
 
+    @Test
+    public void pathMatches() {
+        assertTrue(HeaderUtils.pathMatches("/a/b/c", "/a/b/c"));
```
tkountis, line 278: nit: Consider statically importing the method.

tkountis (commented):
Thanks for the contribution @danfaer 💯 
Some minor comments then it looks good. Nice catch.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
normanmaurer: hmm... should this compare ignoring case ?

tkountis: Good question, I looked it up and its not explicit. Also the previous solution was case sensitive too.

Some of the sources [1][2] treats the components as case sensitive.

1. https://tools.ietf.org/html/draft-ietf-httpbis-p1-messaging-25#section-2.7.3
>... The scheme and host are case-insensitive and normally provided in lowercase; all other components are compared in a case-sensitive manner. ...

2. https://tools.ietf.org/html/rfc3986#section-6.2.2.1
> ... The other generic syntax components are assumed to be case-sensitive unless specifically defined otherwise by the scheme ...

danfaer:
fixed PR comments

idelpivnitskiy (approved):
Thanks @danfaer for the contribution!
Great catch!

PR #1382, opened 2021-02-22
Switch to non-deprecated assertThat

bondolo:
Motivation:
Currently many of the existing tests refer to `org.junit.Assert.assertThat` which has been deprecated for dependency reasons. The preferred implementation for assertThat is now `org.hamcrest.MatcherAssert.assertThat` Since all of the impacted tests also already reference Hamcrest no new dependencies are required.

Modification:
Replace `org.junit.Assert.assertThat` with `org.hamcrest.MatcherAssert.assertThat`

Result:
No change in test performance, the deprecated method simply calls the preferred method but the IDE deprecation warnings are eliminated.


tkountis (approved):
LGTM 🧹 🧹 🧹 

nit: we should also update the copyrights

PR #1383, opened 2021-02-22
Replace CharSequence equality checks with contentEquals

danfaer:
Motivation:
CharSequence.equals does not have well-defined equals behavior. Such a call to equals may return false in cases where equality was expected.
https://errorprone.info/bugpattern/UndefinedEquals

Modifications:
replaced call to CharSequence.equals with CharSequences.contentEquals
temporarily using CharSequences.caseInsensitiveHashCode until CharSequences.hashCode is implemented

Result:
Comparing 2 different implementation of CharSequences created with the same chars returns true

### DIFF
--- a/servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractContentCodec.java
+++ b/servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractContentCodec.java
@@ -15,7 +15,8 @@
  */
 package io.servicetalk.encoding.api;
 
-import java.util.Objects;
+import static io.servicetalk.buffer.api.CharSequences.caseInsensitiveHashCode;
+import static io.servicetalk.buffer.api.CharSequences.contentEquals;
 
 abstract class AbstractContentCodec implements ContentCodec {
 
@@ -30,20 +31,20 @@ abstract class AbstractContentCodec implements ContentCodec {
     }
 
     @Override
-    public boolean equals(final Object o) {
+    public boolean equals(Object o) {
         if (this == o) {
             return true;
         }
-        if (o == null || getClass() != o.getClass()) {
+        if (!(o instanceof AbstractContentCodec)) {
             return false;
         }
         final AbstractContentCodec that = (AbstractContentCodec) o;
-        return name.equals(that.name);
+        return contentEquals(name, that.name);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(name);
+        return caseInsensitiveHashCode(name);
     }
 
     @Override
--- a/servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractContentCodec.java
+++ b/servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractContentCodec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2020 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2020-2021 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
@@ -15,8 +15,6 @@
  */
 package io.servicetalk.http.api;
 
-import javax.annotation.Nonnull;
-
 import static io.servicetalk.buffer.api.CharSequences.caseInsensitiveHashCode;
 import static io.servicetalk.buffer.api.CharSequences.contentEquals;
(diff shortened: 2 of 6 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
```diff
@@ -32,7 +36,7 @@
      * @param cookieName The <a href="https://tools.ietf.org/html/rfc6265#section-4.1.1">cookie-name</a>.
      * @param cookieValue The <a href="https://tools.ietf.org/html/rfc6265#section-4.1.1">cookie-value</a>.
      */
-    public DefaultHttpCookiePair(final CharSequence cookieName, final CharSequence cookieValue) {
+    public DefaultHttpCookiePair(final @Nonnull CharSequence cookieName, final @Nonnull CharSequence cookieValue) {
```
tkountis, line 39: `@Nonnull` by default

tkountis (approved):
Nice thanks for the contribution!
Minor comment then looks good. Also, please consider updating the copyright years to the files you touched.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultHttpCookiePairTest.java
```diff
@@ -41,24 +48,30 @@ public void testEqual() {
...
+                   is(not(new DefaultHttpCookiePair("Foo", "bar"))));
         assertThat(new DefaultHttpCookiePair("foo", "bar").hashCode(),
-                is(not(new DefaultHttpCookiePair("Foo", "bar").hashCode())));
+                is(not(new DefaultHttpCookiePair("fooo", "bar").hashCode())));
+
+        assertThat(new DefaultHttpCookiePair("foo", "bar"),
+                   is(not(new DefaultHttpCookiePair(newAsciiString("Foo"), newAsciiString("bar")))));
+        assertThat(new DefaultHttpCookiePair("foo", "bar").hashCode(),
+                   is(not(new DefaultHttpCookiePair(newAsciiString("fooo"), newAsciiString("bar")).hashCode())));
+
```
tkountis, line 59: Checkstyle error in the build due to extra empty line.

danfaer: Fixed

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpCookiePair.java
danfaer: removed

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractContentCodec.java
```diff
@@ -30,20 +31,20 @@ public final CharSequence name() {
...
         if (this == o) {
             return true;
         }
-        if (o == null || getClass() != o.getClass()) {
+        if (!(o instanceof AbstractContentCodec)) {
             return false;
         }
         final AbstractContentCodec that = (AbstractContentCodec) o;
-        return name.equals(that.name);
+        return contentEquals(name, that.name);
```
idelpivnitskiy, line 42: Great catch! I also found in RFC that all content-coding values are case-insensitive, see https://tools.ietf.org/html/rfc7231#section-3.1.2.1. That means that the `hashCode` impl here is correct 😄 

Let's merge this PR as is (don't want to mix 2 issues in one PR) and I've filed #1385 for the follow-up if you are interested.

idelpivnitskiy (approved):
@danfaer thank you so much for the contributions! ❤️

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/AbstractContentCodec.java
```diff
@@ -30,20 +31,20 @@ public final CharSequence name() {
     }
 
     @Override
-    public boolean equals(final Object o) {
+    public boolean equals(Object o) {
         if (this == o) {
             return true;
         }
-        if (o == null || getClass() != o.getClass()) {
```
bondolo, line 37: comparing the exact class and instanceof is not quite the same. The existing implementation would declare different sub-classes to be non-equal where this implementation might declare them equivalent. There is also the danger of non-reflexive equals() if one of the sub-classes declares an equals() method that matches specifically against their own class.

I would encourage you to leave it getClass() != o.getClass() unless you can make the equals() method final.

idelpivnitskiy: good point!
this class is moved in #1376. let's handle it there and as part of #1386

danfaer: here is the argument against `getClass()` : https://errorprone.info/bugpattern/EqualsGetClass

PR #1384, opened 2021-02-24
`LoadBalancedStreamingHttpClient` should not hide `PayloadInfo` of requests

idelpivnitskiy:
Motivation:

`LoadBalancedStreamingHttpClient` uses `StreamingHttpRequestWithContext`
wrapper to deliver `Runnable` to the underlying HTTP/2 transport level.
However, if we do not unwrap the request as soon as transport receives
it, the wrapper hides `PayloadInfo` interface of the original request
object and the following request processing can not apply required
optimizations for the payload body and flush strategy.

Modifications:

- Unwrap the original request object as soon as
`StreamingHttpRequestWithContext` finished its mission;

Result:

HTTP processing again sees `PayloadInfo` and can apply optimizations for
aggregated API (flush on end).


tkountis (approved):

Scottmitch (approved):
we should add two items to investigate for http api stabilization:

- Wrapping requests/responses may defeat PayloadInfo optimizations
- Reliable way to pass state associated with requests/responses (e.g. AsyncContext costly and can be disabled, Request Context?)

PR #1388, opened 2021-02-26
gRPC Unix Domain Socket Client Support

Scottmitch:
Motivation:
gRPC client builders do not expose methods to easily configure UDS
address.

Modifications:
- Add GrpcClients.forResolvedAddress(..) method that allows for passing
  a UDS type of SocketAddress.

Result:
gRPC Clients can more easily be built for UDS.

### DIFF
--- a/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcClients.java
+++ b/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcClients.java
@@ -84,8 +84,7 @@ public final class GrpcClients {
     }
 
     /**
-     * Creates a {@link GrpcClientBuilder} for an address with default {@link LoadBalancer} and DNS
-     * {@link ServiceDiscoverer}.
+     * Creates a {@link GrpcClientBuilder} for an address with default {@link LoadBalancer}.
      *
      * @param address the {@code ResolvedAddress} to connect to.
      * @return new builder for the address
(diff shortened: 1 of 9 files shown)


Scottmitch:
Build failure attributed to https://github.com/apple/servicetalk/issues/1224

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcClients.java
```diff
@@ -103,6 +106,21 @@ private GrpcClients() {
         return new DefaultGrpcClientBuilder<>(HttpClients.forResolvedAddress(address));
     }
 
+    /**
+     * Creates a {@link GrpcClientBuilder} for an address with default {@link LoadBalancer} and DNS {@link
+     * ServiceDiscoverer}.
```
idelpivnitskiy, line 111: Because it deals with the resolved address, does it make sense to mention DNS ServiceDiscoverer here?
Some question for other `forResolvedAddress` overloads here and in `HttpClients`.

bondolo (commented):
LGTM, but will leave approval for another.

idelpivnitskiy (approved):
🚀

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcClients.java
Scottmitch: agreed, I don't think we need to reference DNS for "resolved" address methods.

PR #1394, opened 2021-03-04
gRPC missing grpc-status error code update

Scottmitch:
Motivation:
If the client receives a response with no grpc-status we return an error
as [required by the spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md).
However we currently return a status code of INTERNAL but the UNKNOWN
status code is more appropriate [1][2].

[1] https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
> Error parsing returned status, UNKNOWN, Client

[2] https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md
> 200 is UNKNOWN because there should be a grpc-status in case of truly OK response.

Modifications:
- Change error status code to UNKNOWN if the grpc-status is not present.

Result:
More appropriate error code when grpc-status is missing from the
response on the client side.


bondolo (approved):

idelpivnitskiy (approved):
🚀

PR #1399, opened 2021-03-04
Clarify AsyncContext put API relative to @Nullable

Scottmitch:
Motivation:
The AsyncContext API and implementation are inconsistent on how null
values are treated for the put(key, value) method.

Modifications:
- Make the AsyncContext and AsyncContextMap APIs consistent with
  java.util.Map API relative to null values. The API allows null values
  and permits the implementation to not support it.

Result:
AsyncContext APIs have the same semantics as java.util.Map for null
values on the put(key, value) method.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AsyncContext.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AsyncContext.java
@@ -92,7 +92,7 @@ public final class AsyncContext {
      * Convenience method for to put all the key/value pairs into the current context.
      *
      * @param map contains the key/value pairs that will be added.
-     * @throws ConcurrentModificationException Done on a best effort basis if {@code entries} is detected to be modified
+     * @throws ConcurrentModificationException done on a best effort basis if {@code entries} is detected to be modified
      * while attempting to put all entries.
      * @throws NullPointerException if {@code key} or {@code value} is {@code null} and the underlying
      * {@link AsyncContextMap} implementation doesn't support {@code null} keys or values.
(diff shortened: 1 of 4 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AsyncContext.java
```diff
@@ -77,16 +78,26 @@ public static AsyncContextMap current() {
...
-    public static <T> void put(AsyncContextMap.Key<T> key, T value) {
+    public static <T> void put(AsyncContextMap.Key<T> key, @Nullable T value) {
         current().put(key, value);
     }
 
     /**
      * Convenience method for to put all the key/value pairs into the current context.
      *
      * @param map contains the key/value pairs that will be added.
+     * @throws ConcurrentModificationException Done on a best effort basis if {@code entries} is detected to be modified
```
tkountis, line 95: Done -> done

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AsyncContextMap.java
```diff
@@ -93,6 +97,8 @@ public String toString() {
      * @param key the key to lookup.
      * @return {@code true} if this context contains a key/value entry corresponding to {@code key}.
      * {@code false} otherwise.
+     * @throws NullPointerException (optional behavior) if {@code key} is {@code null} and the implementation doesn't
+     * support {@code null} keys or values.s
```
tkountis, line 101: values.s -> values

tkountis (approved):
Minor typos

idelpivnitskiy:
Late LGTM

Issue #1294, opened 2020-12-18
DnsServiceDiscovererObserverTest.srvQueryResolutionResult test failure


Scottmitch:
https://github.com/apple/servicetalk/pull/1285/checks?check_run_id=1576011788
```
java.lang.AssertionError: Unexpected TTL value
Expected: is <1>
     but: was <0>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at io.servicetalk.dns.discovery.netty.DnsServiceDiscovererObserverTest.assertResolutionResult(DnsServiceDiscovererObserverTest.java:270)
	at io.servicetalk.dns.discovery.netty.DnsServiceDiscovererObserverTest.srvQueryResolutionResult(DnsServiceDiscovererObserverTest.java:292)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at io.servicetalk.concurrent.internal.ServiceTalkTestTimeout$TimeoutStatement$CallableStatement.call(ServiceTalkTestTimeout.java:171)
	at io.servicetalk.concurrent.internal.ServiceTalkTestTimeout$TimeoutStatement$CallableStatement.call(ServiceTalkTestTimeout.java:163)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

Scottmitch:
[...]

PR #1400, opened 2021-03-04
TcpConnector reduce log noise

Scottmitch:
Motivation:
TcpConnector logs an error message if a connection attempt fails and we
aren't able to propagate the exception. This may commonly happen in the
event of cancellation as Netty will propagate a CancellationException to
Future listeners.

Modifications:
- Remove error log statements from TcpConnector if we are unable to
  complete async control flow

Result:
Less noise in log statements from TcpConnector.


Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/1294

idelpivnitskiy (approved):
🚢

PR #1403, opened 2021-03-04
gRPC Router reduce error log noise

Scottmitch:
Motivation:
GrpcRouter and GrpcServerBuilder will catch exceptions and translate
them into errors which get propagated into trailers. The exception path
is the expected way for servers to return application level errors to
clients. These methods also log error level log messages which is noisy
and can be replicated via a filter or in the application level if
necessary.

Modifications:
- Remove error level logging in GrpcRouter and CatchAllHttpServiceFilter

Result:
Less error level logging on exception paths which maybe expected by the
application.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
@@ -94,8 +94,6 @@ import static java.util.Collections.unmodifiableMap;
  * implementation of a <a href="https://www.grpc.io">gRPC</a> method.
  */
 final class GrpcRouter {
-    private static final Logger LOGGER = LoggerFactory.getLogger(GrpcRouter.class);
-
     private final Map<String, RouteProvider> routes;
     private final Map<String, RouteProvider> streamingRoutes;
     private final Map<String, RouteProvider> blockingRoutes;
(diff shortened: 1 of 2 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -271,13 +269,10 @@ private static void mergeRoutes(final Map<String, RouteProvider> first,
                                                         serializationProvider.serializerFor(responseEncoding,
                                                                 responseClass)))
                                         .recoverWith(cause -> {
-                                            LOGGER.error("Unexpected exception from route: {}, path: {}.", route, path,
```
tkountis, line 274: I liked the logging inside the `recoverWith` operator. it gives path/route info and it can be muted if it causes noise.

tkountis (approved):
LGTM

Scottmitch:

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
idelpivnitskiy: Or we can log it only for unexpected exceptions, but ignore for `GrpcStatusException` because only user can generate `GrpcStatusException` before control flow reaches this place.

Scottmitch: I would prefer to avoid putting debug logging in the core that can be added at higher layers. if more context is required users can always apply their own error recovery to log, record metrics, etc.

bondolo (approved):

PR #1405, opened 2021-03-05
Add javadoc comments to gRPC code generation

Scottmitch:
Motivation:
We currently don't add javadoc comments to the gRPC generated code. This
makes it more difficult for users to understand behavior and
expectations of our generated APIs.

Modifications:
- Add javadoc support to the gRPC protoc plugin. The plugin also
  extracts any comments in the .proto service definitions.
- Clarify ownership on the GrpcPayloadWriter belongs to the user (e.g.
  user must call close()).
- Add an option to the protoc plugin to disable javadoc generation.

Result:
The gRPC protoc plugin generates javadocs.


servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -116,12 +122,14 @@
...
 import static io.servicetalk.grpc.protoc.Words.request;
 import static io.servicetalk.grpc.protoc.Words.requestEncoding;
+import static io.servicetalk.grpc.protoc.Words.responseWriter;
 import static io.servicetalk.grpc.protoc.Words.routes;
 import static io.servicetalk.grpc.protoc.Words.rpc;
 import static io.servicetalk.grpc.protoc.Words.service;
 import static io.servicetalk.grpc.protoc.Words.strategy;
 import static io.servicetalk.grpc.protoc.Words.strategyFactory;
 import static io.servicetalk.grpc.protoc.Words.supportedMessageCodings;
+import static java.lang.System.lineSeparator;
```
bondolo, line 132: I wonder if we should just be using "\n". It is unlikely that we will be generating javadoc on Windows but there it would be "\r\n".

Scottmitch: although we currently don't run CI on windows we should still maintain compatibility. we should enable windows CI in the future (after existing flaky tests are addressed).

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/StringUtils.java
```diff
@@ -100,4 +100,57 @@ static boolean isNullOrEmpty(@Nullable final String v) {
         }
         return options;
     }
+
+    static String escapeJavaDoc(String rawJavaDoc) {
```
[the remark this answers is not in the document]
Scottmitch: good point! added.

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1277

bondolo (approved):
🤩

idelpivnitskiy:
Love the feature! After checking the generated output of `compression.proto` have a couple of questions:

1. Why do we need to enclose the comment inside `<pre>` tag?
2. Can we remove incomplete javadoc from classes that override the service? Example: 

```java
    public static class GreeterServiceFilter implements GreeterService {
        private final GreeterService delegate;

        protected GreeterServiceFilter(final GreeterService delegate) {
            this.delegate = delegate;
        }

        protected GreeterService delegate() {
            return delegate;
        }

        @Override
        public Completable closeAsync() {
            return delegate.closeAsync();
        }

        @Override
        public Completable closeAsyncGracefully() {
            return delegate.closeAsyncGracefully();
        }

        /**
         * @param request the request from the client.
         * @return a {@link Single} which sends the response to the client when it terminates.
         */
        @Override
        public Single<HelloReply> sayHello(final GrpcServiceContext ctx,
                final HelloRequest request) {
            return delegate.sayHello(ctx, request);
        }
    }
```
Javadoc is inherited from the `SayHelloRpc`, we don't need to add a new javadoc section again. Currently added section is incomplete: it doesn't have the initial line and `ctx` param.
The same problem exists with `GreeterClientFilter`, `DefaultGreeterClient`, `FilterableGreeterClientToGreeterClient`, `DefaultBlockingGreeterClient`, `GreeterClientToBlockingGreeterClient`.

Scottmitch:
> Why do we need to enclose the comment (initial line) inside <pre> tag?

This preserves any spaces/newlines/formatting users put in the `.proto` during javadoc html rending. grpc-java also does this by default fwiw.

> Can we remove incomplete javadoc from classes that override the service?

Yes! will do in followup.

> There is usually an empty line before the tags section, can we add one?

sure why not.

Scottmitch:
@idelpivnitskiy - https://github.com/apple/servicetalk/pull/1414

PR #1406, opened 2021-03-05
Require gRPC blocking streaming services to close GrpcPayloadWriter

Scottmitch:
Motivation:
The HTTP blocking streaming APIs passes ownership of the
HttpPayloadWriter to the user to close. The user may process the stream
on another thread after the handle method has returned, or close the
stream with an exception status. The gRPC blocking streaming handle
method always force closes the stream after the user code returns.

Modifications:
- Remove the force closure in gRPC blocking streaming routing layer
- Fix tests which were not invoking close()

Result:
HTTP and gRPC blocking streaming APIs are more consistent, and the user
has the ability to control when the GrpcPayloadWriter is closed.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpService.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpService.java
@@ -16,6 +16,7 @@
 package io.servicetalk.http.api;
 
 import io.servicetalk.concurrent.GracefulAutoCloseable;
+import io.servicetalk.oio.api.PayloadWriter;
 
 /**
  * The equivalent of {@link StreamingHttpService} but with synchronous/blocking APIs instead of asynchronous APIs.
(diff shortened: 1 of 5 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpService.java
```diff
@@ -27,7 +28,8 @@
      *
      * @param ctx Context of the service.
      * @param request to handle.
-     * @param response to send to the client.
+     * @param response to send to the client. The implementation of this method is responsible for calling
```
Scottmitch, line 31: @idelpivnitskiy - this change went in previously and we discussed the trade-offs highlighted in the commit message. Note that I updated the gRPC javadocs to clarify ownership in https://github.com/apple/servicetalk/pull/1405.

bondolo (approved):

idelpivnitskiy (approved):
🚀

idelpivnitskiy:
Btw, do we need a test that verifies the blocking service does not complete successfully if the payload body is not closed?

Scottmitch:
a bunch of our pre-existing tests timeout without the closure so I think we are covered.

PR #1408, opened 2021-03-06
Do not add `transfer-encoding: chunked` inside `transform` method

idelpivnitskiy:
Motivation:

Our current implementation of `transform(TrailersTransformer)` method on
request/response types adds `transfer-encoding: chunked` header with
an assumption that if users use this method, they may add trailers, and the
chunked encoding is required.

However, this method is also used only to access and observe trailers on
the receiver side (e.g., for metrics purposes). As a result, it adds
`transfer-encoding: chunked` header which alters the meaning of the received
message when nothing actually happens.

Modifications:

- Do not add `transfer-encoding: chunked` inside `transform` method, defer this
logic until we prepare a message to be written;
- Improve `addResponseTransferEncodingIfNecessary` and
`addRequestTransferEncodingIfNecessary` to add `transfer-encoding: chunked` if
the message may have trailers;
- Do not add `transfer-encoding: chunked` for HTTP/2 because it will be removed
later (by h2 encoder) anyway;
- Do not add `transfer-encoding: chunked` for HTTP/1.0, because it's not supported;
- `HttpApiConversions#mayHaveTrailers` does not check headers anymore, it's not
necessary;
- Add tests to verify the expected behavior;
- Remove unused `StreamingHttpPayloadHolder#h1TrailersSupported` field;
- Rename `HeaderUtils#shouldAddZeroContentLength` to `responseMayHaveContent`;

Result:

`transform(TrailersTransformer)` does not add `transfer-encoding: chunked` header.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
```diff
@@ -240,21 +230,26 @@ private static void updateResponseContentLength(final int contentLength, final H
...
+        if (canAddResponseTransferEncodingProtocol(response.status().code(), requestMethod) &&
+                canAddTransferEncodingChunked(response)) {
             response.headers().add(TRANSFER_ENCODING, CHUNKED);
         }
-        return response;
     }
 
     static void addRequestTransferEncodingIfNecessary(final StreamingHttpRequest request) {
-        if (canAddRequestTransferEncoding(request)) {
+        if (clientMaySendPayloadBodyFor(request.method()) && canAddTransferEncodingChunked(request)) {
```
Scottmitch, line 242: it is unexpected that the outer method doesn't have `client` in the name, and we are calling a `client` specific method. as discussed offline there are so many methods in this class which do very similar things (e.g. content-length, transfer-encoding) it would be good to clarify in a followup PR.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
```diff
@@ -81,28 +81,18 @@ static boolean canAddRequestContentLength(final StreamingHttpRequest request) {
 
     static boolean canAddResponseContentLength(final StreamingHttpResponse response,
                                                final HttpRequestMethod requestMethod) {
-        return canAddContentLength(response) && shouldAddZeroContentLength(response.status().code(), requestMethod)
+        return canAddContentLength(response) && responseMayHaveContent(response.status().code(), requestMethod)
                 // HEAD requests should either have the content-length already set (= what GET will return) or
                 // have the header omitted when unknown, but never have any payload anyway so don't try to infer it
                 && !isHeadResponse(requestMethod);
```
Scottmitch, line 87: while you are here, can we inline `isHeadResponse`? this method isn't doing anything special

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
```diff
@@ -112,8 +102,8 @@ static boolean canAddResponseTransferEncodingProtocol(final int statusCode,
     }
 
     private static boolean canAddContentLength(final HttpMetaData metadata) {
-        return !hasContentHeaders(metadata.headers()) &&
-                isSafeToAggregate(metadata) && !mayHaveTrailers(metadata);
+        return !hasContentHeaders(metadata.headers()) && isSafeToAggregate(metadata) &&
+                (!mayHaveTrailers(metadata) || metadata.version().major() > 1);
```
Scottmitch, line 106: nit: consider swapping order of `mayHaveTrailers` and `metadata.version().major() > 1` -> short circuit evaluation may avoid casting if h2+

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
```diff
@@ -240,21 +230,26 @@ private static void updateResponseContentLength(final int contentLength, final H
...
-        if (canAddRequestTransferEncoding(request)) {
+        if (clientMaySendPayloadBodyFor(request.method()) && canAddTransferEncodingChunked(request)) {
             request.headers().add(TRANSFER_ENCODING, CHUNKED);
         }
     }
 
-    private static boolean hasContentHeaders(final HttpHeaders headers) {
+    private static boolean canAddTransferEncodingChunked(HttpMetaData metaData) {
+        final HttpHeaders headers = metaData.headers();
+        return !isTransferEncodingChunked(headers) && (!headers.contains(CONTENT_LENGTH) || mayHaveTrailers(metaData));
```
Scottmitch, line 249: should we also take into account the protocol version (if available, similar to `canAddContentLength`)? for h2 we never need to add transfer encoding.

Scottmitch (approved):

idelpivnitskiy:
@Scottmitch your new tests in #1421 helped discover a few issues with my previous approach. I've updated tests and the `canAddTransferEncodingChunked` logic. please, take another look

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpRequest.java
```diff
@@ -175,7 +175,14 @@ default InputStream payloadBodyInputStream() {
 
     /**
      * Returns a {@link BlockingStreamingHttpRequest} with its underlying payload transformed to {@link Buffer}s,
-     * with access to the trailers.
+     * with access to the <a href="https://tools.ietf.org/html/rfc7230#section-4.1.2">trailer</a>s.
+     * <p>
+     * Trailers are supported only when the {@link HttpProtocolVersion#HTTP_1_1 HTTP/1.1} version is used or above.
```
Scottmitch, line 180: `Trailers are supported only when the {@link HttpProtocolVersion#HTTP_1_1 HTTP/1.1} version is used or above.` -> `The minimum required version for trailers is {@link HttpProtocolVersion#HTTP_1_1 HTTP/1.1}`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpRequest.java
```diff
@@ -175,7 +175,14 @@ default InputStream payloadBodyInputStream() {
...
+     * HTTP/1.1 uses <a href="https://tools.ietf.org/html/rfc7230#section-4.1">Chunked Transfer Coding</a> to deliver
```
Scottmitch, line 181:
```java
In order to use trailers with HTTP/1.1, you must use
<a href="https://tools.ietf.org/html/rfc7230#section-4.1">Chunked Transfer Coding</a>.
This method will not modify existing headers so if you intend to add trailers
to the request where non may have previously existed you should do the following:
<ul>
<li>Remove the {@link HttpHeaderNames#CONTENT_LENGTH} header</li>
<li>(optional) Add the {@link HttpHeaderNames#TRANSFER_ENCODING}</li>
<li>(optional) Consider putting the names of each trailer in the {@link HttpHeaderNames#TRAILERS} header</li>
<li>(optional) Validating your peer accepts trailers via the {@link HttpHeaderNames#TE} header</li>
</ul>
```

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpRequest.java
idelpivnitskiy: I've removed these comments because the behavior was reverted.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpRequest.java
idelpivnitskiy: I've removed these comments because the behavior was reverted.

Scottmitch (approved):

PR #1416, opened 2021-03-09
`FromInputStreamPublisher` should apply an upper bound to temporary buffer size

CamW:
Motivation:

An `InputStream` with an arbitrary number of bytes available to be read
must be usable as a supply for payload body on a request.
However, supplying an `InputStream` with available bytes greater than
VM array size limit [1]. Results in `java.lang.OutOfMemoryError:
Requested array size exceeds VM limit` being thrown.

1. https://plumbr.io/outofmemoryerror/requested-array-size-exceeds-vm-limit

Modifications:

- Limited the max size of the buffer into which `InputStream` is read
to 64kb.

Result:

Providing an `InputStream` with available bytes greater than max VM
array size no longer results in an `OutOfMemoryError`.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
@@ -83,7 +83,9 @@ final class FromInputStreamPublisher extends Publisher<byte[]> implements Publis
      */
     FromInputStreamPublisher(final InputStream stream, final int readChunkSize) {
         this.stream = requireNonNull(stream);
-        this.readChunkSize = readChunkSize;
+        if (readChunkSize <= 0) {
+            throw new IllegalArgumentException("readChunkSize: " + readChunkSize + " (expected: >0)");
+        }
     }
 
     @Override
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
@@ -45,6 +45,7 @@ import static java.util.Objects.requireNonNull;
  */
 final class FromInputStreamPublisher extends Publisher<byte[]> implements PublisherSource<byte[]> {
     private static final Logger LOGGER = LoggerFactory.getLogger(FromInputStreamPublisher.class);
+    private static final int DEFAULT_READ_CHUNK_SIZE = 65536;
     private static final AtomicIntegerFieldUpdater<FromInputStreamPublisher> subscribedUpdater =
             AtomicIntegerFieldUpdater.newUpdater(FromInputStreamPublisher.class, "subscribed");
 
@@ -62,6 +63,7 @@ final class FromInputStreamPublisher extends Publisher<byte[]> implements Publis
     private volatile int subscribed;
 
     private final InputStream stream;
+    private final int readChunkSize;
 
     /**
      * A new instance.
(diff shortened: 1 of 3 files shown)


tkountis:
Thanks for the contribution @CamW 
Looking at the solution and wondering whether we should expose the `chunk-size` to the `Publisher#fromInputStream` API. The default is fine of course, but there will certainly be cases where we want more control over it.

idelpivnitskiy (commented):
Thank you @CamW for the great catch and the provided fix 🎖️ 
Agreed with @tkountis, would be nice to add `Publisher.fromInputStream(InputStream, int)` overload that will take the `readChunkSize` from users in case they need to override the default. Otherwise, the PR look great!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
```diff
@@ -62,14 +63,27 @@
...
+    /**
+     * A new instance.
+     *
+     * @param stream the {@link InputStream} to expose as a {@link Publisher}
+     * @param readChunkSize the maximum length of {@code byte[]} chunks which will be read from the {@link InputStream}
+     * and emitted by the {@link Publisher}.
+     */
+    FromInputStreamPublisher(final InputStream stream, final int readChunkSize) {
         this.stream = requireNonNull(stream);
+        this.readChunkSize = readChunkSize;
```
idelpivnitskiy, line 86: Can you please add validation for this param to make sure it's greater than 0? We usually use this consistent format:
```suggestion
        if (readChunkSize <= 0) {
            throw new IllegalArgumentException("readChunkSize: " + readChunkSize + " (expected: >0)");
        }
```

CamW:
Thanks for the reviews @tkountis, @idelpivnitskiy. I've allowed for the chunk size to be configured and added `Publisher.fromInputStream(InputStream, int)`. Happy to make any further changes.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2913,7 +2913,7 @@ protected final void subscribeInternal(Subscriber<? super T> subscriber) {
      * the {@link Subscription} should be "responsive". The responsiveness of the associated {@link Subscription}s will
      * depend upon the behavior of the {@code stream} below. Make sure the {@link Executor} for this execution chain
      * can tolerate this responsiveness and any blocking behavior.
-     * @param stream provides the data in the form of {@code byte[]} to be emitted to the {@link Subscriber} to the
+     * @param stream provides the data in the form of {@code byte[]} to be emitted to the {@link Subscriber} by the
```
idelpivnitskiy, line 2916: 👁️ 🦅

idelpivnitskiy (approved):
Looks great!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
idelpivnitskiy: Looks like I can merge the suggestion from the GH UI, let me try :)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
```diff
@@ -62,14 +63,29 @@
...
+     *
+     * @param stream the {@link InputStream} to expose as a {@link Publisher}
+     * @param readChunkSize the maximum length of {@code byte[]} chunks which will be read from the {@link InputStream}
+     * and emitted by the {@link Publisher}.
+     */
+    FromInputStreamPublisher(final InputStream stream, final int readChunkSize) {
         this.stream = requireNonNull(stream);
+        if (readChunkSize <= 0) {
+            throw new IllegalArgumentException("readChunkSize: " + readChunkSize + " (expected: >0)");
+        }
```
idelpivnitskiy, line 88: 🤦‍♂️ need some practice with GH suggestions
```suggestion
        }
        this.readChunkSize = readChunkSize;
```

idelpivnitskiy:
Thank you @CamW!

PR #1422, opened 2021-03-10
Update Netty gradle dependencies to runtimeOnly

Scottmitch:
Motivation:
We currently have `implementation` dependencies on Netty's native
transport modules with the architecture qualifier. However we only need
a runtime dependency on these artifacts. This isn't expected to change
behavior from an end user perspective but will allow for local Netty
development on one specific architecture without having all native
artifacts (e.g. local snapshot builds).

Modifications:
- All `implementation` scoped dependencies on Netty's native modules
  are divided into two dependencies. An `implementation` which has no
  scope/architecture qualifier (for java compilation), and a
  `runtimeOnly` which does have the scope/architecture qualifier.

Result:
Gradle dependencies on Netty's native modules are more correct, allowing
for easier local Netty development.

### DIFF
--- a/servicetalk-grpc-netty/build.gradle
+++ b/servicetalk-grpc-netty/build.gradle
@@ -51,7 +51,6 @@ dependencies {
   }
   testImplementation "io.grpc:grpc-protobuf:$grpcVersion"
   testImplementation "io.grpc:grpc-stub:$grpcVersion"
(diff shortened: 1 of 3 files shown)


bondolo (approved):

idelpivnitskiy (approved):
🚀

PR #1426, opened 2021-03-11
Deprecate MultiAddress and Partitioned client builder methods

Scottmitch:
Motivation:
PartitionedHttpClientBuilder and MultiAddressHttpClientBuilder replicate
methods that exist on SingleAddressHttpClientBuilder as well as
providing a mechanism to directly access the
SingleAddressHttpClientBuilder for each new client to configure it. We
only need one of these mechanisms and the direct access to
SingleAddressHttpClientBuilder reduces duplication and is the more
flexible approach.

Modifications:
- Deprecate all methods on PartitionedHttpClientBuilder and
  MultiAddressHttpClientBuilder which can be otherwise controlled via
  appendClientBuilderFilter(..).
- Deprecate PartitionHttpClientBuilderConfigurator in favor of
  PartitionedHttpClientBuilder.SingleAddressInitializer which is now
  consistent with MultiAddressHttpClientBuilder.SingleAddressInitializer

Result:
Duplicate methods on PartitionedHttpClientBuilder and
MultiAddressHttpClientBuilder have been deprecated indicating the
preferred alternative.

Deprecated methods will be removed in a future release.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
@@ -49,12 +49,12 @@ import static java.util.Objects.requireNonNull;
 public abstract class MultiAddressHttpClientBuilder<U, R>
         extends HttpClientBuilder<U, R, ServiceDiscovererEvent<R>> {
     /**
-     * Provides a method which can customize {@link SingleAddressHttpClientBuilder} for each new client.
+     * Initializes the {@link SingleAddressHttpClientBuilder} for each new client.
      * @param <U> The unresolved address type.
      * @param <R> The resolved address type.
      */
     @FunctionalInterface
-    public interface SingleAddressConfigurator<U, R> {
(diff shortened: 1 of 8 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
```diff
@@ -49,70 +49,126 @@
...
         extends HttpClientBuilder<U, R, ServiceDiscovererEvent<R>> {
     /**
-     * Provides a method which can customize {@link SingleAddressHttpClientBuilder} for each new client.
+     * Initializes the {@link SingleAddressHttpClientBuilder} for each new client.
      * @param <U> The unresolved address type.
      * @param <R> The resolved address type.
      */
     @FunctionalInterface
-    public interface SingleAddressConfigurator<U, R> {
+    public interface SingleAddressInitializer<U, R> {
```
Scottmitch, line 57: this type hasn't been released yet, so no need to `@deprecate`

idelpivnitskiy (approved):
🚀

PR #1429, opened 2021-03-12
Fix NPE when `TRACE` request doesn't have content-length header for HTTP/2

idelpivnitskiy:
Motivation:

`H2ToStH1ServerDuplexHandler` uses incorrect `>=` check instead of `!= null`
for the nullable `Long` content-length variable.

Modifications:

- Reproduce the scenario in the test;
- Use `contains` to verify if the CL header is present;
- Rename `canAddResponseTransferEncodingProtocol` -> `serverMaySendPayloadBodyFor`;
- In `AbstractH2DuplexHandler` implementations do not add `content-length: 0`
- when it's not expected;

Result:

No NPE when HTTP/2 server receives `TRACE` request.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
@@ -47,8 +47,8 @@ import static io.servicetalk.http.api.HttpResponseMetaDataFactory.newResponseMet
 import static io.servicetalk.http.api.HttpResponseStatus.StatusClass.INFORMATIONAL_1XX;
 import static io.servicetalk.http.netty.H2ToStH1Utils.h1HeadersToH2Headers;
 import static io.servicetalk.http.netty.H2ToStH1Utils.h2HeadersSanitizeForH1;
-import static io.servicetalk.http.netty.HeaderUtils.canAddResponseTransferEncodingProtocol;
 import static io.servicetalk.http.netty.HeaderUtils.responseMayHaveContent;
+import static io.servicetalk.http.netty.HeaderUtils.serverMaySendPayloadBodyFor;
 
 final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
     private boolean readHeaders;
@@ -166,17 +166,16 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
         h2HeadersSanitizeForH1(h2Headers);
         if (httpStatus != null) {
             final int statusCode = httpStatus.code();
-            final Long contentLength = h2Headers.getLong(CONTENT_LENGTH);
-            if (contentLength == null) {
-                if (fullResponse) {
-                    if (responseMayHaveContent(statusCode, method)) {
+            if (!h2Headers.contains(CONTENT_LENGTH)) {
+                if (serverMaySendPayloadBodyFor(statusCode, method)) {
+                    if (fullResponse) {
                         h2Headers.set(CONTENT_LENGTH, ZERO);
+                    } else {
+                        h2Headers.add(TRANSFER_ENCODING, CHUNKED);
                     }
-                } else if (canAddResponseTransferEncodingProtocol(statusCode, method)) {
-                    h2Headers.add(TRANSFER_ENCODING, CHUNKED);
                 }
             } else if (!responseMayHaveContent(statusCode, method)) {
-                throw new IllegalArgumentException("content-length (" + contentLength +
+                throw new IllegalArgumentException("content-length (" + h2Headers.getLong(CONTENT_LENGTH) +
                         ") header is not expected for status code " + statusCode + " in response to " + method.name() +
                         " request");
             }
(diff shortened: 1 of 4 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ServerDuplexHandler.java
```diff
@@ -143,7 +143,7 @@ private NettyH2HeadersToHttpHeaders h2HeadersToH1HeadersServer(Http2Headers h2He
                         h2Headers.add(TRANSFER_ENCODING, CHUNKED);
                     }
                 }
-            } else if (contentLength >= 0L) {
+            } else if (contentLength != null) {
```
Scottmitch, line 146: since we don't need the value lets use `contains` to avoid serialization.

```diff
diff --git a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
index 61b22e7f3..7f90cfd10 100644
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
@@ -166,8 +166,8 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
         h2HeadersSanitizeForH1(h2Headers);
         if (httpStatus != null) {
             final int statusCode = httpStatus.code();
-            final Long contentLength = h2Headers.getLong(CONTENT_LENGTH);
-            if (contentLength == null) {
+            final boolean containsContentLength = h2Headers.contains(CONTENT_LENGTH);
+            if (!containsContentLength) {
                 if (fullResponse) {
                     if (responseMayHaveContent(statusCode, method)) {
                         h2Headers.set(CONTENT_LENGTH, ZERO);
@@ -176,9 +176,8 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
                     h2Headers.add(TRANSFER_ENCODING, CHUNKED);
                 }
             } else if (!responseMayHaveContent(statusCode, method)) {
-                throw new IllegalArgumentException("content-length (" + contentLength +
-                        ") header is not expected for status code " + statusCode + " in response to " + method.name() +
-                        " request");
+                throw new IllegalArgumentException("content-length header is not expected for status code " +
+                        statusCode + " in response to " + method.name() + " request");
             }
         }
         return new NettyH2HeadersToHttpHeaders(h2Headers, headersFactory.validateCookies());

```

Scottmitch: also same issue in `H2ToStH1ServerDuplexHandler` IIUC?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ServerDuplexHandler.java
```diff
@@ -134,17 +135,17 @@ private NettyH2HeadersToHttpHeaders h2HeadersToH1HeadersServer(Http2Headers h2He
...
+                    if (fullRequest && shouldAddZeroContentLength(httpMethod)) {
                         h2Headers.set(CONTENT_LENGTH, ZERO);
                     } else {
                         h2Headers.add(TRANSFER_ENCODING, CHUNKED);
                     }
                 }
-            } else if (contentLength >= 0L) {
-                throw new IllegalArgumentException("content-length (" + contentLength +
+            } else if (containsContentLength) {
+                throw new IllegalArgumentException("content-length (" + h2Headers.getLong(CONTENT_LENGTH) +
```
Scottmitch, line 148: just do `h2Headers.get(CONTENT_LENGTH)` bcz we are using it as a string here anyways.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
```diff
@@ -166,17 +166,16 @@ private NettyH2HeadersToHttpHeaders h2HeadersToH1HeadersClient(Http2Headers h2He
...
                         h2Headers.set(CONTENT_LENGTH, ZERO);
+                    } else {
+                        h2Headers.add(TRANSFER_ENCODING, CHUNKED);
                     }
-                } else if (canAddResponseTransferEncodingProtocol(statusCode, method)) {
-                    h2Headers.add(TRANSFER_ENCODING, CHUNKED);
                 }
             } else if (!responseMayHaveContent(statusCode, method)) {
-                throw new IllegalArgumentException("content-length (" + contentLength +
+                throw new IllegalArgumentException("content-length (" + h2Headers.getLong(CONTENT_LENGTH) +
```
Scottmitch, line 178: just do `h2Headers.get(CONTENT_LENGTH)` bcz we are using it as a string here anyways.

Scottmitch (approved):

PR #1430, opened 2021-03-12
Allow client security configuration without trust material

idelpivnitskiy:
Motivation:

Trust material is optional for the client-side security configuration
and we should allow using the default JDK truststore.

Modifications:

- If users of deprecated `ClientSecurityConfig` did not specify a
trust material, use default `ClientSslConfigBuilder` ctor;

Result:

Users of deprecated `ClientSecurityConfig` can create a security config
with default JDK truststore.


idelpivnitskiy:
The fix is trivial, I will pull this in.

PR #1431, opened 2021-03-12
Add Single.zipWith and Single.zip operators

Scottmitch:
Motivation:
Executing multiple async operations in parallel and combining them in a
type safe way is a common use case, and referred to as zip [1]. We do
not support a zip operator which requires a careful sequence of
operators to replicate.

[1] http://reactivex.io/documentation/operators/zip.html

Modifications:
- Add Single.zipWith
- Add Single.zip overload for 3, 4, and varargs amount of Singles

Result:
Single.zip and Single.zipWith operators exist.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1863,6 +1888,118 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
...
+     * @param <T1> The type for the first {@link Single}.
+     * @param <T2> The type for the second {@link Single}.
+     * @param <T3> The type for the third {@link Single}.
+     * @param <T4> The type for the forth {@link Single}.
+     * @param <R> The result type of the combinator.
+     * @return a new {@link Single} that emits the results of a specified combinator {@link Function} to items emitted
+     * by {@code singles}.
+     * @see <a href="http://reactivex.io/documentation/operators/zip.html">ReactiveX zip operator.</a>
+     */
+    public static <T1, T2, T3, T4, R> Single<R> zip(
```
bondolo, line 1974: Do we think the 3 and 4 cases are going to be common enough to warrant not jumping immediately from 2 -> many? I usually think of these things in terms of method references. What existing functions with the correct arity are likely to be reused as the zipper? If no functions exist, would the new functions be written only for use a zippers or otherwise be useful. 
Some libraries go to even higher arity than four to get the compile type safety.  The typical var args implementation would be purpose built to size check the input array and then read and cast args invoking some method to produce the R type.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleZipper.java
```diff
@@ -0,0 +1,113 @@
...
+    @SuppressWarnings("unchecked")
+    static <T1, T2, T3, T4, R> Single<R> zip(
+            Single<? extends T1> s1, Single<? extends T2> s2, Single<? extends T3> s3, Single<? extends T4> s4,
+            Function4<? super T1, ? super T2, ? super T3, ? super T4, ? extends R> zipper) {
+        return from(s1.map(v -> new ZipArg(0, v)),
+                    s2.map(v -> new ZipArg(1, v)),
+                    s3.map(v -> new ZipArg(2, v)),
+                    s4.map(v -> new ZipArg(3, v)))
+                .flatMapMergeSingle(identity(), 4)
+                .collect(() -> new ArrayList<ZipArg>(4), (list, zipArg) -> {
```
bondolo, line 71: perhaps Collectors.toCollection(ArrayList[]::new)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleZipper.java
```diff
@@ -0,0 +1,113 @@
...
+            mappedSingles[i] = singles[i].map(v -> new ZipArg(finalI, v));
+        }
+        return from(mappedSingles)
+                .flatMapMergeSingle(identity(), mappedSingles.length)
+                .collect(() -> new ArrayList<ZipArg>(mappedSingles.length), (list, element) -> {
+                    list.add(element);
+                    return list;
+                }).map(list -> {
+                    list.sort(ZIP_ARG_COMPARATOR);
+                    Object[] result = new Object[list.size()];
```
bondolo, line 95: perhaps stream().map(e -> e.value).toArray(Object[]::new);

bondolo (commented):
That came together pretty quickly!

Scottmitch:

Scottmitch:

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
Scottmitch: > Do we think the 3 and 4 cases are going to be common enough

Usage is difficult to predict but other rx frameworks do have 8+ arity overrides for this method. The fall-back `n`-arity lack-of generics is pretty verbose/cast-heavy in this case, imho I think it is worth a few extra interfaces to reduce the scenarios where the fall-back operator is necessary.

> would the new functions be written only for use a zippers or otherwise be useful.

rxjava has similar combinator and it is used for `zip`, `combine*`, `withLatestFrom`. I assume `zip` is the most popular out of those.

bondolo (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Function4.java
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.concurrent.api;
+
+import javax.annotation.Nullable;
+
+/**
+ * A functional interface that accepts 3 arguments and generates a return value.
```
tkountis, line 21: 3 arguments -> 4 arguments

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Function4.java
```diff
@@ -0,0 +1,40 @@
...
+ * @param <T1> The type of the first argument.
+ * @param <T2> The type of the second argument.
+ * @param <T3> The type of the third argument.
+ * @param <T4> The type of the forth argument.
```
tkountis, line 25: forth -> fourth

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -535,6 +536,30 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
...
+     * <pre>{@code
+     *      CompletableFuture<T> f1 = ...; // this
+     *      CompletableFuture<T2> other = ...;
+     *      CompletableFuture.allOf(f1, other).get(); // wait for all futures to complete
+     *      return zipper.apply(f1.get(), other.get());
+     * }</pre>
+     * @param other The other {@link Single} to zip with.
+     * @param zipper Used to combine the completed results for each item from {@code singles}.
+     * @param <T2> The type of {@code other}.
+     * @param <R> The result type of the combinator.
```
tkountis, line 553: Should we name the argument combinator to keep it consistent?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -535,6 +536,30 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
...
+     * @return a new {@link Single} that emits the results of a specified combinator {@link Function} to items emitted
```
tkountis, line 554: Function -> BiFunction

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -535,6 +536,30 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
         return new SingleConcatWithPublisher<>(this, next, executor);
     }
 
+    /**
+     * Create a new {@link Single} that emits the results of a specified combinator {@link Function} to items emitted
```
tkountis, line 540: Function -> BiFunction

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1863,6 +1888,118 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
         return amb(singles);
     }
 
+    /**
+     * Create a new {@link Single} that emits the results of a specified combinator {@link Function} to items emitted
```
tkountis, line 1892: Function -> BiFunction

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1863,6 +1888,118 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
...
+     *      CompletableFuture<T2> f2 = ...; // s2
+     *      CompletableFuture.allOf(f1, f2).get(); // wait for all futures to complete
+     *      return zipper.apply(f1.get(), f2.get());
+     * }</pre>
+     * @param s1 The first {@link Single} to zip.
+     * @param s2 The second {@link Single} to zip.
+     * @param zipper Used to combine the completed results for each item from {@code singles}.
+     * @param <T1> The type for the first {@link Single}.
+     * @param <T2> The type for the second {@link Single}.
+     * @param <R> The result type of the combinator.
```
tkountis, line 1907: Same as above, should we name combinator?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1863,6 +1888,118 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
...
+     * @return a new {@link Single} that emits the results of a specified combinator {@link Function} to items emitted
```
tkountis, line 1908: Function -> BiFunction

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1863,6 +1888,118 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
...
+     * by {@code singles}.
+     * @see <a href="http://reactivex.io/documentation/operators/zip.html">ReactiveX zip operator.</a>
+     */
+    public static <T1, T2, R> Single<R> zip(Single<? extends T1> s1, Single<? extends T2> s2,
+                                            BiFunction<? super T1, ? super T2, ? extends R> zipper) {
+        return SingleZipper.zip(s1, s2, zipper);
+    }
+
+    /**
+     * Create a new {@link Single} that emits the results of a specified combinator {@link Function} to items emitted
```
tkountis, line 1918: Link to relevant Function3. A couple more occurrences below.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1863,6 +1888,118 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
...
+     *      CompletableFuture<T2> f2 = ...; // s2
+     *      CompletableFuture<T3> f3 = ...; // s3
+     *      CompletableFuture<T4> f4 = ...; // s3
+     *      CompletableFuture.allOf(f1, f2, f3, f4).get(); // wait for all futures to complete
+     *      return zipper.apply(f1.get(), f2.get(), f3.get(), f4.get());
+     * }</pre>
+     * @param s1 The first {@link Single} to zip.
+     * @param s2 The second {@link Single} to zip.
+     * @param s3 The third {@link Single} to zip.
+     * @param s4 The forth {@link Single} to zip.
```
tkountis, line 1963: fourth

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1863,6 +1888,118 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
...
+     * @param zipper Used to combine the completed results for each item from {@code singles}.
+     * @param <T1> The type for the first {@link Single}.
+     * @param <T2> The type for the second {@link Single}.
+     * @param <T3> The type for the third {@link Single}.
+     * @param <T4> The type for the forth {@link Single}.
```
tkountis, line 1968: fourth

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleZipper.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private SingleZipper() {
+    }
+
+    @SuppressWarnings("unchecked")
+    static <T1, T2, R> Single<R> zip(Single<? extends T1> s1, Single<? extends T2> s2,
+                                     BiFunction<? super T1, ? super T2, ? extends R> zipper) {
+        return from(s1.map(v -> new ZipArg(0, v)), s2.map(v -> new ZipArg(1, v)))
+                .flatMapMergeSingle(identity(), 2)
+                .collect(ZipResult2::new, (result, zipArg) -> {
```
tkountis, line 39: Can we replace the conditionals with a direct array assignment?
example:
```
static <T1, T2, R> Single<R> zip(Single<? extends T1> s1, Single<? extends T2> s2,
                                 BiFunction<? super T1, ? super T2, ? extends R> zipper) {
    return from(s1.map(v -> new ZipArg(0, v)), s2.map(v -> new ZipArg(1, v)))
            .flatMapMergeSingle(identity(), 2)
            .collect(ZipResult2::new, (result, zipArg) -> {
                result.args[zipArg.index] = zipArg;
                return result;
            }).map(result -> zipper.apply((T1) result.args[0].value, (T2) result.args[1].value));
}
```

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/SingleZip5Test.java
```diff
@@ -0,0 +1,225 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+public class SingleZip5Test {
+    @Rule
+    public final Timeout timeout = new ServiceTalkTestTimeout();
+    private final TestSingle<Integer> first = new TestSingle<>();
+    private final TestSingle<Double> second = new TestSingle<>();
+    private final TestSingle<Short> third = new TestSingle<>();
+    private final TestSingle<Byte> forth = new TestSingle<>();
```
tkountis, line 40: fourth :)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleZipper.java
```diff
@@ -0,0 +1,154 @@
...
+                    list.add(zipArg);
+                    return list;
+                }).map(list -> {
+                    list.sort(ZIP_ARG_COMPARATOR);
+                    return zipper.apply((T1) list.get(0).value, (T2) list.get(1).value, (T3) list.get(2).value,
+                            (T4) list.get(3).value);
+                });
+    }
+
+    static <R> Single<R> zip(Function<? super Object[], ? extends R> zipper, Single<?>... singles) {
```
tkountis, line 106: Likewise for the unbounded version

```
static <R> Single<R> zip(Function<? super Object[], ? extends R> zipper, Single<?>... singles) {
    Single<ZipArg>[] mappedSingles = new Single[singles.length];
    for (int i = 0; i < singles.length; ++i) {
        final int finalI = i;
        mappedSingles[i] = singles[i].map(v -> new ZipArg(finalI, v));
    }
    return from(mappedSingles)
            .flatMapMergeSingle(identity(), mappedSingles.length)
            .collect(() -> new Object[singles.length], (list, element) -> {
                list[element.index] = element.value;
                return list;
            }).map(zipper);
}
```

tkountis (commented):
Some suggestions then lgtm

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
Scottmitch: I'll rename to `zipper` for consistency.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleZipper.java
Scottmitch: > Can we replace the conditionals with a direct array assignment?

+1. good suggestion!

> nit: you could probably access the arr elements in reverse order and store them in local fields before calling the zipper, to minimize array bound checks.

I'm not sure I follow. can you drop another comment after I submit the updates if I don't address this?

tkountis: What I meant is this:

```
final Object c = result.args[2].value;
final Object b = result.args[1].value;
final Object a = result.args[0].value;

zipper.apply((T1) a, (T2) b, (T3) c))
```

This should generate code that checks array boundaries only once (1st time) and make the subsequent accesses cheaper (if JIT doesn't handle this case already). This of course, at the cost of extra code, which may confuse readers, so not sure its worth it here.

tkountis (approved):
🚀 🚀 Awesome work!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleZipper.java
Scottmitch: neat. lets proceed without this for now and we can submit a follow-up to use this optimization if it proves necessary/beneficial.

PR #1433, opened 2021-03-15
Skip grpc-accept-encoding when the only option is identity

tkountis:
Motivation:
`grpc-accept-encoding` and `grpc-encoding` headers can be skipped when the only value associated with is `identity`. The spec allows this relaxation and we can save processing and wire time.

Modifications:
Skip the relevant headers when the value is only `identity`.

Result:
Less headers and less processing done to handle unnecessary data.

### DIFF
--- a/servicetalk-encoding-api-internal/src/main/java/io/servicetalk/encoding/api/internal/HeaderUtils.java
+++ b/servicetalk-encoding-api-internal/src/main/java/io/servicetalk/encoding/api/internal/HeaderUtils.java
@@ -17,7 +17,7 @@ package io.servicetalk.encoding.api.internal;
 
 import io.servicetalk.buffer.api.CharSequences;
 import io.servicetalk.encoding.api.ContentCodec;
-import io.servicetalk.encoding.api.ContentCodings;
+import io.servicetalk.encoding.api.Identity;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -43,13 +43,13 @@ public final class HeaderUtils {
      * <p>
      * If no supported encodings are configured then the result is always {@code null}
      * If no accepted encodings are present in the request then the result is always {@code null}
-     * In all other cases, the first matching encoding (that is NOT {@link ContentCodings#identity()}) is preferred,
+     * In all other cases, the first matching encoding (that is NOT {@link Identity#identity()}) is preferred,
      * otherwise {@code null} is returned.
      *
      * @param acceptEncodingHeaderValue The accept encoding header value.
      * @param serverSupportedEncodings The server supported codings as configured.
      * @return The {@link ContentCodec} that satisfies both client and server needs,
-     * null if none found or matched to {@link ContentCodings#identity()}
+     * null if none found or matched to {@link Identity#identity()}
      */
     @Nullable
     public static ContentCodec negotiateAcceptedEncoding(@Nullable final CharSequence acceptEncodingHeaderValue,
@@ -71,13 +71,13 @@ public final class HeaderUtils {
      * on the server side and the incoming header on the request.
      * <p>
      * If no supported encodings are passed then the result is always {@code null}
-     * Otherwise, the first matching encoding (that is NOT {@link ContentCodings#identity()}) is preferred,
+     * Otherwise, the first matching encoding (that is NOT {@link Identity#identity()}) is preferred,
      * or {@code null} is returned.
      *
      * @param clientSupportedEncodings The client supported codings as found in the HTTP header.
      * @param serverSupportedEncodings The server supported codings as configured.
      * @return The {@link ContentCodec} that satisfies both client and server needs,
-     * null if none found or matched to {@link ContentCodings#identity()}
+     * null if none found or matched to {@link Identity#identity()}
      */
     @Nullable
     public static ContentCodec negotiateAcceptedEncoding(final List<ContentCodec> clientSupportedEncodings,
@@ -120,7 +120,7 @@ public final class HeaderUtils {
      * Returns the {@link ContentCodec} that matches the {@code name} within the {@code allowedList}.
      * if {@code name} is {@code null} or empty it results in {@code null} .
      * If {@code name} is {@code 'identity'} this will always result in
-     * {@link ContentCodings#identity()} regardless of its presence in the {@code allowedList}.
+     * {@link Identity#identity()} regardless of its presence in the {@code allowedList}.
      *
      * @param allowedList the source list to find a matching codec from.
      * @param name the codec name used for the equality predicate.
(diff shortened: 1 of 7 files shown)


servicetalk-encoding-api-internal/src/main/java/io/servicetalk/encoding/api/internal/HeaderUtils.java
```diff
@@ -43,13 +43,13 @@ private HeaderUtils() {
      * <p>
      * If no supported encodings are configured then the result is always {@code null}
      * If no accepted encodings are present in the request then the result is always {@code null}
-     * In all other cases, the first matching encoding (that is NOT {@link ContentCodings#identity()}) is preferred,
```
bondolo, line 46: Worth deprecating ContentCodings#identity() ?

tkountis: The class `ContentCodings` is deprecated.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -284,9 +285,10 @@ private static void initResponse(final HttpResponseMetaData response, @Nullable
         final HttpHeaders headers = response.headers();
         headers.set(SERVER, GRPC_USER_AGENT);
         headers.set(CONTENT_TYPE, GRPC_CONTENT_TYPE);
-        if (context != null) {
-            headers.set(GRPC_ACCEPT_ENCODING_KEY,
-                    acceptedEncodingsHeaderValueOrCached(context.supportedMessageCodings()));
+        final CharSequence acceptedEncoding = context == null ? null :
```
Scottmitch, line 288: consider avoiding a `null` conditional if not necessary:
```suggestion
          if (context != null) {
            final CharSequence acceptedEncoding =
                    acceptedEncodingsHeaderValueOrCached(context.supportedMessageCodings());
            if (acceptedEncoding != null) {
                headers.set(GRPC_ACCEPT_ENCODING_KEY, acceptedEncoding);
            }
        }
```

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -347,7 +352,7 @@ private static CharSequence acceptedEncodingsHeaderValue0(final List<ContentCode
             builder.append(codec.name());
         }
 
-        return newAsciiString(builder.toString());
+        return builder.length() > 0 ? newAsciiString(builder.toString()) : null;
```
Scottmitch, line 355: suggestions:
- avoid conditional inside loop
- pre-size the string builder to reduce chances of copy/resize

```java
private static final String CONTENT_ENCODING_SEPARATOR = ", ";
```
```suggestion
        StringBuilder builder = new StringBuilder(codings.size() * (12 + CONTENT_ENCODING_SEPARATOR.length()));
        for (ContentCodec codec : codings) {
            if (codec == identity()) {
                continue;
            }

            builder.append(codec.name()).append(CONTENT_ENCODING_SEPARATOR);
        }
        if (builder.length() > CONTENT_ENCODING_SEPARATOR.length()) {
            builder.setLength(builder.length() - CONTENT_ENCODING_SEPARATOR.length());
            return newAsciiString(builder.toString());
        }
        return null;
```

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcMessageEncodingTest.java
```diff
@@ -421,16 +413,19 @@ private TesterClient newClient(@Nullable final List<ContentCodec> supportedCodin
...
+            throwAsyncErrors();
+            throw t;
         }
-
-        verifyNoErrors();
     }
 
-    private void verifyNoErrors() throws Throwable {
+    private void throwAsyncErrors() throws Throwable {
         if (!errors.isEmpty()) {
```
Scottmitch, line 429: nit: make this a concurrent queue to replace `empty` / `get(0)` (sensitive to race conditions) with `Object o = queue.poll()`

servicetalk-encoding-api-internal/src/main/java/io/servicetalk/encoding/api/internal/HeaderUtils.java
Scottmitch: does this PR remove the last usage of the class?

Scottmitch (approved):
few comments/suggestions then lgtm

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcMessageEncodingTest.java
tkountis: Changed to keep consistent with other test classes. FWIW, there is no race here, the List was `synchronized` and the `throw` is called after the Futures are resolved.

tkountis:
> does this PR remove the last usage of the class?
@Scottmitch yes, there were a few places that weren't caught in the previous refactoring. I fixed them here.

servicetalk-encoding-api-internal/src/main/java/io/servicetalk/encoding/api/internal/HeaderUtils.java
tkountis: yes, there were a few places that weren't caught in the previous refactoring. I fixed them here.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -324,30 +329,33 @@ private static Status getStatusDetails(final HttpHeaders headers) {
...
     private static CharSequence acceptedEncodingsHeaderValueOrCached(final List<ContentCodec> codings) {
         return ENCODINGS_HEADER_CACHE.computeIfAbsent(codings, (__) -> acceptedEncodingsHeaderValue0(codings));
     }
 
+    @Nullable
     private static CharSequence acceptedEncodingsHeaderValue0(final List<ContentCodec> codings) {
-        StringBuilder builder = new StringBuilder();
+        StringBuilder builder = new StringBuilder(codings.size() * (12 + CONTENT_ENCODING_SEPARATOR.length()));
         for (ContentCodec codec : codings) {
             if (codec == identity()) {
```
idelpivnitskiy, line 347: Users may have their own version of "identity" codec. We should implement `equals` and `hashCode` for `IdentityContentCodec` then use `identity().equals(codec)` here and in `HeaderUtils` for HTTP.

tkountis: Agreed, issue: https://github.com/apple/servicetalk/issues/1455

PR #1435, opened 2021-03-16
Publisher, Single, Completable onError* operators

Scottmitch:
Motivation:
ServiceTalk supports a recoverWith operator which allows for error
recovery [1]. However the signature of this method is more complex than
is often required for simple error recovery, and the name implies
"recovery" which is confusing if the return value is intended to
propagate an error.

[1] http://reactivex.io/documentation/operators/catch.html

Modifications:
- Add onError* methods to Publisher, Single, and Completable. The
  signatures are specialized for their respective operation (mapping
  Throwable, swallowing exceptions, recovering with a value) and also
  supports filtering by class type and custom Predicate. This naming
  convention is more commonly used by other JVM based rx implementations
  and is assumed to be more familiar to users.
- Deprecate recoverWith in favor of onErrorResume

Result:
Publisher, Single, and Completable have onError* operators which provide
more APIs tailored toward common error recovery scenarios.


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -402,16 +401,16 @@ public Completable closeAsyncGracefully() {
                         @Override
                         public Publisher<Resp> handle(final GrpcServiceContext ctx, final Publisher<Req> request) {
                             return request.firstOrError()
-                                    .recoverWith(t -> {
+                                    .onErrorMap(t -> {
                                         if (t instanceof NoSuchElementException) {
-                                            return failed(new GrpcStatus(INVALID_ARGUMENT, null,
+                                            return new GrpcStatus(INVALID_ARGUMENT, null,
```
bondolo, line 406: Worth it to initialize the cause?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/OnErrorResumeCompletable.java
```diff
@@ -89,7 +89,6 @@ public void onError(Throwable throwable) {
             try {
                 next = predicate.test(throwable) ? requireNonNull(nextFactory.apply(throwable)) : null;
             } catch (Throwable t) {
-                t.addSuppressed(throwable);
```
bondolo, line 96: Would the supressed be likely to have a different stack root? ie. from another thread with more context? It seems fairly low cost to keep it around unless the suppressed exception is truly redundant.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
Scottmitch: yah good point, not sure if there is a reason for it to be `null`. @idelpivnitskiy - wdyt?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/OnErrorResumeCompletable.java
Scottmitch: I will revert this change.

I think we need to do a more general investigation into `addSuppressed` usage which has the following potential issues:
- we don't use `addSuppressed` consistently
- it creates memory leaks if static/shared exceptions are thrown
- restrictions on usage such as self-add result in throwing which breaks async control flow

Scottmitch:

bondolo (approved):

Scottmitch:

Scottmitch:

Scottmitch:
another failure attributed to https://github.com/apple/servicetalk/issues/1200

PR #1436, opened 2021-03-16
Remove jcenter dependencies from build

Scottmitch:
Motivation:
jcenter is deprecating bintray, and it will soon stop serving artifacts
eventually.

Modifications:
- Remove all references to jcenter from the build

Result:
No more direct references to jcenter from the build avoids dependency on
soon to be decommissioned repository.


Scottmitch:
I will merge this now since jcenter is having a service outage.

idelpivnitskiy:
Late LGTM, thanks!

Issue #1323, opened 2021-01-14
ConnectionCloseHeaderHandlingTest$PipelinedRequestsTest.serverCloseTwoPipelinedRequestsSentBeforeFirstResponse test failure


Scottmitch:
Possibly related: https://github.com/apple/servicetalk/issues/1154, https://github.com/apple/servicetalk/issues/1277

```
ConnectionCloseHeaderHandlingTest$PipelinedRequestsTest.serverCloseTwoPipelinedRequestsSentBeforeFirstResponse[2: useUds=false, viaProxy=true, awaitRequestPayload=false]

java.util.concurrent.ExecutionException: io.servicetalk.transport.netty.internal.CloseHandler$CloseEventObservedException: PROTOCOL_CLOSING_INBOUND(The application protocol closed the read side of this connection. This maybe the result of sending an HTTP header such as Connection: close.) [id: 0xc365f90b, L:/127.0.0.1:50910 - R:localhost/127.0.0.1:36781]
	at io.servicetalk.concurrent.api.SourceToFuture.reportGet(SourceToFuture.java:121)
	at io.servicetalk.concurrent.api.SourceToFuture.get(SourceToFuture.java:92)
	at io.servicetalk.http.netty.ConnectionCloseHeaderHandlingTest$ConnectionSetup.assertResponsePayloadBody(ConnectionCloseHeaderHandlingTest.java:226)
	at io.servicetalk.http.netty.ConnectionCloseHeaderHandlingTest$PipelinedRequestsTest.serverCloseTwoPipelinedRequestsSentBeforeFirstResponse(ConnectionCloseHeaderHandlingTest.java:355)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
[...]

PR #1440, opened 2021-03-16
Deprecate ContentCodec API with offest & length

tkountis:
Motivation:
Deprecate the `ContentCodec` APIs that receive offset & length. User's can slice if needed.

Modifications:
Deprecate API & relevant changes to accommodate the change.

Result:
Less API surface.

### DIFF
--- a/servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
+++ b/servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
@@ -63,6 +63,11 @@ final class NettyChannelContentCodec extends AbstractContentCodec {
     }
 
     @Override
+    public Buffer encode(final Buffer src, final BufferAllocator allocator) {
+        return encode(src, 0, src.readableBytes(), allocator);
+    }
+
+    @Override
     public Buffer encode(final Buffer src, final int offset, final int length, final BufferAllocator allocator) {
         requireNonNull(allocator);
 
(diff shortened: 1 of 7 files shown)


tkountis:
Build failure due https://github.com/apple/servicetalk/issues/1323

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyChannelContentCodec.java
```diff
@@ -177,6 +182,11 @@ public void onComplete() {
                 });
     }
 
+    @Override
+    public Buffer decode(final Buffer src, final BufferAllocator allocator) {
+        return decode(src, 0, src.readableBytes(), allocator);
```
idelpivnitskiy, line 187: Keep the `AbstractZipContentCodec` as it's right now in this PR, but for `NettyChannelContentCodec` consider doing the opposite: implement `decoder(Buffer, BufferAllocator)` and do the slice in this overload.

servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyChannelContentCodecTest.java
```diff
@@ -76,41 +76,31 @@ public NettyChannelContentCodecTest(final ContentCodec codec) {
...
     }
 
     @Test(expected = CodecEncodingException.class)
     public void testEncodeWithOffsetAndZeroLength() {
-        testEncode(DEFAULT_ALLOCATOR, 10, 0);
-    }
-
-    @Test(expected = CodecEncodingException.class)
-    public void testEncodeWithOffsetAndOverflowLength() {
-        testEncode(DEFAULT_ALLOCATOR, 1023, 0);
```
idelpivnitskiy, line 94: Not necessary for this PR (I do not expect anyone uses codec api directly), but for future cases when you deprecate API, consider removing tests at the time you remove the method. It helps to make sure the old API still works as expected before we remove it.

idelpivnitskiy (approved):
One comment for `NettyChannelContentCodec` impl then LGTM

PR #1442, opened 2021-03-16
Github Actions Add CodeQL Analysis

Scottmitch:
Motivation:
Add CodeQL to CI workflow to get additional static code analysis.

Modifications:
- Add codeql-analysis github actions work flow which runs on PRs and
  pushes to the main branch

Result:
More static code analysis.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -6,7 +6,7 @@ on:
     # The branches below must be a subset of the branches above
     branches: [ main ]
   schedule:
-    - cron: '0 13 * * 3'
+    - cron: '0 13 * * 1'
 
 jobs:
   analyze:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,29 @@
+name: CodeQL
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '0 13 * * 3'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'java' ]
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,29 @@
+name: CodeQL
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '0 13 * * 3'
```
bondolo, line 9: Perhaps Tuesdays since our releases are scheduled for Wednesdays.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,29 @@
...
+        language: [ 'java' ]
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v1
+        with:
+          languages: ${{ matrix.language }}
+      - name: Execute gradle build
+        run: ./gradlew --parallel --max-workers=4 clean assemble
```
bondolo, line 27: Any way to make the number of workers automatic?

bondolo (approved):

.github/workflows/codeql-analysis.yml
Scottmitch: good idea. will move to Monday.

.github/workflows/codeql-analysis.yml
Scottmitch: IIRC this was primarily for our older CI systems that were relatively under powered. This config is on all our builds currently so I'll leave it in this PR for consistency but we could experiment by removing this param as a followup.

```bash
--parallel                         Build projects in parallel. Gradle will attempt to determine the optimal number of executor threads to use.
```

Scottmitch:
build failure due to https://github.com/apple/servicetalk/issues/1192 and https://github.com/apple/servicetalk/issues/999

PR #1443, opened 2021-03-16
RangeIntPublisher add explicit cast to int

Scottmitch:
Motivation:
RangeIntPublisher relies upon an implict case from long to int which is
gaurded by a `min(int, long)` (where the int is non-negative). CodeQL
highlights this as a warning.

Modifications:
- Add an explicit cast to int to be more clear this is intentional

Result:
CodeQL warning is resovled.


Scottmitch:
Here are the warnings found: https://github.com/apple/servicetalk/security/code-scanning?query=ref%3Arefs%2Fpull%2F1442%2Fmerge+tool%3ACodeQL

Scottmitch:
trivial change, will merge.

idelpivnitskiy:
The change LGTM, but while I was looking at `range`, why we do not allow a negative stride? Python allows something like `range(25, 2, -2)`: https://www.geeksforgeeks.org/python-range-function/

PR #1444, opened 2021-03-16
Github Actions Remove --max-workers gradle param

Scottmitch:
Motivation:
Our github action scripts explicitly set --max-workers but gradle can
automatically infer the value based upon number of processors [1].

[1] https://docs.gradle.org/current/userguide/command_line_interface.html
> Default is number of processors.

Modifications:
- Remove --max-workers param from github action scripts

Result:
Gradle will automatically determine the number of workers to use based
upon available processors.

### DIFF
--- a/.github/workflows/ci-snapshot.yml
+++ b/.github/workflows/ci-snapshot.yml
@@ -63,7 +63,7 @@ jobs:
           fi
 
           # Execute the gradlew command to publish the build
-          ./gradlew --parallel --max-workers=4$FIRST_GRADLE_TARGETS && ./gradlew$SECOND_GRADLE_TARGETS
+          ./gradlew --parallel$FIRST_GRADLE_TARGETS && ./gradlew$SECOND_GRADLE_TARGETS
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-snapshot.yml
```diff
@@ -63,7 +63,7 @@ jobs:
           fi
 
           # Execute the gradlew command to publish the build
-          ./gradlew --parallel --max-workers=4$FIRST_GRADLE_TARGETS && ./gradlew$SECOND_GRADLE_TARGETS
+          ./gradlew --parallel$FIRST_GRADLE_TARGETS && ./gradlew$SECOND_GRADLE_TARGETS
```
Scottmitch, line 66: note the lack of space is intentional as the variable value always starts with a space

bondolo (approved):

Scottmitch:
I've done 3-4 builds and build time doesn't seem to be impacted. I'll merge and observe if build times change at all over time.

PR #1445, opened 2021-03-17
Add Publisher.timeoutTerminal(Duration) operator

bondolo:
Motivation:
The existing `Publisher.timeout(Duration)` operator restarts the timer for each received item and will only timeout if no items are emitted during the timeout duration. In some cases the total time for all items to be emitted is of more interest. An operator to timeout based on the entire elapsed time since subscribe is needed.

Modifications:
Add a new Publisher operator, `timeoutTerminal(Duration)`, that terminates the Publisher with a TimeoutException if the Publisher does not terminate before the specified time duration.

Result:
A new timeout option is available.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -79,6 +79,15 @@ import static java.util.Objects.requireNonNull;
 public abstract class Publisher<T> {
     private static final Logger LOGGER = LoggerFactory.getLogger(Publisher.class);
 
+    /**
+     * Maximum positive duration which can be expressed as a signed 64-bit number of nanoseconds.
+     */
+    private static final Duration LONG_MAX_NANOS = Duration.ofNanos(Long.MAX_VALUE);
+    /**
+     * Maximum negative duration which can be expressed as a signed 64-bit number of nanoseconds.
+     */
+    private static final Duration LONG_MIN_NANOS = Duration.ofNanos(Long.MIN_VALUE);
+
     private final Executor executor;
     private final boolean shareContextOnSubscribe;
 
@@ -1636,6 +1645,19 @@ public abstract class Publisher<T> {
     }
 
     /**
+     * Converts a {@code Duration} to nanoseconds or if the resulting value would overflow a 64-bit signed integer then
+     * either {@code Long.MIN_VALUE} or {@code Long.MAX_VALUE} as appropriate.
+     *
+     * @param duration The duration to convert
+     * @return The converted nanoseconds value.
+     */
+    private long toNanos(Duration duration) {
+        return duration.compareTo(LONG_MAX_NANOS) < 0 ?
+                duration.compareTo(LONG_MIN_NANOS) > 0 ? duration.toNanos() : Long.MIN_VALUE
+                : Long.MAX_VALUE;
+    }
+
+    /**
      * Creates a new {@link Publisher} that will mimic the signals of this {@link Publisher} but will terminate with a
      * {@link TimeoutException} if time {@code duration} elapses between adjacent {@link Subscriber#onNext(Object)}
      * calls. The timer starts when the returned {@link Publisher} is subscribed.
@@ -1650,7 +1672,7 @@ public abstract class Publisher<T> {
      * @see <a href="http://reactivex.io/documentation/operators/timeout.html">ReactiveX timeout operator.</a>
      */
     public final Publisher<T> timeout(Duration duration, io.servicetalk.concurrent.Executor timeoutExecutor) {
-        return timeout(duration.toNanos(), TimeUnit.NANOSECONDS, timeoutExecutor);
+        return timeout(toNanos(duration), TimeUnit.NANOSECONDS, timeoutExecutor);
     }
 
     /**
@@ -1685,7 +1707,7 @@ public abstract class Publisher<T> {
      * @see <a href="http://reactivex.io/documentation/operators/timeout.html">ReactiveX timeout operator.</a>
      */
     public final Publisher<T> timeoutTerminal(Duration duration, io.servicetalk.concurrent.Executor timeoutExecutor) {
-        return timeoutTerminal(duration.toNanos(), TimeUnit.NANOSECONDS, timeoutExecutor);
+        return timeoutTerminal(toNanos(duration), TimeUnit.NANOSECONDS, timeoutExecutor);
     }
 
     /**
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -1654,6 +1654,41 @@ public abstract class Publisher<T> {
     }
 
     /**
+     * Creates a new {@link Publisher} that will mimic the signals of this {@link Publisher} but will terminate with a
+     * {@link TimeoutException} if time {@code duration} elapses between subscribe and termination. The timer starts
+     * when the returned {@link Publisher} is subscribed.
+     * <p>
+     * In the event of timeout any {@link Subscription} from
+     * {@link Subscriber#onSubscribe(PublisherSource.Subscription)} will be {@link Subscription#cancel() cancelled} and
+     * the associated {@link Subscriber} will be {@link Subscriber#onError(Throwable) terminated}.
+     * @param duration The time duration which is allowed to elapse between {@link Subscriber#onNext(Object)} calls.
+     * @return a new {@link Publisher} that will mimic the signals of this {@link Publisher} but will terminate with a
+     * {@link TimeoutException} if time {@code duration} elapses between {@link Subscriber#onNext(Object)} calls.
+     * @see <a href="http://reactivex.io/documentation/operators/timeout.html">ReactiveX timeout operator.</a>
+     */
+    public final Publisher<T> withTimeout(Duration duration) {
+        return withTimeout(duration, executor);
+    }
+
+    /**
+     * Creates a new {@link Publisher} that will mimic the signals of this {@link Publisher} but will terminate with a
+     * {@link TimeoutException} if time {@code duration} elapses between subscribe and termination. The timer starts
+     * when the returned {@link Publisher} is subscribed.
+     * <p>
+     * In the event of timeout any {@link Subscription} from
+     * {@link Subscriber#onSubscribe(PublisherSource.Subscription)} will be {@link Subscription#cancel() cancelled} and
+     * the associated {@link Subscriber} will be {@link Subscriber#onError(Throwable) terminated}.
+     * @param duration The time duration which is allowed to elapse between {@link Subscriber#onNext(Object)} calls.
+     * @param timeoutExecutor The {@link Executor} to use for managing the timer notifications.
+     * @return a new {@link Publisher} that will mimic the signals of this {@link Publisher} but will terminate with a
+     * {@link TimeoutException} if time {@code duration} elapses between {@link Subscriber#onNext(Object)} calls.
+     * @see <a href="http://reactivex.io/documentation/operators/timeout.html">ReactiveX timeout operator.</a>
+     */
+    public final Publisher<T> withTimeout(Duration duration, io.servicetalk.concurrent.Executor timeoutExecutor) {
+        return new TimeoutPublisher<>(this, executor, duration, false, timeoutExecutor);
+    }
+
+    /**
      * Emits items emitted by {@code next} {@link Publisher} after {@code this} {@link Publisher} terminates
      * successfully.
      * <p>
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -1513,7 +1513,7 @@ public abstract class Publisher<T> {
      */
     @Deprecated
     public final Publisher<T> idleTimeout(long duration, TimeUnit unit) {
-        return idleTimeout(duration, unit, executor);
+        return timeout(duration, unit);
     }
 
     /**
@@ -1533,7 +1533,7 @@ public abstract class Publisher<T> {
      */
     @Deprecated
     public final Publisher<T> idleTimeout(Duration duration) {
-        return idleTimeout(duration.toNanos(), TimeUnit.NANOSECONDS);
+        return timeout(duration);
     }
 
     /**
@@ -1555,7 +1555,7 @@ public abstract class Publisher<T> {
     @Deprecated
     public final Publisher<T> idleTimeout(long duration, TimeUnit unit,
                                           io.servicetalk.concurrent.Executor timeoutExecutor) {
-        return new TimeoutPublisher<>(this, executor, duration, unit, true, timeoutExecutor);
+        return timeout(duration, unit, timeoutExecutor);
     }
 
     /**
@@ -1575,7 +1575,7 @@ public abstract class Publisher<T> {
      */
     @Deprecated
     public final Publisher<T> idleTimeout(Duration duration, io.servicetalk.concurrent.Executor timeoutExecutor) {
-        return idleTimeout(duration.toNanos(), TimeUnit.NANOSECONDS, timeoutExecutor);
+        return timeout(duration, timeoutExecutor);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/TimeoutPublisherTest.java
```diff
@@ -68,13 +76,26 @@
...
+    public TimeoutPublisherTest(boolean untilTermination) {
+        this.untilTermination = untilTermination;
+    }
+
     @Before
     public void setup() {
         testExecutor = executorRule.executor();
     }
 
+    @Parameterized.Parameters(name = "{index} - {0}")
```
idelpivnitskiy, line 91: Consider clarifying what the boolean means: `"{index} - reset timeout at onNext: {0}"`

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/TimeoutPublisherTest.java
```diff
@@ -286,11 +307,14 @@ public Cancellable execute(final Runnable task) throws RejectedExecutionExceptio
...
     private void init() {
-        init(publisher, true);
+        init(publisher, true, untilTermination);
     }
 
-    private void init(Publisher<Integer> publisher, boolean expectOnSubscribe) {
-        toSource(publisher.idleTimeout(1, NANOSECONDS, testExecutor)).subscribe(subscriber);
+    private void init(Publisher<Integer> publisher, boolean expectOnSubscribe, boolean untilTermination) {
+        publisher = untilTermination ?
+                publisher.withTimeout(ONE_NANOSECOND, testExecutor)
```
idelpivnitskiy, line 315: Consider adding a test that verifies the `onNext` does not update the timeout

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/SignalOffloaders.java
```diff
@@ -22,6 +22,9 @@
  */
 public final class SignalOffloaders {
 
+    /**
+     * Uses {@link TaskBasedSignalOffloader} for offloading
```
idelpivnitskiy, line 26: Leftover from other work?

idelpivnitskiy (commented):
Minor suggestions and a couple questions to discuss. Otherwise, LGTM

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -1239,6 +1239,41 @@ public final Completable flatMapCompletableDelayError(Function<? super T, ? exte
...
+     * <p>
+     * In the event of timeout any {@link Subscription} from
+     * {@link Subscriber#onSubscribe(PublisherSource.Subscription)} will be {@link Subscription#cancel() cancelled} and
+     * the associated {@link Subscriber} will be {@link Subscriber#onError(Throwable) terminated}.
+     * @param duration The time duration which is allowed to elapse between {@link Subscriber#onNext(Object)} calls.
+     * @return a new {@link Publisher} that will mimic the signals of this {@link Publisher} but will terminate with a
+     * {@link TimeoutException} if time {@code duration} elapses between {@link Subscriber#onNext(Object)} calls.
+     * @see <a href="http://reactivex.io/documentation/operators/timeout.html">ReactiveX timeout operator.</a>
+     */
+    public final Publisher<T> withTimeout(Duration duration) {
```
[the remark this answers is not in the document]
Scottmitch: `timeoutTerminal` (timeout to receive [terminal signal](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#glossary) -> terms from RS spec)?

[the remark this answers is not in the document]
Scottmitch: also be good to rename `idleTimeout` -> `timeout` (consistent with RxJava, reactor, and http://reactivex.io/documentation/operators/timeout.html). I can submit a followup PR.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -1239,6 +1239,41 @@ public final Completable flatMapCompletableDelayError(Function<? super T, ? exte
...
+     * In the event of timeout any {@link Subscription} from
+     * {@link Subscriber#onSubscribe(PublisherSource.Subscription)} will be {@link Subscription#cancel() cancelled} and
+     * the associated {@link Subscriber} will be {@link Subscriber#onError(Throwable) terminated}.
+     * @param duration The time duration which is allowed to elapse between {@link Subscriber#onNext(Object)} calls.
+     * @param timeoutExecutor The {@link Executor} to use for managing the timer notifications.
+     * @return a new {@link Publisher} that will mimic the signals of this {@link Publisher} but will terminate with a
+     * {@link TimeoutException} if time {@code duration} elapses between {@link Subscriber#onNext(Object)} calls.
+     * @see <a href="http://reactivex.io/documentation/operators/timeout.html">ReactiveX timeout operator.</a>
+     */
+    public final Publisher<T> withTimeout(Duration duration, io.servicetalk.concurrent.Executor timeoutExecutor) {
```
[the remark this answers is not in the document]
bondolo: The Java Util Concurrent time units were (and remain) somewhat popular as a way to specify time before the introduction of the Duration unit. They were liked by people who did concurrency stuff but were always greeted with confusion by enterprise types who were unfamiliar and wanted to use `long` with epoch millis or Date. The downside of the juc duration is that they always require two constants.

I am fine with including juc style durations if desired, especially if we think that users would have to make Durations from existing values in order to use the APIs.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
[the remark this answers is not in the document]
bondolo: I don't feel strongly about the name and following existing patterns is always best. "with" was an attempt at verbing timeout and paired with "recoverWith". I think perhaps `timeout` might be best because it will make the distinction between the new api and the existing idle version more clear.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
```diff
@@ -33,17 +33,44 @@
...
+     * non-negative nanoseconds until the timeout
+     */
     private final long durationNs;
+    /**
+     * If true then reset the timeout at every {@link TimeoutSubscriber#onNext(Object)} otherwise never reset after
+     * subscribe
+     */
+    private final boolean resetAtOnNext;
 
     TimeoutPublisher(final Publisher<T> original,
```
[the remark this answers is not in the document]
bondolo: I am fine with removing them. I had left them alone only out of an abundance of caution.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/SignalOffloaders.java
bondolo: Yes, left behind from other work but (marginally) useful comment so retained.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/TimeoutPublisherTest.java
bondolo: This task has proved to be more difficult than I expected. I was able to verify that it works as expected with a simple program using realistic timeouts but I cannot get it to behave properly using the TestExecutor et al fixtures.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/TimeoutPublisherTest.java
```diff
@@ -166,28 +210,55 @@ public void noDataAndTimeout() {
...
-        assertThat(subscriber.takeOnNext(3), contains(1, 2, 3));
+        Supplier<Throwable> timedOut = null;
+        for (int x = 1; x <= 3; x++) {
+            assertThat(subscriber.pollOnNext(0, MILLISECONDS), is(nullValue()));
+            timedOut = subscriber.pollTerminal(0, MILLISECONDS);
+            if (null != timedOut) {
+                break;
+            }
+            publisher.onNext(x); // may reset timer
+            SECONDS.sleep(1);
```
Scottmitch, line 229: can you reduce the time for this test. takes ~7 seconds for both permutations to run on my local host.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
```diff
@@ -228,7 +233,8 @@ public void run() {
 
             // Instead of recursion we use a 2 level for loop structure.
             for (;;) {
-                final long nextTimeoutNs = parent.durationNs - (nanoTime() - lastOnNextNs);
+                final long now = System.nanoTime();
```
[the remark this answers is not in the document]
bondolo: Sure (I don't believe there would be any difference in performance).

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/TimeoutPublisherTest.java
bondolo: I was worried that it would be too long. I will try 100ms "steps" rather than 1s.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
```diff
@@ -29,53 +28,55 @@
...
 import static io.servicetalk.concurrent.internal.EmptySubscriptions.EMPTY_SUBSCRIPTION;
 import static java.lang.Math.max;
-import static java.lang.System.nanoTime;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 
+/**
+ * Publisher which cancels the subscription after the specified duration if not completed. Two timer modes are offered:
+ * <ul>
+ *     <li>An absolute timeout which requires that the publisher receive all items and complete (or error)
```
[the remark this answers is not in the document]
bondolo: Shouldn't be needed for HTML 5 but I will add them.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
```diff
@@ -29,53 +28,55 @@
...
+ *     <li>A mode which restarts the timer for each items received by {@link Subscriber#onNext(Object) onNext} providing
```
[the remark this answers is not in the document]
bondolo: done.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
```diff
@@ -86,7 +87,7 @@ void handleSubscribe(Subscriber<? super T> subscriber, SignalOffloader signalOff
                 signalOffloader, contextMap, contextProvider);
     }
 
-    private static final class TimeoutSubscriber<X> implements Subscriber<X>, Subscription, Runnable {
```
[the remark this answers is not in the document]
bondolo: A `Runnable` declaration doesn't explain the context under which the `run()` method is to be called. Here `Runnable` is essentially an implementation detail used for an API that the class calls rather than part of the API it offers to callers. That `run()` is a public method also might make it appear that `run()` is part of the API of `TimeoutSubscriber`. Using a direct method reference (and giving the method a useful name) provides better stack traces while providing the same performance and allows removal of an extraneous interface.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
```diff
@@ -29,53 +28,55 @@
...
                      final long duration,
                      final TimeUnit unit,
+                     final boolean restartAtOnNext,
                      final io.servicetalk.concurrent.Executor timeoutExecutor) {
-        this(original, publisherExecutor, unit.toNanos(duration), timeoutExecutor);
+        this(original, publisherExecutor, unit.toNanos(duration), restartAtOnNext, timeoutExecutor);
     }
 
-    private TimeoutPublisher(final Publisher<T> original,
+    TimeoutPublisher(final Publisher<T> original,
```
idelpivnitskiy, line 68: Only one ctor is used by `Publisher`, consider removing the 2nd one.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
```diff
@@ -29,53 +28,55 @@
...
                              final Executor publisherExecutor,
                              final long durationNs,
+                             final boolean restartAtOnNext,
                              final io.servicetalk.concurrent.Executor timeoutExecutor) {
         super(publisherExecutor);
         this.original = requireNonNull(original);
         this.timeoutExecutor = requireNonNull(timeoutExecutor);
         // We use the duration in arithmetic below to determine the expiration time for the "next timer" below. So
         // lets cap this at 0 to simplify overflow at that time.
         this.durationNs = max(0, durationNs);
```
idelpivnitskiy, line 78: It's pre-existing, but wdyt if instead of forcing it to 0 we validate and thow `IllegalArgumentException` when duration is less than 1ns? Or we can do that in a follow-up to extract the behavior change out of this PR.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/TimeoutPublisherTest.java
```diff
@@ -166,28 +210,57 @@ public void noDataAndTimeout() {
...
+        Supplier<Throwable> timedOut = null;
+        for (int x = 1; x <= 3; x++) {
+            assertThat(subscriber.pollOnNext(0, MILLISECONDS), is(nullValue()));
+            timedOut = subscriber.pollTerminal(0, MILLISECONDS);
+            if (null != timedOut) {
+                break;
+            }
+            publisher.onNext(x); // may reset timer
+            MILLISECONDS.sleep(1 * millisMultiplier);
+            testExecutor.advanceTimeBy(1 * millisMultiplier, MILLISECONDS);
```
idelpivnitskiy, line 231: `1 *` is not necessary here

idelpivnitskiy (approved):
We should also add `PublisherTimeoutTerminalTckTest`, similar to `PublisherTimeoutTckTest`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
bondolo: I think the forcing to zero is intended to allow sloppiness around calculated durations. Like the various java.util.concurrent timed operations <= 0 values result in no waiting. It also normalizes the timeout behaviour such that a non-positive value always just times out immediately.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -1632,7 +1641,20 @@ public final Completable flatMapCompletableDelayError(Function<? super T, ? exte
...
+    }
+
+    /**
+     * Converts a {@code Duration} to nanoseconds or if the resulting value would overflow a 64-bit signed integer then
+     * either {@code Long.MIN_VALUE} or {@code Long.MAX_VALUE} as appropriate.
+     *
+     * @param duration The duration to convert
+     * @return The converted nanoseconds value.
+     */
+    private long toNanos(Duration duration) {
```
Scottmitch, line 1654: nit: time utility state/methods can be static and likely moved into another class (publisher has many methods so we try to keep it focused on operators).

bondolo: OK. I made this change because I notice the Java 11 has it built-in and looking at the implementation realized that overflow handling was needed.

PR #1446, opened 2021-03-17
Publisher, Single, Completable rename idleTimeout -> timeout

Scottmitch:
Motivation:
Publisher, Single, Completable have an idleTimeout operator. The "idle"
was meant to highlight that the timeout applies to any signals (e.g. not
time to terminal event). However other rx implementations (reactor,
rxjava) just use "timeout" for this operator name which is consistent
with the [1] reactivex operator naming convention. Using the "idle"
makes the operator more difficult to discover for folks familiar with
existing concepts.

[1] http://reactivex.io/documentation/operators/timeout.html

Modifications:
- Rename idleTimeout to timeout, deprecate idleTimeout

Result:
More consistent operator naming for timeout operator.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -1177,8 +1177,29 @@ public abstract class Publisher<T> {
      * @return a new {@link Publisher} that will mimic the signals of this {@link Publisher} but will terminate with a
      * {@link TimeoutException} if time {@code duration} elapses between {@link Subscriber#onNext(Object)} calls.
      * @see <a href="http://reactivex.io/documentation/operators/timeout.html">ReactiveX timeout operator.</a>
-     * @see #idleTimeout(long, TimeUnit, io.servicetalk.concurrent.Executor)
+     * @see #timeout(long, TimeUnit, io.servicetalk.concurrent.Executor)
      */
+    public final Publisher<T> timeout(long duration, TimeUnit unit) {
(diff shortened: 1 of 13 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -1229,16 +1235,93 @@ public final Completable flatMapCompletableDelayError(Function<? super T, ? exte
...
+     * {@link TimeoutException} if time {@code duration} elapses between {@link Subscriber#onNext(Object)} calls.
+     * @see <a href="http://reactivex.io/documentation/operators/timeout.html">ReactiveX timeout operator.</a>
+     */
+    public final Publisher<T> timeout(long duration, TimeUnit unit,
+                                      io.servicetalk.concurrent.Executor timeoutExecutor) {
+        return new TimeoutPublisher<>(this, executor, duration, unit, timeoutExecutor);
+    }
+
+    /**
+     * Creates a new {@link Publisher} that will mimic the signals of this {@link Publisher} but will terminate with a
```
bondolo, line 1308: This overload would conflict with the usage planned in #1445

Scottmitch: can you clarify how this would conflict? bcz `timeout` is a pre-established concept (what we currently have as `idleTimeout`) I think it makes sense for #1445 to use an alternative/unique name https://github.com/apple/servicetalk/pull/1445#discussion_r596225376

bondolo (approved):

PR #1449, opened 2021-03-19
gRPC deserializer, multiple nio buffers with multiple messages

NiteshKant:
__Motivation__

`ProtoBufSerializationProvider.ProtoDeserializer` when extracting NIO buffers incorrectly reads all bytes from the buffer instead of the length of one message.

__Modification__

Read only the bytes representing a single message (`decodedLengthOfData`) while extracting the NIO buffers.

__Result__

gRPC deserializer works with buffers having multiple NIO buffers and multiple proto messages.


idelpivnitskiy (approved):
Thank you, @NiteshKant! :)

servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProvider.java
```diff
@@ -172,7 +172,7 @@ private static boolean isCompressed(Buffer buffer) throws SerializationException
                             // into it. Later, proto parser will copy data from this temporary ByteBuffer again.
                             // To avoid unnecessary copying, we use newCodedInputStream(buffers, lengthOfData).
                             final ByteBuffer[] buffers = buffer.toNioBuffers(buffer.readerIndex(),
-                                    buffer.readableBytes());
+                                    decodedLengthOfData);
```
Scottmitch, line 175: nit: put on previous line

Scottmitch (approved):

PR #1450, opened 2021-03-20
Add Publisher.scanWithLifetime(ScanWithLifetimeMapper) operator

tkountis:
**Motivation**:
The existing Publisher.scanWith(...) operator is not exposing cancellations to the end user. Use cases that need to control resource lifetimes (ie. clean-up), can't do so on cancel.

**Modifications**:
Add a new Publisher operator, scanWithLifetime(ScanWithLifetimeMapper), that also captures cancellation and invokes an afterFinally callback on all terminal signal (complete, error & cancel).

**Result**:
A new lifetime aware scanWith operator.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimeMapper.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimeMapper.java
@@ -21,7 +21,8 @@ import java.util.function.Supplier;
 
 /**
  * Provides the ability to transform (aka map) signals emitted via
- * the {@link Publisher#scanWithLifetime(Supplier)} operator, and hooks cancellations signals.
+ * the {@link Publisher#scanWithLifetime(Supplier)} operator, as well as the ability to cleanup state
+ * via {@link #afterFinally}.
  * @param <T> Type of items emitted by the {@link Publisher} this operator is applied.
  * @param <R> Type of items emitted by this operator.
  */
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimeMapper.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimeMapper.java
@@ -30,6 +30,8 @@ public interface ScanWithLifetimeMapper<T, R> extends ScanWithMapper<T, R> {
     /**
      * Invoked after a terminal signal {@link PublisherSource.Subscriber#onError(Throwable)} or
      * {@link PublisherSource.Subscriber#onComplete()} or {@link PublisherSource.Subscription#cancel()}.
+     * No further interaction will occur with the {@link ScanWithLifetimeMapper} to prevent use-after-free
+     * on internal state.
      */
-    void onFinalize();
+    void afterFinally();
 }
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimeMapper.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimeMapper.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.concurrent.api;
+
+import io.servicetalk.concurrent.PublisherSource;
+
+import java.util.function.Supplier;
+
+/**
+ * Provides the ability to transform (aka map) signals emitted via
+ * the {@link Publisher#scanWithLifetime(Supplier)} operator, and hooks cancellations signals.
+ * @param <T> Type of items emitted by the {@link Publisher} this operator is applied.
+ * @param <R> Type of items emitted by this operator.
+ */
+public interface ScanWithLifetimeMapper<T, R> extends ScanWithMapper<T, R> {
+
+    /**
+     * Invoked after a terminal signal {@link PublisherSource.Subscriber#onError(Throwable)} or
+     * {@link PublisherSource.Subscriber#onComplete()} or {@link PublisherSource.Subscription#cancel()}.
+     */
+    void onFinalize();
+}
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimePublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimePublisher.java
@@ -58,12 +58,16 @@ final class ScanWithLifetimePublisher<T, R> extends AbstractNoHandleSubscribePub
         private static final int STATE_FINALIZE_PENDING = 2;
         private static final int STATE_FINALIZED = 3;
         /**
-         * Special state to handle the case where a reentry to a terminal (eg. onComplete) signal finalizes
-         * before the root onNext unlocks.
-         * onNext sees the state change while it was busy and assumes that a cancel occurred.
+         * Special state to handle the case where a reentry to a terminal (eg. onComplete) signal needs to defer
+         * finalization to the root frame onNext.
          * This state helps differentiate between cancel and reentrant completion.
          */
-        private static final int STATE_FINALIZED_REENTRY = 4;
+        private static final int STATE_FINALIZE_AFTER_REENTRY = 4;
+        /**
+         * Special state to handle the case where a reentry to a terminal (eg. onComplete) signal needs to defer
+         * a pending finalization to the root frame onNext.
+         */
+        private static final int STATE_FINALIZE_PENDING_AFTER_REENTRY = 5;
 
         @SuppressWarnings("rawtypes")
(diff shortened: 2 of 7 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimePublisher.java
```diff
@@ -0,0 +1,365 @@
...
+import static java.util.concurrent.atomic.AtomicLongFieldUpdater.newUpdater;
+
+final class ScanWithLifetimePublisher<T, R> extends AbstractNoHandleSubscribePublisher<R> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ScanWithLifetimePublisher.class);
+
+    private final Publisher<T> original;
+    private final Supplier<? extends ScanWithLifetimeMapper<? super T, ? extends R>> mapperSupplier;
+
+    ScanWithLifetimePublisher(Publisher<T> original,
```
tkountis, line 42: @Scottmitch i believe we should keep only this operator flavor around, and replace the `ScanWithPublisher` with this and a no-op `onFinalize`. The down side is that this op has two volatile fields, but i don't expect significant perf impacts. I can benchmark this if needed to quantify it.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimePublisher.java
```diff
@@ -0,0 +1,365 @@
...
+
+    @Override
+    void handleSubscribe(final Subscriber<? super R> subscriber, final SignalOffloader signalOffloader,
+                         final AsyncContextMap contextMap, final AsyncContextProvider contextProvider) {
+        original.delegateSubscribe(new ScanWithSubscriber<>(subscriber, mapperSupplier.get(), signalOffloader,
+                contextMap, contextProvider), signalOffloader, contextMap, contextProvider);
+    }
+
+    private static final class ScanWithSubscriber<T, R> implements Subscriber<T> {
+        private static final int NOT_ALLOWED = -1;
```
bondolo, line 58: Does using an int versus an enum have an advantage other than potentially the size of the field? ie. why not use AtomicReferenceFieldUpdater

tkountis: No benefit having an int in this case, so I think I will go with your suggestion. 🚀

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimeMapper.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Provides the ability to transform (aka map) signals emitted via
+ * the {@link Publisher#scanWithLifetime(Supplier)} operator, and hooks cancellations signals.
+ * @param <T> Type of items emitted by the {@link Publisher} this operator is applied.
+ * @param <R> Type of items emitted by this operator.
+ */
+public interface ScanWithLifetimeMapper<T, R> extends ScanWithMapper<T, R> {
+
+    /**
+     * Invoked after a terminal signal {@link PublisherSource.Subscriber#onError(Throwable)} or
```
Scottmitch, line 31: also we should clarify no further interaction with the `ScanWithMapper` will occur after this method, which is intended to prevent use-after-free on internal state.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimePublisher.java
Scottmitch: fwiw `int` is the smallest unit of memory for atomics, enums are classes/references which may consume the same size as `int` but maybe not.

bondolo: Understood, references could be 64-bit in VMs with very large heaps where compressed OOPs cannot be used. It does feel a bit like premature optimization to use an `int` rather than enum but I can understand not wanting to revisit.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimeMapper.java
```diff
@@ -0,0 +1,35 @@
...
+     * {@link PublisherSource.Subscriber#onComplete()} or {@link PublisherSource.Subscription#cancel()}.
+     */
+    void onFinalize();
```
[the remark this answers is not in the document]
tkountis: Finally is a preserved keyword. I went with `afterFinally`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimePublisher.java
```diff
@@ -0,0 +1,254 @@
...
+
+            try {
+                super.onNext(t);
+            } finally {
+                // Re-entry -> don't unlock
+                if (!reentry) {
+                    if (state == STATE_FINALIZE_PENDING_AFTER_REENTRY) {
+                        // If a re-entrant terminal didn't complete due to missing demand, we take over and set
+                        // state to STATE_FINALIZE_PENDING.
+                        if (!stateUpdater.compareAndSet(this, STATE_FINALIZE_PENDING_AFTER_REENTRY,
```
Scottmitch, line 122: if `onCancel` happens concurrently we may double-invoke `finalize()`. we should be more explicit about state transitions.

I extended your work here and to demo explicit state transitions which avoid concurrent finalization https://github.com/scottmitch/servicetalk-1/tree/scan_with_lifetime_demo.

servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/PublisherScanWithLifetimeTckTest.java
```diff
@@ -0,0 +1,57 @@
...
+            @Nullable
+            @Override
+            public String mapOnError(final Throwable cause) throws Throwable {
+                return null;
+            }
+
+            @Nullable
+            @Override
+            public String mapOnComplete() {
+                return valueOf(-1);
```
Scottmitch, line 48: nit: any reason why this can't be `null` like `mapOnError`?

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/ScanWithPublisherTest.java
```diff
@@ -41,6 +44,7 @@
 import static org.hamcrest.Matchers.nullValue;
 
 public class ScanWithPublisherTest {
+
```
Scottmitch, line 47: can you some tests where the subscription and subscriber are on different threads, and verify the finalization is only invoked once?

Scottmitch (changes requested):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithLifetimePublisher.java
tkountis: FWIW there is no double `finalize()` involved here, double checked the code. Let me know if you spend more time on it and figure out a path that leads to that behavior, would be nice to know if I missed sth.
Regardless, after our offline chat, we agreed to bring in your changes which demonstrate better clarity. They do have extra CAS in some cases but that can be an optimization we can look for later.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/ScanWithPublisherTest.java
tkountis: Added

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithPublisher.java
```diff
@@ -140,13 +147,25 @@ public void onNext(@Nullable final T t) {
...
+            onError0(t);
+        }
+
+        @Override
+        public void onComplete() {
+            onComplete0();
+        }
+
+        /**
+         * Return true if onError0 had sufficient demand to deliver.
```
Scottmitch, line 159: nit: consider using regular java docs here (e.g. add `@param t`, `@return`, `true` -> `{@code true}`)

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/ScanWithPublisherTest.java
```diff
@@ -201,68 +307,220 @@ public boolean mapTerminal() {
...
+            }
+
+            @Override
+            public boolean mapTerminal() {
+                if (interleaveCancellation) {
+                    checkpoint.countDown();
+                    try {
+                        resume.await();
+                    } catch (InterruptedException e) {
+                        e.printStackTrace();
```
Scottmitch, line 395: `e.printStackTrace();` -> `throw new AssertionError(e)`?

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/ScanWithPublisherTest.java
```diff
@@ -201,68 +307,220 @@ public boolean mapTerminal() {
...
+            Subscription s = subscriber.awaitSubscription();
+
+            s.request(2);
+            requested.countDown();
+
+            if (interleaveCancellation) {
+                try {
+                    checkpoint.await();
+                } catch (InterruptedException e) {
+                    e.printStackTrace();
```
Scottmitch, line 419: a common practice for tests with other threads where tasks may throw is as follows:
```java
ExecutorService e = ...;
try {
  Future<Void> f = e.submit(() -> {
    try {
      methodThatMayThrow();
     } catch (Exception e) {
        throw new AssertionError(e);
     }
    return null;
   });

   f.get();
} finally {
  e.shutdown();
}
```

this provides visibility to the test thread (and therefore the junit report) if an exception occurs in another thread, and also allows you to wait for the task to complete if required.

Scottmitch (approved):

PR #1453, opened 2021-03-23
Ensure that onError with timeout is called after subscription cancel

bondolo:
Motivation:
In some cases the `cancel()` which resulted from a timeout did not
result in the `onError()` also being called with the timeout exception
because the cancel operation changed the subscriber state.

Modifications:
If the operator had not completed at the time of the timeout then the
`onError()` should be called regardless of the state changes which
happen during `cancel()`.

Result:
The reason for the cancel is exposed via the `onError()`.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
@@ -232,7 +232,7 @@ final class TimeoutPublisher<T> extends AbstractNoHandleSubscribePublisher<T> {
                     offloadTimeout(new TimeoutException("timeout after " + NANOSECONDS.toMillis(parent.durationNs) +
                             "ms"));
                     return;
-                } else {
+                } else { // reschedule
                     final Cancellable nextTimerCancellable;
                     try {
                         nextTimerCancellable = requireNonNull(
@@ -282,13 +282,16 @@ final class TimeoutPublisher<T> extends AbstractNoHandleSubscribePublisher<T> {
         private void processTimeout(Throwable cause) {
             final Subscription subscription = subscriptionUpdater.getAndSet(this, EMPTY_SUBSCRIPTION);
             // The timer is started before onSubscribe so the subscription may actually be null at this time.
-            if (subscription != null) {
-                subscription.cancel();
-                // onErrorFromTimeout will protect against concurrent access on the Subscriber.
-            } else {
-                target.onSubscribe(EMPTY_SUBSCRIPTION);
+            try {
+                if (subscription != null) {
+                    subscription.cancel();
+                    // onErrorFromTimeout will protect against concurrent access on the Subscriber.
+                } else {
+                    target.onSubscribe(EMPTY_SUBSCRIPTION);
+                }
+            } finally {
+                target.processOnError(cause);
             }
-            target.processOnError(cause);
         }
 
         private void stopTimer() {
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutSingle.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutSingle.java
@@ -184,13 +184,16 @@ final class TimeoutSingle<T> extends AbstractNoHandleSubscribeSingle<T> {
                                 contextProvider.contextMap()));
                 // The timer is started before onSubscribe so the oldCancellable may actually be null at this time.
                 if (oldCancellable != null) {
-                    oldCancellable.cancel();
-
                     // We already know that this.onSubscribe was called because we have a valid Cancellable. We need to
                     // know that the call to target.onSubscribe completed so we don't interact with the Subscriber
                     // concurrently.
-                    if (subscriberStateUpdater.getAndSet(this, STATE_TIMED_OUT_ERROR) == STATE_ON_SUBSCRIBE_DONE) {
-                        offloadedTarget.onError(newTimeoutException());
+                    int stateWas = subscriberStateUpdater.getAndSet(this, STATE_TIMED_OUT_ERROR);
+                    try {
+                        oldCancellable.cancel();
+                    } finally {
+                        if (stateWas == STATE_ON_SUBSCRIBE_DONE) {
+                            offloadedTarget.onError(newTimeoutException());
+                        }
                     }
                 } else {
                     // If there is no Cancellable, that means this.onSubscribe wasn't called before the timeout. In this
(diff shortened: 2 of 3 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutSingle.java
```diff
@@ -184,13 +184,16 @@ public void run() {
...
                     // We already know that this.onSubscribe was called because we have a valid Cancellable. We need to
                     // know that the call to target.onSubscribe completed so we don't interact with the Subscriber
                     // concurrently.
-                    if (subscriberStateUpdater.getAndSet(this, STATE_TIMED_OUT_ERROR) == STATE_ON_SUBSCRIBE_DONE) {
-                        offloadedTarget.onError(newTimeoutException());
+                    int stateWas = subscriberStateUpdater.getAndSet(this, STATE_TIMED_OUT_ERROR);
+                    try {
+                        oldCancellable.cancel();
+                    } finally {
+                        if (stateWas == STATE_ON_SUBSCRIBE_DONE) {
```
idelpivnitskiy, line 194: The `subscriberState` changes only in 2 places: here and inside `onSubscribe`. How is it possible that `cancel()` modifies the state?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
```diff
@@ -282,13 +282,16 @@ private void offloadTimeout(Throwable cause) {
...
-                target.onSubscribe(EMPTY_SUBSCRIPTION);
+            try {
+                if (subscription != null) {
+                    subscription.cancel();
+                    // onErrorFromTimeout will protect against concurrent access on the Subscriber.
+                } else {
+                    target.onSubscribe(EMPTY_SUBSCRIPTION);
+                }
+            } finally {
+                target.processOnError(cause);
```
idelpivnitskiy, line 293: Why `finally` is required here? For cases when `onSubscribe/cancel` throws?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutSingle.java
bondolo: That part was possibly not essential. The main motivation was for the `cancel()` throws case and specifically when it is throwing the channel closed exception and we want the timeout exception "true cause" as the onError result.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
bondolo: Yes, to ensure that the original "true cause" is not lost.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutSingle.java
idelpivnitskiy: Interesting, the RS spec says that `cancel()` must [return normally](https://github.com/reactive-streams/reactive-streams-jvm#3.15). If you have seen that `cancel()` may throw, we should investigate the cause and fix it. Can you please provide more context when it happens? Do you have a reproducer?

bondolo: I think I have a better explanation now. It is not that cancel() might throw an exception. You are correct, it should never throw. The problem this addresses is that the call to cancel() might cause an onError() propagation and this can only be done once. By updating the subscriber state before calling cancel() the updater reserves the right for themselves to call the onError() with the cause they intended rather than someone else in the cancel() chain calling it with their cause.

In the particular case of timeout the exposed cause via onError() was showing up as StacklessClosedChannelException because the channel threw that when canceled. By updating the subscription state before calling cancel() we are suppressing any onError() causes being propagated up from within the cancel(). When the cancel() completes we call onError() with the cause we want to expose, TimeoutException.

bondolo: In a perfect world any onError() which occurs while cancel() is being called would be captured as a suppressed exception for the TimeoutException.

Scottmitch (approved):
we will need similar changes for `Publisher` to make sure the timeout exception is propagated but we can do that in a followup PR as the state machine is more involved.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutSingle.java
idelpivnitskiy: I don't want to block this PR, feel free to merge if it's already approved. I will appreciate if someone can help me to understand what I'm missing, but we can do that after the PR is merged or offline.

Thanks for the detailed explanation. This is approximately what I thought originally after reading the commit description. Hence my question: 

> How is it possible that `cancel()` modifies the state?

`target.onError(t)` is protected by two states: `subscriberStateUpdater` and `cancellableUpdater`. If we are inside the current `if` statement of the `run()` method, we flip the cancellable to `LOCAL_IGNORE_CANCEL` at the beginning. As the result, if the `oldCancellable.cancel()` generates `onError`, it will be ignored because the `onError` also checks the current `cancellable` and propagates it only if it's not `LOCAL_IGNORE_CANCEL`:
```java
        @Override
            if (cancellableUpdater.getAndSet(this, LOCAL_IGNORE_CANCEL) != LOCAL_IGNORE_CANCEL) {
                try {
                    stopTimer();
                } finally {
                    target.onError(t);
                }
            }
        }
```

From what you describe it looks like `StacklessClosedChannelException` is correctly propagated because it's generated **before** the timeout expires (before `run()` executes).

Inside the `timeout-http` branch I reverted this commit and repeatedly run the `ResponseTimeoutTest`. Could not reproduce an issue.

Is it possible to write a repeatable reproducer for the problem this PR tries to fix?

Scottmitch (commented):
@idelpivnitskiy brings up a good point, lets hold off on merging this until we have a reproducer which demonstrates the case this fixes.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutSingle.java
bondolo: I am unable to reproduce the original problem as well.

bondolo:
This change does not appear to be necessary. Closing without integration.

PR #1456, opened 2021-03-25
Improve timeout support for HTTP

bondolo:
Motivation:
Some uses of the HTTP API require progress or completion within specified time limits.

Modifications:
New or enhanced filters are provided for client and server to manage timeout events.
Existing client filter `TimeoutHttpRequesterFilter` is enhanced to support additional
timeout options for the entire HTTP request/response transaction. A new server filter
`TimeoutHttpServiceFilter` is provided for managing timeouts by services.

Result:
HTTP API provides request/response transaction timeout capabilities

### DIFF
--- a/servicetalk-examples/http/helloworld/src/main/java/io/servicetalk/examples/http/helloworld/async/HelloWorldClient.java
+++ b/servicetalk-examples/http/helloworld/src/main/java/io/servicetalk/examples/http/helloworld/async/HelloWorldClient.java
@@ -18,38 +18,31 @@ package io.servicetalk.examples.http.helloworld.async;
 import io.servicetalk.http.api.HttpClient;
 import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
 import io.servicetalk.http.netty.HttpClients;
-import io.servicetalk.http.utils.TimeoutHttpRequesterFilter;
 import io.servicetalk.transport.api.HostAndPort;
 
 import java.net.InetSocketAddress;
-import java.time.Duration;
 import java.util.concurrent.CountDownLatch;
 
 import static io.servicetalk.http.api.HttpSerializationProviders.textDeserializer;
-import static io.servicetalk.http.utils.TimeoutHttpRequesterFilter.useDefaultTimeout;
 
 public final class HelloWorldClient {
 
     public static void main(String[] args) throws Exception {
         SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> builder =
-                HttpClients.forSingleAddress("localhost", 8080)
-                // (optional) All requests must fully complete within 2 minutes
-                .appendClientFilter(new TimeoutHttpRequesterFilter(
-                        useDefaultTimeout(Duration.ofMinutes(2)), true));
+                HttpClients.forSingleAddress("localhost", 8080);
         try (HttpClient client = builder.build()) {
             // This example is demonstrating asynchronous execution, but needs to prevent the main thread from exiting
             // before the response has been processed. This isn't typical usage for a streaming API but is useful for
             // demonstration purposes.
             CountDownLatch responseProcessedLatch = new CountDownLatch(1);
             client.request(client.get("/sayHello"))
-                    // (optional) Requests must complete within 1 minute
-                    .timeout(Duration.ofMinutes(1))
                     .afterFinally(responseProcessedLatch::countDown)
                     .subscribe(resp -> {
                         System.out.println(resp.toString((name, value) -> value));
                         System.out.println(resp.payloadBody(textDeserializer()));
                     });
 
+            // block until request is complete and afterFinally() is called
             responseProcessedLatch.await();
         }
     }
--- a/servicetalk-examples/http/helloworld/src/main/java/io/servicetalk/examples/http/helloworld/async/HelloWorldClient.java
+++ b/servicetalk-examples/http/helloworld/src/main/java/io/servicetalk/examples/http/helloworld/async/HelloWorldClient.java
@@ -16,21 +16,34 @@
 package io.servicetalk.examples.http.helloworld.async;
 
 import io.servicetalk.http.api.HttpClient;
+import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
 import io.servicetalk.http.netty.HttpClients;
+import io.servicetalk.http.utils.TimeoutHttpRequesterFilter;
+import io.servicetalk.transport.api.HostAndPort;
 
+import java.net.InetSocketAddress;
+import java.time.Duration;
 import java.util.concurrent.CountDownLatch;
 
 import static io.servicetalk.http.api.HttpSerializationProviders.textDeserializer;
+import static io.servicetalk.http.utils.TimeoutHttpRequesterFilter.useDefaultTimeout;
 
 public final class HelloWorldClient {
 
     public static void main(String[] args) throws Exception {
-        try (HttpClient client = HttpClients.forSingleAddress("localhost", 8080).build()) {
+        SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> builder =
+                HttpClients.forSingleAddress("localhost", 8080)
+                // (optional) All requests must fully complete within 2 minutes
+                .appendClientFilter(new TimeoutHttpRequesterFilter(
+                        useDefaultTimeout(Duration.ofMinutes(2)), true));
+        try (HttpClient client = builder.build()) {
             // This example is demonstrating asynchronous execution, but needs to prevent the main thread from exiting
             // before the response has been processed. This isn't typical usage for a streaming API but is useful for
             // demonstration purposes.
             CountDownLatch responseProcessedLatch = new CountDownLatch(1);
             client.request(client.get("/sayHello"))
+                    // (optional) Requests must complete within 1 minute
+                    .timeout(Duration.ofMinutes(1))
                     .afterFinally(responseProcessedLatch::countDown)
                     .subscribe(resp -> {
                         System.out.println(resp.toString((name, value) -> value));
--- a/servicetalk-examples/http/helloworld/src/main/java/io/servicetalk/examples/http/helloworld/async/HelloWorldClient.java
+++ b/servicetalk-examples/http/helloworld/src/main/java/io/servicetalk/examples/http/helloworld/async/HelloWorldClient.java
@@ -16,11 +16,8 @@
 package io.servicetalk.examples.http.helloworld.async;
 
 import io.servicetalk.http.api.HttpClient;
-import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
 import io.servicetalk.http.netty.HttpClients;
-import io.servicetalk.transport.api.HostAndPort;
 
-import java.net.InetSocketAddress;
 import java.util.concurrent.CountDownLatch;
 
 import static io.servicetalk.http.api.HttpSerializationProviders.textDeserializer;
@@ -28,9 +25,7 @@ import static io.servicetalk.http.api.HttpSerializationProviders.textDeserialize
 public final class HelloWorldClient {
 
     public static void main(String[] args) throws Exception {
-        SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> builder =
-                HttpClients.forSingleAddress("localhost", 8080);
-        try (HttpClient client = builder.build()) {
+        try (HttpClient client = HttpClients.forSingleAddress("localhost", 8080).build()) {
             // This example is demonstrating asynchronous execution, but needs to prevent the main thread from exiting
             // before the response has been processed. This isn't typical usage for a streaming API but is useful for
             // demonstration purposes.
--- a/servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutClient.java
+++ b/servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutClient.java
@@ -54,7 +54,7 @@ public final class TimeoutClient {
             // first request, with default timeout from HttpClient
             client.request(client.get("/sayHello"))
                     .afterFinally(responseProcessedLatch::countDown)
-                    .afterOnError(System.out::println)
+                    .afterOnError(System.err::println)
                     .subscribe(resp -> {
                         System.out.println(resp.toString((name, value) -> value));
                         System.out.println(resp.payloadBody(textDeserializer()));
@@ -65,7 +65,7 @@ public final class TimeoutClient {
                     // This request and response must complete within 3 seconds or the request will be cancelled.
                     .timeout(Duration.ofSeconds(3))
                     .afterFinally(responseProcessedLatch::countDown)
-                    .afterOnError(System.out::println)
+                    .afterOnError(System.err::println)
                     .subscribe(resp -> {
                         System.out.println(resp.toString((name, value) -> value));
                         System.out.println(resp.payloadBody(textDeserializer()));
--- a/servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutClient.java
+++ b/servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutClient.java
@@ -42,8 +42,7 @@ public final class TimeoutClient {
                 HttpClients.forSingleAddress("localhost", 8080)
                         // Filter enforces that requests made with this client must fully complete
                         // within 10 seconds or will be cancelled.
-                        .appendClientFilter(new TimeoutHttpRequesterFilter(
-                                Duration.ofSeconds(10), true));
+                        .appendClientFilter(new TimeoutHttpRequesterFilter(Duration.ofSeconds(10), true));
 
         try (HttpClient client = builder.build()) {
             // This example is demonstrating asynchronous execution, but needs to prevent the main thread from exiting
@@ -51,7 +50,7 @@ public final class TimeoutClient {
             // demonstration purposes.
             CountDownLatch responseProcessedLatch = new CountDownLatch(2);
 
-            // first request, with default timeout from HttpClient
+            // first request, with default timeout from HttpClient (this will succeed)
             client.request(client.get("/sayHello"))
                     .afterFinally(responseProcessedLatch::countDown)
                     .afterOnError(System.err::println)
@@ -60,7 +59,7 @@ public final class TimeoutClient {
                         System.out.println(resp.payloadBody(textDeserializer()));
                     });
 
-            // second request, with custom timeout that is lower than the client default
+            // second request, with custom timeout that is lower than the client default (this will timeout)
             client.request(client.get("/sayHello"))
                     // This request and response must complete within 3 seconds or the request will be cancelled.
                     .timeout(Duration.ofSeconds(3))
@@ -71,7 +70,7 @@ public final class TimeoutClient {
                         System.out.println(resp.payloadBody(textDeserializer()));
                     });
 
-            // block until request is complete and afterFinally() is called
+            // block until requests are complete and afterFinally() has been called
             responseProcessedLatch.await();
         }
     }
--- a/servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutClient.java
+++ b/servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutClient.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.examples.http.timeout;
+
+import io.servicetalk.http.api.HttpClient;
+import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
+import io.servicetalk.http.netty.HttpClients;
+import io.servicetalk.http.utils.TimeoutHttpRequesterFilter;
+import io.servicetalk.transport.api.HostAndPort;
+
+import java.net.InetSocketAddress;
+import java.time.Duration;
+import java.util.concurrent.CountDownLatch;
+
+import static io.servicetalk.http.api.HttpSerializationProviders.textDeserializer;
+
+/**
+ * Extends the async 'Hello World!' example to demonstrate use of timeout filters and timeout operators. If a single
+ * timeout can be applied to all transactions then the timeout should be applied using the
+ * {@link TimeoutHttpRequesterFilter}. If only some transactions require a timeout then the timeout should be applied
+ * using a {@link io.servicetalk.concurrent.api.Single#timeout(Duration)} Single.timeout()} or a
+ * {@link io.servicetalk.concurrent.api.Publisher#timeoutTerminal(Duration)} (Duration)} Publisher.timeoutTerminal()}
+ * operator.
+ */
+public final class TimeoutClient {
+
+    public static void main(String[] args) throws Exception {
+        SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> builder =
+                HttpClients.forSingleAddress("localhost", 8080)
+                        // Filter enforces that requests made with this client must fully complete
+                        // within 2 minutes or will be cancelled.
+                        .appendClientFilter(new TimeoutHttpRequesterFilter(
+                                Duration.ofMinutes(2), true));
+
+        try (HttpClient client = builder.build()) {
+            // This example is demonstrating asynchronous execution, but needs to prevent the main thread from exiting
+            // before the response has been processed. This isn't typical usage for a streaming API but is useful for
+            // demonstration purposes.
+            CountDownLatch responseProcessedLatch = new CountDownLatch(1);
+            client.request(client.get("/sayHello"))
+                    // This request and response must complete within 1 minute or the request will be cancelled.
+                    .timeout(Duration.ofMinutes(1))
+                    .afterFinally(responseProcessedLatch::countDown)
+                    .subscribe(resp -> {
+                        System.out.println(resp.toString((name, value) -> value));
+                        System.out.println(resp.payloadBody(textDeserializer()));
+                    });
+
+            // block until request is complete and afterFinally() is called
+            responseProcessedLatch.await();
+        }
+    }
+}
--- a/servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutClient.java
+++ b/servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutClient.java
@@ -41,19 +41,31 @@ public final class TimeoutClient {
         SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> builder =
                 HttpClients.forSingleAddress("localhost", 8080)
                         // Filter enforces that requests made with this client must fully complete
-                        // within 2 minutes or will be cancelled.
+                        // within 10 seconds or will be cancelled.
                         .appendClientFilter(new TimeoutHttpRequesterFilter(
-                                Duration.ofMinutes(2), true));
+                                Duration.ofSeconds(10), true));
 
         try (HttpClient client = builder.build()) {
             // This example is demonstrating asynchronous execution, but needs to prevent the main thread from exiting
             // before the response has been processed. This isn't typical usage for a streaming API but is useful for
             // demonstration purposes.
-            CountDownLatch responseProcessedLatch = new CountDownLatch(1);
+            CountDownLatch responseProcessedLatch = new CountDownLatch(2);
+
+            // first request, with default timeout from HttpClient
             client.request(client.get("/sayHello"))
-                    // This request and response must complete within 1 minute or the request will be cancelled.
-                    .timeout(Duration.ofMinutes(1))
                     .afterFinally(responseProcessedLatch::countDown)
+                    .afterOnError(System.out::println)
+                    .subscribe(resp -> {
+                        System.out.println(resp.toString((name, value) -> value));
+                        System.out.println(resp.payloadBody(textDeserializer()));
+                    });
+
+            // second request, with custom timeout that is lower than the client default
+            client.request(client.get("/sayHello"))
+                    // This request and response must complete within 3 seconds or the request will be cancelled.
+                    .timeout(Duration.ofSeconds(3))
+                    .afterFinally(responseProcessedLatch::countDown)
+                    .afterOnError(System.out::println)
                     .subscribe(resp -> {
                         System.out.println(resp.toString((name, value) -> value));
                         System.out.println(resp.payloadBody(textDeserializer()));
--- a/servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutServer.java
+++ b/servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutServer.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.examples.http.timeout;
+
+import io.servicetalk.http.netty.HttpServers;
+import io.servicetalk.http.utils.TimeoutHttpServiceFilter;
+
+import java.time.Duration;
+
+import static io.servicetalk.concurrent.api.Single.succeeded;
+import static io.servicetalk.http.api.HttpSerializationProviders.textSerializer;
+
+/**
+ * Extends the async 'Hello World!' example to demonstrate use of timeout filter.
+ */
+public final class TimeoutServer {
+    public static void main(String[] args) throws Exception {
+        HttpServers.forPort(8080)
+                // Filter enforces that responses must complete within 2 minutes or will be cancelled.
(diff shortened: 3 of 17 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseTimeoutTest.java
```diff
@@ -0,0 +1,311 @@
...
+import static io.servicetalk.http.utils.TimeoutHttpRequesterFilter.useDefaultTimeout;
+import static io.servicetalk.transport.netty.internal.AddressUtils.localAddress;
+import static io.servicetalk.transport.netty.internal.AddressUtils.serverHostAndPort;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.fail;
+
+@RunWith(Parameterized.class)
+public class ResponseTimeoutTest {
```
tkountis, line 78: Leftover comments

tkountis:
Minor comments.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
```diff
@@ -0,0 +1,92 @@
...
+import io.servicetalk.http.api.StreamingHttpService;
+import io.servicetalk.http.api.StreamingHttpServiceFilter;
+import io.servicetalk.http.api.StreamingHttpServiceFilterFactory;
+
+import java.time.Duration;
+import java.time.Instant;
+import javax.annotation.Nullable;
+
+/**
+ * A {@link StreamingHttpServiceFilter} that adds support for request/response timeouts.
```
tkountis, line 36: Should we comment about retry/timeout filter ordering? If a retry filters comes before a timeout filter, then the timeout will be retried too, potentially ending up spending more time than the user initially wanted. Example:

With a timeout of 5 seconds, and an operation that throws a retryable exception after 3 seconds, will be retried with another timeout of 5 seconds. If it finishes after 3 seconds, the total time spend is >5seconds.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -30,46 +34,104 @@
...
  */
 public final class TimeoutHttpRequesterFilter implements StreamingHttpClientFilterFactory,
                                                          StreamingHttpConnectionFilterFactory,
                                                          HttpExecutionStrategyInfluencer {
-    private final Duration duration;
+    /**
+     * HTTP timeout is stored in context as a deadline so that when propagated to a new client request the remaining
+     * time available for that request can be calculated.
+     */
+    static final AsyncContextMap.Key<Instant> HTTP_DEADLINE_KEY = AsyncContextMap.Key.newKey("http-timeout-deadline");
```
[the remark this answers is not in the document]
bondolo: Yes. It had been used by the TimeoutHttpServiceFilter in the same package.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -30,46 +34,104 @@
...
+
+    /**
+     * Establishes the timeout for a given request
+     */
+    private final TimeoutFromRequest timeoutForRequest;
```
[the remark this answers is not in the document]
bondolo: In this case the advantage over Function<HttpRequest,Duration> was the ability to include @Nullable annotations.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -100,4 +162,58 @@ public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strat
...
+     * Returns timeout duration calculated based on the remaining time until the deadline in the context or, if absent,
+     * the provided default.
+     *
+     * <p>The timeout, if any, will be added to the context for additional requests initiated within this context
+     *
+     * @param defaultDuration default timeout duration or null for no timeout
+     * @return a timeout based on the context deadline or specified default (which may be null)
+     */
+    public static @Nullable Duration useContextDeadlineOrDefault(@Nullable Duration defaultDuration) {
+        Instant deadline = AsyncContext.get(HTTP_DEADLINE_KEY);
```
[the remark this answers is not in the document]
bondolo: I discovered that it didn't work when I disabled AsyncContext for debugging. I am not sure it needs to be documented because disabling AsyncContext isn't something that applications would do (I could be wrong in this assumption) and that when the ServiceTalks devs disable it they would understand that there might be behavioral differences.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseTimeoutTest.java
bondolo: Do you mean the commented out code for turning on wireLogging, disabling AsyncContext and disabling offloading?

tkountis: Yes sorry, i could have left the comment on the relevant lines.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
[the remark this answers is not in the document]
tkountis: afaik AsyncContext [can be disabled by users too ](https://apple.github.io/servicetalk/servicetalk/SNAPSHOT/async-context.html#usage) if its not needed (ie. tracing), since its rather costly. Requiring it for timeouts (especially when this info isn't propagated to other requests 'yet') seems unreasonable. 

Me: _thinking out loud_
It looks to me that we need it only if we want to apply `remaining` duration for timeout on other actions upstream? Is there any other benefit? Can this functionality be user defined rather a filter default?

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
bondolo: Assuming the AsyncContext is enabled, correct handling of the total timeout would be accomplished by the deadline added to the context.

tkountis: OK I see, so if someone uses the `useDefaultTimeout` function, but what about any other function eg. `(__) -> ofSeconds(1)`

bondolo: If they write their own function then they would be responsible for ensuring the ordering of filters accomplishes what they desire. One could even imagine two layers of timeout filters, all attempts and per attempt. (at least one of which would have to define a custom RequestToTimeout function).

tkountis: Agreed, shouldn't we clarify that in the javadoc and explain that ordering matters and why so?
eg. https://github.com/apple/servicetalk/blob/main/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ContentCodingHttpServiceFilter.java#L51

servicetalk-examples/http/helloworld/src/main/java/io/servicetalk/examples/http/helloworld/async/HelloWorldClient.java
```diff
@@ -16,21 +16,34 @@
...
 import static io.servicetalk.http.api.HttpSerializationProviders.textDeserializer;
+import static io.servicetalk.http.utils.TimeoutHttpRequesterFilter.useDefaultTimeout;
 
 public final class HelloWorldClient {
 
     public static void main(String[] args) throws Exception {
-        try (HttpClient client = HttpClients.forSingleAddress("localhost", 8080).build()) {
+        SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> builder =
+                HttpClients.forSingleAddress("localhost", 8080)
+                // (optional) All requests must fully complete within 2 minutes
```
Scottmitch, line 36: lets keep the hello world example simple, and add a new example if we want to demonstrate new features.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -100,4 +167,59 @@ public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strat
         // No influence since we do not block.
         return strategy;
     }
+
+    @FunctionalInterface
+    public interface TimeoutFromRequest extends Function<HttpRequestMetaData, Duration> {
```
Scottmitch, line 172: do we need this additional interface? IIUC I wonder if it is just for `@Nullable` annotations, and we have deferred adding new types for this in the past. I we think this provides enough value should we add a more general API (e.g. `NullReturnFunction`) so we don't have to create new types for each use case?

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -100,4 +167,59 @@ public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strat
...
+
+        if (null != deadline) {
+            return Duration.between(Instant.now(), deadline);
+        } else if (null != defaultDuration) {
+            // Convert it to a deadline so that if reused for a new client request the timeout will reflect actual
+            // remaining time.
+            try {
+                AsyncContext.put(HTTP_DEADLINE_KEY, Instant.now().plus(defaultDuration));
+            } catch (UnsupportedOperationException ignored) {
+                // ignored
```
Scottmitch, line 219: any reason we don't re-throw this rather than silently swallow? Also in the future when we have RequestContext we don't really want folks to use this method IIUC. perhaps we should just internalize this in gRPC where required for now.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -30,46 +34,109 @@
...
 
     private Single<StreamingHttpResponse> request(final StreamingHttpRequester delegate,
                                                   final HttpExecutionStrategy strategy,
                                                   final StreamingHttpRequest request) {
-        return timeoutExecutor != null ? delegate.request(strategy, request).timeout(duration, timeoutExecutor) :
-                delegate.request(strategy, request).timeout(duration);
+
+        Single<StreamingHttpResponse> response = delegate.request(strategy, request);
+
+        Duration timeout = timeoutForRequest.apply(request);
```
Scottmitch, line 120: the server has this function invoked inside the `defer` which means it will be re-invoked on each re-subscribe/retry, but here it is outside the scope of `defer`. can we make them consistent?

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
idelpivnitskiy: +1 for moving this logic to gRPC. At HTTP level, an ability to provide a function is enough. If users need deadlines for HTTP, they can implement their function to support it. There is no much complexity in this function and it probably won't fit all use-cases.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -100,4 +167,59 @@ public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strat
...
+    }
+
+    /**
+     * Returns a function which uses {@link #useContextDeadlineOrDefault(Duration) useContextDeadlineOrDefault()} with
+     * the provided default duration to determine the timeout duration to be used for a provided request.
+     *
+     * @param duration default timeout duration or null for no timeout
+     * @return a function to produce a timeout based on the context deadline or specified default
+     */
+    public static TimeoutFromRequest useDefaultTimeout(@Nullable Duration duration) {
```
idelpivnitskiy, line 192: If we remove `useContextDeadlineOrDefault` based on discussion in https://github.com/apple/servicetalk/pull/1456/files/19674baef2572fc09a1dfc25c25b22b34ccc9cab..f4365832f371db9c976fba790b24b1b86c5b8bd6#r602650058, consider removing this method as well.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+public final class TimeoutHttpServiceFilter
+        implements StreamingHttpServiceFilterFactory, HttpExecutionStrategyInfluencer {
+
+    private final TimeoutHttpRequesterFilter.TimeoutFromRequest timeoutForRequest;
+
+    @Nullable
+    private final Executor timeoutExecutor;
+
+    public TimeoutHttpServiceFilter(TimeoutHttpRequesterFilter.TimeoutFromRequest timeoutForRequest) {
```
idelpivnitskiy, line 50: Ctors should have javadoc

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -30,46 +34,109 @@
...
+
+    /**
+     * Creates a new instance.
+     *
+     * @param timeoutForRequest function for extracting timeout from request
+     * @param fullRequestResponse if true then timeout is for full request/response transaction otherwise only the
+     * response metadata must arrive before the timeout.
+     */
+    public TimeoutHttpRequesterFilter(final TimeoutFromRequest timeoutForRequest, final boolean fullRequestResponse) {
+        this.timeoutForRequest = timeoutForRequest;
```
idelpivnitskiy, line 93: Consider using `requireNonNull` for all ctors that take `timeoutForRequest` function.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -30,46 +34,109 @@
...
+     * @param fullRequestResponse if true then timeout is for full request/response transaction otherwise only the
+     * response metadata must arrive before the timeout.
+     * @param timeoutExecutor the {@link Executor} to use for managing the timer notifications
+     */
+    public TimeoutHttpRequesterFilter(final TimeoutFromRequest timeoutForRequest,
+                                      final boolean fullRequestResponse,
+                                      final Executor timeoutExecutor) {
+        this.timeoutForRequest = timeoutForRequest;
+        this.fullRequestResponse = fullRequestResponse;
+        this.timeoutExecutor = timeoutExecutor;
```
idelpivnitskiy, line 111: Consider using `requireNonNull` for all ctors that take an `Executor`

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
```diff
@@ -0,0 +1,96 @@
...
+        return new StreamingHttpServiceFilter(service) {
+            @Override
+            public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
+                                                        final StreamingHttpRequest request,
+                                                        final StreamingHttpResponseFactory responseFactory) {
+
+                return Single.defer(() -> {
+                    Duration timeout = timeoutForRequest.apply(request);
+
+                    Single<StreamingHttpResponse> response = delegate().handle(ctx, request, responseFactory);
```
idelpivnitskiy, line 72: The way it gets the response single is the only difference between client and server. Logic related to applying timeout for response payload body is similar to client-side filter. There should be an opportunity to share the code.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
```diff
@@ -0,0 +1,96 @@
...
+    @Nullable
+    private final Executor timeoutExecutor;
+
+    public TimeoutHttpServiceFilter(TimeoutHttpRequesterFilter.TimeoutFromRequest timeoutForRequest) {
+        this.timeoutForRequest = timeoutForRequest;
+        this.timeoutExecutor = null;
+    }
+
+    public TimeoutHttpServiceFilter(TimeoutHttpRequesterFilter.TimeoutFromRequest timeoutForRequest,
+                                    Executor timeoutExecutor) {
```
idelpivnitskiy, line 56: We should provide a similar option for users to decide if the timeout applied for response meta-data or the full response completion

servicetalk-examples/http/helloworld/src/main/java/io/servicetalk/examples/http/helloworld/async/HelloWorldClient.java
bondolo: OK, easily done.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
bondolo: In prior discussions the important part was to ensure that client requests made by servers were bound by the active time limit. In basic HTTP this will come from the TimeoutServiceFilter but in gRPC it will propagate from the originating client to the server and then be incorporated in to client requests.

Keep this feature for HTTP or not worth it? The issue for servers making client requests is determining the amount of time still available and the async context still seems to be the best option for that.

bondolo: Regarding not throwing: I discovered needing to catch the UnsupportedOperationException when I disabled the AsyncContext. The problem with re-throwing is that doing so would break the usage/application completely. Breaking the feature, passing the timeout in the async context, seemed preferable and a reasonable response to the AsyncContext being disabled.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
bondolo: The `@Nullable` and documentation are the primary value. Removing the `@Nullable` means that static analysis follows the potentially null reference and may issue warnings if you try to use the result without checking for null. The use in the `TimeoutHttpRequesterFilter` and `TimeoutHttpServiceFilter` both have explicit null checks so are fine. I wouldn't expect this type to be used many other places, if at all.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -30,46 +34,109 @@
...
+    }
+
+    /**
+     * Creates a new instance.
+     *
+     * @param timeoutForRequest function for extracting timeout from request
+     * @param fullRequestResponse if true then timeout is for full request/response transaction otherwise only the
+     * response metadata must arrive before the timeout.
+     */
+    public TimeoutHttpRequesterFilter(final TimeoutFromRequest timeoutForRequest, final boolean fullRequestResponse) {
```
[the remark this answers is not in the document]
bondolo: 1. Yes earlier iterations had the `TimeoutHttpRequesterFilter(Duration, boolean)` overload but when the propagation of timeout from server to client requests was added the `useDuration()` became the preferred option so as to include that behaviour. 
2. No, I assumed a given app would want one or the other but probably not both.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Construct a new instance.
+     *
+     * @param duration the timeout {@link Duration}
+     * @param timeoutExecutor the {@link Executor} to use for managing the timer notifications
+     */
+    public TimeoutHttpServiceFilter(Duration duration,
+                                    Executor timeoutExecutor) {
```
idelpivnitskiy, line 68: Can we make client/server filters consistent and also add an overload with `fullRequestResponse`? Sounds like a useful feature for the server-side as well to decide where timeout should be applied.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
```diff
@@ -0,0 +1,131 @@
...
+        this(simpleDurationTimeout(duration), timeoutExecutor);
+    }
+
+    /**
+     * Construct a new instance.
+     *
+     * @param timeoutForRequest function for extracting timeout from request which may also determine the timeout using
+     * other sources. If no timeout is to be applied then the function should return null.
+     */
+    public TimeoutHttpServiceFilter(TimeoutHttpRequesterFilter.TimeoutFromRequest timeoutForRequest) {
```
idelpivnitskiy, line 78: If you prefer a custom interface, consider making it a top level interface to avoid dependency between `TimeoutHttpServiceFilter` and `TimeoutHttpRequesterFilter`.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
```diff
@@ -0,0 +1,131 @@
...
+                    Single<StreamingHttpResponse> response = delegate().handle(ctx, request, responseFactory);
+
+                    Duration duration = timeoutForRequest.apply(request);
+
+                    if (null != duration) {
+                        Instant beganAt = Instant.now();
+                        response = (null != timeoutExecutor ?
+                                response.timeout(duration, timeoutExecutor) : response.timeout(duration))
+                                .map(resp -> resp.transformMessageBody(body -> Publisher.defer(() -> {
+                                    Duration remaining = duration.minus(Duration.between(beganAt, Instant.now()));
```
idelpivnitskiy, line 114: Handle negative `remaining` result

servicetalk-examples/http/timeout/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,32 @@
...
+    <Console name="Console" target="SYSTEM_OUT">
+      <PatternLayout pattern="%d %30t [%-5level] %-30logger{1} - %msg%n"/>
+    </Console>
+  </Appenders>
+  <Loggers>
+    <!-- Enables all wire logging messages -->
+    <Logger name="servicetalk-examples-wire-logger" level="TRACE"/>
+
+    <!-- Use `-Dservicetalk.logger.level=INFO` to change the root logger level via command line  -->
+    <Root level="${sys:servicetalk.logger.level:-DEBUG}">
```
idelpivnitskiy, line 28: This example doesn't use wire-logging and should log at `INFO` level by default.
Consider copying `log4j2.xml` from original http hello-world example.

servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutClient.java
```diff
@@ -0,0 +1,66 @@
...
+                                Duration.ofMinutes(2), true));
+
+        try (HttpClient client = builder.build()) {
+            // This example is demonstrating asynchronous execution, but needs to prevent the main thread from exiting
+            // before the response has been processed. This isn't typical usage for a streaming API but is useful for
+            // demonstration purposes.
+            CountDownLatch responseProcessedLatch = new CountDownLatch(1);
+            client.request(client.get("/sayHello"))
+                    // This request and response must complete within 1 minute or the request will be cancelled.
+                    .timeout(Duration.ofMinutes(1))
```
idelpivnitskiy, line 55: It's great idea to demonstrate both options (filter and operator). Consider making 2 requests inside this example (similar to compression example) when the first users a filter, the second also applies an operator. And clarify that when operator has a lower timeout, it will be triggered earlier than the timeout applied by the filter.

Alternative will be to build 2 different clients: one with a filter, one without. But it should be ok to demonstrate an operator with the same client.

servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutServer.java
```diff
@@ -0,0 +1,39 @@
...
+import static io.servicetalk.http.api.HttpSerializationProviders.textSerializer;
+
+/**
+ * Extends the async 'Hello World!' example to demonstrate use of timeout filter.
+ */
+public final class TimeoutServer {
+    public static void main(String[] args) throws Exception {
+        HttpServers.forPort(8080)
+                // Filter enforces that responses must complete within 2 minutes or will be cancelled.
+                .appendServiceFilter(new TimeoutHttpServiceFilter(Duration.ofMinutes(2)))
```
idelpivnitskiy, line 33: Duration in "minutes" looks too hight. Consider using seconds here and in the client filter.

idelpivnitskiy (commented):
Couple polishing comments and ship it. Can't wait to use these enhancements instead of a custom filters

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
```diff
@@ -0,0 +1,131 @@
...
+     * Construct a new instance.
+     *
+     * @param timeoutForRequest function for extracting timeout from request which may also determine the timeout using
+     * other sources. If no timeout is to be applied then the function should return null.
+     * @param timeoutExecutor the {@link Executor} to use for managing the timer notifications
+     */
+    public TimeoutHttpServiceFilter(TimeoutHttpRequesterFilter.TimeoutFromRequest timeoutForRequest,
+                                    Executor timeoutExecutor) {
+        this.timeoutForRequest = timeoutForRequest;
+        this.timeoutExecutor = timeoutExecutor;
```
idelpivnitskiy, line 93: `requireNonNull` for the function and executor

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Construct a new instance.
+     *
+     * @param timeoutForRequest function for extracting timeout from request which may also determine the timeout using
+     * other sources. If no timeout is to be applied then the function should return null.
+     */
+    public TimeoutHttpServiceFilter(TimeoutHttpRequesterFilter.TimeoutFromRequest timeoutForRequest) {
+        this.timeoutForRequest = timeoutForRequest;
```
idelpivnitskiy, line 79: requireNonNull(timeoutForRequest)

servicetalk-gradle-plugin-internal/src/main/groovy/io/servicetalk/gradle/plugin/internal/ServiceTalkLibraryPlugin.groovy
```diff
@@ -174,6 +174,8 @@ final class ServiceTalkLibraryPlugin extends ServiceTalkCorePlugin {
           showStandardStreams = true
         }
 
+		ignoreFailures = true
```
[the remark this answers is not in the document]
bondolo: It is for continuing past flaky tests. I keep accidentally adding it to the current changeset. I will remove it.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -97,7 +184,52 @@ public StreamingHttpConnectionFilter create(final FilterableStreamingHttpConnect
...
+         *
+         * @param request the current request
+         * @return The timeout or null for no timeout
+         */
+        @Override
+        @Nullable
+        Duration apply(HttpRequestMetaData request);
+
+        @Override
+        HttpExecutionStrategy influenceStrategy(HttpExecutionStrategy strategy);
```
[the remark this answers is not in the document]
bondolo: I use option (1) for the moment. Using option (2) would allow simple lambda or method reference implementations which would be nice.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
bondolo: I have refactored this code in the most recent revision to make the client and server even more similar. I didn't attempt, so far, to share code.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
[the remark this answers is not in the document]
bondolo: I think I prefer the first option for discoverability and documentation reasons.

servicetalk-examples/http/timeout/src/main/java/io/servicetalk/examples/http/timeout/TimeoutClient.java
```diff
@@ -0,0 +1,78 @@
...
+                        System.out.println(resp.toString((name, value) -> value));
+                        System.out.println(resp.payloadBody(textDeserializer()));
+                    });
+
+            // second request, with custom timeout that is lower than the client default
+            client.request(client.get("/sayHello"))
+                    // This request and response must complete within 3 seconds or the request will be cancelled.
+                    .timeout(Duration.ofSeconds(3))
+                    .afterFinally(responseProcessedLatch::countDown)
+                    .afterOnError(System.out::println)
```
idelpivnitskiy, line 68: Consider using `System.err` for errors

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -30,46 +32,161 @@
...
+            Single<StreamingHttpResponse> response = delegate.request(strategy, request);
+
+            Duration timeout = timeoutForRequest.apply(request);
+
+            if (null != timeout) {
+                Single<StreamingHttpResponse> timeoutResponse = timeoutExecutor != null ?
+                        response.timeout(timeout, timeoutExecutor) : response.timeout(timeout);
+
+                response = fullRequestResponse ?
+                        Single.defer(() -> {
```
idelpivnitskiy, line 172: You are already inside the `Single.defer` that handles `Single<StreamingHttpResponse>`, there is not need for another one

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
bondolo: The interface now also extends `HttpExecutionStrategyInfluencer` and for documentation, discoverability reasons  and (to be honest) so that it can be implemented by an anonymous inner class I chose to keep it as an explicit interface.

servicetalk-gradle-plugin-internal/src/main/groovy/io/servicetalk/gradle/plugin/internal/ServiceTalkLibraryPlugin.groovy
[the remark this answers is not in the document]
idelpivnitskiy: Reminder about this one

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
idelpivnitskiy: Any thoughts here?

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
bondolo: thank you.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpServiceFilter.java
bondolo: I have added fullRequestResponse support to the Service filter as well and the implementations are even more similar than before. I haven't tried to share the code between them yet as they implement slightly different interfaces and get the response in slightly different ways.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -132,57 +137,38 @@ public TimeoutHttpRequesterFilter(final TimeoutFromRequest timeoutForRequest,
...
+                if (null != timeout) {
+                    Single<StreamingHttpResponse> timeoutResponse = timeoutExecutor == null ?
+                            response.timeout(timeout) : response.timeout(timeout, timeoutExecutor);
+
+                    if (fullRequestResponse) {
+                        Instant deadline = Instant.now().plus(timeout);
+                        response = timeoutResponse.map(resp -> resp.transformMessageBody(body ->
+                                Publisher.defer(() -> {
+                                    Duration remaining = Duration.between(Instant.now(), deadline);
+                                    return (Duration.ZERO.compareTo(remaining) < 0 ?
```
idelpivnitskiy, line 160: Should it be `<= 0` here and in service filter?

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
```diff
@@ -51,4 +51,28 @@ default HttpExecutionStrategy influenceStrategy(HttpExecutionStrategy strategy)
...
     }
+
+    /**
+     * Returns a function which returns the provided default duration as the timeout duration to be used for any
+     * request.
+     *
+     * @param duration timeout duration or null for no timeout
+     * @return a function to produce a timeout using specified duration
+     */
+    static TimeoutFromRequest simpleDurationTimeout(@Nullable Duration duration) {
```
idelpivnitskiy, line 62: Any specific reason for making it public now (all members of a public interface are public)? The function is trivial and does not provide much value when constructors just take `Duration`. Let's move it back as a pkg-private static method of a class.

idelpivnitskiy (approved):
Let's merge this PR after `simpleDurationTimeout` is made pkg-private. The PR looks correct from the implementation and API point of view. Code sharing can be done in a follow-up PR, if ever necessary.
Thank you for the great improvement!

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
bondolo: Indeed

PR #1458, opened 2021-03-25
Publisher#scanWith(ScanWithMapper) incorrect/missing terminal after cancel

Scottmitch:
Motivation:
Publisher#scanWith(ScanWithMapper) may not deliver a terminal event
after cancel is invoked. The operator may also deliver an internally
generated error due to state management incorrectly indicated a terminal
event has already been seen (e.g. duplicate terminal with violates RS
spec) or invalid requestN demand was previously seen.

Modifications:
- cancel should not update state to TERMINATED

Result:
Publisher#scanWith(ScanWithMapper) will deliver terminal events after
cancel and won't deliver internal errors at incorrect times.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithPublisher.java
@@ -35,31 +35,7 @@ final class ScanWithPublisher<T, R> extends AbstractNoHandleSubscribePublisher<R
 
     ScanWithPublisher(Publisher<T> original, Supplier<R> initial, BiFunction<R, ? super T, R> accumulator,
                       Executor executor) {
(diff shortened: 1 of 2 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ScanWithPublisher.java
```diff
@@ -35,31 +35,7 @@
 
     ScanWithPublisher(Publisher<T> original, Supplier<R> initial, BiFunction<R, ? super T, R> accumulator,
                       Executor executor) {
-        this(original, () -> new ScanWithMapper<T, R>() {
```
Scottmitch, line 38: this change was just to consolidate conversion code outside the constructor (e.g. less noise) and add `requireNonNull` checks. I can break it out into a separate PR but was convenient while I was here and should have no external impact.

bondolo (approved):

tkountis (approved):

PR #1461, opened 2021-03-26
Update dependencies for site documentation

bondolo:
Motivation:
The documentation produced for the servicetalk.io website uses plugins
and tools which have been updated. Where practical recent versions of
these dependencies should be used.

Modifications:
Plugins and tool dependencies used for generating and verifying the
servicetalk.io site are updated to current or more recent versions.

Result:
Updated versions are used for generating documentation

### DIFF
--- a/docs/generation/gradle/validateSite.gradle
+++ b/docs/generation/gradle/validateSite.gradle
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019,2021 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -38,7 +38,6 @@ configurations {
 
 dependencies {
   nuValidator("nu.validator:validator:$nuValidatorVersion") {
-    exclude group: "org.eclipse.jetty"
     exclude group: "javax.servlet"
     exclude group: "commons-fileupload"
   }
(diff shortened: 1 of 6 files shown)


bondolo:
I was unable to update Antora to the latest version, 2.3.4, as the local site generated contains broken links. There appears to have been a change in how the site base URL is calculated beginning in 2.2.0.

bondolo:
I found the Antora 2.2.0 release notes which helped me figure out what needed to be changed so I was able to update to latest Antora, 2.3.4.

I compared the generated output of the current 2.0.0 version and the 2.3.4 and there is no significant difference obvious in the HTML, it has surprisingly few differences.

docs/generation/gradle/validateSite.gradle
```diff
@@ -74,7 +73,8 @@ for (location in ["local", "remote"]) {
 
     classpath = configurations.nuValidator
     main = "nu.validator.client.SimpleCommandLineValidator"
-    args "--skip-non-html" // --also-check-css doesn't work (still checks css as html), so limit to html files
+    args "--also-check-css"
```
tkountis, line 76: 🚀

tkountis:
> I was unable to update Antora to the latest version, 2.3.4, as the local site generated contains broken links. There appears to have been a change in how the site base URL is calculated beginning in 2.2.0.

Should we open a GH issue for that?

tkountis (approved):
Checkstyle issue, then LGTM

docs/generation/gradle/validateSite.gradle
bondolo: Just one of the many benefits of updating!

idelpivnitskiy:
Looks great! Thanks for taking care of it ❤️

PR #1462, opened 2021-03-27
Add client and server support for gRPC deadlines

bondolo:
Motivation:
Some gRPC usages require completion of calls within specified time limits. The gRPC protocol provides capability for specifying timeout (aka deadline) for completion of calls.

Modifications:
Features are added to support specification of call timeouts. 

Result:
ServiceTalk gRPC API now supports gRPC timeout functionality.


servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
```diff
@@ -168,4 +227,91 @@ public void doAppendHttpClientFilter(final Predicate<StreamingHttpRequest> predi
...
+            @Nullable
+            public Duration apply(HttpRequestMetaData request) {
+                Duration duration = readTimeoutHeader(request);
+
+                if (null != duration) {
+                    // use timeout from request
+                    return duration;
+                }
+
+                Instant deadline = AsyncContext.get(GRPC_DEADLINE_KEY);
```
idelpivnitskiy, line 250: Looks like you try to extract the deadline from `AsyncContext` twice: here and in `DefaultGrpcClientCallFactory#timeoutForRequest`. Can we do it only once?

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
```diff
@@ -168,4 +227,91 @@ public void doAppendHttpClientFilter(final Predicate<StreamingHttpRequest> predi
...
+                    assert false : "Bad digit: " + digitChar;
+                    return null;
+                } else {
+                    runningTotal = runningTotal * 10L + (long) (digitChar - '0');
+                }
+            }
+
+            return Duration.of(runningTotal, TIMEOUT_CHRONOUNITS[unitsIdx]);
+        } else {
+            assert null == value : "bad timeout value: " + value;
```
idelpivnitskiy, line 313: Why this assertion is required? It's legit to have no `grpc-timeout` on the request if users passed a value higher then the max value.

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
```diff
@@ -168,4 +227,91 @@ public void doAppendHttpClientFilter(final Predicate<StreamingHttpRequest> predi
...
+
+    /**
+     * Parse a gRPC HTTP timeout header value as a duration if present.
+     *
+     * @param value the text value of {@link #GRPC_TIMEOUT_HEADER_KEY} header to be parsed to a timeout duration or
+     * null if header was present
+     * @return The non-negative timeout duration which may null if not present or malformed
+     */
+    static @Nullable Duration parseTimeoutHeader(@Nullable CharSequence value) {
+        if (null != value && value.length() >= 2 && value.length() <= 9) {
```
idelpivnitskiy, line 286: We should probably throw if the value is malformed instead of returning a `null`

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
```diff
@@ -155,23 +186,71 @@ protected void doAppendHttpServiceFilter(final Predicate<StreamingHttpRequest> p
...
+                        // use server default (if any)
+                        : defaultTimeout;
+
+                if (null != duration) {
+                    try {
+                        AsyncContext.put(GRPC_DEADLINE_KEY, Instant.now().plus(duration));
+                    } catch (UnsupportedOperationException ignored) {
+                        // ignored -- async context has probably been disabled.
+                        // Timeout propagation will be partially disabled.
+                        // cancel()s will still happen which will accomplish the same effect though less efficiently
```
idelpivnitskiy, line 222: Consider logging when this happens. It's good to notify users when they use a feature that depends on `AsyncContext`, but it's disabled. It may accidentelly happen if they pull a dependency that disables `AsyncContext`. 
Not sure what will be a good level for that, maybe `debug` is sufficient.

idelpivnitskiy (changes requested):

idelpivnitskiy:
The overall approach looks great! My comments mostly about details

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -101,6 +154,28 @@ static void initRequest(final HttpRequestMetaData request, final List<ContentCod
         }
     }
 
+    /**
+     * Make a timeout header value from the specified duration.
+     *
+     * @param timeout the timeout duration
+     * @return The timeout header text value or null for infinite timeouts
+     */
+    static @Nullable CharSequence makeTimeoutHeader(Duration timeout) {
```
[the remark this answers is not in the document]
bondolo: Seems like a mistake as the annotation applies to the return type not the method.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -101,6 +154,28 @@ static void initRequest(final HttpRequestMetaData request, final List<ContentCod
...
+        if (GrpcMetadata.GRPC_MAX_TIMEOUT.compareTo(timeout) >= 0) {
+            // convert assuming that negative timeout is the result of an already missed deadline.
+            // cannot overflow as we have already checked against safe maximum
+            long timeoutValue = timeout.isNegative() ? 0 : timeout.toNanos();
+            int units = 0;
+            while (timeoutValue > GrpcMetadata.EIGHT_NINES) {
+                timeoutValue = CONVERTERS[units].applyAsLong(timeoutValue);
+                units++; // cannot go past end of units array as we have already range checked
+            }
+            return Long.toString(timeoutValue) + TIMEOUT_UNIT_CHARS[units];
```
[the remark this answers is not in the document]
bondolo: OK, I am curious why CharSequences#newAscciString doesn't take a CharSequence argument as the fromAscii method it calls does and in some cases a string conversion could be avoided.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
```diff
@@ -65,10 +68,33 @@
...
+            status = GrpcStatusCode.CANCELLED;
+        } else if (cause instanceof TimeoutException) {
+            // local timeout
+            status = GrpcStatusCode.DEADLINE_EXCEEDED;
+        } else {
+            // something else
+            status = GrpcStatusCode.UNKNOWN;
+        }
+
+        GrpcStatusException grpc = new GrpcStatusException(new GrpcStatus(status, cause), () -> null);
```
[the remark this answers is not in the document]
bondolo: From what I saw in `ProtocolCompatibilityTest` it appears to always throw the `StatusRuntimeException` for gRPC failures.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
```diff
@@ -65,10 +82,32 @@
...
+            statusCode = GrpcStatusCode.CANCELLED;
+        } else if (cause instanceof TimeoutException) {
+            // local timeout
+            statusCode = GrpcStatusCode.DEADLINE_EXCEEDED;
+        } else {
+            // something else
+            statusCode = GrpcStatusCode.UNKNOWN;
+        }
+
+        return new GrpcStatusException(new GrpcStatus(statusCode, cause), () -> null);
```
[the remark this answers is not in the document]
bondolo: It does not throw the original cause. From what I saw in ProtocolCompatibilityTest it appears to always throw the StatusRuntimeException for gRPC failures including deadline exceeded.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
```diff
@@ -124,11 +164,19 @@
...
+                final HttpResponse response = strategy == null ? client.request(httpRequest) :
+                        client.request(strategy, httpRequest);
+                return validateResponseAndGetPayload(response,
+                        serializationProvider.deserializerFor(
+                                readGrpcMessageEncoding(response, supportedCodings), responseClass));
+            } catch (Exception all) {
+                if (all instanceof RuntimeException) {
+                    throw all;
+                } else {
+                    throw toGrpcException(all);
```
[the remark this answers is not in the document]
bondolo: It was mostly to avoid additional wrapping and unwrapping but since the exceptions thrown aren't easily documented it seems reasonable to only throw `GrpcStatusException`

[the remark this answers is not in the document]
bondolo: It would be nice if `BlockingClientCall::request(GrpcClientMetadata metadata, Req request) throws Exception;` was declared to throw a narrower exception, perhaps just `GrpcStatusException`. Throwing raw Exception doesn't encourage good handling of exceptions.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcMetadata.java
```diff
@@ -15,15 +15,48 @@
...
+
+    /**
+     * Timeout after which the client no longer wants response.
+     *
+     * @return {@link Duration} of associated timeout. All durations greater than {@link #GRPC_MAX_TIMEOUT} will be
+     * treated as infinite (no deadline).
+     * @see <a href="https://grpc.io/blog/deadlines/">gRPC Deadlines</a>
+     */
+    default Duration timeout() {
+        return INFINITE_TIMEOUT;
```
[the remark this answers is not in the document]
bondolo: A default was provided to prevent breakage of existing code but if we don't think there are GrpcMetadata implementations I will remove the default.

bondolo (commented):
Thank you for the review feedback. I haven't resolved all of the issues yet so there will be additional changesets eventually.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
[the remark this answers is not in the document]
idelpivnitskiy: Good question! I think we can relax the argument to `CharSequence`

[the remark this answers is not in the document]
bondolo: I will file a separate PR for CharSequences#newAsciiString(Charsequence)

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
bondolo: This assert is associated with the case that the input is less than 2 characters or more than 9 characters long. The assert would not trigger if there was no Grpc-timeout header which it accomplishes, confusingly, but asserting `null == value`.

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
bondolo: Yes, reasonable. Is it acceptable to throw IllegalArgumentException in this and the other malformed header cases?

bondolo: In this case the IllegalArgumentException would be thrown from the `Single.defer()` returned by `TimeoutHttpRequesterFilter.request()` so will show upon the subscribe.

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
bondolo: now replaced by IllegalArgumentException

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
```diff
@@ -41,12 +48,22 @@
 import static java.util.Objects.requireNonNull;
 
 final class DefaultGrpcClientCallFactory implements GrpcClientCallFactory {
+    /**
+     * gRPC timeout is stored in context as a deadline so that when propagated to a new request the remaining time to be
+     * included in the request can be calculated.
+     */
+    static final AsyncContextMap.Key<Instant> GRPC_DEADLINE_KEY = AsyncContextMap.Key.newKey("grpc-deadline");
```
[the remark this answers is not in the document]
bondolo: I avoided adding a new module by add protected versions of the constant to both `GrpcServerBuilder` and `GrpcClientBuilder`. ie.

```
    /**
     * gRPC timeout is stored in context as a deadline so that when propagated to a new request the remaining time to be
     * included in the request can be calculated.
     */
    static final AsyncContextMap.Key<Instant> PKG_GRPC_DEADLINE_KEY = AsyncContextMap.Key.newKey("grpc-deadline");;

    /**
     * for sub-classes in other packages.
     */
    protected static final AsyncContextMap.Key<Instant> GRPC_DEADLINE_KEY = PKG_GRPC_DEADLINE_KEY;
```

I may refactor to abstract the key to a single class as you suggest.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
```diff
@@ -41,12 +48,22 @@
...
+
     private final StreamingHttpClient streamingHttpClient;
     private final GrpcExecutionContext executionContext;
+    @Nullable
+    private final Duration defaultTimeout;
```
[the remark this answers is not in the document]
bondolo: I have added an LRU cache of timeout values in `GrpcUtils` to avoid recreation of the `CharSequence` instances.

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
bondolo: I think it can be entirely eliminated from GrpcClientBuilder. The server operation already has a timeout filter running so client requests shouldn't need to redundantly add their own timeout filters. If the server operation times out and is  cancelled any pending client requests should be cancelled as well without them needing to have their own independent timers.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
```diff
@@ -204,14 +231,33 @@ public Completable onClose() {
...
     }
+
+    /**
+     * Determines the timeout for a new request using three potential sources; the client default, the client metadata
+     * or the async context. The lowest timeout will be used.
+     *
+     * @param metaDataTimeout the timeout specified in client metadata
+     * @return The timeout {@link Duration}, potentially negative.
+     */
+    private Duration timeoutForRequest(Duration metaDataTimeout) {
```
idelpivnitskiy, line 252: Hm, I think the manually provided `metaDataTimeout` should override the default value configured on the builder, when `GRPC_DEADLINE_KEY` is empty. Users may have higher timeout expectation for a specific endpoint, while others are fine with lower default value.

The `GRPC_DEADLINE_KEY` should have the highest priority if it's the lowest value, because otherwise the service filter will trigger cancellation.

`defaultTimeout` is more like a fallback if non of the above is set. It's not "maxTimeout", it's just a default.

WDYT about this order of priorities: `min(Duration.between(Instant.now(), AsyncContext.get(GRPC_DEADLINE_KEY)), metaDataTimeout) || defaultTimeout`?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -83,15 +119,61 @@ protected HttpHeaders payloadComplete(final HttpHeaders trailers) {
...
+                        throw new GrpcStatusException(new GrpcStatus(CANCELLED, cause), () -> null);
+                    }
+
+                    // local timeout
+                    if (cause instanceof TimeoutException) {
+                        // omit the cause unless logging has been configured at TRACE
+                        throw new GrpcStatusException(
+                                LOGGER.isTraceEnabled() ?
+                                    new GrpcStatus(DEADLINE_EXCEEDED, cause) :
+                                    DEADLINE_EXCEEDED.status(),
```
idelpivnitskiy, line 139: Why in case of `DEADLINE_EXCEEDED` the cause is not useful? The exception message provides a good idea abot what was the configured timeout value.
If you still prefer a condition based on the logger level, consider using `DEBUG` level. Users rarely use `TRACE` for all packages, usually only for specific logger names.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -167,14 +271,33 @@ static void setStatus(final HttpHeaders trailers, final Throwable cause, final B
...
-            GrpcStatus status = new GrpcStatus(UNIMPLEMENTED, cause, "Message encoding '" +
-                    msgEncException.encoding() + "' not supported ");
-            setStatus(trailers, status, null, allocator);
+            status = new GrpcStatus(UNIMPLEMENTED, cause, "Message encoding '" + msgEncException.encoding()
+                    + "' not supported ");
+        } else if (cause instanceof CancellationException) {
+            status = new GrpcStatus(CANCELLED, cause);
+        } else if (cause instanceof TimeoutException) {
+            // omit cause unless logging configured for trace
+            status = LOGGER.isTraceEnabled() ? new GrpcStatus(DEADLINE_EXCEEDED, cause) : DEADLINE_EXCEEDED.status();
```
idelpivnitskiy, line 289: same

idelpivnitskiy (commented):
Looks great! Couple more things to discuss and ready to be shipped

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
```diff
@@ -177,4 +255,57 @@ public void doAppendHttpClientFilter(final Predicate<StreamingHttpRequest> predi
...
+     */
+    static Duration parseTimeoutHeader(CharSequence grpcTimeoutValue) {
+        if (grpcTimeoutValue.length() < 2 || grpcTimeoutValue.length() > 9) {
+            throw new IllegalArgumentException("grpcTimeoutValue: " + grpcTimeoutValue +
+                    " (expected 2-9 characters)");
+        }
+
+        char unitChar = grpcTimeoutValue.charAt(grpcTimeoutValue.length() - 1);
+        int unitsIdx = 0;
+        while (unitsIdx < TIMEOUT_UNIT_CHARS.length && unitChar != TIMEOUT_UNIT_CHARS[unitsIdx]) {
```
Scottmitch, line 287: nit: `TIMEOUT_UNIT_CHARS` introduces some additional conditionals/loop that we could avoid with a `switch`:

```java
        switch (unitChar) {
            case 'n':
                units = ChronoUnit.NANOS;
                break;
            case 'u':
                units = ChronoUnit.MICROS;
                break;
                // ...
            default:
                throw new IllegalArgumentException("grpcTimeoutValue: " + grpcTimeoutValue +
                        " (Bad unit '" + unitChar + "')");
        }
```

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
```diff
@@ -165,23 +200,72 @@ protected void doAppendHttpServiceFilter(final Predicate<StreamingHttpRequest> p
...
+
+                duration = (null != duration) ?
+                        // As specified in the request
+                        duration
+                        // use server default (if any)
+                        : defaultTimeout;
+
+                if (null != duration) {
+                    try {
+                        AsyncContext.put(GRPC_DEADLINE_KEY, Instant.now().plus(duration));
```
Scottmitch, line 232: `Instant.now()` -> this is not monotonic IIUC, which may lead to durations being missed or unexpectedly delayed due to local time shifts. `Instant` has an association with a "clock" and IIRC none of the default implementations provide a monotonic option. We can avoid this by using `System.nanoTime()`:

```java
TimerState {
  long startTime = System.nanoTime();
  long nsDuration = duration.toNanos();
}
// set the timer
AsyncContext.put(GRPC_DEADLINE_KEY, new TimerState());


// compute remaining time
State s = AsyncContext.get(GRPC_DEADLINE_KEY);
long remianingNs = s.nsDuration - (System.nanoTime() - s.startTime);
```

There is a `TimeSource` (and `NormalizedTimeSource`) utility to help with time management, but it is currently internal to test utilities as it needs some work. `NormalizedTimeSource` allows for absolute offsets to be added and later checked (so you don't have to carry around start time).

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
idelpivnitskiy: Because if there is a `GRPC_DEADLINE_KEY ` value in async context it means that we are inside a service which has a timeout filter and that timeout will be triggered regardless of what user explicitly set on the meta-data. Therefore, there is not reason to propagate the higher value that user provided to downstream services if the current service is going to timeout earlier.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -71,10 +86,31 @@
...
     private static final ConcurrentMap<List<ContentCodec>, CharSequence> ENCODINGS_HEADER_CACHE =
             new ConcurrentHashMap<>();
     private static final CharSequence CONTENT_ENCODING_SEPARATOR = ", ";
+    /**
+     * Unlike {@link #CONTENT_ENCODING_SEPARATOR} it is possible or (depending on usage) likely that we will encounter a
+     * very large number of different duration values so we need to implement an LRU cache to limit the number of
+     * cached header values.
+     */
+    private static final Map<Duration, CharSequence> TIMEOUT_HEADER_VALUES_CACHE =
+            Collections.synchronizedMap(new LinkedHashMap<Duration, CharSequence>() {
```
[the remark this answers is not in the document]
idelpivnitskiy: What if we remove the cache now and consider it as a follow-up internal improvement? We should not overcomplicate for cases when the value is computed because the probability it will hit the cache is very low.

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
idelpivnitskiy: > `Instant.now()` -> this is not monotonic IIUC

If this is true, we should fix `TimeoutHttpRequesterFilter` and `TimeoutHttpServiceFilter` implementations as well

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcMetadata.java
```diff
@@ -15,15 +15,35 @@
...
     String path();
+
+    /**
+     * Timeout after which the client no longer wants response.
+     *
+     * @return {@link Duration} of associated timeout. All durations greater than {@link #GRPC_MAX_TIMEOUT} will be
+     * treated as infinite (no deadline).
+     * @see <a href="https://grpc.io/blog/deadlines/">gRPC Deadlines</a>
+     */
+    Duration timeout();
```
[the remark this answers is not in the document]
bondolo: Done

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
```diff
@@ -165,23 +200,72 @@ protected void doAppendHttpServiceFilter(final Predicate<StreamingHttpRequest> p
         httpServerBuilder.appendServiceFilter(predicate, factory);
     }
 
+    private HttpServerBuilder preBuild() {
+        if (!invokedBuild && null != defaultTimeout && GRPC_MAX_TIMEOUT.compareTo(defaultTimeout) >= 0) {
+            doAppendHttpServiceFilter(new TimeoutHttpServiceFilter(grpcDetermineTimeout()));
```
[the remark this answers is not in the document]
bondolo: 1. Historical reasons. `TimeoutHttpRequesterFilter` originally only supported the `false` behaviour

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
bondolo: The cause message is useful but the stack trace is probably not. In the examples I have seen the stack is quite large and there was not much point in ever potentially transmitting it or even keeping the stack trace. I did consider adding the cause message as the status details but that doesn't seem to be consistently used. I will switch it to DEBUG.

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
bondolo: Will fix this and create a separate issue for the TimeoutHttp filters

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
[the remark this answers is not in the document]
bondolo: I will remove the cache for now.

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
bondolo: OK

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
```diff
@@ -51,6 +54,18 @@
...
+     * gRPC timeout is stored in context as a deadline so that when propagated to a new request the remaining time to be
+     * included in the request can be calculated.
+     */
+    static final AsyncContextMap.Key<Instant> PKG_GRPC_DEADLINE_KEY = AsyncContextMap.Key.newKey("grpc-deadline");
+
+    /**
+     * gRPC timeout is stored in context as a deadline so that when propagated to a new request the remaining time to be
+     * included in the request can be calculated.
+     */
+    protected static final AsyncContextMap.Key<Instant> GRPC_DEADLINE_KEY = PKG_GRPC_DEADLINE_KEY;
```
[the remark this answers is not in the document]
bondolo: Doing so would require adding another module "servicetalk-grpc-internal' and a class to share the "secret" async context key between modules. I avoided that for now on the expectation that there aren't going to be user implementations of the GrpcServerBuilder and GrcpcClientBuilder

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientMetadata.java
```diff
@@ -26,42 +28,126 @@
  */
 public class DefaultGrpcClientMetadata extends DefaultGrpcMetadata implements GrpcClientMetadata {
 
+    /**
+     * No timeout or infinite timeout
+     */
+    static final Duration INFINITE_TIMEOUT = java.time.Duration.ofSeconds(Long.MAX_VALUE, 999_999_999);
```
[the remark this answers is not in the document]
bondolo: I tried too hard to keep null out of the public API.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
bondolo: I am fine with the suggestion.

idelpivnitskiy:
LGTM, thanks for the great work!

PR #1465, opened 2021-03-30
Reduce initial size of listeners queue in `WriteStreamSubscriber`

idelpivnitskiy:
Motivation:

Most of the write HTTP messages have 3 listeners, because it consist of
headers, one payload body chunk (can be empty), and trailers. `ArrayDeque`
defines `MIN_INITIAL_CAPACITY = 8` internally. It's sufficient for most
use-cases. Messages with large streaming payload body may have more
listeners, but those messages have higher overhead anyway (flush-on-each,
etc.). Therefore, it's ok for them to resize the queue, if necessary.

Modifications:

- Define the default size of 8 for `listenersOnWriteBoundaries` in
`WriteStreamSubscriber.AllWritePromise`;

Result:

Less allocation per each message write.


bondolo (approved):
Seems reasonable.

PR #1468, opened 2021-04-02
Update JMH 1.28 -> 1.29 and me.champeau.jmh plugin 0.5.3 -> 0.6.3

idelpivnitskiy:
Motivation:

0.5.x plugin version can not run newer JMH versions.

### DIFF
--- a/servicetalk-benchmarks/build.gradle
+++ b/servicetalk-benchmarks/build.gradle
@@ -16,7 +16,7 @@
 
 plugins {
   id "com.github.johnrengelman.shadow" version "6.1.0"
-  id "me.champeau.gradle.jmh" version "0.5.3"
+  id "me.champeau.jmh" version "0.6.3"
 }
 
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-library"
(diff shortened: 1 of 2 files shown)


servicetalk-benchmarks/build.gradle
```diff
@@ -43,11 +43,11 @@ dependencies {
 }
 
 jmh {
-  include = ".*Benchmark"
+  includes = [".*Benchmark"]
   jmhVersion = "$jmhCoreVersion"
-  jvmArgsPrepend = "-Dio.netty.maxDirectMemory=9223372036854775807 " +
-                   "-Djmh.executor=CUSTOM " +
-                   "-Djmh.executor.class=io.servicetalk.benchmark.concurrent.AsyncContextFriendlyExecutor"
+  jvmArgsPrepend = ["-Dio.netty.maxDirectMemory=9223372036854775807 " +
```
idelpivnitskiy, line 48: The plugin was reworked. It uses a new method of running JMH now, which does not cause issues with gradle cache, but the API for configuration is a bit buggy. For example, it requires an array now, but still needs manual string assembly instead of adding properties as separate array elements.

bondolo (approved):
Having to use array syntax without getting any benefit is kind of odd, but looks fine. 🤷‍♂️

idelpivnitskiy:
I tried to re-run this build 6+ times and it constantly fails with one of the flaky tests, usually #1040 or #999. Because it doesn't change any code related to the main project, I did verify it locally, and #1469 passed the PRB (includes this commit), I will merge it as-is.

PR #1469, opened 2021-04-02
Implement `CharSequences#parseLong(CharSequence)`

idelpivnitskiy:
Motivation:

Our `HttpHeaders` API uses `CharSequence`. In order for users to parse
the value (like, `Content-Length`) they have to convert `CharSequence`
to `String` for `Long.parseLong(String)` API. This approach does an extra
copy of data from underlying `Buffer`. We can avoid it.

Modifications:

- Add `CharSequences#parseLong(CharSequence)` API that does not force
`CharSequence#toString()` conversion;
- Add tests to validate new API;
- Use the new API in all places when we parse an `AsciiBuffer` or
`AsciiString`;

Result:

Less allocations of `String`.

This PR depends on #1468 (ignore first commit).

### DIFF
--- a/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
+++ b/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
@@ -443,7 +443,8 @@ public final class CharSequences {
     /**
      * Parses the {@link CharSequence} argument as a signed decimal {@code long}.
      *
-     * <p> This is the equivalent of {@link Long#parseLong(String)} that does not require to {@link String} conversion.
+     * <p> This is the equivalent of {@link Long#parseLong(String)} that does not require to
+     * {@link CharSequence#toString()} conversion.
      *
      * @param cs a {@code CharSequence} containing the {@code long} value to be parsed
      * @return {code long} representation of the passed {@link CharSequence}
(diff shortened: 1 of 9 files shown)


servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
```diff
@@ -420,4 +439,125 @@ static boolean contentEqualsIgnoreCaseUnknownTypes(final CharSequence a, final C
...
+        int i = 0;
+        final char firstCh = cs.charAt(i);
+        final boolean negative = firstCh == '-';
+        if ((negative || firstCh == '+') && ++i == length) {
+            throw illegalInput(cs);
+        }
+
+        long result = 0;
+        while (i < length) {
+            final int digit = Character.digit(cs.charAt(i++), RADIX_10);
```
bondolo, line 477: Character.digit will also consider unicode characters which shouldn't be allowed in headers and could be used for nefarious purposes by exploiting the parsing differences. Should consider only ASCII digits I think. (and it is probably faster)

bondolo (approved):
Looks good except for the one use of Character.digit()

servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/CharSequences.java
idelpivnitskiy: Because it parses the `CharSequence` and the method is located in non-http package, we should probably be generic and support any encoding by default (identical to `Long.parseLong`). Only when we know that it's `AsciiBuffer` we can apply ascii digits decoding.

PR #1470, opened 2021-04-05
CharSequences.newAsciiString accept CharSequence

bondolo:
Motivation:
Currently the method `CharSequences.newAsciiString` takes a String
argument but the implementation is fine with any CharSequence. Allowing
the method to accept CharSquence avoids some cases of unncessary
conversion to String.

Modifications:
The String argument is replaced with a more general CharSequence.
Existing code using this method will need to be recompiled.

Result:
CharSequences.newAsciiString() generality is improved.


idelpivnitskiy (approved):
🚀

PR #1479, opened 2021-04-13
Optimize HTTP messages with empty payload body

idelpivnitskiy:
Motivation:

In many cases users do not set payload body when they create an HTTP
request (like `GET`) or response (like, 204). We can optimize flush
strategy and HTTP/1.1 encoding in these cases.

Modifications:

- Introduce `PayloadInfo#isEmpty()` flag that tells internal layers when
an HTTP message contains payload body;
- Use this flag to choose a flush strategy and set `content-length` header;
- Adjust existing `content-length` tests to verify new use-cases with
empty payload body;
- Enhance existing `FlushStrategyOnServerTest` to verify empty responses
flush on end;
- Filter out empty payload body buffer when converting between types;
- Add more tests to verify users can keep using different ways to set payload;

Behavior changes:
- Users cannot write more data into a payload body `Buffer` of new
request/response objects after types conversion:
```
HttpRequest request = streamingHttpClient().asClient().post(SVC_ECHO);
StreamingHttpRequest streamingRequest = request.toStreamingRequest();
// Too late to take the buffer from the original aggregated request object, modifications won't be visible:
Buffer payload = request.payloadBody();
payload.writeAscii(CONTENT);
assertResponse(makeRequest(streamingRequest), request.version(), OK, 0);
```
- Streaming HTTP/1.1 requests and responses with empty payload body are
encoded with `content-length: 0` instead of `transfer-encoding: chunked`;

Result:

Less flushes + non-chunked HTTP/1.1 encoding when there is no payload body
results in better throughput (+30% RPS) for these use-cases, latency improves
by ~15%.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpApiConversions.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpApiConversions.java
@@ -184,6 +184,16 @@ public final class HttpApiConversions {
     }
 
     /**
+     * Checks whether a request/response payload body is empty.
+     *
+     * @param metadata The request/response to check.
+     * @return {@code true} is the request/response payload body is empty, {@code false} otherwise.
+     */
+    public static boolean isPayloadEmpty(HttpMetaData metadata) {
+        return (metadata instanceof PayloadInfo && ((PayloadInfo) metadata).isEmpty());
+    }
+
+    /**
      * Checks whether a request/response payload is safe to aggregate, which may allow for writing a `content-length`
      * header.
      *
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/PayloadInfo.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/PayloadInfo.java
@@ -24,6 +24,15 @@ import io.servicetalk.concurrent.api.Publisher;
  */
 interface PayloadInfo {
     /**
+     * Returns {@code true} if and only if, the payload body {@link Publisher} was never assigned, changed, or modified,
+     * and therefore remains empty.
+     *
+     * @return {@code true} if and only if, the payload body {@link Publisher} was never assigned, changed, or modified
+     * and therefore remains empty.
+     */
+    boolean isEmpty();
+
+    /**
      * Returns {@code true} if and only if, the {@link Publisher} associated with this {@link PayloadInfo} can be safely
      * aggregated to bring all data in memory. Inputs to this decision is left to the user of the API.
      *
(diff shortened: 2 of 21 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/PayloadInfo.java
```diff
@@ -23,6 +23,15 @@
  * {@link StreamingHttpResponse}.
  */
 interface PayloadInfo {
+    /**
+     * Returns {@code true} if and only if, the payload body {@link Publisher} was never assigned, changed, or modified,
+     * and therefore remains empty.
+     *
+     * @return {@code true} if and only if, the payload body {@link Publisher} was never assigned, changed, or modified
```
bondolo, line 30: What would be result for an explicitly empty payload such as empty String or empty byte array? 

I am thinking of the case were app has their own empty payload response and they use it as the payload for a response they intend to be empty.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
```diff
@@ -119,9 +123,9 @@ private static void updateRequestContentLengthNonZero(final int contentLength, f
         }
     }
 
-    private static void updateRequestContentLength(final int contentLength, final HttpHeaders headers) {
```
bondolo, line 122: Surprised to see that `contentLength` is not a `long` argument

bondolo (approved):
Seems like a quite worthwhile improvement!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpApiConversions.java
```diff
@@ -183,6 +183,16 @@ public static ServiceAdapterHolder toStreamingHttpService(BlockingHttpService se
...
     }
 
+    /**
+     * Checks whether a request/response payload body is empty.
+     *
+     * @param metadata The request/response to check.
+     * @return {@code true} is the request/response payload body is empty, {@code false} otherwise.
+     */
+    public static boolean isPayloadEmpty(HttpMetaData metadata) {
+        return (metadata instanceof PayloadInfo && ((PayloadInfo) metadata).isEmpty());
```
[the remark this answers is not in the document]
idelpivnitskiy: `PayloadInfo` is an internal interface. If users wrap the request/response types they do not have access to `PayloadInfo` to implement it. As the result, they will loose all optimizations we apply. I have plans to think about making the `PayloadInfo` public in future.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
idelpivnitskiy: This code executes only when we aggregate payload body into a single `Buffer`. The `Buffer` has the maximum length of `int`: `readableBytes()` returns `int`. So, we should not care about int overflow here.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/PayloadInfo.java
idelpivnitskiy: Unfortunately, in order to support [this](https://github.com/apple/servicetalk/pull/1479/files#diff-06e897b4cfb03b8c7b26e0a72492fae2bc8406a983f0a1e529afb0172027c529R56-R62) use-case, we can not apply the optimization in the scenarios you highlighted, because users can expand it later. If in future we will agree to drop support for this way, we will be able to optimize any empty payload body.

However, if users set a read-only empty buffer, we can support it. Fixed: https://github.com/apple/servicetalk/pull/1479/commits/69e962fce3da0d562daa6e6d4ffad8094ec12c55

tkountis (approved):

PR #1481, opened 2021-04-14
Use only monotonic clock in deadline calculations

bondolo:
Motivation:
The default clock used by `java.time.Instant` is not monotonic. This
could result in deadlines not being correctly managed if the clock is
adjusted during the course of timeout executing.
Modifications:
Replace use of `Instant` with `System.nanoTime()`
Result:
Timeout HTTP filters have monotonic behaviour.


idelpivnitskiy (approved):
🚀

PR #1485, opened 2021-04-15
Enable back SHA256 and SHA512 checksums for published artifacts

idelpivnitskiy:
Motivation:

We had to disable SHA256 and SHA512 checksums until Sonatype and Maven Central
support them. The referenced issues were closed. We can start publishing artifacts
with SHA256 and SHA512 checksum files.

Modification:

- Remove a parameter from `gradle.properties` that did disable this feature;

Result:

Gradle will generate  SHA256 and SHA512 checksum files for published artifacts.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -64,9 +64,3 @@ commonsLangVersion=2.6
 grpcVersion=1.37.0
 javaxAnnotationsApiVersion=1.3.5
 jsonUnitVersion=2.8.1
-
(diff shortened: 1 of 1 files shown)


bondolo:
I did a spot check and found that there were artifacts in the repo we use that include SHA256 hashes so I believe this should work.

bondolo (approved):

tkountis (approved):

PR #1486, opened 2021-04-15
Use consistent defaults for HTTP Timeout filters

bondolo:
Motivation:
There are two timeout filters for HTTP, `TimeoutHttpRequesterFilter`
and `TimeoutHttpServiceFilter` but they have differing defaults for the
behaviour of whether the timeout applies to the entire response or only
the response metadata.
Modifications:
Both filters now have the same set of constructors and defaults.
Result:
Reduced opportunity for API misuse due to inconsistent behaviour.

### DIFF
--- a/servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/GatewayServer.java
+++ b/servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/GatewayServer.java
@@ -122,7 +122,7 @@ public final class GatewayServer {
                                 .maxRetries(3)
                                 .buildWithExponentialBackoffDeltaJitter(ofMillis(100), ofMillis(50), ofSeconds(30)))
                         // Apply a timeout filter for the client to guard against latent clients.
-                        .appendClientFilter(new TimeoutHttpRequesterFilter(ofMillis(500)))
+                        .appendClientFilter(new TimeoutHttpRequesterFilter(ofMillis(500), false))
                         // Apply a filter that returns an error if any response status code is not 200 OK
                         .appendClientFilter(new ResponseCheckingClientFilter(backendName))
                         .ioExecutor(ioExecutor)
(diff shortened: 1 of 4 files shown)


servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -60,25 +60,6 @@
...
-        this(simpleDurationTimeout(duration), false);
-    }
-
-    /**
-     * Creates a new instance which requires only that the response metadata be received before the timeout.
-     *
-     * @param duration the timeout {@link Duration}
-     * @param timeoutExecutor the {@link Executor} to use for managing the timer notifications
-     */
-    public TimeoutHttpRequesterFilter(final Duration duration, final Executor timeoutExecutor) {
```
idelpivnitskiy, line 78: We can not remove these two contractors because they were present in the previous version.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -90,20 +71,6 @@ public TimeoutHttpRequesterFilter(final Duration duration, final boolean fullReq
...
-     * Creates a new instance.
-     *
-     * @param duration the timeout {@link Duration}
-     * @param fullRequestResponse if true then timeout is for full request/response transaction otherwise only the
-     * response metadata must arrive before the timeout.
-     * @param timeoutExecutor the {@link Executor} to use for managing the timer notifications
-     */
-    public TimeoutHttpRequesterFilter(final Duration duration,
-                                      final boolean fullRequestResponse,
-                                      final Executor timeoutExecutor) {
```
idelpivnitskiy, line 103: This looks like a useful ctor to me when users don't want to care about `TimeoutFromRequest` function.

idelpivnitskiy (changes requested):
I think the set of constructors for `TimeoutHttpRequesterFilter` was reasonable:
```
Duration
Duration, boolean
TimeoutFromRequest, boolean
```
\+ variants with `Executor` for each. Can we make `TimeoutHttpServiceFilter` constructors identical?

bondolo:
The two classes are now synchronized in both using `false` as default for fullRequestResponse and both now have the same full set of constructors.

idelpivnitskiy (approved):
❤️

PR #1487, opened 2021-04-15
[DO NOT MERGE] Reproduce ProtocolCompatibilityTest.serviceTalkToGrpcJavaTimeout test failure

idelpivnitskiy:
Trying to reproduce https://github.com/apple/servicetalk/pull/1485/checks?check_run_id=2348341117 and get more logs from CI.


idelpivnitskiy:
We found the reason, it's described in https://github.com/apple/servicetalk/issues/1489#issuecomment-820563188

PR #1491, opened 2021-04-15
Fix duration comparison bug in timeout filters and reduce code duplication

idelpivnitskiy:
Motivation:

`Duration#compareTo` is hard to understand, the result of comparison
depends on the order of arguments. This lead to a bug in
`TimeoutHttpRequesterFilter` and `TimeoutHttpServiceFilter` which always
fails payload body publisher with `TimeoutException`. Use a shared
utility to make `isPositive` check for `Duration` consistent across all modules.
Also, `TimeoutHttpRequesterFilter` and `TimeoutHttpServiceFilter` logic
is identical, we can reduce duplication and share the logic.

Modifications:

- Introduce a shared `DurationUtils` class for common operations with
`Duration`;
- Use this utility to reduce mistakes when we check that the passed
`Duration` must be positive;
- Introduce `AbstractTimeoutHttpFilterTest` to share timeout logic
between `TimeoutHttpRequesterFilter` and `TimeoutHttpServiceFilter`;
- Add tests to verify behavior correctness for both timeout filters;
- Adjust how the timeout filters are created in `ResponseTimeoutTest`
to avoid NPE;
- Minor improvements to make javadoc consistent between
`TimeoutHttpRequesterFilter` and `TimeoutHttpServiceFilter`;

Result:

1. `TimeoutHttpRequesterFilter` and `TimeoutHttpServiceFilter` do not
fail payload body publisher before the actual timeout expires.
2. No code duplication between timeout filters.
3. Shared utility to consistently validate `Duration` in all modules.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
@@ -168,8 +168,8 @@ public class H2ResponseCancelTest extends AbstractNettyHttpServerTest {
                         // ignore
                     }
                 })
-                .expectError()
-                // .thenCancel()
+                // FIXME: use thenCancel() after await() instead of cancelling from inside then(...) + expectError()
+                .expectError(IllegalStateException.class)   // should never happen
                 .verify();
 
         assertThat("Unexpected responses", responses, is(empty()));
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
@@ -169,6 +169,7 @@ public class H2ResponseCancelTest extends AbstractNettyHttpServerTest {
                     }
                 })
                 // FIXME: use thenCancel() after await() instead of cancelling from inside then(...) + expectError()
+                // https://github.com/apple/servicetalk/issues/1492
                 .expectError(IllegalStateException.class)   // should never happen
                 .verify();
 
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.utils;
+
+import io.servicetalk.concurrent.api.Executor;
+import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.http.api.HttpExecutionStrategy;
+import io.servicetalk.http.api.HttpExecutionStrategyInfluencer;
+import io.servicetalk.http.api.HttpRequestMetaData;
+import io.servicetalk.http.api.StreamingHttpRequest;
+import io.servicetalk.http.api.StreamingHttpResponse;
+
+import java.time.Duration;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Function;
+import javax.annotation.Nullable;
+
+import static io.servicetalk.concurrent.api.Publisher.defer;
+import static io.servicetalk.concurrent.api.Publisher.failed;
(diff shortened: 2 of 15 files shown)


servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/DurationUtils.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.time.Duration;
+
+import static java.time.Duration.ZERO;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Helper utilities for {@link Duration}.
+ */
+public final class DurationUtils {
```
idelpivnitskiy, line 26: This class will have more utility methods in a follow-up PRs

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
```diff
@@ -0,0 +1,128 @@
...
+                if (fullRequestResponse) {
+                    final long deadline = System.nanoTime() + timeout.toNanos();
+                    response = timeoutResponse.map(resp -> resp.transformMessageBody(body -> defer(() -> {
+                        final Duration remaining = ofNanos(deadline - System.nanoTime());
+                        if (isPositive(remaining)) {
+                            return (timeoutExecutor == null ?
+                                    body.timeoutTerminal(remaining) : body.timeoutTerminal(remaining, timeoutExecutor))
+                                    .subscribeShareContext();
+                        }
+                        return failed(new TimeoutException("timeout after " + timeout.toMillis() + "ms"));
```
idelpivnitskiy, line 94: @bondolo I've removed `subscribeShareContext()` from here because it's not necessary to apply it in case we return failure right away. It's a static async source that does not contain any user code that can modify the context.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
```diff
@@ -168,8 +168,8 @@ private void requestCancellationResetsStreamButNotParentConnection(StreamingHttp
                         // ignore
                     }
                 })
-                .expectError()
-                // .thenCancel()
+                // FIXME: use thenCancel() after await() instead of cancelling from inside then(...) + expectError()
```
Scottmitch, line 171: can you create an issue and link it here?

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
```diff
@@ -0,0 +1,128 @@
...
+    public final HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strategy) {
+        return timeoutForRequest.influenceStrategy(strategy);
+    }
+
+    final Single<StreamingHttpResponse> withTimeout(final StreamingHttpRequest request,
+                final Function<StreamingHttpRequest, Single<StreamingHttpResponse>> responseFunction) {
+
+        return Single.defer(() -> {
+            final Duration timeout = timeoutForRequest.apply(request);
+            if (null != timeout && !isPositive(timeout)) {
```
Scottmitch, line 76: do we need todo the pre-checks? ideally the timeout operators handle this case and we can simplify here. we would pay a bit more cost to construct the operators, but if the operators don't handle this everywhere we apply a timeout would have todo a pre-check (which is not ideal).

idelpivnitskiy: currently - yes. it's a feature of `timeoutForRequest`. if it returns `null`, it means no timeout. If it returns `ZERO` or negative => users want timeout right away.
I will keep it as-is for now. If we will move this logic to timeout operators in a follow-up, we can simplify here. For now, limiting the scope to the filters only

bondolo: The operators do handle it, they explicitly accept negative duration which is expected to occur as a result of calculated timeouts. The appearance here of pre-checks was as a result of earlier review feedback to simplify the flow.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
bondolo: Understood (now). Thank you for the explanation.

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/DurationUtils.java
bondolo: I will move the private `Publisher.toNanos(Duration)` here.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
```diff
@@ -0,0 +1,128 @@
+/*
+ * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
```
bondolo, line 2: Year

bondolo (approved):
Very nice refactoring and much improved tests.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
idelpivnitskiy: https://github.com/apple/servicetalk/issues/1492
Also added this link for `FIXME` comments: https://github.com/apple/servicetalk/pull/1491/commits/43dff7cf63df3fa914e492fb4ebe3dcdb017bfc5

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
Scottmitch: `null` is is intended to impact control flow (e.g. `null` means "no timeout") so it is expected to check for `null`. however if there is a timeout value the timeout operators should be able to handle it.

idelpivnitskiy: I would consider that as a follow-up only if the operator improved. Right now operator will always schedule a task, which creates more overhead compare to this check.

Scottmitch: followup sounds good: https://github.com/apple/servicetalk/pull/1494

PR #1493, opened 2021-04-15
Move `Publisher#toNanos` to `DurationUtils`

idelpivnitskiy:
Motivation:

`Publisher` is already a complicated class. We should not expand its
scope by additional utilities, even if they are internal. We have a
better place for `Publisher#toNanos` now - `DurationUtils`.

Modifications:

- Move `Publisher#toNanos` to `DurationUtils`;

Result:

Less utilities in `Publisher`, all `compareTo` operations for
`Duration` are in a single place.

### DIFF
--- a/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/DurationUtils.java
+++ b/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/DurationUtils.java
@@ -18,12 +18,23 @@ package io.servicetalk.utils.internal;
 import java.time.Duration;
 
 import static java.time.Duration.ZERO;
+import static java.time.Duration.ofNanos;
 import static java.util.Objects.requireNonNull;
 
 /**
  * Helper utilities for {@link Duration}.
  */
 public final class DurationUtils {
+
+    /**
+     * Maximum positive duration which can be expressed as a signed 64-bit number of nanoseconds.
+     */
+    private static final Duration LONG_MAX_NANOS = ofNanos(Long.MAX_VALUE);
+
+    /**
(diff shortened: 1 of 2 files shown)


servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/DurationUtils.java
```diff
@@ -54,4 +65,17 @@ public static Duration ensurePositive(final Duration duration, final String name
...
     }
+
+    /**
+     * Converts a {@code Duration} to nanoseconds or if the resulting value would overflow a 64-bit signed integer then
+     * either {@code Long.MIN_VALUE} or {@code Long.MAX_VALUE} as appropriate.
+     *
+     * @param duration the duration to convert
+     * @return the converted nanoseconds value
+     */
+    public static long toNanos(final Duration duration) {
```
Scottmitch, line 76: does it make sense to qualify the name to indicate under/overflow protection? For example in `FlowControlUtils` we use [WithUnderOverflowProtection](https://github.com/apple/servicetalk/blob/main/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/FlowControlUtils.java#L122).

Scottmitch (approved):

bondolo (approved):

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/DurationUtils.java
idelpivnitskiy: +1, let's keep as-is. It's an internal method anyway

PR #1494, opened 2021-04-15
HttpTimeoutFilter include original timeout on message body timeout

Scottmitch:
Motivation:
AbstractTimeoutHttpFilter can now apply a timeout on the metadata and
the message body level. If the message body level timeout triggers the
interval is taken as the user will see a value which is different than
the originally supplied timeout duration.

Modifications:
- Add a MappedTimeoutException that can wrap a TimeoutException and
  allows for specifying the original timeout value
- Remove pre-check conditions on timeout duration which is enforced by
  the timeout operators.

Result:
HttpTimeoutFilter exception includes the original timeout duration.

### DIFF
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
@@ -29,9 +29,7 @@ import java.util.function.Function;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.api.Publisher.defer;
-import static io.servicetalk.concurrent.api.Publisher.failed;
 import static io.servicetalk.utils.internal.DurationUtils.ensurePositive;
-import static io.servicetalk.utils.internal.DurationUtils.isPositive;
 import static java.time.Duration.ofNanos;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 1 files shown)


servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
```diff
@@ -73,10 +71,6 @@ public final HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy
 
         return Single.defer(() -> {
             final Duration timeout = timeoutForRequest.apply(request);
-            if (null != timeout && !isPositive(timeout)) {
-                return Single.failed(new TimeoutException("non-positive timeout of " + timeout.toMillis() + "ms"));
```
Scottmitch, line 77: doing the short-circuit evaluation also bypasses the `timeoutExecutor`. perhaps we want todo this but ideally we can consolidate the behavior in the operators to ensure consistency.

bondolo (commented):
I am unsure on the short-circuiting issue. I would like to think that it should be possible for this kind of short circuiting to safely take place as long as it can be done safely. Having to invoke the entire timeout machinery so that the threading/control flow would be consistent seems too sensitive.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/AbstractTimeoutHttpFilter.java
Scottmitch: Agreed we should be more explicit about offloading and executor selection w.r.t to timeouts (especially when we apply timeout filters in the background). I'll create an issue to track that. However this PR doesn't change existing behavior and just ensures we always delegate to the timeout operator to make the decision. When we do revisit this topic all we have todo is make sure we supply the right executor, and the executor can determine if it needs to schedule a task if the duration is `<=0` or invoke the runnable synchronously.

Also note that @bondolo is re-working the offloading which will have impacts here.

idelpivnitskiy: Agreed, let's make the behavior consistent with 0.38.1 and we can work on further improvements.

idelpivnitskiy (approved):
Thanks!

PR #1495, opened 2021-04-15
Fix unsubscribed message-body when isEmpty

tkountis:
Motivation:
With a recent optimization we missed a message-body subscription which had as a result some operations to not run.

Modifications:
Subscribe in the empty message-body.

Result:
All message-body operators should now run.


tkountis:
Build failure due to https://github.com/apple/servicetalk/issues/1192

idelpivnitskiy (approved):
Great catch! 🎖️ 
In a follow-up, please add a test for this use-case.

PR #1496, opened 2021-04-15
Add `DurationUtils#isInfinite` utility

idelpivnitskiy:
Motivation:

gRPC module checks if the timeout is considered infinite or not in
multiple places. All use different way of `compareTo` check.

Modifications:
- Introduce `DurationUtils#isInfinite` utility and test it;
- Use the new utility in all places where required;

Result:

Consistent approach to verify when timeout should be handled as infinite.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientMetadata.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientMetadata.java
@@ -23,6 +23,7 @@ import javax.annotation.Nullable;
 
 import static io.servicetalk.encoding.api.Identity.identity;
 import static io.servicetalk.utils.internal.DurationUtils.ensurePositive;
+import static io.servicetalk.utils.internal.DurationUtils.isInfinite;
 
 /**
  * Default implementation for {@link DefaultGrpcClientMetadata}.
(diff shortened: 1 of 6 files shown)


bondolo (commented):
I am less convinced of the value of this utility, it seems to incorporate too much specific semantics of the use. I would almost have it include the replacement value.

idelpivnitskiy:
The goal is to consolidate the logic and avoid different ways of getting the same result, like
```
null != timeout && timeout.compareTo(GRPC_MAX_TIMEOUT) <= 0
null == timeout || timeout.compareTo(GRPC_MAX_TIMEOUT) > 0
null == defaultTimeout || GRPC_MAX_TIMEOUT.compareTo(defaultTimeout) < 0
```
It's easier to read the method name instead of thinking about the result of the `if` statement.

bondolo:
I understand the sentiment. Perhaps it should be specific to the gRPC implementation and include the value to map to if "infinite"?

idelpivnitskiy:
Because it's used in `grpc-api` and `grpc-netty` I do not want to create a module just for a single method. `DurationUtils` is already in internal module. Should be ok to have it there.
The 2nd argument for `GRPC_MAX_TIMEOUT` looked as a nice compromise, but open for other suggestions.

bondolo:
+1 for not introducing another module

bondolo (approved):

PR #1497, opened 2021-04-15
Do not concatenate `Publisher.empty()` message body

idelpivnitskiy:
Motivation:

For `HeaderUtils#setContentLength(...)`, when the `messageBody` is exactly
`Publisher.empty()` there are no operators applied for the `messageBody`.
In this case we do not need concatenation.

Modifications:

- Skip concatenation when `messageBody == empty()`;
- Add tests when payloadBody is not set at all;

Result:

Less operators applied during request/response processing when not
necessary.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
@@ -177,7 +177,11 @@ final class HeaderUtils {
                                                       final BiIntConsumer<HttpHeaders> contentLengthUpdater) {
         if (messageBody == empty() || (isPayloadEmpty(metadata) && !mayHaveTrailers(metadata))) {
             contentLengthUpdater.apply(0, metadata.headers());
-            return from(metadata, EmptyHttpHeaders.INSTANCE).concat(messageBody.ignoreElements());
+            return messageBody == empty() ?
+                    from(metadata, EmptyHttpHeaders.INSTANCE) :
+                    // Subscribe to the messageBody publisher to trigger any applied transformations, but ignore its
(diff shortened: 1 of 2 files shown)


idelpivnitskiy:
@tkountis minor improvement, didn't think about it during review

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
```diff
@@ -177,7 +177,11 @@ public boolean mapTerminal() {
                                                       final BiIntConsumer<HttpHeaders> contentLengthUpdater) {
         if (messageBody == empty() || (isPayloadEmpty(metadata) && !mayHaveTrailers(metadata))) {
             contentLengthUpdater.apply(0, metadata.headers());
-            return from(metadata, EmptyHttpHeaders.INSTANCE).concat(messageBody.ignoreElements());
+            return messageBody == empty() ?
```
Scottmitch, line 180: nit: does it make sense to cache this?

```java
final boolean messageBodyEmpty = messageBody == empty();
if (messageBodyEmpty || ..) {
  return messageBodyEmpty ? ...;
}
```

Scottmitch (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
idelpivnitskiy: because it's a comparison of a reference address, not sure cache will be helpful

idelpivnitskiy: also hope that compiler is smart enough :)

PR #1498, opened 2021-04-16
release.sh: identify the remote before switching to `DRYRUN` mode

idelpivnitskiy:
Motivation:

Current version of the release script identifies the remote after the
`DRYRUN` mode overrides `git` command. As the result, logs do not show
which remote will be used when the `DRYRUN` mode will be off.

Modifications:

- Move lines that identify `remote` before the `DRYRUN` mode check;
- Print a message when `DRYRUN` mode is enabled;

Result:

`DRYRUN` mode correctly shows the git remote.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -49,17 +49,17 @@ fi
 echo "Releasing version $version"
 version_majorminor="${version%.*}"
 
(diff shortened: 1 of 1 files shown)


tkountis (approved):
🚀

PR #1512, opened 2021-04-22
Respect missing B3 sampled headers

tkountis:
Motivation:
Missing B3 sampled header (OpenTracing) from upstream translates to no sampling but according to the [spec](https://github.com/openzipkin/b3-propagation#sampling-state-1) absent should translate to receiver's choice.

Modifications:
Split class relation between span and trace-state and allow for trace state to have a Nullable sampled field, which in turn allows the sampler to take the right decision. The decision will propagate downstream.

Result:
Sampled field is now not required and sampler is allowed the make the right choice. That choice propagates downstream.

### DIFF
--- a/servicetalk-opentracing-http/src/main/java/io/servicetalk/opentracing/http/TracingHttpHeadersFormatter.java
+++ b/servicetalk-opentracing-http/src/main/java/io/servicetalk/opentracing/http/TracingHttpHeadersFormatter.java
@@ -29,6 +29,7 @@ import javax.annotation.Nullable;
 import static io.servicetalk.buffer.api.CharSequences.contentEqualsIgnoreCase;
 import static io.servicetalk.buffer.api.CharSequences.newAsciiString;
 import static io.servicetalk.opentracing.internal.HexUtils.validateHexBytes;
+import static java.lang.Boolean.TRUE;
 import static java.lang.String.valueOf;
 
 final class TracingHttpHeadersFormatter implements InMemoryTraceStateFormat<HttpHeaders> {
@@ -65,7 +66,9 @@ final class TracingHttpHeadersFormatter implements InMemoryTraceStateFormat<Http
         if (parentSpanIdHex != null) {
             carrier.set(PARENT_SPAN_ID, parentSpanIdHex);
         }
-        carrier.set(SAMPLED, state.isSampled() ? "1" : "0");
+        if (state.isSampled() != null) {
+            carrier.set(SAMPLED, TRUE.equals(state.isSampled()) ? "1" : "0");
+        }
     }
 
     @Nullable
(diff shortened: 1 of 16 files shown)


servicetalk-opentracing-http/src/main/java/io/servicetalk/opentracing/http/TracingHttpHeadersFormatter.java
```diff
@@ -58,14 +60,15 @@ private TracingHttpHeadersFormatter(boolean verifyExtractedValues) {
...
+    public void inject(final InMemorySpanContext context, final HttpHeaders carrier) {
+        final InMemoryTraceState state = context.traceState();
         carrier.set(TRACE_ID, state.traceIdHex());
         carrier.set(SPAN_ID, state.spanIdHex());
         String parentSpanIdHex = state.parentSpanIdHex();
         if (parentSpanIdHex != null) {
             carrier.set(PARENT_SPAN_ID, parentSpanIdHex);
         }
-        carrier.set(SAMPLED, state.isSampled() ? "1" : "0");
+        carrier.set(SAMPLED, context.isSampled() ? "1" : "0");
```
tkountis, line 71: Forwarding the local decision downstream

servicetalk-opentracing-inmemory-api/src/main/java/io/servicetalk/opentracing/inmemory/api/InMemorySpan.java
```diff
@@ -26,10 +26,35 @@
...
-public interface InMemorySpan extends Span, InMemoryTraceState {
+public interface InMemorySpan extends Span {
     @Override
     InMemorySpanContext context();
 
+    /**
+     * The hex representation of the traceId.
+     * @return hex representation of the traceId.
+     */
+    String traceIdHex();
```
Scottmitch, line 37: if we are going to break APIs wdyt about keeping these methods in a single location:

- `*IdHex` -> `context().traceState()` (already on the `InMemoryTraceState` interface).
- `*isSampled` -> `context.isSampled()` (already on the `InMemorySpanContext` interface).

servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/DefaultInMemorySpanContext.java
```diff
@@ -37,7 +37,7 @@
     final boolean isSampledOverride;
 
     DefaultInMemorySpanContext(InMemoryTraceState state) {
-        this(state, state.isSampled());
+        this(state, state.isSampled() != null && requireNonNull(state.isSampled()));
```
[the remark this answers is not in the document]
tkountis: leftover, thx

servicetalk-opentracing-inmemory-api/src/main/java/io/servicetalk/opentracing/inmemory/api/InMemorySpan.java
tkountis: hum, sounds good. I kinda liked the shortcut, but its have no other meaning, so yes getting rid of them.

Scottmitch (commented):
1 more API related comment then lgtm

Scottmitch (approved):
@tkountis - I should have clarified, the API I had a question about isn't impacted by this PR so feel free to merge. however while we are here can you tackle this in a followup PR?

PR #1513, opened 2021-04-22
fix: set span.kind for producer/consumer spans

eddie4941:
Motivation:

The zipkin publisher was not setting the span.kind for producer/consumer
spans. This causes some issues using open telemetry collector which
requires the top level span.kind field to be set. Zipkin doesn't have
this issue as it seems it falls back to reading from the tags to find
the kind. To ensure maximum compatibility, lets set the span.kind top
level field.

Modifications:

- set the span.kind top level field for missing span.kinds
(producer/consumer)

Result:

All spans sent via zipkin publisher will have top level span
.kind field set for all span.kinds.


servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/ZipkinPublisher.java
```diff
@@ -178,6 +178,12 @@ public void onSpanFinished(final InMemorySpan span, long durationMicros) {
             builder.kind(Span.Kind.SERVER);
         } else if (Tags.SPAN_KIND_CLIENT.equals(type)) {
             builder.kind(Span.Kind.CLIENT);
+        } else if (Tags.SPAN_KIND_PRODUCER.equals(type)) {
+            builder.kind(Span.Kind.PRODUCER);
+        } else if (Tags.SPAN_KIND_CONSUMER.equals(type)) {
+            builder.kind(Span.Kind.CONSUMER);
+        } else {
+            LOGGER.warn("Found unsupported kind '{}' in span tags. Not setting span.kind", type);
```
tkountis, line 186: Lets skip the warning here. Span kind is not a required field so no-value is also valid.

tkountis: Please consider updating the copyright of the file 2018 -> 2018, 2021

tkountis:
Thanks for the contribution @eddie4941!!
Minor comment, the change looks good to me otherwise.

tkountis (approved):

tkountis:
Once again thanks for the contribution @eddie4941 🎉 🎉

PR #1514, opened 2021-04-23
RoundRobinLoadBalancer reduce copy/resize for connection add/remove

Scottmitch:
Motivation:
RoundRobinLoadBalancer's Host class maintains a list of active
connections in a copy-on-write fashion. However when an item is added or
removed their is an additional copy operation due to initial state not
accounting for 1 more/less element.

Modifications:
- Host class uses an array for copy-on-write storage and allocates
  exactly the required amount of space on add/remove.

Result:
Less resize/copy operations on connection established/closed.

### DIFF
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018-2020 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2018-2021 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -53,8 +53,9 @@ import static io.servicetalk.client.api.LoadBalancerReadyEvent.LOAD_BALANCER_NOT
 import static io.servicetalk.client.api.LoadBalancerReadyEvent.LOAD_BALANCER_READY_EVENT;
 import static io.servicetalk.concurrent.api.AsyncCloseables.newCompositeCloseable;
 import static io.servicetalk.concurrent.api.AsyncCloseables.toAsyncCloseable;
-import static io.servicetalk.concurrent.api.Completable.mergeAllDelayError;
+import static io.servicetalk.concurrent.api.Completable.completed;
 import static io.servicetalk.concurrent.api.Processors.newPublisherProcessorDropHeadOnOverflow;
+import static io.servicetalk.concurrent.api.Publisher.from;
 import static io.servicetalk.concurrent.api.Single.defer;
 import static io.servicetalk.concurrent.api.Single.failed;
 import static io.servicetalk.concurrent.api.Single.succeeded;
@@ -336,7 +337,7 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
         return activeHosts.stream().map(Host::asEntry).collect(toList());
     }
 
-    private static class Host<Addr, C extends ListenableAsyncCloseable> implements AsyncCloseable {
+    private static final class Host<Addr, C extends ListenableAsyncCloseable> implements AsyncCloseable {
         @SuppressWarnings("rawtypes")
         private static final AtomicReferenceFieldUpdater<Host, Object[]> connectionsUpdater =
                 AtomicReferenceFieldUpdater.newUpdater(Host.class, Object[].class, "connections");
@@ -421,9 +422,11 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
 
         @SuppressWarnings("unchecked")
         private Completable doClose(final Function<? super C, Completable> closeFunction) {
-            return defer(() -> succeeded(connectionsUpdater.getAndSet(this, CLOSED_ARRAY)))
-                    .flatMapCompletable(array -> mergeAllDelayError(
-                            Stream.of(array).map(conn -> closeFunction.apply((C) conn))::iterator));
+            return Completable.defer(() -> {
+                final Object[] connections = connectionsUpdater.getAndSet(this, CLOSED_ARRAY);
+                return connections == CLOSED_ARRAY ? completed() :
+                        from(connections).flatMapCompletableDelayError(conn -> closeFunction.apply((C) conn));
+            });
         }
 
         @Override
@@ -436,6 +439,8 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
     }
 
     private static final class StacklessNoAvailableHostException extends NoAvailableHostException {
+        private static final long serialVersionUID = 5942960040738091793L;
+
         private StacklessNoAvailableHostException(final String message) {
             super(message);
         }
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -39,6 +39,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.concurrent.ThreadLocalRandom;
@@ -46,6 +47,7 @@ import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 import static io.servicetalk.client.api.LoadBalancerReadyEvent.LOAD_BALANCER_NOT_READY_EVENT;
 import static io.servicetalk.client.api.LoadBalancerReadyEvent.LOAD_BALANCER_READY_EVENT;
@@ -86,6 +88,8 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
 
     private static final Logger LOGGER = LoggerFactory.getLogger(RoundRobinLoadBalancer.class);
     private static final List<?> CLOSED_LIST = new ArrayList<>(0);
+    private static final Object[] CLOSED_ARRAY = new Object[0];
+    private static final Object[] EMPTY_ARRAY = new Object[0];
 
     @SuppressWarnings("rawtypes")
     private static final AtomicReferenceFieldUpdater<RoundRobinLoadBalancer, List> activeHostsUpdater =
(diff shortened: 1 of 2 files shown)


servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -403,15 +439,16 @@ public Completable closeAsyncGracefully() {
 
         @SuppressWarnings("unchecked")
         private Completable doClose(final Function<? super C, Completable> closeFunction) {
-            return defer(() -> succeeded((List<C>) connectionsUpdater.getAndSet(this, CLOSED_LIST)))
-                    .flatMapCompletable(list -> mergeAllDelayError(list.stream().map(closeFunction)::iterator));
+            return defer(() -> succeeded(connectionsUpdater.getAndSet(this, CLOSED_ARRAY)))
+                    .flatMapCompletable(array -> mergeAllDelayError(
+                            Stream.of(array).map(conn -> closeFunction.apply((C) conn))::iterator));
```
idelpivnitskiy, line 444: Consider:
```java
return defer(() -> {
    final Object[] connections = connectionsUpdater.getAndSet(this, CLOSED_ARRAY);
    if (connections == CLOSED_ARRAY) {
        return completed();
    }
    return Publisher.from(connections).flatMapCompletableDelayError(conn -> closeFunction.apply((C) conn));
});
```

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -333,62 +356,75 @@ public Completable closeAsyncGracefully() {
...
                 }
             }).subscribe();
             return true;
         }
 
         // Used for testing only
+        @SuppressWarnings("unchecked")
         Entry<Addr, List<C>> asEntry() {
-            return new SimpleImmutableEntry<>(address, new ArrayList<>(connections));
+            return new SimpleImmutableEntry<>(address, Stream.of(connections).map(conn -> (C) conn).collect(toList()));
```
idelpivnitskiy, line 427: Consider:
```java
return new SimpleImmutableEntry<>(address, Arrays.asList((C[]) connections));
```

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -86,6 +88,9 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(RoundRobinLoadBalancer.class);
     private static final List<?> CLOSED_LIST = new ArrayList<>(0);
+    private static final Object[] CLOSED_ARRAY = new Object[0];
```
idelpivnitskiy, line 91: Consider updating copyright year for this class

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -260,15 +248,50 @@ public void onComplete() {
...
+                        final C connection = (C) connections[index];
+                        if (selector.test(connection)) {
+                            return succeeded(connection);
+                        }
+                    } while (++i < connections.length);
+                    break;
+                }
+            }
+        } else {
+            for (int i = 0; i < LARGE_LIST_SELECTION_ATTEMPTS; i++) {
```
idelpivnitskiy, line 289: Can we keep the previous logic in this case: `(int) (connections.length * 0.75f)`? 64 attempts looks low if users have 1000s of connections

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -260,15 +248,50 @@ public void onComplete() {
...
+        final Object[] connections = host.connections;
+        if (connections.length <= 64) {
+            // With small enough search space, attempt all connections.
+            // Back off after exploring most of the search space, it gives diminishing returns.
+            int i = 0;
+            long indexMask = 0;
+            int collisions = 0;
+            final int collisionThreshold = connections.length >>> 1;
+            while (i < connections.length) {
+                int index = rnd.nextInt(connections.length);
```
idelpivnitskiy, line 260: This algorithm is a bit involved and may have collisions. In the worst case it iterates over all connections. Consider taking a random value only for the first test and then continue in a loop over all connections.

idelpivnitskiy (commented):
Nice improvements!

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -403,15 +439,16 @@ public Completable closeAsyncGracefully() {
...
+                    .flatMapCompletable(array -> mergeAllDelayError(
+                            Stream.of(array).map(conn -> closeFunction.apply((C) conn))::iterator));
         }
 
         @Override
         public String toString() {
             return "Host{" +
                     "address=" + address +
-                    ", removed=" + (connections == CLOSED_LIST) +
+                    ", removed=" + (connections == CLOSED_ARRAY) +
```
idelpivnitskiy, line 451: While you are here, please add `serialVersionUID` for `StacklessNoAvailableHostException`

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: casting to `C[]` will result in class cast exception.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: I will remove the selection algorithm part of this PR and move to another PR.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: done

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: done

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: I will experiment with this, and I'll move the selection changes to another PR.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -333,62 +356,75 @@ public Completable closeAsyncGracefully() {
...
+                        if (existing[i].equals(connection)) {
+                            break;
+                        }
+                    }
+                    if (i == existing.length) {
                         break;
+                    } else {
+                        Object[] newList = new Object[existing.length - 1];
+                        System.arraycopy(existing, 0, newList, 0, i);
+                        System.arraycopy(existing, i + 1, newList, i, newList.length - i);
```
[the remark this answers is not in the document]
idelpivnitskiy: WDYT about this one?

idelpivnitskiy (approved):

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -331,64 +337,77 @@ public Completable closeAsyncGracefully() {
...
+    private static final class Host<Addr, C extends ListenableAsyncCloseable> implements AsyncCloseable {
         @SuppressWarnings("rawtypes")
-        private static final AtomicReferenceFieldUpdater<Host, List> connectionsUpdater =
-                AtomicReferenceFieldUpdater.newUpdater(Host.class, List.class, "connections");
+        private static final AtomicReferenceFieldUpdater<Host, Object[]> connectionsUpdater =
+                AtomicReferenceFieldUpdater.newUpdater(Host.class, Object[].class, "connections");
 
         final Addr address;
-        private volatile List<C> connections = emptyList();
+        private volatile Object[] connections = EMPTY_ARRAY;
```
bondolo, line 346: This could be ListenableAsyncCloseable

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -331,64 +337,77 @@ public Completable closeAsyncGracefully() {
...
                 }
-                ArrayList<C> connectionAdded = new ArrayList<>(existing);
-                connectionAdded.add(connection);
-                if (connectionsUpdater.compareAndSet(this, existing, connectionAdded)) {
+                Object[] newList = Arrays.copyOf(existing, existing.length + 1);
+                newList[existing.length] = connection;
+                if (connectionsUpdater.compareAndSet(this, existing, newList)) {
                     break;
                 }
             }
```
bondolo, line 377: do {} while(cU.compareAndSet()) would avoid the break

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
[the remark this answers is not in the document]
Scottmitch: I'm not sure it is worth it bcz `arraycopy` already does bounds checking. we could also specialize other cases (e.g. `i == existing.length - 1`) but lets save for followup investigation.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -331,64 +337,77 @@ public Completable closeAsyncGracefully() {
...
+                    } else {
+                        Object[] newList = new Object[existing.length - 1];
+                        System.arraycopy(existing, 0, newList, 0, i);
+                        System.arraycopy(existing, i + 1, newList, i, newList.length - i);
+                        if (connectionsUpdater.compareAndSet(this, existing, newList)) {
+                            break;
+                        }
                     }
                 }
             }).subscribe();
```
bondolo, line 403: Perhaps make this a standalone removeConnectionMethod() and call that in beforeFinally?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -331,64 +337,77 @@ public Completable closeAsyncGracefully() {
...
 
             // Instrument the new connection so we prune it on close
             connection.onClose().beforeFinally(() -> {
                 for (;;) {
-                    final List<C> existing = connections;
-                    if (existing == CLOSED_LIST) {
+                    final Object[] existing = this.connections;
+                    if (existing == CLOSED_ARRAY) {
                         break;
                     }
```
bondolo, line 385: Not really necessary as CLOSED_ARRAY is always empty.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
bondolo, line 346: How big is connections likely to be and how much contention is there likely to be? Could perhaps just use or extend CopyOnWriteArraySet

idelpivnitskiy: Its internal `CopyOnWriteArrayList` uses `ReentrantLock` that we prefer to avoid as it may content on the event-loop threads.

PR #1515, opened 2021-04-23
Add Single#zipDelayError operator

Scottmitch:
Motivation:
We have Single#zip but not the zipDelayError variant.

Modifications:
- Add Single#zipDelayError operator to match Single#zip use cases

Result:
Single#zipDelayError operator exists.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
@@ -899,6 +899,31 @@ public abstract class Single<T> {
     }
 
     /**
+     * Create a new {@link Single} that emits the results of a specified zipper {@link BiFunction} to items emitted by
+     * {@code singles}. All operations will terminate (even if there are failures) before the returned {@link Single}
+     * terminates.
+     * <p>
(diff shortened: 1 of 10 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -898,6 +898,31 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
         return zip(this, other, zipper);
     }
 
+    /**
+     * Create a new {@link Single} that emits the results of a specified zipper {@link BiFunction} to items emitted by
+     * {@code singles}. All operations will terminate (even if there are failures) before the returned {@link Single}
+     * terminates.
+     * <p>
```
bondolo, line 905: Perhaps should say more about what the error terminations look like. What does `onError` receive if both `onError`? Either `onError`? Is the item lost when the other Single errors?

Scottmitch: I updated the javadoc. ptal.

bondolo (approved):

idelpivnitskiy (commented):
Late LGTM!

PR #1518, opened 2021-04-25
Optimize encoding of HTTP/2 frames to write less on the wire

idelpivnitskiy:
Motivation:

We can infer from the `PayloadInfo` when HTTP message does not contain
payload body or trailers. In this case, we can write it in a single frame.
Also, we do not need to write `Http2DataFrame`(s) for empty `Buffer`(s)
or process `HttpHeaders` when it's empty.

Modifications:

- Detect if the `HttpMetaData` can be written in a single frame;
- Emit `CloseHandler#protocolPayloadEndOutbound` on `endStream`;
- Ignore empty `Buffer`(s) when writing `Http2DataFrame`(s);
- Do not process `HttpHeaders` when they are empty;
- Add tests for new behavior;

Result:

Less `Http2StreamFrame`(s) allocated and written on the wire, results in
+10% RPS, -12% latency for some HTTP/2 use-cases.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
@@ -33,6 +33,8 @@ import io.netty.handler.codec.http2.Http2DataFrame;
 import io.netty.handler.codec.http2.Http2Headers;
 import io.netty.handler.codec.http2.Http2ResetFrame;
 import io.netty.util.ReferenceCountUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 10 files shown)


tkountis (approved):
Consider updating the copyrights. 
🚀 nice perf figures.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
```diff
@@ -65,18 +73,51 @@ public final void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
         }
     }
 
+    final void writeMetaData(ChannelHandlerContext ctx, HttpMetaData metaData, Http2Headers h2Headers,
+                             ChannelPromise promise) {
+        endStream = !mayHaveTrailers(metaData) && isPayloadEmpty(metaData);
+        if (endStream) {
+            closeHandler.protocolPayloadEndOutbound(ctx, promise);
```
idelpivnitskiy, line 80: @Scottmitch `HttpTransportObserverTest.serverHandlerError` helped to discover a race between netty closing the `StreamChannel` (emits `channelInactive`, `channelUnregistered`, and removes the handler) and receiving trailers from the offloaded thread.

So, I made this change: in case of `endStream`, I trigger this callback here and do nothing in `writeTrailers`. IIUC the control flow of `WriteStreamSubscriber`, it will ignore any following writes (see `doWrite`) if the source is already terminated. 
Do you see any issues with existing approach?

idelpivnitskiy: Discussed with @Scottmitch offline. `WriteStreamSubscriber` will correctly account for the current state and no further writes or terminal signals will be propagated. All good.

PR #1522, opened 2021-05-03
Use `UniformStreamByteDistributor` for HTTP/2 connections

idelpivnitskiy:
Motivation:

HTTP/2 spec defines priorities for streams. However, in ST we do not
expose API to assign priorities for streams and this feature is rarely
used in general. `Http2FrameCodecBuilder` creates connections with
`WeightedFairQueueByteDistributor` by default that takes priorities into
account. We can use `UniformStreamByteDistributor` to reduce the
overhead of managing HTTP/2 streams.

Modifications:

- Extend `Http2FrameCodecBuilder` to use `UniformStreamByteDistributor`
for new `Http2Connection`(s);

Result:

Less CPU usage for HTTP/2 ~1-1.5% according to the benchmarks.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,7 +17,7 @@
 group=io.servicetalk
 version=0.40.0-SNAPSHOT
 
-nettyVersion=4.1.63.Final
+nettyVersion=4.1.64.Final-SNAPSHOT
 tcnativeVersion=2.0.38.Final
 
 jsr305Version=3.0.2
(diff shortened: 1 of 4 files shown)


gradle.properties
```diff
@@ -17,7 +17,7 @@
 group=io.servicetalk
 version=0.40.0-SNAPSHOT
 
-nettyVersion=4.1.63.Final
+nettyVersion=4.1.64.Final-SNAPSHOT
```
idelpivnitskiy, line 20: Reminder for myself: update netty version when it's available.

idelpivnitskiy:
Used the wrong branch, reopened in #1523.

PR #1524, opened 2021-05-03
Config test ignoreFailures via system property

bondolo:
Motivation:
The default behaviour of Gradle is to stop the build with an error in
the first module with a test failure. In some cases it is useful to
run all tests to get a survey of which tests are failing. It can also
be annoying if flaky tests are stopping the build from completing.

Modifications:
This change allows configuration of the Gradle test `ignoreFailures`
option to continue the build after test failures. The `ignoreFailures`
test configuration is initialized from the system property 
`servicetalk.test.ignoreFailures` allowing for command line
configuration. `./gradlew -Dservicetalk.test.ignoreFailures=true build`

Result:
Dynamically configurable ignoreFailures. This feature should not be
used for CI builds as it will mask build failures. Build results
should always be manually examined for test failures if this option
is used.


idelpivnitskiy (approved):
Useful for development, thanks!

PR #1527, opened 2021-05-05
Deprecate publishAndSubscribeOnOverride methods

bondolo:
**Motivation**:
Refactoring of offloading is expected to remove these methods so it is
important to give users forewarning to migrate to other methods.

**Modifications**:
The `publishOnOverride`, `subscribeOnOverride` and
`publishAndSubscribeOnOverride` methods of `Publisher`, `Single` and
`Completable` are deprecated as well as the internal methods they call
are marked deprecated and an alternative suggestion is provided.

**Result**:
Users are forewarned of upcoming API change.


idelpivnitskiy (approved):
🚀

PR #1528, opened 2021-05-06
Support wrapping read-only `ByteBuffer`(s) by `ServiceTalkBufferAllocator`

chemicL:
Motivation:

When user provides a read-only ByteBuffer, the
ServiceTalkBufferAllocator should be able to wrap it without issues.
Currently it fails.

Modifications:

- Fork of `io.netty.buffer.ReadOnlyByteBufferBuf` from Netty source as
`UnreleasableReadOnlyByteBufferBuf` which disables reference counting;
- Support in `ServiceTalkBufferAllocator.wrap method` for read-only
`ByteBuffer`;

Result:

Read-only direct buffers can be wrapped via `ServiceTalkBufferAllocator`
without issue.


servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ReadOnlyByteBufferBuf.java
```diff
@@ -0,0 +1,508 @@
+/*
+ * Copyright © 2018-2021 Apple Inc. and the ServiceTalk project authors
```
idelpivnitskiy, line 2: For new files, copyright year should be current

servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ReadOnlyByteBufferBuf.java
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    @Override
+    public ByteBuf unwrap() {
+        return null;
+    }
+
+    @Override
+    public boolean isReadOnly() {
+        return buffer.isReadOnly();
```
idelpivnitskiy, line 351: The following comments for this class will ask you to change netty's impl. However, I think those changes are correct and worth contributing them back to netty:

This can always return `true` because we apply read-only enforcement at current level

servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ReadOnlyByteBufferBuf.java
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    @Override
+    public ByteBuf getBytes(int index, OutputStream out, int length) throws IOException {
+        ensureAccessible();
+        if (length == 0) {
+            return this;
+        }
+
+        if (buffer.hasArray()) {
```
idelpivnitskiy, line 366: `ByteBuffer#hasArray()` always return `false` for read-only implementations

servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ReadOnlyByteBufferBuf.java
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    @Override
+    public final boolean isContiguous() {
+        return true;
+    }
+
+    @Override
+    public boolean hasArray() {
+        return buffer.hasArray();
```
idelpivnitskiy, line 486: We can return false from here bcz any read-only impl always returns false

servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ReadOnlyByteBufferBuf.java
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    @Override
+    public byte[] array() {
+        return buffer.array();
```
idelpivnitskiy, line 491: We should do:
```
throw new ReadOnlyBufferException();
```
in this case

servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ReadOnlyByteBufferBuf.java
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    @Override
+    public int arrayOffset() {
+        return buffer.arrayOffset();
```
idelpivnitskiy, line 496: We should do:
```
throw new ReadOnlyBufferException();
```
in this case

servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/UnreleasableReadOnlyByteBufferBuf.java
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+package io.servicetalk.buffer.netty;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
+
+import java.nio.ByteBuffer;
+
+public class UnreleasableReadOnlyByteBufferBuf extends ReadOnlyByteBufferBuf {
```
idelpivnitskiy, line 23: I understood your motivation to keep the copied class as close as possible to the original version from netty, but I don't think it's necessary and we do not really follow this pattern. Consider using only one class - `UnreleasableReadOnlyByteBufferBuf` (by merging it together with `ReadOnlyByteBufferBuf`) and making it `final` + pkg-private.

idelpivnitskiy (changes requested):
Congrats on the first PR!

servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/UnreleasableReadOnlyByteBufferBuf.java
chemicL: Thanks for the clarification. I was hesitant for a while to keep the separation of concerns, but in our case it actually makes more sense to merge these classes the way you describe.

servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ReadOnlyByteBufferBuf.java
chemicL: Perfect, that makes sense.

normanmaurer: sounds correct

PR #1529, opened 2021-05-06
Unconditionally add timeout filter to gRPC server

bondolo:
Motivation:
If there is no default timeout set then a client specified timeout will
currently be ignored because the filter is not present.
Modifications:
Unconditionally add the timeout filter in the gRPC server builder.
Result:
Timeout filter is present in all connections.


idelpivnitskiy:
What about the server-side? It should also have a filter in case an incoming request enforces timeout, isn't it?

idelpivnitskiy:
🤦 I thought you added that to the client-builder only.
Ok, then reversed question: should we do the same for the client-side? Bcz deadline is exposed on the `ClientMetaData` or the value can be in async-context. But if users did not configure the default, non of the previous two will work.

bondolo:
Looking at  [`DefaultGrpcClientCallFactory::timeoutForRequest`](https://github.com/apple/servicetalk/blob/f5ed264c1ff65405db04d09b36915450d4edba27/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java#L250-L261) the default is only used if neither request metadata or async context timeout is present. Were you looking at some other portion of the code which gave you a different impression?

idelpivnitskiy:
I'm trying to say that if the `GrpcClientBuilder#defaultTimeout` is not set => the `TimeoutHttpRequesterFilter` won't be added => neither `GrpcClientMetadata#timeout()` nor value from `AsyncContext` will work. This will be surprising for users if they use `GrpcClientMetadata` to set the value or expect that server propagates deadline to the clients it uses. Hard to understand why it's not working if you do not know about `GrpcClientBuilder#defaultTimeout`.

bondolo:
I have updated the PR to unconditionally install the timeout filter for gRPC client and server.

idelpivnitskiy (approved):

Issue #1530, opened 2021-05-06
Flaky test: SecureTcpTransportObserverErrorsTest#testSslErrors


idelpivnitskiy:
https://github.com/apple/servicetalk/runs/2521994553

```
SecureTcpTransportObserverErrorsTest > testSslErrors[errorReason=NOT_MATCHING_PROTOCOLS, clientProvider=JDK, serverProvider=JDK] FAILED
    org.junit.runners.model.TestTimedOutException: test timed out after 90000 milliseconds
        at java.lang.Thread.sleep(Native Method)
        at org.mockito.internal.verification.VerificationOverTimeImpl.sleep(VerificationOverTimeImpl.java:117)
        at org.mockito.internal.verification.VerificationOverTimeImpl.handleVerifyException(VerificationOverTimeImpl.java:100)
        at org.mockito.internal.verification.VerificationOverTimeImpl.verify(VerificationOverTimeImpl.java:89)
        at org.mockito.internal.verification.VerificationWrapper.verify(VerificationWrapper.java:19)
        at org.mockito.internal.verification.MockAwareVerificationMode.verify(MockAwareVerificationMode.java:27)
        at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:72)
        at org.mockito.internal.handler.NullResultGuardian.handle(NullResultGuardian.java:29)
        at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:35)
        at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:61)
        at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:49)
        at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor$DispatcherDefaultingToRealMethod.interceptAbstract(MockMethodInterceptor.java:126)
        at io.servicetalk.transport.api.ConnectionObserver$SecurityHandshakeObserver$MockitoMock$664195828.handshakeFailed(Unknown Source)
        at io.servicetalk.tcp.netty.internal.SecureTcpTransportObserverErrorsTest.testSslErrors(SecureTcpTransportObserverErrorsTest.java:206)

SecureTcpTransportObserverErrorsTest > testSslErrors[errorReason=NOT_MATCHING_PROTOCOLS, clientProvider=OPENSSL, serverProvider=JDK] FAILED
[...]

PR #1532, opened 2021-05-07
Debug #1530

idelpivnitskiy:


idelpivnitskiy:
The root cause was found, fixed in #1533.

PR #1539, opened 2021-05-11
Don't load netty-transport-native classes when native lib isn't available

idelpivnitskiy:
Motivation:

1. Do not try to load native libraries when it's known that OS does not
support it.
2. Log when OS supports native transport, but native lib cannot be
loaded.
3. Prevent class loading of any classes related to native transport when
it's not available.

Modifications:

- Enhance `NativeTransportUtils` to protect class loading when it's
known that OS does not support a specific type of transport;
- Log when native transport can not be loaded;

Result:

1. Less unused classes are loaded in memory.
2. Users are notified when native transport can be used but is not
loaded properly.


tkountis (approved):
LGTM
nit: consider replacing the wording "should be used" -> "could be used" in the `useEpoll` and `useKQueue` APIs.

normanmaurer (approved):

PR #1555, opened 2021-05-12
Jersey example with multipart file upload support

tkountis:
Motivation:

Include an example of a Jersey resource that supports `mulitpart/form-data` requests.

Modifications:

A new handler in the existing greetings resource.

Result:

More examples for users who want to support multipart file uploads.


ddossot (approved):
LG!

idelpivnitskiy (approved):
🚀

PR #1558, opened 2021-05-13
Disable auto ack for HTTP/2 PING frames by netty

idelpivnitskiy:
Motivation:

`KeepAliveManager#pingReceived` always acks PING frames. Netty's auto
ack produces additional PING-ack frames that are not necessary and
confusing.

Modifications:

- Use `Http2FrameCodecBuilder#autoAckPingFrame(false)` on client and
server sides;

Result:

No duplicated PING frames written with ack=true.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
@@ -48,6 +48,8 @@ final class H2ClientParentChannelInitializer implements ChannelInitializer {
                 // The max concurrent streams is made available via a publisher and may be consumed asynchronously
                 // (e.g. when offloading is enabled), so we manually control the SETTINGS ACK frames.
                 .autoAckSettingsFrame(false)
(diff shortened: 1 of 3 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
```diff
@@ -48,6 +48,8 @@ public void init(final Channel channel) {
                 // The max concurrent streams is made available via a publisher and may be consumed asynchronously
                 // (e.g. when offloading is enabled), so we manually control the SETTINGS ACK frames.
                 .autoAckSettingsFrame(false)
+                // We ack PING frames in KeepAliveManager#pingReceived.
+                .autoAckPingFrame(false)
```
idelpivnitskiy, line 52: All PING frames are handled here and always ack'ed:
https://github.com/apple/servicetalk/blob/6b4a2407b6952e0242b1129a728b8146c015efd8/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java#L142-L159

bondolo (approved):

PR #1559, opened 2021-05-13
Do not write GOAWAY frame if HTTP/2 connection is already closed

idelpivnitskiy:
Motivation:

`KeepAliveManager#gracefulCloseWriteSecondGoAway()` can be triggered on
PING-ack frame at the time when the connection is already closed,
because there were no active streams.

Modifications:

- Do not write GOAWAY frame if `gracefulCloseState == CLOSED`;

Result:

No failure on trying to write a GOAWAY frame after connection is already
closed.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
@@ -274,7 +274,7 @@ final class KeepAliveManager {
     private void gracefulCloseWriteSecondGoAway() {
         assert channel.eventLoop().inEventLoop();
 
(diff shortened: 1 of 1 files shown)


bondolo (approved):

idelpivnitskiy:
The actual reason for executing `gracefulCloseWriteSecondGoAway()` at the time when `gracefulCloseState == CLOSED` was duplicated PINK-ack frame, see #1558. After eliminating that issue, this never happens again. Even if the remote peer is malformed and sends two PING-ack frames instead of one, writing a `GO_AWAY` frame after the channel is closed does not cause any issues, because `Http2ConnectionHandler#goAway` checks if the frame should be written or not. The result in this case is just noop. We don't need this additional check.

PR #1561, opened 2021-05-13
Allow using `null` peer host for TLS and disabling host and port inference and SNI altogether.

chemicL:
Motivation:

There are cases when clients require connecting to a secure server which doesn't
identify itself in the TLS certificate with the same address as used for
connection establishment, or multiple identities can be presented from
diffrent underlying backends. In such case the hostname verification would fail
and would need to be disabled. However, the peerHost which is inferred from the
connection address would unnecessarily cache SSL sessions, which could lead
to issues in session resumption. Therefore, we want to allow explicitly disabling
peerHost inference and either allow setting it to a desired address or to null.

Modifications:

- Added ClientSslConfig#isPeerHostSet method,
- ClientSslConfigBuilder#setPeerHost now accepts null and disables address inference,
- Added test in SslAndNonSslConnectionsTest validating the mentioned scenario,
- Added test in SslAndNonSslConnectionsTest for ensuring hostname validation
  works as expected when hostname is used (existing tests used only IP).

Result:

It is now possible to explicitly use `null` for `peerHost`
and disable its' inference.


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ClientSslConfig.java
```diff
@@ -39,14 +39,23 @@
...
     @Nullable
     String peerHost();
 
+    /**
+     * Determine if the user explicitly set the {@link #peerHost()}.
+     *
+     * @return {@code true} the user explicitly set {@link #peerHost()} to some value (including {@code null}).
+     */
+    default boolean isPeerHostSet() {
+        return false;
```
Scottmitch, line 56: can you clarify rational for the default being `false` vs `peerHost() != null`?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ClientSslConfig.java
```diff
@@ -39,14 +39,23 @@
...
      */
     @Nullable
     String peerHost();
 
+    /**
+     * Determine if the user explicitly set the {@link #peerHost()}.
+     *
+     * @return {@code true} the user explicitly set {@link #peerHost()} to some value (including {@code null}).
+     */
+    default boolean isPeerHostSet() {
```
idelpivnitskiy, line 55: 1. `ClientSslConfig` is a relatively new API and we do not expect people implementing their own version. Not necessary to make it `default`.
2. Instead of introducing a new API method, can we consider using an empty string as a way to reset the peerHost?

@Scottmitch `SSLEngine` takes `peerHost` and `peerPort` together. Imho, our API should also keep them as one method that takes/returns `HostAndPort`. Otherwise, it's easier to make a mistake of setting one, but not another. Or at least have two variants:
```java
public ClientSslConfigBuilder peerHost(String);
public ClientSslConfigBuilder peerHostAndPort(HostAndPort);
```
What can be a use-case when users want to override the `peerPort`, but not the `peerHost`?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslAndNonSslConnectionsTest.java
```diff
@@ -185,6 +187,20 @@ public void singleAddressClientWithSslToSecureServer() throws Exception {
...
     }
 
+    @Test
+    public void singleAddressClientWithSslToSecureServerWithoutPeerHost() throws Exception {
+        assert secureServerCtx != null;
+        try (BlockingHttpClient client = HttpClients.forSingleAddress(serverHostAndPort(secureServerCtx))
+                .sslConfig(new ClientSslConfigBuilder(DefaultTestCerts::loadServerCAPem)
+                        .peerHost(null)
+                        // if verification is not disabled, identity check fails against the undefined address
+                        .hostnameVerificationAlgorithm("")
```
idelpivnitskiy, line 197: @Scottmitch if this approach also works we can consider removing `disableHostnameVerification()`. If we prefer `disableHostnameVerification()`, we should consider enforcing a non-empty verification algorithm. Having two ways to do the same thing is ambigious.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ClientSslConfig.java
chemicL: Great catch, @idelpivnitskiy. I think so far everything turned out fine for HTTP. The host AND port inference from target address would not have been disabled even when the user set `peerHost` to something of their desire. In that case the host would have changed, but the port would have been inferred. In the implementation I suggested, not setting the port would result in using `-1` even if multiple ports were used for connections, which would disable SSL caching of sessions ([check JDK implementation](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/sun/security/ssl/SSLSessionContextImpl.java#L178)).

Therefore, I agree with your suggestion for creating a method `peerHostAndPort(HostAndPort)`. This would make the behaviour easier to understand from the user's perspective.

Let's assume an empty string represents the situation when the user explicitly set `peerHost` to be empty.

I would imagine, that setting it to empty also disables port inference (effectively, as shown by JDK implementation, if peerHost is not set then the port is not used for caching either. For hostname verification, only the peerHost/sni is used.

The questions that arise are:
* if the user set `peerHost(String)` to non-empty but has not fiddled with `peerPort`, do we keep inferring the port?
* if the user explicitly wants to set `peerPort(int)` to some disabled value to disable the inference – what value would that be?

With existing API for `peerHost` the `String` type allows three independent value types: `null`, `""` and non-empty strings. For `int` it is different, we could potentially explain `-1` stands for "go ahead and infer", while `< -1` stands for "use -1 and don't infer the port". I feel the interface would be getting clunky at this point.

---

There is another way to approach the problem, as suggested by @Scottmitch – introducing new API methods in `SingleAddressHttpClientBuilder` to disable inference of `peerHost`/`peerPort`, and possibly disabling `sniHostname` inference and SNI use. However, I think that those settings are integral to `ClientSslConfig` regardless of the protocol (currently there's only HTTP, but potentially raw TCP/UDP could utilize those settings).

Having all that in mind, I'd suggest two options for changes to `ClientSslConfig`:

**(1)**

* Add method `Supplier<HostAndPort> peerHostAndPort()`: when null -> default behaviour = infer both from target address; when returned `Supplier.get()` returns null -> disable inference = use default empty values (`null`, `-1`); when not `null` -> use whatever the user provided,
* Add method `Supplier<String> getSniHostname()`: when null -> default behaviour = infer from target address; when returned `Supplier.get()` returns null -> disable SNI; when not `null` -> use for SNI Hostname,
* Deprecate/remove `String sniHostname()`,
* Deprecate/remove `String peerHost()`,
* Deprecate/remove `int peerPort()`.

**(2)**
* Leave `String sniHostname()`, `String peerHost()`, `int peerPort()` in their current form,
* Add methods `boolean isPeerHostSet()`, `boolean isPeerPortSet()` which allow determining whether the user set the values and wants to disable any default inference for any of them,
* `isSniEnabled()` -> returns `true` by default or when user provided a non- `null`/empty value. Returns `false` when user explicitly set `null`. The builder should disallow empty strings. As a result: `isSniEnabled() && sniHostname() == null` results in address inference for SNI, `isSniEnabled() && sniHostname() != null) results in using the provided hostname and disabling SNI for `!isSniEnabled()`.

Please let me know whether any of those options work and I'll get back to the implementation.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslAndNonSslConnectionsTest.java
chemicL: I think the intention was to block users from disabling the verification, but I believe there are cases where it makes sense. So agree with @idelpivnitskiy, best to leave just one way to disable verification.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ClientSslConfig.java
Scottmitch: It seems like the `ClientSslConfig` API provides what it needs to, and the confusion is introduced at the HTTP layer which attempts to infer and apply overrides under some circumstances. If we allow the user to disable the http infer/override and just use the `ClientSslConfig` verbatim, would that address the original issue? This would avoid over complicating the `ClientSslConfig` API.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslAndNonSslConnectionsTest.java
Scottmitch: my vote: remove `disableHostnameVerification` update javadocs for `hostnameVerificationAlgorithm`

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ClientSslConfig.java
chemicL: Agreed, the implementation will be simpler. My only worry is that encapsulation of SSL configuration is lost and the user needs to interact with two builders:

- `ClientSslConfigBuilder` to set the config values,
- and `SingleAddressHttpClientBuilder` to tell the `HttpClient` what to do with the provided values.

Sample use that makes the client use SNI which is inferred, but does not verify the certificates and disables session caching:

```
        HttpClients.forSingleAddress(serverHostAndPort(serverCtx))
                .sslConfig(
                        new ClientSslConfigBuilder(DefaultTestCerts::loadServerCAPem)
                                .hostnameVerificationAlgorithm("null")
                                .build()
                )
                .disablePeerHostInference()
                .buildBlocking())
```

I think this depicts a reasonable use case. If that is fine by both of you, @Scottmitch and @idelpivnitskiy, I'll go ahead and make changes.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
```diff
@@ -144,4 +144,26 @@
...
+     */
+    public abstract SingleAddressHttpClientBuilder<U, R> disablePeerPortInference();
+
+    /**
+     * Effectively disables <a href="https://datatracker.ietf.org/doc/html/rfc6066#section-3">SNI</a>
+     * and possible SNI hostname inference from client's address if not explicitly specified
+     * via {@link #sslConfig(ClientSslConfig)}.
+     * @return {@code this}
+     */
+    public abstract SingleAddressHttpClientBuilder<U, R> disableSni();
```
Scottmitch, line 168: `disableSni` -> `disableSniInference`? Can you clarify why this method is different than the other new methods?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ClientSecurityConfig.java
```diff
@@ -256,7 +256,7 @@ public ClientSslConfig asSslConfig() {
         }
 
         if (hostnameVerificationAlgorithm == null) {
-            builder.disableHostnameVerification();
+            builder.hostnameVerificationAlgorithm("");
```
idelpivnitskiy, line 259: Can you please add at the end of `@deprecated` description of `ClientSslConfigBuilder#disableHostnameVerification()` that if users are sure they need to disable it, they can use `hostnameVerificationAlgorithm("")`.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
idelpivnitskiy: Consider the following names to allow users set/unset the value on the builder:
```
peerHostInference(boolean)
peerPortInference(boolean)
sniHostnameInference(boolean)
```
or
```
inferPeerHost(boolean)
inferPeerPort(boolean)
inferSniHostname(boolean)
```

chemicL: @Scottmitch my reasoning was the following: the inference for SNI is less straightforward than for peerHost – internally, the `X509TrustManagerImpl` does a fallback to peerHost when checking the certificate identity. The internals also [explicitly add](https://github.com/openjdk/jdk8u/blob/4a4236a366eeb961baf157f0938634c1647c447f/jdk/src/share/classes/sun/security/ssl/SSLSocketImpl.java#L1260-L1265) the peerHost to the SNI list.
The existing fallback may seem confusing, hence I decided disabling SNI is the functionality that might be useful. Yet still, my attempt at allowing disabling is not really effective because of the fallback in JDK and Netty. I'll change to the inference related method.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ClientSecurityConfig.java
chemicL: Done.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClientConfig.java
```diff
@@ -119,6 +137,20 @@ public int peerPort() {
                 public String sniHostname() {
                     return sniHostname;
                 }
+
+                @Nullable
+                private String evaluatePeerHost() {
+                    return (configPeerHost == null && inferPeerHost) ? fallbackPeerHost : configPeerHost;
```
Scottmitch, line 143: consider inlining these definitions or making the anonymous type non-anonymous and adding constructor to clarify we don't need to retain references to `configPeerHost` and related variables outside the scope of this class.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SniTest.java
```diff
@@ -97,9 +100,30 @@ void noSniClientDefaultServerFallbackSuccess() throws Exception {
...
+    @Test
+    void noSniClientDefaultServerFallbackFailExpected() throws Exception {
+        try (ServerContext serverContext = HttpServers.forAddress(localAddress(0))
+                .sslConfig(
+                        untrustedServerConfig(),
+                        singletonMap(InetAddress.getLoopbackAddress().getHostName(), trustedServerConfig())
+                )
+                .listenBlockingAndAwait((ctx, request, responseFactory) -> responseFactory.ok());
+             BlockingHttpClient client = HttpClients.forSingleAddress(
+                     HostAndPort.of(
```
Scottmitch, line 116: nit: consider removing `HostAndPort.of(` and use the method overload that takes `String, int`

Scottmitch (commented):
one comment then lgtm

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
```diff
@@ -144,4 +144,29 @@
...
      */
     public abstract SingleAddressHttpClientBuilder<U, R> sslConfig(ClientSslConfig sslConfig);
+
+    /**
+     * Toggle inference of value to use instead of {@link ClientSslConfig#peerHost()}
+     * from client's address when peer host is not specified. By default, inference is enabled.
+     * @param shouldInfer value indicating whether inference is on ({@code true}) or off ({@code false}).
+     * @return {@code this}
+     */
+    public abstract SingleAddressHttpClientBuilder<U, R> inferPeerHost(boolean shouldInfer);
```
idelpivnitskiy, line 154: We previously discussed in https://github.com/apple/servicetalk/pull/1561#discussion_r632269087, but doesn't look like we come to a conclusion. Does it make sense to join peer host and port together on `ClientSslConfig` API into a single method `HostAndPort peerHostAndPort()`? If yes, we can have a single method here: `inferPeerHostAndPort(boolean)` instead of two.

idelpivnitskiy (commented):
Good work!

Doesn't look like this PR breaks any API now. Should we remove `breaking-change` label now?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
idelpivnitskiy: @Scottmitch wdyt about single `ClientSslConfig#peerHostAndPort()` and inferPeerHostAndPort(boolean)?

chemicL: To my understanding the peer port is not used when peer host is not inferred. I think it's reasonable to provide an API where both are affected. And yes, the breaking API change would need to go to `ClientSslConfig`. Perhaps we could do it in a separate PR and for now add this feature with no breakage?

idelpivnitskiy: I was thinking that if we merge host and port into one method, we can provide only one "infer" method and won't need to change client API later.
Let's make the final decision tomorrow morning. Everything else looks good in the PR.

idelpivnitskiy: Discussed with @Scottmitch offline. He suggested to keep host and port independent because sometimes users don't know what is the port number (SRV resolutions or use of Unix Domain Sockets).

idelpivnitskiy (approved):
Great work!

```
FAILURE: Build failed with an exception.
428 actionable tasks: 353 executed, 75 up-to-date
What went wrong:
Execution failed for task ':servicetalk-http-netty:pmdTest'.
> 4 PMD rule violations were found. See the report at: file:///home/runner/work/servicetalk/servicetalk/servicetalk-http-netty/build/reports/pmd/test.html
```

The build fails because of #1596. Please, fix PMD warnings and 🚢

chemicL:
Fixed PMD errors, merged with `main` branch and ready to go.

PR #1563, opened 2021-05-14
Fix ProxyTunnel to always use the address instead of the hostname.

tkountis:
Motivation:

ProxyTunnel was returning the `localhost` hostname instead of the IP-address, which in some configurations
where the IPv4 is preferred but the DNS resolves in both IPv4 and IPv6 for the loopback hostname,
cause connectivity issues.

Modifications:

Modified the ProxyTunnel to always return the IP-address for the clients to connect to.

Result:

More stable test.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
@@ -85,7 +85,8 @@ final class ProxyTunnel implements AutoCloseable {
             }
             return null;
         });
-        return HostAndPort.of(getLoopbackAddress().getHostAddress(), serverSocket.getLocalPort());
+        InetSocketAddress serverSocketAddress = (InetSocketAddress) serverSocket.getLocalSocketAddress();
+        return HostAndPort.of(serverSocketAddress.getAddress().getHostAddress(), serverSocketAddress.getPort());
     }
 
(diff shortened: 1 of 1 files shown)


normanmaurer (approved):

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
```diff
@@ -86,7 +85,7 @@ HostAndPort startProxy() throws IOException {
             }
             return null;
         });
-        return HostAndPort.of((InetSocketAddress) serverSocket.getLocalSocketAddress());
+        return HostAndPort.of(getLoopbackAddress().getHostAddress(), serverSocket.getLocalPort());
```
idelpivnitskiy, line 88:
```suggestion
        InetSocketAddress serverSocketAddress = (InetSocketAddress) serverSocket.getLocalSocketAddress();
        return HostAndPort.of(serverSocketAddress.getAddress().getHostAddress(), serverSocketAddress.getPort());
```

We should still prefer using `serverSocket`, but query different IP to get the resolved IP.

idelpivnitskiy (changes requested):
Good catch!
For tests, we should always use IP, this is why we have `AddressUtils`. One suggestion:

idelpivnitskiy (approved):

idelpivnitskiy:
Thank you!

PR #1565, opened 2021-05-15
Add SingleThreadedExecutor

bondolo:
Motivation:
For some applications it is important that all signals be handled on
the same thread, sometimes called "Thread Affinity". This is typically
needed because of the use of `ThreadLocal` to pass state between
invocations.

Modifications:
Add a SingleThreadedExecutor implementation that can execute one task
at a time. While concurrent tasks are generally rejected, the
`execute(Runnable)` method provides special behavior if the invoking
thread is the executor worker thread: the `Runnable` will be executed
immediately on the calling thread.

Result:
Another option for handling of signals where thread affinity is
desired.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AsyncContextHolderThread.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AsyncContextHolderThread.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright © 2018-2019, 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.concurrent.api;
+
+import javax.annotation.Nullable;
+
+class AsyncContextHolderThread extends Thread implements AsyncContextMapHolder {
+    @Nullable
+    private AsyncContextMap asyncContextMap;
+
+    AsyncContextHolderThread(String name) {
+        super(name);
+    }
+
+    AsyncContextHolderThread(Runnable target, String name) {
+        super(target, name);
+    }
+
+    @Override
+    public void asyncContextMap(@Nullable final AsyncContextMap asyncContextMap) {
+        this.asyncContextMap = asyncContextMap;
+    }
+
+    @Nullable
+    @Override
+    public AsyncContextMap asyncContextMap() {
+        return asyncContextMap;
+    }
+}
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Executors.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Executors.java
@@ -49,6 +49,29 @@ public final class Executors {
     }
 
     /**
+     * Creates a new {@link Executor} that has a single dedicated worker thread. If {@link Executor#execute(Runnable)}
+     * method is invoked on the worker thread the {@link Runnable#run()} method will be executed immediately, but
+     * concurrent use is otherwise rejected.
+     *
+     * @return A new {@link Executor} that will use a single thread.
+     */
+    public static Executor newSingleThreadedExecutor() {
+        return from(new SingleThreadedExecutor());
+    }
+
+    /**
+     * Creates a new {@link Executor} that has a single dedicated worker thread. If {@link Executor#execute(Runnable)}
+     * method is invoked on the worker thread the {@link Runnable#run()} method will be executed immediately, but
+     * concurrent use is otherwise rejected.
+     *
+     * @param namePrefix A prefix for the worker thread name. A sequence number will always be appended.
+     * @return A new {@link Executor} that will use a single worker thread.
+     */
+    public static Executor newSingleThreadedExecutor(String namePrefix) {
+        return from(new SingleThreadedExecutor(namePrefix));
+    }
+
+    /**
      * Creates a new {@link Executor} that has a fixed number of threads as specified by the {@code size}.
      *
      * @param size Number of threads used by the newly created {@link Executor}.
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
@@ -19,11 +19,13 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.lang.ref.WeakReference;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.function.Supplier;
 
 import static java.util.concurrent.atomic.AtomicIntegerFieldUpdater.newUpdater;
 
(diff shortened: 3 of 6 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AsyncContextHolderThread.java
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.concurrent.api;
+
+import javax.annotation.Nullable;
+
+class AsyncContextHolderThread extends Thread implements AsyncContextMapHolder {
+    @Nullable
+    private AsyncContextMap asyncContextMap;
```
bondolo, line 22: It makes me feel itchy that this field is not volatile.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+        if (Thread.currentThread() == thread) {
+            try {
+                command.run();
+            } catch (Throwable all) {
+                LOGGER.warn("Uncaught throwable from command " + command, all);
+            }
+        } else {
+            try {
+                if (nonblocking) {
+                    if (!thread.queue.offer(command, 100, TimeUnit.MICROSECONDS)) {
```
chemicL, line 70: How do we know this number makes sense in all usage scenarios? Are there any reasons why it should not be configurable?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+                        throw new RejectedExecutionException("Executor is busy");
```
chemicL, line 71: I'd definitely like to see some hints as a user why this rejection happened. Perhaps adding the hint about the `nonblocking` flag being true to the message would suffice?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+
+            if (Thread.NORM_PRIORITY != getPriority()) {
+                setPriority(Thread.NORM_PRIORITY);
+            }
+        }
+
+        public void run() {
+            while (OPEN == state && null != owner.get()) {
+                try {
+                    Runnable command = queue.poll(1L, TimeUnit.MINUTES);
```
chemicL, line 123: What is the rationale behind polling? Would it not make sense to just call `queue.take()` and block until a task is ready or the thread is interrupted?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+                    if (null == command) {
+                        continue;
+                    }
+                    try {
+                        command.run();
+                    } catch (Throwable all) {
```
chemicL, line 129: In the ST codebase usually this `Throwable` uses `t` as identifier. TBH the name `all` makes more sense in boolean situations. Looking at a particular caught `Throwable` it's not an 'all throwable', but 'a throwable'. I'd vouch for `t`.

chemicL (changes requested):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
bondolo: Yes, it could be made configurable with this value as the default.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/DefaultExecutorTest.java
```diff
@@ -176,6 +211,8 @@ public void execution(ExecutorParam executorParam) throws Throwable {
     @EnumSource(ExecutorParam.class)
     public void longRunningTasksDoesNotHaltOthers(ExecutorParam executorParam) throws Throwable {
         executor = executorParam.get();
+        assumeTrue(executorParam.size() == UNBOUNDED || executorParam.size() > 1,
+                () -> "Ignoring executor: " + executorParam + ", it has only a single thread");
```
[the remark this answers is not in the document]
bondolo: OK

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+
+import static java.util.concurrent.atomic.AtomicIntegerFieldUpdater.newUpdater;
+
+/**
+ * Executor that can execute one task at a time on a dedicated worker thread. While concurrent tasks are generally
+ * rejected, the `execute(Runnable)` method provides special behavior if the invoking thread is the executor worker
+ * thread: the `Runnable` will be executed immediately on the calling thread.
```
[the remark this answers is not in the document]
bondolo: Unit test makes sense. I will add it.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+ */
+class SingleThreadedExecutor implements java.util.concurrent.Executor, AutoCloseable {
+    static final Logger LOGGER = LoggerFactory.getLogger(SingleThreadedExecutor.class);
+    private static final String DEFAULT_NAME_PREFIX = "servicetalk-solo";
+    private static final AtomicInteger INSTANCE_COUNT = new AtomicInteger();
```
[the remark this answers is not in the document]
bondolo: Depending on the application it is possible that there would be hundreds of instances. The count is shared so it is slightly confusing when the prefix changes but that's better than not being able to distinguish between threads because they all have the same name.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+        thread.start();
+    }
+
+    @Override
+    public void execute(final Runnable command) {
+        if (Thread.currentThread() == thread) {
+            try {
+                command.run();
+            } catch (Throwable all) {
+                LOGGER.warn("Uncaught throwable from command " + command, all);
```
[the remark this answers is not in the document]
bondolo: OK. For consistency I will make the change. If there is a performance concern about a particular log statement I usually add a `LOGGER.isXXXXEnabled()` guard. In this case warn is unlikely to be disabled and the exception should be very rare so performance isn't too much of a concern.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
bondolo: I don't understand the distinction you are making about this usage. I find the `t` identifier to not be very expressive of intent so would prefer to use `all`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Executors.java
```diff
@@ -48,6 +48,29 @@ public static Executor immediate() {
...
+
+    /**
+     * Creates a new {@link Executor} that has a single dedicated worker thread. If {@link Executor#execute(Runnable)}
+     * method is invoked on the worker thread the {@link Runnable#run()} method will be executed immediately, but
+     * concurrent use is otherwise rejected.
+     *
+     * @param namePrefix A prefix for the worker thread name. A sequence number will always be appended.
+     * @return A new {@link Executor} that will use a single worker thread.
+     */
+    public static Executor newSingleThreadedExecutor(String namePrefix) {
```
idelpivnitskiy, line 70: Are we confident that this is the approach to move forward? WDYT about exposing it only in some `internal` package or, let's say in this package but choose a different factory class: `ExperimentalExecutors`.
That way we can test how apps behave without risking that users depend on it.

idelpivnitskiy (commented):
Smells like more offloading fun is coming soon :)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AsyncContextHolderThread.java
bondolo: If it is only ever set/read on the thread for itself then agreed there is no issue.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Executors.java
bondolo: Certainly reasonable at this point. As long we can easily move or expose it later.

tkountis (approved):

tkountis:
didn't mean to approve, ignore that.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
```diff
@@ -0,0 +1,202 @@
...
+                    }
+                } catch (InterruptedException woken) {
+                    Thread.interrupted();
+                    // closing?
+                } catch (Throwable all) {
+                    LOGGER.warn("Uncaught throwable in worker thread", all);
+                }
+            }
+            int was = stateUpdater.getAndSet(this, null != owner.get() ? ORPHANED : EXITED);
+            LOGGER.debug("WorkerThread exiting " + (CLOSED == was ? "cleanly" : "messily"));
```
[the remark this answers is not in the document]
bondolo: Probably the log should be a warning if the exit was not clean. The extra states are indeed for debuggability, so that a conditional breakpoint can be triggered on the value being set or being present at a particular point.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
```diff
@@ -0,0 +1,202 @@
...
+import java.util.function.Supplier;
+
+import static java.util.concurrent.atomic.AtomicIntegerFieldUpdater.newUpdater;
+
+/**
+ * Executor that can execute one task at a time on a dedicated worker thread. While concurrent tasks are generally
+ * rejected, the `execute(Runnable)` method provides special behavior if the invoking thread is the executor worker
+ * thread: the `Runnable` will be executed immediately on the calling thread.
+ */
+class SingleThreadedExecutor implements java.util.concurrent.Executor, AutoCloseable {
```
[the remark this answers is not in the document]
idelpivnitskiy: Not sure about `java.io.Closeable` as it has nothing to do with IO.
Consider overriding `close()` method with more appropriate exception based on what we learned from a use-case with grpc #1566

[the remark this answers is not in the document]
bondolo: Since it is not a public API class I don't find the need to mark it (or the thread subclass) as `final`. The historical performance reasons will be handled by the JIT so dispatch should be no slower that if it were declared `final`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleThreadedExecutor.java
```diff
@@ -0,0 +1,200 @@
...
+    public void execute(final Runnable command) throws RejectedExecutionException {
+        if (WorkerThread.OPEN != thread.state) {
+            throw new RejectedExecutionException("Executor closed");
+        }
+
+        if (Thread.currentThread() == thread) {
+            try {
+                command.run();
+            } catch (Throwable all) {
+                LOGGER.warn("Uncaught throwable from command {}", command, all);
```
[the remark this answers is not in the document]
bondolo: As much as I don't like having the exception swallowed I wanted to make the behaviour of the immediate execution be as similar as possible to the asynchronous/offloaded execution. The issue would be that if an app comes to expect that an implementation will throw an exception and then another executor is used or a different code path is invoked the exception might be thrown or not thrown violating that expectation.

idelpivnitskiy:
@bondolo what is the current state of this PR? Is it ready for review or should it be marked as a draft for now?

bondolo:
I should have marked it draft or withdrawn it for now. I have marked it draft.

PR #1566, opened 2021-05-18
[HOLD] `GracefulAutoCloseable` renamed to `GracefulCloseable` to eliminate compiler warnings

chemicL:
Motivation:

`GracefulAutoCloseable` interface inherited from `AutoCloseable`, which has [certain limitations](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html#close--) regarding thrown `Exception`s that are validated at compile-time. Compiling GRPC-generated classes yielded warnings, which turn to errors when compiled with `-Xlint:try` option.

Modification:

- `GracefulAutoCloseable` renamed to `GracefulCloseable`,
- `GracefulCloseable` extends `java.io.Closeable` instead of `java.lang.AutoCloseable` which declares `IOException` as thrown from the `close()` method. That in turn doesn't produce compilation warnings,
- `AutoCloseables` renamed to `Closeables` as it operated on `GracefulAutoCloseable`s only,
- `Generator` class generates code responsible for wrapping thrown exceptions from the `close()` and `closeGracefully()` methods into `IOException`.

Result:

Related code and GRPC-generated code don't produce compiler warnings for the try-with-resources blocks when `GracefulCloseable` is used.

### DIFF
--- a/servicetalk-concurrent/src/main/java/io/servicetalk/concurrent/GracefulCloseable.java
+++ b/servicetalk-concurrent/src/main/java/io/servicetalk/concurrent/GracefulCloseable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019-2021 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,10 +15,13 @@
  */
 package io.servicetalk.concurrent;
 
+import java.io.Closeable;
+import java.io.IOException;
+
 /**
- * An extension of {@link AutoCloseable} to add graceful closure semantics.
+ * An extension of {@link Closeable} to add graceful closure semantics.
  */
-public interface GracefulAutoCloseable extends AutoCloseable {
+public interface GracefulCloseable extends Closeable {
 
     /**
      * Used to close/shutdown a resource, similar to {@link #close()}, but attempts to cleanup state before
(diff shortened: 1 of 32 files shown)


bondolo (commented):
Have we considered just narrowing the exceptions declared to be thrown rather than changing the types?

servicetalk-concurrent/src/main/java/io/servicetalk/concurrent/GracefulCloseable.java
```diff
@@ -15,10 +15,13 @@
...
 
+import java.io.Closeable;
+import java.io.IOException;
+
 /**
- * An extension of {@link AutoCloseable} to add graceful closure semantics.
+ * An extension of {@link Closeable} to add graceful closure semantics.
  */
-public interface GracefulAutoCloseable extends AutoCloseable {
+public interface GracefulCloseable extends Closeable {
```
tkountis, line 24: Some notes from our offline discussion.

Since we are breaking API, I was considering whether the module owner of this class makes sense.
`GracefulCloseable` seems to be used in the context of I/O (ie. Http / gRPC). What is its correlation with the concurrent package? It would probably be a better fit, in some transport-related package.
WDYT? /cc @bondolo @idelpivnitskiy

chemicL:
> An example of the lint warning would be useful.

Sample warning:

```
/home/dariusz/development/projects/github/servicetalk/servicetalk-examples/grpc/helloworld/build/generated/sources/proto/main/java/io/grpc/examples/helloworld/Greeter.java:68: warning: [try] auto-closeable resource BlockingSayHelloRpc has a member method close() that could throw InterruptedException
    public interface BlockingSayHelloRpc extends BlockingGrpcService {
           ^
```

> Have we considered just narrowing the exceptions declared to be thrown rather than changing the types?

I was wondering what narrowing we could try to apply. The pattern used throughout the test codebase and examples is by using try-with-resources blocks on top of the main resource or by using a `get` on a `Future` on the `ServerContext` which heavily uses `AsyncCloseable` with `CompositeCloseable`. My undestanding is that this is how cleanup upon error/interruption is performed and it's a pattern used for any reasource. Unfortunately, these interfaces are exposed as public and can be implemented by our users. Which means we have no control over the call site. Unless we narrow down to essentially no checked exceptions, I'm not sure whether there is a meaningful subclass of `Exception` which would be generic enough but not inclusive of `InterruptedException` to eliminate the warnings.

chemicL:
This PR is out of date. The subject can be revisited at some point, but for now I'm closing it.

PR #1521, opened 2021-05-03
junit4 to junit5 migration. step 2 - modules `servicetalk-http-api` and `servicetalk-http-netty`

danfaer:
Motivation:
 
- JUnit 5 leverages features from Java 8 or later, such as lambda functions, making tests more powerful and easier to maintain.
- JUnit 5 has added some very useful new features for describing, organizing, and executing tests. For instance, tests get better display names and can be organized hierarchically.
- JUnit 5 is organized into multiple libraries, so only the features you need are imported into your project. With build systems such as Maven and Gradle, including the right libraries is easy.
- JUnit 5 can use more than one extension at a time, which JUnit 4 could not (only one runner could be used at a time). This means you can easily combine the Spring extension with other extensions (such as your own custom extension).

[migrating-from-junit-4-to-junit-5-important-differences-and-benefits](https://blogs.oracle.com/javamagazine/migrating-from-junit-4-to-junit-5-important-differences-and-benefits)

Modifications:
   Updated unit tests 
   
Result:
   modules`servicetalk-http-api` and `servicetalk-http-netty` run tests using junit5 engine

PR #1570, opened 2021-05-21
Make `http-api:testFixtures` usable in other modules

idelpivnitskiy:
Motivation:

`testFixtures` are expected to be used by other modules.
After #1521 refactoring, visibility of some methods was reduced.

Modifications:

- Revert `protected` modifiers for `AbstractHttpRequesterFilterTest` and
`AbstractHttpServiceFilterTest` methods;
- Do not override `sslSession()` without a special need;

Result:

Modules that depend on these testFixtures classes can continue to work.


idelpivnitskiy:
Change is trivial and changes only test classes. Let me merge it to unblock myself and I will send a follow-up PR if you have any comments.

PR #1572, opened 2021-05-21
Test `BufferInputStream#skip(long)`

idelpivnitskiy:
Motivation:

Had an assumption that `BufferInputStream#skip(long)` has a bug.

Modifications:

- Add `BufferInputStreamTest` class that tests `skip(long)` method;

Result:

Confidence that `BufferInputStream#skip(long)` is implemented correctly.


idelpivnitskiy:
Trivial test, no changes for the main codebase. Merging

PR #1573, opened 2021-05-24
Produce a TrailersOnly response when gRPC endpoints throw an error.

tkountis:
Motivation:
On some erroneous cases we can optimise the gRPC response to return TrailerOnly [1] when headers aren't written to the wire yet. 

Modifications:
gRPC router was altered to generate a TrailersOnly response when an error happens before headers are sent out.

Result:
An optimised response that contains only a single frame.

[1] https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019-2020 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019-2021 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -74,8 +74,8 @@ import static io.servicetalk.grpc.api.GrpcStatus.fromCodeValue;
 import static io.servicetalk.grpc.api.GrpcStatusCode.INVALID_ARGUMENT;
 import static io.servicetalk.grpc.api.GrpcStatusCode.UNIMPLEMENTED;
 import static io.servicetalk.grpc.api.GrpcUtils.negotiateAcceptedEncoding;
-import static io.servicetalk.grpc.api.GrpcUtils.newErrorResponse;
 import static io.servicetalk.grpc.api.GrpcUtils.newResponse;
+import static io.servicetalk.grpc.api.GrpcUtils.newTrailersOnlyErrorResponse;
 import static io.servicetalk.grpc.api.GrpcUtils.readGrpcMessageEncoding;
 import static io.servicetalk.grpc.api.GrpcUtils.setStatus;
 import static io.servicetalk.grpc.api.GrpcUtils.setStatusOk;
@@ -97,7 +97,7 @@ final class GrpcRouter {
 
     private static final GrpcStatus STATUS_UNIMPLEMENTED = fromCodeValue(UNIMPLEMENTED.value());
     private static final StreamingHttpService NOT_FOUND_SERVICE = (ctx, request, responseFactory) -> {
-        final StreamingHttpResponse response = newResponse(responseFactory, null, STATUS_UNIMPLEMENTED,
+        final StreamingHttpResponse response = newTrailersOnlyErrorResponse(responseFactory, STATUS_UNIMPLEMENTED,
                 ctx.executionContext().bufferAllocator());
         response.version(request.version());
         return succeeded(response);
@@ -264,10 +264,11 @@ final class GrpcRouter {
                                                 .payloadBody(rawResp,
                                                         serializationProvider.serializerFor(responseEncoding,
                                                                 responseClass)))
-                                        .onErrorReturn(cause -> newErrorResponse(responseFactory,
+                                        .onErrorReturn(cause ->
+                                                newTrailersOnlyErrorResponse(responseFactory,
                                                 finalServiceContext, cause, ctx.executionContext().bufferAllocator()));
                             } catch (Throwable t) {
-                                return succeeded(newErrorResponse(responseFactory, serviceContext, t,
+                                return succeeded(newTrailersOnlyErrorResponse(responseFactory, serviceContext, t,
                                         ctx.executionContext().bufferAllocator()));
                             }
                         }
@@ -320,7 +321,7 @@ final class GrpcRouter {
                                     serializationProvider.serializerFor(responseEncoding, responseClass),
                                     ctx.executionContext().bufferAllocator()));
                         } catch (Throwable t) {
-                            return succeeded(newErrorResponse(responseFactory, serviceContext, t,
+                            return succeeded(newTrailersOnlyErrorResponse(responseFactory, serviceContext, t,
                                     ctx.executionContext().bufferAllocator()));
                         }
                     }
@@ -465,7 +466,7 @@ final class GrpcRouter {
                                         ctx.executionContext().bufferAllocator()).payloadBody(response,
                                                 serializationProvider.serializerFor(responseEncoding, responseClass));
                             } catch (Throwable t) {
-                                return newErrorResponse(responseFactory, serviceContext, t,
+                                return newTrailersOnlyErrorResponse(responseFactory, serviceContext, t,
                                         ctx.executionContext().bufferAllocator());
                             }
                         }
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
@@ -74,8 +74,8 @@ import static io.servicetalk.grpc.api.GrpcStatus.fromCodeValue;
 import static io.servicetalk.grpc.api.GrpcStatusCode.INVALID_ARGUMENT;
 import static io.servicetalk.grpc.api.GrpcStatusCode.UNIMPLEMENTED;
 import static io.servicetalk.grpc.api.GrpcUtils.negotiateAcceptedEncoding;
+import static io.servicetalk.grpc.api.GrpcUtils.newErrorResponse;
 import static io.servicetalk.grpc.api.GrpcUtils.newResponse;
-import static io.servicetalk.grpc.api.GrpcUtils.newTrailersOnlyErrorResponse;
 import static io.servicetalk.grpc.api.GrpcUtils.readGrpcMessageEncoding;
 import static io.servicetalk.grpc.api.GrpcUtils.setStatus;
 import static io.servicetalk.grpc.api.GrpcUtils.setStatusOk;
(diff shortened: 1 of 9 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -97,7 +97,7 @@
 
     private static final GrpcStatus STATUS_UNIMPLEMENTED = fromCodeValue(UNIMPLEMENTED.value());
     private static final StreamingHttpService NOT_FOUND_SERVICE = (ctx, request, responseFactory) -> {
-        final StreamingHttpResponse response = newResponse(responseFactory, null, STATUS_UNIMPLEMENTED,
+        final StreamingHttpResponse response = newTrailersOnlyErrorResponse(responseFactory, STATUS_UNIMPLEMENTED,
```
idelpivnitskiy, line 100: Assume all changes in this class and `GrpcServerBuilder` will be reverted after the method name is changed back to  `newResponse`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
```diff
@@ -135,8 +135,9 @@ void transformMessageBody(UnaryOperator<Publisher<?>> transformer) {
 
     <T> void transform(final TrailersTransformer<T, Buffer> trailersTransformer) {
         if (messageBody == null) {
-            messageBody = from(trailersTransformer.payloadComplete(trailersTransformer.newState(),
-                    headersFactory.newEmptyTrailers()));
+            messageBody = defer(() ->
+                from(trailersTransformer.payloadComplete(trailersTransformer.newState(),
+                        headersFactory.newEmptyTrailers())));
```
idelpivnitskiy, line 140: This change is not related to the described issue. Can you please open a separate PR with a test for this use-case?

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/TrailersOnlyErrorTest.java
```diff
@@ -0,0 +1,254 @@
...
+    public void testRouteThrows() throws Exception {
+        final BlockingQueue<Throwable> asyncErrors = new LinkedBlockingDeque<>();
+        final ServerContext serverContext = GrpcServers.forAddress(localAddress(0))
+                .listenAndAwait(new TesterProto.Tester.ServiceFactory(mockTesterService()));
+
+        final GrpcClientBuilder<HostAndPort, InetSocketAddress> clientBuilder =
+                GrpcClients.forAddress(serverHostAndPort(serverContext))
+                .appendHttpClientFilter(__ -> true, setupResponseVerifierFilter(asyncErrors));
+
+        Greeter.GreeterClient client = clientBuilder.build(new Greeter.ClientFactory());
```
idelpivnitskiy, line 83: 1. Because you do not have before/after methods to set/clean the state, consider using try-with-resources for `ServerContext` and `GreeterClient` to automatically close them in all tests.
2. Consider adding an imports to avoid `Greeter.` prefix.
3. It took some time until I understood why the server returns `UNIMPLEMENTED` :) Please, add a comment highlighting that you use another client to talk to `TesterProto` service to similate `UNIMPLEMENTED`.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/TrailersOnlyErrorTest.java
```diff
@@ -0,0 +1,254 @@
...
+                .build()).concat(never())).toFuture(), UNKNOWN);
+        assertNoErrors(asyncErrors);
+
+        verifyException(client.testBiDiStream(from(TesterProto.TestRequest.newBuilder()
+                .build())).toFuture(), UNKNOWN);
+        assertNoErrors(asyncErrors);
+    }
+
+    private static void assertNoErrors(final BlockingQueue<Throwable> errors) {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
idelpivnitskiy, line 180: We already have a utility like this in `H2PriorKnowledgeFeatureParityTest` and `AsyncContextHttpFilterVerifier`. `servicetalk-test-resources` might be a good place to share it.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/TrailersOnlyErrorTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+    private static void verifyException(final Throwable cause, final GrpcStatusCode expectedCode) {
+        assertNotNull(cause);
+        assertThat(assertThrows(GrpcStatusException.class, () -> {
+            throw cause;
+        }).status().code(), equalTo(expectedCode));
+    }
+
+    private static TesterProto.Tester.TesterService mockTesterService() {
+        TesterProto.Tester.TesterService filter = mock(TesterProto.Tester.TesterService.class);
```
idelpivnitskiy, line 208: Consider using imports to avoid repeated `TesterProto.Tester.` prefix

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/TrailersOnlyErrorTest.java
```diff
@@ -0,0 +1,254 @@
...
+        final ServerContext serverContext = GrpcServers.forAddress(localAddress(0))
+                .listenAndAwait(new TesterProto.Tester.ServiceFactory(mockTesterService()));
+
+        final GrpcClientBuilder<HostAndPort, InetSocketAddress> clientBuilder =
+                GrpcClients.forAddress(serverHostAndPort(serverContext))
+                .appendHttpClientFilter(__ -> true, setupResponseVerifierFilter(asyncErrors));
+
+        Greeter.GreeterClient client = clientBuilder.build(new Greeter.ClientFactory());
+        assertNoErrors(asyncErrors);
+        verifyException(client.sayHello(HelloRequest.newBuilder().build()).toFuture(), UNIMPLEMENTED);
```
idelpivnitskiy, line 85: In this order, `assertNoErrors` does not verify anything because you haven't sent a request yet. You need to swap these two statements.

idelpivnitskiy (changes requested):
Thanks for the additional test coverage!

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -167,18 +165,31 @@ static HttpResponse newResponse(final HttpResponseFactory responseFactory,
...
+        initResponse(response, null);
+        setStatus(response.headers(), status, null, allocator);
+        return response;
+    }
+
+    static HttpResponse newTrailersOnlyErrorResponse(final HttpResponseFactory responseFactory,
+                                                     @Nullable final GrpcServiceContext context,
+                                                     final Throwable cause, final BufferAllocator allocator) {
+        final HttpResponse response = responseFactory.ok();
+        initResponse(response, context);
```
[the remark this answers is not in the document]
tkountis: According to the spec [1] the server MAY send the Message-Accept-Encoding to the client. Its mostly useful for cases where an incoming encoding isn't supported so you advertise to the peer which ones are supported. We can tune it down and only produce it in case of a compression error. Thoughts?

[2] https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses
> Response-Headers → HTTP-Status [Message-Encoding] [Message-Accept-Encoding] Content-Type *Custom-Metadata

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -167,18 +165,31 @@ static HttpResponse newResponse(final HttpResponseFactory responseFactory,
         return response;
     }
 
-    static HttpResponse newErrorResponse(final HttpResponseFactory responseFactory,
-                                         @Nullable final GrpcServiceContext context,
-                                         final Throwable cause, final BufferAllocator allocator) {
-        HttpResponse response = newResponse(responseFactory, context, allocator);
-        setStatus(response.trailers(), cause, allocator);
+    static StreamingHttpResponse newTrailersOnlyErrorResponse(final StreamingHttpResponseFactory responseFactory,
```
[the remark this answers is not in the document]
tkountis: The overload was for the status vs cause fields not for the context. I changed it per your recommendation, but the overload was useful due to the many places we now pass null.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
tkountis: Issue: https://github.com/apple/servicetalk/issues/1576
I will spend some more time on that and follow with a PR if needed.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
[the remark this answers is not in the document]
idelpivnitskiy: I missed that part, thanks for pointing. Let's keep as-is for now.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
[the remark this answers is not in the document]
idelpivnitskiy: I see now, thanks! Missed that it used `newResponse` before, not `newErrorResponse`.
Up to you then. We can keep `null`s or add another overload. Thanks for clarifying.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -167,18 +165,32 @@ static HttpResponse newResponse(final HttpResponseFactory responseFactory,
...
 
-    static HttpResponse newErrorResponse(final HttpResponseFactory responseFactory,
-                                         @Nullable final GrpcServiceContext context,
-                                         final Throwable cause, final BufferAllocator allocator) {
-        HttpResponse response = newResponse(responseFactory, context, allocator);
-        setStatus(response.trailers(), cause, allocator);
+    static HttpResponse newErrorResponse(
+            final HttpResponseFactory responseFactory, @Nullable final GrpcServiceContext context,
+            @Nullable final GrpcStatus status, @Nullable final Throwable cause, final BufferAllocator allocator) {
+        assert status != null || cause != null;
```
idelpivnitskiy, line 171: If you want to check that only one of them is set, assertion should be more strict:
```java
assert (status != null && cause == null) || (status == null && cause != null);
```

idelpivnitskiy (approved):
Thank you!

servicetalk-test-resources/src/main/java/io/servicetalk/test/resources/TestUtils.java
```diff
@@ -0,0 +1,48 @@
...
+        // No instances
+    }
+
+    /**
+     * Helper method to check if a given {@link Queue} contains any errors,
+     * usually produced through async sources. For ALL {@link Throwable}s found in the queue,
+     * a suppressed error will be captured and surface through a single {@link AssertionError}.
+     * @param errors The queue of captured errors.
+     */
+    public static void assertNoAsyncErrors(final Queue<Throwable> errors) {
```
bondolo, line 35: Could this be Iterable? There a cases where `List` is used for capturing async errors

tkountis: The reason this was a Queue originally is that it offers direct access to multi-threaded implementations. That's not to say that you can't have a multithreaded list of course. I guess we can relax this to an iterable. I will follow up.

Issue #1576, opened 2021-05-25
StreamingHttpPayloadHolder transform invokes the `payloadComplete` on the caller thread.


tkountis:
The transform method in the StreamingHttpPayloadHolder calls payloadComplete on the caller thread which should be called when the message-body is subscribed on.

PR #1577, opened 2021-05-25
Defer trailer transformer callbacks on null msg-body to subscription.

tkountis:
Motivation:

When we have a null msg-body and call transform, then the callbacks of the trailer
transformer are invoked on the calling thread rather than the subsciption.

Modifications:

Used defer to defer the invocation.

Result:

Better error handling in case of erroneous trailer transformer (ie. trailers validations)
and deferred execution of the callbacks.

Fix: https://github.com/apple/servicetalk/issues/1576


idelpivnitskiy (commented):
I looked at the CI failures: https://github.com/apple/servicetalk/pull/1577/checks?check_run_id=2666777501
Never saw `ConnectionCloseHeaderHandlingTest$NonPipelinedRequestsTest` being flaky before this PR. Let's hold it for now?

tkountis:
Build failure due https://github.com/apple/servicetalk/issues/1579

tkountis:
@idelpivnitskiy as we discussed offline, we have seen this error again outside the context of this change. 
PTAL at the change.

idelpivnitskiy (approved):
🚀

PR #1580, opened 2021-05-26
Preparing for 0.41.0-SNAPSHOT development

tkountis:

PR #1582, opened 2021-05-26
release.sh: check that the next version doesn't equal to the current one

idelpivnitskiy:
Motivation:

To prevent scenarios like #1580, let's enhance validation of input
parameter for the `release.sh` script.

Modifications:

- Compare passed next version with the current version from
`gradle.properties`, fails fast if they are equal;

Result:

Enhanced validation of next version parameter.


tkountis (approved):
🚢

PR #1583, opened 2021-05-26
assertNoAsyncErrors more queue types and msg arg

bondolo:
Motivation:
Converting some existing code to use `assertNoAsyncErrors` I found
cases where the error queue was `AssertionError` not `Throwable` and
some cases where the message was used to provide more specific info
about the source/context of the errors.
Modifications:
Add additional overload for message and allow wildcard for queue type.
Result:
More useful `assertNoAsyncErrors`.

### DIFF
--- a/servicetalk-test-resources/src/main/java/io/servicetalk/test/resources/TestUtils.java
+++ b/servicetalk-test-resources/src/main/java/io/servicetalk/test/resources/TestUtils.java
@@ -15,6 +15,7 @@
  */
 package io.servicetalk.test.resources;
 
+import java.util.Objects;
 import java.util.Queue;
 
 /**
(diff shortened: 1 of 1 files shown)


servicetalk-test-resources/src/main/java/io/servicetalk/test/resources/TestUtils.java
```diff
@@ -32,12 +33,23 @@ private TestUtils() {
      * a suppressed error will be captured and surface through a single {@link AssertionError}.
      * @param errors The queue of captured errors.
      */
-    public static void assertNoAsyncErrors(final Queue<Throwable> errors) {
```
bondolo, line 35: Regarding my earlier request to consider using `Iterable`. This turns out to not be needed, queue is fine.

tkountis (approved):

bondolo:
Failure is due to #1579

PR #1588, opened 2021-05-28
Allow JUnit 5 tests to execute concurrently

bondolo:
Motivation:
The JUnit5 tests are currently all run serially which is slower than
would generally be possible on most configurations.
Modifications:
Add junit-platform execution parameters for allowing tests to run
concurrently if suitably configured to improve test execution
performance.
Result:
Tests are annotated with `@Execution(CONCURRENT)` will execute
concurrently. There are no changes for existing un-annotated
tests which not explicitly opt-in to parallel execution.

### DIFF
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleToCompletionStageTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleToCompletionStageTest.java
@@ -1038,7 +1038,8 @@ class SingleToCompletionStageTest {
     }
 
     private static void verifyInJUnitThread() {
-        if (!currentThread().getName().startsWith(JUNIT_THREAD_PREFIX)) {
+        if (!currentThread().getName().startsWith(JUNIT_THREAD_PREFIX) &&
+                !currentThread().getName().startsWith(JDK_FORK_JOIN_THREAD_NAME_PREFIX)) {
             throw new AssertionError("unexpected thread: " + currentThread());
         }
     }
@@ -1052,8 +1053,9 @@ class SingleToCompletionStageTest {
 
     private static void verifyInStOrJUnitThread() {
         if (!currentThread().getName().startsWith(ST_THREAD_PREFIX_NAME) &&
-                !currentThread().getName().startsWith(JUNIT_THREAD_PREFIX)) {
-            throw new IllegalStateException("unexpected thread: " + currentThread());
+                !currentThread().getName().startsWith(JUNIT_THREAD_PREFIX) &&
+                !currentThread().getName().startsWith(JDK_FORK_JOIN_THREAD_NAME_PREFIX)) {
+            throw new AssertionError("unexpected thread: " + currentThread());
         }
     }
 
--- a/servicetalk-test-resources/src/main/resources/junit-platform.properties
+++ b/servicetalk-test-resources/src/main/resources/junit-platform.properties
@@ -1,0 +1,8 @@
+#
+# See https://junit.org/junit5/docs/snapshot/user-guide/index.html#writing-tests-parallel-execution
+#
+junit.jupiter.execution.parallel.enabled = true
+junit.jupiter.execution.parallel.mode.default = same_thread
+junit.jupiter.execution.parallel.mode.classes.default = same_thread
+junit.jupiter.execution.parallel.config.strategy = dynamic
+junit.jupiter.execution.parallel.config.dynamic.factor = 1.5
--- a/servicetalk-test-resources/src/main/resources/junit-platform.properties
+++ b/servicetalk-test-resources/src/main/resources/junit-platform.properties
@@ -5,4 +5,4 @@ junit.jupiter.execution.parallel.enabled = true
 junit.jupiter.execution.parallel.mode.default = same_thread
 junit.jupiter.execution.parallel.mode.classes.default = same_thread
 junit.jupiter.execution.parallel.config.strategy = dynamic
-junit.jupiter.execution.parallel.config.dynamic.factor = 1.5
+junit.jupiter.execution.parallel.config.dynamic.factor = 1.0


bondolo:
This draft PR is open for collaboration. Feel free to pull it locally, run a build to see which tests break, fix a test and update the PR. Code review is welcome, as always, but won't begin in earnest until we have all tests running cleanly with concurrent execution.

servicetalk-test-resources/src/main/resources/junit-platform.properties
```diff
@@ -0,0 +1,5 @@
+junit.jupiter.execution.parallel.enabled = true
+junit.jupiter.execution.parallel.mode.default = same_thread
+junit.jupiter.execution.parallel.mode.classes.default = concurrent
+junit.jupiter.execution.parallel.config.strategy = dynamic
+junit.jupiter.execution.parallel.config.dynamic.factor = 2
```
idelpivnitskiy, line 5: Interesting, I thought that gradle runs tests concurrently because we use `--parallel` option: 
https://github.com/apple/servicetalk/blob/a51d2f275592f7fa2eb1a822ac4964dd24bdba0d/.github/workflows/ci-prb.yml#L45

Two questions I have:
1. Does the parallel execution at jUnit 5 level give us additional benefits to gradle's `--parallel` flag?
2. How does it work for modules that still use jUnit4 but via jUnit5 vintage engine we apply it in the plugin?
https://github.com/apple/servicetalk/blob/ad7e8e1e6abe1bbc024b03c93c376962a87e4be9/servicetalk-gradle-plugin-internal/src/main/groovy/io/servicetalk/gradle/plugin/internal/ServiceTalkLibraryPlugin.groovy#L197-L200

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleToCompletionStageTest.java
```diff
@@ -1032,27 +1032,29 @@ private static void verifyInJdkForkJoinThread() {
...
         }
     }
 
     private static void verifyInStOrJUnitThread() {
         if (!currentThread().getName().startsWith(ST_THREAD_PREFIX_NAME) &&
-                !currentThread().getName().startsWith(JUNIT_THREAD_PREFIX)) {
-            throw new IllegalStateException("unexpected thread: " + currentThread());
+                !currentThread().getName().startsWith(JUNIT_THREAD_PREFIX) &&
+                !currentThread().getName().startsWith(JDK_FORK_JOIN_THREAD_NAME_PREFIX)) {
+            throw new AssertionError("unexpected thread: " + currentThread());
```
idelpivnitskiy, line 1057: If those are improvements for tests, consider opening a separate PR.

idelpivnitskiy (commented):
The only hesitation I have is the flaky tests. If we apply more parallelism, we may see more flaky test failures. Which may be good and let us discover issues, if we can prioritize that work.

servicetalk-test-resources/src/main/resources/junit-platform.properties
bondolo: 1. Gradle `--parallel` is only parallelism at the Gradle task level, it doesn't affect parallelism within a task (at least as far as I can tell). This would mean that javadoc can be built at the same time as spotbugs runs, etc. or any other tasks that don't have an dependency. Trying a build locally using `--parallel`, which I thought was the default based upon seeing multiple Gradle daemons running, the results are **much faster** builds, 5.5 minutes vs about 11-12 minutes.  So I am curious why we use `--parallel` in some of the GitHub action yaml config files, but do not have `org.gradle.parallel=true` in our `gradle.properties` file. If it works well enough for CI it should work in all circumstances.
We should see further build performance improvements by also using test execution parallelism.
2. I do not know what impact it has for JUnit 4. Investigation will be needed.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleToCompletionStageTest.java
bondolo: The test parallelism causes JUnit to run the tests on the default ForkJoinPool so the thread observed changes from `Test-` to `ForkJoinPool-`

servicetalk-test-resources/src/main/resources/junit-platform.properties
idelpivnitskiy: Thanks for #1591!

Read the docs: https://junit.org/junit5/docs/snapshot/user-guide/index.html#writing-tests-parallel-execution
I like that they allow this type of configuration:
```
junit.jupiter.execution.parallel.mode.default = same_thread
junit.jupiter.execution.parallel.mode.classes.default = concurrent
```
This mode makes perfect sense for our tests.

Couple comments:
1. Please, add a link to the jUnit5 docs at the beginning of the file.
2. Does `dynamic.factor=2` provide more speed up? It may make sense for CI to have the factor == 2, but for local development it may consume too much CPU and won't allow developers to do other work when tests are running. Let's start with 1?

bondolo: > Read the docs: https://junit.org/junit5/docs/snapshot/user-guide/index.html#writing-tests-parallel-execution
> I like that they allow this type of configuration:
> 
> ```
> junit.jupiter.execution.parallel.mode.default = same_thread
> junit.jupiter.execution.parallel.mode.classes.default = concurrent
> ```
> 
> This mode makes perfect sense for our tests.
> 
> Couple comments:
> 
> 1. Please, add a link to the jUnit5 docs at the beginning of the file.

Can do.

> 2. Does `dynamic.factor=2` provide more speed up? It may make sense for CI to have the factor == 2, but for local development it may consume too much CPU and won't allow developers to do other work when tests are running. Let's start with 1?

That was from previous experience with JUnit4/Maven parallelization and may have been specific to that project. The tests there had some blocking portions so running at a little higher concurrency kept the CPUs busier. The ServiceTalks tests may not benefit much and you may be right that it would completely saturate a developer system. I will revert to 1 and then do comparisons once the tests are actually running in parallel.

Scottmitch:
there are some tests that share state or use static state (e.g. async context), is there a way to tests which have these characteristics always run in a serial fashion?

servicetalk-test-resources/src/main/resources/junit-platform.properties
```diff
@@ -1,5 +1,8 @@
+#
+# See https://junit.org/junit5/docs/snapshot/user-guide/index.html#writing-tests-parallel-execution
+#
 junit.jupiter.execution.parallel.enabled = true
 junit.jupiter.execution.parallel.mode.default = same_thread
-junit.jupiter.execution.parallel.mode.classes.default = concurrent
+junit.jupiter.execution.parallel.mode.classes.default = same_thread
```
idelpivnitskiy, line 6: having `same_thread` on both won't do anything concurrently, see the picture here: https://junit.org/junit5/docs/snapshot/user-guide/index.html#writing-tests-parallel-execution
What you had previously was the right choice. It executes all tests withing a single class sequentially, but runs independent test classes concurrently:
```
junit.jupiter.execution.parallel.enabled = true
junit.jupiter.execution.parallel.mode.default = same_thread
junit.jupiter.execution.parallel.mode.classes.default = concurrent
```

bondolo: As proposed, the options are intended to run tests in the same manner as our current non-parallel execution. Only the first line, `junit.jupiter.execution.parallel.enabled = true`, is intended to have any effect. The rest of the lines just make the defaults explicit (I hope, though it is certainly possible I screwed it up). So the effect of the first line is minor in that it merely allows parallel execution, it doesn't cause parallel execution. This configuration works with our current tests, some of which fail when executed in parallel due to interdependencies between cases, but does allow us to start adding `@Execution` annotations to individual suites/tests that can be parallelized. There are some rather large suites parameterized test that can hopefully be accelerated.

jbduncan: Hi all - I found my way here from a [thread](https://gitter.im/junit-team/junit5?at=60f721a2926ce249e576a3b1) started on the JUnit 5 Gitter.

I'm rather curious, have you already tried running the tests with all the `parallel.mode` sub-options set to `concurrent`?

Either way, I completely understand why you're taking the cautious approach. :)

bondolo: @jbduncan wrote:
> I'm rather curious, have you already tried running the tests with all the `parallel.mode` sub-options set to `concurrent`?

Yes, we have tried `concurrent` as the default and some of our tests do fail. We will have to incrementally mark tests using `@Execution(CONCURRENT)` and work on fixing individual tests. There are a few "tentpole" tests that consume a lot of time due to sequential execution (and hold up subsequent parallel portions of the build) so we will probably focus on those first.

bondolo:
Now ready for review and integration. This PR will not change any observable behavior for test execution but will allow individual test suites to opt-in to parallel test execution via use of the `@Execution(CONCURRENT)` annotation.

servicetalk-test-resources/src/main/resources/junit-platform.properties
```diff
@@ -0,0 +1,8 @@
+#
+# See https://junit.org/junit5/docs/snapshot/user-guide/index.html#writing-tests-parallel-execution
+#
+junit.jupiter.execution.parallel.enabled = true
+junit.jupiter.execution.parallel.mode.default = same_thread
+junit.jupiter.execution.parallel.mode.classes.default = same_thread
+junit.jupiter.execution.parallel.config.strategy = dynamic
+junit.jupiter.execution.parallel.config.dynamic.factor = 1.5
```
idelpivnitskiy, line 8: How about starting with 1 as we discussed before: https://github.com/apple/servicetalk/pull/1588#discussion_r643575718?

idelpivnitskiy (approved):

servicetalk-test-resources/src/main/resources/junit-platform.properties
bondolo: Sure.

PR #1602, opened 2021-06-02
Remove deprecated `idleTimeout()` methods

bondolo:
Motivation:
Several methods were renamed from `idleTimeout` to simply `timeout` and
the old names were retained but deprecated. The old names can be
removed as the names have now been deprecated for several releases.
Modifications:
The deprecated `idleTimeout` methods are removed from `Completable`,
`Publisher`, and `Single`.
Result:
Deprecated renamed methods removed in favor of new name `tiemout`.


idelpivnitskiy (commented):
Thanks for opening a PR!
LGTM when we are ready to merge breaking changes.

idelpivnitskiy (approved):
LGTM after resolving conflicts

Scottmitch (approved):

Issue #1568, opened 2021-05-20
Convert jUnit4 tests to jUnit5


idelpivnitskiy:
@danfaer has done huge work by migrating the biggest and the most complex modules (`concurrent-api`, `http-api`, and `http-netty`) from jUnit4 to jUnit5 in #1415 and #1521.

For those, who want to learn ServiceTalk API, experience it as a user, and start contributing it might be a good exercise to do a similar conversion for other modules.

Each pull request can convert only one module. There are plenty of modules that contain only a few test classes. Please, look at the referenced pull requests and make your work consistent with their style (how dependencies are declared, how parametrized tests are used, how timeout is configured, etc.).

The whole list of modules with tests:

- [x] `servicetalk-buffer-api` #1592
- [x] `servicetalk-buffer-netty` #1593
- [x] `servicetalk-client-api-internal` #1608
- [x] `servicetalk-client-api` #1612
- [x] `servicetalk-concurrent-api-internal` #1608
- [x] `servicetalk-concurrent-api` #1415
- [x] `servicetalk-concurrent-api-internal` #1658
- [x] `servicetalk-concurrent-api-test` #1658 
- [x] `servicetalk-concurrent-internal` #1779
- [x] `servicetalk-concurrent-jdkflow` #1624
- [x] `servicetalk-concurrent-reactivestreams` #1628
- [x] `servicetalk-concurrent-test-internal` #1658 
- [x] `servicetalk-data-jackson-jersey` #1645
- [x] `servicetalk-data-jackson` #1664
- [x] `servicetalk-data-protobuf` #1664
- [x] `servicetalk-dns-discovery-netty` #1607
- [x] `servicetalk-encoding-api` #1525
- [x] `servicetalk-encoding-netty` #1664
- [x] `servicetalk-grpc-api` #1660
- [x] `servicetalk-grpc-internal` #1501
- [x] `servicetalk-grpc-netty` #1660
- [x] `servicetalk-grpc-protobuf` #1660
- [x] `servicetalk-grpc-protoc` #1660
- [x] `servicetalk-http-api` #1521
- [x] `servicetalk-http-netty` #1521
- [x] `servicetalk-http-router-jersey` #1645
- [x] `servicetalk-http-router-predicate` #1659
- [x] `servicetalk-http-security-jersey` #1659
- [x] `servicetalk-http-utils` #1659
- [X] `servicetalk-loadbalancer`#1762
- [x] `servicetalk-log4j2-mdc-utils` #1664
[...]

PR #1605, opened 2021-06-02
Migrate servicetalk-opentracing-log4j2 tests to JUnit 5 (#1568)

kashike:
Motivation:

- JUnit 5 leverages features from Java 8 or later, such as lambda functions, making tests more powerful and easier to maintain.
- JUnit 5 has added some very useful new features for describing, organizing, and executing tests. For instance, tests get better display names and can be organized hierarchically.
- JUnit 5 is organized into multiple libraries, so only the features you need are imported into your project. With build systems such as Maven and Gradle, including the right libraries is easy.
- JUnit 5 can use more than one extension at a time, which JUnit 4 could not (only one runner could be used at a time). This means you can easily combine the Spring extension with other extensions (such as your own custom extension).

Modifications:

- Unit tests have been migrated from JUnit 4 to JUnit 5

Result:

Module servicetalk-opentracing-log4j2 now runs tests using JUnit 5

### DIFF
--- a/servicetalk-opentracing-log4j2/build.gradle
+++ b/servicetalk-opentracing-log4j2/build.gradle
@@ -28,8 +28,10 @@ dependencies {
   implementation "org.slf4j:slf4j-api:$slf4jVersion"
 
   testImplementation testFixtures(project(":servicetalk-log4j2-mdc-utils"))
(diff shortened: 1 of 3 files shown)


idelpivnitskiy (approved):
🚀

PR #1610, opened 2021-06-04
Prevent memory leak if HTTP/1.1 encoding fails

idelpivnitskiy:
Motivation:

When `HttpObjectEncoder` encodes meta-data or trailers an error may
happen, which will cause a leak of internal `ByteBuf`. It may happen if
any of the encoding object throw or if it tries to expand the size of
that internal `ByteBuf` without success.

Modifications:

- try-catch meta-data and trailers encoding and release the tmp
`ByteBuf` in case of any failures;

Result:

Internal `ByteBuf` does not leak in case of an error in
`HttpObjectEncoder`.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectEncoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectEncoder.java
@@ -130,19 +130,25 @@ abstract class HttpObjectEncoder<T extends HttpMetaData> extends ChannelOutbound
             // to a socket. In order to do the write to the socket the memory typically needs to be allocated in direct
             // memory and will be copied to direct memory if not. Using a direct buffer will avoid the copy.
             ByteBuf byteBuf = ctx.alloc().directBuffer((int) headersEncodedSizeAccumulator);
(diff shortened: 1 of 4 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectEncoder.java
```diff
@@ -258,8 +264,13 @@ private static void encodeChunkedContent(ChannelHandlerContext ctx, Buffer msg,
         if (contentLength > 0) {
             String lengthHex = toHexString(contentLength);
             ByteBuf buf = ctx.alloc().directBuffer(lengthHex.length() + 2);
-            buf.writeCharSequence(lengthHex, US_ASCII);
-            writeShortBE(buf, CRLF_SHORT);
+            try {
+                buf.writeCharSequence(lengthHex, US_ASCII);
+                writeShortBE(buf, CRLF_SHORT);
+            } catch (Exception e) {
```
idelpivnitskiy, line 270: This code is unlikely to fail. Added try-catch for consistency, more safety, and better sleep :)

NiteshKant (approved):
One question but LGTM

PR #1614, opened 2021-06-10
Publisher#flatMapMergeSingle catch an propagate unexpected errors downstream

Scottmitch:
Motivation:
Publisher#flatMapMergeSingle relies upon the upstream source to catch
unexpected exceptions, cleanup necessary state, and propagate an error
back downstream. However in the event the upstream source is a Single it
will not be able to propagate an error downstream because it has already
sent a terminal signal (multiple terminal signals not allowed by the
RS spec). This means the upstream can at best log/swallow the error and
downstream may not be notified of the failure.

Modifications:
- Publisher#flatMapMergeSingle exception handling is modified to instead
  of propagating the exception upstream, exceptions are caught, upstream
  is cancelled, and the exception is propagated downstream.

Result:
Publisher#flatMapMergeSingle usages where downstream operators may throw
will cancel upstream and propagate and error downstream.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
@@ -306,7 +306,7 @@ final class PublisherFlatMapSingle<T, R> extends AbstractAsynchronousPublisherOp
             }
         }
 
-        private void enqueueFailed(Object item) {
+        private static void enqueueFailed(Object item) {
             LOGGER.error("Queue should be unbounded, but an offer failed for item {}!", item);
             // Note that we throw even if the item represents a terminal signal (even though we don't expect another
             // terminal signal to be delivered from the upstream source because we are already terminated). If we fail
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
@@ -44,9 +44,7 @@ import static io.servicetalk.concurrent.internal.SubscriberUtils.checkDuplicateS
 import static io.servicetalk.concurrent.internal.SubscriberUtils.isRequestNValid;
 import static io.servicetalk.concurrent.internal.TerminalNotification.complete;
 import static io.servicetalk.concurrent.internal.TerminalNotification.error;
-import static io.servicetalk.concurrent.internal.ThrowableUtils.catchUnexpected;
 import static io.servicetalk.utils.internal.PlatformDependent.newUnboundedMpscQueue;
-import static io.servicetalk.utils.internal.PlatformDependent.throwException;
 import static java.lang.Math.min;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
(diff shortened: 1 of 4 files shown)


Scottmitch:
Ideally `Publisher#flatMapMerge` and `Publisher#flatMapMergeSingle` remain as consistent as possible. However I think there is a case to be made here for a slight variation due propagating exceptions upstream to `Single` doesn't allow for error propagation. We have cases where `flatMapMerge` error propagation upstream allows for more contextual cleanup (e.g. in `DnsClient` rather than just cancel).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
```diff
@@ -237,15 +235,28 @@ private void onError0(Throwable throwable, boolean cancelUpstream) {
...
-                } finally {
-                    if (!releaseLock(emittingUpdater, this)) {
-                        drainPending();
-                    }
+                } catch (Throwable cause) {
+                    onErrorHoldingLock(cause);
+                    return; // Poison emittingUpdater. We prematurely terminated, other signals should be ignored.
+                }
+                // Release lock after we handle errors, because error handling needs to poison the lock.
+                if (!releaseLock(emittingUpdater, this)) {
```
idelpivnitskiy, line 258: Here is a risk that smth inside `onErrorHoldingLock` may throw and then lock won't be released

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapMergeTest.java
```diff
@@ -102,6 +103,18 @@ static void afterClass() throws Exception {
         }
     }
 
+    @Test
+    void singleToPublisherOnNextErrorPropagated() {
+        toSource(publisher.flatMapMerge(x -> succeeded(x).toPublisher(), 2)
```
idelpivnitskiy, line 108: Consider adding a test for `flatMapMergeDelayError`

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapSingleTest.java
```diff
@@ -85,6 +86,27 @@ static void afterClass() throws Exception {
...
+        onNextErrorPropagated(x -> executor.submit(() -> x));
+    }
+
+    @Test
+    void succeededSingleOnNextErrorPropagated() {
+        onNextErrorPropagated(Single::succeeded);
+    }
+
+    private void onNextErrorPropagated(Function<? super Integer, ? extends Single<? extends Integer>> func) {
+        toSource(source.flatMapMergeSingle(func, 2)
```
idelpivnitskiy, line 100: Consider adding a test for `flatMapMergeSingleDelayError`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
```diff
@@ -261,30 +272,25 @@ private void enqueueAndDrain(Object item) {
...
-            }
-
-            if (delayedCause != null) {
-                throwException(delayedCause);
-            }
-
-            if (drainCount != 0) {
-                doDrainPostProcessing(drainCount);
+                // Release lock after we handle errors, because error handling needs to poison the lock.
+                tryAcquire = !releaseLock(emittingUpdater, this);
```
idelpivnitskiy, line 293: 1. Looks like this `releaseLock` invocation can also be impacted by an exception thrown from `onErrorHoldingLock`.
2. While you are here, please make `enqueueFailed` method below `static`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
Scottmitch: that is intentional. when `onErrorHoldingLock` is called the lock is "poisoned" (aka not unlocked) to prevent future signal delivery.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
Scottmitch: > Looks like this releaseLock invocation can also be impacted by an exception thrown from onErrorHoldingLock.

lets discuss above in https://github.com/apple/servicetalk/pull/1614#discussion_r649654072

> While you are here, please make enqueueFailed method below static.

done

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
idelpivnitskiy: I see, thanks for clarifying!

idelpivnitskiy (approved):
🚀

Scottmitch:

Scottmitch:
another build failure attributed to https://github.com/apple/servicetalk/issues/1579

PR #1615, opened 2021-06-10
Single#concat(Publisher) onNext error propagation

Scottmitch:
Motivation:
Single#concat(Publisher) doesn't catch exceptions from downstream
Subscriber#onNext and relies upon upstream to propagate exceptions.
However upstream is a Single and has already terminated by the time the
exception is caught so it cannot send a duplicate terminal signal.

Modifications:
- Single#concat(Publisher) catches exceptions from downstream
  Subscriber#onNext and propagates an error.

Result:
Single#concat(Publisher) is more robust with respect to exception
handling.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
@@ -152,7 +152,12 @@ final class SingleConcatWithPublisher<T> extends AbstractNoHandleSubscribePublis
         }
 
         private void emitSingleSuccessToTarget(@Nullable final T result) {
-            target.onNext(result);
+            try {
+                target.onNext(result);
+            } catch (Throwable cause) {
+                target.onError(cause);
+                return;
+            }
             next.subscribeInternal(this);
         }
     }
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleConcatWithPublisherTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleConcatWithPublisherTest.java
@@ -15,6 +15,7 @@
  */
 package io.servicetalk.concurrent.api.single;
 
+import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.concurrent.api.TestCancellable;
 import io.servicetalk.concurrent.api.TestPublisher;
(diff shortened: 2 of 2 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
```diff
@@ -152,7 +152,12 @@ public void cancel() {
...
 
         private void emitSingleSuccessToTarget(@Nullable final T result) {
-            target.onNext(result);
+            try {
+                target.onNext(result);
+            } catch (Throwable cause) {
+                target.onError(cause);
+                return;
+            }
             next.subscribeInternal(this);
```
idelpivnitskiy, line 161: If we terminate in case of an error, do we need to subscribe to the `next` publisher? Consider enhancing the test to verify we do not subscribe to the publisher.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
```diff
@@ -152,7 +152,12 @@ public void cancel() {
         }
 
         private void emitSingleSuccessToTarget(@Nullable final T result) {
-            target.onNext(result);
+            try {
+                target.onNext(result);
+            } catch (Throwable cause) {
+                target.onError(cause);
```
idelpivnitskiy, line 158: Do we need to use `SubscriberUtils#safeOnError` here?

idelpivnitskiy:
Nice catch!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
idelpivnitskiy: I missed the `return` statement above but still nice to validate in a test

idelpivnitskiy (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
Scottmitch: upstream should handle exceptions being thrown and won't propagate a duplicate terminal.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
Scottmitch: done

Scottmitch:

Issue #1562, opened 2021-05-14
BufferInputStream should support mark/reset


idelpivnitskiy:
`BufferInputStream` does not implement mark/reset methods of the `InputStream`, while we can. We should implement it to make it close to `ByteArrayInputStream` and `ByteBufInputStream` (from netty) behavior.

SivaTharun:
@idelpivnitskiy can i work on this issue, i see that the class `src/main/java/io/servicetalk/buffer/api/BufferInputStream.java`, has to implement mar and reset methods. can you also add the expectation of the method implementation, what should it essentially do.

idelpivnitskiy:
Thank you for the interest @SivaTharun!
Absolutely, you can take it.
The goal is to implement the standard behavior of `mark/reset` methods defined by Java community in [InputStream javadoc](https://github.com/openjdk/jdk/blob/79b39445f6fcd005744c5de89ba2cd2ed5bc0a54/src/java.base/share/classes/java/io/InputStream.java#L649-L750). You can refer to [java.io.ByteArrayInputStream](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/io/ByteArrayInputStream.java) and [io.netty.buffer.ByteBufInputStream](https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/ByteBufInputStream.java) to get the basic idea of the implementation.
Please, also add tests to verify new behavior.

PR #1616, opened 2021-06-13
[WIP] 1562 - provide functionality for mark and reset method

SivaTharun:
PR for addressing #1562 

- added the mark method and reset method signature
- introduced the bufferedPos, count and mark variables, to keep track of the current buffer position


bondolo (commented):
I don't see where the read position in the buffer is altered in response to a reset or bufferPosition is used to adjust the read position. It seems the mark/reset alters the number of bytes available to read but not what is read.

tkountis:
@SivaTharun any updates on this work, let us know if we can help by providing hints/direction as needed.

bondolo:
This PR will be closed as inactive if no progress/response is made within the next week.

PR #1620, opened 2021-06-16
run GitHub release action publish task without parallel

bondolo:
Motivation:
The sonatype repository appears to potentially exhibit a race condition when uploading
artifacts that causes multiple staging repositories to be created.
Modifications:
This change hopefully avoids the apparent race condition by executing publish tasks
serially.
Result:
Successful sonatype upload.

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -43,7 +43,7 @@ jobs:
           # Build arguments to feed to the single gradlew publish command
           if [ "${{ matrix.java }}" = "8" ]; then
             FIRST_GRADLE_TARGETS=" clean check"
-            SECOND_GRADLE_TARGETS=" publish"
+            SECOND_GRADLE_TARGETS=" --no-parallel publish"
           else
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-release.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           # Build arguments to feed to the single gradlew publish command
           if [ "${{ matrix.java }}" = "8" ]; then
             FIRST_GRADLE_TARGETS=" clean check"
-            SECOND_GRADLE_TARGETS=" publish"
+            SECOND_GRADLE_TARGETS=" --no-parallel publish"
```
idelpivnitskiy, line 46: I think you also need to remove the parallel flag from `gradle.properties` file

idelpivnitskiy: or does the explicit flag override it?

bondolo: The explicit command line flag takes precedence. I checked with local build.

tkountis (approved):

PR #1622, opened 2021-06-17
Remove fields from Single

bondolo:
`Single` has two fields, `executor` and `shareContextOnSubscribe` that provide
limited value because the contents are unused in all but a few operators. Eliminating
them will reduce allocation and eliminate the held references which could be preventing
GC. This change is also an incremental step in simplifying offloading as `Executor` state
will progressively be reduced to only operators that involve offloading.
Modifications:
Fields are removed from `Single` base and defined locally in sub-classes which have
actual values for the fields.
Result:
Lower allocation, fewer unused local fields.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnSingles.java
```diff
@@ -40,38 +41,77 @@ private PublishAndSubscribeOnSingles() {
     }
 
     static <T> Single<T> publishAndSubscribeOn(Single<T> original, Executor executor) {
-        return original.executor() == executor ? original : new PublishAndSubscribeOn<>(executor, original);
+        return original.executor() == executor || immediate() == executor ?
```
Scottmitch, line 44: IIUC this will eventually go away as `original.executor()` will no longer exist?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -2599,10 +2575,20 @@ void handleSubscribe(Subscriber<? super T> subscriber, SignalOffloader signalOff
     /**
      * Returns the {@link Executor} used for this {@link Single}.
      *
-     * @return {@link Executor} used for this {@link Single} via {@link #Single(Executor)}.
+     * @return {@link Executor} used for this {@link Single}.
+     */
+    Executor executor() {
```
Scottmitch, line 2580: inline with above comments, this will eventually go away too IIUC?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnSingles.java
bondolo: Correct, the `immediate() == executor` part might remain.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
bondolo: Yes. once the SignalOffloader creation is removed from `subscribeWithContext` this can go away.

Scottmitch (approved):

PR #1623, opened 2021-06-17
Publisher remove fields

bondolo:
Motivation:
`Publisher` has two fields, `executor` and `shareContextOnSubscribe` that provide
limited value because the contents are unused in all but a few operators. Eliminating
them will reduce allocation and eliminate the held references which could be preventing
GC. This change is also an incremental step in simplifying offloading as `Executor` state
will progressively be reduced to only operators that involve offloading.

Modifications:
Fields are removed from `Publisher` base and defined locally in sub-classes which have
actual values for the fields.

Result:
Lower allocation, fewer unused local fields.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AbstractAsynchronousPublisherOperator.java
```diff
@@ -35,11 +35,15 @@
...
     private final Publisher<T> original;
 
-    AbstractAsynchronousPublisherOperator(Publisher<T> original, Executor executor) {
-        super(executor);
+    AbstractAsynchronousPublisherOperator(Publisher<T> original) {
         this.original = requireNonNull(original);
     }
 
+    @Override
+    Executor executor() {
```
Scottmitch, line 43: (for visibility): the plan is to eventually remove the `executor()` methods in subsequent PRs, correct?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
```diff
@@ -73,6 +72,11 @@
         this.restartAtOnNext = restartAtOnNext;
     }
 
+    @Override
```
Scottmitch, line 75: the timeout operators are exceptional in that they require an executor to execute the timeout callback, so these operators will need to retain an `Executor`, but they should be able todo that as `private final` fields. fine to defer to followup PR if that is where you intended to do more cleanup.

Scottmitch (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AbstractAsynchronousPublisherOperator.java
bondolo: Correct.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
bondolo: Yes, this will need to be cleaned up when `SignalOffloader` is removed.

PR #1626, opened 2021-06-18
Handle reentry for n-arity Publisher.from

tkountis:
Motivation:
FromPublisher(s) with n-arity have no safety net for re-entrance. This can result in emitting duplicate items.

Modifications:
Handle re-entry.

Result:
No duplicates.


idelpivnitskiy (changes requested):
Great catch! 🎖️

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/From3Publisher.java
```diff
@@ -74,60 +78,34 @@ public void request(final long n) {
...
-            } else if (state == DELIVERED_V1) {
-                if (n > 1) {
-                    deliverV2V3();
+            if (requested == 3) {
+                return;
+            }
+            requested = min(3, addWithOverflowProtection(requested, n));
+            while (delivered < requested) {
+                delivered++;
+                if (state == INIT) {
```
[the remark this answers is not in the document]
tkountis: I was also considering removing it, but then again 2-3 arguments will be the most common use-case so its probably worth the special version for it. Haven't done any perf testing though to see if the benefits are even visible.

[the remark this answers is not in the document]
tkountis: I run a quick JMH on the creation of the publisher, and the GC pressure is quite visible under the extremes of a benchmark. Still I think there is visible benefit for var-args < 5 arguments so I am leaning towards keeping the 3 args as well.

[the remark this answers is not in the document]
bondolo: Thinking about it from the other direction, if only `FromArrayPublisher` existed how likely would it be that it would appear as the bottleneck in someone's flame graph? We know that `FromXPublisher`s for X < 5 **could** be faster but are we sure it would make any difference outside of microbenchmarks?

An alternative implementation could also increase code reuse by abusing the initialization conditions so that From2Publisher and From3Publisher used the same code but the from-2 pattern initialized it such that value 1 was "already delivered".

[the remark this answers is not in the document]
tkountis: I generally agree with this line of thinking, and most of the times its not worth the effort and the duplication (potential more space for bugs). 

Having said that, _playing the devils advocate_, people do run benchmarks and they do compare us with alternative products, so its always good to not have to justify why we are generating more garbage. 
Also, outside the microbenchmarking world, I do thing generating less garbage is a good mental model, and in this case the gc rate was many folds bigger (10x); so if you use `Publisher#from` directly or indirectly (through zip) in your hot-path, you do pay for that additional cost.

clarification: there is no argument about speed here, its just gc pressure.
WDYT?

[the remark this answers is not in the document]
idelpivnitskiy: We use `From3Publisher` in two places: `zip` operator for 3 singles and `HeaderUtils#setContentLength` to flatten the HTTP message. Let's keep this optimization. However, I like the suggestion to share code between `From2Publisher` and `From3Publisher`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/From3Publisher.java
```diff
@@ -45,12 +47,9 @@ void doSubscribe(final Subscriber<? super T> subscriber) {
     }
 
     private final class ThreeValueSubscription implements Subscription {
-        private static final byte INIT = 0;
-        private static final byte DELIVERED_V1 = 1;
-        private static final byte DELIVERED_V2 = 2;
-        private static final byte CANCELLED = 3;
-        private static final byte TERMINATED = 4;
-        private byte state;
+        private final int TERMINATED = -1;
```
idelpivnitskiy, line 50: Minor suggestions:
```patch
Index: servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/From3Publisher.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/From3Publisher.java b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/From3Publisher.java
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/From3Publisher.java	(revision ef7b3c5da3192be0fe61308121f0d46122e2de89)
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/From3Publisher.java	(date 1624337438778)
@@ -47,8 +47,8 @@
     }
 
     private final class ThreeValueSubscription implements Subscription {
-        private final int TERMINATED = -1;
-        private long requested;
+        private static final int TERMINATED = 3;
+        private int requested;
         private int state;
         private final Subscriber<? super T> subscriber;
 
@@ -71,36 +71,30 @@
                 subscriber.onError(newExceptionForInvalidRequestN(n));
                 return;
             }
-            if (requested == 3) {
-                return;
-            }
-            requested = min(3, addWithOverflowProtection(requested, n));
+            requested = (int) min(TERMINATED, addWithOverflowProtection(requested, n));
             while (state < requested) {
                 if (state == 0) {
-                    deliver(v1, 1);
+                    deliver(v1);
                 } else if (state == 1) {
-                    deliver(v2, 2);
+                    deliver(v2);
                 } else if (state == 2) {
-                    if (deliver(v3, 3)) {
+                    if (deliver(v3)) {
                         subscriber.onComplete();
                     }
-                } else {
-                    // Cancelled
-                    break;
                 }
             }
         }
 
-        private boolean deliver(@Nullable T value, int newState) {
-            state = newState;
+        private boolean deliver(@Nullable T value) {
+            ++state;
             try {
                 subscriber.onNext(value);
                 return true;
             } catch (Throwable cause) {
                 state = TERMINATED;
                 subscriber.onError(cause);
-            }
-            return false;
+            }
         }
     }
 }
Index: servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/From3PublisherTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/From3PublisherTest.java b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/From3PublisherTest.java
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/From3PublisherTest.java	(revision ef7b3c5da3192be0fe61308121f0d46122e2de89)
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/From3PublisherTest.java	(date 1624334022108)
@@ -180,6 +180,11 @@
         reentry(2, false, 1);
     }
 
+    @Test
+    void reentryOneFirstNoReentry() {
+        reentry(1, false, 2);
+    }
+
     @Test
     void reentryMaxFirst() {
         reentry(MAX_VALUE, true, 1);
@@ -190,6 +195,16 @@
         reentry(1, true, MAX_VALUE);
     }
 
+    @Test
+    void reentryMaxFirstNoReentry() {
+        reentry(MAX_VALUE, false, 1);
+    }
+
+    @Test
+    void reentryMaxLaterNoReentry() {
+        reentry(1, false, MAX_VALUE);
+    }
+
     void reentry(final long firstDemand, final boolean nextWithReentry, final long nextDemand) {
         final int[] emitted = {0};
         final boolean[] completed = {false};
```

idelpivnitskiy (commented):
+1 to @bondolo's idea to reuse `From3Publisher` for 2 elements, it will need just to start with `state = 1`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/From3Publisher.java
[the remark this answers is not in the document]
tkountis: Merged the two publishes in one. 👍

tkountis:
Build failure https://github.com/apple/servicetalk/issues/1630

idelpivnitskiy (approved):
Less code 🎉 
LGTM! I have an optional suggestion, but up to you which approach to take. Everything works great

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromNPublisher.java
```diff
@@ -17,47 +17,58 @@
...
         private final Subscriber<? super T> subscriber;
 
-        private TwoValueSubscription(final Subscriber<? super T> subscriber) {
+        private NValueSubscription(final Subscriber<? super T> subscriber) {
             this.subscriber = subscriber;
         }
 
         @Override
         public void cancel() {
-            if (state != TERMINATED) {
```
Scottmitch, line 58: iirc this check was added so that terminal state could be propagated regardless of cancellation or not (e.g. an error).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromNPublisher.java
```diff
@@ -17,47 +17,58 @@
...
+import static java.lang.Math.min;
 
-final class From2Publisher<T> extends AbstractSynchronousPublisher<T> {
+final class FromNPublisher<T> extends AbstractSynchronousPublisher<T> {
     @Nullable
     private final T v1;
     @Nullable
     private final T v2;
+    @Nullable
+    private final T v3;
```
Scottmitch, line 32: The two independent classes were introduced to reduce memory consumption as they are commonly used cases. After the classes are combined are there any benefit relative to `FromArrayPublisher` anymore? Looks like the `From2` will use more memory. Can you fix the issues in the same code structure and explore combining memory impact as followup PR?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromNPublisher.java
idelpivnitskiy: this is only valid for handling invalid `request(n)`. the spec says that after cancelation it should be [nop](https://github.com/reactive-streams/reactive-streams-jvm#3.6). validation of `N` is not required at this point. hence, better to minimize number of states

tkountis: I believe @Scottmitch is referring to this case:

```
void testErrorWhenCancelInterleavesDemandCompletion() {
        final int[] emitted = {0};
        final boolean[] completed = {false};
        final boolean[] error = {false};
        final Subscription[] subscription = new Subscription[1];
        toSource(fromPublisher()).subscribe(new Subscriber<Integer>() {
            @Override
            public void onSubscribe(final Subscription subscription1) {
                subscription[0] = subscription1;
                subscription1.request(3);
            }

            @Override
            public void onNext(@Nullable final Integer integer) {
                subscription[0].cancel();
                if (requireNonNull(integer) == 2) {
                    throw new IllegalStateException();
                } else {
                    emitted[0]++;
                }
            }

            @Override
                error[0] = true;
            }

            @Override
            public void onComplete() {
                completed[0] = true;
            }
        });

        assertThat(emitted[0], is(1));
        assertThat(completed[0], is(false));
        assertThat(error[0], is(true));
    }
```

Allowing the terminal signal to be propagated, especially when this can happen async. The spec says that the `cancel` could be eventual. I think the spec allows both behaviours, but allowing an error to be delivered I think is preferable. 
Thoughts?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromNPublisher.java
tkountis: @Scottmitch I did some changes, share your thoughts. 

**Before** 

```
io.servicetalk.concurrent.api.From2Publisher object internals:
OFF  SZ   TYPE DESCRIPTION               VALUE
  0   8        (object header: mark)     N/A
  8   4        (object header: class)    N/A
 12   4      T From2Publisher.v1         N/A
 16   4      T From2Publisher.v2         N/A
 20   4        (object alignment gap)    
Instance size: 24 bytes
Space losses: 0 bytes internal + 4 bytes external = 4 bytes total
```

```
io.servicetalk.concurrent.api.From2Publisher.TwoValueSubscription object internals:
OFF  SZ                    TYPE DESCRIPTION                       VALUE
  0   8                         (object header: mark)             N/A
  8   4                         (object header: class)            N/A
 12   1                    byte TwoValueSubscription.state        N/A
 13   3                         (alignment/padding gap)           
 16   4   Subscriber<? super T> TwoValueSubscription.subscriber   N/A
 20   4                         (object alignment gap)            
Instance size: 24 bytes
Space losses: 3 bytes internal + 4 bytes external = 7 bytes total
```

------

```
io.servicetalk.concurrent.api.From3Publisher object internals:
OFF  SZ   TYPE DESCRIPTION               VALUE
  0   8        (object header: mark)     N/A
  8   4        (object header: class)    N/A
 12   4      T From3Publisher.v1         N/A
 16   4      T From3Publisher.v2         N/A
 20   4      T From3Publisher.v3         N/A
Instance size: 24 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```

```
io.servicetalk.concurrent.api.From3Publisher.ThreeValueSubscription object internals:
OFF  SZ                    TYPE DESCRIPTION                         VALUE
  0   8                         (object header: mark)               N/A
  8   4                         (object header: class)              N/A
 12   1                    byte ThreeValueSubscription.state        N/A
 13   3                         (alignment/padding gap)             
 16   4   Subscriber<? super T> ThreeValueSubscription.subscriber   N/A
 20   4                         (object alignment gap)              
Instance size: 24 bytes
Space losses: 3 bytes internal + 4 bytes external = 7 bytes total
```

-----

```
io.servicetalk.concurrent.api.FromArrayPublisher object internals:
OFF  SZ   TYPE DESCRIPTION                 VALUE
  0   8        (object header: mark)       N/A
  8   4        (object header: class)      N/A
 12   4    T[] FromArrayPublisher.values   N/A
Instance size: 16 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```

```
io.servicetalk.concurrent.api.FromArrayPublisher.FromArraySubscription object internals:
OFF  SZ                    TYPE DESCRIPTION                            VALUE
  0   8                         (object header: mark)                  N/A
  8   4                         (object header: class)                 N/A
 12   4                     int FromArraySubscription.beginOffset      N/A
 16   4                     int FromArraySubscription.endOffset        N/A
 20   1                 boolean FromArraySubscription.ignoreRequests   N/A
 21   3                         (alignment/padding gap)                
 24   4   Subscriber<? super T> FromArraySubscription.subscriber       N/A
 28   4                     T[] FromArraySubscription.values           N/A
Instance size: 32 bytes
Space losses: 3 bytes internal + 0 bytes external = 3 bytes total
```
-----

**After**

```
io.servicetalk.concurrent.api.FromNPublisher object internals:
OFF  SZ   TYPE DESCRIPTION               VALUE
  0   8        (object header: mark)     N/A
  8   4        (object header: class)    N/A
 12   4      T FromNPublisher.v1         N/A
 16   4      T FromNPublisher.v2         N/A
 20   4      T FromNPublisher.v3         N/A
Instance size: 24 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```

```
io.servicetalk.concurrent.api.FromNPublisher object internals:
OFF  SZ   TYPE DESCRIPTION               VALUE
  0   8        (object header: mark)     N/A
  8   4        (object header: class)    N/A
 12   4      T FromNPublisher.v1         N/A
 16   4      T FromNPublisher.v2         N/A
 20   4      T FromNPublisher.v3         N/A
Instance size: 24 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```

----

With that in mind, before & after have no difference in heap cost between them. The new version still saves 8bytes per subscription instance, but if to fwiw it doesn't justify the loss of code readability. Last, the n-arity operator performs better against the array version when it comes to allocation rate.

Scottmitch: thanks for providing this analysis. as long as no memory increase lgtm!

PR #1629, opened 2021-06-21
Load balancer should not close existing connections

chemicL:
Motivation:

When dealing with DNS based discovery, the RoundRobinLoadBalancer
eagerly closes connections to hosts not returned in the DNS response. It
makes sense to keep established connections open until the server
decides to close them, but allow traffic shifting by no opening new
connections to unavailable hosts. This change allows configuring the
RoundRobinLoadBalancer to choose the desired behaviour.

Modifications:

* Added a boolean flag to RoundRobinLoadBalancer that allows treating
  unavailable hosts as expired and continue using its' connections.
* Created a Builder for RoundRobinLoadBalancerFactory that allows
  configuring the RoundRobinLoadBalancer with the new flag.
* Moved `RoundRobinLoadBalancerFactory` to the top level class.
* Deprecated `RoundRobinLoadBalancer` and
`RoundRobinLoadBalancer.RoundRobinLoadBalancerFactory`.
* Enhanced `RoundRobinLoadBalancer` to do deduplication of addresses.
* Added tests to verify new behavior.

Result:

It is possible to either maintain existing behaviour of
RoundRobinLoadBalancer's flag or to keep connections open to unavailable
hosts.


tkountis (commented):
good work @chemicL 🎉 
Skimmed through the PR, I still need to have a deeper look.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -202,6 +223,42 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
                 }
             }
 
+            @SuppressWarnings("unchecked")
+            private Host<ResolvedAddress, C> createHost(ResolvedAddress addr) {
+                Host<ResolvedAddress, C> host = new Host<>(addr);
+                host.onClose().afterFinally(() ->
```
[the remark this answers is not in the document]
chemicL: Is that indeed so? `ListenableAsyncCloseable#onClose` javadoc:

```
* Returns a {@link Completable} that is notified once the {@link ListenableAsyncCloseable} was closed.
     * @return the {@code Completable} that is notified on close.
```

My intention was the following:

When the last connection closes, the host is closed (line 513).
Following that, upon closing of the host, the host is removed from the array of active hosts. It needs to be added here because of the visibility of the `activeHostsUpdater` and the cohesion – the LB creates the `Host`s and allows interactions with its own internal state by adding this hook.

Please let me know if that clarifies it. The tests confirm the behaviour is as expected.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -164,29 +187,28 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
-                                final Host<ResolvedAddress, C> host = oldHostsTyped.get(i);
-                                if (host.address.equals(addr)) {
-                                    host.markInactive();
-                                    for (int x = i + 1; x < oldHostsTyped.size(); ++x) {
-                                        newHosts.add(oldHostsTyped.get(x));
+                            if (eagerConnectionShutdown) {
+                                return removeFromHostsList(oldHosts, addr);
+                            } else {
+                                for (int i = 0; i < oldHostsTyped.size(); ++i) {
+                                    final Host<ResolvedAddress, C> host = oldHostsTyped.get(i);
```
bondolo, line 204: Strange to see an old style indexed for-loop used for this.

chemicL: The pattern from the above condition was replicated, but whilst it made sense above, here it's actually pointless. I replaced it with an enhanced for loop.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -391,6 +506,11 @@ boolean addConnection(C connection) {
                     }
                     if (i == existing.length) {
                         break;
+                    } else if (existing.length == 1 && isExpired) {
+                        // We're closing the last connection, close the Host.
+                        // Closing the host will trigger the Host's onClose method,
+                        // which will remove the host from active hosts list.
+                        Host.this.closeAsyncGracefully().subscribe();
```
idelpivnitskiy, line 513: I see that `doClose` will set `connections` to `CLOSED_ARRAY`. However, it looks like here can be a race between reading `isExpired` and updating `isExpired` + creating a new connection. Should we try to CAS `existing` with `CLOSED_ARRAY` before closing the host, WDYT?

If we set `CLOSED_ARRAY`, it guarantees no more connections can be added for this host. Therefore, we don't need to close the host gracefully because it's known that there are no connections left.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -337,16 +437,31 @@ public Completable closeAsyncGracefully() {
...
+                public Completable closeAsync() {
+                    return doClose(AsyncCloseable::closeAsync);
+                }
+
+                @Override
+                public Completable closeAsyncGracefully() {
+                    return doClose(AsyncCloseable::closeAsyncGracefully);
+                }
+            });
+            this.closeable = toListenableAsyncCloseable(ac);
```
idelpivnitskiy, line 464: You can replace this whole thing with:
```java
this.closeable = toAsyncCloseable(graceful -> graceful ? doClose(AsyncCloseable::closeAsyncGracefully) :
        doClose(AsyncCloseable::closeAsync));
```
It will be consistent with how `asyncCloseable` is initialized at the end of RRLB ctor.

idelpivnitskiy (changes requested):

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -202,6 +223,42 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
+                return host;
+            }
+
+            private List<Host<ResolvedAddress, C>> removeFromHostsList(
+                    List<Host<ResolvedAddress, C>> oldHosts, ResolvedAddress addr) {
+                @SuppressWarnings("unchecked")
+                final List<Host<ResolvedAddress, C>> oldHostsTyped = (List<Host<ResolvedAddress, C>>) oldHosts;
+                if (oldHostsTyped.isEmpty()) {
+                    // this can happen when an expired host is removed,
+                    // but all of its connections have already been closed
```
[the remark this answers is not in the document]
chemicL: That is true. The code here could be used to reflect a future scenario where `expired` events can be emitted from SD. Followed by an unavailable event which would actually mean removal, which in such case would cause entering this check. For now, this is just a double check. I will update the comment as recommended.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: As a matter of fact, the `doClose` method does the CAS and is called from `closeAsyncGracefully`. The purpose of closing the `Host` here is triggering the callback which will remove it from the `activeHosts` list. So we achieve two purposes at once.

The race condition that's worth considering is when a connection is currently created (c1) (let's say just a moment before `unavailable` event arrives) and the Host has just one connection (c0) open:
1. t0: Host is available, connection c1 is being created.
2. t1: SD sends unavailable event, host is marked as expired.
3. t2: Remote server closes existing connections, last one (c0) triggers closing the Host.
4. t3: New connection, a successfully established one (c1) is added to the `connections` array.

If t2 happens before t3: `addConnection` uses CAS for adding the connection and will spot if the array is already closed, rejecting the connection.
If t3 happens before t2: c0's close callback is triggered and can either see `existing.length == 1` or `existing.length == 2`. If it's `2`, nothing bad will happen. If it's `1`, `doClose` will kill the newly created connection (c1) and the Host will be removed from `activeHosts`.

In this scenario doing the CAS to `CLOSED_ARRAY` will allow us to see that in fact c0 is not the last connection, do another loop and simply remove c0 from the array.

Got it, the updated code is the following:

```
                    } else if (existing.length == 1 && isExpired) {
                        // We're closing the last connection, close the Host.
                        // Closing the host will trigger the Host's onClose method,
                        // which will remove the host from active hosts list.
                        // If a race condition appears and a new connection was added in the meantime,
                        // the CAS operation will allow for determining that and not closing the Host yet and removing
                        // the previously last connection from the array.
                        if (connectionsUpdater.compareAndSet(this, existing, CLOSED_ARRAY)) {
                            Host.this.closeAsyncGracefully().subscribe();
                        }
                    }
```

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -164,29 +187,27 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
-                                if (host.address.equals(addr)) {
-                                    host.markInactive();
-                                    for (int x = i + 1; x < oldHostsTyped.size(); ++x) {
-                                        newHosts.add(oldHostsTyped.get(x));
+                            if (eagerConnectionShutdown) {
+                                return removeFromHostsList(oldHosts, addr);
+                            } else {
+                                for (Host<ResolvedAddress, C> host : oldHostsTyped) {
+                                    if (host.address.equals(addr)) {
+                                        host.isExpired = true;
```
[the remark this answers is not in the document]
chemicL: One interesting feature/bug of the current implementation of the RRLB can be observed in the tests in the latest commit addressing the issue. Please check a new test I added. The behaviour is just consistent with the flaws of allowing duplicates and I would not address the issue in this PR just yet. But for a non-duplicate scenario, we will be removing such zombies, thanks for spotting it!

chemicL:
@idelpivnitskiy I refactored the Factory and Builder as recommended and deprecated the public constructor of RRLB together with the static method for creating the factory. Please re-review after all the changes :-)

tkountis (approved):
Changes LGTM 🚀 🚀

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/EagerRoundRobinLoadBalancerTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        serviceDiscoveryPublisher.onError(DELIBERATE_EXCEPTION);
+        assertAddresses(lb.activeAddresses(), "address-1");
+    }
+
+    @Test
+    public void hostDownGracefulCloseConnection() throws Exception {
+        sendServiceDiscoveryEvents(upEvent("address-1"));
+        TestLoadBalancedConnection conn = lb.selectConnection(any()).toFuture().get();
+        sendServiceDiscoveryEvents(downEvent("address-1"));
```
idelpivnitskiy, line 84: Consider adding tests that verify:
1. If `address-1` has 2+ connections, all will be closed on "unavailable" event.
2. If RRLB has a few addresses, only connections related to the "unavailable" address close, other remain open.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -79,6 +79,12 @@
  * This can lead to situations where connections will be used to their maximum capacity (for example in the context of
  * pipelining) before new connections are created.</li>
  * <li>Closed connections are automatically pruned.</li>
+ * <li>By default, connections to addresses marked as {@link ServiceDiscovererEvent#isAvailable() unavailable}
+ * are used for requests, but no new connections are created for them. In case the address' connections are busy,
+ * another host is tried. If all hosts are busy, selection fails with a {@link ConnectionRejectedException}.
+ * If {@link RoundRobinLoadBalancer#RoundRobinLoadBalancer(Publisher, ConnectionFactory, boolean)}
```
idelpivnitskiy, line 85: Avoid referencing pkg-private ctor

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -130,9 +136,28 @@
      *
      * @param eventPublisher    provides a stream of addresses to connect to.
      * @param connectionFactory a function which creates new connections.
+     * @deprecated Use {@link io.servicetalk.loadbalancer.RoundRobinLoadBalancerFactory} to build instances
+     * of {@link RoundRobinLoadBalancer}.
```
idelpivnitskiy, line 140: Avoid referencing class that will be deprecated

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -320,15 +406,18 @@ public Completable closeAsyncGracefully() {
      *
      * @param <ResolvedAddress> The resolved address type.
      * @param <C> The type of connection.
+     * @deprecated Use {@link io.servicetalk.loadbalancer.RoundRobinLoadBalancerFactory} to build instances
+     * of {@link RoundRobinLoadBalancer}.
```
idelpivnitskiy, line 410: Avoid referencing class that will be deprecated

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -164,29 +189,30 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
-                                    for (int x = i + 1; x < oldHostsTyped.size(); ++x) {
-                                        newHosts.add(oldHostsTyped.get(x));
+                            if (eagerConnectionShutdown) {
+                                return removeFromHostsList(oldHosts, addr);
+                            } else {
+                                for (Host<ResolvedAddress, C> host : oldHostsTyped) {
+                                    if (host.address.equals(addr)) {
+                                        host.isExpired = true;
+                                        if (host.connections.length == 0) {
+                                            host.closeAsyncGracefully().subscribe();
```
idelpivnitskiy, line 209: I think this is still racy. Because there is a gap between checking the length and swapping the `connections` with `CLOSED_ARRAY` inside `doClose`, another thread that creates a new connection can add it to the `connections` array and return `true` from `addConnection`. It will mean that the connection is selected for request execution. Due to offloading, `closeAsyncGracefully()` may close the connection before the meta-data are taken by event-loop, and the state is updated inside `RequestResponseCloseHandler`. As the result, request will fail because connection is already closed.

Consider holding a ref to `host.connections` in a local variable. Then check the length and if it's 0, try to CAS with `CLOSED_ARRAY`. Do `closeAsyncGracefully()` only if CAS succeeded. Actually, we can do `closeAsync()` in this case.

Another approach can be to CAS `EMPTY_ARRAY` with `CLOSED_ARRAY`, without length check. But it will require changing `connection.onClose().beforeFinally` to always set connections to  `EMPTY_ARRAY` if it handles closure of the last connection when the host is not expired.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -391,6 +485,16 @@ boolean addConnection(C connection) {
...
                         break;
+                    } else if (existing.length == 1 && isExpired) {
+                        // We're closing the last connection, close the Host.
+                        // Closing the host will trigger the Host's onClose method,
+                        // which will remove the host from active hosts list.
+                        // If a race condition appears and a new connection was added in the meantime,
+                        // the CAS operation will allow for determining that and not closing the Host yet and removing
+                        // the previously last connection from the array.
+                        if (connectionsUpdater.compareAndSet(this, existing, CLOSED_ARRAY)) {
+                            Host.this.closeAsyncGracefully().subscribe();
```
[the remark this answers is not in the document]
idelpivnitskiy: Because `connections` array is already set to `CLOSED_ARRAY`, we can close non-gracefully. WDYT?

idelpivnitskiy (changes requested):
Great progress!
Only one concern about a possible race, all other comments are minor suggestions:

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
```diff
@@ -491,11 +478,12 @@ private TestLoadBalancedConnection newConnection(final String address) {
         };
     }
 
-    private interface TestLoadBalancedConnection extends ListenableAsyncCloseable, LoadBalancedConnection {
+    protected interface TestLoadBalancedConnection extends ListenableAsyncCloseable, LoadBalancedConnection {
```
idelpivnitskiy, line 481: `ListenableAsyncCloseable` is inherited from `LoadBalancedConnection`

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -0,0 +1,104 @@
...
+        public Builder() {
+        }
+
+        /**
+         * Configures the {@link RoundRobinLoadBalancerFactory} to produce a {@link LoadBalancer} with
+         * a setting driving eagerness of connection shutdown. By default, the created {@link LoadBalancer} does
+         * not close connections when a host become {@link ServiceDiscovererEvent#isAvailable() unavailable}.
+         * When configured with {@code true} as an argument, the connections will be closed gracefully on such event.
+         *
+         * @param eagerConnectionShutdown when true, connections will be shut down upon receiving
```
idelpivnitskiy, line 85: `true` -> `{@code true}`

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/LingeringRoundRobinLoadBalancerTest.java
```diff
@@ -0,0 +1,166 @@
...
+        lb.selectConnection(connectionFilter).toFuture().get();
+
+        assertAddresses(lb.activeAddresses(), "address-1", "address-2");
+
+        sendServiceDiscoveryEvents(downEvent("address-3"));
+        assertAddresses(lb.activeAddresses(), "address-1", "address-2");
+
+        sendServiceDiscoveryEvents(upEvent("address-1"));
+        // At this moment in time, duplicates are allowed and the down event removes just the first address
+        assertAddresses(lb.activeAddresses(), "address-1", "address-1", "address-2");
```
idelpivnitskiy, line 104: Sure, the duplicates are allowed at this moment. However, there is a difference between seeing two available events in a row for the same address and going back and forth between states ("available" -> "unavailable" -> "available") for the same address.

`DefaultServiceDiscoveryRetryStrategy` guarantees that it won't propagate duplicates, but the status change is not considered as a duplicate. Next "available" event should clear `isExpired` flag and make this connection availalbe again. Only if an address is already "availalbe", the next "available" event can add a new `Host` entry. Same for two "unavailable" events in a row:
```
upEvent("address-1")
// open connection
upEvent("address-1")
// open connection
downEvent("address-1")
downEvent("address-1")
``` 
^ in this case both have to be marked as "expired". The next `upEvent("address-1")` will clear `isExpired` flag from the first host only.

idelpivnitskiy (changes requested):
Sorry, somehow missed the duplicate problem in the previous reviews. Everything else looks good to me.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -391,6 +484,16 @@ boolean addConnection(C connection) {
...
                     if (i == existing.length) {
                         break;
+                    } else if (existing.length == 1 && isExpired) {
+                        // We're closing the last connection, close the Host.
+                        // Closing the host will trigger the Host's onClose method,
+                        // which will remove the host from active hosts list.
+                        // If a race condition appears and a new connection was added in the meantime,
+                        // the CAS operation will allow for determining that and not closing the Host yet and removing
+                        // the connection, previously considered as the last one, from the array.
+                        if (connectionsUpdater.compareAndSet(this, existing, CLOSED_ARRAY)) {
```
idelpivnitskiy, line 494: Actually, I was thinking about this race one more time. Let's draw an example:
```
thread-a: close the last connection for this expired host, we go inside `existing.length == 1 && isExpired` branch
sd-thread: mark this host as available
thread-b: select this host to open a new connection
thread-a: swaps the "connections" array with CLOSED_ARRAY, removes this host from the list of active hosts
thread-b: fails to add a new connection because the host is already closed, closes the new connection
```
I'm ok with the `thread-b` closing the newly created connection. But it doesn't look good if we remove the host from the list of active hosts that recently became available 🤔 
WDYT if instead of a boolean `isExpired` we will have an enum `state` variable with 3 possible values: ACTIVE, EXPIRED, CLOSING?
Your previous `if` check will change to:
```java
...
} else if (existing.length == 1) {
    if (stateUpdater.compareAndSet(this, EXPIRED, CLOSING)) {
        this.closeAsync().subscribe();
        break;
    }
    if (connectionsUpdater.compareAndSet(this, existing, EMPTY_ARRAY)) {
        break;
    }
} else {
...
```
It will help to avoid a race between `thread-a` and `sd-thread`. `sd-thread` will be able to see that the current host is closing and will add a new entry for this address. The old host entry will be removed by `thread-a`. The internal state won't have duplicates.

chemicL: I see your point, thank you for bringing this up. While addressing your concern about duplicates I tried to encapsulate Host's state a bit and that led me to come up with an idea for avoiding the state machine. Possibly the state machine would be helpful if the SD events were not serialised, but as I understand they currently are, which means we can use that assumption to our advantage. I came up with with a guarding mechanism in the form of a new method called `tryToMarkActive`. It uses CAS on the `connectionsUpdater` to determine CLOSING/CLOSED state and will instruct the caller to add a new host when that's the case. New commit coming right up, please have a look @idelpivnitskiy.

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/LingeringRoundRobinLoadBalancerTest.java
chemicL: That's a great point. I'm shocked to have missed it. The next commit covers this case and simplifies the branching logic of handling SD events too.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
idelpivnitskiy: > if the SD events were not serialised, but as I understand they currently are, which means we can use that assumption to our advantage

Sure, SD-events are emitted serially. The problem is that there is a race between thread that closes the connection and a thread that emits a new SD-event. You can discard `thread-b` from my example above to simplify it.

Doesn't look like current variant of `tryToMarkActive()` helps because it can be executed on `sd-thread` in between `existing.length == 1 && isExpired` (line 526) and `connectionsUpdater.compareAndSet(this, existing, CLOSED_ARRAY)` (line 533). I've added a test, see `closureOfLastConnectionDoesntRaceWithNewAvailableEvent()`. But because it's a concurrency test, it does not fail all the time. Use "repeat until failure" feature in Intellij IDEA. For me it fails within ~10k iterations.

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/LingeringRoundRobinLoadBalancerTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+        sendServiceDiscoveryEvents(downEvent("address-1"));
+        // Because the first address (new ones are added at the end) has an open connection,
+        // both addresses stay in the collection
+        assertAddresses(lb.activeAddresses(), "address-1", "address-1", "address-2");
+
+        sendServiceDiscoveryEvents(downEvent("address-2"));
+        // This host has a connection open, so it stays.
+        assertAddresses(lb.activeAddresses(), "address-1", "address-1", "address-2");
```
idelpivnitskiy, line 151: I've played with the tests. Please, make sure that all tests pass after the following changes:

```patch
Index: servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/EagerRoundRobinLoadBalancerTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/EagerRoundRobinLoadBalancerTest.java b/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/EagerRoundRobinLoadBalancerTest.java
--- a/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/EagerRoundRobinLoadBalancerTest.java	(revision a486e819a22c225237794acc1cb81be1db03cb95)
+++ b/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/EagerRoundRobinLoadBalancerTest.java	(date 1626325084267)
@@ -58,7 +58,7 @@
         assertAddresses(lb.activeAddresses(), "address-2");
 
         sendServiceDiscoveryEvents(upEvent("address-1"));
-        assertAddresses(lb.activeAddresses(), "address-1", "address-2");
+        assertAddresses(lb.activeAddresses(), "address-2", "address-1");
 
         sendServiceDiscoveryEvents(downEvent("address-1"));
         assertAddresses(lb.activeAddresses(), "address-2");
Index: servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/LingeringRoundRobinLoadBalancerTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/LingeringRoundRobinLoadBalancerTest.java b/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/LingeringRoundRobinLoadBalancerTest.java
--- a/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/LingeringRoundRobinLoadBalancerTest.java	(revision a486e819a22c225237794acc1cb81be1db03cb95)
+++ b/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/LingeringRoundRobinLoadBalancerTest.java	(date 1626331141590)
@@ -16,10 +16,13 @@
 package io.servicetalk.loadbalancer;
 
 import io.servicetalk.client.api.NoAvailableHostException;
+import io.servicetalk.concurrent.api.Executor;
+import io.servicetalk.concurrent.api.Executors;
 
 import org.junit.Test;
 
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.function.Predicate;
 
 import static io.servicetalk.concurrent.internal.DeliberateException.DELIBERATE_EXCEPTION;
@@ -27,6 +30,7 @@
 import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.Assert.assertThrows;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
@@ -35,7 +39,16 @@
 public class LingeringRoundRobinLoadBalancerTest extends RoundRobinLoadBalancerTest {
 
     @Test
-    public void hostDownDoesntCloseConnection() throws Exception {
+    public void hostDownDoesntCloseConnectionCloseLB() throws Exception {
+        hostDownDoesntCloseConnection(true);
+    }
+
+    @Test
+    public void hostDownDoesntCloseConnectionCloseLBGracefully() throws Exception {
+        hostDownDoesntCloseConnection(true);
+    }
+
+    private void hostDownDoesntCloseConnection(boolean gracefulClosure) throws Exception {
         sendServiceDiscoveryEvents(upEvent("address-1"));
         TestLoadBalancedConnection conn = lb.selectConnection(any()).toFuture().get();
         sendServiceDiscoveryEvents(downEvent("address-1"));
@@ -43,8 +56,15 @@
         verify(conn, never()).closeAsync();
 
         // But connection is cleaned when LB is closed
-        lb.closeAsync().toFuture().get();
-        verify(conn, times(1)).closeAsync();
+        if (gracefulClosure) {
+            lb.closeAsyncGracefully().toFuture().get();
+            verify(conn, times(1)).closeAsyncGracefully();
+            verify(conn, never()).closeAsync();
+        } else {
+            lb.closeAsync().toFuture().get();
+            verify(conn, times(1)).closeAsync();
+            verify(conn, never()).closeAsyncGracefully();
+        }
         assertAddresses(lb.activeAddresses(), EMPTY_ARRAY);
     }
 
@@ -67,6 +87,28 @@
         assertAddresses(lb.activeAddresses(), EMPTY_ARRAY);
     }
 
+    @Test
+    public void closureOfLastConnectionDoesntRaceWithNewAvailableEvent() throws Exception {
+        Executor executor = Executors.newFixedSizeExecutor(1);
+        try {
+            sendServiceDiscoveryEvents(upEvent("address-1"));
+            TestLoadBalancedConnection conn = lb.selectConnection(alwaysNewConnectionFilter()).toFuture().get();
+            sendServiceDiscoveryEvents(downEvent("address-1"));
+            assertConnectionCount(lb.activeAddresses(), connectionsCount("address-1", 1));
+
+            Future<Object> f = executor.submit(() -> {
+                sendServiceDiscoveryEvents(upEvent("address-1"));
+                return null;
+            }).toFuture();
+
+            conn.closeAsync().toFuture().get();
+            f.get();
+            assertConnectionCount(lb.activeAddresses(), connectionsCount("address-1", 0));
+        } finally {
+            executor.closeAsync().toFuture().get();
+        }
+    }
+
     @Test
     public void closedConnectionDoesntRemoveActiveHost() throws Exception {
         sendServiceDiscoveryEvents(upEvent("address-1"));
@@ -89,11 +131,11 @@
         assertAddresses(lb.activeAddresses(), "address-1");
 
         lb.selectConnection(any()).toFuture().get();
-
         assertConnectionCount(lb.activeAddresses(), connectionsCount("address-1", 1));
 
         sendServiceDiscoveryEvents(downEvent("address-1"));
-        assertAddresses(lb.activeAddresses(), "address-1");
+        assertConnectionCount(lb.activeAddresses(), connectionsCount("address-1", 1));
+        assertThat(lb.selectConnection(any()).toFuture().get(), is(notNullValue()));
 
         // We validate the host is expired by attempting to create a new connection
         final Predicate<TestLoadBalancedConnection> createNewConnection = alwaysNewConnectionFilter();
@@ -101,18 +143,15 @@
                 lb.selectConnection(createNewConnection).toFuture().get());
         assertThat(e.getCause(), instanceOf(NoAvailableHostException.class));
 
-        lb.selectConnection(any()).toFuture().get();
-
+        // When the host becomes available again, new connections can be created
         sendServiceDiscoveryEvents(upEvent("address-1"));
         assertAddresses(lb.activeAddresses(), "address-1");
         lb.selectConnection(createNewConnection).toFuture().get();
-
         assertConnectionCount(lb.activeAddresses(), connectionsCount("address-1", 2));
     }
 
     @Test
     public void handleDiscoveryEventsForConnectedHosts() throws Exception {
-        assertAddresses(lb.activeAddresses(), EMPTY_ARRAY);
         assertThat(lb.activeAddresses(), is(empty()));
 
         final Predicate<TestLoadBalancedConnection> connectionFilter = alwaysNewConnectionFilter();
@@ -137,27 +176,30 @@
         sendServiceDiscoveryEvents(upEvent("address-1"));
         assertAddresses(lb.activeAddresses(), "address-1", "address-2");
 
+        // When all hosts are active, new event creates a duplicate
         sendServiceDiscoveryEvents(upEvent("address-1"));
-        // At this moment in time, duplicates are allowed and the down event will remove just the first address
-        assertAddresses(lb.activeAddresses(), "address-1", "address-1", "address-2");
+        assertAddresses(lb.activeAddresses(), "address-1", "address-2", "address-1");
 
-        sendServiceDiscoveryEvents(downEvent("address-1"));
-        // Because the first address (new ones are added at the end) has an open connection,
+        // Because the first address (new ones are added at the end) has an open connection it's marked as "expired",
         // both addresses stay in the collection
-        assertAddresses(lb.activeAddresses(), "address-1", "address-1", "address-2");
+        sendServiceDiscoveryEvents(downEvent("address-1"));
+        assertAddresses(lb.activeAddresses(), "address-1", "address-2", "address-1");
 
+        // This host has a connection open, so it stays as "expired".
         sendServiceDiscoveryEvents(downEvent("address-2"));
-        // This host has a connection open, so it stays.
-        assertAddresses(lb.activeAddresses(), "address-1", "address-1", "address-2");
+        assertAddresses(lb.activeAddresses(), "address-1", "address-2", "address-1");
+
+        // Another "down" event removes the second entry of "address-1" that does not have active connections
+        sendServiceDiscoveryEvents(downEvent("address-1"));
+        assertAddresses(lb.activeAddresses(), "address-1", "address-2");
 
         // Let's make sure that an SD failure doesn't compromise LB's internal state
         serviceDiscoveryPublisher.onError(DELIBERATE_EXCEPTION);
-        assertAddresses(lb.activeAddresses(), "address-1", "address-1", "address-2");
+        assertAddresses(lb.activeAddresses(), "address-1", "address-2", "address-1");
     }
 
     @Test
     public void handleDiscoveryEventsForNotConnectedHosts() {
-        assertAddresses(lb.activeAddresses(), EMPTY_ARRAY);
         assertThat(lb.activeAddresses(), is(empty()));
 
         sendServiceDiscoveryEvents(upEvent("address-1"));
@@ -173,7 +215,7 @@
         assertAddresses(lb.activeAddresses(), "address-2");
 
         sendServiceDiscoveryEvents(upEvent("address-1"));
-        assertAddresses(lb.activeAddresses(), "address-1", "address-2");
+        assertAddresses(lb.activeAddresses(), "address-2", "address-1");
 
         sendServiceDiscoveryEvents(downEvent("address-1"));
         assertAddresses(lb.activeAddresses(), "address-2");
Index: servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java b/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
--- a/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java	(revision a486e819a22c225237794acc1cb81be1db03cb95)
+++ b/servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java	(date 1626325104652)
@@ -79,7 +79,6 @@
 import static org.hamcrest.Matchers.both;
 import static org.hamcrest.Matchers.contains;
-import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.emptyIterable;
 import static org.hamcrest.Matchers.equalTo;
@@ -413,7 +412,7 @@
     }
 
     @SafeVarargs
-    protected final <T> void assertConnectionCount(
+    protected static <T> void assertConnectionCount(
             Iterable<T> addresses, Map.Entry<String, Integer>... addressAndConnCount) {
         final Matcher<? super T>[] args = (Matcher<? super T>[]) Arrays.stream(addressAndConnCount)
@@ -422,7 +421,7 @@
                 .collect(Collectors.toList())
                 .toArray(new Matcher[] {});
         final Matcher<Iterable<? extends T>> iterableMatcher =
-                addressAndConnCount.length == 0 ? emptyIterable() : containsInAnyOrder(args);
+                addressAndConnCount.length == 0 ? emptyIterable() : contains(args);
         assertThat(addresses, iterableMatcher);
     }
 
@@ -437,7 +436,7 @@
                 .collect(Collectors.toList())
                 .toArray(new Matcher[] {});
         final Matcher<Iterable<? extends T>> iterableMatcher =
-                address.length == 0 ? emptyIterable() : containsInAnyOrder(args);
+                address.length == 0 ? emptyIterable() : contains(args);
         assertThat(addresses, iterableMatcher);
     }
 

```

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * Builder for {@link RoundRobinLoadBalancerFactory}.
+     *
+     * @param <ResolvedAddress> The resolved address type.
+     * @param <C> The type of connection.
+     */
+    public static final class Builder<ResolvedAddress, C extends LoadBalancedConnection> {
+        private boolean eagerConnectionShutdown;
```
idelpivnitskiy, line 71: I'm a bit hesitant of enforcing the new behavior by default without proper validation. WDYT if we make it opt-in in the next release, validate with some users, and if everything works as expected we can make it opt-out in the following releases.

idelpivnitskiy (changes requested):

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
chemicL: Agreed, my original idea was to introduce it as an opt-in. I committed the change.

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/LingeringRoundRobinLoadBalancerTest.java
chemicL: Huge thanks for putting so much effort into this. I reworked the concurrency as you suggested with more states but a single volatile variable that controls the lifetime. Please have a look. The test passes 10k iterations now.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -202,6 +204,76 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
                 }
             }
 
+            private List<Host<ResolvedAddress, C>> markHostAsExpired(
+                    final List<Host<ResolvedAddress, C>> oldHostsTyped, final ResolvedAddress addr) {
+                for (Host<ResolvedAddress, C> host : oldHostsTyped) {
+                    if (host.address.equals(addr)) {
+                        // Host removal will be handled by the Host's onClose::afterFinally callback
+                        host.markExpired();
```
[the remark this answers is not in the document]
chemicL: Fixed.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -162,31 +174,21 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
-                                    host.markInactive();
-                                    for (int x = i + 1; x < oldHostsTyped.size(); ++x) {
-                                        newHosts.add(oldHostsTyped.get(x));
-                                    }
-                                    return newHosts.isEmpty() ? emptyList() : newHosts;
-                                } else {
-                                    newHosts.add(host);
-                                }
+                            if (event.isAvailable()) {
+                                return addHostToList(oldHostsTyped, addr, true);
```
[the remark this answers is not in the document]
chemicL: I'd actually not go that direction. To me it seems easier to reason about behaviour the way it is now.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -202,6 +204,76 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
+
+            private List<Host<ResolvedAddress, C>> removeFromHostsList(
+                    List<Host<ResolvedAddress, C>> oldHostsTyped, ResolvedAddress addr) {
+                if (oldHostsTyped == CLOSED_LIST) {
+                    // this can happen when an expired host is removed during closing of the RoundRobinLoadBalancer,
+                    // but all of its connections have already been closed
+                    return oldHostsTyped;
+                }
+                if (oldHostsTyped.isEmpty()) {
+                    return emptyList();
```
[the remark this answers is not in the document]
chemicL: I think there were some previous distinctions between the static `EMPTY_LIST` vs a new instance of `emptyList()` but it looks like they both boil down to the same case.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -162,31 +174,27 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
-                                if (host.address.equals(addr)) {
-                                    host.markInactive();
-                                    for (int x = i + 1; x < oldHostsTyped.size(); ++x) {
-                                        newHosts.add(oldHostsTyped.get(x));
-                                    }
-                                    return newHosts.isEmpty() ? emptyList() : newHosts;
-                                } else {
-                                    newHosts.add(host);
-                                }
+                            if (event.isAvailable()) {
```
Scottmitch, line 191: nit: combine with outer `else` clause to reduce scoping.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -255,48 +335,68 @@ public void onComplete() {
...
-            if (selector.test(connection)) {
-                return succeeded(connection);
+        Host<ResolvedAddress, C> pickedHost = null;
+        for (int i = 0; i < activeHosts.size(); ++i) {
+            // for a particular iteration we maintain a local cursor without contention with other requests
+            int localCursor = (cursor + i) % activeHosts.size();
+            final Host<ResolvedAddress, C> host = activeHosts.get(localCursor);
+            assert host != null : "Host can't be null.";
+
+            final ThreadLocalRandom rnd = ThreadLocalRandom.current();
```
Scottmitch, line 351: nit: since we are caching this in a local variable, it can be lifted outside the loop.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -255,48 +335,68 @@ public void onComplete() {
...
+            for (int j = 0; j < attempts; ++j) {
+                @SuppressWarnings("unchecked")
+                final C connection = (C) connections[rnd.nextInt(connections.length)];
+                if (selector.test(connection)) {
+                    return succeeded(connection);
+                }
             }
-        }
 
+            // don't open new connections for expired hosts, try a different one
```
Scottmitch, line 367: can you elaborate on why this is done after the connection selection has already been attempted? seems like we may select a connection even if the host is not active.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -255,48 +335,68 @@ public void onComplete() {
...
+            if (host.isActive()) {
```
Scottmitch, line 368: the name of the collection which holds all the SD events is called `activeHosts`, it is strange for this method to return false and still be retained in the collection, consider clarifying the naming to disambiguate (previously when hosts went inactive they were removed).

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -359,12 +483,32 @@ void markInactive() {
             }
         }
 
+        @SuppressWarnings("PMD.CollapsibleIfStatements")
+        void markExpired() {
+            stateUpdater.set(this, connections.length == 0 ? State.CLOSED : State.EXPIRED);
```
Scottmitch, line 488: since there are two independent volatiles involved, can you add a comment to clarify why accessing independent volatile state is safe?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -359,12 +483,32 @@ void markInactive() {
...
+            return state == State.ACTIVE;
+        }
+
+        boolean isExpired() {
+            return state == State.EXPIRED;
         }
 
         boolean addConnection(C connection) {
             for (;;) {
+                if (state == State.CLOSED) {
```
Scottmitch, line 509: if the state is closed won't this host be deferenced (eventually)? if so do we need this check here on every `addConnection`? at best this is only a best effort check anyways due to independent volatile variables IIUC.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -391,6 +535,17 @@ boolean addConnection(C connection) {
                     }
                     if (i == existing.length) {
                         break;
+                    } else if (existing.length == 1 && state == State.EXPIRED) {
```
Scottmitch, line 538: IIUC if we don't observe `state == State.EXPIRED` we may retain a stale entry in our collection until another SD event comes in pertaining to this host?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: The selection until this point tries to reuse existing connections. Expired hosts take part in the round robin selection for active connections – otherwise they'd be useless in the process. The idea here is that – if a host is still serving the traffic but is considered expired, we don't close the connections, but continue using them. We don't open new connections to expired hosts though, that's why the check is performed at the moment a new connection needs to be opened. This allows for shifting traffic – if the expired host is going away it will close the connections from the client. In effect, the client will also see this and will remove the expired host entirely.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: I consolidated the naming - the collection was renamed `activeHosts` -> `usedHosts`, while the `AVAILABLE` state is represented consistently. Other states also have more consistent presence.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: Please review the added comment.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: That is true, I added a new test for concurrent behaviour that validates removing this best effort check does not cause unexpected behaviour.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: If the state is not `EXPIRED` it is either `CLOSED` or `AVAILABLE`. For `CLOSED` the host is already being removed and its connections are either removed or just about to be removed. For `AVAILABLE` it's not a stale entry – the host can be selected and a new connection will be created upon demand.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: makes sense!

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -251,52 +329,73 @@ public void onComplete() {
...
                         return succeeded(newCnx);
                     }
-                    return newCnx.closeAsync().concat(this.activeHosts == CLOSED_LIST ? failedLBClosed() :
-                            failed(new ConnectionRejectedException("Failed to add newly created connection for host: " +
-                                    host.address + ", host inactive? " + host.isInactive())));
+                    return newCnx.closeAsync().concat(this.usedHosts == CLOSED_LIST ? failedLBClosed() :
+                            failed(new ConnectionRejectedException(
+                                    "Failed to add newly created connection for host: "
+                                            + host.address
+                                            + ", host active? " + host.isActive()
```
Scottmitch, line 396: Can you use `Host#toString()` here? Then you could make `isClosed()` and `isExpired()` private.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: I think we have a race condition which may lea
```
Host A {connection1, connection2}

T1 - SD event {A unavailable}, A.markExpired(), connections.length == 2
T2 - c1 closed, existing.length == 2, state == ACTIVE, new = {c2}, connectionsUpdater.cas(old, new)
T2 - c2 closed, existing.length == 1, state == ACTIVE, new = {}, connectionsUpdater.cas(old, new)
T1 - state = expired
```

note that `Host#closeAsync()` is never called, and `createHost`'s `host.onClose().afterFinally(() -> ..)` may never run to cleanup `usedHostsUpdater`. I think there are also scenarios where we may create connections after SD has told us the host is no longer available. I think we should stick to a single variable for the state and connections, otherwise it is difficult to eliminate race conditions.

```java
        @SuppressWarnings("rawtypes")
        private static final AtomicReferenceFieldUpdater<Host, ConnState> connStateUpdater =
                AtomicReferenceFieldUpdater.newUpdater(Host.class, ConnState.class, "connState");
        private static final class ConnState {
            final Object[] connections;
            final State state;

            ConnectionsState(Object[] connections, State state) {
                this.connections = connections;
                this.state = state;
            }
        }
        private static final ConnState EMPTY_CONN_STATE = new ConnState(EMPTY_ARRAY, ...);
        
        private volatile ConnState connState = EMPTY_CONN_STATE;
```

Scottmitch (changes requested):

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -220,9 +223,12 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
                 Host<ResolvedAddress, C> host = new Host<>(addr);
                 if (!eagerConnectionShutdown) {
-                    host.onClose().afterFinally(() ->
-                            activeHostsUpdater.updateAndGet(RoundRobinLoadBalancer.this,
-                                    previousHosts -> removeFromHostsList(previousHosts, addr))
+                    host.onClose().afterFinally(() -> {
+                                usedHostsUpdater.updateAndGet(RoundRobinLoadBalancer.this,
+                                        previousHosts ->
+                                                listWithHostRemoved(previousHosts, current -> current == host));
+                                host.markClosed();
```
idelpivnitskiy, line 230: IIUC, this callback can be invoked only if the host is already marked as closed, no need to mark it again.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
idelpivnitskiy: 1. Consider saving `state` in a local variable and break earlier if it's already `CLOSED`.
2. Remove `state == State.EXPIRED` check from here and go inside this branch for all `existing.length == 1` cases. Then try to `cas(EXPIRED, CLOSED)`, if fails try `cas(existing, EMPTY_ARRAY)`. It will help to avoid allocating a new object array if host is active.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: Thank you for spotting that and for the excellent suggestion. I've implemented it the way you suggest, will push a commit soon.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: With the new approach suggested by @Scottmitch this optimisation won't be necessary.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -316,115 +410,189 @@ public Completable closeAsyncGracefully() {
...
             for (Object conn : toRemove) {
                 @SuppressWarnings("unchecked")
                 final C cConn = (C) conn;
                 cConn.closeAsyncGracefully().subscribe();
             }
         }
 
-        boolean isInactive() {
-            return connections == CLOSED_ARRAY;
+        @SuppressWarnings("PMD.CollapsibleIfStatements")
```
Scottmitch, line 490: I think this can be removed now?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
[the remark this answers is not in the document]
Scottmitch: @idelpivnitskiy - IIUC your suggestion is as follows:

```java
if (host.address.equals(addr) && host.markExpired()) {
  break;
}
```

I think this makes sense in general (even if duplicates are not present we would want to bail when we found a match).

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -154,57 +166,118 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
...
+                    }
+                }
+                return oldHostsTyped;
+            }
+
+            @SuppressWarnings("unchecked")
+            private Host<ResolvedAddress, C> createHost(ResolvedAddress addr) {
+                Host<ResolvedAddress, C> host = new Host<>(addr);
+                if (!eagerConnectionShutdown) {
+                    host.onClose().afterFinally(() -> {
```
Scottmitch, line 226: replace lambda with expression.

also consider reducing scope of `@SuppressWarnings("unchecked")` to site of warning. this can help if the method evolves to avoid unintended warnings from being suppressed.

```java
previousHosts -> {
  List<Host<ResolvedAddress, C>> cast = (List<Host<ResolvedAddress, C>>) previousHosts;
  return listWithHostRemoved(cast, current -> current == host);
})
```

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -316,115 +410,189 @@ public Completable closeAsyncGracefully() {
...
+                this.connections = connections;
+                this.state = state;
+            }
+        }
 
         final Addr address;
-        private volatile Object[] connections = EMPTY_ARRAY;
+        private final ListenableAsyncCloseable closeable;
+
+        @SuppressWarnings("rawtypes")
```
Scottmitch, line 460: I thought we had quality rules for this but typically the `private static final` methods are all grouped together at the top of the class, and non-static state is below that.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -0,0 +1,107 @@
...
+
+    static final boolean EAGER_CONNECTION_SHUTDOWN_ENABLED = true;
+    private final boolean eagerConnectionShutdown;
+
+    private RoundRobinLoadBalancerFactory(boolean eagerConnectionShutdown) {
+        this.eagerConnectionShutdown = eagerConnectionShutdown;
+    }
+
+    @Override
+    public <T extends C> LoadBalancer<T> newLoadBalancer(
```
Scottmitch, line 59: do we need a factory and a builder? is just a builder sufficient? if just a builder is sufficient we could have it be an inner type of `RoundRobinLoadBalancer` and avoid deprecation of the existing type.

Scottmitch (commented):
few questions but approach generally lgtm

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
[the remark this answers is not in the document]
idelpivnitskiy: Correct, we want to break anyway. The only difference will be: if duplicates are supported, `markExpired()` returns `true` only if we transition from `ACTIVE` to `EXPIRED`. If duplicates are not supported, we can break at any address match, regardless of its state.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
idelpivnitskiy: All our API at the client level target `LoadBalancerFactory`. It's hard to operate with the `LoadBalancer` directly because you have to coordinate events publisher and connection factory. So, `LoadBalancerFactory` is a helpful abstraction.
Here we have a builder for the `LoadBalancerFactory` because we want to set some configuration options (we plan to add more in a follow-up PR). Users can use the result of this builder to create as many clients as necessary, reusing a single `LoadBalancerFactory`.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
idelpivnitskiy: Unfortunately, check style does not have this type of validation (to my knowledge).

idelpivnitskiy (commented):
Great progress! All looks good, this is the last discussion: https://github.com/apple/servicetalk/pull/1629/files#r670155425

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: We have https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/DeclarationOrderCheck.html configured with `ignoreModifiers=true`. The check follows the original Sun Java Code Conventions document and according to it all methods go below class/instance fields.

chemicL: Our checkstyle configuration uses https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/DeclarationOrderCheck.html with `<property name="ignoreModifiers" value="true"/>`, which disables such detection. Should we re-enable it? When that property is not set, the order is properly verified.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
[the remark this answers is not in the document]
chemicL: I agree breaking is in general desired. My concern with duplicate entries is the following:
When there was no `EXPIRED` state, duplicates made sense - every `AVAILABLE` event would create an entry, which would remain in the array until an `UNAVAILABLE` event arrives to remove the entry. Each event would need to have a pair. Still, I'm already lost on the semantics here – if an address is unavailable - why would it reside in two states at some point – some available and some not? I'd assume all duplicate entries concern the same host - if the host is not available, it is effectively unavailable in all duplicate occurrences. Hence, I'd iterate over all duplicates. However, if every event has a match, I'd assume everything goes well and remove just one entry at a time, so be it.
For `EXPIRED` hosts, the entries can go away in the meantime when connections close. It's hard to mentally keep track of the pairs of events – when marking a host as available after it was expired there is no mental match of which event corresponds to which entry as their presence in the array is not guaranteed. To me, duplicate `AVAILABLE` events made sense for increasing the frequency of selecting a particular host, but pairing `UNAVAILABLE` events sounds less solid to me. In face of `EXPIRED`, even duplicate `AVAILABLE` events no longer signify the same thing – they can either mean "increase the frequency of selection" or "mark as no longer expired". And even that statement depends on the current state (the former when the previously expired host went away and the latter when we still have connections).

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -484,76 +489,75 @@ void markClosed() {
...
+                    for (; i < connections.length; ++i) {
+                        if (connections[i].equals(connection)) {
                             break;
                         }
                     }
-                    if (i == existing.length) {
+                    if (i == connections.length) {
                         break;
-                    } else if (existing.length == 1 && state == State.EXPIRED) {
+                    } else if (connections.length == 1 && currentConnState.state == State.EXPIRED) {
```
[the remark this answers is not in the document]
chemicL: Thanks, now I see the optimisation you were thinking of. I'll change it to what you suggest.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: @chemicL - lets consider check style updates in a followup/isolated PR, but yes this seems interesting if it leads to more standardization.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -239,11 +240,18 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
                     return singletonList(createHost(addr));
                 }
 
-                if (handleExpired) {
-                    for (Host<ResolvedAddress, C> host : oldHostsTyped) {
-                        if (host.address.equals(addr) && host.tryToMarkActive()) {
-                            return oldHostsTyped;
+                // duplicates are not allowed
```
Scottmitch, line 243: do we need to actively prevent/detect duplicates unconditionally? the API currently doesn't allow for duplicates and our most popular discoverer (DNS) doesn't send duplicates. consider reverting to retain the `handleExpired` to avoid O(n) iteration before copy/add, and/or pre-allocate `newHosts` and add while iterating (duplicates are not expected/supported).

Scottmitch (commented):
last comment then lgtm

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -169,7 +169,7 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
                 final List<Host<ResolvedAddress, C>> usedAddresses =
                     usedHostsUpdater.updateAndGet(RoundRobinLoadBalancer.this, oldHosts -> {
                         if (oldHosts == CLOSED_LIST) {
-                            return CLOSED_LIST;
+                            return oldHosts;
```
idelpivnitskiy, line 172: Intellij IDEA had a warning that there is no cast if `CLOSED_LIST` is used, easier just to return `oldHosts` and not worry about types.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -243,9 +244,7 @@ public void onNext(final ServiceDiscovererEvent<ResolvedAddress> event) {
                 // duplicates are not allowed
                 for (Host<ResolvedAddress, C> host : oldHostsTyped) {
                     if (host.address.equals(addr)) {
-                        if (!handleExpired) {
-                            LOGGER.warn("Duplicate ACTIVE event for host " + host + ". Duplicates are disallowed.");
```
idelpivnitskiy: Not necessary to warn here, because duplicates will be expected when we remove deduplication logic from `DefaultServiceDiscoveryRetryStrategy`. Duplicates will be propagated if you retry SD events publisher. We can just silently ignore them or add trace/debug level logging later, if necessary.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -557,7 +556,7 @@ boolean addConnection(C connection) {
                                 // remove the connection (previously considered as the last one) from the array
                                 // in the next iteration.
                                 && connStateUpdater.compareAndSet(this, currentConnState, CLOSED_CONN_STATE)) {
-                            closeAsync().subscribe();
+                            this.closeAsync().subscribe();
```
idelpivnitskiy, line 559: Just for consistency with the other case where you invoke `closeAsync()`

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -607,6 +606,7 @@ private Completable doClose(final Function<? super C, Completable> closeFunction
 
         @Override
         public String toString() {
+            final ConnState connState = this.connState;
```
idelpivnitskiy, line 609: Local variable helps to make sure, `connState.state` and `connState.connections` relate to the same `connState` instance.

idelpivnitskiy (approved):
Awesome job! It was not easy, but I hope we learned a lot on the way.

I would like to include this in the next release. So, pushed minor changes: https://github.com/apple/servicetalk/pull/1629/commits/cbc5707079f455b00020fbb56bfcf3acc2d734d7

PR #1578, opened 2021-05-25
Deprecate `append` method on all filter factories

idelpivnitskiy:
Motivation:

All `*FilterFactory` interfaces currently implement default `append`
method that helps to chain filters in order of append. Because
internally it uses wrapping approach, it can hide
`HttpExecutionStrategyInfluencer` implementation of the passed filter
factory. To avoid that, users have to use a method on the client/server
builder instead.

Modifications:

- Deprecate `append` method on all filter factories;
- Add `FilterFactoryUtils` test fixture as the replacement utility for
tests;

Result:

Users won't hide `HttpExecutionStrategyInfluencer` impl when they chain
filters.

PR #1631, opened 2021-06-21
Remove `append` methods that were deprecated in #1578

idelpivnitskiy:
Motivation:

`append` method on each filter factory was deprecated in #1578 and can
be removed now.

Modifications:

- Remove `append` method from `ConnectionFactoryFilter`,
`GrpcServiceFilterFactory`, `MultiAddressHttpClientFilterFactory`,
`StreamingHttpClientFilterFactory`, `StreamingHttpConnectionFilterFactory`,
`StreamingHttpServiceFilterFactory`.

Result:

No deprecated methods on filter factory API.


bondolo (approved):
LGTM assuming checks pass.

PR #1634, opened 2021-06-21
Remove deprecated CharSequences methods

bondolo:
Motivation:
The `newAsciiString(String)` and `isAsciiString(CharSequence)`
methods have been previously deprecated and may now be removed.
The `newAsciiString(String)` is not marked as deprecated because
it is provided for binary compatibility only; the removal is source
compatible because the more general newAsciiString(CharSequence)
method is provided.

Modifications:
Deprecated methods of `CharSequences` are removed.

Result:
Obsolete methods no longer supported.


idelpivnitskiy (approved):
🚀

PR #1640, opened 2021-06-25
Fix B3 sampling API to support missing values

tkountis:
Motivation:

B3 sampling handling was supposed to support both present and not values.
However, in practise it only supported present values, and treated non existing as sampling=off.

Modifications:

Support Boolean values for handling sampling which now allows for not present values in the carrier,
and the local sampler can make the final decision on whether to enable sampling for the local span or not.

The change unfortunately broke API because of the type change, thus we took advantage of that and cleaned up
some problematic relationships between classes which results in a cleaner more consistent API, which also
follows closer the spec.

- There is no `InMemoryTraceState` since it was mainly duplication of `SpanContext` and the spec suggests that all these variables can/should be part the context.
- Span has no accessors for IDs. To keep everything consistent, the accessors are now only available through the `context`. see. https://github.com/opentracing/specification/blob/master/rfc/trace_identifiers.md

Result:

Correct sampling & cleaner API

Link: https://github.com/apple/servicetalk/pull/1512


Scottmitch:
@eddie4941 - FYI. PTAL.

servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/AbstractInMemoryTracer.java
```diff
@@ -80,10 +76,14 @@
...
+     * @param spanId       span ID
+     * @param parentSpanId parent span ID, optional
+     * @param sampled      whether the trace is sampled
      *
-     * @param state The state for the trace.
-     * @return span context
+     * @return span context {@link InMemorySpanContext}
      */
-    protected abstract InMemorySpanContext newSpanContext(InMemoryTraceState state);
+    protected abstract InMemorySpanContext newSpanContext(String traceId, String spanId, String parentSpanId,
```
Scottmitch, line 87: this method is no longer called in this class, make it private to `DefaultInMemoryTracer`?

servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/DefaultInMemorySpanContext.java
```diff
@@ -16,54 +16,73 @@
...
 
 import static io.servicetalk.opentracing.inmemory.SingleLineValue.format;
-import static java.util.Objects.requireNonNull;
 
 /**
- * Span object used by the {@link DefaultInMemoryTracer}.
+ * SpanContext object used by the {@link DefaultInMemoryTracer}.
  */
-class DefaultInMemorySpanContext implements InMemorySpanContext {
+public class DefaultInMemorySpanContext implements InMemorySpanContext {
```
Scottmitch, line 32: can we make this class final for now (more restrictive access pattern to minimize our API surface)?

Scottmitch (commented):
few comments then lgtm

PR #1641, opened 2021-06-25
Mark any exception before the first message is written as retryable

idelpivnitskiy:
Motivation:

If the `Channel` is already closed before we attempt the first write, we
wrap it with `AbortedFirstWrite` in `WriteStreamSubscriber` and re-wrap
with `RetryableClosureException` in `DefaultNettyConnection`. This
approach has a couple of disadvantages:
1. `AbortedFirstWrite` is not a `RetryableException` and if the
`closeReason` is not known in `DefaultNettyConnection`, the original
exception will be propagated as-is without `RetryableException` marker;
2. `WriteStreamSubscriber` prematurely sets `written` flag before we
attempt the first write, that may also fail even if the `Channel` is not
closed;

Modifications:
- `WriteStreamSubscriber`: mark internal state as `written` only after
the first msg successfully goes through the outbound channel pipeline;
- Rename `AbortedFirstWrite` -> `AbortedFirstWriteException` and make it
retryable `IOException` because it indicates only IO failures;
- Wrap any error that happened before the first msg is written with
`AbortedFirstWriteException`;
- Adjust `DefaultNettyConnection#enrichError` to correctly unwrap
`AbortedFirstWriteException`;

Result:

Any failure before the first message is written is marked as retryable.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
@@ -142,7 +142,7 @@ final class HeaderUtils {
 
     static ScanWithMapper<Object, Object> insertTrailersMapper() {
         return new ScanWithMapper<Object, Object>() {
-            @Nullable
+
             private boolean sawHeaders;
 
             @Nullable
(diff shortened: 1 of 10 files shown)


idelpivnitskiy:
Build failed due to #1579

chemicL (approved):

PR #1643, opened 2021-06-25
`Single.concat(Publisher)` defer subscribe to `Publisher` until requested

idelpivnitskiy:
Motivation:

Current version of `Single.concat(Publisher)` subscribes to the next
`Publisher` as soon as the `Single` completes. If the passed `Publisher`
does not support multiple subscribes, users can not apply retries if the
further processing of the single result fails.

Modifications:

- Add `Single.concat(Publisher, boolean)` overload that tells if we
need to subscribe to the next `Publisher` asap or defer subscribe until
more items are requested;
- Implement another `ConcatDeferNextSubscriber` variant that defers
subscribe to the next publisher, share common code in
`AbstractConcatSubscriber`;
- Enhance tests for new operator variant;

Result:

If processing of the first item failed and no more items were requested,
it's safe to apply retry operator even with a `Publisher` that do not
support re-subscribe.

- [x] Depends on #1642.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -740,7 +740,30 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
      * all elements from {@code next} {@link Publisher}.
```
bondolo, line 740: This method should document that it is equivalent to concat(next, false) and explain that it will try to subscribe before the next item is needed.

I am also wondering how common it is for a Publisher to not allow concurrent subscribes. If it is at least somewhat common then perhaps the default should be true, just for safety. What are the execution or performance implications of deferring the subscribe? Would there be a downside to always deferring the subscribe? Could/should Publishers be typed with a marker interface to indicate that they don't support concurrent subscribes?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
```diff
@@ -43,100 +47,141 @@ Executor executor() {
...
-            target.onError(t);
+        public final void cancel() {
+            // We track cancelled here because we need to make sure if cancel() happens subsequent calls to request(n)
+            // are NOOPs [1].
+            // [1] https://github.com/reactive-streams/reactive-streams-jvm#3.6
+            mayBeResult = CANCELLED;
+            super.cancel();
+        }
+
+        final void superCancel() {
```
bondolo, line 120: this seems a little odd, what problem is it trying to solve?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
```diff
@@ -145,24 +190,136 @@ public void request(long n) {
                 }
             }
         }
+    }
+
+    private static final class ConcatDeferNextSubscriber<T> extends AbstractConcatSubscriber<T> {
+        /**
+         * If only one item was {@link #request(long) requested} before {@link #onSuccess(Object)}.
+         */
+        private static final Object REQUESTED_ONE = new Object();
```
bondolo, line 199: I like defining toString for these to the name of the field. Some people use an enum if they have a bunch of them to define as the toString is free. Instance variables can use either the enum type or Object and if the enum type, it is nice to be able to use switch.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleConcatWithPublisherDeferSubscribeTest.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.concurrent.api.single;
+
+class SingleConcatWithPublisherDeferSubscribeTest extends SingleConcatWithPublisherTest {
+
+    @Override
+    boolean deferSubscribe() {
```
bondolo, line 21: Parameterized test instead? I don't understand why this should be two tests

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
idelpivnitskiy: When we discussed with @Scottmitch the approach offline we agreed to keep the existing behavior as-is in case users expect it and have the new behavior opt-in. The new operator is more complex and expensive. Maybe in future we will keep only new behavior and make it the default. But probably we will need to rework all other operators to behave in the same way (there are more `concat` variants for other sources).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
idelpivnitskiy: haha, the name of the method is funny :) 
the intend is to invoke `DelayedCancellableThenSubscription#cancel()` from subclasses of `AbstractConcatSubscriber`. Kind of `super.super.cancel()`. I will add a comment

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
idelpivnitskiy: I like an idea for using an `enum` in general, but it didn't fit well for this use-case because:
- set of states is different for each subscriber impl
- `mayBeResult` may also result of the `Single`
I will keep it consistent with all other operators and we can refactor them altogether if necessary

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -738,9 +742,34 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
...
+     * <pre>{@code
+     *     List<T> results = new ...;
+     *     results.add(resultOfThisSingle());
+     *     results.addAll(nextStream());
+     *     return results;
+     * }</pre>
+     * @param next {@link Publisher} to concat.
+     * @param deferSubscribe if {@code true} subscribe to the {@code next} {@link Publisher} will be deferred until
+     * demand is received. Otherwise, it subscribes to the {@code next} {@link Publisher} as soon as {@code this}
+     * {@link Single} completes successfully.
```
bondolo, line 767: I think that some explanation of why you would choose either false or true should be included. "Choosing the deferred (true) behavior is important if this Publisher does not or might not support multiple concurrent subscribers." and any other reasons that might be relevant. Choosing the immediate subscription (false) behavior may have better performance for Publishers that support it and should be used whenever possible." WDYT?

bondolo (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
```diff
@@ -43,100 +47,144 @@ Executor executor() {
...
                     if (oldVal != INITIAL) {
                         @SuppressWarnings("unchecked")
                         final T tVal = (T) oldVal;
-                        emitSingleSuccessToTarget(tVal);
+                        if (tryEmitSingleSuccessToTarget(tVal)) {
+                            next.subscribeInternal(this);
+                        }
                     }
                     if (n != 1) {
                         super.request(n - 1);
```
idelpivnitskiy, line 190: FYI: I observed this issue with pre-existing behavior of ``: https://github.com/apple/servicetalk/runs/2915774996
It does not happen if I move this `super.request(n - 1)` before `if (oldVal != INITIAL)`.

idelpivnitskiy: Changed order in https://github.com/apple/servicetalk/pull/1643/commits/dc592f2d5d065d0945e8a085bbad247d1c9edee2

idelpivnitskiy: Issue was fixed in #1646

PR #1644, opened 2021-06-25
HTTP request may become non-retryable when first write fails

idelpivnitskiy:
Motivation:

Because `WriteStreamSubscriber` requests >1 items for new requests and
`meta-data.concat(messageBody)` subscribes to the `messageBody` asap,
it's not safe to retry these requests if the payload body does not
support multiple subscribes (non-replayable).

Modifications:

- Reproduce described behavior in a test;
- `WriteStreamSubscriber` requests only one item on the client-side and
requests more only if the first write succeeds;
- Use new operator `Single.concat(Publisher, boolean)` for
`AbstractStreamingHttpConnection` and `NettyHttpServer`;
- Remove unused `requested` counter from `WriteStreamSubscriber`;

Result:

Client subscribes to the message body only if the request meta-data is
written successfully. Otherwise, `RetryableException` is propagated.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
@@ -241,7 +241,7 @@ final class WriteStreamSubscriber implements PublisherSource.Subscriber<Object>,
 
     private void initialRequestN(Subscription subscription) {
         if (isClient) {
-            if (channel.isWritable()) {
+            if (promise.isWritable()) {
                 subscription.request(1L);   // Request meta-data only
             }
         } else {
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
@@ -21,7 +21,6 @@ import io.servicetalk.concurrent.PublisherSource;
 import io.servicetalk.concurrent.PublisherSource.Subscription;
 import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.concurrent.internal.ConcurrentSubscription;
-import io.servicetalk.concurrent.internal.FlowControlUtils;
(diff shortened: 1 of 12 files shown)


idelpivnitskiy:
@chemicL I tried your suggestion to edit the PR and change the target branch to #1641, but it doesn't work because `AbortedFirstWrite` is in my fork, not in the main `servicetalk` repo. Please, use GH UI to select only the last commit for review: https://github.com/apple/servicetalk/pull/1644/commits/fbccead52edaae73abe65d2df624575b9d2e0d5d

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -248,16 +239,26 @@ public void cancel() {
...
+
+    private void requestMoreIfRequired(@Nullable Subscription subscription, long bytesBeforeUnwritable) {
         // subscription could be null if channelWritable is invoked before onSubscribe.
         // If promise is not writable, then we will not be able to write anyways, so do not request more.
         if (subscription == null || subscription == CANCELLED || !promise.isWritable()) {
             return;
         }
 
-        long n = demandEstimator.estimateRequestN(channel.bytesBeforeUnwritable());
+        long n = demandEstimator.estimateRequestN(bytesBeforeUnwritable >= 0 ? bytesBeforeUnwritable :
```
chemicL, line 259: Why is the check required? When `bytesBeforeUnwritable` is >= it's equal to `channel.bytesBeforeUnwritable()` (line 164), otherwise, `channel.bytesBeforeUnwritable()` is used directly.

chemicL (approved):
Just one question, otherwise great job 👍

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
idelpivnitskiy: The most frequent method that invokes `requestMoreIfRequired` is `doWrite`, where we already have `capacityAfter`. It's a way to avoid the second invocation of `channel.bytesBeforeUnwritable()` that will return the same result because we are on a single event-loop thread here. `bytesBeforeUnwritable()` is not super cheap because it does some math and volatile read internally.

PR #1645, opened 2021-06-28
Migrate servicetalk-data-jackson-jersey and servicetalk-http-router-jersey tests to junit5

danfaer:
Motivation:

JUnit 5 leverages features from Java 8 or later, such as lambda functions, making tests more powerful and easier to maintain.
JUnit 5 has added some very useful new features for describing, organizing, and executing tests. For instance, tests get better display names and can be organized hierarchically.
JUnit 5 is organized into multiple libraries, so only the features you need are imported into your project. With build systems such as Maven and Gradle, including the right libraries is easy.
JUnit 5 can use more than one extension at a time, which JUnit 4 could not (only one runner could be used at a time). This means you can easily combine the Spring extension with other extensions (such as your own custom extension).
Modifications:

Unit tests have been migrated from JUnit 4 to JUnit 5
Upgraded junit5 to version 5.0.8 as it supports declarative test suits

Result:

Modules servicetalk-data-jackson-jersey and servicetalk-http-router-jersey now run tests and test suits using JUnit 5

https://github.com/apple/servicetalk/issues/1568


gradle.properties
```diff
@@ -56,7 +56,8 @@ jmhPluginVersion=0.6.4
 jmhShadowPluginVersion=6.1.0
 
 junitVersion=4.13.2
-junit5Version=5.7.1
+junitPlatformVersion=1.7.2
+junit5Version=5.8.0-M1
```
idelpivnitskiy, line 60:
```suggestion
junit5Version=5.7.1
```

Do you depend on any feature that is available only in 5.8.x version? If not, please don't change the version number. We usually submit separate PRs for version upgrades. In this case, don't think it's necessary. It's safer to stay with GA release instead of risking with M1.

servicetalk-data-jackson-jersey/src/test/java/io/servicetalk/data/jackson/jersey/JerseyDataJacksonTestSuite.java
```diff
@@ -17,18 +17,8 @@
...
-        SingleJsonResourcesTest.class,
-        PublisherJsonResourcesTest.class,
-        JsonMicroformatResourcesTest.class,
-        CustomJacksonSerializationProviderTest.class,
-        FilterStreamingJsonTest.class,
-        BaseJerseyRouterTestSuite.class
-})
-public class JerseyDataJacksonTestSuite {
+@SuiteDisplayName("Run JerseyRouterTestSuite with ServiceTalkJacksonSerializerAutoDiscoverable")
+public class JerseyDataJacksonTestSuite extends BaseJerseyRouterTestSuite {
```
idelpivnitskiy, line 23: 👁️ 🦅 🎖️

idelpivnitskiy (approved):
Massive work, thank you @danfaer!

gradle.properties
danfaer: yes, 5.8 adds support for test suits

idelpivnitskiy: Sounds good, reverting...

PR #1646, opened 2021-06-28
`Single#concat(Publisher)` potential demand deadlock fix

Scottmitch:
Motivation:
Single#concat(Publisher) may result in deadlock if demand comes after
the Single completes, and the next Publisher blocks in its subscribe
method. This may happen if blocking APIs are used (e.g.
ConnectablePayloadWriter) and the control flow involves
Single#concat(Publisher). ServiceTalk currently only uses this operator
for use in the Publisher#buffer(..) operator.

Modifications:
- SingleConcatWithPublisher#request(long) should deliver demand to super
  class before delivering downstream demand and subscribing. Ordering of
  downstream signals will be preserved because
  DelayedCancellableThenSubscription won't propagate demand upstream
  until onSubscribe is called.

Result:
If downstream requests sufficient demand it will be available to the
concat Publisher as soon as it is subscribed to, and therefore less
potential for deadlock when using blocking APIs and
Single#concat(Publisher).


idelpivnitskiy (approved):
🚀

idelpivnitskiy:
I will take this in and adjust #1643. Thank you for the test!

PR #1647, opened 2021-06-30
Removing deprecated methods from MultiAddressHttpClientBuilder

chemicL:
Motivation:

The overridden methods from BaseHttpBuilder have been deprecated and
it's safe to remove them. A `SingleAddressInitializer` serves the
purpose of configuring the individual address' clients.

Modifications:

- Removed inheritance of `HttpClientBuilder` and transitive
  `BaseHttpBuilder` from `MultiAddressHttpClientBuilder`,
- Removed deprecated methods.

Result:

The old methods are gone and the dependency on a base class meant for
single address builders is also removed. The new mechanism using
`SingleAddressInitializer` is the only correct way to configure
individual underlying clients used by `MultiAddressHttpClientBuilder`.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018-2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2018-2019, 2021 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,17 +15,8 @@
  */
 package io.servicetalk.http.api;
 
-import io.servicetalk.buffer.api.BufferAllocator;
-import io.servicetalk.client.api.AutoRetryStrategyProvider;
-import io.servicetalk.client.api.ConnectionFactoryFilter;
-import io.servicetalk.client.api.ServiceDiscoverer;
-import io.servicetalk.client.api.ServiceDiscovererEvent;
-import io.servicetalk.logging.api.LogLevel;
-import io.servicetalk.transport.api.IoExecutor;
+import io.servicetalk.transport.api.ExecutionContext;
 
-import java.net.SocketOption;
-import java.util.function.BooleanSupplier;
-import java.util.function.Function;
 import java.util.function.Predicate;
 
 import static io.servicetalk.http.api.StrategyInfluencerAwareConversions.toMultiAddressConditionalFilterFactory;
@@ -42,8 +33,7 @@ import static java.util.Objects.requireNonNull;
  * @param <R> the type of address after resolution (resolved address)
  * @see <a href="https://tools.ietf.org/html/rfc7230#section-5.3.2">absolute-form rfc7230#section-5.3.2</a>
  */
-public abstract class MultiAddressHttpClientBuilder<U, R>
-        extends HttpClientBuilder<U, R, ServiceDiscovererEvent<R>> {
+public abstract class MultiAddressHttpClientBuilder<U, R> {
     /**
      * Initializes the {@link SingleAddressHttpClientBuilder} for each new client.
      * @param <U> The unresolved address type.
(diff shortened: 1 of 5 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
```diff
@@ -336,4 +131,38 @@
      * @return {@code this}.
      */
     public abstract MultiAddressHttpClientBuilder<U, R> maxRedirects(int maxRedirects);
+
+    /**
+     * Builds a new {@link StreamingHttpClient}, using a default {@link ExecutionContext}.
+     *
+     * @return A new {@link StreamingHttpClient}
+     */
+    public abstract StreamingHttpClient buildStreaming();
```
chemicL, line 140: I considered adding a new interface just for these actual build methods. For now I went with copying, but if it makes sense, those methods could reside as default implementations in a new interface that is implemented by all other client builders. I'd appreciate opinions whether that makes sense.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -334,100 +319,19 @@ public StreamingHttpRequest newRequest(final HttpRequestMethod method, final Str
...
 
-    @Override
-    public MultiAddressHttpClientBuilder<HostAndPort, InetSocketAddress> unresolvedAddressToHost(
-            Function<HostAndPort, CharSequence> unresolvedAddressToHostFunction) {
-        this.unresolvedAddressToHostFunction = requireNonNull(unresolvedAddressToHostFunction);
-        return this;
-    }
-
     @Override
     public MultiAddressHttpClientBuilder<HostAndPort, InetSocketAddress> appendClientFilter(
```
chemicL, line 336: I am wondering whether this method should now become deprecated as this behaviour can be achieved using the `SingleAddressInitializer` and the code can be simplified.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
bondolo: Would the interface be just for the default implementations or would the interface be used in API, field or variable types?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
bondolo: If you have questions then no doubt others do as well. I prefer one mechanism especially when there is ambiguity about differences in behavior between multiple options.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
chemicL: HttpClientBuildFinalizer.java:
```
interface HttpClientBuildFinalizer {
    StreamingHttpClient buildStreaming();
    
    default HttpClient build() {
        return buildStreaming().asClient();
    }
    
    default BlockingStreamingHttpClient buildBlockingStreaming() {
        return buildStreaming().asBlockingStreamingClient();
    }
    
    default BlockingHttpClient buildBlocking() {
        return buildStreaming().asBlockingClient();
    }
}
```

used in MultiAddressHttpClientBuilder.java:

```
public abstract class MultiAddressHttpClientBuilder<U, R> implements HttpClientBuildFinalizer {
    // ...
}
```

implemented method in DefaultMultiAddressUrlHttpClientBuilder.java:
```
@Override
    public StreamingHttpClient buildStreaming() {
        // ...
    }
```

For other builders I believe it would be the same. So no, no exposure apart from the methods that are already there with those same signatures. Only binary compatibility breakage from this change's perspective.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
chemicL: What I can suggest now:
1. create a separate PR: add `@Deprecated` annotation, leave existing logic, release soon and let users digest the change
1. merge this PR in a breaking-change release
1. another PR, another release removes the recently `@Deprecated`  method and cleans up the logic

Points 2. and 3. can be merged if we anyways have to wait until the deprecation is released.

idelpivnitskiy: +1, those also have to be deprecated, including `MultiAddressHttpClientFilterFactory` interface. We can cherry-pick (1) into 0.41 branch and do (2) and (3) together in this PR.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
```diff
@@ -54,8 +55,9 @@ void buildWithDefaults() throws Exception {
     @Test
     void buildAggregatedWithDefaults() throws Exception {
         HttpRequester newAggregatedRequester = HttpClients.forMultiAddressUrl()
-                .ioExecutor(CTX.ioExecutor())
-                .executionStrategy(defaultStrategy(CTX.executor()))
+                .initializer((scheme, address, builder) -> builder
+                        .ioExecutor(CTX.ioExecutor())
```
idelpivnitskiy, line 59: Hm... this becomes interesting. Because there is no single `IoExecutor` / `ExecutionStrategy` / `BufferAllocator` for all internal clients anymore, not sure what resulting `HttpClient#executionContext()` should return for multi-address client. We cannot return `UnsupportedOperationException` because `executionContext()` is frequently used to access `BufferAllocator`. @Scottmitch wdyt?

chemicL:
This PR can be discarded – a follow up PR addresses the concerns expressed here: https://github.com/apple/servicetalk/pull/1721

PR #1649, opened 2021-06-30
Remove publishAndSubscribeOn

bondolo:
Motivation:
The previously deprecated `publishAndSubscribeOn` operators should be
removed.
Modifications:
Removes the deprecated `publishAndSubscribeOn`operators. Tests using
these operators are updated.
Result:
Deprecated operators removed.


Scottmitch (approved):

bondolo:
For sequencing reasons the deprecation change will be applied first.

PR #1650, opened 2021-07-01
`Publisher#buffer` doesn't emit if count is met or original terminates

### LINKED ISSUE #1048: Publisher#buffer may delay emissions
Current approach of `Publisher#buffer()` only emits when a boundary is emitted from the boundaries `Publisher`. The motivation is to simplify the implementation trading-off emission delays if an item is emitted at the same time when a boundary is emitted. Following is the sequence of events that will delay emission:

```
Publisher<T> original
Publisher<A> boundaries
Publisher<B> buffered

Thread 1: emits an item on original, changes state to ADDING

Thread 2: emits an Accumulator on boundaries , changes state from ADDING to the emitted Accumulator, returns

Thread 1: adds item to old Accumulator and sets back the Accumulator to the state. This delays emissions of old Accumulator

Thread 3: emits and Accumulator on boundaries, finishes and emits the old Accumulator set by Thread 1 above.
```

We can optimize this behavior by handing-off emission to Thread 1 but this will complicate the state machine as `Accumulator#add()` can take arbitrarily long times and we may have multiple boundary emissions during that time.

### LINKED ISSUE #1168: Publisher#buffer doesn't emit if count is meet
`Publisher#buffer` operator doesn't emit if the count is the first boundary to be exceeded.

```java
    @Test
    public void emitJustCount() throws Exception {
        Integer[] expected = new Integer[] {1, 2, 3};
        assertThat(Publisher.from(expected)
                .buffer(BufferStrategies.forCountOrTime(expected.length, Duration.ofDays(1)))
                .toFuture().get(), contains(expected));
    }
```

### LINKED ISSUE #1259: BufferStrategiesTest.sizeOrDurationConcurrent test failure
https://github.com/apple/servicetalk/runs/1521278514

```
2020-12-09T01:47:11.8036134Z     "Time-limited test" #41 waiting on java.util.concurrent.CountDownLatch$Sync@27c94c64
2020-12-09T01:47:11.8037423Z       java.lang.Thread.State: WAITING
2020-12-09T01:47:11.8038496Z     	  at java.base@11.0.9/jdk.internal.misc.Unsafe.park(Native Method)
2020-12-09T01:47:11.8040231Z     	  at java.base@11.0.9/java.util.concurrent.locks.LockSupport.park(LockSupport.java:194)
2020-12-09T01:47:11.8044291Z     	  at java.base@11.0.9/java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:885)
2020-12-09T01:47:11.8050637Z     	  at java.base@11.0.9/java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1039)
2020-12-09T01:47:11.8086901Z     	  at java.base@11.0.9/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1345)
2020-12-09T01:47:11.8091154Z     	  at java.base@11.0.9/java.util.concurrent.CountDownLatch.await(CountDownLatch.java:232)
2020-12-09T01:47:11.8095061Z     	  at app//io.servicetalk.concurrent.api.BufferStrategiesTest.sizeOrDurationConcurrent(BufferStrategiesTest.java:117)
2020-12-09T01:47:11.8099116Z     	  at java.base@11.0.9/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
2020-12-09T01:47:11.8131797Z     	  at java.base@11.0.9/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
2020-12-09T01:47:11.8135694Z     	  at java.base@11.0.9/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
2020-12-09T01:47:11.8138297Z     	  at java.base@11.0.9/java.lang.reflect.Method.invoke(Method.java:566)
2020-12-09T01:47:11.8140151Z     	  at app//org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
2020-12-09T01:47:11.8142993Z     	  at app//org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
2020-12-09T01:47:11.8147789Z     	  at app//org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
2020-12-09T01:47:11.8150667Z     	  at app//org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
2020-12-09T01:47:11.8153366Z     	  at app//org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
2020-12-09T01:47:11.8156781Z     	  at app//io.servicetalk.concurrent.internal.ServiceTalkTestTimeout$TimeoutStatement$CallableStatement.call(ServiceTalkTestTimeout.java:171)
2020-12-09T01:47:11.8169370Z     	  at app//io.servicetalk.concurrent.internal.ServiceTalkTestTimeout$TimeoutStatement$CallableStatement.call(ServiceTalkTestTimeout.java:163)
2020-12-09T01:47:11.8172489Z     	  at java.base@11.0.9/java.util.concurrent.FutureTask.run(FutureTask.java:264)
2020-12-09T01:47:11.8174123Z     	  at java.base@11.0.9/java.lang.Thread.run(Thread.java:834)
```

idelpivnitskiy:
Motivation:

1. `Publisher#buffer(BufferStrategy)` doesn't emit if count is met when
`BufferStrategies#forCountOrTime` is used. It happens because the count
based boundary is emitted from inside `accumulate` on the same thread
while we `ADDING`. As the result, this new boundary is discarded.
2. It also doesn't emit if original publisher terminates before the count
boundary met. Because it won't emit any new items, count boundary will
never be triggered.
3. If the downstream subscriber requests only 1 accumulation, internal
state machine drains the original publisher till termination and accumulates
all values internally.

Modifications:
- Adjust `PublisherBuffer` internal `State` to accommodate described
use-cases: use `NextAccumulatorHolder` to notify thread that does
accumulation that a new boundary was emitted, ignore new boundaries
if more than one boundary was emitted while accumulating, use
`ItemsTerminated` to notify thread that processes `nextAccumulator` if
original publisher terminates while `nextAccumulator` delivers `onNext`;
- Inside the internal `State` account for the demand to avoid requesting
more items from the original publisher and to avoid delivering the last
boundary without demand;
- Improve javadoc for `Publisher#buffer(BufferStrategy)` and 
`BufferStrategy#bufferSizeHint()`;
- Add more tests to verify new state machine and different
`BufferStrategies`.

Result:

Fixes #1048.
Fixes #1168.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -249,57 +290,111 @@ public void onComplete() {
...
+                    if (maybeAccumulatorUpdater.compareAndSet(this, cMaybeAccumulator, itemsTerminated)) {
+                        return;
+                    }
+                } else if (maybeAccumulatorUpdater.compareAndSet(this, cMaybeAccumulator, TERMINATED)) {
+                    assert cMaybeAccumulator instanceof CountingAccumulator;
+                    // Deliver the last boundary only if there are some items pending and demand is present
+                    @SuppressWarnings("unchecked")
+                    final CountingAccumulator<T, B> accumulator = (CountingAccumulator<T, B>) cMaybeAccumulator;
+                    terminateTarget(accumulator.size() > 0 && pendingUpdater.decrementAndGet(this) >= 0 ?
+                            accumulator : null, target, terminalNotification, bCancellable);
```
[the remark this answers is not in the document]
tkountis: I was experimenting with the work and this test doesn't behave as expected

```
    @Test
    void test() throws InterruptedException {
        final AtomicInteger idSeq = new AtomicInteger();
        final CountDownLatch completion = new CountDownLatch(1);

        Publisher<Integer> tPublisher = Publisher.range(1, 1000);
        Publisher<Accumulator<Integer, Integer>> boundaries = new Publisher<Accumulator<Integer, Integer>>() {
            @Override
            protected void handleSubscribe(final Subscriber<? super Accumulator<Integer, Integer>> subscriber) {
                subscriber.onSubscribe(new Subscription() {
                    @Override
                    public void request(final long n) {
                        int id = idSeq.incrementAndGet();
                        subscriber.onNext(new Accumulator<Integer, Integer>() {

                            private int sum;

                            @Override
                            public void accumulate(@Nullable final Integer item) {
                                System.err.println("Accumulating on " + id + " " + item + " sum: " + sum);
                                sum += item;
                            }

                            @Override
                            public Integer finish() {
                                System.err.println("Sum for " + id + " is: " + sum);
                                return sum;
                            }
                        });
                    }

                    @Override
                    public void cancel() {
                        //ignore
                    }
                });
            }
        };

        BlockingQueue<Integer> buffers = new LinkedBlockingDeque<>();
        toSource(tPublisher.buffer(new TestBufferStrategy(boundaries, 3)))
                .subscribe(new Subscriber<Integer>() {
                    @Override
                    public void onSubscribe(Subscription s) {
                        s.request(1);
                    }

                    @Override
                    public void onNext(@Nullable Integer integer) {
                        assert integer != null;
                        System.err.println("Downstream received " + integer);
                        buffers.add(integer);
                    }

                    @Override
                        t.printStackTrace();
                    }

                    @Override
                    public void onComplete() {
                        completion.countDown();
                    }
                });

        completion.await();
    }
```

If you only return 1 boundary it works, if you allow it to return 2 (as is now) then the 1st boundary will return a 0 sum, and the 2nd one doesn't emit (even though it *does* accumulate).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -249,57 +290,111 @@ public void onComplete() {
...
-                        ItemsTerminated<T, B> itemsTerminated = (ItemsTerminated<T, B>) cMaybeAccumulator;
-                        terminateTarget(itemsTerminated.accumulator, target, itemsTerminated.terminalNotification,
-                                bSubscription);
+                    if (maybeAccumulatorUpdater.compareAndSet(this, ADDING,
+                            new NextAccumulatorHolder<>(nextAccumulator))) {
+                        // Thread that is ADDING will observe the change and emit the current accumulator
                         return;
                     }
+                } else if (NextAccumulatorHolder.class.equals(cMaybeAccumulator.getClass())) {
+                    // Hand-off emission of one more nextAccumulator to the thread that is still ADDING, discarding it.
```
tkountis, line 308: Similarly here about discarding emitted boundaries. I am not saying we should fix it, but I believe at the very least we should include some warning in the javadoc, maybe: 
"boundaries should *not* be used as lifecycle aware objects, `finish` is not guaranteed to be called."

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -249,57 +290,111 @@ public void onComplete() {
...
+                                    // Keep "prev" state because we observed `itemsTerminated` or `accumulate` took
+                                    // ownership of the `holder`
+                                    toCounting(nextAccumulator), (prev, next) -> prev == holder ? next : prev);
+                            if (ItemsTerminated.class.equals(nextState.getClass())) {
+                                maybeAccumulator = TERMINATED;
+                                @SuppressWarnings("unchecked")
+                                final ItemsTerminated<T, B> it = (ItemsTerminated<T, B>) nextState;
+                                // Deliver the last boundary only if there are some items pending and demand is present
+                                final Accumulator<T, B> accumulator = it.accumulator.size() > 0 &&
+                                        pendingUpdater.decrementAndGet(this) >= 0L ? it.accumulator : null;
```
Scottmitch, line 329: if we have data queued and we received a terminal event, we should deliver the queued elements when we get more demand.

```diff
diff --git a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherBufferTest.java b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherBufferTest.java
index acf6b8a4c..b951a05ff 100644
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherBufferTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherBufferTest.java
@@ -25,6 +25,7 @@ import org.hamcrest.Matcher;
 import org.junit.jupiter.api.Test;
 
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
@@ -48,17 +49,63 @@ class PublisherBufferTest {
 
     private static final int EMPTY_ACCUMULATOR_VAL = -1;
     static final int BUFFER_SIZE_HINT = 8;
-    private final TestPublisher<Integer> original = new TestPublisher<>();
-    private final TestPublisher<Accumulator<Integer, Integer>> boundaries = new TestPublisher<>();
+    private final TestSubscription tSubscription = new TestSubscription();
+    private final TestPublisher<Integer> original = new TestPublisher.Builder<Integer>().disableAutoOnSubscribe()
+            .build(sub -> {
+                sub.onSubscribe(tSubscription);
+                return sub;
+            });
+    private final TestSubscription bSubscription = new TestSubscription();
+    private final TestPublisher<Accumulator<Integer, Integer>> boundaries =
+            new TestPublisher.Builder<Accumulator<Integer, Integer>>().disableAutoOnSubscribe().build(sub -> {
+                sub.onSubscribe(bSubscription);
+                return sub;
+            });
     private final TestPublisherSubscriber<Integer> bufferSubscriber = new TestPublisherSubscriber<>();
 
-    PublisherBufferTest() {
+    PublisherBufferTest() throws Exception {
         toSource(original.buffer(new TestBufferStrategy(boundaries, BUFFER_SIZE_HINT))).subscribe(bufferSubscriber);
         bufferSubscriber.awaitSubscription().request(1); // get first boundary
+        bSubscription.awaitRequestN(1);
         boundaries.onNext(new SumAccumulator());
         assertThat(bufferSubscriber.pollOnNext(10, MILLISECONDS), is(nullValue()));
     }
 
+    @Test
+    void terminalBeforeRequestedDelivers() throws Exception {
+        Executor executor = Executors.newCachedThreadExecutor();
+        try {
+            Future<Void> f = executor.submit(() -> {
+                try {
+                    tSubscription.awaitRequestN(3);
+                } catch (Exception e) {
+                    throw new AssertionError(e);
+                }
+
+                original.onNext(1);
+
+                try {
+                    bSubscription.awaitRequestN(2);
+                } catch (Exception e) {
+                    throw new AssertionError(e);
+                }
+                boundaries.onNext(new SumAccumulator());
+
+                original.onNext(2);
+                original.onComplete();
+            }).toFuture();
+
+            assertThat(bufferSubscriber.takeOnNext(), is(1));
+
+            f.get();
+            bufferSubscriber.awaitSubscription().request(1);
+            assertThat(bufferSubscriber.takeOnNext(), is(2));
+            bufferSubscriber.awaitOnComplete();
+        } finally {
+            executor.closeAsync().toFuture().get();
+        }
+    }
+
     @Test
     void invalidBufferSizeHint() {
         TestPublisherSubscriber<Integer> bufferSubscriber = new TestPublisherSubscriber<>();

```

Scottmitch (changes requested):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BufferStrategies.java
```diff
@@ -290,6 +299,10 @@ public void accumulate(@Nullable final T item) {
         public B finish() {
             return delegate.finish();
         }
+
+        int size() {
```
Scottmitch, line 303: looks like we could just expose `isEmpty()` as we don't really need the size.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BufferStrategies.java
```diff
@@ -279,9 +287,10 @@ void beforeNewAccumulatorEmitted(final Accumulator<T, B> newAccumulator) {
 
         @Override
         public void accumulate(@Nullable final T item) {
-            size++;
+            size = addWithOverflowProtection(size, 1);
```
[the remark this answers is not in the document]
idelpivnitskiy: In this case, if the `sizeThreshold == 1`, until the next `accumulate` is invoked. After thinking about it more, don't think that we need any protection because after `countThresholdBreached` is invoked, we do not expect new invocations of `accumulate` method. I will add an assertion, but may change it to `IllegalStateException` if that will make more sense.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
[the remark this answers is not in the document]
idelpivnitskiy: Discussed offline:
1. The last empty accumulation is discarded because a user usually does not request it. Internally we request at least +1 boundary from what the user requested. It helps to avoid confusion when users accumulate 3 items of `Publisher.from(1, 2, 3)`. Instead of receiving 2 accumulations, they will get one.
2. This is a bit different compare to empty accumulations in between because those are usually driven by a timer. With count-based accumulation users won't see empty accumulations.
3. Clarified in javadoc under what conditions boundaries can be discarded and that the `finish` method won't be invoked.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
idelpivnitskiy: Clarified in javadoc.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BufferStrategies.java
```diff
@@ -279,9 +286,14 @@ void beforeNewAccumulatorEmitted(final Accumulator<T, B> newAccumulator) {
 
         @Override
         public void accumulate(@Nullable final T item) {
-            size++;
+            if (sizeThreshold > 0 || isEmpty()) {
+                // Increment `size` only when the `sizeThreshold` is used or for the first accumulation
```
Scottmitch, line 290: Can we simplify this logic and just unconditionally increment? The conditional logic is more involved to understand how it relates to the other state and the `delegate`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -249,97 +302,182 @@ public void onComplete() {
...
-                        return;
-                    }
+                // decrementAndGet alternative without update of the MIN_VALUE state:
+                final long pending = pendingUpdater.accumulateAndGet(this, 1,
+                        (prev, next) -> prev == MIN_VALUE ? prev : prev - 1);
+                assert pending >= 0;
+                target.onNext(accumulator.finish());
+                return true;
+            } catch (Throwable t) {
+                terminate(t, target, bCancellable, tCancellable);
```
Scottmitch, line 409: if you are in the subscriber thread the exception typically propagates up to the source, and the source sends a terminal `onError` signal back down (instead of forcing a cancel). what is done here won't leak resources bcz we the corresponding subscription, however the control flow to upstream operators won't be the same (e.g. if there is a retry/recover they won't run bcz we force a cancel inside this operator).

if we do let the exception propagate then we will have to be careful the state machine allows "re-entry" for error delivery.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -249,97 +302,182 @@ public void onComplete() {
...
+                        terminateTarget(null, target, terminalNotification);
+                        bCancellable.cancel();
+                        return;
+                    }
+                } else {
+                    @SuppressWarnings("unchecked")
+                    final CountingAccumulator<T, B> accumulator = (CountingAccumulator<T, B>) cMaybeAccumulator;
+                    if (maybeAccumulatorUpdater.compareAndSet(this, cMaybeAccumulator,
+                            new ItemsTerminated<>(accumulator, terminalNotification))) {
+                        bCancellable.cancel();
```
Scottmitch, line 384: above the cancel is done after the downstream termination. can you be consistent, and put this in a `try/finally` (so we always terminate)?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -249,97 +302,182 @@ public void onComplete() {
...
+                        NextAccumulatorHolder.class.equals(cMaybeAccumulator.getClass())) {
+                    // Terminated while `nextAccumulator` delivers onNext, notify that thread using ItemsTerminated
+                    // wrapper. There is no more concurrency because neither `itemsTerminated` nor `accumulate` will be
+                    // invoked again.
+                    @SuppressWarnings("unchecked")
+                    final ItemsTerminated<T, B> itemsTerminated = new ItemsTerminated<>(
+                            ((NextAccumulatorHolder<T, B>) cMaybeAccumulator).accumulator, terminalNotification);
+                    if (maybeAccumulatorUpdater.compareAndSet(this, cMaybeAccumulator, itemsTerminated)) {
+                        bCancellable.cancel();
+                        return;
```
Scottmitch, line 370: if errors from the subscriber thread are allowed to propagate upstream (see other comment), we may need to propagate the terminal here.

```
1. (subscriber thread) subscriber.onNext -> CAS(accumulator, ADDING)
2. (boundary thread) boundary.onNext -> CAS(ADDING, NextAccumulator)
3. (subscriber thread) target.onNext -> throws e
4. (subscriber thread) subscriber.onError(e) -> /* we should deliver the terminal event here */
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BufferStrategies.java
idelpivnitskiy: Ok, I will keep the `size = addWithOverflowProtection(size, 1);` then

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
idelpivnitskiy: There is a case when we cancel in the current thread (line 369), but deliver terminal from another thread that does `nextAccumulator`. I can keep it consistent and always cancel boundaries before termination. Is you suggestion is to move termination to a `finally` block in case `bCancellable.cancel()` throws?

https://github.com/apple/servicetalk/pull/1650/files/d1d008ff9a19a8364e5114efcbcbf8edc79576f7#diff-a02114146aee33b0d470d21bdd07e3e9017f55b6aea426a264ee4c8d62a6efaaR360-R370

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BufferStrategies.java
idelpivnitskiy: Discussed offline, this does not require any protection due to how we manage the demand for items publisher. Changed to `++size`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
idelpivnitskiy: I made cancellation consistent and always cancel boundaries before propagating termination. Also added `try-finally` block to make sure the terminal event is propagated.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
idelpivnitskiy: Makes sense, let exceptions to propagate and added a test with retry: https://github.com/apple/servicetalk/pull/1650/commits/99c1d6dd60593cbbc7382fede3b721c841f29c1b

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
idelpivnitskiy: Not necessary in this case. If the `onNext` throws, I change the state to `Accumulator`. So, this `if` won't be used in the described use-case.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -77,91 +80,120 @@ public void onComplete() {
...
                     } finally {
                         tSubscription.cancel();
                     }
                 }
+            }));
+            // Wrap bSubscription to count number of requested items excluding internal demand for discarded boundaries.
+            target.onSubscribe(new Subscription() {
+                @Override
+                public void request(final long n) {
+                    final boolean needRequestItems = state.requested(n, target);
```
Scottmitch, line 180: any time there is invalid demand we should propagate it upstream (which is expected to propagate an error back downstream), to avoid accounting for negative values in the `pendingUpdater` state machine you can do something like this:

```java
@Override
                public void request(final long n) {
                    if (SubscriberUtils.isRequestNValid(n)) {
                        final boolean needRequestItems = state.requested(n, target);
                        bSubscription.request(n);
                        if (needRequestItems) {
                            tSubscription.request(state.itemsRequestN);
                        }
                    } else {
                        try {
                            bSubscription.request(n);
                        } finally {
                            tSubscription.request(n);
                        }
                    }
                }
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -190,57 +222,113 @@ public void onComplete() {
...
+
+        <T, B> boolean requested(long n, final Subscriber<? super B> target) {
+            final long oldPending = pendingUpdater.getAndAccumulate(this, n, (prev, nValue) -> {
+                if (nValue <= 0L || prev == NEED_TERMINATE) {
+                    return prev;
+                }
+                return prev == NEED_REQUEST_ITEMS ? nValue : addWithOverflowProtection(prev, nValue);
+            });
+            if (oldPending == NEED_TERMINATE && n > 0L) {
+                @SuppressWarnings("unchecked")
```
Scottmitch, line 270: what if we are in `NEED_TERMINATE` state and someone does `request(1); request(1);`? we should check `ItemsTerminated instance of ItemsTerminated ` to avoid class cast (we should be careful to avoid duplicate termination too).

add an additional `bufferSubscriber.awaitSubscription().request(1);` to `terminalBeforeRequestedDelivers` (line 571) to test this issue.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -77,91 +80,120 @@ public void onComplete() {
...
             this.tSubscription = tSubscription;
         }
 
         @Override
-        public void onSubscribe(final Subscription bSubscription) {
-            subscription = wrap(new Subscription() {
+        public void onSubscribe(final Subscription subscription) {
+            // We may cancel from multiple threads and DelayedSubscription will atomically swap if a cancel occurs but
+            // it will not prevent concurrent access between request(n) and cancel() on the original subscription.
+            bSubscription.delayedSubscription(ConcurrentSubscription.wrap(new Subscription() {
```
Scottmitch, line 161: `tSubscription` is already wrapped in `ConcurrentSubscription` did you consider just wrapping `subscription` here?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -190,57 +222,113 @@ public void onComplete() {
...
-                    // next boundary emission.
-                    maybeAccumulatorUpdater.accumulateAndGet(this, accumulator,
-                            (prev, next) -> prev == TERMINATED ? TERMINATED : next);
+                    final Object nextState = maybeAccumulatorUpdater.accumulateAndGet(this, accumulator,
+                            (prev, next) -> prev == ADDING ? next : prev);
+                    if (nextState == accumulator || nextState == TERMINATED) {
+                        return;
+                    }
+                    // Received the next boundary while adding, deliver current accumulator and unwrap the next boundary
+                    assert nextState instanceof NextAccumulatorHolder : "Expected next boundary";
```
Scottmitch, line 317: nit: we will get a class cast on the next line anyways, consider removing this assertion.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
```diff
@@ -249,97 +337,175 @@ public void onComplete() {
...
-            try {
-                if (accumulator != null) {
-                    try {
-                        target.onNext(accumulator.finish());
-                    } catch (Throwable t) {
-                        safeOnError(target, t);
-                        return;
-                    }
+        private <T, B> void deliverOnNext(final Accumulator<T, B> accumulator, final Subscriber<? super B> target) {
+            // an alternative of `decrementAndGet` without update of the special states:
```
Scottmitch, line 445: did you consider adjusting the states so we can just use `decrementAndGet` as a light weight alternative? For example would using `-1` to indicate "needs more demand", `-2` would be "got demand and delivered" (only set here), and `-3+` be terminated work? worth consideration since this is invoked on each `onNext`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
Scottmitch: got it, seems like we handle it when the stack unwinds iiiuc.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherBuffer.java
idelpivnitskiy: The only difference between `accumulateAndGet` and `decrementAndGet` is the ternary operator for `prev <= LAST_SPECIAL_STATE` check which I believe is not a bottleneck.
Let me keep it as-is. Don't want to mess with another state transitioning in my head and it will be safer to use `Long.MIN_VALUE` range for the special states to avoid situations when boundaries do not respect demand and may deliver more than requested.

idelpivnitskiy:
@tkountis @Scottmitch thanks for the provided review and all corner cases. I believe all comments were addressed. If any of them are missed, I can submit a follow-up

Issue #1652, opened 2021-07-02
`SplittingWriteEventsListener` expects that an async source limits recursion depth to 1


idelpivnitskiy:
`SplittingWriteEventsListener` expects the following sequence of HTTP messages:
```
meta-data → Start  → Payload → InProgress → trailers → End
```

If the source does not limit recursion to a depth of 1 when it delivers data (see [PublisherFrom2TckTest](https://github.com/apple/servicetalk/blob/4bf0a96991286abb350ff10f9bd4f8bd8a6041c2/servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/PublisherFrom2TckTest.java#L38) and [PublisherFrom3TckTest](https://github.com/apple/servicetalk/blob/4bf0a96991286abb350ff10f9bd4f8bd8a6041c2/servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/PublisherFrom3TckTest.java#L38)), the sequence of events can be any, depending on how `WriteDemandEstimator` requests items.

Steps to reproduce:
1. Change `WriteDemandEstimators#newDefaultEstimator()` to always return 1L from `estimateRequestN`.
2. Run `ContentLengthAndTrailersTest`

By playing with `WriteDemandEstimator` you can achieve any sequence of events. Examples:
```
meta-data → trailers → End → Start
meta-data  → Payload → trailers → End  → InProgress → Start
meta-data  → Payload → InProgress → Start → Payload → trailers → End  → InProgress
```

Because the sequent can be any, `SplittingWriteEventsListener` and `CountingFlushStrategyProvider` can not correctly detect message boundaries. As the result, data may hang in the socket buffer and will never be flushed.

This is less problematic when task-based offloading is enabled because each onNext will be on a different task thread, but causes issues when offloading is disabled.

Workaround:  `flushOnEach()` strategy, which will trigger the flush for any item, regardless of the boundary. Users can opt-out from our internal flush strategy optimizations by using streaming API or applying `transform` operation on the HTTP message in a filter (it will reset the "aggregated" and "empty" flags).

Ways to reduce likelihood of this issue for `flushOnEnd()`:
- Limit recursion depth to 1 for all async sources and operators.

PR #1654, opened 2021-07-02
`SingleConcatWithPublisher`: limit recursion depth to 1

idelpivnitskiy:
Motivation:

To mitigate #1652 issue, temporary limit the mutual recursion between
`onNext` and `request` to a depth of 1 for `ConcatDeferNextSubscriber`.

Modifications:

- Introduce additional `SINGLE_DELIVERING` state to understand if the
`request` was invoked while we deliver result of the `Single`;
- Remove `boundedDepthOfOnNextAndRequestRecursion()` override from
`SingleConcatWithPublisherDeferSubscribeTckTest`;

Result:

1. `SingleConcatWithPublisher` has a maximum recursion depth of 1.
2. Less risk of encountering issue #1652.


idelpivnitskiy:
This was already reviewed in #1643. We agreed with @Scottmitch to move this change into a separate PR to revert it later after #1652 is fixed. Will merge after CI passes all checks.

PR #1656, opened 2021-07-02
Fixes issues with GRPC Examples documentation

bondolo:
Motivation:
Documentation validation fails due to several errors in the GRPC
Examples documentation. These problems should be fixed.
Modifications:
Link targets are adjusted to pass validation.
Result:
Validation succeeds.


idelpivnitskiy (approved):
❤️

PR #1657, opened 2021-07-03
Remove SNAPSHOT from validation path for docs

bondolo:
Motivation:
When validating local docs validate everything but only validate
the current SNAPSHOT when validating remote docs. Remote docs
likely include legacy docs which will fail validation but cannot
be updated.
Modifications:
HTML files path is now customized based on validation location.
Result:
Cleaner, narrower validation.


idelpivnitskiy (approved):
🚀

tkountis (approved):

PR #1661, opened 2021-07-07
Add `awaitSubscribed()` to test async sources

bondolo:
Motivation:
For some tests it is useful to be able to wait the source being
subscribed.
Modifications:
add `awaitSubscribed()` method to `TestCompletable`, `TestSingle`, and
`TestPublisher`. Shared code is moved to new `Utils` class. Adds features
to `TestExecutor.execute(Runnable)` to name the thread and adversarially
clear the AsyncContextMap while running tasks.
Result:
Additional testing capability.

### DIFF
--- a/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCompletable.java
+++ b/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCompletable.java
@@ -17,7 +17,7 @@ package io.servicetalk.concurrent.api;
 
 import io.servicetalk.concurrent.Cancellable;
 import io.servicetalk.concurrent.CompletableSource;
-import io.servicetalk.concurrent.test.internal.Utils;
+import io.servicetalk.concurrent.test.internal.AwaitUtils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -82,7 +82,7 @@ public final class TestCompletable extends Completable implements CompletableSou
      * {@link Thread#isInterrupted()} will be set upon return.
      */
     public void awaitSubscribed() {
-        Utils.awaitUninterruptibly(subscriberLatch);
+        AwaitUtils.awaitUninterruptibly(subscriberLatch);
     }
 
     @Override
--- a/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCompletable.java
+++ b/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCompletable.java
@@ -17,12 +17,14 @@ package io.servicetalk.concurrent.api;
 
 import io.servicetalk.concurrent.Cancellable;
 import io.servicetalk.concurrent.CompletableSource;
+import io.servicetalk.concurrent.test.internal.Utils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.function.Function;
@@ -53,6 +55,7 @@ public final class TestCompletable extends Completable implements CompletableSou
     private final Function<Subscriber, Subscriber> subscriberFunction;
     private final List<Throwable> exceptions = new CopyOnWriteArrayList<>();
     private volatile Subscriber subscriber = new WaitingSubscriber();
+    private final CountDownLatch subscriberLatch = new CountDownLatch(1);
 
     /**
      * Create a {@code TestCompletable} with the defaults. See <b>Defaults</b> section of class javadoc.
@@ -74,6 +77,14 @@ public final class TestCompletable extends Completable implements CompletableSou
         return !(subscriber instanceof WaitingSubscriber);
     }
 
+    /**
+     * Awaits until this {@link TestCompletable} is subscribed, even if interrupted. If interrupted the
+     * {@link Thread#isInterrupted()} will be set upon return.
+     */
+    public void awaitSubscribed() {
+        Utils.awaitUninterruptibly(subscriberLatch);
+    }
+
     @Override
     protected void handleSubscribe(final Subscriber subscriber) {
         try {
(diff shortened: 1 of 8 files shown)


servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
```diff
@@ -314,8 +317,14 @@ private static boolean executeOne(Queue<RunnableWrapper> tasks, AtomicInteger ta
...
 
-    // Wraps Runnables to ensure that object-equality (and hashcode) is used for removal from Lists.
-    // Also ensures a unique object each time, so the same Runnable can be executed multiple times.
+    /**
+     *  Wraps Runnables to ensure that object-equality (and hashcode) is used for removal from Lists.
+     *  Also ensures a unique object each time, so the same Runnable can be executed multiple times.
+     *  Sets the thread name to {@code TestExecutor} while running the task so that capturing the thread name makes
+     *  sense and during debugging the execution context is more obvious.
+     *  Pessimistically set the {@link AsyncContextMap} to null to ensure that any use of {@link AsyncContextMap}
+     *  within the context of the Runnable includes appropriate setting/restoring.
```
idelpivnitskiy, line 326: I'm not sure about this change. We expect `AsyncContext` to be propagated through the whole reactive chain, regardless of on which executor each operator is executed. Will this change the behavior? Can we reuse `ContextPreservingRunnable` in this case to be consistent with other executors?

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
```diff
@@ -325,7 +334,29 @@ private RunnableWrapper(final Runnable delegate) {
 
         @Override
         public void run() {
-            delegate.run();
+            Thread current = Thread.currentThread();
+            String oldName = current.getName();
+            current.setName("TestExecutor");
```
idelpivnitskiy, line 339: Consider adding a `factoryCount` for each new instance of `TestExecutor`, similar to how we generate names in `DefaultThreadFactory`. The `threadCount` probably is not necessary, but the `factoryCount` can help to distinguish between multiple instances of `TestExecutor`.

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/Utils.java
```diff
@@ -0,0 +1,113 @@
...
+package io.servicetalk.concurrent.test.internal;
+
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+public final class Utils {
```
idelpivnitskiy, line 25: Consider `AwaitUtils` or `ConcurrentUtils`

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/Utils.java
```diff
@@ -0,0 +1,113 @@
...
+                }
+            } while (true);
+        } finally {
+            if (interrupted) {
+                Thread.currentThread().interrupt();
+            }
+        }
+    }
+
+    public static <T> T takeUninterruptibly(BlockingQueue<T> queue) {
```
idelpivnitskiy, line 72: `takeUninterruptibly` and `pollUninterruptibly` is not used by any other code, except `TestPublisherSubscriber`. Can we keep them there for now and share only when necessary?

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCompletable.java
```diff
@@ -74,6 +77,14 @@ public boolean isSubscribed() {
         return !(subscriber instanceof WaitingSubscriber);
     }
 
+    /**
+     * Awaits until this {@link TestCompletable} is subscribed, even if interrupted. If interrupted the
+     * {@link Thread#isInterrupted()} will be set upon return.
+     */
+    public void awaitSubscribed() {
```
idelpivnitskiy, line 84: Can you please elaborate the use-case? `subscribe` is a synchronous operation. Usually this await is not required because the source is subscribed as soon as you do `subscribe`. Is it for the case when you initiate subscribe from another thread?

idelpivnitskiy (commented):
It's ok to keep this PR as-is. For future, please send separate PRs for unrelated changes. Otherwise, it's hard to understand from PR title what was changed.

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
bondolo: The intent of this change is to clear the current async context from the thread which will execute the `Runnable` to ensure that if the `Runnable` does depend upon `AsyncContext` it is properly wrapped in `ContextPreservingRunnable`. It is an "adversarial" testing behaviour intended to simulate what an offloaded task would experience if it were run on a fresh thread. Alternately we could also add a new `AsyncContextMap` implementation which asserts if any of the `get()`/`put()` methods are called rather than (potentially) a new `AsyncContextMap` being created as happens under some scenarios.

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCompletable.java
bondolo: The `subscribe()` is being offloaded because of the `subscribeOn()` operator. The thread which initiate the `subscribe()` regains control before the `subscribe` completes on the offloaded thread. `awaitSubscribed()` allows it to wait until the subscribe is completed.

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/Utils.java
bondolo: It seemed re-use would be more likely by exposing them. This is still only in the test-internal module so not part of the public API.

Is the motivation for keeping them local based on YAGNI, the behaviour of the methods (which some may disagree with), or other reasons?

idelpivnitskiy: The only motivation was to not move code until necessary to minimize git changes, but either way works fine for me in this case.

idelpivnitskiy (approved):
Thank you!

PR #1662, opened 2021-07-09
Enforce correct AsyncContext use and name "thread"

bondolo:
Motivation:
`TestExecutor` can improve the testing of the code under test by
requiring that tasks it runs correctly handle saving and restoring of
the `AsyncContext`. It would also be useful to set the thread name
so that it can be captured by the tasks and to make the execution
context clear for logging or debugging.
Modifications:
Before any task is run the active (if any) `AsyncContext` is saved and
a special hostile instance is substituted that detects improper
access. The name of the executing thread is also adjusted while running
the task to identify the `TestExecutor` instance.
Result:
Additional testing capability for `TestExecutor` users.

### DIFF
--- a/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
+++ b/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
@@ -54,7 +54,7 @@ public class TestExecutor implements Executor {
     private final CompletableProcessor closeProcessor = new CompletableProcessor();
     private final AtomicInteger tasksExecuted = new AtomicInteger();
     private final AtomicInteger scheduledTasksExecuted = new AtomicInteger();
-    String instanceName = getClass().getSimpleName() + "-" + INSTANCES.incrementAndGet();
+    final String instanceName = getClass().getSimpleName() + "-" + INSTANCES.incrementAndGet();
 
     /**
      * Create a new instance.
--- a/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
+++ b/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
@@ -34,10 +34,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiPredicate;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
```diff
@@ -314,18 +321,118 @@ private static boolean executeOne(Queue<RunnableWrapper> tasks, AtomicInteger ta
...
+
+        @Nullable
+        @Override
+        public Key<?> forEach(final BiPredicate<Key<?>, Object> consumer) {
+            throw new AssertionError("Invalid access of AsyncContextMap");
+        }
+
+        @Override
+        public AsyncContextMap copy() {
+            throw new AssertionError("Invalid access of AsyncContextMap");
```
idelpivnitskiy, line 435: Consider moving the `throw new ...` into a private method to avoid duplicating the cause message.

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
```diff
@@ -314,18 +321,118 @@ private static boolean executeOne(Queue<RunnableWrapper> tasks, AtomicInteger ta
...
         }
 
         @Override
         public void run() {
-            delegate.run();
+            Thread current = Thread.currentThread();
+            String oldName = current.getName();
+            current.setName(threadName);
+            AsyncContextMap tlPrev = contextThreadLocal.get();
+            contextThreadLocal.set(InvalidAsyncContextMap.INSTANCE);
```
idelpivnitskiy, line 348: Can we just keep the thread name change here, and wrap this `RunnableWrapper` with `ContextPreservingRunnable`?

If not, consider changing `contextThreadLocal` only if the thread is not an instance of `AsyncContextMapHolder`, to be consistent with `ContextPreservingRunnable`.

bondolo: Yes, switching to providing `InvalidAsyncContext` instance rather than `null` allowed wrapping with `AsyncContext.provider().wrapRunnable(delegate, InvalidAsyncContextMap.INSTANCE)` and greatly simplifies the implementation to just the thread naming.

idelpivnitskiy (approved):
Great change!

bondolo:
0.41 (242a6d8d24ca46b46a12c64764cc72300463d20f)

PR #1664, opened 2021-07-09
Migrate tests to junit5 - Step 4

danfaer:
Motivation:

JUnit 5 leverages features from Java 8 or later, such as lambda functions, making tests more powerful and easier to maintain.
JUnit 5 has added some very useful new features for describing, organizing, and executing tests. For instance, tests get better display names and can be organized hierarchically.
JUnit 5 is organized into multiple libraries, so only the features you need are imported into your project. With build systems such as Maven and Gradle, including the right libraries is easy.
JUnit 5 can use more than one extension at a time, which JUnit 4 could not (only one runner could be used at a time). This means you can easily combine the Spring extension with other extensions (such as your own custom extension).
Modifications:

Unit tests have been migrated from JUnit 4 to JUnit 5

Result:

Converted modules now run tests using JUnit 5

#1568

### DIFF
--- a/servicetalk-serialization-api/src/test/java/io/servicetalk/serialization/api/DefaultSerializerDeserializationTest.java
+++ b/servicetalk-serialization-api/src/test/java/io/servicetalk/serialization/api/DefaultSerializerDeserializationTest.java
@@ -21,13 +21,11 @@ import io.servicetalk.concurrent.BlockingIterator;
 import io.servicetalk.concurrent.CloseableIterable;
 import io.servicetalk.concurrent.CloseableIterator;
 import io.servicetalk.concurrent.api.Publisher;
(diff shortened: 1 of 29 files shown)


servicetalk-serialization-api/src/test/java/io/servicetalk/serialization/api/DefaultSerializerDeserializationTest.java
```diff
@@ -302,15 +297,15 @@ public void deserializeIncompleteAggregatedWithType() {
 
         try {
             deserialized.iterator().hasNext();
-            fail();
+            Assertions.fail();
```
tkountis, line 300: Try to use static-imports for things within the context that are easy to understand.

tkountis (approved):
🚀

PR #1666, opened 2021-07-13
Specify PITest version in `Versions.groovy`

bondolo:
Motivation:
For Gradle modules which are not part of ServiceTalk butuse the ServiceTalk Gradle plugin it is necessary to include the versions referenced in the `Versions.groovy` file because they cannot reference the `gradle.properties` of the main ServiceTalk project.
Modifications:
`PITEST_VERSION` and `PITEST_PLUGIN_JUNIT5_VERSION` added to `Versions.groovy`
Result:
Modules using servicetalk-gradle-plugin build

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -14,7 +14,6 @@
 # limitations under the License.
 #
 
-
 org.gradle.parallel=true
 
 group=io.servicetalk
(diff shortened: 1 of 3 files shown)


idelpivnitskiy (approved):
❤️

PR #1667, opened 2021-07-13
`ClientEffectiveStrategyTest` and `ServerEffectiveStrategyTest` cleanups

bondolo:
Motivation:
The parameterization pattern used for both `ClientEffectiveStrategyTest` and `ServerEffectiveStrategyTest` makes it difficult to understand the configuration under test because the configuration is built imperatively rather than as a set of declarative options.
Modifications:
Restructured test parameterization. One specific case is improved; for the cases where no offloading to provided executor is to be performed the test now explicitly ensures that the test fails if the executor is used.
Result:
Easier to understand test with easier to understand test cases.


idelpivnitskiy (approved):
Thanks!

PR #1668, opened 2021-07-14
Convert "ASCII-art" arrows to unicode arrows

bondolo:
Motivation:
The javadoc for quite a few methods includes use of `=&gt;` to
approximate an arrow for diagrams in small figures. Converting these to
unicode characters is cleaner, makes the source the same as the
rendered output and is more accessible.
Modifications:
Use of `=&gt;` is replaced with `⇒`
Result:
Cleaner source.


idelpivnitskiy (approved):
🚀

PR #1669, opened 2021-07-15
Deprecate OpenTracing APIs that will be removed later through a breaking-api release.

tkountis:
Motivation:

We are going through an effort to re-visit the OpenTracing APIs and the implementations
as part of a fix related to B3 sampling.

Modifications:

Deprecating APIs that will be removed or replaced where needed in a follow up breaking-api release.

Result:

Cleaner API, consistent & leaner state without duplication.

Link for breaking changes: https://github.com/apple/servicetalk/pull/1640


tkountis:
Failing due to https://github.com/apple/servicetalk/issues/1663

Scottmitch (approved):

PR #1670, opened 2021-07-15
[Port to main] Deprecate OpenTracing APIs that will be removed later through a breaking-api release.

tkountis:
Motivation:

We are going through an effort to re-visit the OpenTracing APIs and the implementations
as part of a fix related to B3 sampling.

Modifications:

Deprecating APIs that will be removed or replaced where needed in a follow up breaking-api release.

Result:

Cleaner API, consistent & leaner state without duplication.

(cherry picked from commit b0612524904907e5df25a03fb42ed86dd752da57)


Scottmitch (approved):
lgtm after clean build

PR #1671, opened 2021-07-15
Publisher#flatMapConcatIterable propagate error even if no onSubscribe

Scottmitch:
Motivation:
The ReactiveStreams specification says that onSubscribe(Subscription)
must be called prior to any other signals being delivered [1]. However
in practice user code may throw in onSubscribe which may stop the
propagation. In this case Publisher#flatMapConcatIterable has an
assertion which may prevent the error from propagating downstream and
result in a control flow deadlock.

[1] https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#1.9

Modifications:
- Move assertion in Publisher#flatMapConcatIterable to allow for null
  subscription, add a runtime check before it is used.

Result:
More robust error propagation for Publisher#flatMapConcatIterable in the
precense of Reactive Streams specification violations.


idelpivnitskiy (approved):
🎖️

Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/1672

Scottmitch:
0.41 (5b66b7ec4)

PR #1673, opened 2021-07-16
New Serializer APIs, consolidation of ContentCodec, and gRPC MethodDescriptor

Scottmitch:
Motivation:
Supporting a custom transport for gRPC currently requires
reflection and special knowledge of types for serialization. The
generated code has information related to types and serialization that
can be exposed to the runtime which also allows supporting custom
transports.

This task requires exposing serializer APIs into the public
API. Our existing serializer APIs were heavily influced by streaming use
cases and JSON serialization which can serialize any type and generally
doesn't have type restrictions. This lead to a general
SerializationProvider API which is impossible to constrain types due to
the generics being at the method level, which doesn't account for
implementing serialization for type constrained approached (protobuf
MessageLite). Also the serialization APIs couple scalar and streaming
use cases which in practice are more commonly decoupled into scalar
(individual objects) and streaming (which applies some framing around
individual objects). These concepts are coupled which leads to confusing
specializations (deserializeAggregatedSingle, deserializeAggregated)
which are easy to misuse and lead to invalid serialization in practice
(streaming {form url, string} encoding are not properly framed and may
not be deserialized correctly). The existing Serializer APIs also do not
promote composability in that if you can serialize one object you should
be able to re-use that serialization for a stream of objects with some
frameing applied on top (required for gRPC).

Modifications:
- Introduce a new servicetalk-serializer-api package with new
  Serializer APIs.
- Introduce a new servicetalk-serializer-utils package with common
  utilities for serialization such as streaming framing (fixed and
  varint length prefixed)
- Introduce BufferEncoder APIs in servicetalk-encoding-api for
  compression and decompression. This also comes with
  NettyBufferEncoders and NettyCompression to create Netty backed
  implementations.
- Introduce HttpStreamingSerializer and new HttpSerializer APIs in
  servicetalk-http-api, and modify all request/response APIs to use the
  new serializer APIs. Introduce new methods to deserialize and process
  trailers.
- Introduce HttpSerializers in servicetalk-http-api for implementations
  of the new APIs.
- Introduce JacksonSerializerCache in servicetalk-http-api as the new
  entry point to create/cache serializers for JSON.
- Introduce ProtobufSerializerCache in servicetalk-data-protobuf as the
  new entry point to create/cache serializers for protobuf.
- Introduce MethodDescriptor API into servicetalk-grpc-api and update
  code generation to use this new API.
- Update all examples avoid usage of deprecated APIs and leverage new
  APIs.
- Deprecate the following APIs and related methods in favor of the new
  types descrbied above:
  - SerializationProvider, JacksonSerializationProvider,
    ProtobufSerializationProvider, GrpcSerializationProvider,
    HttpSerializationProvider
  - ContentCodec, CodecDecodingException, CodecEncodingException,
    ContentCodings
  - GrpcMetadata#path(), code generated types that extend
    DefaultGrpcClientMetadata

Result:
gRPC API supports MethodDescriptor, which uses new Serializer APIs, and
the entire code base has been updated to use the new APIs consistently.


Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/1579

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1663

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1663

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/FixedLengthStreamingSerializer.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override
+    public Publisher<T> deserialize(final Publisher<Buffer> serializedData, final BufferAllocator allocator) {
+        return serializedData.liftSync(new FramedDeserializerOperator<>(serializer, LengthDeframer::new, allocator))
+                .flatMapConcatIterable(identity());
+    }
+
+    @Override
+    public Publisher<Buffer> serialize(final Publisher<T> toSerialize, final BufferAllocator allocator) {
+        return toSerialize.map(t -> {
+            Buffer buffer = allocator.newBuffer(4 + bytesEstimator.applyAsInt(t));
```
bondolo, line 61: Integer.BYTES

servicetalk-serializer-api/src/main/java/io/servicetalk/serializer/api/StreamingDeserializer.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Deserialize a {@link Publisher} of {@link Buffer} to {@link Publisher} of {@link Object}.
+ * @param <T> The type of objects that can be deserialized.
+ */
+@FunctionalInterface
+public interface StreamingDeserializer<T> {
+    Publisher<T> deserialize(Publisher<Buffer> serializedData, BufferAllocator allocator);
+
+    default BlockingIterable<T> deserialize(Iterable<Buffer> serializedData, BufferAllocator allocator) {
```
bondolo, line 33: Any advantage to providing this over what the application could do? I ask because it begs the questions of the other two flavors of Publisher<T>/Iterable<T>

bondolo: Same comment for `StreamingSerializer`

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/StringCharsetSerializer.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Map;
+
+/**
+ * Serialize/deserialize {@link String}s encoded with a {@link Charset}.
+ * @see StringAsciiSerializer
+ * @see StringUtf8Serializer
+ */
+public final class StringCharsetSerializer implements SerializerDeserializer<String> {
+    private static final Map<Charset, Integer> MAX_BYTES_PER_CHAR_MAP = new HashMap<>();
+    static {
```
bondolo, line 33: There are often hundreds of mostly garbage charsets. Perhaps computeIfAbsent?

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/StringCharsetSerializer.java
```diff
@@ -0,0 +1,69 @@
...
+    private final Charset charset;
+    private final int maxBytesPerChar;
+
+    /**
+     * Create a new instance.
+     * @param charset The charset used for encoding.
+     */
+    public StringCharsetSerializer(final Charset charset) {
+        this.charset = charset;
+        maxBytesPerChar = MAX_BYTES_PER_CHAR_MAP.getOrDefault(charset, 1);
```
bondolo, line 48: if you use the computeIfAbsent suggestion then you won't need a default.

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/StringUtf8Serializer.java
```diff
@@ -0,0 +1,52 @@
...
+import io.servicetalk.buffer.api.Buffer;
+import io.servicetalk.buffer.api.BufferAllocator;
+import io.servicetalk.serializer.api.SerializerDeserializer;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * Serialize/deserialize {@link String}s with {@link java.nio.charset.StandardCharsets#US_ASCII} encoding.
+ */
+public final class StringUtf8Serializer implements SerializerDeserializer<String> {
```
bondolo, line 27: I don't see any advantage over extending `StringCharsetSerializer`

bondolo (commented):
70/304 files so far…

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaderValues.java
```diff
@@ -153,10 +153,15 @@
...
+    /**
+     * {@code "text/plain"}
+     */
+    public static final CharSequence TEXT_PLAIN_US_ASCII = newAsciiString("text/plain; charset=US-ASCII");
     /**
      * {@code "text/plain"}
      */
     public static final CharSequence TEXT_PLAIN_UTF_8 = newAsciiString("text/plain; charset=UTF-8");
+
     /**
```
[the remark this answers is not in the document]
Scottmitch: This newline wasn't intentional, I'll revert this change.

this list was a subset of Netty's [HttpHeaderValues](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderValues.java) and we can add new header values as required (or the user can create their own).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpDeserializer2.java
```diff
@@ -0,0 +1,35 @@
...
+public interface HttpDeserializer2<T> {
+    /**
+     * Deserialize a single {@link Object} into a {@link T}.
+     * @param headers The {@link HttpHeaders} associated with the {@code payload}.
+     * @param allocator Used to allocate intermediate {@link Buffer}s if required.
+     * @param payload The {@link Object} to deserialize. The contents are assumed to be in memory, otherwise this method
+     * may block.
+     * @return The result of the deserialization.
+     */
+    T deserialize(HttpHeaders headers, BufferAllocator allocator, Buffer payload);
```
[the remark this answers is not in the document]
Scottmitch: I started with the approach of using `HttpDeserializer` and `HttpSerializer`, as these are the desired names longer term, but the usage of this API on `HttpRequest` and `HttpResponse` is such that it should only be the aggregated methods. The new implementations of these APIs don't make sense for the streaming paths (e.g. no framing is specified) so I could implement these to fail with UnsupportedOperationException, but then users may accidentally pass implementations to streaming APIs and get exceptions at runtime. Having a distinct API avoids passing impls to the wrong method and partially implemented interface, but negative is that the method name isn't desirable long term. However since it is intended to be only "temporary" (likely 3+ releases bcz these changes are central) maybe this is OK... wdyt?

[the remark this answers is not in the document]
bondolo: Since the interface is intended to be used only temporarily the naming is not a problem.

servicetalk-serializer-api/src/main/java/io/servicetalk/serializer/api/StreamingDeserializer.java
Scottmitch: Gives the `StreamingDeserializer` the option to natively do serialization with the `Iterable` if it wants to and avoid the intermediate translation.

`PayloadWriter` on the `StreamingSerializer` is harder to wrap and convert at higher levels without support on the base API (see `DefaultHttpStreamingSerializerDeserializer`).

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/StringCharsetSerializer.java
Scottmitch: can you clarify the benefit of `computeIfAbsent`? are you implying there will be `Charset` whose hash/equals collide and the first one should win?

bondolo: Ah, commented on the wrong line. 

The problem is that `Charset.availableCharsets().values()` may return hundreds of charsets, mostly goofy Windows ones, obscure CJK charsets and oddballs like EBCDIC. The OpenJDK default is about 125 charsets, SAP and IBM in particular include many more in their JDKs (400+ if I remember correctly). Not sure about other vendors. With `computeIf Absent(charset, aCharset -> (int) charset.newEncoder().maxBytesPerChar());` you'll only compute the charsets you need and there is no need for the `getOrDefault(charset, 1)` should you encounter a charset not in `Charset.availableCharsets()`, which is possible.

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/StringUtf8Serializer.java
Scottmitch: there are specialized methods on `Buffer` for `writeUtf8` and `writeUsAscii` but good point I can internalize that from the API perspective.

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/StringCharsetSerializer.java
Scottmitch: Lets defer this for later if more exotic Charsets are used in practice and if we dynamically modify the map in the constructor we have to worry about concurrent access which I'm assuming this "missing charset" case will be very rare (most common usage is in `StandardCharset` which I assume would also be in `availableCharsets`).

Scottmitch:
another https://github.com/apple/servicetalk/issues/1579

Scottmitch:
another https://github.com/apple/servicetalk/issues/1663

Scottmitch:
another https://github.com/apple/servicetalk/issues/1663

Scottmitch:
another https://github.com/apple/servicetalk/issues/1579 and https://github.com/apple/servicetalk/issues/1663

Scottmitch:
https://github.com/apple/servicetalk/issues/1507 and https://github.com/apple/servicetalk/issues/1663

Scottmitch:

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/StringCharsetSerializer.java
bondolo: Yes, `StandardCharset` will be in `availableCharsets`. I am still worried about the cost of computing the result for all available charsets when most (probably 95%+) will never be used and resulting impact on startup time. I noticed that another class, `FormUrlEncodedSerializer` also builds a map for all charsets. A possible alternative is to continue to use HashMap, preload it with the StandardCharsets, and when another charset is encountered, copy the map, add the new charset and replace the entire map. Lighter weight than a full `ConcurrentMap` and only a volatile read more expensive on average.

Scottmitch: I will add a utility targeting standardard charset (I don't know of an existing collection which makes these available from the JDK). Since no COW map exists in the JDK I'll just stick to read only map for now and we can enhance later if necessary.

servicetalk-examples/http/serialization/src/main/java/io/servicetalk/examples/http/serialization/SerializerUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+import io.servicetalk.http.api.HttpSerializerDeserializer;
+import io.servicetalk.http.api.HttpStreamingSerializerDeserializer;
+
+import static io.servicetalk.data.jackson.JacksonSerializerCache.INSTANCE;
+import static io.servicetalk.http.api.HttpSerializers.jsonSerializer;
+import static io.servicetalk.http.api.HttpSerializers.jsonStreamingSerializer;
+
+/**
+ * Utilities to cache serializer instances.
```
bondolo, line 26: …cache POJO to JSON serializer…

bondolo (approved):
approve for examples portion

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ContentEncodingHttpServiceFilter.java
```diff
@@ -0,0 +1,224 @@
...
+/**
+ * A {@link StreamingHttpService} that adds encoding / decoding functionality for responses and requests respectively,
+ * as these are specified by the spec
+ * <a href="https://tools.ietf.org/html/rfc7231#section-3.1.2.2">Content-Encoding</a>.
+ *
+ * <p>
+ * Append this filter before others that are expected to to see compressed content for this request/response, and after
+ * other filters that expect to see/manipulate the original payload.
+ */
+public final class ContentEncodingHttpServiceFilter
```
tkountis, line 53: nit: Coding as noun has both the meaning of encoding or decoding. I think the name before made more sense. The comment is worth considering across the whole encoding/decoding APIs (didn't want to duplicate it).

servicetalk-serializer-api/src/main/java/io/servicetalk/serializer/api/Serializer.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.serializer.api;
+
+import io.servicetalk.buffer.api.Buffer;
+import io.servicetalk.buffer.api.BufferAllocator;
+
+/**
+ * Serialize from {@link Object} to {@link Buffer} and deserialize objects from {@link Buffer} to {@link Object}.
```
tkountis, line 22: javadoc out of sync

servicetalk-serializer-api/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+
+  implementation project(":servicetalk-annotations")
+  implementation project(":servicetalk-concurrent-api-internal")
+  implementation project(":servicetalk-oio-api-internal")
+  implementation "com.google.code.findbugs:jsr305:$jsr305Version"
+
+  testImplementation testFixtures(project(":servicetalk-concurrent-api"))
+  testImplementation testFixtures(project(":servicetalk-concurrent-internal"))
+  testImplementation project(":servicetalk-concurrent-test-internal")
+  testImplementation "junit:junit:$junitVersion"
```
tkountis, line 32: are they needed? no tests

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/AbstractStringSerializer.java
```diff
@@ -0,0 +1,73 @@
...
+import io.servicetalk.serializer.api.SerializerDeserializer;
+
+import java.nio.charset.Charset;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+import static io.servicetalk.utils.internal.CharsetUtils.standardCharsets;
+
+abstract class AbstractStringSerializer implements SerializerDeserializer<String> {
```
tkountis, line 29: Serializer/Deserializer in name? same for subclass

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/VarIntLengthStreamingSerializer.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.function.BiFunction;
+import java.util.function.ToIntFunction;
+import javax.annotation.Nullable;
+
+import static java.util.Objects.requireNonNull;
+import static java.util.function.Function.identity;
+
+/**
+ * A {@link StreamingSerializerDeserializer} that uses a {@link SerializerDeserializer} and frames each object by
+ * preceding it with the length in bytes. The length component is variable length an encoded as
```
tkountis, line 34: "variable length an encoded as" -> "variable length and encoded as"

servicetalk-data-jackson/src/main/java/io/servicetalk/data/jackson/JacksonSerializer.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.io.IOException;
+
+import static io.servicetalk.buffer.api.Buffer.asOutputStream;
+
+/**
+ * Serializes and deserializes to/from JSON via jackson.
+ * @param <T> The type of objects to serialize.
+ */
+final class JacksonSerializer<T> implements SerializerDeserializer<T> {
```
tkountis, line 37: Consider Serialiazer/Deserializer in name

servicetalk-data-jackson/src/main/java/io/servicetalk/data/jackson/JacksonStreamingSerializer.java
```diff
@@ -0,0 +1,412 @@
...
+import javax.annotation.Nullable;
+
+import static com.fasterxml.jackson.databind.node.JsonNodeFactory.instance;
+import static io.servicetalk.concurrent.internal.EmptySubscriptions.EMPTY_SUBSCRIPTION;
+import static io.servicetalk.data.jackson.JacksonSerializer.doSerialize;
+import static java.util.Collections.singletonList;
+import static java.util.function.Function.identity;
+
+/**
+ * Serialize a stream of JSON objects.
```
tkountis, line 59: Serialize and deserialize

servicetalk-data-jackson/src/main/java/io/servicetalk/data/jackson/JacksonStreamingSerializer.java
```diff
@@ -0,0 +1,412 @@
...
+
+        private DeserializeOperator(ObjectReader reader) {
+            this.reader = reader;
+        }
+
+        @Override
+        public Subscriber<? super Buffer> apply(final Subscriber<? super Iterable<T>> subscriber) {
+            final JsonParser parser;
+            try {
+                // TODO(scott): ByteBufferFeeder is currently not supported by jackson, and the current API throws
```
tkountis, line 106: Do we want this in now? Shall we open a radar?

tkountis (commented):
1st round.
General comments: 
- that I didn't want to repeat, it's probably worth sticking to ser/deser naming convention to all implementations.
- new APIs are certainly better and I love the effort with varint streaming support.

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/AbstractStringSerializer.java
chemicL: A shorthand in the form of SerDe is frequently used in the community.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpMessageBodyUtils.java
```diff
@@ -0,0 +1,155 @@
...
+
+            @Nullable
+            @Override
+            public HttpHeaders trailers() {
+                return msgBody.trailers();
+            }
+        };
+    }
+
+    private static final class BlockingStreamingHttpMessageBodyAdapter<I> implements BlockingIterator<Buffer> {
```
chemicL, line 73: Should access to this class' state be made thread-safe? `BlockingIterator` comes from the `io.servicetalk.concurrent` package. I'd assume it can be used in a concurrent context. The parent class has no mention of it requiring single-thread access.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ContentEncodingHttpServiceFilter.java
```diff
@@ -0,0 +1,224 @@
...
+        // Content-Length or Transfer-Encoding header fields received in
+        // such a message.
+        // ...
+        final int code = response.status().code();
+        return INFORMATIONAL_1XX.contains(code) || code == NO_CONTENT.code() || code == NOT_MODIFIED.code() ||
+                (method == HEAD || (method == CONNECT && SUCCESSFUL_2XX.contains(code)));
+    }
+
+    @Nullable
+    static <T> T matchAndRemoveEncoding(final List<T> supportedEncoders,
```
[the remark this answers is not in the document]
Scottmitch: I tried sharing code between these two methods but they are doing different jobs and I couldn't find a good way. I considered at least moving them into the same `HeaderUtils` file, to keep similar code close but the `matchAndRemoveEncoding` can be kept package private if it stays here.

`matchAndRemoveEncoding` -> needs to remove the encoding header that is matched at preserve ordering (bcz the body has been decoded and the header no longer applies). first header must supported values or else we can't decode and its an error.
`negotiateAcceptedEncodingRaw` -> finds a matching encoding, so it can be applied (no removal of headers). any match is acceptable.

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/AbstractStringSerializer.java
Scottmitch: The strategy I used was to be explicit with the interface names (`SerializerDeserializer`, `BufferEncoderDecoder`) but just use the "serializer" variant for concrete method/type names that fill both roles for brevity (`StringSerializer`, `NettyBufferEncoders#bufferEncoder(..)`, `HttpSerializers#textSerializerUtf8()`). In practice I'm not sure lengthening the names of concrete types/methods, or using abbreviations (`SerDe`), will improve understandability/readability.

servicetalk-data-jackson/src/main/java/io/servicetalk/data/jackson/JacksonStreamingSerializer.java
Scottmitch: https://github.com/apple/servicetalk/issues/1711 ... this is long standing from 3+ years ago just cpy/pasta to use the new APIs.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpMessageBodyUtils.java
Scottmitch: No this isn't necessary. `BlockingIterator` is in `concurrent` package because it is how we convert to/from `java.util.Iterator` related types. This type just adds timeouts at the call site and makes iterable `Closable` (which will cancel the underlying publisher if it isn't fully consumed).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ContentEncodingHttpServiceFilter.java
```diff
@@ -0,0 +1,224 @@
...
+     * The encodings can differ for requests and responses, allowing a server that supports different compressions for
+     * requests and different ones for responses.
+     * <p>
+     * The order of the codecs provided, affect selection priority alongside the order of the incoming
+     * <a href="https://tools.ietf.org/html/rfc7231#section-5.3.4">accept-encoding</a> header from the client.
+     *
+     * @param decompressors used to decompress client requests if compressed.
+     * @param compressors used to compress server responses if client accepts them.
+     */
+    public ContentEncodingHttpServiceFilter(final BufferDecoderGroup decompressors,
```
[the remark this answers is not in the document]
Scottmitch: done

Scottmitch:

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/BufferDecoderGroupBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Create a new instance.
+     */
+    public BufferDecoderGroupBuilder() {
+        this(2);
+    }
+
+    /**
+     * Create a new instance.
+     * @param encodersSizeEstimate estimate as to how many {@link BufferDecoder} will be included in the
```
idelpivnitskiy, line 42: `encodersSizeEstimate` -> `decodersSizeEstimate`?

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/BufferDecoderGroupBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+import javax.annotation.Nullable;
+
+import static io.servicetalk.buffer.api.CharSequences.newAsciiString;
+import static java.util.Collections.emptyList;
+
+/**
+ * Builder for {@link BufferDecoderGroup}s.
+ */
+public final class BufferDecoderGroupBuilder {
+    private static final char CONTENT_ENCODING_SEPARATOR = ',';
```
idelpivnitskiy, line 29: Consider adding a space before comma. Helps with readability when users debug and will be consistent with how it's formatted in RFC

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/BufferEncodingException.java
```diff
@@ -0,0 +1,64 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.encoding.api;
+
+import io.servicetalk.serializer.api.SerializationException;
+
+/**
+ * A specialization of {@link SerializationException} used to indicate a encoding exception.
```
idelpivnitskiy, line 21: `a` -> `an`

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/CodecDecodingException.java
```diff
@@ -17,7 +17,9 @@
 
 /**
  * Exception thrown when something goes wrong during decoding.
+ * @deprecated Use {@link BufferEncodingException}.
  */
+@Deprecated
 public final class CodecDecodingException extends RuntimeException {
```
idelpivnitskiy, line 23: Same question here, should we make this and `CodecEncodingException` subtypes of `SerializationException`?

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/IdentityBufferEncoderDecoder.java
```diff
@@ -0,0 +1,135 @@
...
+import io.servicetalk.serializer.api.Deserializer;
+import io.servicetalk.serializer.api.Serializer;
+import io.servicetalk.serializer.api.SerializerDeserializer;
+import io.servicetalk.serializer.api.StreamingDeserializer;
+import io.servicetalk.serializer.api.StreamingSerializer;
+import io.servicetalk.serializer.api.StreamingSerializerDeserializer;
+
+import static io.servicetalk.buffer.api.CharSequences.newAsciiString;
+
+final class IdentityBufferEncoderDecoder implements BufferEncoderDecoder {
```
idelpivnitskiy, line 33: It's important to implement `equals` and `hashCode` that follow RFC. See how those are implemented in `IdentityContentCodec`.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/DefaultBufferEncoderDecoder.java
```diff
@@ -0,0 +1,66 @@
...
+import io.servicetalk.serializer.api.Deserializer;
+import io.servicetalk.serializer.api.Serializer;
+import io.servicetalk.serializer.api.SerializerDeserializer;
+import io.servicetalk.serializer.api.StreamingDeserializer;
+import io.servicetalk.serializer.api.StreamingSerializer;
+import io.servicetalk.serializer.api.StreamingSerializerDeserializer;
+
+import static java.util.Objects.requireNonNull;
+
+final class DefaultBufferEncoderDecoder implements BufferEncoderDecoder {
```
idelpivnitskiy, line 29: Same, see `AbstractContentCodec`

idelpivnitskiy (commented):
I will submit my review in chunks:

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/FixedLengthStreamingSerializer.java
```diff
@@ -0,0 +1,94 @@
...
+public final class FixedLengthStreamingSerializer<T> implements StreamingSerializerDeserializer<T> {
+    private final SerializerDeserializer<T> serializer;
+    private final ToIntFunction<T> bytesEstimator;
+
+    /**
+     * Create a new instance.
+     * @param serializer The {@link SerializerDeserializer} used to serialize/deserialize individual objects.
+     * @param bytesEstimator Provides the length in bytes for each {@link T} being serialized.
+     */
+    public FixedLengthStreamingSerializer(final SerializerDeserializer<T> serializer,
```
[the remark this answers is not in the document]
idelpivnitskiy: The alternative approach I was thinking about is to take `Serializer` and `Deserializer` separately for more complex utilities. And provide one utility that can merge them both into `SerializerDeserializer`:
```
FixedLengthStreamingSerializer(Serializer<T> s)
FixedLengthStreamingSerializer(Deserializer<T> d)
DefaultSerializerDeserializer(Serializer<T> s, Deserializer<T> d)
```
We can also use static factories instead of ctors.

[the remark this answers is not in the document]
Scottmitch: The combined utility is sufficient for current use cases and we can add followup utilities on demand. In practice the serializer/deserializer APIs are useful to have defined independently so accepting API contracts can be more specific, but implementations will often need to be able todo both serialization/deserialization.

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/AbstractStringSerializer.java
```diff
@@ -0,0 +1,73 @@
...
+    private final Charset charset;
+    private final int maxBytesPerChar;
+
+    /**
+     * Create a new instance.
+     * @param charset The charset used for encoding.
+     */
+    AbstractStringSerializer(final Charset charset) {
+        this.charset = charset;
+        maxBytesPerChar = MAX_BYTES_PER_CHAR_MAP.getOrDefault(charset, 1);
```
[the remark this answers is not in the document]
Scottmitch: Discussed with @bondolo above in https://github.com/apple/servicetalk/pull/1673#discussion_r675131444 ... lets keep the map read only for now to avoid concurrency issues. we can enhance cache in the future as necessary.

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/FramedDeserializerOperator.java
```diff
@@ -0,0 +1,156 @@
...
+
+import static java.util.Collections.singletonList;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Utility which helps implementations of {@link StreamingDeserializer} leverage a {@link Deserializer} and apply a
+ * framing to define the boundaries of each object.
+ * @param <T> The type to serialize/deserialize.
+ */
+public final class FramedDeserializerOperator<T> implements PublisherOperator<Buffer, Iterable<T>> {
```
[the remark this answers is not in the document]
Scottmitch: 👍 yes, done.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpSerializers.java
```diff
@@ -0,0 +1,318 @@
...
+
+    /**
+     * Creates a {@link HttpStreamingSerializerDeserializer} that serializes {@link String}s with
+     * {@link StandardCharsets#UTF_8} encoding using fixed {@code int} length delimited framing.
+     *
+     * @return a {@link HttpStreamingSerializerDeserializer} that serializes {@link String}s with
+     * {@link StandardCharsets#UTF_8} encoding using fixed {@code int} length delimited framing.
+     * @see FixedLengthStreamingSerializer
+     */
+    public static HttpStreamingSerializerDeserializer<String> textSerializerUtf8FixLen() {
```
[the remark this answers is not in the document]
Scottmitch: The serialization isn't a problem, I will add a utility method to create a `HttpStreamingSerializer` given a `Serializer`. This maybe useful when folks have content in non-`Buffer` types (e.g. `String`, `byte[]`) and know the content-type a-prior (e.g. json/xml). However they have to use that a-prior content-type knowledge to select the corresponding `HttpStreamingDeserializer` on the other side. For example you can't just serialize a raw `String` and deserialize them in the same fashion on the other side without some framing (e.g. how many bytes is in each `String`).

> I saw many cases when users use string/text serialization to send JSON or XML

The HttpSerializers will enforce headers checks to ensure content type is expected. So `HttpSerializationProviders#textSerializer()` wouldn't work to "stream raw data from `String`" OOTB without stripping/manipulating headers.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpSerializers.java
```diff
@@ -0,0 +1,318 @@
...
+import static io.servicetalk.http.api.HttpHeaderValues.TEXT_PLAIN_UTF_8;
+import static io.servicetalk.serializer.utils.StringSerializer.stringSerializer;
+import static java.nio.charset.StandardCharsets.US_ASCII;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * Factory for creation of {@link HttpSerializerDeserializer} and {@link HttpStreamingSerializerDeserializer}.
+ */
+public final class HttpSerializers {
+    private static final CharSequence APPLICATION_TEXT_FIXED = newAsciiString("application/text-fix-int");
```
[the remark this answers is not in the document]
Scottmitch: I think `application` I more appropriate bcz it isn't strictly human readable (binary data for integer encoding).

https://www.rfc-editor.org/rfc/rfc2046#section-3
> text -- textual information. ... In
          particular, formats that employ embeddded binary
          formatting information are not considered directly
          readable.

> application -- some other kind of data, typically
          either uninterpreted binary data or information to be
          processed by an application. ... and word
          processing formats that are not directly readable.

https://www.rfc-editor.org/rfc/rfc2046#section-4.1
> The "text" media type is intended for sending material which is
   principally textual in form.

https://datatracker.ietf.org/doc/html/rfc3023#section-3
> If an XML document -- that is, the unprocessed, source XML document
   -- is readable by casual users, text/xml is preferable to
   application/xml.

servicetalk-encoding-api/src/main/java/io/servicetalk/encoding/api/BufferDecoderGroupBuilder.java
Scottmitch: I prefer to leave it as is. less space on the wire, less white space stripping, negligible readability impact, and consistent with other implementations (e.g. [grpc-java](https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/DecompressorRegistry.java#L38))

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcDeserializer.java
```diff
@@ -0,0 +1,71 @@
...
+import io.servicetalk.serializer.api.SerializationException;
+
+import javax.annotation.Nullable;
+
+import static io.servicetalk.grpc.api.GrpcStreamingDeserializer.isCompressed;
+import static io.servicetalk.grpc.api.GrpcStreamingSerializer.METADATA_SIZE;
+import static java.util.Objects.requireNonNull;
+
+final class GrpcDeserializer<T> implements Deserializer<T> {
+    private final Deserializer<T> serializer;
```
chemicL, line 31: deserializer

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcDeserializer.java
```diff
@@ -0,0 +1,71 @@
...
+    @Nullable
+    private final BufferDecoder compressor;
```
chemicL, line 33: Should the variable name reflect the decoding character? Perhaps decoder or decompressor would be more appropriate?

servicetalk-serializer-utils/src/main/java/io/servicetalk/serializer/utils/FixedLengthStreamingSerializer.java
```diff
@@ -0,0 +1,94 @@
...
+import static java.lang.Integer.BYTES;
+import static java.util.Objects.requireNonNull;
+import static java.util.function.Function.identity;
+
+/**
+ * A {@link StreamingSerializerDeserializer} that uses a {@link SerializerDeserializer} and frames each object by
+ * preceding it with the length in bytes. The length component is fixed and always consumes 4 bytes.
+ * @param <T> The type of object to serialize.
+ */
+public final class FixedLengthStreamingSerializer<T> implements StreamingSerializerDeserializer<T> {
```
tkountis, line 38: Did you consider the option to support header/footer for the streaming approach, that would help the reading side verify that the input is indeed fixed-length and validate checksum at the end. 
Optional of course, it can still be added later and maybe even with an additional operator.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpSerializers.java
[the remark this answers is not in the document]
idelpivnitskiy: Ok, because we have a length prefix, it's not 100% text data. I would agree, `application/...` is more appropriate. Users can configure their own header value, if necessary. Thanks for the links!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ContentEncodingHttpServiceFilter.java
Scottmitch: duplicate of https://github.com/apple/servicetalk/pull/1673#discussion_r681224512 ... lets discuss there if necessary.

servicetalk-data-jackson/src/main/java/io/servicetalk/data/jackson/JacksonSerializer.java
Scottmitch: duplicate of https://github.com/apple/servicetalk/pull/1673#discussion_r681224512 ... lets discuss there if necessary.

servicetalk-grpc-protoc/build.gradle
```diff
@@ -27,6 +27,10 @@ apply plugin: "com.google.protobuf"
 
 dependencies {
   implementation project(":servicetalk-annotations")
+
+  // runtimeOnly is enough for this module in isolation, but we want transitive modules to include at compile time too.
+  api project(":servicetalk-data-protobuf")
```
Scottmitch, line 32: not sure if there is a better way to represent this type of dependency. it is somewhat unique bcz we are doing code generation. basically this module doesn't expose it in its API, it doesn't need it to compile this module, however we don't want modules that depend upon `servicetalk-grpc-protoc` to explicitly include this dependency. note that `implementation` doesn't do the trick:

```
/Users/scott/git/servicetalk/servicetalk-examples/grpc/compression/build/generated/sources/proto/main/java/io/grpc/examples/compression/Greeter.java:51: error: package ProtobufSerializerCache does not exist
    private static final MethodDescriptor<HelloRequest, HelloReply> BlockingSayHelloRpc_MD = MethodDescriptors.newMethodDescriptor("/compression.Greeter/SayHello", "sayHello", false, true, HelloRequest.class, "+proto", ProtobufSerializerCache.INSTANCE.serializerDeserializer(HelloRequest.parser()), HelloRequest::getSerializedSize, false, true, HelloReply.class, "+proto", ProtobufSerializerCache.INSTANCE.serializerDeserializer(HelloReply.parser()), HelloReply::getSerializedSize);
```

Scottmitch:

Scottmitch:
another https://github.com/apple/servicetalk/issues/1579

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpSerializers.java
```diff
@@ -0,0 +1,375 @@
...
+
+    /**
+     * Create a {@link HttpStreamingSerializer} that serializes {@code byte[]}. This method is useful if the payload
+     * body is provided in {@code byte[]} and the {@link HttpHeaderNames#CONTENT_TYPE} is known a-prior
+     * (e.g. streaming raw json data from a stream of {@code byte[]}s). Deserialization should be done using
+     * the a-prior knowledge to use a compatible {@link HttpStreamingDeserializer}.
+     * @param headersSerializeConsumer Sets the headers to indicate the appropriate encoding and content type.
+     * @return a {@link HttpStreamingSerializer} that uses a {@link Serializer} for serialization.
+     */
+    public static HttpStreamingSerializer<byte[]> bytesStreamingSerializer(
```
idelpivnitskiy, line 293: Can we also support deserializer?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpSerializers.java
```diff
@@ -0,0 +1,375 @@
...
+     * if the payload body is provided in non-{@link Buffer} type and the {@link HttpHeaderNames#CONTENT_TYPE} is known
+     * a-prior (e.g. streaming raw json data from a stream of {@link String}s). Deserialization should be done using
+     * the a-prior knowledge to use a compatible {@link HttpStreamingDeserializer}.
+     * @param serializer Used to serialize each {@link T} chunk.
+     * @param bytesEstimator Provides an estimate of how many bytes to allocate for each {@link Buffer} to serialize to.
+     * @param headersSerializeConsumer Sets the headers to indicate the appropriate encoding and content type.
+     * @param <T> Type of object to serialize.
+     * @return a {@link HttpStreamingSerializer} that uses a {@link Serializer} for serialization.
+     */
+    public static <T> HttpStreamingSerializer<T> streamingSerializer(
```
idelpivnitskiy, line 371: There is a method above that returns `HttpStreamingSerializerDeserializer`, this method returns `HttpStreamingSerializer`. Do we need a method that returns `HttpStreamingDeserializer`?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpPayloadHolder.java
```diff
@@ -133,6 +134,61 @@ void transformMessageBody(UnaryOperator<Publisher<?>> transformer) {
...
 
+    <T, S> void transform(final TrailersTransformer<T, S> trailersTransformer,
+                          final HttpStreamingDeserializer<S> serializer) {
+        if (messageBody == null) {
+            messageBody = defer(() ->
+                    from(trailersTransformer.payloadComplete(trailersTransformer.newState(),
+                            headersFactory.newEmptyTrailers())).subscribeShareContext());
+        } else {
+            final Publisher<?> oldMessageBody = messageBody;
+            messageBody = defer(() -> {
```
idelpivnitskiy, line 145: 1. The `payloadInfo.setEmpty(false)` is missed compare to the `transform(TrailersTransformer)` overload.
2. Is there any way to share code between both `transform` overloads?

idelpivnitskiy (commented):
So far the approach looks good! Lots of improvements. 
I still need to look at grpc. Everything else is reviewed. But don't be blocked by that.

servicetalk-data-protobuf/src/main/java/io/servicetalk/data/protobuf/ProtobufSerializerCache.java
```diff
@@ -0,0 +1,68 @@
...
+     * @return a {@link SerializerDeserializer}.
+     */
+    @SuppressWarnings("unchecked")
+    public <T extends MessageLite> SerializerDeserializer<T> serializerDeserializer(Parser<T> parser) {
+        return serializerMap.computeIfAbsent(parser, parser2 -> new ProtobufSerializer<>((Parser<T>) parser2));
+    }
+
+    /**
+     * Get a {@link StreamingSerializerDeserializer} which supports &lt;fixed length, value&gt; encoding as described in
+     * <a href="https://developers.google.com/protocol-buffers/docs/techniques">Protobuf Streaming</a>.
```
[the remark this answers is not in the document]
Scottmitch: Good catch! I also stumbled on this when I was adding tests in response to your Jackson request ... fixed -> varint ftw

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpMessageBody.java
```diff
@@ -0,0 +1,45 @@
...
+
+import io.servicetalk.concurrent.BlockingIterator;
+
+import javax.annotation.Nullable;
+
+/**
+ * Get the <a href="https://tools.ietf.org/html/rfc7230#section-3.3">message-body</a>.
+ * @param <T> The type of the payload body.
+ */
+public interface BlockingStreamingHttpMessageBody<T> {
```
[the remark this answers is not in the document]
Scottmitch: The functional APIs such as `TrailersTransformer` doesn't lend themselves to the sequential programming style of traditional blocking APIs. I meant to deprecate the `transform` methods on the blocking request/response APIs but forgot (will push this change).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
```diff
@@ -232,17 +233,11 @@ static boolean hasContentLength(final HttpHeaders headers) {
...
-        }
-    }
-
-    static void setContentEncoding(final HttpHeaders headers, CharSequence encoding) {
+    static void addContentEncoding(final HttpHeaders headers, CharSequence encoding) {
         // H2 does not support TE / Transfer-Encoding, so we rely in the presentation encoding only.
         // https://tools.ietf.org/html/rfc7540#section-8.1.2.2
-        headers.set(CONTENT_ENCODING, encoding);
-        headers.set(VARY, CONTENT_ENCODING);
+        headers.add(CONTENT_ENCODING, encoding);
```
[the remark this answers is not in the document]
Scottmitch: content encodings can be layered. if we do a `set` we lose information and the peer may not be able to decode it.

https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.2.2
> Content-Encoding = 1#content-coding
> If one or more encodings have been applied to a representation, the
   sender that applied the encodings MUST generate a Content-Encoding
   header field that lists the content codings in the order in which
   they were applied.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpStreamingSerializerDeserializer.java
```diff
@@ -0,0 +1,111 @@
...
+        this.headersSerializeConsumer = requireNonNull(headersSerializeConsumer);
+        this.headersDeserializePredicate = requireNonNull(headersDeserializePredicate);
+    }
+
+    @Override
+    public BlockingIterable<Buffer> serialize(final HttpHeaders headers, final BlockingIterable<T> value,
+                                              final BufferAllocator allocator) {
+        // Do the headers modification eagerly, otherwise if this is done lazily the headers would have already been
+        // written and the modification will not be applied in time.
+        headersSerializeConsumer.accept(headers);
```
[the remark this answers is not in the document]
Scottmitch: the logic pattern is similar but the logic isn't shared. A base class could have the `Consumer<HttpHeaders>` and `Predicate<HttpHeaders>` but I don't think it is worth having a common hierarchy just for this.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpSerializers.java
Scottmitch: we can as a followup. we don't expose `byte[]` so I'm not sure its necessary. this method is to accommodate "stream content with a-prior known framing in pre-existing data type" and the other side is expected to use the corresponding deserialization directly (e.g. json/xml/etc.).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpSerializers.java
Scottmitch: This was intended to support the same use case as described in https://github.com/apple/servicetalk/pull/1673#discussion_r684544839 ... if folks need deserialization specific variants we can always add more methods later, or they can use the `HttpStreamingSerializerDeserializer` variants and no-op the serialization side as short term approach.

servicetalk-examples/http/serialization/src/main/java/io/servicetalk/examples/http/serialization/async/streaming/PojoStreamingClient.java
```diff
@@ -39,9 +35,9 @@ public static void main(String[] args) throws Exception {
 
             client.request(client.post("/pojos")
                     .payloadBody(from("value1", "value2", "value3").map(CreatePojoRequest::new),
-                            serializer.serializerFor(CreatePojoRequest.class)))
+                            REQ_STREAMING_SERIALIZER))
```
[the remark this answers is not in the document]
Scottmitch: Behavior hasn't changed here but we don't cache the HTTP level `HttpStreamingSerializerDeserializer`. This object is intended to be shared and we don't necessarily want folks to create a new one each time (and look up in caches if the type is static). If folks have a more dynamic case they can add a map to cache instances (or we can enhance the factory and also cache). I understand the trade-off of APIs being 1 level removed, but lets leave as is for this PR and we can iterate if there is a better way.

servicetalk-data-jackson/src/main/java/io/servicetalk/data/jackson/JacksonSerializerCache.java
```diff
@@ -0,0 +1,126 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Caches instances of {@link SerializerDeserializer} and {@link StreamingSerializerDeserializer} for jackson.
+ */
+public final class JacksonSerializerCache {
```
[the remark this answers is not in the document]
Scottmitch: I wanted to emphasize there is caching in this implementation, and we could potentially add a "factory" interface at some point ... however "factory" is fine I'll change it.

servicetalk-data-protobuf/src/main/java/io/servicetalk/data/protobuf/ProtobufSerializerCache.java
```diff
@@ -0,0 +1,68 @@
...
+import com.google.protobuf.Parser;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Caches instances of {@link SerializerDeserializer} and {@link StreamingSerializerDeserializer} for
+ * <a href="https://developers.google.com/protocol-buffers/">protocol buffer</a>.
+ */
+public final class ProtobufSerializerCache {
```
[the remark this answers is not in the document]
Scottmitch: Updated to "Factory".

servicetalk-examples/http/serialization/src/main/java/io/servicetalk/examples/http/serialization/SerializerUtils.java
```diff
@@ -0,0 +1,43 @@
...
+import static io.servicetalk.data.jackson.JacksonSerializerCache.INSTANCE;
+import static io.servicetalk.http.api.HttpSerializers.jsonSerializer;
+import static io.servicetalk.http.api.HttpSerializers.jsonStreamingSerializer;
+
+/**
+ * Utilities to cache POJO to JSON serializer instances.
+ */
+public final class SerializerUtils {
+    public static final HttpSerializerDeserializer<CreatePojoRequest> REQ_SERIALIZER =
+            jsonSerializer(INSTANCE.serializerDeserializer(CreatePojoRequest.class));
```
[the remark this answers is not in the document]
Scottmitch: I'll use `JACKSON` for brevity.

servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/backends/MetadataBackend.java
```diff
@@ -51,13 +44,13 @@ private MetadataBackend(HttpSerializationProvider serializer) {
 
         // Create random names and author for the metadata
         Metadata metadata = new Metadata(entityId, randomString(15), randomString(5));
-        return succeeded(responseFactory.ok().payloadBody(metadata, serializer.serializerFor(Metadata.class)));
+        return succeeded(responseFactory.ok().payloadBody(metadata, METADATA_SERIALIZER));
```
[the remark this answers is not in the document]
Scottmitch: dup comment of https://github.com/apple/servicetalk/pull/1673#discussion_r684565068 .. I'll leave as is for now and we can iterate.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpRequest.java
```diff
@@ -49,14 +50,38 @@ default InputStream payloadBodyInputStream() {
...
+     * @param <T> The resulting type of the deserialization operation.
+     * @return The results of the deserialization operation.
+     */
+    <T> BlockingIterable<T> payloadBody(HttpStreamingDeserializer<T> deserializer);
+
+    /**
+     * Get the {@link HttpMessageBodyIterable} for this request.
+     * @return the {@link HttpMessageBodyIterable} for this request.
+     */
+    HttpMessageBodyIterable<Buffer> messageBody();
```
Scottmitch, line 75: continuation of https://github.com/apple/servicetalk/pull/1673#discussion_r683957514

@idelpivnitskiy @chemicL - note that I removed `BlockingStreamingHttpMessageBody` originally introduced in this PR in favor of `HttpMessageBodyIterable`. The motivation is the functional APIs such as TrailersTransformer doesn't lend themselves to the sequential programming style of traditional blocking APIs. I meant to deprecate the transform methods on the blocking request/response APIs but forgot (will push this change). The `BlockingIterable` types are still cumbersome to create but I think that can be improved over time (more utilities) and also we have a task to investigate removal of `BlockingIterable` (in favor of `ClosableIterable`).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpMessageBody.java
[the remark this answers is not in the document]
Scottmitch: this code has been updated ... comment continued here: https://github.com/apple/servicetalk/pull/1673#discussion_r684566005

servicetalk-examples/http/helloworld/src/main/java/io/servicetalk/examples/http/helloworld/async/streaming/HelloWorldStreamingClient.java
```diff
@@ -32,7 +32,7 @@ public static void main(String[] args) throws Exception {
             CountDownLatch responseProcessedLatch = new CountDownLatch(1);
             client.request(client.get("/sayHello"))
                     .beforeOnSuccess(response -> System.out.println(response.toString((name, value) -> value)))
-                    .flatMapPublisher(resp -> resp.payloadBody(textDeserializer()))
+                    .flatMapPublisher(resp -> resp.payloadBody(textSerializerUtf8FixLen()))
```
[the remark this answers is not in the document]
Scottmitch: The streaming API requires a `HttpStreamingDeserializer` which `textSerializerUtf8` doesn't satisfy. I'll clarify the method name (`textSerializerUtf8FixLen` -> `appSerializerUtf8FixLen`) and the javadocs of these methods to clarify.

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1294

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1507

PR #1675, opened 2021-07-16
Permit copy for `TestExecutor` `AsyncContextMap`

bondolo:
As observed by @chemicL

Motivation:
The `TestExecutor` was recently enhanced to install an invalid
`AsyncContextMap` when executing `Runnable` tasks to check for tasks
which did not properly save/restore the appropriate async context. The
invalid map should allow itself to be copied, producing an empty map,
for usages which want to initiate a new subscribe operation.
Modifications:
`copy()` returns a new empty map rather than generating an exception.
Result:
Now possible to subscribe() within a `Runnable` without providing an
`AsyncContextMap`.

### DIFF
--- a/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
+++ b/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
@@ -418,7 +418,7 @@ public class TestExecutor implements Executor {
 
         @Override
         public AsyncContextMap copy() {
-            throw invalidAccess();
+            return this;
         }
     }
 }


servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
```diff
@@ -418,7 +418,8 @@ public void clear() {
 
         @Override
         public AsyncContextMap copy() {
-            throw invalidAccess();
+            // Create a new non-broken empty map as this is most likely a subscribe()
+            return new CopyOnWriteAsyncContextMap();
```
Scottmitch, line 422: here is an alternative which uses the public APIs:

```java
AsyncContextMap map = AsyncContext.current().copy();
map.clear();
return map;
```

bondolo: In most cases the current map would be this invalid one so copying it would be just recursively calling the same method.

tkountis: I think the only problem with this approach is that we are 'enforcing' one implementation, whereas in the prod code this can be in the future a CHM.

tkountis (approved):

chemicL (approved):

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
bondolo: There are two alternative directions to take. Simple and fancy:

- Simple: Have `copy()` return `this` just as `NoopAsyncContextMap` does.
- Fancy: Add a `newContextMap()` method to the `AsyncContextProvider` interface with suitable implementation to abstract the map creation.

Opinions?

bondolo: I opted to "simple" in the PR.

Scottmitch (approved):

bondolo:
0.41 (cc76f151b91ed4c6c6ea804d42bd7fa333544545)

PR #1679, opened 2021-07-19
Write of HTTP/2 message may never complete

### LINKED ISSUE #1663: Flaky test: HttpTransportObserverTest#serverHandlerError (protocol=HTTP_2)
https://github.com/apple/servicetalk/pull/1660/checks?check_run_id=2996316294

```
java.util.concurrent.TimeoutException: serverHandlerError(io.servicetalk.http.netty.HttpProtocol) timed out after 30 seconds
...
Suppressed: java.lang.RuntimeException: Thread sleep has been interrupted
		at org.mockito.internal.verification.VerificationOverTimeImpl.sleep(VerificationOverTimeImpl.java:119)
		at org.mockito.internal.verification.VerificationOverTimeImpl.handleVerifyException(VerificationOverTimeImpl.java:100)
		at org.mockito.internal.verification.VerificationOverTimeImpl.verify(VerificationOverTimeImpl.java:86)
		at org.mockito.internal.verification.VerificationWrapper.verify(VerificationWrapper.java:19)
		at org.mockito.internal.verification.MockAwareVerificationMode.verify(MockAwareVerificationMode.java:27)
		at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:72)
		at org.mockito.internal.handler.NullResultGuardian.handle(NullResultGuardian.java:29)
		at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:35)
		at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:61)
		at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor.doIntercept(MockMethodInterceptor.java:49)
		at org.mockito.internal.creation.bytebuddy.MockMethodInterceptor$DispatcherDefaultingToRealMethod.interceptAbstract(MockMethodInterceptor.java:126)
		at io.servicetalk.transport.api.ConnectionObserver$WriteObserver$MockitoMock$798461370.writeComplete(Unknown Source)
		at io.servicetalk.http.netty.HttpTransportObserverTest.testRequestResponse(HttpTransportObserverTest.java:232)
		at io.servicetalk.http.netty.HttpTransportObserverTest.serverHandlerError(HttpTransportObserverTest.java:209)
		at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
		at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
		at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
		at java.base/java.lang.reflect.Method.invoke(Method.java:566)
		at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:688)
		at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
		at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
		at org.junit.jupiter.engine.extension.TimeoutInvocation.proceed(TimeoutInvocation.java:46)
		... 136 more
	Caused by: java.lang.InterruptedException: sleep interrupted
		at java.base/java.lang.Thread.sleep(Native Method)
		at org.mockito.internal.verification.VerificationOverTimeImpl.sleep(VerificationOverTimeImpl.java:117)
		... 157 more
```

idelpivnitskiy:
Motivation:

In #1518 we optimized HTTP/2 to write less frames when it's known that
the message body is empty. However, that change and the #1644 follow-up
for `AbstractH2DuplexHandler` incorrectly handle end of outbound stream.
`AbstractH2DuplexHandler` writes the `endStream` flag and waits for an
empty trailers. Netty h2 codec will close the stream as soon as it sees
`endStream` for request and response and will invoke
`WriteStreamSubscriber#channelClosed(...)`. If offloading is enabled,
message body can be cancelled before trailers reached the flush
strategy. As the result, `flushOnEnd` will never be triggered and some
tests hang waiting for the response that will never be flushed. It also
affects tests related to `TransportObserver` because `writeComplete`
may never be triggered.

Modifications:

- Enhance `AbstractStreamingHttpConnection` and `NettyHttpServer` to
use only meta-data when it creates a flat stream of HTTP message if the
message body is empty;
- `AbstractH2DuplexHandler`: invoke
`CloseHandler#protocolPayloadEndOutbound(...)` before writing a frame
with `endStream` flag;
- Enhance tests for `AbstractH2DuplexHandler` to verify new behavior;

Result:

Fixes #1663.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
@@ -33,6 +33,8 @@ import io.netty.handler.codec.http2.Http2DataFrame;
 import io.netty.handler.codec.http2.Http2Headers;
 import io.netty.handler.codec.http2.Http2ResetFrame;
 import io.netty.util.ReferenceCountUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
@@ -45,10 +47,13 @@ import static io.servicetalk.transport.netty.internal.ChannelCloseUtils.channelE
 
 abstract class AbstractH2DuplexHandler extends ChannelDuplexHandler {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractH2DuplexHandler.class);
+
     final BufferAllocator allocator;
     final HttpHeadersFactory headersFactory;
     final CloseHandler closeHandler;
     private final StreamObserver observer;
+    private boolean endStream;
 
     AbstractH2DuplexHandler(BufferAllocator allocator, HttpHeadersFactory headersFactory, CloseHandler closeHandler,
                             StreamObserver observer) {
@@ -69,7 +74,7 @@ abstract class AbstractH2DuplexHandler extends ChannelDuplexHandler {
 
     final void writeMetaData(ChannelHandlerContext ctx, HttpMetaData metaData, Http2Headers h2Headers,
                              ChannelPromise promise) {
-        final boolean endStream = emptyMessageBody(metaData);
+        endStream = emptyMessageBody(metaData);
         if (endStream) {
             closeHandler.protocolPayloadEndOutbound(ctx, promise);
         }
@@ -95,6 +100,7 @@ abstract class AbstractH2DuplexHandler extends ChannelDuplexHandler {
             if (h2Headers.isEmpty()) {
                 writeEmptyEndStream(ctx, promise);
             } else {
+                endStream = true;
                 ctx.write(new DefaultHttp2HeadersFrame(h2Headers, true), promise);
             }
         }
@@ -149,5 +155,14 @@ abstract class AbstractH2DuplexHandler extends ChannelDuplexHandler {
             observer.streamClosed(t);
         }
         ctx.fireChannelInactive();
+    }
+
+    final boolean endStreamSent() {
+        return endStream;
+    }
+
+    static void logMsgAfterEndStream(final Object msg) {
+        LOGGER.debug("Unexpected attempt to write a message after endStream flag has been sent: {}",
+                msg.getClass().getName());
     }
 }
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
@@ -33,8 +33,6 @@ import io.netty.handler.codec.http2.Http2DataFrame;
 import io.netty.handler.codec.http2.Http2Headers;
 import io.netty.handler.codec.http2.Http2ResetFrame;
 import io.netty.util.ReferenceCountUtil;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
@@ -47,13 +45,10 @@ import static io.servicetalk.transport.netty.internal.ChannelCloseUtils.channelE
 
 abstract class AbstractH2DuplexHandler extends ChannelDuplexHandler {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractH2DuplexHandler.class);
-
     final BufferAllocator allocator;
     final HttpHeadersFactory headersFactory;
     final CloseHandler closeHandler;
     private final StreamObserver observer;
-    private boolean endStream;
 
     AbstractH2DuplexHandler(BufferAllocator allocator, HttpHeadersFactory headersFactory, CloseHandler closeHandler,
                             StreamObserver observer) {
@@ -74,7 +69,7 @@ abstract class AbstractH2DuplexHandler extends ChannelDuplexHandler {
 
     final void writeMetaData(ChannelHandlerContext ctx, HttpMetaData metaData, Http2Headers h2Headers,
                              ChannelPromise promise) {
-        endStream = emptyMessageBody(metaData);
+        final boolean endStream = emptyMessageBody(metaData);
         if (endStream) {
             closeHandler.protocolPayloadEndOutbound(ctx, promise);
         }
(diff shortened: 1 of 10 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -266,7 +266,8 @@ public String toString() {
             this.splittingFlushStrategy = new SplittingFlushStrategy(connection.defaultFlushStrategy(),
                     itemWritten -> {
                         if (itemWritten instanceof HttpResponseMetaData) {
-                            return Start;
+                            final HttpResponseMetaData metadata = (HttpResponseMetaData) itemWritten;
+                            return emptyMessageBody(metadata) && metadata.version().major() > 1 ? End : Start;
```
idelpivnitskiy, line 270: @Scottmitch I understand that a different format of the flattened HTTP message needs additional coordination for places that expect trailers at the end. However, the perf bump of reducing the number of h2 frames is significant enough (see description of #1518) to justify these adjustments. Also, I hope that this `SplittingFlushStrategy` logic will go away after we rework the server-side internals.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
```diff
@@ -75,7 +69,10 @@ public final void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
 
     final void writeMetaData(ChannelHandlerContext ctx, HttpMetaData metaData, Http2Headers h2Headers,
                              ChannelPromise promise) {
-        endStream = !mayHaveTrailers(metaData) && isPayloadEmpty(metaData);
+        final boolean endStream = emptyMessageBody(metaData);
+        if (endStream) {
+            closeHandler.protocolPayloadEndOutbound(ctx, promise);
```
Scottmitch, line 74: `CloseHandler` API doesn't expect this method will be called multiple times. The `emptyMessageBody` method seems like an indirect way to infer there will be no further writes, but feels a bit "fragile" as it is disconnected from the actual write sequencing. I guess if we add another `concat(EMPTY_TRAILERS)` into the mix, we would already have problems bcz we would be writing on a stream we already set "endOfStream" on...

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -266,7 +266,8 @@ public String toString() {
             this.splittingFlushStrategy = new SplittingFlushStrategy(connection.defaultFlushStrategy(),
                     itemWritten -> {
                         if (itemWritten instanceof HttpResponseMetaData) {
-                            return Start;
+                            final HttpResponseMetaData metadata = (HttpResponseMetaData) itemWritten;
+                            return emptyMessageBody(metadata) && protocol().major() > 1 ? End : Start;
```
Scottmitch, line 270: nit: consider flipping the order of evaluation to short circuit on `protocol().major() > 1` which should be the less expensive to evaluate.

Scottmitch: https://github.com/apple/servicetalk/pull/1679#discussion_r672638667

> @Scottmitch I understand that a different format of the flattened HTTP message needs additional coordination for places that expect trailers at the end. However, the perf bump of reducing the number of h2 frames is significant enough (see description of #1518) to justify these adjustments.

It is not ideal to have different semantics on the flattened stream depending upon protocol. The `{headers, payload*, trailers}` sequencing was devised bcz we couldn't inspect the stream contents and didn't have the meta data retained in `PayloadInfo`. Another option is to use the `LastHttpChunk` interface style approach (we use this in Netty) to delineate message boundaries. This may incur object allocation overhead to wrap, but would may make the internal ST<->http transport API more explicit (and agnostic to protocol version).

> Also, I hope that this SplittingFlushStrategy logic will go away after we rework the server-side internals.

me too 🙏

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
Scottmitch: I wasn't able to comment on this during review, so I moved this down to here: https://github.com/apple/servicetalk/pull/1679#discussion_r672732144

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
idelpivnitskiy: Correct, I can keep `endStream` as a class variable as an additional protection from further writes.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
idelpivnitskiy: I would propose to defer alternative approaches because:
1. They will require a lot more refactoring at different layer, it may be easy to miss something.
2. Extra allocation and wrapping/unwrapping is not desirable.
3. If `SplittingFlushStrategy` will go away, we won't need any marker for the last chunk because it can be managed by publisher completion.

Scottmitch: yip fine to defer, lets just make sure this planned work is tracked somewhere (with reference back to this discussion to restore context if necessary).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
idelpivnitskiy: Discussed offline, additional protection is not necessary because netty will fire an exception anyway if we write after `endStream` was already written. Reverted: https://github.com/apple/servicetalk/pull/1679/commits/6cea6d42a70f475c9c2d9aac00f0ec22ff215509

idelpivnitskiy:
PR was approved offline by @Scottmitch, merging.

idelpivnitskiy:
cherry-picked to 0.41: https://github.com/apple/servicetalk/commit/4a829ec946293d96f23937ee5e69760de6bfbda3

PR #1682, opened 2021-07-20
Remove gRPC not performance tested statement from docs

Scottmitch:
Motivation:
The docs say that gRPC is an active work in progress and not yet
performance tested. Nightly benchmarks are run on HTTP transport now,
and this statement is no longer accurate.

Modifications:
- Remove caveat about not being performance tested from docs

Result:
Docs on gRPC more accurately reflect the current state.


bondolo (approved):

Scottmitch:
0.41 (78b5c1234)

PR #1689, opened 2021-07-21
Conditional offload `publishOn` & `subscribeOn`

bondolo:
Motivation:
There are many cases where offloading is not always required or could
be omitted because the execution is already sufficiently offloaded.
Making offload conditional allows omitting it when not required.
Modifications:
Provide `publishOn()` and `subscribeOn()` operators that take a
`BooleanSupplier` that will determine if offloading is necessary. In
most cases once the supplier returns `true` all subsequent offload
points will be offloaded.
Result:
Reduced offloading for cases which do not require offloading.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -73,6 +87,8 @@ void handleSubscribe(Subscriber subscriber, AsyncContextMap contextMap,
         private static final AtomicIntegerFieldUpdater<AbstractOffloadedSingleValueSubscriber> stateUpdater =
                 newUpdater(AbstractOffloadedSingleValueSubscriber.class, "state");
 
+        boolean hasOffloaded;
+        final BooleanSupplier offload;
```
Scottmitch, line 91: make this private?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -41,17 +42,30 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(TaskBasedAsyncCompletableOperator.class);
 
+    private volatile boolean hasOffloaded;
```
Scottmitch, line 45: this state is duplicated in `AbstractOffloadedSingleValueSubscriber#hasOffloaded`. I would suggest we keep the state at the lower layer it is harder to understand what threads access the state, and in practice I don't think it needs to be volatile.

For example the `SubscribeOn#handleSubscribe` usage of `offload` doesn't need to share state with the `Subscription` (it will happen before, ordering doesn't really matter).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -92,7 +109,11 @@ public final void onSubscribe(final Cancellable cancellable) {
                 LOGGER.trace("executing {} onSubscribe on {}",
                         this instanceof CompletableSubscriberOffloadedTerminals ? "Completable" : "Single",
                         executor);
-                executor.execute(this::deliverSignals);
+                if (offload.getAsBoolean()) {
+                    executor.execute(this::deliverSignals);
```
Scottmitch, line 113: set `hasOffloaded = true` before `executor.execute(..)`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -92,7 +109,11 @@ public final void onSubscribe(final Cancellable cancellable) {
...
+                } else {
+                    deliverSignals();
+                }
             } catch (Throwable t) {
                 // As a policy, we call the target in the calling thread when the executor is inadequately
                 // provisioned. In the future we could make this configurable.
```
Scottmitch, line 119: I missed these in previous reviews, can you remove the debugging log statements below (`LOGGER.trace("offloaded on {}", executor);`, `LOGGER.trace("terminated for {}", executor);`, etc.)?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -73,6 +87,8 @@ void handleSubscribe(Subscriber subscriber, AsyncContextMap contextMap,
         private static final AtomicIntegerFieldUpdater<AbstractOffloadedSingleValueSubscriber> stateUpdater =
                 newUpdater(AbstractOffloadedSingleValueSubscriber.class, "state");
 
+        boolean hasOffloaded;
```
Scottmitch, line 90: make this private?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -153,7 +174,11 @@ final void terminal(final Object terminal) {
                     // have not seen onSubscribe and there is a sequencing issue on the Subscriber. Either way we avoid
                     // looping and deliver the terminal event.
                     try {
-                        executor.execute(this::deliverSignals);
+                        if (hasOffloaded || offload.getAsBoolean()) {
+                            executor.execute(this::deliverSignals);
```
Scottmitch, line 178: consider adding a comment to the effect ... `no need to set hasOffloaded because this is a terminal signal`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -292,27 +321,33 @@ public void onError(final Throwable t) {
...
-                executor.execute(() -> safeCancel(cancellable));
-            } catch (Throwable t) {
-                LOGGER.warn("Failed to execute task on the executor {}. " +
-                                "Invoking Cancellable (cancel()) in the caller thread. Cancellable {}. ",
-                        executor, cancellable, t);
-                // As a policy, we call the target in the calling thread when the executor is inadequately
-                // provisioned. In the future we could make this configurable.
+            if (offload.getAsBoolean()) {
+                try {
+                    LOGGER.debug("executing Completable cancel on {}", executor);
```
Scottmitch, line 337: can we remove the debug/development logs? we can always instrument with operators if necessary (e.g. `afterCancel(...)`).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncPublisherOperator.java
```diff
@@ -102,13 +118,15 @@ protected void handleSubscribe(Subscriber<? super T> subscriber,
...
         }
 
-        OffloadedSubscriber(final Subscriber<? super T> target, final Executor executor,
+        OffloadedSubscriber(final Subscriber<? super T> target, final BooleanSupplier offload, final Executor executor,
                             final int publisherSignalQueueInitialCapacity) {
             this.target = target;
+            this.offload = offload;
             this.executor = executor;
             // Queue is bounded by request-n
             signals = newUnboundedSpscQueue(publisherSignalQueueInitialCapacity);
```
Scottmitch, line 132: line 143: `public void onNext(final T t)` -> `public void onNext(@Nullable final T t)`

Scottmitch: also plz remove debug/development logs below (`LOGGER.trace("offloading Publisher onSubscribe on {}", executor);`, ..). operators allow for instrumentation/visibility if required.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncPublisherOperator.java
```diff
@@ -246,8 +264,12 @@ private void offerSignal(Object signal) {
             }
 
             try {
-                LOGGER.debug("delivering Subscriber signals on {}", executor);
-                executor.execute(this::deliverSignals);
+                if (offload.getAsBoolean()) {
+                    LOGGER.trace("delivering Subscriber signals on {}", executor);
```
Scottmitch, line 268: remove debug/development logs? operators allow for instrumentation/visibility if required.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncPublisherOperator.java
```diff
@@ -357,8 +383,12 @@ private void enqueueTaskIfRequired(boolean forRequestN) {
             final int oldState = stateUpdater.getAndSet(this, STATE_ENQUEUED);
             if (oldState == STATE_IDLE) {
                 try {
-                    LOGGER.debug("executing {} on {}", forRequestN ? "request" : "cancel", executor);
-                    executor.execute(this::executeTask);
+                    if (offload.getAsBoolean()) {
+                        LOGGER.debug("executing {} on {}", forRequestN ? "request" : "cancel", executor);
```
Scottmitch, line 387: remove debug/development logs? operators allow for instrumentation/visibility if required.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncPublisherOperator.java
```diff
@@ -56,20 +57,34 @@ public String toString() {
         }
     };
 
+    private volatile boolean hasOffloaded;
```
Scottmitch, line 60: inline with comment above ... lets move this into the implementations to clarify expected thread access pattern and avoid shared state between subscribe/subscription when not necessary.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncSingleOperator.java
```diff
@@ -51,20 +52,33 @@ public String toString() {
         }
     };
 
+    private volatile boolean hasOffloaded;
```
Scottmitch, line 55: inline with comment above ... lets move this into the implementations to clarify expected thread access pattern and avoid shared state between subscribe/subscription when not necessary.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1387,7 +1388,23 @@ public final Completable liftAsync(CompletableOperator operator) {
...
+    }
+
+    /**
+     * Creates a new {@link Completable} that will use the passed {@link Executor} to invoke all {@link Subscriber}
+     * methods.
+     * This method does <strong>not</strong> override preceding {@link Executor}s, if any, specified for {@code this}
+     * {@link Completable}. Only subsequent operations, if any, added in this execution chain will use this
+     * {@link Executor}.
+     *
+     * @param offload Offload only if supplier returns true otherwise continue execution on the same thread.
```
Scottmitch, line 1401: also clarify the return value serves as a hint. for example even if this returns `{@code false}` offloading may still be done (e.g. to preserve ordering).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncSingleOperator.java
bondolo: I put it in the `TaskBasedAsync(Completable|Publisher|Single)Operator` because it is referenced by the `SubscribeOn::handleSubscribe` in addition to the signals offloading classes.

Scottmitch: https://github.com/apple/servicetalk/pull/1689#discussion_r674351132

> For example the SubscribeOn#handleSubscribe usage of offload doesn't need to share state with the Subscription (it will happen before, ordering doesn't really matter).

I'm not sure it needs to be shared. Does `SubscribeOn#handleSubscribe` have to care about the `hasOffloaded` state?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncPublisherOperator.java
bondolo: I have fixed a few cases where `debug` was used rather than `trace` but am reluctant to remove them entirely. They do provide more information than the operators can access.

Scottmitch: in cases where we can add instrumentation at higher layers lets take advantage of that as we have more flexibility (customize how info is captured ... logging, format, metrics, etc.). In this case we have visibility into the executor but in most cases you can derive that info at higher layers from the currentThread() IIUC.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncPublisherOperator.java
```diff
@@ -68,8 +72,8 @@ Executor executor() {
     }
 
     @Override
-    protected void handleSubscribe(Subscriber<? super T> subscriber,
-                                   AsyncContextMap contextMap, AsyncContextProvider contextProvider) {
+    protected void handleSubscribe(final Subscriber<? super T> subscriber,
```
Scottmitch, line 75: nit: package private

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
```diff
@@ -110,8 +107,11 @@ void handleSubscribe(final Subscriber subscriber,
                 Subscriber wrapped = contextProvider.wrapCancellable(subscriber, contextMap);
 
                 // offload the remainder of subscribe()
-                LOGGER.trace("Offloading Completable subscribe() on {}", executor());
-                executor().execute(() -> super.handleSubscribe(wrapped, contextMap, contextProvider));
+                if (shouldOffload.getAsBoolean()) {
```
Scottmitch, line 110: `public final Publisher<T> subscribeOn(Executor executor, BooleanSupplier shouldOffload)` -> `public final Publisher<T> subscribeOn(Executor executor, Supplier<? extends BooleanSupplier> shouldOffload)`

inlined with our discussion yesterday I think we can change the operator signature to a `Supplier` and just override `handleSubscribe` such that you don't have to call `apply`:

```java
private static final class SubscribeOn<T> extends AbstractNoHandleSubscribePublisher<T> {
  private final Supplier<? extends BooleanSupplier> shouldOffloadSupplier; // set in constructor

  @Override
  final void handleSubscribe(Subscriber<? super R> subscriber,
                               AsyncContextMap contextMap, AsyncContextProvider contextProvider) {
     BooleanSupplier shouldOffload = shouldOffloadSupplier.get();

     // Note that the Executor is wrapped by default to preserve async context, so if the executor is
     // wrapped we don't have to re-wrap the Subscriber. Also we only "need" to preserve async context
     // when offloading happens, as otherwise this operator is synchronous. add tests for this :)
     Subscriber<? super T> wrapped = new OffloadedSubscriptionSubscriber<>(subscriber, shouldOffload, executor());
     if (shouldOffload.getAsBoolean()) {
       executor().execute(() -> super.handleSubscribe(wrapped, contextMap, contextProvider));
     } else {
       super.handleSubscribe(wrapped, contextMap, contextProvider);
     }
  }
}
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
```diff
@@ -110,8 +107,11 @@ void handleSubscribe(final Subscriber subscriber,
                 Subscriber wrapped = contextProvider.wrapCancellable(subscriber, contextMap);
```
Scottmitch, line 107: (can't comment on `AbstractAsynchronousPublisherOperator`). I just noticed that `AbstractAsynchronousPublisherOperator` allows you to override `handleSubscribe` however the overriding class to un-do the AsyncContext wrapping. We should make `AbstractAsynchronousPublisherOperator#handleSubscribe` final (just like `AbstractSynchronousPublisherOperator`) to avoid this confusion.

Scottmitch (changes requested):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -73,26 +81,40 @@ void handleSubscribe(Subscriber subscriber, AsyncContextMap contextMap,
...
         private volatile int state = STATE_INIT;
+        private boolean hasOffloaded;
 
-        AbstractOffloadedSingleValueSubscriber(final Executor executor) {
+        AbstractOffloadedSingleValueSubscriber(final BooleanSupplier shouldOffload, final Executor executor) {
+            this.shouldOffload = shouldOffload;
             this.executor = executor;
         }
 
+        private boolean offload() {
```
Scottmitch, line 99: nit: `offload` -> `shouldOffload`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -292,27 +313,33 @@ public void onError(final Throwable t) {
...
-            try {
-                LOGGER.debug("executing Completable cancel on {}", executor);
-                executor.execute(() -> safeCancel(cancellable));
-            } catch (Throwable t) {
-                LOGGER.warn("Failed to execute task on the executor {}. " +
-                                "Invoking Cancellable (cancel()) in the caller thread. Cancellable {}. ",
-                        executor, cancellable, t);
-                // As a policy, we call the target in the calling thread when the executor is inadequately
-                // provisioned. In the future we could make this configurable.
+            if (shouldOffload.getAsBoolean()) {
```
Scottmitch, line 328: since this is user code we should include this in the try/catch.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -73,26 +81,40 @@ void handleSubscribe(Subscriber subscriber, AsyncContextMap contextMap,
...
         // Visibility: Task submitted to executor happens-before task execution.
         private Cancellable cancellable;
         @Nullable
         private Object terminal;
         private volatile int state = STATE_INIT;
+        private boolean hasOffloaded;
 
-        AbstractOffloadedSingleValueSubscriber(final Executor executor) {
+        AbstractOffloadedSingleValueSubscriber(final BooleanSupplier shouldOffload, final Executor executor) {
+            this.shouldOffload = shouldOffload;
```
Scottmitch, line 95: `requireNonNull(shouldOffload)`?

Scottmitch (changes requested):
few more changes requested but looking good. I avoided duplicate comments as much as possible, but most of them apply equally to all 3 publisher/single/completable.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1387,7 +1388,25 @@ public final Completable liftAsync(CompletableOperator operator) {
...
+     * {@link Executor}.
+     *
+     * @param executor {@link Executor} to use.
+     * @param shouldOffload provides a hint whether offloading to executor can be omitted. If true is ever returned then
+     * all subsequent offload opportunities will be offloaded to the provided executor. If only false is returned then
+     * offloading to executor may still occur, for example, to preserve item or invocation ordering.
+     * @return A new {@link Completable} that will use the passed {@link Executor} to invoke all methods on the
+     * {@link Subscriber}.
+     */
+    public final Completable publishOn(Executor executor, Supplier<BooleanSupplier> shouldOffload) {
```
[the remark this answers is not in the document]
bondolo: Sure. Seems of limited value for this particular case but good for consistency where the supplied type might be more "interesting".

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
bondolo: I put it in the `BooleanSupplier shouldOffload()` method instead.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
```diff
@@ -79,7 +75,11 @@ void handleSubscribe(final Subscriber subscriber,
             // re-wrap the subscriber so that async context is restored during offloading.
             Subscriber wrapped = contextProvider.wrapCompletableSubscriber(subscriber, contextMap);
```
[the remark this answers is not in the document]
bondolo: No, I haven't yet. Sorry. I forgot to before the last PR update.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncPublisherOperator.java
```diff
@@ -59,21 +59,22 @@ public String toString() {
...
+                                    final Supplier<? extends BooleanSupplier> shouldOffloadSupplier,
+                                    final Executor executor) {
         this.original = original;
         this.shouldOffloadSupplier = shouldOffloadSupplier;
         this.executor = executor;
     }
 
-    BooleanSupplier shouldOffload() {
-        return shouldOffloadSupplier.get();
+    final BooleanSupplier shouldOffload() {
```
Scottmitch, line 73: nit: consider renaming (e.g. `shouldOffloadSupplier`) to avoid overloaded naming with the other internal `shouldOffload`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
```diff
@@ -127,23 +127,16 @@ public void accept(final Runnable runnable) {
...
 
     /**
-     * {@link AutoCloseable} interface will invoke {@link ExecutorService#shutdown()}.
+     * {@link Runnable} interface will invoke {@link ExecutorService#shutdown()}.
      */
-    private interface InternalExecutor extends Function<Runnable, Cancellable>, AutoCloseable {
-
-        @Override
-        void close();
+    private interface InternalExecutor extends Function<Runnable, Cancellable>, Runnable {
```
Scottmitch, line 132: doesn't this revert a change you recently make?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
```diff
@@ -64,21 +64,22 @@ static Completable subscribeOn(Completable original, Executor executor) {
...
         }
 
         @Override
         void handleSubscribe(final Subscriber subscriber,
                              final AsyncContextMap contextMap, final AsyncContextProvider contextProvider) {
             // re-wrap the subscriber so that async context is restored during offloading.
             Subscriber wrapped = contextProvider.wrapCompletableSubscriber(subscriber, contextMap);
-            super.handleSubscribe(wrapped, contextMap, contextProvider);
+
+            BooleanSupplier shouldOffload = shouldOffloadSupplier();
```
Scottmitch, line 78: looks like https://github.com/apple/servicetalk/pull/1689#discussion_r675885472 was collapsed but still not resolved

> since we are invoking user code here, we should wrap in try/catch like in `SubscribeOn` below.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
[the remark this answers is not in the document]
Scottmitch: wrapping was removed for `PublishOn` in publisher/single. should be removed here too, IIUC?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnSingles.java
```diff
@@ -60,21 +60,19 @@ private PublishAndSubscribeOnSingles() {
...
-        }
-
-        @Override
-        public void handleSubscribe(final Subscriber<? super T> subscriber,
+        void handleSubscribe(final Subscriber<? super T> subscriber,
                                     final AsyncContextMap contextMap, final AsyncContextProvider contextProvider) {
-            // re-wrap the subscriber so that async context is restored during offloading.
-            Subscriber<? super T> wrapped = contextProvider.wrapSingleSubscriber(subscriber, contextMap);
-            super.handleSubscribe(wrapped, contextMap, contextProvider);
+            BooleanSupplier shouldOffload = shouldOffloadSupplier();
```
Scottmitch, line 71: this is user code, we should wrap in try/catch

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -292,27 +314,42 @@ public void onError(final Throwable t) {
...
 
         @Override
         public void cancel() {
+            boolean offloadCancel;
             try {
-                LOGGER.debug("executing Completable cancel on {}", executor);
-                executor.execute(() -> safeCancel(cancellable));
+                offloadCancel = shouldOffload.getAsBoolean();
             } catch (Throwable t) {
-                LOGGER.warn("Failed to execute task on the executor {}. " +
```
Scottmitch, line 308: we should still log a warning as throwing isn't supported and we make an assumption about offloading that may have impacts on the application.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -292,27 +314,42 @@ public void onError(final Throwable t) {
...
-                                "Invoking Cancellable (cancel()) in the caller thread. Cancellable {}. ",
-                        executor, cancellable, t);
-                // As a policy, we call the target in the calling thread when the executor is inadequately
-                // provisioned. In the future we could make this configurable.
+                // As a policy, we call the target in the calling thread when the "hint" fails. In the future we could
+                // make this configurable.
+                offloadCancel = false;
```
Scottmitch, line 335: lets play it safe here and offload. if folks are blocking and we skip offloading this may lead to bad times (blocking IO thread, increase latency, etc.), if folks are non-blocking and we offload they will have to use the thread pool (assumed less impact).

Scottmitch (changes requested):
last round of changes :)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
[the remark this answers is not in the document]
bondolo: Yes, somehow I missed that one.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
bondolo: Yes, weird. I had not noticed this strange revert.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
bondolo: Now checked all 6 spots and all are wrapped as well as all sites where the `BooleanSupplier` is invoked.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
bondolo: Added

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
bondolo: OK.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
bondolo: Yes, very strange. I had completely missed that it reverted the other change.

bondolo:
Depends on #1699. There will be at least one more update/rebase on `main` to remove the TestExecutor changes in this PR.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java
```diff
@@ -292,27 +314,43 @@ public void onError(final Throwable t) {
...
+                offloadCancel = shouldOffload.getAsBoolean();
             } catch (Throwable t) {
-                LOGGER.warn("Failed to execute task on the executor {}. " +
-                                "Invoking Cancellable (cancel()) in the caller thread. Cancellable {}. ",
-                        executor, cancellable, t);
-                // As a policy, we call the target in the calling thread when the executor is inadequately
-                // provisioned. In the future we could make this configurable.
+                // As a policy, we offload to call the target when the "hint" fails. In the future we could make this
+                // configurable.
+                LOGGER.warn("Offloading disabled because {} threw ", shouldOffload, t);
```
Scottmitch, line 335: this log statement isn't consistent with what the code is doing. consider rewording:

`LOGGER.warn("Offloading hint Supplier {} threw", shouldOffload, t);`

Scottmitch (approved):

bondolo:
Test failure was expected due to removal of `TestExecutor` changes from PR. #1699 is now merged.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1387,7 +1388,24 @@ public final Completable liftAsync(CompletableOperator operator) {
...
+     * {@link Completable}. Only subsequent operations, if any, added in this execution chain will use this
+     * {@link Executor}.
+     *
+     * @param executor {@link Executor} to use.
+     * @param shouldOffload provides a hint whether offloading to executor can be omitted. Offloading may still occur
+     * even if {@code false} is returned in order to preserve signal ordering.
+     * @return A new {@link Completable} that will use the passed {@link Executor} to invoke all methods on the
+     * {@link Subscriber}.
+     */
+    public final Completable publishOn(Executor executor, Supplier<? extends BooleanSupplier> shouldOffload) {
```
idelpivnitskiy, line 1407: 1. Based on the test that utilizes this method, it looks like the `Supplier` is not required. Users who want to have a `BooleanSupplier` per subscribe, can wrap this operator with `defer`. Can we simplify API and remove the `Supplier`?
2. `BooleanSupplier` doesn't give users any information about which method is invoked. Did you consider  `Predicate` instead that takes an enum value saying what method is invoked? Examples: users may want to offload only terminal signals for the `Publisher` or they may want to offload `onComplete()` but not `onError`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnPublishers.java
```diff
@@ -88,26 +94,26 @@ public void handleSubscribe(final Subscriber<? super T> subscriber,
...
-                        contextProvider.wrapSubscription(subscriber, contextMap);
-
-                // offload the remainder of subscribe()
-                LOGGER.trace("Offloading Publisher subscribe() on {}", executor());
-                executor().execute(() -> super.handleSubscribe(wrapped, contextMap, contextProvider));
+                BooleanSupplier shouldOffload = shouldOffloadSupplier();
+                final Subscriber<? super T> upstreamSubscriber =
+                        new OffloadedSubscriptionSubscriber<>(subscriber, shouldOffload, executor());
+
+                if (shouldOffload.getAsBoolean()) {
```
idelpivnitskiy, line 110: General comment for all async sources: I see that inside `publishOn` operator we log and ignore exceptions from the `BooleanSupplier`, but `subscribeOn` does not ignore them and terminates the `Subscriber`. Can we be consistent?

idelpivnitskiy (commented):
Like new feature! Submitted a follow-up to improve javadoc #1713 and have a few questions:

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
bondolo: > * Based on the test that utilizes this method, it looks like the `Supplier` is not required. Users who want to have a `BooleanSupplier` per subscribe, can wrap this operator with `defer`. Can we simplify API and remove the `Supplier`?

I like this idea as it is a nice simplification.

> * `BooleanSupplier` doesn't give users any information about which method is invoked. Did you consider  `Predicate` instead that takes an enum value saying what method is invoked? Examples: users may want to offload only terminal signals for the `Publisher` or they may want to offload `onComplete()` but not `onError`.

The intention was not to provide such fine-grained decisions but rather to base the decision to offload on the context such as "currently on the io thread". There doesn't yet seem to be a clear case where this is necessary but it could be added if we find a use case.

bondolo: @Scottmitch WDYT about removing the extra layer of `Supplier` since the few who might want a `BooleanSupplier` per subscribe can use defer?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnPublishers.java
bondolo: This may be moot if we remove the `Supplier<? extends BooleanSupplier>` layer as in the publishOn case it was only catching the possible exception from the `Supplier.get()`. Sufficient?

idelpivnitskiy: I was referring to this method:
https://github.com/apple/servicetalk/blob/603b2dc4f2136c1acd153d21afb94d960fb137c2/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncCompletableOperator.java#L348-L355

It's used to invoke the `BooleanSupplier`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
idelpivnitskiy: +1, we may also revisit `Supplier` for `scanWith` and `scanWithLifetime`, those can also use `defer`. Non of the other operators use a `Supplier` to provide a function (map/flatMap/etc) per subscribe. So, better to be consistent.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnPublishers.java
bondolo: Gotcha. I agree that it should cancel and the existing comments in the methods calling `shouldOffload()` indicate that is the required policy.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
idelpivnitskiy: `Supplier` was removed in #1723, thank you @bondolo!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnPublishers.java
idelpivnitskiy: This was improved in #1723, thank you @bondolo!

PR #1691, opened 2021-07-21
`DefaultExecutor` internally use `AutoCloseable`

bondolo:
Motivation:
`DefaultExecutor.InternalExecutor` currently implements `Runnable`
which is used to close the executor. This is odd.
Modifications:
Instead implement the `AutoCloseable` interface and rename the `apply`
method to `schedule`.
Result:
More conventional interface and method name use.


bondolo:
This change is internal only; the most likely place people would encounter the odd method names is in stack traces.

Scottmitch (approved):

bondolo:
0.41 (a9916956928406fe93062ef8d2ca0b33a2cd1307)

PR #1692, opened 2021-07-21
AbstractRetryingFilterBuilder to require jitter

Scottmitch:
Motivation:
AbstractRetryingFilterBuilder currently allows configuration with no
jitter, but this will trigger an assertion error at build time. In
practice most use cases are encouraged to use some form of jitter so we
should deprecate the non-jitter build methods and make sure they don't
trigger an assertion error.

Modifications:
- Deprecate buildWithImmediateRetries, buildWithConstantBackoff, and
  buildWithConstantBackoff methods on AbstractRetryingFilterBuilder.
- Enhance jitter bounds upfront validation in RepeatStrategies and
  RetryStrategies to throw early if we can easily detect invalid
  arguments.

Result:
No more assertion error in AbstractRetryingFilterBuilder due to lack of
jitter, users encouraged to use jitter in retry strategies.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/AbstractRetryingFilterBuilder.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/AbstractRetryingFilterBuilder.java
@@ -34,7 +34,7 @@ import static io.servicetalk.concurrent.api.RetryStrategies.retryWithConstantBac
 import static io.servicetalk.concurrent.api.RetryStrategies.retryWithExponentialBackoffDeltaJitter;
 import static io.servicetalk.concurrent.api.RetryStrategies.retryWithExponentialBackoffFullJitter;
 import static java.time.Duration.ofDays;
-import static java.time.Duration.ofNanos;
+import static java.time.Duration.ofMillis;
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -49,7 +49,7 @@ import static java.util.Objects.requireNonNull;
 public abstract class AbstractRetryingFilterBuilder<Builder
         extends AbstractRetryingFilterBuilder<Builder, Filter, Meta>, Filter, Meta> {
     private static final Duration FULL_JITTER = ofDays(1024);
-    private static final Duration NULL_JITTER = ofNanos(1);
+    private static final Duration NULL_JITTER = ofMillis(1);
     private int maxRetries;
     @Nullable
     private BiPredicate<Meta, Throwable> retryForPredicate;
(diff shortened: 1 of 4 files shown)


servicetalk-client-api/src/main/java/io/servicetalk/client/api/AbstractRetryingFilterBuilder.java
```diff
@@ -48,6 +49,7 @@
 public abstract class AbstractRetryingFilterBuilder<Builder
         extends AbstractRetryingFilterBuilder<Builder, Filter, Meta>, Filter, Meta> {
     private static final Duration FULL_JITTER = ofDays(1024);
+    private static final Duration NULL_JITTER = ofNanos(1);
```
bondolo, line 52: Perhaps MIN_JITTER as it is still some (albeit trifling) jitter. Would a larger value, ofMillis(1) perhaps, be a better default for the cases where we are providing jitter value?

servicetalk-client-api/src/main/java/io/servicetalk/client/api/AbstractRetryingFilterBuilder.java
```diff
@@ -287,7 +294,7 @@ private ReadOnlyRetryableSettings(final int maxRetries,
             this.initialDelay = initialDelay;
             this.timerExecutor = timerExecutor;
             this.exponential = exponential;
-            this.jitter = jitter;
+            this.jitter = requireNonNull(jitter);
```
bondolo, line 297: Perhaps range check here rather than in Strategies?

servicetalk-client-api/src/main/java/io/servicetalk/client/api/AbstractRetryingFilterBuilder.java
```diff
@@ -287,7 +294,7 @@ private ReadOnlyRetryableSettings(final int maxRetries,
             this.initialDelay = initialDelay;
```
bondolo, line 294: Perhaps range check this here as well?

servicetalk-client-api/src/main/java/io/servicetalk/client/api/AbstractRetryingFilterBuilder.java
Scottmitch: I would prefer to keep the value validation in a single spot to avoid duplicating logic.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/AbstractRetryingFilterBuilder.java
Scottmitch: same as above I would prefer to keep validation in a single spot to avoid duplicating logic (and getting out of date).

tkountis (approved):
lgtm

chemicL (approved):

servicetalk-client-api/src/main/java/io/servicetalk/client/api/AbstractRetryingFilterBuilder.java
bondolo: Ok, the suggestion was about validating at creation/build vs usage. Build time validation should usually be preferred where possible because the usage may be temporally and organizationally disconnected from the error and thus harder to diagnose and resolve.

Scottmitch:
0.41 (813ab139a)

PR #1695, opened 2021-07-23
Simplify `HttpExecutionStrategy `

bondolo:
Motivation:
The `Executor` instance provided within the `HttpExecutionStrategy` is of
limited value as it is frequently duplicative of the `Executor`
instance in the `ExecutionContext` and inconsistently used. When
another executor is need it can be provided in the filter or service
without complicating normal usage.
Modifications:
Removes `Executor` and `MergeStrategy` from `HttpExecutionStrategy `
simplifying the behavior.
Result:
Simplified `Executor` behaviour.


bondolo:
`PredicateRouterOffloadingTest` is a good example of the changes that might be needed in apps as a result of the removal of `Executor` from `ExecutionStrategy`.

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
```diff
@@ -496,13 +495,6 @@ private static HttpExecutionStrategy difference(final Executor fallback,
             return right;
         }
 
-        final Executor rightExecutor = right.executor();
```
Scottmitch, line 499: how does this relate to `HttpExecutionStrategies#difference`, and does this method still need to exist?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategies.java
```diff
@@ -131,11 +115,6 @@ public static HttpExecutionStrategy difference(final Executor fallback,
         if (noOffloads(left)) {
```
Scottmitch, line 115: `final Executor fallback,` parameter is no longer used

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
Scottmitch: `final Executor fallback,` parameter is no longer used

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategies.java
bondolo: removed

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
bondolo: It is now the same as the `HttpExecutionStrategies#difference` implementation

bondolo: Now removed.

bondolo:
`JerseyRouterTestSuite` has a lot of failures but I can't find where the assumptions/behaviour has changed.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
```diff
@@ -45,40 +40,9 @@
...
-                        // (meta + data) stream. In this case, we need to preserve the service contract and hence
-                        // have to offload both meta and data separately.
-                        resp.map(r -> r.transformMessageBody(
-                                p -> p.subscribeOn(e, shouldOffload))).subscribeOn(e, shouldOffload) :
-                        resp;
-            }
-        };
+        StreamingHttpService offloadedService =
+                StreamingHttpService.offloadService(strategy, null, shouldOffload, offloaded.create(service));
+        return new StreamingHttpServiceFilter(offloadedService) { };
```
Scottmitch, line 45: `new StreamingHttpServiceFilter(offloadedService) { };` -> `new StreamingHttpServiceFilter(offloadedService);`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpService.java
```diff
@@ -46,4 +52,94 @@
...
+     * Wraps the passed {@link StreamingHttpService} to apply the provided {@link HttpExecutionStrategy} for offloading.
+     *
+     * @param strategy {@link HttpExecutionStrategy} to use for offloading.
+     * @param executor {@link Executor} to use as executor or {@code null} to use Execution context executor.
+     * @param shouldOffload If {@link BooleanSupplier} returns {@code true} then offload to executor otherwise continue
+     * execution on calling thread.
+     * @param service {@link StreamingHttpService} to wrap.
+     * @return Wrapped {@link StreamingHttpService}.
+     */
+    static StreamingHttpService offloadService(final HttpExecutionStrategy strategy,
```
Scottmitch, line 66: lets move this into a different utility file. most users aren't expected to know this exists and don't need to know about it.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpService.java
```diff
@@ -46,4 +52,94 @@
...
+                                                                final StreamingHttpResponseFactory responseFactory) {
+                        // We compute the difference between the ExecutionStrategy from the current ExecutionContext and
+                        // this ExecutionStrategy to understand if we need to offload more than we already offloaded:
+                        final HttpExecutionStrategy diff =
+                                difference(ctx.executionContext().executionStrategy(), strategy);
+
+                        Executor useExecutor = null != executor ? executor : ctx.executionContext().executor();
+
+                        // The service should see this ExecutionStrategy inside the ExecutionContext:
+                        final HttpServiceContext wrappedCtx =
```
Scottmitch, line 83: ideally we can avoid allocation/wrapping on each request. I guess if we need to we could cache this object in the future.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractStreamingHttpConnection.java
```diff
@@ -119,8 +119,20 @@ public final HttpConnectionContext connectionContext() {
...
             }
-            return strategy.invokeClient(executionContext.executor(), flatRequest,
-                    determineFlushStrategyForApi(request), this).subscribeShareContext();
+
+            if (strategy.isSendOffloaded()) {
+                flatRequest = flatRequest.subscribeOn(executionContext.executor());
+            }
+            Single<StreamingHttpResponse> resp = invokeClient(flatRequest, determineFlushStrategyForApi(request));
+            if (strategy.isMetadataReceiveOffloaded()) {
+                resp = resp.publishOn(executionContext.executor());
```
Scottmitch, line 128: I would expect each of the `subscribeOn` and `publishOn` in this method to be using something like `Thread.currentThread() instanceof IoThreadFactory.IoThread` for the `BooleanSupplier`, wdyt?

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/JerseyRouteExecutionStrategy.java
```diff
@@ -0,0 +1,112 @@
...
+import io.servicetalk.http.api.HttpExecutionStrategy;
+
+import javax.annotation.Nullable;
+
+import static java.util.Objects.requireNonNull;
+
+class JerseyRouteExecutionStrategy implements HttpExecutionStrategy {
+
+    private final HttpExecutionStrategy delegate;
+    @Nullable
```
Scottmitch, line 30: is this necessary? looks like the usage of this class prevents `@Nullable`. if it is intended then update the accessor method to have this annotation too and protect against `null` for usages (e.g. in equals, hashCode).

servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/InOrderRouter.java
```diff
@@ -68,9 +68,9 @@
                 StreamingHttpService service = pair.service();
                 final HttpExecutionStrategy strategy = pair.routeStrategy();
                 if (strategy != null) {
-                    Executor se = strategy.executor();
-                    Executor executor = null == se ? ctx.executionContext().executor() : se;
-                    service = strategy.offloadService(executor, service);
+                    Executor executor = ctx.executionContext().executor();
+                    service = StreamingHttpService.offloadService(strategy,
+                            executor, Boolean.TRUE::booleanValue, service);
```
Scottmitch, line 73: `Boolean.TRUE::booleanValue` -> should this be sensitive to `IoThread`?

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/JerseyRouteExecutionStrategy.java
bondolo: It does not need to allow `@Nullable`, removed.

servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/InOrderRouter.java
bondolo: I had left those cases alone after refactoring to use the same implementation for both `OffloadFilter` and the removed `HttpExecutionStrategy::offloadService`. I have converted them now and they work correctly so they are now conditionally offloading.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpService.java
bondolo: I was pretty sure the context never changed, but it could for conditional filters, so I didn't implement caching. If we think it is worthwhile I can add it.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractStreamingHttpConnection.java
bondolo: Seems reasonable.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpService.java
bondolo: Now in `StreamingHttpServiceToOffloadedStreamingHttpService`. The naming matches the naming of the other adapters to blocking and aggregating.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -161,8 +161,10 @@ private static void populateRoutes(final ExecutionContext executionContext,
             final String path = entry.getKey();
             final ServiceAdapterHolder adapterHolder = entry.getValue().buildRoute(executionContext);
             final StreamingHttpService route = closeable.append(adapterHolder.adaptor());
-            verifyNoOverrides(allRoutes.put(path, adapterHolder.serviceInvocationStrategy()
-                    .offloadService(executionContext.executor(), route)), path, emptyMap());
+            verifyNoOverrides(allRoutes.put(path,
+                    StreamingHttpService.offloadService(adapterHolder.serviceInvocationStrategy(),
+                            executionContext.executor(), Boolean.TRUE::booleanValue, route)),
```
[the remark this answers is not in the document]
bondolo: It could. I didn't "upgrade" them while making the rest of the renovations to avoid adding more potential problems to the mix.

Now done for all but the test cases which want to ensure that the offloading actually takes place so need to be unconditional

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
```diff
@@ -40,16 +39,6 @@ public static GrpcExecutionStrategy defaultStrategy() {
         return Builder.DEFAULT;
     }
 
-    /**
-     * The default {@link GrpcExecutionStrategy} using the passed {@link Executor}.
-     *
-     * @param executor {@link Executor} to use.
-     * @return Default {@link GrpcExecutionStrategy}.
-     */
-    public static GrpcExecutionStrategy defaultStrategy(final Executor executor) {
```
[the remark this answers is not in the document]
bondolo: Deprecations now posted in #1781

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcExecutionStrategy.java
```diff
@@ -75,29 +51,23 @@ public HttpExecutionStrategy merge(final HttpExecutionStrategy other) {
...
     @Override
-    public <T> Single<T> offloadSend(final Executor fallback, final Single<T> original) {
-        return delegate.offloadSend(fallback, original);
-    }
-
-    @Override
-    public <T> Single<T> offloadReceive(final Executor fallback, final Single<T> original) {
-        return delegate.offloadReceive(fallback, original);
+    public <T> Single<T> offloadSend(final Executor executor, final Single<T> original) {
+        return delegate.offloadSend(executor, original);
```
Scottmitch, line 55: future pr: conditional offloading was missed at the `ExecutionStrategy` level. Can you cover this with the removal of `offload*` methods?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
```diff
@@ -161,8 +163,13 @@ private static void populateRoutes(final ExecutionContext executionContext,
             final String path = entry.getKey();
             final ServiceAdapterHolder adapterHolder = entry.getValue().buildRoute(executionContext);
             final StreamingHttpService route = closeable.append(adapterHolder.adaptor());
-            verifyNoOverrides(allRoutes.put(path, adapterHolder.serviceInvocationStrategy()
-                    .offloadService(executionContext.executor(), route)), path, emptyMap());
+            verifyNoOverrides(allRoutes.put(path,
+                            StreamingHttpServiceToOffloadedStreamingHttpService.offloadService(
+                                    adapterHolder.serviceInvocationStrategy(),
+                                    executionContext.executor(),
+                                    () -> Thread.currentThread() instanceof IoThreadFactory.IoThread,
```
Scottmitch, line 170: should this take into account `isIoThreadSupported` (same question for all other usages)? Seems like we should have a utility method somewhere (e.g. `io.servicetalk.transport.api`?) to share code as this repeats in a few areas.

```java
   static BooleanSupplier shouldOffload(IoExecutor executor) {
        return executor.isIoThreadSupported() ?
                () -> Thread.currentThread() instanceof IoThreadFactory.IoThread : Boolean.TRUE::booleanValue;
    }
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractStreamingHttpConnection.java
```diff
@@ -102,11 +103,15 @@ public final HttpConnectionContext connectionContext() {
                 this::newSplicedResponse));
     }
 
+    private static boolean shouldOffload() {
```
Scottmitch, line 106: can we leverage `BooleanSupplier shouldOffload(IoExecutor executor)` here?

servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/InOrderRouter.java
```diff
@@ -68,9 +70,9 @@
                 StreamingHttpService service = pair.service();
                 final HttpExecutionStrategy strategy = pair.routeStrategy();
                 if (strategy != null) {
-                    Executor se = strategy.executor();
-                    Executor executor = null == se ? ctx.executionContext().executor() : se;
-                    service = strategy.offloadService(executor, service);
+                    Executor executor = ctx.executionContext().executor();
+                    service = StreamingHttpServiceToOffloadedStreamingHttpService.offloadService(strategy,
+                            executor, () -> Thread.currentThread() instanceof IoThreadFactory.IoThread, service);
```
Scottmitch, line 75: can we leverage `BooleanSupplier shouldOffload(IoExecutor executor)` here?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -573,6 +564,10 @@ public final ServerContext listenBlockingStreamingAndAwait(
         return doListen(connectionAcceptor, serviceContext, filteredService, drainRequestPayloadBody);
     }
 
+    private static boolean shouldOffload() {
```
Scottmitch, line 567: can we leverage `BooleanSupplier shouldOffload(IoExecutor executor)` here?

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
```diff
@@ -486,28 +519,27 @@ public Method getResourceMethod() {
...
-        if (rightExecutor != null && rightExecutor != left.executor() && rightExecutor != fallback) {
-            // Since the original offloads were done on a different executor, we need to offload again
-            // and for this we assume that the intention is to offload only the call to handle
-            return customStrategyBuilder().offloadReceiveMetadata().executor(rightExecutor).build();
+        if (right instanceof JerseyRouteExecutionStrategy) {
+            final Executor rightExecutor = ((JerseyRouteExecutionStrategy) right).executor();
+            if (rightExecutor != null && rightExecutor != contextExecutor) {
```
Scottmitch, line 534: `rightExecutor != null` -> always true?

Scottmitch (commented):
looking good ... last few questions

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRouter.java
bondolo: Yes, I would be willing. There was such a method in an earlier revision but it was banished. 😊It seems common enough to include.On key goal is to be able to use a method reference rather than a lambda so that the lambda is self-identifying in the debugger.

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
bondolo: yes, now no longer possibly null

servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/InOrderRouter.java
bondolo: yes

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractStreamingHttpConnection.java
bondolo: yes

Scottmitch (approved):

bondolo:
build failed due to automatic mis-merge.

PR #1696, opened 2021-07-23
`AllWritesPromise` should not be writable after channel is closed

idelpivnitskiy:
Motivation:

`AllWritesPromise` is closed when a channels becomes inactive or the
output side of the channel is shutdown. We should not try to write
more messages after that, even if a subscriber is not terminated yet.

Modifications:

- Consider `AllWritesPromise` as non-writable if any of the `state` flag
is set, including `CLOSE_OUTBOUND_ON_SUBSCRIBER_TERMINATION`;
- Add a test to verify further writes are discarded after
`WSS#channelClosed` is invoked;

Result:

No failed writes on netty pipeline after the channel is closed.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
@@ -337,7 +337,8 @@ final class WriteStreamSubscriber implements PublisherSource.Subscriber<Object>,
 
         boolean isWritable() {
             assert channel.eventLoop().inEventLoop();
-            return !hasAnyFlags(CHANNEL_CLOSED, SUBSCRIBER_TERMINATED, SOURCE_TERMINATED);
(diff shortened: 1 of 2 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -337,7 +337,8 @@ public final ChannelPromise removeListeners(
 
         boolean isWritable() {
             assert channel.eventLoop().inEventLoop();
-            return !hasAnyFlags(CHANNEL_CLOSED, SUBSCRIBER_TERMINATED, SOURCE_TERMINATED);
+            return !hasAnyFlags(CHANNEL_CLOSED, SUBSCRIBER_TERMINATED, SOURCE_TERMINATED,
```
Scottmitch, line 340: nit: this could be `state != 0` based upon the current state machine. (understood if you want to be more explicit about this with the current code).

Scottmitch (approved):

PR #1639, opened 2021-06-23
Offload operators

bondolo:
This change replaces use of `SignalOffloader` with operators that implement the `TaskBasedSignalOffloader` style of offloading.

There are some significant differences in behaviour:
- Non-offloading operators no longer know what `Executor` they might be executed on. This impacts `timeout()` and `liftAsync()` which may need rework.
- `subscribe()` (or `internalSubscribe()`) no longer offloads directly. Offloading only begins at the `subscribeOn() operator. This impacts primarily offloading of `before*()`, `after*()`, and `when*()` operators which must precede the offloading operator in order for them to be offloaded.

PR #1697, opened 2021-07-24
Follow-up for `SignalOffloader` removal in #1639

idelpivnitskiy:
Motivation:

Minor changes for the late review of #1639.

Modifications:

- Removed comments that are not valid anymore;
- Skip wrapping of `handleSubscribe` when we share the context;
- Remove unused methods and `Utils` class;
- Make `executor()` method `final` in `TaskBasedAsync*Operator`;
- Fix incorrect logger names;

Result:

Minor improvements.


Scottmitch (approved):

bondolo (commented):
Thank you for the cleanups. Some of this duplicates changes in #1689 and whichever PR is completed second is going to have a merge mess to clean up.

The inlining of the wrapping seems unnecessary. In part it had been retained to reduce diff chatter in the original offloading #1639 PR. Looking at the generated bytecode for an example it is identical (there is an additional local var entry in the method debug info).

idelpivnitskiy:
@bondolo sounds good, I can revert inlining to reduce the diff. The primary intend was to clean up comments and unused code. Can you please reply [there](https://github.com/apple/servicetalk/pull/1639#discussion_r675916900) and I will incorporate changes based on your answers in this PR?

bondolo (approved):

PR #1698, opened 2021-07-27
Minor javadoc updates and generics usage fix

Scottmitch:
Motivation:
Some javadocs in Publisher, Completable, Single could be clarified. One
usage of a generic type is missing the dimond operator.

Modifications:
- Enhance javadocs for "amb" related operators.
- Add diamond operator for generic type usage in ScanWithPublisher

Result:
Minor code/javadoc cleanup.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
@@ -1425,7 +1425,8 @@ public abstract class Completable {
     /**
      * Creates a new {@link Completable} that terminates with the result (either success or error) of either this
      * {@link Completable} or the passed {@code other} {@link Completable}, whichever terminates first. Therefore the
-     * result is said to be <strong>ambiguous</strong> relative to which source it originated from.
+     * result is said to be <strong>ambiguous</strong> relative to which source it originated from. After the first
+     * source terminates the non-terminated source will be cancelled.
      * <p>
      * From a sequential programming point of view this method is roughly equivalent to the following:
      * <pre>{@code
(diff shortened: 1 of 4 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1471,9 +1472,10 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
      *      }
      * }</pre>
      *
-     * @param other {@link Single} with which the result of this {@link Single} is to be ambiguated.
+     * @param other {@link Single} to subscribe to and race with this {@link Single} to propagate to the return value.
```
bondolo, line 1475: Error result will also be propagated. What is effect of cancel?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1895,9 +1899,10 @@ public static Completable mergeAllDelayError(int maxConcurrency, Completable...
      *      }
      * }</pre>
      *
-     * @param completables {@link Completable}s the result of which are to be ambiguated.
+     * @param completables {@link Completable}s to subscribe to and race to propagate to the return value.
```
bondolo, line 1902: Perhaps "result" rather than "return value".

bondolo (approved):

tkountis (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
Scottmitch: >Creates a new {@link Single} that terminates with the result (either success or error) of either this
     * {@link Single} or the passed {@code other} {@link Single}, whichever terminates first.

I'll add something on cancellation
> After the first source terminates the non-terminated source will be cancelled.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
Scottmitch: We use "return value" elsewhere in these docs, so I'll use that for consistency.

Scottmitch:
<delta too large, change is minimal, skipping back port>

PR #1699, opened 2021-07-27
TestExecutor Runnable preserve AsyncContextMap

bondolo:
Motivation:
The earlier PR #1662 (and followon #1675),  were intended to resolve
issues with handling of AsyncContextMap and
`TestExecutor.execute(Runnable)` but the nature of the problem was
misunderstood. In those cases the observed problem was that causing
tasks to execute using `TestExecutor.executeNextTask()` from a
different thread than that which enqueued the task resulted in a
new empty AsyncContextMap being used rather than the correct
AsyncContextMap. The solution in #1662 was intended to detect the
case where the map was not correctly being set by the wrapping of
the Subscription or Subscriber in the operator chain. It missed
that the behavior of TestExecutor was itself inconsistent with other
ServiceTalk executors. Executors are supposed to propagate the
AsyncContextMap from the thread invoking `Executor.execute()` to
the thread which executes the `Runnable`. This behavior was missing
in TestExecutor which originally did not set the AsyncContextMap and
after #1662 set the map to an intentionally invalid map.
Modifications:
TestExecutor captures the active `AsyncContextMap` at task creation and
when the task is eventually executed the `Runnable` is invoked with
the same map active.
Result:
TestExecutor behavior matches other executors.

### DIFF
--- a/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
+++ b/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
@@ -327,12 +327,10 @@ public class TestExecutor implements Executor {
     private static final class RunnableWrapper implements Runnable {
         private final String threadName;
         private final Runnable delegate;
-        private final AsyncContextMap contextMap;
 
         private RunnableWrapper(final String threadName, final Runnable delegate) {
             this.threadName = threadName;
-            this.delegate = delegate;
-            this.contextMap = AsyncContext.current();
+            this.delegate = AsyncContext.provider().wrapRunnable(delegate, AsyncContext.current());
         }
 
         @Override
@@ -340,12 +338,9 @@ public class TestExecutor implements Executor {
             Thread current = Thread.currentThread();
             String oldName = current.getName();
             current.setName(threadName);
-            AsyncContextMap currentMap = AsyncContext.current();
             try {
-                AsyncContext.provider().contextMap(contextMap);
                 delegate.run();
             } finally {
-                AsyncContext.provider().contextMap(currentMap);
                 current.setName(oldName);
             }
         }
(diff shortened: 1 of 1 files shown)


idelpivnitskiy (approved):
Thank you!

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
```diff
@@ -323,102 +321,33 @@ private static boolean executeOne(Queue<RunnableWrapper> tasks, AtomicInteger ta
...
+     *  thread which called {@link #execute(Runnable)}.
      */
     private static final class RunnableWrapper implements Runnable {
         private final String threadName;
         private final Runnable delegate;
+        private final AsyncContextMap contextMap;
 
         private RunnableWrapper(final String threadName, final Runnable delegate) {
             this.threadName = threadName;
             this.delegate = delegate;
```
Scottmitch, line 334: you can just do `this.delegate = AsyncContext.wrapRunnable(delegate);` and remove `AsyncContextMap`. this lets the `AsyncContext` take care of the impl details.

Scottmitch (changes requested):

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
bondolo: The intent was to avoid creating another wrapper layer and since there is already a try/finally for the thread naming it seemed worthwhile.

Scottmitch: For test utilities I think we should prioritize minimum code duplication and readability. I don't think wrapping here will be an issue. Also `AsyncContext` does some internal checking related to enable/disabled and thread type, etc. that we don't have to replicate here.

tkountis (approved):

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestExecutor.java
bondolo: OK

bondolo:
0.41 (81374425a850ef669cc6a310a671acbaab435872)

PR #1703, opened 2021-07-28
Relax OpenTracing HttpHeaders formatter to allow custom containers

tkountis:
Motivation:

Currently the HttpHeaders formatter is non-public API and strongly typed around our own HttpHeaders interface.
Users who want to use our OpenTracing API and utilize a formatter with their own carrier they end up duplicating
the same code.

Modifications:

Deprecated current solution, and introduced a KeyValue formatter that allows a custom injector and extractor.

Result:

Additional API exposed for OpenTracing users.


tkountis:
@eddie4941 FYI

servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/KeyValueFormatter.java
```diff
@@ -31,15 +29,16 @@
 import static io.servicetalk.opentracing.internal.HexUtils.validateHexBytes;
 import static java.lang.String.valueOf;
 
-final class TracingHttpHeadersFormatter implements InMemorySpanContextFormat<HttpHeaders> {
-    private static final Logger logger = LoggerFactory.getLogger(TracingHttpHeadersFormatter.class);
+public final class KeyValueFormatter implements InMemorySpanContextFormat<CharSequenceKeyValueAccessor> {
```
eddie4941, line 32: Since were breaking the name always, it might a good idea to throw the name `zipkin` or `b3` in there since this is formatter the trace context in b3/zipkin format. 

another suggestion is that we somehow provided a way for users to control the key names.

servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/CharSequenceKeyValueAccessor.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+package io.servicetalk.opentracing.inmemory;
+
+import javax.annotation.Nullable;
+
+/**
+ * An Key Value accessor interface that works with {@link CharSequence} types.
+ */
+public interface CharSequenceKeyValueAccessor {
```
eddie4941, line 23: nit: 

The typical pattern in open tracing repo is to call these `<myType>Extract` or `<myType>Inject` i.e. 

we would split this class up into two interfaces and call them `CharSequenceKeyValue(Extract|Inject)`. 

leaving the comment here in case we want to be consistent as users of open tracing may be familiar with that naming scheme.

servicetalk-opentracing-http/src/main/java/io/servicetalk/opentracing/http/HttpHeadersKeyValueAccessor.java
```diff
@@ -0,0 +1,47 @@
...
+package io.servicetalk.opentracing.http;
+
+import io.servicetalk.http.api.HttpHeaders;
+import io.servicetalk.opentracing.inmemory.CharSequenceKeyValueAccessor;
+
+import javax.annotation.Nullable;
+
+import static java.util.Objects.requireNonNull;
+
+final class HttpHeadersKeyValueAccessor implements CharSequenceKeyValueAccessor {
```
eddie4941, line 25: nit: The typical pattern is to name these `<myType>Adapter` as we see in the open tracing repo: https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/propagation/TextMapExtractAdapter.java

it may be a good idea to follow that pattern so that users familiar with those apis don't get confused.

servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/KeyValueFormatter.java
tkountis: Agreed on 'B3'. 
Regarding the keys I don't want to relax the API yet inside ST especially since this is now B3 compliant. The possibility of altering the names can still happen through the carrier set/get methods (overriding the keys).

servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/KeyValueFormatter.java
```diff
@@ -49,16 +48,17 @@
...
      * correct format.
      */
-    private TracingHttpHeadersFormatter(boolean verifyExtractedValues) {
+    private KeyValueFormatter(boolean verifyExtractedValues) {
         this.verifyExtractedValues = verifyExtractedValues;
     }
 
-    static InMemorySpanContextFormat<HttpHeaders> traceStateFormatter(boolean validateTraceKeyFormat) {
+    public static InMemorySpanContextFormat<CharSequenceKeyValueAccessor>
+    traceStateFormatter(boolean validateTraceKeyFormat) {
```
Scottmitch, line 56: `CharSequenceKeyValueAccessor` likely requires allocation on each extraction/insertion for pre-existing types (e.g. for servicetalk). An alternative approach is to make an abstract base class with consumer/function to inject/extract values from the carrier type (see below).

```java
abstract class KeyValueFormatter<T> implements InMemorySpanContextFormat<T> {
    private final TriConsumer<T, CharSequence, CharSequence> carrierInjector;
    private final BiFunction<T, CharSequence, CharSequence> carrierExtractor;

    // may make sense to put into concurrent package or somewhere else, not sure
    // if the is an existing interface in the jdk we can leverage...
    interface TriConsumer<X, Y, Z> {
        void accept(X x, Y y, Z z);
    }

   @Override
    public void inject(final InMemorySpanContext context, final T carrier) {
        carrierInjector.accept(carrier, TRACE_ID, context.toTraceId());
        ....
    }

    @Nullable
    @Override
    public InMemorySpanContext extract(final T carrier) {
        CharSequence traceId = carrierExtractor.apply(carrier, TRACE_ID);
        ....
    }
}
```

servicetalk-opentracing-inmemory/src/main/java/io/servicetalk/opentracing/inmemory/CharSequenceKeyValueAccessor.java
tkountis: The classes are now gone. They B3KeyValue formatter allows for generic injector/extractors to be passed in.

servicetalk-opentracing-inmemory/build.gradle
```diff
@@ -18,8 +18,10 @@ apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-library"
 
 dependencies {
   api project(":servicetalk-opentracing-inmemory-api")
+  api project(":servicetalk-concurrent")
```
Scottmitch, line 21: we have `BiIntPredicate` and `Function3` (and friends) in `concurrent.api` ... any reason why `TriConsumer` isn't in the same place?

PR #1705, opened 2021-07-28
Conditional offload consistent exception handling

Scottmitch:
Motivation:
If the BooleanSupplier throws in some situations we don't offload and in
others we do offload. This is a user error, but it may result in
inconsistent ordering of signals. This can be prevented if we always
offload in the event an exception is thrown.

Modifications:
- Introduce a safeShouldOffload method that returns true if an exception
  is thrown, and use it consistently in TaskBasedAsync*Operator

Result:
More consistency around exception handling for conditional offloading.


Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1706

bondolo (approved):

PR #1708, opened 2021-07-28
Add examples for redirect handling

idelpivnitskiy:
Motivation:

Demonstrate how users can handle redirects with ServiceTalk, describe
what is not copied from the original request automatically, and how
users can redirect with headers and payload body.

Modifications:
- Add `servicetalk-examples-http-redirects` module;
- Add links in documentation for the new example;

Result:

Users can see different ways how to handle redirects.

### DIFF
--- a/servicetalk-examples/docs/modules/ROOT/pages/http/index.adoc
+++ b/servicetalk-examples/docs/modules/ROOT/pages/http/index.adoc
@@ -175,7 +175,7 @@ the same across all the HTTP APIs.
 [#Redirects]
 == Redirects
 
-Extends the async "Hello World" example to demonstrate different ways how users can support redirects in ServiceTalk
+Extends the async "Hello World" example to demonstrate different ways that users can support redirects in ServiceTalk
 applications. You should read and understand the async "Hello World" example first to understand the additions this
 example adds. No separate example is needed for the other API variants as the usage of the debugging features are the
 same for all API styles.
--- a/servicetalk-examples/docs/modules/ROOT/pages/http/index.adoc
+++ b/servicetalk-examples/docs/modules/ROOT/pages/http/index.adoc
@@ -172,6 +172,24 @@ Using the following classes:
 NOTE: This example uses the link:#blocking-aggregated[blocking + aggregated] API, as the TLS/SSL configuration API is
 the same across all the HTTP APIs.
 
+[#Redirects]
+== Redirects
+
+Extends the async "Hello World" example to demonstrate different ways how users can support redirects in ServiceTalk
+applications. You should read and understand the async "Hello World" example first to understand the additions this
+example adds. No separate example is needed for the other API variants as the usage of the debugging features are the
+same for all API styles.
+
+* link:{source-root}/servicetalk-examples/http/redirects/src/main/java/io/servicetalk/examples/http/redirects/RedirectingServer.java[RedirectingServer] -
+Starts two servers, one of them (HTTP) redirects to another (HTTPS).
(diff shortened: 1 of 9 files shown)


servicetalk-examples/http/redirects/src/main/java/io/servicetalk/examples/http/redirects/ManualRedirectClient.java
```diff
@@ -0,0 +1,122 @@
...
+                        });
+
+                responseProcessedLatch.await();
+
+                // Redirect of a POST request with a payload body:
+                responseProcessedLatch = new CountDownLatch(1);
+                HttpRequest originalPost = client.post("http://localhost:8080/sayHello")
+                        .payloadBody(client.executionContext().bufferAllocator().fromAscii("some_content"));
+                client.request(originalPost)
+                        .flatMap(response -> {
```
bondolo, line 87: Handling for multiple redirects?

servicetalk-examples/http/redirects/src/main/java/io/servicetalk/examples/http/redirects/ManualRedirectClient.java
```diff
@@ -0,0 +1,122 @@
...
+                                return redirectClient.request(redirectClient
+                                        .newRequest(originalPost.method(), location.toString())
+                                        .payloadBody(originalPost.payloadBody()));
+                            }
+                            // Decided not to follow redirect, return the original response or an error:
+                            return succeeded(response);
+                        })
+                        .afterFinally(responseProcessedLatch::countDown)
+                        .subscribe(resp -> {
+                            System.out.println(resp.toString((name, value) -> value));
```
bondolo, line 100: Is there any way to indicate whether redirect was taken in response or is it lost?

servicetalk-examples/http/redirects/src/main/java/io/servicetalk/examples/http/redirects/ManualRedirectClient.java
idelpivnitskiy: It's possible, but in the case of a manual redirect users have to take care of it. By playing around, users can write a method that will apply this `flatMap` operator, then re-use the method recursively from inside `flatMap`. It's also possible to pass a counter there, which will be used to limit the number of recursive calls. I will leave all these enhancements to the users' decision. Otherwise, the scope of this example may grow.

servicetalk-examples/http/redirects/src/main/java/io/servicetalk/examples/http/redirects/ManualRedirectClient.java
idelpivnitskiy: It prints all meta-data of the response, the status code will also be there. Users will see 200 instead of 301.
Because it's just an example, let's avoid adding testing functionality and keep it consistent with the hello-world example.

servicetalk-examples/docs/modules/ROOT/pages/http/index.adoc
```diff
@@ -172,6 +172,24 @@ Using the following classes:
 NOTE: This example uses the link:#blocking-aggregated[blocking + aggregated] API, as the TLS/SSL configuration API is
 the same across all the HTTP APIs.
 
+[#Redirects]
+== Redirects
+
+Extends the async "Hello World" example to demonstrate different ways how users can support redirects in ServiceTalk
```
tkountis, line 178: "ways how" -> "ways that"

tkountis (approved):

PR #1709, opened 2021-07-29
Load balancer unhealthy hosts

chemicL:
This PR is still WIP, the clear description is still pending. The basic idea is to remove unhealthy hosts from connection establishment in round robin load balancing for some time until a connection can be established again.

### DIFF
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -41,13 +41,13 @@ import io.servicetalk.concurrent.internal.ThrowableUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.time.Duration;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.concurrent.ThreadLocalRandom;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.function.Function;
@@ -89,6 +89,10 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
     static final String BACKGROUND_PROCESSING_EXECUTOR_NAME = "round-robin-load-balancer-executor";
     static final Executor SHARED_EXECUTOR = Executors.newFixedSizeExecutor(1,
             new DefaultThreadFactory(BACKGROUND_PROCESSING_EXECUTOR_NAME));
+    static final Duration DEFAULT_HEALTH_CHECK_INTERVAL = Duration.ofSeconds(1);
+    static final int DEFAULT_HEALTH_CHECK_FAILED_CONNECTIONS_THRESHOLD = 5; // higher than default for AutoRetryStrategy
+    static final HealthCheckConfig SHARED_HEALTH_CHECK_CONFIG = new HealthCheckConfig(SHARED_EXECUTOR,
+            DEFAULT_HEALTH_CHECK_INTERVAL, DEFAULT_HEALTH_CHECK_FAILED_CONNECTIONS_THRESHOLD);
 
     private static final Logger LOGGER = LoggerFactory.getLogger(RoundRobinLoadBalancer.class);
     private static final List<?> CLOSED_LIST = new ArrayList<>(0);
@@ -125,7 +129,6 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
     private final Publisher<Object> eventStream;
     private final SequentialCancellable discoveryCancellable = new SequentialCancellable();
     private final ConnectionFactory<ResolvedAddress, ? extends C> connectionFactory;
-    private final Executor healthCheckExecutor;
     private final ListenableAsyncCloseable asyncCloseable;
 
     /**
@@ -139,7 +142,7 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
     @Deprecated
     public RoundRobinLoadBalancer(final Publisher<? extends ServiceDiscovererEvent<ResolvedAddress>> eventPublisher,
                                   final ConnectionFactory<ResolvedAddress, ? extends C> connectionFactory) {
-        this(eventPublisher, connectionFactory, EAGER_CONNECTION_SHUTDOWN_ENABLED, SHARED_EXECUTOR);
+        this(eventPublisher, connectionFactory, EAGER_CONNECTION_SHUTDOWN_ENABLED, SHARED_HEALTH_CHECK_CONFIG);
     }
 
     /**
@@ -155,11 +158,10 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
     RoundRobinLoadBalancer(final Publisher<? extends ServiceDiscovererEvent<ResolvedAddress>> eventPublisher,
                            final ConnectionFactory<ResolvedAddress, ? extends C> connectionFactory,
                            final boolean eagerConnectionShutdown,
-                           final Executor healthCheckExecutor) {
+                           final HealthCheckConfig healthCheckConfig) {
         Processor<Object, Object> eventStreamProcessor = newPublisherProcessorDropHeadOnOverflow(32);
         this.eventStream = fromSource(eventStreamProcessor);
         this.connectionFactory = requireNonNull(connectionFactory);
-        this.healthCheckExecutor = healthCheckExecutor;
 
         toSource(eventPublisher).subscribe(new Subscriber<ServiceDiscovererEvent<ResolvedAddress>>() {
 
@@ -233,7 +235,7 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
             }
 
             private Host<ResolvedAddress, C> createHost(ResolvedAddress addr) {
-                Host<ResolvedAddress, C> host = new Host<>(addr);
+                Host<ResolvedAddress, C> host = new Host<>(addr, healthCheckConfig);
                 if (!eagerConnectionShutdown) {
                     host.onClose().afterFinally(() ->
                             usedHostsUpdater.updateAndGet(RoundRobinLoadBalancer.this, previousHosts -> {
@@ -380,8 +382,7 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
             }
 
             // Don't open new connections for expired or unhealthy hosts, try a different one.
-            // Unhealthy hosts can in fact have open connections – that's why we don't fail earlier.
-            // When a host accepts a limited amount of connections, we will try to use those that have been established.
+            // Unhealthy hosts have no open connections – that's why we don't fail earlier, the loop will not progress.
             if (host.isActiveAndHealthy()) {
                 pickedHost = host;
                 break;
@@ -398,7 +399,7 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
         // Users can apply a ConnectionFactoryFilter if they need to override this "null" value with TransportObserver.
         return connectionFactory.newConnection(host.address, null)
                 // Schedule health check before returning
-                .beforeOnError(t -> host.registerHealthCheck(healthCheckExecutor, connectionFactory, selector))
+                .beforeOnError(t -> host.markUnhealthy(connectionFactory, selector))
                 .flatMap(newCnx -> {
                     // Invoke the selector before adding the connection to the pool, otherwise, connection can be
                     // used concurrently and hence a new connection can be rejected by the selector.
@@ -449,13 +450,25 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
                 final Publisher<? extends ServiceDiscovererEvent<ResolvedAddress>> eventPublisher,
                 final ConnectionFactory<ResolvedAddress, T> connectionFactory) {
             return new RoundRobinLoadBalancer<>(eventPublisher, connectionFactory,
-                    EAGER_CONNECTION_SHUTDOWN_ENABLED, SHARED_EXECUTOR);
+                    EAGER_CONNECTION_SHUTDOWN_ENABLED, SHARED_HEALTH_CHECK_CONFIG);
         }
     }
 
     // Visible for testing
     List<Entry<ResolvedAddress, List<C>>> usedAddresses() {
         return usedHosts.stream().map(Host::asEntry).collect(toList());
+    }
+
+    static final class HealthCheckConfig {
+        private final Executor executor;
+        private final Duration healthCheckInterval;
+        private final int failedThreshold;
+
+        HealthCheckConfig(final Executor executor, final Duration healthCheckInterval, final int failedThreshold) {
+            this.executor = executor;
+            this.healthCheckInterval = healthCheckInterval;
+            this.failedThreshold = failedThreshold;
+        }
     }
 
     private static final class Host<Addr, C extends LoadBalancedConnection> implements ListenableAsyncCloseable {
@@ -470,25 +483,27 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
         private static final AtomicReferenceFieldUpdater<Host, ConnState> connStateUpdater =
                 AtomicReferenceFieldUpdater.newUpdater(Host.class, ConnState.class, "connState");
 
-        private static class HealthCheck<ResolvedAddress, C extends LoadBalancedConnection>
+        @SuppressWarnings("rawtypes")
+        private static final AtomicIntegerFieldUpdater<Host> connectionFailuresUpdater =
+                AtomicIntegerFieldUpdater.newUpdater(Host.class, "connectionFailures");
+
+        private static final class HealthCheck<ResolvedAddress, C extends LoadBalancedConnection>
                 implements Runnable, Cancellable {
-            private Executor executor;
-            private Predicate<C> selector;
-            private ConnectionFactory<ResolvedAddress, ? extends C> connectionFactory;
-            private Host<ResolvedAddress, C> host;
+            private final Predicate<C> selector;
+            private final ConnectionFactory<ResolvedAddress, ? extends C> connectionFactory;
+            private final Host<ResolvedAddress, C> host;
             private volatile boolean cancelled;
 
-            public HealthCheck(final Executor executor, final Predicate<C> selector,
+            private HealthCheck(final Predicate<C> selector,
                                final ConnectionFactory<ResolvedAddress, ? extends C> connectionFactory,
                                final Host<ResolvedAddress, C> host) {
-                this.executor = executor;
                 this.selector = selector;
                 this.connectionFactory = connectionFactory;
                 this.host = host;
             }
 
             public void schedule() {
-                executor.schedule(this, 1, TimeUnit.SECONDS);
+                host.healthCheckConfig.executor.schedule(this, host.healthCheckConfig.healthCheckInterval);
             }
 
             @Override
@@ -533,11 +548,14 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
         }
 
         final Addr address;
+        private final HealthCheckConfig healthCheckConfig;
         private final ListenableAsyncCloseable closeable;
         private volatile ConnState connState = EMPTY_CONN_STATE;
+        private volatile int connectionFailures;
 
-        Host(Addr address) {
+        Host(Addr address, HealthCheckConfig healthCheckConfig) {
             this.address = requireNonNull(address);
+            this.healthCheckConfig = requireNonNull(healthCheckConfig);
             this.closeable = toAsyncCloseable(graceful ->
                     graceful ? doClose(AsyncCloseable::closeAsyncGracefully) : doClose(AsyncCloseable::closeAsync));
         }
@@ -615,6 +633,8 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
                     break;
                 }
             }
+
+            connectionFailuresUpdater.set(this, 0);
 
             // Instrument the new connection so we prune it on close
             connection.onClose().beforeFinally(() -> {
@@ -696,20 +716,18 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
             });
         }
 
-        void registerHealthCheck(Executor executor, ConnectionFactory<Addr, ? extends C> connectionFactory, Predicate<C> selector) {
+        void markUnhealthy(ConnectionFactory<Addr, ? extends C> connectionFactory, Predicate<C> selector) {
+            connectionFailuresUpdater.getAndUpdate(this, val -> val == Integer.MAX_VALUE ? val : val + 1);
             for (;;) {
                 ConnState previous = connStateUpdater.get(this);
-                if (previous.state != STATE_ACTIVE) {
+                if (previous.state != STATE_ACTIVE
+                        || previous.connections.length > 0
+                        || connectionFailuresUpdater.get(this) < this.healthCheckConfig.failedThreshold) {
                     break;
                 }
 
-                // TODO(dj): if we mark the host as unhealthy for any connection failure, transient failures will
-                // potentially block clients from performing any requests for the health check interval.
-                // There needs to be more logic for determining when to take the host out of the pool, considering
-                // the actual pool and the state of other hosts.
-
                 final ConnState nextState = new ConnState(previous.connections,
-                        new HealthCheck<Addr, C>(executor, selector, connectionFactory, this));
+                        new HealthCheck<Addr, C>(selector, connectionFactory, this));
                 if (connStateUpdater.compareAndSet(this, previous, nextState)) {
                     @SuppressWarnings("unchecked")
                     final HealthCheck<Addr, C> state = (HealthCheck<Addr, C>) nextState.state;
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -49,7 +49,6 @@ import java.util.Map.Entry;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -257,9 +256,9 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
                 // duplicates are not allowed
                 for (Host<ResolvedAddress, C> host : oldHostsTyped) {
                     if (host.address.equals(addr)) {
-                        if (handleExpired && !host.tryToMarkActive()) {
-                            // If the new state is not ACTIVE, the host is already in CLOSED state, we should create
-                            // a new entry. For duplicate ACTIVE events or for repeated activation due to failed CAS
+                        if (handleExpired && !host.markActiveIfNotClosed()) {
+                            // If the host is already in CLOSED state, we should create a new entry.
+                            // For duplicate ACTIVE events or for repeated activation due to failed CAS
                             // of replacing the usedHosts array the marking succeeds so we will not add a new entry.
                             break;
                         }
@@ -460,25 +459,74 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
     }
 
     private static final class Host<Addr, C extends LoadBalancedConnection> implements ListenableAsyncCloseable {
-        private static final ConnState EMPTY_CONN_STATE = new ConnState(EMPTY_ARRAY, State.ACTIVE);
-        private static final ConnState CLOSED_CONN_STATE = new ConnState(EMPTY_ARRAY, State.CLOSED);
+        private static final Object STATE_ACTIVE = new Object();
+        private static final Object STATE_EXPIRED = new Object();
+        private static final Object STATE_CLOSED = new Object();
+
+        private static final ConnState EMPTY_CONN_STATE = new ConnState(EMPTY_ARRAY, STATE_ACTIVE);
+        private static final ConnState CLOSED_CONN_STATE = new ConnState(EMPTY_ARRAY, STATE_CLOSED);
 
         @SuppressWarnings("rawtypes")
         private static final AtomicReferenceFieldUpdater<Host, ConnState> connStateUpdater =
                 AtomicReferenceFieldUpdater.newUpdater(Host.class, ConnState.class, "connState");
 
-        private enum State {
-            ACTIVE,
-            UNHEALTHY, // only active hosts can be unhealthy, expired hosts should not attempt opening new connections
-            EXPIRED,
-            CLOSED
+        private static class HealthCheck<ResolvedAddress, C extends LoadBalancedConnection>
+                implements Runnable, Cancellable {
+            private Executor executor;
+            private Predicate<C> selector;
+            private ConnectionFactory<ResolvedAddress, ? extends C> connectionFactory;
+            private Host<ResolvedAddress, C> host;
+            private volatile boolean cancelled;
+
+            public HealthCheck(final Executor executor, final Predicate<C> selector,
+                               final ConnectionFactory<ResolvedAddress, ? extends C> connectionFactory,
+                               final Host<ResolvedAddress, C> host) {
+                this.executor = executor;
+                this.selector = selector;
+                this.connectionFactory = connectionFactory;
+                this.host = host;
+            }
+
+            public void schedule() {
+                executor.schedule(this, 1, TimeUnit.SECONDS);
+            }
+
+            @Override
+            public void run() {
+                if (cancelled) {
+                    return;
+                }
+                connectionFactory.newConnection(host.address, null)
+                        .flatMapCompletable(newCnx -> {
+                            if (!selector.test(newCnx)) {
+                                return newCnx.closeAsync().concat(Completable.failed(new RuntimeException()));
+                            }
+                            if (host.addConnection(newCnx)) {
+                                host.markHealthy();
+                            } else {
+                                return newCnx.closeAsync().concat(Completable.failed(new RuntimeException()));
+                            }
+                            return completed();
+                        })
+                        .afterOnError(e -> {
+                            if (!cancelled) {
+                                schedule();
+                            }
+                        })
+                        .subscribe();
+            }
+
+            @Override
+            public void cancel() {
+                this.cancelled = true;
+            }
         }
 
         private static final class ConnState {
             final Object[] connections;
-            final State state;
+            final Object state;
 
-            ConnState(final Object[] connections, final State state) {
+            ConnState(final Object[] connections, final Object state) {
                 this.connections = connections;
                 this.state = state;
             }
@@ -486,10 +534,7 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
 
         final Addr address;
         private final ListenableAsyncCloseable closeable;
-
         private volatile ConnState connState = EMPTY_CONN_STATE;
-
-        private final AtomicReference<HealthCheck> healthCheck = new AtomicReference<>();
 
         Host(Addr address) {
             this.address = requireNonNull(address);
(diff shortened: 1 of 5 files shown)


Scottmitch (commented):
took a quick look and left a few initial comments.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -441,54 +462,125 @@ public Completable closeAsyncGracefully() {
...
-                if (oldConnState.state == State.EXPIRED) {
-                    return new ConnState(oldConnState.connections, State.ACTIVE);
+        boolean markActiveIfNotClosed() {
+            final Object oldState = connStateUpdater.getAndUpdate(this, oldConnState -> {
+                if (oldConnState.state == STATE_EXPIRED) {
+                    return new ConnState(oldConnState.connections, STATE_ACTIVE);
                 }
                 // If oldConnState.state == State.ACTIVE this could mean either a duplicate event,
                 // or a repeated CAS operation. We could issue a warning, but as we don't know, we don't log anything.
+                // UNHEALTHY state is treated similarly to ACTIVE
```
idelpivnitskiy, line 574: This comment is a bit confusing. Consider saying: UNHEALTHY state cannot transition to ACTIVE without passing the health check.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -441,54 +462,125 @@ public Completable closeAsyncGracefully() {
...
+                        .flatMapCompletable(newCnx -> {
+                            if (!selector.test(newCnx)) {
+                                return newCnx.closeAsync().concat(Completable.failed(new RuntimeException()));
+                            }
+                            if (host.addConnection(newCnx)) {
+                                host.markHealthy();
+                            } else {
+                                return newCnx.closeAsync().concat(Completable.failed(new RuntimeException()));
+                            }
+                            return completed();
```
idelpivnitskiy, line 527: Consider aadding more debug logs when:
1. heath-check is sceduled
2. result of each health-check attempt
3. when health-check is cancelled

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -598,21 +712,67 @@ public Completable onClose() {
...
+        void markUnhealthy(ConnectionFactory<Addr, ? extends C> connectionFactory, Predicate<C> selector) {
+            if (healthCheckConfig == null) {
+                return;
+            }
+            connectionFailuresUpdater.getAndUpdate(this, val -> val == Integer.MAX_VALUE ? val : val + 1);
+            for (;;) {
+                ConnState previous = connStateUpdater.get(this);
+                if (previous.state != STATE_ACTIVE
+                        || previous.connections.length > 0
+                        || connectionFailuresUpdater.get(this) < this.healthCheckConfig.failedThreshold) {
```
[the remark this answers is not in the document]
chemicL: Thank you for the suggestion. I considered the race and decided an optimistic assumption that the race only causes lost increments, but still would trigger a health check. I understand it makes sense to avoid any other races and will iterate over this idea. However, you suggest the increments happen on unhealthy state, which previously meant that the background check is already scheduled. Now, the HealthCheck could mean ACTIVE, but with count. For clarity of state, I'd go for creating objects of ACTIVE type with a counter and include your suggestions. Does that seem fair?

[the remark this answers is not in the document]
Scottmitch: > Now, the HealthCheck could mean ACTIVE, but with count.

Yes I didn't take into account `failedThreshold` here so that makes sense. Seems like we actually only need the count on the `ACTIVE` state, as when we are unhealthy we don't need to count any more. We should also use an atomic inc/update/cas to compare against `failedThreshold` (instead of two independent operations).

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -498,17 +590,37 @@ void markClosed() {
...
-        boolean isActive() {
-            return connState.state == State.ACTIVE;
+        void markHealthy() {
+            Object oldState = connStateUpdater.getAndUpdate(this, previous -> {
+                if (previous.state.getClass().equals(HealthCheck.class)) {
+                    return new ConnState(previous.connections, STATE_ACTIVE);
+                }
+                return previous;
+            }).state;
+            cancelIfHealthCheck(oldState);
```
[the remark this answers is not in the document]
chemicL: Will add a comment explaining why it's necessary.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -95,13 +111,83 @@ public Builder() {
...
+        }
+
+        /**
+         * Configure an interval for health checking a host that failed to open connections. To disable the health
+         * checking logic, a negative value can be provided.
+         * @param intervalMillis interval (in milliseconds) at which a background health check will be scheduled.
+         * When the value is negative, the health checking mechanism is disabled.
+         * @return {@code this}.
+         */
+        public RoundRobinLoadBalancerFactory.Builder<ResolvedAddress, C> healthCheckIntervalMillis(int intervalMillis) {
```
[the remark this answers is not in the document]
chemicL: Makes sense. I guess it's a habit from the days where those would come from configuration files. I'll change the API here and make the disabling rely on the failed connections threshold, I think it makes more sense.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -95,13 +111,83 @@ public Builder() {
...
+         * to open a connection in more consecutive attempts than the specified value, the host will be marked as
+         * unhealthy and a connection establishment will take place in the background. Until finished, the host will
+         * not take part in load balancing selection.
+         * @param threshold number of consecutive connection failures to consider a host unhealthy and elligible for
+         * background health checking.
+         * @return {@code this}.
+         */
+        public RoundRobinLoadBalancerFactory.Builder<ResolvedAddress, C> healthCheckFailedConnectionsThreshold(
+                int threshold) {
+            this.healthCheckFailedConnectionsThreshold = threshold;
```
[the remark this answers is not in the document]
chemicL: Actually, after giving it a bit of thought, I'll treat this configuration as means to disable the functionality.
`< 0` means health checking disabled
`0` uses default threshold
`> 0` will be used to configure the threshold

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -441,54 +462,125 @@ public Completable closeAsyncGracefully() {
...
+            this.executor = executor;
+            this.healthCheckInterval = healthCheckInterval;
+            this.failedThreshold = failedThreshold;
+        }
+    }
+
+    private static final class Host<Addr, C extends LoadBalancedConnection> implements ListenableAsyncCloseable {
+        private static final Object STATE_ACTIVE = new Object();
+        private static final Object STATE_EXPIRED = new Object();
+        private static final Object STATE_CLOSED = new Object();
```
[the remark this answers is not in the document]
chemicL: As a matter of fact, with the latest suggestion to keep a counter in the active state object, this will not be a feasible option. `STATE_EXPIRED`, and `STATE_CLOSED` would be the only constants and e.g. `ActiveState` and `HealthCheck` will be new instances upon each update to the state.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -598,21 +712,67 @@ public Completable onClose() {
...
             });
         }
 
+        void markUnhealthy(ConnectionFactory<Addr, ? extends C> connectionFactory, Predicate<C> selector) {
+            if (healthCheckConfig == null) {
+                return;
+            }
+            connectionFailuresUpdater.getAndUpdate(this, val -> val == Integer.MAX_VALUE ? val : val + 1);
+            for (;;) {
+                ConnState previous = connStateUpdater.get(this);
```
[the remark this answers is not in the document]
chemicL: For consistency's sake I decided to always use the updater as the proxy to the field. There's no penalty at present: https://shipilev.net/blog/2015/faster-atomic-fu/

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
```diff
@@ -381,6 +400,177 @@ public void newConnectionIsClosedWhenSelectorRejects() throws Exception {
...
+
+        // From test main thread, wait until the host becomes UNHEALTHY, which is apparent from NoHostAvailableException
+        // being thrown from selection AFTER a health check was scheduled by any thread.
+        try {
+            awaitIndefinitely(lb.selectConnection(any()).retryWhen(retryWithConstantBackoffFullJitter((t) ->
+                    // DeliberateException comes from connection opening, check for that first
+                    // Next, NoAvailableHostException is thrown when the host is unhealthy,
+                    // but we still wait until the health check is scheduled and only then stop retrying.
+                    t instanceof DeliberateException || testExecutor.scheduledTasksPending() == 0,
+            // try to prevent stack overflow
```
chemicL, line 546: When I ran this 1000 times, 1 test failed due to stack overflow caused by the nature of resubscribing in the retry operator. I believe adding a small delay will allow the concurrent executor trigger the health check before an overflow happens.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -70,6 +91,10 @@ private RoundRobinLoadBalancerFactory(boolean eagerConnectionShutdown) {
      */
     public static final class Builder<ResolvedAddress, C extends LoadBalancedConnection> {
         private boolean eagerConnectionShutdown = EAGER_CONNECTION_SHUTDOWN_ENABLED;
+        @Nullable
+        private Executor backgroundExecutor;
+        private Duration healthCheckInterval = Duration.ZERO;
+        private int healthCheckFailedConnectionsThreshold;
```
idelpivnitskiy, line 97: Instead of checking at the `build` time, you can initialize all defaults right here. Validation of input params will prevent misconfiguration.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -95,13 +120,99 @@ public Builder() {
...
-            return new RoundRobinLoadBalancerFactory<>(eagerConnectionShutdown);
+            if (this.healthCheckFailedConnectionsThreshold < 0) {
+                return new RoundRobinLoadBalancerFactory<>(eagerConnectionShutdown, null);
+            }
+            assert !healthCheckInterval.isNegative();
+
+            if (this.backgroundExecutor == null && this.healthCheckInterval.isZero()
+                    && this.healthCheckFailedConnectionsThreshold == 0) {
+                return new RoundRobinLoadBalancerFactory<>(eagerConnectionShutdown,
+                        SharedHealthCheckConfig.getInstance());
```
idelpivnitskiy, line 181: I would say this `if` and `SharedHealthCheckConfig` are unnecessary. Just check the executor and always allocate a new `HealthCheckConfig`.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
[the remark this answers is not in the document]
idelpivnitskiy: heap dump shows class names, so `ActiveState` and `HealthCheck` will be readable, but `Object` won't. Can we have a enum for `expired` and `closed`?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -598,21 +691,121 @@ public Completable onClose() {
...
+            ActiveState() {
+                this(0);
+            }
+
+            private ActiveState(int failedConnections) {
+                this.failedConnections = failedConnections;
+            }
+
+            ActiveState forNextFailedConnection() {
+                return new ActiveState(FlowControlUtils.addWithOverflowProtection(this.failedConnections, 1));
```
idelpivnitskiy, line 747: Consider using static import

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -498,30 +526,95 @@ void markClosed() {
...
+                        break;
+                    }
+                    // another thread won the race, try again
+                    continue;
+                }
+
+                final HealthCheck<Addr, C> healthCheck = new HealthCheck<>(connectionFactory, this);
+                final ConnState nextState = new ConnState(previous.connections, healthCheck);
+                if (connStateUpdater.compareAndSet(this, previous, nextState)) {
+                    LOGGER.debug("Triggering health check for address {}", address);
```
idelpivnitskiy, line 591: Consider: Triggering health check for address {} after {} failed attempts to open a new connection

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -543,11 +636,11 @@ boolean addConnection(C connection) {
                     if (i == connections.length) {
                         break;
                     } else if (connections.length == 1) {
-                        if (currentConnState.state == State.ACTIVE) {
-                            if (connStateUpdater.compareAndSet(this, currentConnState, EMPTY_CONN_STATE)) {
+                        if (ActiveState.class.equals(currentConnState.state.getClass())) {
+                            if (connStateUpdater.compareAndSet(this, currentConnState, ACTIVE_EMPTY_CONN_STATE)) {
```
idelpivnitskiy, line 640: Why closure of a connection resets the failure counter?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -598,21 +691,121 @@ public Completable onClose() {
...
                     "address=" + address +
-                    ", state=" + connState.state +
+                    ", state=" + describeState(connState.state) +
                     ", #connections=" + connState.connections.length +
                     '}';
         }
+
+        private String describeState(Object state) {
+            if (ActiveState.class.equals(state.getClass())) {
+                return "active";
```
idelpivnitskiy, line 723: Will be useful to implement `toString()` for `ActiveState` to see the current failure counter. Then you can simplify this method to:
```java
if (HealthCheck.class.equals(state.getClass())) {
    return "unhealthy";
} else {
    state.toString();
}
```

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -598,21 +691,121 @@ public Completable onClose() {
...
             final ConnState connState = this.connState;
             return "Host{" +
                     "address=" + address +
-                    ", state=" + connState.state +
+                    ", state=" + describeState(connState.state) +
                     ", #connections=" + connState.connections.length +
                     '}';
         }
+
+        private String describeState(Object state) {
```
idelpivnitskiy, line 721: It can be `static`

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -598,21 +691,121 @@ public Completable onClose() {
...
+                                .concat(reconnect()
+                                        .retryWhen(retryWithConstantBackoffFullJitter(
+                                                cause -> cause == RESCHEDULE_SIGNAL,
+                                                host.healthCheckConfig.healthCheckInterval,
+                                                host.healthCheckConfig.executor)))
+                                .subscribe());
+            }
+
+            public Completable reconnect() {
+                return defer(() -> connectionFactory.newConnection(host.address, null))
```
idelpivnitskiy, line 781: Doesn't look like `defer` is required here

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -598,21 +691,121 @@ public Completable onClose() {
...
+                            return RESCHEDULE_SIGNAL;
+                        })
+                        .flatMapCompletable(newCnx -> {
+                            if (host.addConnection(newCnx)) {
+                                LOGGER.debug("Health check passed for address {}.", host.address);
+                                host.markHealthy();
+                            } else {
+                                LOGGER.debug("Health check failed for address {}. Host rejected connection.",
+                                        host.address);
+                                return newCnx.closeAsync().concat(Completable.failed(RESCHEDULE_SIGNAL));
```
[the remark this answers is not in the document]
chemicL: Makes sense.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -598,21 +691,121 @@ public Completable onClose() {
...
+                                                cause -> cause == RESCHEDULE_SIGNAL,
+                                                host.healthCheckConfig.healthCheckInterval,
+                                                host.healthCheckConfig.executor)))
+                                .subscribe());
+            }
+
+            public Completable reconnect() {
+                return defer(() -> connectionFactory.newConnection(host.address, null))
+                        .onErrorMap(cause -> {
+                            LOGGER.debug("Health check failed for address {}. Cause: {}", host.address, cause);
```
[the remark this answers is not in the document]
chemicL: True, forgot about that :)

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -95,13 +120,99 @@ public Builder() {
...
+        }
+
+        /**
+         * Configure an interval for health checking a host that failed to open connections.
+         * @param interval interval at which a background health check will be scheduled.
+         * @return {@code this}.
+         */
+        public RoundRobinLoadBalancerFactory.Builder<ResolvedAddress, C> healthCheckInterval(Duration interval) {
+            this.healthCheckInterval = requireNonNull(interval);
+            if (interval.isNegative()) {
```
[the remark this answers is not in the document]
chemicL: 1. Thanks for pointing it out. Regarding `requireNonNull`, I'm leaving it as a wrapper on top as `requireNonNull(interval).isNegative()`. As the package has the `ElementsAreNonnullByDefault` an explicit null check causes warnings in the IDE. But when the API is called from a language (e.g. Kotlin) that potentially ignores that annotation we still need to check for null. Am I right?
2. Zero is allowed, it uses the default. I wouldn't impose a minimum value. The API should be flexible here, it's difficult to come up with a minimum that will work in all cases. Assuming there's network traffic involved, the limiting factor will be the time it takes to repeat a connection attempt. If users want to set a value that's not `Duration.ZERO`, but too small for them, they should know what they're during. We provide a sensible default, anything different than that should be in users' hands.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -95,13 +120,99 @@ public Builder() {
...
+         * This {@link LoadBalancer} monitors hosts to which connection establishment has failed
+         * using health checks that run in the background. The health check tries to establish a new connection
+         * and if it succeeds, the host is returned to the load balancing pool. As long as the connection
+         * establishment fails, the host is not considered for opening new connections for processed requests.
+         *
+         * @param backgroundExecutor {@link Executor} on which to schedule health checking.
+         * @return {@code this}.
+         */
+        public RoundRobinLoadBalancerFactory.Builder<ResolvedAddress, C> backgroundExecutor(
+                Executor backgroundExecutor) {
```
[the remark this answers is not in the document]
chemicL: `retryWithConstantBackoffFullJitter` used in the `HealthCheck` requires `io.servicetalk.concurrent.api.Executor`. I don't see an easy way to overcome this limitation.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
chemicL: The idea her was to utilize lazy initialization as suggested by yourself in a previous comment. If I initialize the `backgroundExecutor` here, the whole point is missed. Were you thinking about something different though?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
chemicL: Sure.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: Good catch. This logic was an optimization. Now It wouldn't hurt anyone as the trigger for counting is when there's 0 connections. But in the future if we make the trigger point configurable that could be a hidden bug.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: As a matter of fact it is. It might feel it isn't and in most cases there won't be any issue. But the fact that the call `connectionFactory.newConnection(host.address, null)` can return a different `Single` every time it's called makes the re-subscribing mechanism prone to errors if that call is not re-exercised upon every retry. A practical example can be observed when you run the RRLB tests after removing the `defer()` call. But aside from the quirks of the test, I just believe it's semantically incorrect to assume the factory returns a proper re-subscribable `Single` that would behave the same way when a new one was obtained via calling `newConnection`.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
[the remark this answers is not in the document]
idelpivnitskiy: 1. If the `interval` is `null`, `interval.isNegative()` will throw NPE, which gives the same result as `requireNonNull(interval).isNegative()`.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
idelpivnitskiy: It makes sense to make `backgroundExecutor` initialization lazy, in this comment I was referring only to `healthCheckInterval` and `healthCheckFailedConnectionsThreshold`. Sorry for confusion

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
idelpivnitskiy: In Reactive Streams each re-subscribe should trigger new evaluation of the `Single` and produce the meaningful result. The result might be different (like a connection with new id), but it's still a result that should be treated as any previous result. In this case, your test is written incorrectly, without properly following RS spec:
```java
        Function<String, Single<TestLoadBalancedConnection>> factory =
                new Function<String, Single<TestLoadBalancedConnection>>() {

            @Override
            public Single<TestLoadBalancedConnection> apply(final String s) {
                if (s.equals(failingHost)) {
                    requests.incrementAndGet();
                    if (momentInTime.get() >= connections.size()) {
                        return properConnection;
                    }
                    return connections.get(momentInTime.get());
                }
                return properConnection;
            }
        };
```
Your `apply` method returns a `Single`, but it creates a state outside of the `Single`, which is not executed on each re-subscribe. Look at any other filter. For example, `RequestTargetEncoderHttpRequesterFilter`. It uses `defer` to recompute `requestTarget` on each re-subscribe. Otherwise, retry or repeat operator won't work as expected. You should do the same here and wrap the internals of your `apply` function with defer instead of using `defer` inside RRLB.

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
```diff
@@ -381,6 +400,177 @@ public void newConnectionIsClosedWhenSelectorRejects() throws Exception {
         awaitIndefinitely(connection.onClose());
     }
 
+    @Test
+    public void unhealthyHostTakenOutOfPoolForSelection() throws Exception {
+        serviceDiscoveryPublisher.onComplete();
```
idelpivnitskiy, line 405: Here and in other tests, you complete `serviceDiscoveryPublisher` but later invoke `onNext`. It violates the spec. Interesting, that `TestPublisher` doesn't validate such misuse. Please, open a GH issue for that.

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
```diff
@@ -381,6 +400,177 @@ public void newConnectionIsClosedWhenSelectorRejects() throws Exception {
...
+
+        sendServiceDiscoveryEvents(upEvent("address-1"));
+        sendServiceDiscoveryEvents(upEvent("address-2"));
+
+        for (int i = 0; i < DEFAULT_HEALTH_CHECK_FAILED_CONNECTIONS_THRESHOLD * 2; ++i) {
+            try {
+                final TestLoadBalancedConnection selectedConnection = lb.selectConnection(any()).toFuture().get();
+                assertThat(selectedConnection, equalTo(properConnection.toFuture().get()));
+            } catch (Exception e) {
+                assertThat(e.getCause(), instanceOf(DELIBERATE_EXCEPTION.getClass()));
```
idelpivnitskiy, line 423: Here and in other tests we can simplify to
```java
assertThat(e.getCause(), is(DELIBERATE_EXCEPTION));
```

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
```diff
@@ -381,6 +400,177 @@ public void newConnectionIsClosedWhenSelectorRejects() throws Exception {
...
+        // From test main thread, wait until the host becomes UNHEALTHY, which is apparent from NoHostAvailableException
+        // being thrown from selection AFTER a health check was scheduled by any thread.
+        try {
+            awaitIndefinitely(lb.selectConnection(any()).retryWhen(retryWithConstantBackoffFullJitter((t) ->
+                    // DeliberateException comes from connection opening, check for that first
+                    // Next, NoAvailableHostException is thrown when the host is unhealthy,
+                    // but we still wait until the health check is scheduled and only then stop retrying.
+                    t instanceof DeliberateException || testExecutor.scheduledTasksPending() == 0,
+            // try to prevent stack overflow
+            Duration.ofMillis(30), io.servicetalk.concurrent.api.Executors.newFixedSizeExecutor(1))));
```
idelpivnitskiy, line 547: We should properly shutdown all executors that we use in tests.

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
```diff
@@ -381,6 +400,177 @@ public void newConnectionIsClosedWhenSelectorRejects() throws Exception {
...
+            unhealthyHostConnectionFactory.advanceTime(testExecutor);
+
+            // Assert still unhealthy
+            Exception e = assertThrows(ExecutionException.class, () -> lb.selectConnection(any()).toFuture().get());
+            assertThat(e.getCause(), instanceOf(NoAvailableHostException.class));
+        }
+
+        // Shutdown the concurrent validation of unhealthiness.
+        executor.shutdownNow();
+        executor.awaitTermination(10, SECONDS);
```
idelpivnitskiy, line 566: If the test throws earlier, executor won't be shutdown

idelpivnitskiy (approved):
Last comments then LGTM. Good job!

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: That's a great explanation, thanks for taking the time to elaborate. That makes sense, I did as you recommended.

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
chemicL: Interesting, I actually copied the pattern from `selectStampedeUnsaturableConnection`, but I thought I'm re-creating the `serviceDiscoveryPublisher` in the method creating a new load balancer. But that might have gotten lost. We should improve this behaviour, you're right.

chemicL: https://github.com/apple/servicetalk/issues/1733

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
chemicL: Thanks, I missed that in the late addition.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -141,10 +152,13 @@ public RoundRobinLoadBalancer(final Publisher<? extends ServiceDiscovererEvent<R
      * set to {@code false} should be eagerly closed. When {@code false}, the expired addresses will be used
      * for sending requests, but new connections will not be requested, allowing the server to drive
      * the connection closure and shifting traffic to other addresses.
+     * @param healthCheckConfig configuration for the health checking mechanism, which monitors hosts that
```
bondolo, line 155: `healthCheckConfig` is nullable but it is not explained what happens when it is null or why you might not provide a health check config.

chemicL: Thanks for noticing, I added a note about it.

bondolo: What is the result of disabling it? Will the hosts never become eligible?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -95,13 +121,89 @@ public Builder() {
...
+         * <p>
+         * {@link #healthCheckFailedConnectionsThreshold(int)} can be used to disable the health checking mechanism
+         * and always consider all hosts for establishing new connections.
+         * </p>
+         * @param interval interval at which a background health check will be scheduled.
+         * @return {@code this}.
+         * @see #healthCheckFailedConnectionsThreshold(int)
+         */
+        public RoundRobinLoadBalancerFactory.Builder<ResolvedAddress, C> healthCheckInterval(Duration interval) {
+            if (requireNonNull(interval).isNegative()) {
```
bondolo, line 155: Should also disallow zero and perhaps nonsensical values (<1s or >1 day)

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
idelpivnitskiy: Thanks for the created issue! In your current tests, please adjust the behavior to be correct (i.e. don't terminate the publisher if you intend to use it later). You can skip `onComplete()` in your tests, it's not necessary for tests that are not focused on SD termination. In general, termination of SD events stream is not expected by RRLB.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -95,13 +121,89 @@ public Builder() {
...
+         * <p>
+         * {@link #healthCheckFailedConnectionsThreshold(int)} can be used to disable this mechanism and always
+         * consider all hosts for establishing new connections.
+         * </p>
+         *
+         * @param backgroundExecutor {@link Executor} on which to schedule health checking.
+         * @return {@code this}.
+         * @see #healthCheckFailedConnectionsThreshold(int)
+         */
+        public RoundRobinLoadBalancerFactory.Builder<ResolvedAddress, C> backgroundExecutor(
```
bondolo, line 138: The shared executor which is used if no executor is provided should be mentioned.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -95,13 +121,89 @@ public Builder() {
...
+            return this;
+        }
+
+        /**
+         * Configure an interval for health checking a host that failed to open connections.
+         * <p>
+         * {@link #healthCheckFailedConnectionsThreshold(int)} can be used to disable the health checking mechanism
+         * and always consider all hosts for establishing new connections.
+         * </p>
+         * @param interval interval at which a background health check will be scheduled.
```
bondolo, line 150: Should mention that providing an interval is optional and an uspecified default value will be used if not specified.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -95,13 +121,89 @@ public Builder() {
...
+         * Configure a threshold for consecutive connection failures to a host. When the {@link LoadBalancer} fails
+         * to open a connection in more consecutive attempts than the specified value, the host will be marked as
+         * unhealthy and a connection establishment will take place in the background. Until finished, the host will
+         * not take part in load balancing selection.
+         * Use a negative value of the argument to disable health checking.
+         * @param threshold number of consecutive connection failures to consider a host unhealthy and eligible for
+         * background health checking. Use negative value to disable the health checking mechanism.
+         * @return {@code this}.
+         * @see #backgroundExecutor(Executor)
+         * @see #healthCheckFailedConnectionsThreshold(int)
```
bondolo, line 172: Should point at #healthCheckInterval(Duration)

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -95,13 +121,89 @@ public Builder() {
...
+            return this;
+        }
+
+        /**
+         * Configure a threshold for consecutive connection failures to a host. When the {@link LoadBalancer} fails
+         * to open a connection in more consecutive attempts than the specified value, the host will be marked as
+         * unhealthy and a connection establishment will take place in the background. Until finished, the host will
+         * not take part in load balancing selection.
+         * Use a negative value of the argument to disable health checking.
+         * @param threshold number of consecutive connection failures to consider a host unhealthy and eligible for
```
bondolo, line 168: What does zero mean? Does 1 mean a single failure or one additional failure following a failure?

bondolo (approved):
Looks good to me, just javadoc clarifications

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
chemicL: I decided to point to the factory, which is a public API and contains a more thorough explanation.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
chemicL: I'm not sure about the bounds. I'd say it's up to the user to determine what is nonsensical – in some scenarios <1s might make sense. Not sure about >1 day, but then again, is 23 hrs good?
For zero, in my initial proposal it served the purpose of restoring the default. But I'm ok with preventing against zero too.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
chemicL: Thanks for this comment. Actually there were inconsistencies here. I adjusted the description (# consecutive failures greater or equal to provided value triggers health check).
I'm throwing an exception for 0 now. 1 means any failure triggers the health check. Hope it's clearer now.

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
chemicL: As a matter of fact this call is necessary for tests to behave correctly whenever a new instance of the load balancer is created. The reason for it is the fact that TestPublisher prevents simultaneous subscriptions by default. Calling `onComplete()` allows the new LB to reuse the existing `serviceDiscoveryPublisher`. The `onComplete()` method is not actually shutting down the `Publisher` but its' `Subscriber`:

```
 /**
     * Completes the {@link Subscriber}.
     *
     * @see Subscriber#onComplete()
     */
    public void onComplete()
```

I think the expected changes can be discussed in the created issue.

PR #1710, opened 2021-07-29
Add support for nonOffloading HTTP Server Filters

bondolo:
Motivation:
Some filters may wish to run early in the request process before the request has been offloaded.
Modifications:
A new API, `HttpServerBuilder.appendNonOffloadingServiceFilter(StreamingHttpServiceFilterFactory)` is provided for registering filters which will not be offloaded. Offloading will will be handled in the filter chain by the new 
`OffloadingFilter` which is logically inserted between the non-offloading and offloading filters.
Result:
New options for filters.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
@@ -310,16 +310,10 @@ class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
 
         final DefaultHttpExecutionStrategy that = (DefaultHttpExecutionStrategy) o;
 
-        if (offloads != that.offloads) {
-            return false;
-        }
-        if (threadAffinity != that.threadAffinity) {
-            return false;
-        }
-        if (!Objects.equals(executor, that.executor)) {
-            return false;
-        }
-        return mergeStrategy == that.mergeStrategy;
+        return offloads == that.offloads &&
+                threadAffinity == that.threadAffinity &&
+                Objects.equals(executor, that.executor) &&
+                mergeStrategy == that.mergeStrategy;
     }
 
     @Override
@@ -333,7 +327,7 @@ class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
 
     @Override
     public String toString() {
-        return "DefaultHttpExecutionStrategy{" +
+        return getClass().getSimpleName() + "{" +
                 "executor=" + executor +
                 ", offloads=" + offloads +
                 ", mergeStrategy=" + mergeStrategy +
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
@@ -254,11 +254,6 @@ final class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
     }
 
     // Visible for testing
-    boolean hasThreadAffinity() {
-        return false;
-    }
-
-    // Visible for testing
     boolean offloaded(byte flag) {
         return (offloads & flag) == flag;
     }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
@@ -34,7 +34,7 @@ import static java.util.function.Function.identity;
 /**
  * Default implementation for {@link HttpExecutionStrategy}.
  */
-class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
+final class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
 
     static final byte OFFLOAD_RECEIVE_META = 1;
     static final byte OFFLOAD_RECEIVE_DATA = 2;
@@ -43,14 +43,12 @@ class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
     private final Executor executor;
     private final byte offloads;
     private final MergeStrategy mergeStrategy;
-    private final boolean threadAffinity;
 
     DefaultHttpExecutionStrategy(@Nullable final Executor executor, final byte offloads,
                                  final MergeStrategy mergeStrategy) {
         this.mergeStrategy = mergeStrategy;
         this.executor = executor;
         this.offloads = offloads;
-        this.threadAffinity = false;
     }
 
     DefaultHttpExecutionStrategy(byte offloadOverride, HttpExecutionStrategy original) {
@@ -62,7 +60,6 @@ class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
         } else {
             mergeStrategy = Merge;
         }
-        this.threadAffinity = false;
     }
 
     @Nullable
@@ -99,7 +96,7 @@ class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
                                                         StreamingHttpRequest request,
                                                         final StreamingHttpResponseFactory responseFactory) {
                 // We compute the difference between the ExecutionStrategy from the current ExecutionContext and this
-                // ExecutionStrategy to understand if we need to offload more then we already offloaded:
+                // ExecutionStrategy to understand if we need to offload more than we already offloaded:
                 final HttpExecutionStrategy diff = difference(fallback, ctx.executionContext().executionStrategy(),
                         DefaultHttpExecutionStrategy.this);
                 // The service should see this ExecutionStrategy inside the ExecutionContext:
@@ -205,19 +202,13 @@ class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
         }
 
         final byte otherOffloads;
-        final boolean otherThreadAffinity;
         final MergeStrategy otherMergeStrategy;
         otherOffloads = generateOffloadsFlag(other);
-        otherThreadAffinity = extractThreadAffinity(otherExecutor);
         otherMergeStrategy = Merge; // always default to merge
 
-        return (otherOffloads == offloads && executor == otherExecutor && otherThreadAffinity == threadAffinity &&
+        return (otherOffloads == offloads && executor == otherExecutor &&
                 otherMergeStrategy == mergeStrategy) ? this :
                 new DefaultHttpExecutionStrategy(executor, (byte) (otherOffloads | offloads), otherMergeStrategy);
-    }
-
-    private static boolean extractThreadAffinity(@Nullable final Executor otherExecutor) {
-        return false;
     }
 
     private static byte generateOffloadsFlag(final HttpExecutionStrategy strategy) {
@@ -259,13 +250,12 @@ class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
 
     private Executor executor(final Executor fallback) {
         requireNonNull(fallback);
-        assert !threadAffinity : "Thread affinity not supported";
         return executor == null ? fallback : executor;
     }
 
     // Visible for testing
     boolean hasThreadAffinity() {
-        return threadAffinity;
+        return false;
     }
 
     // Visible for testing
@@ -278,14 +268,13 @@ class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
         if (this == o) {
             return true;
         }
-        if (o == null || getClass() != o.getClass()) {
+        if (!(o instanceof DefaultHttpExecutionStrategy)) {
             return false;
         }
 
         final DefaultHttpExecutionStrategy that = (DefaultHttpExecutionStrategy) o;
 
         return offloads == that.offloads &&
-                threadAffinity == that.threadAffinity &&
                 Objects.equals(executor, that.executor) &&
                 mergeStrategy == that.mergeStrategy;
     }
@@ -295,7 +284,6 @@ class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
         int result = executor != null ? executor.hashCode() : 0;
         result = 31 * result + offloads;
         result = 31 * result + mergeStrategy.hashCode();
-        result = 31 * result + (threadAffinity ? 1 : 0);
         return result;
     }
 
@@ -305,7 +293,6 @@ class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
                 "executor=" + executor +
                 ", offloads=" + offloads +
                 ", mergeStrategy=" + mergeStrategy +
-                ", threadAffinity=" + threadAffinity +
                 '}';
     }
 }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
@@ -22,7 +22,6 @@ import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.http.api.HttpExecutionStrategies.Builder.MergeStrategy;
 
 import java.util.Objects;
-import java.util.function.BiFunction;
 import java.util.function.Function;
 import javax.annotation.Nullable;
 
@@ -86,31 +85,6 @@ class DefaultHttpExecutionStrategy implements HttpExecutionStrategy {
         }
         if (offloaded(OFFLOAD_RECEIVE_DATA)) {
             resp = resp.map(response -> response.transformMessageBody(payload -> payload.publishOn(e)));
-        }
-        return resp;
-    }
-
-    @Override
-    public Publisher<Object> invokeService(
-            final Executor fallback, StreamingHttpRequest request,
-            final Function<StreamingHttpRequest, Publisher<Object>> service,
-            final BiFunction<Throwable, Executor, Publisher<Object>> errorHandler) {
-        final Executor e = executor(fallback);
-        if (offloaded(OFFLOAD_RECEIVE_DATA)) {
-            request = request.transformMessageBody(payload -> payload.publishOn(e));
-        }
-        Publisher<Object> resp;
-        if (offloaded(OFFLOAD_RECEIVE_META)) {
-            final StreamingHttpRequest r = request;
-            resp = e.submit(() -> service.apply(r).subscribeShareContext())
-                    .onErrorReturn(cause -> errorHandler.apply(cause, e))
-                    // exec.submit() returns a Single<Publisher<Object>>, so flatten the nested Publisher.
-                    .flatMapPublisher(identity());
-        } else {
-            resp = service.apply(request);
-        }
-        if (offloaded(OFFLOAD_SEND)) {
-            resp = resp.subscribeOn(e);
         }
         return resp;
     }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DelegatingHttpServiceContext.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DelegatingHttpServiceContext.java
@@ -40,7 +40,12 @@ public class DelegatingHttpServiceContext extends HttpServiceContext {
         this.delegate = other;
     }
 
-    public HttpServiceContext getDelegate() {
+    /**
+     * Returns the delegate {@link HttpServiceContext}.
+     *
+     * @return the delegate {@link HttpServiceContext}.
+     */
+    public HttpServiceContext delegate() {
         return delegate;
     }
 
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DelegatingHttpServiceContext.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DelegatingHttpServiceContext.java
@@ -40,6 +40,16 @@ public class DelegatingHttpServiceContext extends HttpServiceContext {
         this.delegate = other;
     }
 
+    // public for tests
+    public HttpServiceContext getDelegate() {
+        return delegate;
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "{delegate=" + delegate + "}";
+    }
+
     @Override
     public SocketAddress localAddress() {
         return delegate.localAddress();
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DelegatingHttpServiceContext.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DelegatingHttpServiceContext.java
@@ -40,7 +40,6 @@ public class DelegatingHttpServiceContext extends HttpServiceContext {
         this.delegate = other;
     }
 
-    // public for tests
     public HttpServiceContext getDelegate() {
         return delegate;
     }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ExecutionContextOverridingServiceContext.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ExecutionContextOverridingServiceContext.java
@@ -18,7 +18,7 @@ package io.servicetalk.http.api;
 import io.servicetalk.concurrent.api.Executor;
 
 final class ExecutionContextOverridingServiceContext extends DelegatingHttpServiceContext {
-    public final HttpExecutionContext context;
+    private final HttpExecutionContext context;
 
     ExecutionContextOverridingServiceContext(final HttpServiceContext ctx, final HttpExecutionStrategy strategy,
                                              final Executor executor) {
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ExecutionContextOverridingServiceContext.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ExecutionContextOverridingServiceContext.java
@@ -18,7 +18,7 @@ package io.servicetalk.http.api;
 import io.servicetalk.concurrent.api.Executor;
 
 final class ExecutionContextOverridingServiceContext extends DelegatingHttpServiceContext {
-    private final HttpExecutionContext context;
+    public final HttpExecutionContext context;
 
     ExecutionContextOverridingServiceContext(final HttpServiceContext ctx, final HttpExecutionStrategy strategy,
                                              final Executor executor) {
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategy.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategy.java
@@ -45,26 +45,6 @@ public interface HttpExecutionStrategy extends ExecutionStrategy {
                                                     @Nullable FS flushStrategy, ClientInvoker<FS> client);
 
     /**
-     * Invokes the passed {@link Function} and applies the necessary offloading of request and response for a server.
-     *
-     * @param fallback {@link Executor} to use as a fallback if this {@link HttpExecutionStrategy} does not define an
-     * {@link Executor}.
-     * @param request {@link StreamingHttpRequest} for which the offloading is to be applied.
-     * @param service A {@link Function} that executes a {@link StreamingHttpRequest} and returns a flattened
-     * {@link Publisher} containing all data constituting an HTTP response.
-     * @param errorHandler In case there is an error before calling the passed {@code service}, this {@link BiFunction}
-     * will be called to generate an error response.
-     * @return A flattened {@link Publisher} containing all data constituting an HTTP response.
-     *
-     * @deprecated Will be removed as part of improved offloading strategy. If you need this API please file an issue
-     * explaining the use case.
-     */
-    @Deprecated
-    Publisher<Object> invokeService(Executor fallback, StreamingHttpRequest request,
-                                    Function<StreamingHttpRequest, Publisher<Object>> service,
-                                    BiFunction<Throwable, Executor, Publisher<Object>> errorHandler);
-
-    /**
      * Invokes a service represented by the passed {@link Function}.
      * <p>
      * This method does not apply the strategy on the object returned from the {@link Function}, if that object is an
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
@@ -23,8 +23,6 @@ import static io.servicetalk.http.api.DefaultStreamingStrategyInfluencer.DEFAULT
 @FunctionalInterface
 public interface HttpExecutionStrategyInfluencer {
 
-    HttpExecutionStrategyInfluencer NO_INFLUENCE = strategy -> strategy;
-
     /**
      * Optionally modify the passed {@link HttpExecutionStrategy} to a new {@link HttpExecutionStrategy} that suits
      * this {@link HttpExecutionStrategyInfluencer}.
@@ -33,6 +31,15 @@ public interface HttpExecutionStrategyInfluencer {
      * @return {@link HttpExecutionStrategy} that suits this {@link HttpExecutionStrategyInfluencer}
      */
     HttpExecutionStrategy influenceStrategy(HttpExecutionStrategy strategy);
+
+    /**
+     * Returns the execution strategy required by this influencer.
+     *
+     * @return the execution strategy required by this influencer.
+     */
+    default HttpExecutionStrategy requiredStrategy() {
+        return influenceStrategy(HttpExecutionStrategies.noOffloadsStrategy());
+    }
 
     /**
      * Returns an {@link HttpExecutionStrategyInfluencer} to be used for the default streaming programming model.
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
@@ -23,6 +23,8 @@ import static io.servicetalk.http.api.DefaultStreamingStrategyInfluencer.DEFAULT
 @FunctionalInterface
 public interface HttpExecutionStrategyInfluencer {
 
+    HttpExecutionStrategyInfluencer NO_INFLUENCE = strategy -> strategy;
+
     /**
      * Optionally modify the passed {@link HttpExecutionStrategy} to a new {@link HttpExecutionStrategy} that suits
      * this {@link HttpExecutionStrategyInfluencer}.
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
@@ -33,15 +33,6 @@ public interface HttpExecutionStrategyInfluencer {
     HttpExecutionStrategy influenceStrategy(HttpExecutionStrategy strategy);
 
     /**
-     * Returns the execution strategy required by this influencer.
-     *
-     * @return the execution strategy required by this influencer.
-     */
-    default HttpExecutionStrategy requiredStrategy() {
-        return influenceStrategy(HttpExecutionStrategies.noOffloadsStrategy());
-    }
-
-    /**
      * Returns an {@link HttpExecutionStrategyInfluencer} to be used for the default streaming programming model.
      *
      * @return An {@link HttpExecutionStrategyInfluencer} to be used for the default streaming programming model.
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
@@ -16,7 +16,6 @@
 package io.servicetalk.http.api;
 
 import io.servicetalk.buffer.api.BufferAllocator;
-import io.servicetalk.concurrent.api.AsyncContext;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.http.api.HttpApiConversions.ServiceAdapterHolder;
 import io.servicetalk.logging.api.LogLevel;
@@ -30,6 +29,8 @@ import io.servicetalk.transport.api.TransportObserver;
 
 import java.net.SocketOption;
 import java.net.StandardSocketOptions;
+import java.util.ArrayDeque;
+import java.util.Deque;
 import java.util.Map;
 import java.util.function.BooleanSupplier;
 import java.util.function.Predicate;
@@ -37,6 +38,7 @@ import javax.annotation.Nullable;
 
 import static io.servicetalk.http.api.BlockingUtils.blockingInvocation;
 import static io.servicetalk.http.api.HttpApiConversions.toStreamingHttpService;
+import static io.servicetalk.http.api.HttpExecutionStrategies.OFFLOAD_NONE_STRATEGY;
 import static io.servicetalk.http.api.HttpExecutionStrategies.defaultStrategy;
 import static io.servicetalk.http.api.HttpExecutionStrategyInfluencer.defaultStreamingInfluencer;
 import static io.servicetalk.http.api.StrategyInfluencerAwareConversions.toConditionalServiceFilterFactory;
@@ -50,11 +52,17 @@ public abstract class HttpServerBuilder {
 
     @Nullable
     private ConnectionAcceptorFactory connectionAcceptorFactory;
-    @Nullable
-    private StreamingHttpServiceFilterFactory serviceFilter;
+    private final Deque<StreamingHttpServiceFilterFactory> noOffloadServiceFilters = new ArrayDeque<>();
+    private final Deque<StreamingHttpServiceFilterFactory> serviceFilters = new ArrayDeque<>();
     private HttpExecutionStrategy strategy = defaultStrategy();
-    private final StrategyInfluencerChainBuilder influencerChainBuilder = new StrategyInfluencerChainBuilder();
     private boolean drainRequestPayloadBody = true;
+
+    public HttpServerBuilder() {
+        // Async context clear goes before everything else.
+        StreamingHttpServiceFilterFactory asyncContextFilter = new ClearAsyncContextHttpServiceFilter();
+        assert OFFLOAD_NONE_STRATEGY == influenceStrategy(asyncContextFilter, OFFLOAD_NONE_STRATEGY);
+        noOffloadServiceFilters.add(asyncContextFilter);
+    }
 
     /**
      * Configurations of various HTTP protocol versions.
@@ -164,8 +172,10 @@ public abstract class HttpServerBuilder {
      * <p>
      * The order of execution of these filters are in order of append. If 3 filters are added as follows:
      * <pre>
-     *     builder.appendConnectionAcceptorFilter(filter1).appendConnectionAcceptorFilter(filter2).
-     *     appendConnectionAcceptorFilter(filter3)
+     *     builder
+     *          .appendConnectionAcceptorFilter(filter1)
+     *          .appendConnectionAcceptorFilter(filter2)
+     *          .appendConnectionAcceptorFilter(filter3)
      * </pre>
      * accepting a connection by a filter wrapped by this filter chain, the order of invocation of these filters will
      * be:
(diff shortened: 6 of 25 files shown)


bondolo:
One test is currently failing in `InsufficientlySizedExecutorHttpTest`. This failure indicates that more work is needed on error handling.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DelegatingHttpServiceContext.java
```diff
@@ -40,6 +40,16 @@ public DelegatingHttpServiceContext(final HttpServiceContext other) {
         this.delegate = other;
     }
 
+    // public for tests
```
Scottmitch, line 43: can we avoid adding public methods to public APIs just for tests?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ExecutionContextOverridingServiceContext.java
```diff
@@ -18,7 +18,7 @@
 import io.servicetalk.concurrent.api.Executor;
 
 final class ExecutionContextOverridingServiceContext extends DelegatingHttpServiceContext {
-    private final HttpExecutionContext context;
+    public final HttpExecutionContext context;
```
Scottmitch, line 21: reminder: revert

servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/InOrderRouter.java
```diff
@@ -67,7 +68,10 @@
                 StreamingHttpService service = pair.service();
                 final HttpExecutionStrategy strategy = pair.routeStrategy();
                 if (strategy != null) {
-                    service = strategy.offloadService(ctx.executionContext().executor(), service);
+                    Executor executor = null == strategy.executor() ?
```
Scottmitch, line 71: `strategy.executor()` in theory may change in-between checking against null and then saving the reference. cache the reference in a local variable to avoid this.

Scottmitch: also isn't the plan to remove `executor()` from the strategy? why do we need to change this code to use the `strategy`'s executor?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public StreamingHttpServiceFilter create(StreamingHttpService service) {
+        return new StreamingHttpServiceFilter(offloaded.create(service)) {
+
+            @Override
+            public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
+                                                        StreamingHttpRequest request,
+                                                        final StreamingHttpResponseFactory responseFactory) {
+                Executor e = null != strategy.executor() ? strategy.executor() : ctx.executionContext().executor();
```
Scottmitch, line 48: `strategy.executor()` in theory may change in-between checking against null and then saving the reference. cache the reference in a local variable to avoid this.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
```diff
@@ -0,0 +1,74 @@
...
+
+                // The service should see our ExecutionStrategy inside the ExecutionContext:
+                final HttpServiceContext wrappedCtx =
```
Scottmitch, line 51: if `ExecutionStrategy` interface is reduced to just the state (e.g. boolean methods), will we need `ExecutionContext executionStrategy()` on the `ExecutionContext` interface (and by extension `HttpExecutionContext`)? perhaps this is still necessary for routers to see which strategy has been used thus far and do a diff?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
```diff
@@ -0,0 +1,74 @@
...
+                        new ExecutionContextOverridingServiceContext(ctx, strategy, e);
+
+                if (strategy.isDataReceiveOffloaded()) {
```
Scottmitch, line 54: reminder: remove `invokeService` API from `HttpExecutionStrategy` now that the code has been moved to here.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DelegatingHttpServiceContext.java
bondolo: One of the Netty tests was assuming that the context would be a particular class which instead now gets wrapped by `ExecutionContextOverridingServiceContext`. The easiest resolution was to allow the test to accept either the class it expected or `ExecutionContextOverridingServiceContext` and extract the Netty context from the delegating. Due to module separation the `getDelegate()` method can't be package private.

servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/InOrderRouter.java
bondolo: It was changed for consistency. Yes, it will soon be obsolete.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
bondolo: invokeService is still used by Netty in the path where there is no offloading filter installed. I am not sure it is safe to remove.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DelegatingHttpServiceContext.java
Scottmitch: can we adjust the test and avoid adding new APIs?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
Scottmitch: the goal of this PR IIUC is to move the offloading responsibility from `NettyHttpServer` into a filter, so we can control the ordering in which offloading is applied. Why do we need `HttpExecutionStrategy#invokeService`?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
bondolo: Yes, the diff is why it is still needed. I believe there are cases where this will be needed.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
bondolo: Not all the code paths which use it have been removed yet.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DelegatingHttpServiceContext.java
bondolo: Probably. This was expedient for a draft PR.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ExecutionContextOverridingServiceContext.java
bondolo: done

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
Scottmitch: IIUC the plan is todo this in this PR? maybe I'll just wait until the PR is out of draft before reviewing 😄

bondolo: Review is always welcome at any point in the process.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DelegatingHttpServiceContext.java
bondolo: I didn't find a way to easily fix the `NettyHttpServer` to not require the addition of an accessor to `DelegatingHttpServiceContext`.  It could access the field via reflection if we really did not want to add the `getDelegate()` method.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
bondolo: Currently there are no strategies with non-final executors and with the soon-to-follow #1695 it will be moot.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -412,20 +445,74 @@ public final ServerContext listenBlockingStreamingAndAwait(
     private Single<ServerContext> listenForService(StreamingHttpService rawService, HttpExecutionStrategy strategy) {
         ConnectionAcceptor connectionAcceptor = connectionAcceptorFactory == null ? null :
                 connectionAcceptorFactory.create(ACCEPT_ALL);
-        StreamingHttpServiceFilterFactory currServiceFilter = serviceFilter;
-        if (!AsyncContext.isDisabled()) {
-            StreamingHttpServiceFilterFactory asyncContextFilter = new AsyncContextAwareHttpServiceFilter();
-            currServiceFilter = currServiceFilter == null ? asyncContextFilter :
-                    appendFilter(asyncContextFilter, currServiceFilter);
+
+        Deque<StreamingHttpServiceFilterFactory> nonOffloadingFilters = new ArrayDeque<>(noOffloadServiceFilters);
```
idelpivnitskiy, line 449: I would say this copying is not necessary. We never expect concurrency at the build time.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -412,20 +445,74 @@ public final ServerContext listenBlockingStreamingAndAwait(
...
+                // We are going to have to offload, even if just to the raw service
+                nonOffloadingFilters.add(new OffloadingFilter(strategy,
+                        buildFactory(serviceFilters)));
+                strategy = null != strategy.executor() ?
+                        HttpExecutionStrategies.customStrategyBuilder()
+                                .offloadNone().executor(strategy.executor()).build() :
+                        HttpExecutionStrategies.noOffloadsStrategy();
+            } else {
+                // All the filters can be appended.
+                nonOffloadingFilters.addAll(serviceFilters);
```
idelpivnitskiy, line 470: My stream-fu isn't great, so asking: will `Stream.concat(nonOffloadingFilters.stream, serviceFilters.stream())` do better than `addAll` here?

idelpivnitskiy (commented):
Great progress!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
Scottmitch: IIUC after `Executor` is removed wrapping is only necessary if `!ctx.executionContext().executionStrategy().equals(strategy)`? I expect this `equals` comparison should be cheap after this point (coming like `int` bit mask comparison).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -260,9 +258,8 @@ public String toString() {
             this.headersFactory = headersFactory;
             executionContext = new DefaultHttpExecutionContext(connection.executionContext().bufferAllocator(),
                     connection.executionContext().ioExecutor(), connection.executionContext().executor(),
-                    strategy);
+                    HttpExecutionStrategies.noOffloadsStrategy());
```
Scottmitch, line 261: this looks suspicious. is there a chance we may return an unexpected value for `executionStrategy ()`? do we need `HttpExecutionStrategy executionStrategy()` on `HttpExecutionContext`?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ServerEffectiveStrategyTest.java
```diff
@@ -218,7 +218,7 @@ void blocking(final ServerStrategyCase strategyCase) throws Exception {
         BlockingHttpClient client = params.startBlocking();
         client.request(client.get("/")
                 .payloadBody(client.executionContext().bufferAllocator().fromAscii("Hello")));
-        params.verifyOffloads(ServiceType.Blocking);
+        //params.verifyOffloads(ServiceType.Blocking);
```
Scottmitch, line 221: restore/fix/remove assertions?

servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/InOrderRouter.java
Scottmitch, line 71: on the subject of removing `Executor` from `HttpExecutionStrategy` ... looks like the routers are leveraging this to pass the `Executor` + strategy maybe overriden on the route level (e.g. `RouteExecutionStrategy`, `RouteExecutionStrategyUtils`, ..., `JerseyRouteExecutionStrategyUtils#getRouteExecutionStrategy`, `EndpointEnhancingRequestFilter#defineEndpoint`). Does this impact plans to simplify the `HttpExecutionStrategy` interface? I'm not even sure we need to support specifying different executors on the route level as we apply capacity management independent of the Executor.

servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/InOrderRouter.java
```diff
@@ -67,7 +68,10 @@
...
+                            ctx.executionContext().executor() :
+                            strategy.executor();
+                    service = strategy.offloadService(executor, service);
```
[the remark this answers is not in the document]
Scottmitch: fyi duplicate of https://github.com/apple/servicetalk/pull/1710#discussion_r679983842 and https://github.com/apple/servicetalk/pull/1710#discussion_r679986201

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
```diff
@@ -0,0 +1,79 @@
...
+                            // exec.submit() returns a Single<Single<response>>, so flatten the nested Single.
+                            .flatMap(identity());
+                } else {
+                    resp = delegate().handle(wrappedCtx, request, responseFactory);
+                }
+                return strategy.isSendOffloaded() ?
+                        // This is different from invokeService() where we just offload once on the  flattened
+                        // (meta + data) stream. In this case, we need to preserve the service contract and hence
+                        // have to offload both meta and data separately.
+                        resp.map(r -> r.transformMessageBody(p -> p.subscribeOn(e))).subscribeOn(e) : resp;
```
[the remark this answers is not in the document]
Scottmitch: `invokeService` logic was moved to this filter. `offloadService` will eventually be moved to utility method and targeted at routers usage (which includes difference, and potentially wrapping of service + execution context, etc.). good to re-evaluate after the full scope comes into picture.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -50,12 +52,18 @@
...
+    private final Deque<StreamingHttpServiceFilterFactory> serviceFilters = new ArrayDeque<>();
     private HttpExecutionStrategy strategy = defaultStrategy();
-    private final StrategyInfluencerChainBuilder influencerChainBuilder = new StrategyInfluencerChainBuilder();
     private boolean drainRequestPayloadBody = true;
 
+    public HttpServerBuilder() {
+        // Async context clear goes before everything else.
+        StreamingHttpServiceFilterFactory asyncContextFilter = new ClearAsyncContextHttpServiceFilter();
+        assert OFFLOAD_NONE_STRATEGY == influenceStrategy(asyncContextFilter, OFFLOAD_NONE_STRATEGY);
+        noOffloadServiceFilters.add(asyncContextFilter);
```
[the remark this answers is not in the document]
idelpivnitskiy: Users can enable/disable `AsyncContext` at runtime. My 2c: it's still worth adding a filter unconditionally. Its overhead is tiny.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -412,20 +445,74 @@ public final ServerContext listenBlockingStreamingAndAwait(
...
+                        buildFactory(serviceFilters)));
+                strategy = null != strategy.executor() ?
+                        HttpExecutionStrategies.customStrategyBuilder()
+                                .offloadNone().executor(strategy.executor()).build() :
+                        HttpExecutionStrategies.noOffloadsStrategy();
+            } else {
+                // All the filters can be appended.
+                nonOffloadingFilters.addAll(serviceFilters);
+            }
+            filteredService = buildService(nonOffloadingFilters, rawService);
```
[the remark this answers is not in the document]
idelpivnitskiy: Agreed with the comment. IIUC, current PR preserves existing behavior. So, I did not comment on that. We can re-consider what objects are checked for `HttpExecutionStrategyInfluencer` in a follow-up when we re-work `HttpExecutionStrategyInfluencer`.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -50,12 +52,18 @@
...
-    private StreamingHttpServiceFilterFactory serviceFilter;
+    private final Deque<StreamingHttpServiceFilterFactory> noOffloadServiceFilters = new ArrayDeque<>();
+    private final Deque<StreamingHttpServiceFilterFactory> serviceFilters = new ArrayDeque<>();
     private HttpExecutionStrategy strategy = defaultStrategy();
-    private final StrategyInfluencerChainBuilder influencerChainBuilder = new StrategyInfluencerChainBuilder();
     private boolean drainRequestPayloadBody = true;
 
+    public HttpServerBuilder() {
+        // Async context clear goes before everything else.
+        StreamingHttpServiceFilterFactory asyncContextFilter = new ClearAsyncContextHttpServiceFilter();
```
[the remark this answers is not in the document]
Scottmitch: fyi duplicate of https://github.com/apple/servicetalk/pull/1710#discussion_r682297571

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
bondolo: Correct. The comparison between execution strategies is already fairly cheap anyway due to short-circuiting.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
```diff
@@ -310,16 +284,10 @@ public boolean equals(final Object o) {
...
-        }
-        if (threadAffinity != that.threadAffinity) {
-            return false;
-        }
-        if (!Objects.equals(executor, that.executor)) {
-            return false;
-        }
-        return mergeStrategy == that.mergeStrategy;
+        return offloads == that.offloads &&
+                threadAffinity == that.threadAffinity &&
```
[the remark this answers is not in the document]
bondolo: Yes. Leaving this unresolved until deprecation is merged in `main`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
```diff
@@ -23,6 +23,8 @@
 @FunctionalInterface
 public interface HttpExecutionStrategyInfluencer {
 
+    HttpExecutionStrategyInfluencer NO_INFLUENCE = strategy -> strategy;
```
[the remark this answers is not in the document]
bondolo: This was moved here and then the usage was removed. Removing.

servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/InOrderRouter.java
bondolo: TBH, I am struggling with exactly that issue in #1695 at the moment. Some of what the routers expect can be handled by using the router executor as the context executor.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
```diff
@@ -0,0 +1,79 @@
...
+import org.slf4j.LoggerFactory;
+
+import static java.util.function.Function.identity;
+
+/**
+ * An {@link StreamingHttpServiceFilterFactory} implementation which offloads filters using a provided strategy.
+ */
+final class OffloadingFilter implements StreamingHttpServiceFilterFactory {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(OffloadingFilter.class);
```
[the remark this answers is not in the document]
bondolo: removed

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -412,20 +445,74 @@ public final ServerContext listenBlockingStreamingAndAwait(
...
+
+    private static StreamingHttpService buildService(Deque<StreamingHttpServiceFilterFactory> filters,
+                                                     StreamingHttpService service) {
+        return filters.stream()
+                .reduce((prev, filter) -> svc -> prev.create(filter.create(svc)))
+                .map(factory -> (StreamingHttpService) factory.create(service))
+                .orElse(service);
+    }
+
+    private static HttpExecutionStrategyInfluencer buildInfluencer(Deque<StreamingHttpServiceFilterFactory> filters,
```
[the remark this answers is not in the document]
bondolo: Originally it was replaced for the optimization of moving filters from offloading to non-offloading and since the data structure was redundant with the filter list  Beyond that it was redundant with the filter list. At the moment the removal is an unnecessary simplification.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
bondolo: OK.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
bondolo: Yes, now converted to Stream.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
```diff
@@ -310,16 +284,10 @@ public boolean equals(final Object o) {
...
-            return false;
-        }
-        if (threadAffinity != that.threadAffinity) {
-            return false;
-        }
-        if (!Objects.equals(executor, that.executor)) {
-            return false;
-        }
-        return mergeStrategy == that.mergeStrategy;
+        return offloads == that.offloads &&
```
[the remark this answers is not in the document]
bondolo: My preference is for exact types where there may be significant behavior differences between implementations and/or a different equals definition in other implementations. This would seem to be likely or possible with `HttpExecutionStrategy` implementations so I'd prefer to keep the current narrow definition. With #1695 it won't be possible as we won't control `equals()` and enum equals is always identity equals.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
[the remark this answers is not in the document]
bondolo: I agree about adding it unconditionally. I was tempted to have the `create()` method do the conditional check but the `NoopAsyncContext.clear()` is cheap.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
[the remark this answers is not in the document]
bondolo: originally it was for safety and because we added filters to it, but now removed.

servicetalk-http-router-predicate/src/main/java/io/servicetalk/http/router/predicate/InOrderRouter.java
Scottmitch: discussed offline ... plan is to continue with `HttpExecutionStrategy#executor()` removal and focus on routers specifying strategy portions first (required to be "safe"), and we can later add support to specify an `executor` at the route level if required.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -205,12 +215,38 @@ public final HttpServerBuilder appendConnectionAcceptorFilter(final ConnectionAc
      * @param factory {@link StreamingHttpServiceFilterFactory} to append.
      * @return {@code this}
      */
-    public final HttpServerBuilder appendServiceFilter(final StreamingHttpServiceFilterFactory factory) {
+    public final HttpServerBuilder appendNonOffloadingServiceFilter(final StreamingHttpServiceFilterFactory factory) {
         requireNonNull(factory);
-        serviceFilter = appendFilter(serviceFilter, factory);
-        if (!influencerChainBuilder.appendIfInfluencer(factory)) {
-            influencerChainBuilder.append(defaultStreamingInfluencer());
+        HttpExecutionStrategy factoryStrategy = influenceStrategy(factory, OFFLOAD_NONE_STRATEGY);
```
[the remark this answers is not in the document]
bondolo: I agree that requiring the user to implement `HttpExecutionStrategyInfluencer` might just cause them to write a GIGO implementation without being honest. I added the checks in append to help avoid honest mistakes, but it is not possible to prevent intentional misuse.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
bondolo: The strategy for the server context is always now noOffloads since offloading is handled by the OffloadingFilter. It does seem that it could possibly be removed.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ServerEffectiveStrategyTest.java
bondolo: The `InstrumentedExecutionStrategy` used to record the `invokeService` call and can no longer do so. I have disabled the verification.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategy.java
```diff
@@ -44,22 +44,6 @@
...
-     * @param fallback {@link Executor} to use as a fallback if this {@link HttpExecutionStrategy} does not define an
-     * {@link Executor}.
-     * @param request {@link StreamingHttpRequest} for which the offloading is to be applied.
-     * @param service A {@link Function} that executes a {@link StreamingHttpRequest} and returns a flattened
-     * {@link Publisher} containing all data constituting an HTTP response.
-     * @param errorHandler In case there is an error before calling the passed {@code service}, this {@link BiFunction}
-     * will be called to generate an error response.
-     * @return A flattened {@link Publisher} containing all data constituting an HTTP response..
-     */
-    Publisher<Object> invokeService(Executor fallback, StreamingHttpRequest request,
```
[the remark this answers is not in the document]
bondolo: Deprecation PR now posted #1720

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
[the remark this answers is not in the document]
bondolo: Specifying either on the builder via rawService being an `HttpExecutionStrategyInfluence` is OK. I somewhat prefer the later, that each element specify the strategy they require. If the strategy is set on the builder it is less clear who needs that strategy.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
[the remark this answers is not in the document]
idelpivnitskiy: Ok, I don't worry much about custom impls, just want to make sure that users won't see the difference in behavior if they pass `noOffloadsStrategy()` instead of `customStrategyBuilder().offloadNone().build()`. If that will be improved in #1695, I'm fine with that.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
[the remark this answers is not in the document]
idelpivnitskiy: Agreed, users can incorrectly implement the influencer method. Or influencer may not be updated after the filter code changes. However, it's a bigger problem that will also impact `appendServiceFilter` with API conversions. 
For now, we can trust that users implemented the contract correctly and require filters to implement influencer for `appendNonOffloadingServiceFilter`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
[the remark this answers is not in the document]
idelpivnitskiy: If there is no change in behavior, would be great to have less duplicated logic between `buildInfluencer` and `StrategyInfluencerChainBuilder`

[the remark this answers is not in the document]
Scottmitch: Agreed with @idelpivnitskiy. We now have `StrategyInfluencerChainBuilder` and this `ArrayDeque` approach. Can you confirm the end sate will involve a single way this is accomplished?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
Scottmitch: we should have a plan for offloading methods that may call back user code on the IOThread (e.g. `closeAsync()`, `closeAsyncGracefully()`, `onClose()`) for the `HttpServiceContext` passed to the user's service (and filters). It maybe nice to defer this until the offloading filters bcz we have some internal use cases where we don't want offloading to be applied (like `ChannelSet` that we disabled offloading for).

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/InsufficientlySizedExecutorHttpTest.java
```diff
@@ -69,21 +66,18 @@
...
-            // Arguments.of(1, true),
-            Arguments.of(1, false)
+            Arguments.of(0),
+            Arguments.of(1)
         );
     }
 
-    @ParameterizedTest(name = "{displayName} {index} - capacity: {0} thread based: {1}")
+    @ParameterizedTest(name = "{displayName} {index} - capacity: {0}")
     @MethodSource("executors")
```
Scottmitch, line 75: `@ValueSource(ints = {0, 1})`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/InsufficientlySizedExecutorHttpTest.java
```diff
@@ -97,55 +91,43 @@ void insufficientClientCapacityStreaming(final int capacity, final boolean threa
         }
     }
 
-    @ParameterizedTest(name = "{displayName} {index} - capacity: {0} thread based: {1}")
+    @ParameterizedTest(name = "{displayName} {index} - capacity: {0}")
     @MethodSource("executors")
```
Scottmitch, line 95: `@ValueSource(ints = {0, 1})`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/InsufficientlySizedExecutorHttpTest.java
```diff
@@ -97,55 +91,43 @@ void insufficientClientCapacityStreaming(final int capacity, final boolean threa
...
-        insufficientServerCapacityStreaming0(capacity, threadBased);
+    void insufficientServerCapacityStreaming(final int capacity) throws Exception {
+        initWhenServerUnderProvisioned(capacity, false);
+        insufficientServerCapacityStreaming0(capacity);
     }
 
     @Disabled("https://github.com/apple/servicetalk/issues/336")
-    @ParameterizedTest(name = "{displayName} {index} - capacity: {0} thread based: {1}")
+    @ParameterizedTest(name = "{displayName} {index} - capacity: {0}")
     @MethodSource("executors")
```
Scottmitch, line 103: `@ValueSource(ints = {0, 1})`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ServerEffectiveStrategyTest.java
```diff
@@ -340,22 +340,22 @@ void initStateHolderUserStrategyNoExecutor() {
...
 
         void initStateHolderUserStrategyNoOffloadsNoExecutor() {
-            verifyStrategyUsed = !addFilter;
+            verifyStrategyUsed = false;
             newRecorder(noOffloadsStrategy());
         }
 
         void initStateHolderCustomUserStrategyNoExecutor() {
-            verifyStrategyUsed = !addFilter;
+            verifyStrategyUsed = false;
```
Scottmitch, line 358: looks like `verifyStrategyUsed` is always false, can it be removed?

Scottmitch (changes requested):

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -263,14 +338,14 @@ public final HttpServerBuilder appendServiceFilter(final Predicate<StreamingHttp
      * @return {@code this}.
      */
     public final HttpServerBuilder executionStrategy(HttpExecutionStrategy strategy) {
-        this.strategy = strategy;
+        this.strategy = requireNonNull(strategy, "strategy");
```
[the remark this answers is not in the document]
bondolo: Use of "strategy" is basically free, just an extra arg, as the string is already in the constant pool for the parameter name.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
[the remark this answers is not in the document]
bondolo: `StrategyInfluencerChainBuilder` is also used by the client builder and won't be refactored in the current PR. My preference is still for the approach now used in `HttpServerBuilder` as the implementation is simpler and the "rules" are better consolidated to two methods vs divided between `HttpServerBuilder` and `StrategyInfluencerChainBuilder` implementation. The "rules" are:
1. If a service filter implements `HttpStrategyInfluencer` apply the influence in result otherwise apply `defaultStreamingInfluencer()` influence
2. The service is prepended as an influencer if it implements `HttpStrategyInfluencer` otherwise it has no influence
3. The influencer used for non-async streaming adaptor is the combination of the filters' StrategyInfluence (1) and the service's influence (2) applied to the builder strategy. 
4. The builder influencer (3) is used to influence the strategy used by the adaptor using `HttpExecutionStrategy.merge()`

The `StrategyInfluencerChainBuilder` spreads these rules out across `HttpServerBuilder` append methods, listen methods the final `build` method which makes it hard to be sure the behavior is even consistent. The implementation could be simplified a bit, calculating the influence directly rather than building the influencer and then calculating but that didn't seem worth the extra time.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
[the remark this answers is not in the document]
bondolo: I opted for the former because that choice anticipates the #1695 behavior.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/InsufficientlySizedExecutorHttpTest.java
```diff
@@ -155,10 +137,9 @@ private void initWhenServerUnderProvisioned(final int capacity,
         client = forSingleAddress(serverHostAndPort(server)).buildStreaming();
     }
 
-    private HttpExecutionStrategy newStrategy(final boolean threadBased) {
+    private HttpExecutionStrategy newStrategy() {
         final HttpExecutionStrategies.Builder strategyBuilder = customStrategyBuilder().offloadAll().executor(executor);
```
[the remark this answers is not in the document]
bondolo: executor is still used, for now.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -50,12 +52,18 @@
 
     @Nullable
     private ConnectionAcceptorFactory connectionAcceptorFactory;
-    @Nullable
-    private StreamingHttpServiceFilterFactory serviceFilter;
+    private final Deque<StreamingHttpServiceFilterFactory> noOffloadServiceFilters = new ArrayDeque<>();
+    private final Deque<StreamingHttpServiceFilterFactory> serviceFilters = new ArrayDeque<>();
```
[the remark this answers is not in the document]
bondolo: Originally I used `ArrayDeque` because there were additions at both ends which it handles better than `ArrayList`. If you don't need `List` interface and will be adding to both ends then `ArrayDeque` is more efficient.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
bondolo: Acknowledged.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
```diff
@@ -285,13 +250,12 @@ private static byte generateOffloadsFlag(final HttpExecutionStrategy strategy) {
 
     private Executor executor(final Executor fallback) {
         requireNonNull(fallback);
-        assert !threadAffinity : "Thread affinity not supported";
         return executor == null ? fallback : executor;
     }
 
     // Visible for testing
     boolean hasThreadAffinity() {
```
Scottmitch, line 257: remove this method (only used in a single test, just kill the state that uses it too)?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -186,6 +198,80 @@ public final HttpServerBuilder appendConnectionAcceptorFilter(final ConnectionAc
...
+     * </pre>
+     *
+     * @param <FF> Super-type of the filter factory which must implement both {@link StreamingHttpServiceFilterFactory}
+     * and {@link HttpExecutionStrategyInfluencer} though non-offloading filters must not influence strategy.
+     * @param predicate the {@link Predicate} to test if the filter must be applied. This must not block.
+     * @param factory {@link StreamingHttpServiceFilterFactory} to append.
+     * @return {@code this}
+     * @throws IllegalArgumentException if the provided filter or predicate requires offloading.
+     */
+    public final <FF extends StreamingHttpServiceFilterFactory & HttpExecutionStrategyInfluencer> HttpServerBuilder
```
Scottmitch, line 260: `FF` isn't used. should we avoid the generics and just address the `HttpExecutionStrategyInfluencer` being explicit in a followup PR to remote consistent with `appendServiceFilter` (question for both `appendNonOffloadingServiceFilter` methods) ?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -186,6 +198,80 @@ public final HttpServerBuilder appendConnectionAcceptorFilter(final ConnectionAc
...
+                                                 final StreamingHttpServiceFilterFactory factory) {
+        HttpExecutionStrategy predicateStrategy = influenceStrategy(predicate, OFFLOAD_NONE_STRATEGY);
+        if (!OFFLOAD_NONE_STRATEGY.equals(predicateStrategy)) {
+            throw new IllegalArgumentException("Non-offloading predicate required offloading " + predicateStrategy);
+        }
+        HttpExecutionStrategy factoryStrategy = influenceStrategy(factory, OFFLOAD_NONE_STRATEGY);
+        if (!OFFLOAD_NONE_STRATEGY.equals(factoryStrategy)) {
+            throw new IllegalArgumentException("Non-offloading filter required offloading " + factoryStrategy);
+        }
+        appendServiceFilter(service -> new ConditionalHttpServiceFilter(predicate, factory.create(service), service));
```
Scottmitch, line 271: calling `appendServiceFilter` doesn't look quite right. questions:

- does the test verify that no-offloading is done and order is preserved relative to offload filters?
- can you use the code from https://github.com/apple/servicetalk/pull/1710/files#r684634612 (we either need to call `appendNonOffloadingServiceFilter` or use `noOffloadServiceFilters.add`)?

```diff
diff --git a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
index 8771e3ab0..9323a9243 100644
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
@@ -215,14 +215,12 @@ public abstract class HttpServerBuilder {
      *     filter2 ⇒ filter1 ⇒ filter3 ⇒ service
      * </pre>
      *
-     * @param <FF> Super-type of the filter factory which must implement both {@link StreamingHttpServiceFilterFactory}
      * and {@link HttpExecutionStrategyInfluencer} though non-offloading filters must not influence strategy.
      * @param factory {@link StreamingHttpServiceFilterFactory} to append.
      * @return {@code this}
      * @throws IllegalArgumentException if the provided filter requires offloading.
      */
-    public final <FF extends StreamingHttpServiceFilterFactory & HttpExecutionStrategyInfluencer> HttpServerBuilder
-                appendNonOffloadingServiceFilter(final FF factory) {
+    public final HttpServerBuilder appendNonOffloadingServiceFilter(final StreamingHttpServiceFilterFactory factory) {
         requireNonNull(factory);
         HttpExecutionStrategy factoryStrategy = influenceStrategy(factory, OFFLOAD_NONE_STRATEGY);
         if (!OFFLOAD_NONE_STRATEGY.equals(factoryStrategy)) {
@@ -250,26 +248,20 @@ public abstract class HttpServerBuilder {
      *     filter2 ⇒ filter1 ⇒ filter3 ⇒ service
      * </pre>
      *
-     * @param <FF> Super-type of the filter factory which must implement both {@link StreamingHttpServiceFilterFactory}
      * and {@link HttpExecutionStrategyInfluencer} though non-offloading filters must not influence strategy.
      * @param predicate the {@link Predicate} to test if the filter must be applied. This must not block.
      * @param factory {@link StreamingHttpServiceFilterFactory} to append.
      * @return {@code this}
      * @throws IllegalArgumentException if the provided filter or predicate requires offloading.
      */
-    public final <FF extends StreamingHttpServiceFilterFactory & HttpExecutionStrategyInfluencer> HttpServerBuilder
-                appendNonOffloadingServiceFilter(final Predicate<StreamingHttpRequest> predicate,
-                                                 final StreamingHttpServiceFilterFactory factory) {
+    public final HttpServerBuilder appendNonOffloadingServiceFilter(final Predicate<StreamingHttpRequest> predicate,
+                                                                    final StreamingHttpServiceFilterFactory factory) {
         HttpExecutionStrategy predicateStrategy = influenceStrategy(predicate, OFFLOAD_NONE_STRATEGY);
         if (!OFFLOAD_NONE_STRATEGY.equals(predicateStrategy)) {
             throw new IllegalArgumentException("Non-offloading predicate required offloading " + predicateStrategy);
         }
-        HttpExecutionStrategy factoryStrategy = influenceStrategy(factory, OFFLOAD_NONE_STRATEGY);
-        if (!OFFLOAD_NONE_STRATEGY.equals(factoryStrategy)) {
-            throw new IllegalArgumentException("Non-offloading filter required offloading " + factoryStrategy);
-        }
-        appendServiceFilter(service -> new ConditionalHttpServiceFilter(predicate, factory.create(service), service));
-        return this;
+        return appendNonOffloadingServiceFilter(service ->
+                new ConditionalHttpServiceFilter(predicate, factory.create(service), service));
     }
 
     /**

```

Scottmitch (changes requested):

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
bondolo: Removed for now.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
bondolo: Yes, it was busted good. Now corrected. The test is `HttpServerFilterOrderTest` which infers offloading based upon filter execution order but doesn't definitively check it. (I couldn't figure out how to do that will mocks).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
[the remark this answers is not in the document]
Scottmitch: not saying the duplication needs to be addressed in this PR, and no objection to the new implementation ... just want to make sure we will eventually have one implementation. since we are introducing duplication in this PR consider filing an issue associated with the current offloading API scope for followup.

Scottmitch (approved):

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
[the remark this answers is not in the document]
bondolo: This will likely need to be addressed in #1695 with the changes to `HttpStrategyInfluencer`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -186,6 +198,65 @@ public final HttpServerBuilder appendConnectionAcceptorFilter(final ConnectionAc
...
+     * </pre>
+     *
+     * @param predicate the {@link Predicate} to test if the filter must be applied. This must not block.
+     * @param factory {@link StreamingHttpServiceFilterFactory} to append.
+     * @return {@code this}
+     * @throws IllegalArgumentException if the provided filter or predicate requires offloading.
+     */
+    public final HttpServerBuilder appendNonOffloadingServiceFilter(final Predicate<StreamingHttpRequest> predicate,
+                                                                    final StreamingHttpServiceFilterFactory factory) {
+        checkNonOffloading("Non-offloading predicate", noOffloadsStrategy(), predicate);
```
idelpivnitskiy, line 253: This is a great catch that we also have to check the predicate! However, we don't do anything for a predicate in `appendServiceFilter`. Any plans to improve that?

idelpivnitskiy (commented):
Great work!
Minor javadoc improvements in #1729.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
bondolo: Good comment. This comment suggested a bunch of changes to `ConditionalFilterFactory` and related classes, now in PR #1730.

PR #1712, opened 2021-08-02
gRPC protoc plugin proto3 optional support

Scottmitch:
Motivation:
protobuf 3.15 changed optional field support from experimental to
on-by-default. It also requires plugins to opt-in indicating the feature
is supported or else it throws an error and prevents usage of the
plugin.

[1] https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0

Modifications:
- indicate the feature is supported, although services don't interact
  with the optional keyword

Result:
ServiceTalk gRPC protoc plugin can be used when the "optional" proto3
keyword is present.


bondolo (approved):

idelpivnitskiy (approved):

Scottmitch:
0.41 (25afec5dc)

PR #1714, opened 2021-08-03
Fix javadoc in example

idelpivnitskiy:
Motivation:

`./gradlew javadoc` command fails due to some javadoc rules violations
in examples.

Modifications:

- Don't use self-closing tags;
- Fix a link attribute for `a` tag;
- Use `<pre>` + `{@code ...}` to not worry about manual chars escape;

Result:

`./gradlew javadoc` command is happy.


bondolo (approved):

bondolo:
Even though we do not publish the examples JavaDocs should we perhaps generate it during all builds (it won't take much time, probably < 1second) to ensure that it remains compliant?

idelpivnitskiy:
@bondolo makes sense! We can handle it separately. I've created an issue #1715 to keep track of it.

PR #1717, opened 2021-08-04
Update gradlew 7.0.2 -> 7.1.1

Scottmitch:


chemicL (approved):

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.0.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.1.1-bin.zip
```
idelpivnitskiy, line 3: Please, also upgrade the version in the `docs` subproject

idelpivnitskiy (changes requested):

gradle/wrapper/gradle-wrapper.properties
Scottmitch: good catch. why do we need an independent/duplicate Gradle project in the first place?

idelpivnitskiy: I don't remember the full context. Likely because there is no java code at all and we need to avoid any java-plugin from the main project

PR #1718, opened 2021-08-04
Update gradlew wrapper and jar -> 7.1.1

Scottmitch:
Motivation:
Historically the `./gradlew wrapper` command with `--distribution-type`
would make updates to gradlew script and related jar [1]. However that
didn't happen this time and the `--gradle-distribution-url` parameter
was necessary.

[1] `./gradlew wrapper --gradle-version=7.1.1 --distribution-type=bin`
[2] `./gradlew wrapper --gradle-version=7.1.1
--gradle-distribution-url=https://services.gradle.org/distributions/gradle-7.1.1-bin.zip`


bondolo:
Odd, I tried the `--distribution-type=bin` and it still worked for me.  I think it should be preferred if it works since it is less explicit.

Scottmitch:
agreed the more brief option is preferred, not sure why it didn't work this time. maybe I did something wrong, we can re-asses on next update.

PR #1721, opened 2021-08-06
MultiAddressHttpClientBuilder API cleanup

chemicL:
Motivation:

Some overridden methods from `BaseHttpBuilder` have been deprecated and it's safe to remove them. `SingleAddressInitializer` serves the purpose of configuring the individual address' clients.

Some methods are no longer deprecated though, as they allow configuring the global execution context as perceived from the user of the client, e.g. specifying a `BufferAllocator`, which is used from the composite client when preparing the payload.

Modifications:

* Removed inheritance of `HttpClientBuilder` and transitive `BaseHttpBuilder` from `MultiAddressHttpClientBuilder`,
* Introduced an interface for building the final client – `HttpClientBuildFinalizer` which groups the `build*` methods from `HttpClientBuilder`,
* Removed most of deprecated methods,
* Removed deprecation of `ExecutionContext` related methods (`ioExecutor`, `executionStrategy`, `bufferAllocator`). This point can be discarded when the deprecation is included in a PR that is merged before this one, targeting the 0.41 release line,
* Removed `MultiAddressHttpClientFilterFactory` and its' references.

Result:

Most of the deprecated methods are gone and the dependency on a base class meant for single address builders is also removed. The new mechanism using `SingleAddressInitializer` is the only correct way to configure individual underlying clients used by `MultiAddressHttpClientBuilder`. Still, some global execution context can be set on the outer client.


servicetalk-http-api/src/main/java/io/servicetalk/http/api/ExecutionContextAwareHttpBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.api;
+
+import io.servicetalk.buffer.api.BufferAllocator;
+import io.servicetalk.transport.api.IoExecutor;
+
+public interface ExecutionContextAwareHttpBuilder<ResolvedAddress> {
```
Scottmitch, line 21: Can this interface be package private. Ideally we can minimize complexity related to class hierarchy of builders. If not, can you add javadoc?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
```diff
@@ -42,8 +34,8 @@
  * @param <R> the type of address after resolution (resolved address)
  * @see <a href="https://tools.ietf.org/html/rfc7230#section-5.3.2">absolute-form rfc7230#section-5.3.2</a>
  */
-public abstract class MultiAddressHttpClientBuilder<U, R>
-        extends HttpClientBuilder<U, R, ServiceDiscovererEvent<R>> {
+public abstract class MultiAddressHttpClientBuilder<U, R> implements HttpClientBuildFinalizer,
+                                                                     ExecutionContextAwareHttpBuilder {
```
Scottmitch, line 38: `ExecutionContextAwareHttpBuilder` -> `ExecutionContextAwareHttpBuilder<R>`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
```diff
@@ -294,10 +100,16 @@
...
+     *   {@link #initializer(SingleAddressInitializer)} and
+     *   {@link SingleAddressHttpClientBuilder#appendClientFilter(StreamingHttpClientFilterFactory)}
+     *   on the last argument of
+     *   {@link SingleAddressInitializer#initialize(String, Object, SingleAddressHttpClientBuilder)}.
      * @param factory {@link MultiAddressHttpClientFilterFactory} to decorate a {@link StreamingHttpClient} for the
      * purpose of filtering.
      * @return {@code this}
      */
+    @Deprecated
     public abstract MultiAddressHttpClientBuilder<U, R> appendClientFilter(
```
Scottmitch, line 113: can we remove `appendClientFilter*` methods from this PR? IIUC the intention is https://github.com/apple/servicetalk/pull/1722 will deprecate it, and can be back ported to 0.41 ... but this PR won't be back ported so we should be able to remove.

Scottmitch (commented):
few questions but generally looks good!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ExecutionContextAwareHttpBuilder.java
chemicL: That's an omission, should definitely be package private. I'll skip the javadoc as it's not intended to be exposed to users.

chemicL: Re: minimize complexity – I believe short, domain oriented interfaces achieve that purpose. I didn't minimise the amount of interfaces, but rather introduced more decoupling and specialization of interfaces.
As a result of my changes, abstract class hierarchy:
    BaseHttpBuilder <- HttpClientBuilder <- BaseSingleAddressHttpClientBuilder
can be squashed to one abstract class. I'm not sure whether that's what we in fact need, but merging `BaseHttpBuilder` with `HttpClientBuilder` sounds reasonable. WDYT?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
chemicL: That's true, thanks for keeping an eye on that.

chemicL: I removed the `MultiAddressHttpClientFilterFactory` and references to it in the latest commit. #1722 will be updated with relevant deprecations too.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ExecutionContextAwareHttpBuilder.java
Scottmitch: +1 if we can reduce the hierarchy depth and additional types

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientFilterFactory.java
```diff
@@ -27,9 +28,17 @@
 /**
  * A factory for {@link StreamingHttpClientFilter} to filter clients for different unresolved addresses.
  *
+ * @deprecated Use
```
Scottmitch, line 31: should we deprecate in 0.41.x and remove in this PR (or followup PR for 0.42)? looks like this PR removes all practical usages of this interface so it won't be very useful anymore (and other APIs related to it have already been broken).

Scottmitch (commented):
https://github.com/apple/servicetalk/pull/1721#discussion_r685606346

> +1 if we can reduce the hierarchy depth and additional types

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientFilterFactory.java
chemicL: Correct, this class is now removed.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuildFinalizer.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.api;
+
+interface HttpClientBuildFinalizer {
```
idelpivnitskiy, line 18: It can be marked as `@FunctionalInterface`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuildFinalizer.java
```diff
@@ -0,0 +1,32 @@
...
+    StreamingHttpClient buildStreaming();
+
+    default HttpClient build() {
```
idelpivnitskiy, line 21: Please, add javadoc for each method

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuilder.java
```diff
@@ -43,37 +45,126 @@
...
+     * Appends the filter to the chain of filters used to decorate the {@link StreamingHttpConnection} created by this
+     * builder, for every request that passes the provided {@link Predicate}.
+     * <p>
+     * Filtering allows you to wrap a {@link StreamingHttpConnection} and modify behavior during request/response
+     * processing
+     * Some potential candidates for filtering include logging, metrics, and decorating responses.
+     * <p>
+     * The order of execution of these filters are in order of append. If 3 filters are added as follows:
+     * <pre>
+     *     builder.append(filter1).append(filter2).append(filter3)
```
idelpivnitskiy, line 142: We recently improved javadoc for `append*`methods in #1710 and #1729. While you are here, can you please use the actual method name in these cases? For example, here it will be:
```
* builder.appendConnectionFilter(filter1).appendConnectionFilter(filter2).appendConnectionFilter(filter3)
```

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
```diff
@@ -44,8 +44,9 @@
     @Test
     void buildWithDefaults() throws Exception {
         StreamingHttpRequester newRequester = HttpClients.forMultiAddressUrl()
-                .ioExecutor(CTX.ioExecutor())
-                .executionStrategy(defaultStrategy(CTX.executor()))
+                .initializer((scheme, address, builder) -> builder
+                        .ioExecutor(CTX.ioExecutor())
+                        .executionStrategy(defaultStrategy(CTX.executor())))
```
idelpivnitskiy, line 49: 1. Can we revert these changes and use the same methods on `MultiAddressHttpClientBuilder`?
2. Consider adding a separet test to verify that `MultiAddressHttpClientBuilder` correctly sets a use-provided execution context for `client.executionContext()`.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
```diff
@@ -42,8 +29,8 @@
  * @param <R> the type of address after resolution (resolved address)
  * @see <a href="https://tools.ietf.org/html/rfc7230#section-5.3.2">absolute-form rfc7230#section-5.3.2</a>
  */
-public abstract class MultiAddressHttpClientBuilder<U, R>
-        extends HttpClientBuilder<U, R, ServiceDiscovererEvent<R>> {
+public abstract class MultiAddressHttpClientBuilder<U, R> implements HttpClientBuildFinalizer,
```
idelpivnitskiy, line 32: Because this PR is a breaking change anyway, wdyt about making it an `interface` instead of an `abstract class`?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuilder.java
chemicL: I understand you refer to the documentation of the server filters, as I haven't noticed any changes in the client filters in the PRs you linked. I will change the name as requested, but I'm not sure how to proceed about the referred PRs, please let me know.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
chemicL: Agree.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
chemicL: I reverted the changes as requested. I also added a more elaborate test that validates both the external client and the internal one with regards to all the possible execution context settings.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuildFinalizer.java
```diff
@@ -15,17 +15,41 @@
  */
 package io.servicetalk.http.api;
 
+import io.servicetalk.transport.api.ExecutionContext;
+
+@FunctionalInterface
 interface HttpClientBuildFinalizer {
+
+    /**
+     * Builds a new {@link StreamingHttpClient}, using a default {@link ExecutionContext}.
```
idelpivnitskiy, line 24: Does `using a default {@link ExecutionContext}` part make sense? `HttpClientBuildFinalizer` doesn't know if users customized `ExecutionContext` or not.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
```diff
@@ -41,10 +60,13 @@
...
     static final ExecutionContextExtension CTX = immediate();
 
+    @RegisterExtension
+    static final ExecutionContextExtension INTERNAL_CLIENT_CTX = cached();
+
     @Test
     void buildWithDefaults() throws Exception {
         StreamingHttpRequester newRequester = HttpClients.forMultiAddressUrl()
-                .ioExecutor(CTX.ioExecutor())
+                        .ioExecutor(CTX.ioExecutor())
```
idelpivnitskiy, line 69: nit: shifted indentation in all tests

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
```diff
@@ -87,10 +109,78 @@ void buildWithProvidedServiceDiscoverer() throws Exception {
...
+        try (ServerContext serverContext = HttpServers.forAddress(localAddress(0))
+                .executionStrategy(noOffloadsStrategy())
+                .listenStreamingAndAwait((ctx, request, responseFactory) -> succeeded(responseFactory.ok()))) {
+            AtomicReference<BufferAllocator> actualInternalBufferAllocator = new AtomicReference<>();
+            AtomicReference<IoExecutor> actualInternalIoExecutor = new AtomicReference<>();
+            AtomicReference<Executor> actualInternalExecutor = new AtomicReference<>();
+            AtomicReference<ExecutionStrategy> actualInternalExecutionStrategy = new AtomicReference<>();
+
+            final StreamingHttpClient streamingHttpClient = HttpClients.forMultiAddressUrl()
+                    .initializer(((scheme, address, builder) ->
```
idelpivnitskiy, line 145: nit: extra parentheses around lambda

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ExecutionContextAwareHttpBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.api;
+
+import io.servicetalk.buffer.api.BufferAllocator;
+import io.servicetalk.transport.api.IoExecutor;
+
+interface ExecutionContextAwareHttpBuilder<ResolvedAddress> {
```
idelpivnitskiy, line 21: The `ResolvedAddress` param here looks unnecessary. Can we remove it?

idelpivnitskiy (approved):
LGTM! Feel free to merge after the last comments addressed.

P.S. https://github.com/apple/servicetalk/pull/1721#discussion_r687308050
You don't need to refer those PRs, was asking just to use actual method names instead of "append". You did everything correctly, thanks!

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
chemicL: nice catch :)

PR #1724, opened 2021-08-07
Http conditional offload

bondolo:
Draft, for now, until #1710 is completed because it refactors `invokeService` in to `OffloadingFilter`. More work is also still needed to remove the temporarily exposed `isIoExecutorThread` from the public API.


bondolo:
Depends upon #1723

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
```diff
@@ -38,4 +38,12 @@
...
      */
     boolean isFileDescriptorSocketAddressSupported();
+
+    /**
+     * Returns true if the provided thread is an IOExecutor thread.
+     *
+     * @param thread the thread of interest
+     * @return {@code true} if an IOExecutor thread otherwise {@code false}.
+     */
+    boolean isIoExecutorThread(Thread thread);
```
Scottmitch, line 48: We shouldn't need this method and instead we should move this as an implementation detail of the `netty` packages. I think this can be accomplished via the `HttpServerBuilder` (which lives in `http.api`, and will apply the offloading filter in https://github.com/apple/servicetalk/pull/1710) accepting a `Predicate<Thread>` which is translated into the `BooleanSupplier` via:

```java
Predicate<Thread> predicate = thread -> thread instanceof AsyncContextHolderNettyThread;
BooleanSupplier shouldOffload = () -> predicate.test(Thread.currentThread());
```

Also the granularity of this method is off as the `Thread` may or may not be known to this `IoExecutor`, so a single instance of `IoExecutor` won't be authoritative on all `Threads.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultHttpExecutionStrategyTest.java
```diff
@@ -214,11 +217,26 @@ void wrapServiceThatWasNotOffloaded(final Params params) throws Exception {
...
+                    .toFuture().get();
+            } catch (InterruptedException | ExecutionException e) {
+                fail("Unexepected exception", e);
+            } finally {
+              awaitComplete.countDown();
+            }
+        };
+        if (params.offloadSend || params.offloadReceiveMeta || params.offloadReceiveData) {
+            NettyIoExecutor ioExecutor = (NettyIoExecutor) contextRule.ioExecutor();
+            ioExecutor.asExecutor().execute(runHandle);
```
Scottmitch, line 235: you can remove the `awaitComplete` latch if you change this line to `ioExecutor.asExecutor().submit(runHandle).toFuture().get();`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
```diff
@@ -125,20 +126,22 @@ public StreamingHttpService offloadService(final Executor fallback, final Stream
                                                         StreamingHttpRequest request,
                                                         final StreamingHttpResponseFactory responseFactory) {
```
Scottmitch, line 127: `invokeService` should also be updated, or build this PR on top of  https://github.com/apple/servicetalk/pull/1710 and update the `OffloadFilter`.

Scottmitch (changes requested):

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
bondolo: Yes, something similar (or exactly) is the transition plan for this API. The method on IoExecutor is temporary.

bondolo: I have marked it as deprecated in the draft PR.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
bondolo: Since #1710 is nearing completion and #1723 is completed I will rebase this on top of #1710

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
Scottmitch: If this change is easier to accomplish in combination with changes from another PR than can we combine them or sequence the changes differently (e.g. this change built on top of the other)? I don't think it makes sense to have this temporary API merged in, and it would be easier to review/evaluate the end state with out temporary states.

bondolo: The next draft PR is base upon #1710 and now uses the `IoThreadFactory` static method via `NettyHttpServerBuilder`

bondolo:
There are changes to internal APIs which result in this being a "breaking change" but public APIs are not impacted.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -480,6 +480,13 @@ public final ServerContext listenBlockingStreamingAndAwait(
         return listenForService(adapterHolder.adaptor(), adapterHolder.serviceInvocationStrategy());
     }
 
+    /**
+     * Returns true if offloading is appropriate for the execution context.
+     *
+     * @return true if offloading is appropriate for the execution context.
+     */
+    protected abstract boolean shouldOffload();
```
idelpivnitskiy, line 488: Can we defer adding this `protected` method? Consider starting with hardcoded check inside this class and we can make it `protected` on-demand.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
```diff
@@ -24,6 +24,15 @@
  */
 public interface IoExecutor extends ListenableAsyncCloseable {
 
+    /**
+     * Marker interface for IO Threads. All threads created by {@link IoExecutor}s are expected to implement this
+     * interface.
+     *
+     * @see #isIoExecutorThread(Thread)
+     */
+    interface IoThread {
```
idelpivnitskiy, line 33: My biggest worry is that there is no way for users to use `IoThreadFactory`, because it's an internal class. Anyone, who is creating a custom `IoExecutor` (I'm sure there are such users), will end up being non-offloaded. 

Did you consider the approach we discussed offline of adding `IoThreadFactory` interface in `transport-api`?
```java
    interface IoThreadFactory<T extends Thread & IoThread> extends ThreadFactory {

        @Override
        T newThread(Runnable r);
    }
``` 
Then changing `io.servicetalk.transport.netty.NettyIoExecutors` to require it.

Ideally, we need to introduce `IoThreadFactory` and and methods for `NettyIoExecutors` in 0.41 branch. Then in `main` branch we can remove old static factories and make sure we won't break users.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
```diff
@@ -38,4 +47,14 @@
...
     boolean isFileDescriptorSocketAddressSupported();
+
+    /**
+     * Returns true if the provided thread is an IOExecutor thread.
+     *
+     * @param thread the thread of interest
+     * @return {@code true} if an IOExecutor thread otherwise {@code false}.
+     */
+    static boolean isIoExecutorThread(Thread thread) {
+        return thread instanceof IoThread;
```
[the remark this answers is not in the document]
bondolo: The intention of providing it as a method was for the abstraction value, the implementation could change and user code would not need to change.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/IoThreadFactory.java
```diff
@@ -92,4 +93,8 @@ public AsyncContextMap asyncContextMap() {
             return asyncContextMap;
         }
     }
+
+    public static boolean isIOThread(Thread thread) {
+        return thread instanceof AsyncContextHolderNettyThread;
```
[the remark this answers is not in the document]
bondolo: Removed. All instances should have been replaced with the `IoExecutor` method.

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/NettyChannelPublisherTest.java
```diff
@@ -97,7 +113,8 @@ public void read(ChannelHandlerContext ctx) throws Exception {
                     readRequested = true;
                     super.read(ctx);
                 }
-            }), OFFLOAD_ALL_STRATEGY, mock(Protocol.class), NoopConnectionObserver.INSTANCE, true).toFuture().get();
+            }), offloadStrategy.strategy(),
```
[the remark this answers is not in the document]
bondolo: Neither `OFFLOAD_ALL_STRATEGY` or `OFFLOAD_FROM_IO_STRATEGY` did anything and neither is referenced outside of tests. This does seem to be the right place to test it (I could be wrong) but I could not understand what causes the tests to behave differently than how production would. Nor could I find the production code equivalent of what the tests are doing with `OFFLOAD_ALL_STRATEGY`.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
bondolo: Originally the method was going to be specific to Netty implementation and this was the abstraction and I am still thinking that perhaps it should be. The `IoExecutor.IoThread` marker interface doesn't feel like it should be public.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
bondolo: This does seem like a good option.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/OffloadAllExecutionStrategy.java
```diff
@@ -31,23 +31,23 @@ private OffloadAllExecutionStrategy() {
     }
 
     @Override
-    public <T> Single<T> offloadSend(final Executor fallback, final Single<T> original) {
-        return original.subscribeOn(fallback);
+    public <T> Single<T> offloadSend(final Executor executor, final Single<T> original) {
```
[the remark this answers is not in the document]
bondolo: What equivalent class is used for non-test code?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
bondolo: It seems like this might only be possible as a hard breaking change since there will be no graceful support for IO threads that do not implement IoThread marker, ServiceTalk would just break if you tried it because necessary offloading would not be done.

It does seem to be the right answer though.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/OffloadAllExecutionStrategy.java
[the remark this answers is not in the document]
idelpivnitskiy: You introduced `OffloadFromIOExecutionStrategy` instead of `OffloadAllExecutionStrategy`. As the result, `OffloadAllExecutionStrategy` is not used by any production code anymore.

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/NettyChannelPublisherTest.java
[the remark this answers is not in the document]
idelpivnitskiy: Discussed offline, these tests don't actually use offloading strategy, so doesn't matter what strategy we use here.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
idelpivnitskiy: We do not have any known use-cases when users extend `HttpServerBuilder`. Therefore, proposing to defer expanding public API until we have a use-case.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
idelpivnitskiy: Discussed offline. We will introduce `IoThreadFactory` interface in 0.41, add new factory methods for `NettyIoExecutors`, and deprecate old factories. Then we can break API in 0.42+

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/OffloadAllExecutionStrategy.java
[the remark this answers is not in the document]
bondolo: Moved `OffloadAllExecutionStrategy` to test sources.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
bondolo: Leaving this open until PR for 0.41 branch is created.

bondolo: Deprecation PR now available as #1738

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoExecutors.java
```diff
@@ -101,6 +111,17 @@ public static NettyIoExecutor fromNettyEventLoop(EventLoop eventLoop) {
...
     }
 
+    /**
+     * Creates a new instance of {@link NettyIoExecutor} using the passed {@link EventLoop}.
+     *
+     * @param eventLoop {@link EventLoop} to use to create a new {@link NettyIoExecutor}.
+     * @param isIoThreadSupported if {@code true} then event loop threads are marked as {@link IoThreadFactory.IoThread}
+     * @return New {@link NettyIoExecutor} using the passed {@link EventLoop}.
+     */
+    public static NettyIoExecutor fromNettyEventLoop(EventLoop eventLoop, boolean isIoThreadSupported) {
```
idelpivnitskiy, line 121: Consider adding a similar overload for `fromNettyEventLoopGroup`, in case someone uses it to convert `EventLoopGroup` to `IoExecutor`

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoExecutor.java
```diff
@@ -33,4 +34,11 @@
      * @return an {@link Executor} which will use an {@link IoExecutor} thread for execution.
      */
     Executor asExecutor();
+
+    /**
+     * Determine if threads used by this {@link IoExecutor} are marked with {@link IoThreadFactory.IoThread} interface.
+     *
+     * @return {@code true} if supported
+     */
+    boolean isIoThreadSupported();
```
idelpivnitskiy, line 43: Don't have a strong preference, but since we introduced this flag specifically for EventLoop use-cases, consider moving it to `EventLoopAwareNettyIoExecutor` interface.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -261,9 +263,11 @@ protected void handleSubscribe(
                 final DelayedCancellable delayedCancellable;
                 try {
                     delayedCancellable = new DelayedCancellable();
-                    DefaultNettyConnection<Read, Write> connection = new DefaultNettyConnection<>(channel, allocator,
-                            executor, terminalPredicate, closeHandler, flushStrategy, idleTimeoutMs,
-                            executionStrategy, protocol, null, null, NoopDataObserver.INSTANCE, isClient,
+                    DefaultExecutionContext executionContext = new DefaultExecutionContext(allocator,
+                            fromNettyEventLoop(channel.eventLoop()), executor, executionStrategy);
```
idelpivnitskiy, line 267: We should also check & pass the `isIoThreadSupported` here

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
```diff
@@ -99,17 +104,19 @@ public StreamingHttpService offloadService(final Executor fallback, final Stream
...
                 // The service should see this ExecutionStrategy inside the ExecutionContext:
                 final HttpServiceContext wrappedCtx =
                         new ExecutionContextOverridingServiceContext(ctx, DefaultHttpExecutionStrategy.this, e);
                 if (diff == null) {
                     return service.handle(wrappedCtx, request, responseFactory);
                 } else {
                     if (diff.isDataReceiveOffloaded()) {
-                        request = request.transformMessageBody(p -> p.publishOn(e));
+                        request = request.transformMessageBody(p ->
+                                p.publishOn(e, DefaultHttpExecutionStrategy::shouldOffload));
```
idelpivnitskiy, line 116: This `shouldOffload` does not verify `isIoThreadSupported()`, consider checking what is in `ctx`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -111,20 +113,41 @@ public HttpServerBuilder bufferAllocator(final BufferAllocator allocator) {
...
+
+    @Override
+    protected HttpExecutionContext buildExecutionContext(final HttpExecutionStrategy strategy) {
+        executionContextBuilder.executionStrategy(strategy);
+        return executionContextBuilder.build();
+    }
+
+    @Override
+    protected BooleanSupplier shouldOffload(IoExecutor ioExecutor) {
+        return ioExecutor instanceof NettyIoExecutor && ((NettyIoExecutor) ioExecutor).isIoThreadSupported() ?
```
idelpivnitskiy, line 130: Instead of adding these new public methods, can `OffliadingFilter` compute the `BooleanSupplied` based on `HttpServiceContext` it receives with each request? `DefaultHttpExecutionStrategy` will need to do the same.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
```diff
@@ -121,7 +128,10 @@ public StreamingHttpService offloadService(final Executor fallback, final Stream
                             // This is different as compared to invokeService() where we just offload once on the
                             // flattened (meta + data) stream. In this case, we need to preserve the service contract
                             // and hence have to offload both meta and data separately.
-                            resp.map(r -> r.transformMessageBody(p -> p.subscribeOn(e))).subscribeOn(e) : resp;
+                            resp.map(r -> r.transformMessageBody(
+                                    p -> p.subscribeOn(e, DefaultHttpExecutionStrategy::shouldOffload)))
```
idelpivnitskiy, line 132: Will the client-side offliading be implemented in a follow-up PR?

PR #1727, opened 2021-08-11
Fix gradle grpc protoc project local dependency sequencing

Scottmitch:
Motivation:
servicetalk-grpc-protoc has tests in the local directory that depend
upon the compileJava task to complete and the buildExecutable task to
complete. However the dynamically generated generateTestProto task
didn't have an explicit ordering/sequencing dependency on the
buildExecutable task so the build would occasionally fail due to lack of
grpc-protoc artifact.

Modifications:
- Add an explicit dependency from the generateTestProto task on the
  buildExecutable task
- Use a consistent strategy to specify the dependency for other
  sub-projects in the same gradle projects.

Result:
No more spurious build failures due to missing grpc-protoc artifacts.

### DIFF
--- a/servicetalk-examples/grpc/compression/build.gradle
+++ b/servicetalk-examples/grpc/compression/build.gradle
@@ -41,15 +41,21 @@ protobuf {
   }
   plugins {
     servicetalk_grpc {
-      //// Users are expected to use "artifact" instead of "path". we use "path"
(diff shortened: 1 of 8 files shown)


Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1507

tkountis (approved):

PR #1728, opened 2021-08-11
Safer offloading handleSubscribe invocation

bondolo:
Motivation:
The current offloading operators have the possibility to call
`onSubscribe` twice if the `onSubscribe` method throws an exception.
This should be avoided.
Modifications:
Introduces a new safeHandleSubscribe static method that is invoked in
a slightly different context that avoids the possibility of double
calls to `onSubscribe`.
Result:
Better conformance to reactive specification.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
@@ -23,6 +23,7 @@ import java.util.function.BooleanSupplier;
 
 import static io.servicetalk.concurrent.api.Executors.immediate;
 import static io.servicetalk.concurrent.internal.SubscriberUtils.deliverErrorFromSource;
+import static io.servicetalk.concurrent.internal.SubscriberUtils.safeOnError;
 
 /**
  * A set of factory methods that provides implementations for the various publish/subscribeOn methods on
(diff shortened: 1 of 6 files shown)


bondolo:
Thank you @idelpivnitskiy for noticing this problem and describing the solution.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
```diff
@@ -44,6 +45,20 @@ static void deliverOnSubscribeAndOnError(Subscriber subscriber,
         deliverErrorFromSource(contextProvider.wrapCompletableSubscriber(subscriber, contextMap), cause);
     }
 
+    @FunctionalInterface
+    interface HandleSubscribe {
```
idelpivnitskiy, line 49: Consider making the interface and static method `private` in all variants.

idelpivnitskiy (approved):
LGTM after comments addressed.
Thanks for teaching me a trick with `@FunctionalInterface`!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
```diff
@@ -71,17 +86,20 @@ static Completable subscribeOn(final Completable original,
...
         void handleSubscribe(final Subscriber subscriber,
                              final AsyncContextMap contextMap, final AsyncContextProvider contextProvider) {
+            final Subscriber upstreamSubscriber;
             try {
                 BooleanSupplier shouldOffload = shouldOffload();
-                Subscriber upstreamSubscriber =
+                upstreamSubscriber =
                         new CompletableSubscriberOffloadedTerminals(subscriber, shouldOffload, executor());
 
                 // Note that the Executor is wrapped by default to preserve AsyncContext, so we don't have to re-wrap
```
chemicL, line 95: This comment was linked to the line that is now moved further down, consider moving it too.

chemicL (approved):

chemicL:
LGTM. Perhaps worth adding a test case to avoid regressions?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublishAndSubscribeOnCompletables.java
bondolo: It is actually a comment at the site where the (unnecessary, per the comment) async context wrapping would be done. I believe it is in the correct location since if the wrapping was needed it would be in that try/catch block.

PR #1730, opened 2021-08-12
Conditional filter support for StrategyInfluencer

bondolo:
Motivation:
If a filter implements `HttpExecutionStrategyInfluencer` this needs to
be fully supported by `ConditionalFilterFactory` and other wrappers.
The `Predicate` should also be considered as a potential influencer.
Modifications:
 `ConditionalFilterFactory` now implements
 `HttpExecutionStrategyInfluencer ` and both it and
 `StrategyInfluencerAwareConversions` consider the `Predicate` as an
 influence source.
Result:
More accurate strategy influence.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/StrategyInfluencerAwareConversions.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/StrategyInfluencerAwareConversions.java
@@ -25,43 +25,37 @@ final class StrategyInfluencerAwareConversions {
         // No instances.
     }
 
+    @Deprecated
     static <U> StreamingHttpClientFilterFactory toClientFactory(final U address,
                                                                 final MultiAddressHttpClientFilterFactory<U> original) {
         requireNonNull(address);
-        if (original instanceof HttpExecutionStrategyInfluencer) {
-            HttpExecutionStrategyInfluencer influencer = (HttpExecutionStrategyInfluencer) original;
-            return new StrategyInfluencingStreamingClientFilterFactory() {
-                @Override
-                public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strategy) {
-                    return influencer.influenceStrategy(strategy);
-                }
+        return new StrategyInfluencingStreamingClientFilterFactory() {
+            @Override
+            public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strategy) {
+                return HttpExecutionStrategyInfluencer.applyInfluence(original, strategy);
+            }
 
-                @Override
-                public StreamingHttpClientFilter create(final FilterableStreamingHttpClient client) {
-                    return original.create(address, client);
-                }
-            };
-        }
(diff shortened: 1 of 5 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/StrategyInfluencerAwareConversions.java
```diff
@@ -25,134 +25,116 @@ private StrategyInfluencerAwareConversions() {
...
-                public StreamingHttpClientFilter create(final FilterableStreamingHttpClient client) {
-                    return new ConditionalHttpClientFilter(predicate, original.create(client), client);
-                }
-            };
-        }
-        return client -> new ConditionalHttpClientFilter(predicate, original.create(client), client);
+        return new StrategyInfluencingStreamingClientFilterFactory() {
+            @Override
+            public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strategy) {
+                return HttpExecutionStrategyInfluencer.applyInfluence(predicate,
```
idelpivnitskiy, line 107: While it's right to include `predicate` into `HttpExecutionStrategyInfluencer` analysis, I don't see how it will match with our intention to make `HttpExecutionStrategyInfluencer` more explicit/visible. For a predicate, users usually use lambda. Should we think a bit more about the approach here?

idelpivnitskiy (commented):
That was too fast :) 
I didn't expect to see the PR right away, and thought about the predicate more as part of `HttpExecutionStrategyInfluencer` API change and enhancements.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StrategyInfluencerAwareConversions.java
bondolo: Agreed that it is concerning and this issue has been a good exercise. I've been thinking about what should be potentially considered an influencer since the `TimeoutHttpServiceFilter` `TimeoutFromRequest` interface. Any bit of user code that will be executed could potentially be an influencer (or potentially block). Requiring things that would naturally be functional interfaces to implement the interface seems onerous. I think that perhaps we should use specific purpose interfaces like the `TimeoutFromRequest` and provide currying utility  methods to convert between simple functional interfaces that people want to use and the special purpose interfaces. The assistance we provide could either be `default` methods providing a basic behaviour or 

```
static TimeoutFromRequest create(Function<HttpRequestMetaData, Duration> function, 
                                                             HttpExecutionStrategy requiredStrategy) {
        return new TimeoutFromRequest() {
            @Nullable
            @Override
            public Duration apply(final HttpRequestMetaData request) {
                return function.apply(request);
            }

            @Override
            public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strategy) {
                return strategy.merge(requiredStrategy);
            }
        };
    }
```

I want to avoid a solution which requires users to write all that tedious boiler plate to use our interfaces.

bondolo:
This PR was an exploration, elements will be brought forward in other PRs.

PR #1732, opened 2021-08-16
Introduce ClientIpUtils to help extracting actual client IP from requests.

tkountis:
Motivation:

Requests that go through proxies can end up affecting the remote_address of the incoming request's connection.
The developer leveraing ST APIs can not access the original Client IP in that case unless they manually parse the headers and check
for forwarding info.

Modifications:

Introduce a helper inside servicetalk-http-utils that can parse X-FORWARDED-FOR and FORWARDED headers to extract the original client IP if available.

Result:

Public implementation of extracting the original client IP instead of custom ones in the wild.

### DIFF
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/ClientIpUtils.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/ClientIpUtils.java
@@ -1,0 +1,215 @@
+/*
+ * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.utils;
+
+import io.servicetalk.http.api.HttpHeaderNames;
+import io.servicetalk.http.api.HttpHeaders;
+import io.servicetalk.transport.api.ConnectionContext;
+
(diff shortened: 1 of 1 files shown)


chemicL:
Have you considered adding some basic unit tests?

tkountis:
> Have you considered adding some basic unit tests?

There is testing that verifies this behaviour but its not in this repo. 
I will do so.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/ClientIpUtils.java
```diff
@@ -0,0 +1,215 @@
...
+    static CharSequence extractClientIp(final HttpHeaders headers,
+                                        final ConnectionContext ctx,
+                                        final boolean fallbackToSocketAddress) {
+
+        final CharSequence ip = parseXForwardedFor(headers.get(X_FORWARDED_FOR));
+        if (ip != null) {
+            return ip;
+        }
+
+        final CharSequence ipFromForwarded = parseForwarded(headers.get(FORWARDED));
```
chemicL, line 61: Shouldn't the `Forwarded` header be considered first if it exists as it's a standardized header (https://datatracker.ietf.org/doc/html/rfc7239) and only if it's missing should the `X-Forwarded-For` be considered?
Also, the RFC recommends mapping `XFF` to `Forwarded` and have a unified parsing strategy: https://datatracker.ietf.org/doc/html/rfc7239#section-7.4

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/ClientIpUtils.java
```diff
@@ -0,0 +1,215 @@
...
+        int endIdx = computeForParamEndIdx(str, beginIdx);
+        if (endIdx - beginIdx < 7) {
+            // not enough data to extract the shortest valid origin IP address: "[::1]" or 0.0.0.0
+            return null;
+        }
+
+        if (str.charAt(beginIdx) == '"') {
+            // unwrap quoted value
+            beginIdx++;
+            endIdx--;
```
chemicL, line 125: Worth checking if the corresponding character is also `"`. If it's improperly formatted it might escape some prior validation and can be used to trick the util to read a spoofed IP.

PR #1734, opened 2021-08-16
Deprecate `IoExecutor` create w/ `ThreadFactory`

bondolo:
Motivation:
Future versions of ServiceTalk will require that all IoExecutor threads
implement the `IoThreadFactory.IoThread` marker interface so that the
threads can be recognized for offloading. Usages pf existing APIs which
accept a `ThreadFactory` need to be replaced to use variants that
accept an `IoThreadFactory`.
Modifications:
Introduces `IoThreadFactory` which will be required in the future for
`IoExecutor`. Adds new methods accepting `IoThreadFactory` and
deprecates existing `IoExecutor` creation APIs which accept
`ThreadFactory`.
Result:
API and deprecations anticipate future required behavior.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoThreadFactory.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoThreadFactory.java
@@ -15,6 +15,8 @@
  */
 package io.servicetalk.transport.api;
 
+import io.servicetalk.transport.api.IoThreadFactory.IoThread;
+
 import java.util.concurrent.ThreadFactory;
 
 /**
@@ -22,7 +24,8 @@ import java.util.concurrent.ThreadFactory;
  *
  * @param <T> Type of threads created
  */
-public interface IoThreadFactory<T extends Thread & IoThreadFactory.IoThread> extends ThreadFactory {
+@FunctionalInterface
+public interface IoThreadFactory<T extends Thread & IoThread> extends ThreadFactory {
 
     /**
      * Marker interface for IO Threads. All threads created by a {@link IoThreadFactory} are expected to implement this
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoThreadFactory.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoThreadFactory.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.transport.api;
+
+import java.util.concurrent.ThreadFactory;
+
+/**
+ * Thread factory for use with {@link IoExecutor}.
+ *
+ * @param <T> Type of threads created
+ */
+public interface IoThreadFactory<T extends Thread & IoThreadFactory.IoThread> extends ThreadFactory {
+
+    /**
+     * Marker interface for IO Threads. All threads created by a {@link IoThreadFactory} are expected to implement this
+     * interface.
+     */
+    interface IoThread {
+    }
+
+    @Override
+    T newThread(Runnable r);
+}
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoExecutors.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoExecutors.java
@@ -16,7 +16,7 @@
 package io.servicetalk.transport.netty.internal;
 
 import io.servicetalk.transport.api.IoExecutor;
-import io.servicetalk.transport.api.IoThreadFactory;
+import io.servicetalk.transport.api.IoThreadFactory.IoThread;
 
 import io.netty.channel.EventLoop;
 import io.netty.channel.EventLoopGroup;
@@ -44,12 +44,12 @@ public final class NettyIoExecutors {
      * Create a new {@link NettyIoExecutor} with the default number of {@code ioThreads}.
      *
      * @param <T> Type of the IO thread instances created by factory.
-     * @param threadFactory the {@link ThreadFactory} to use. If possible you should use an instance of
-     * {@link NettyIoThreadFactory} as it allows internal optimizations.
+     * @param threadFactory the {@link io.servicetalk.transport.api.IoThreadFactory} to use. If possible you should use
+     * an instance of {@link IoThreadFactory} as it allows internal optimizations.
      * @return The created {@link IoExecutor}
      */
-    public static <T extends Thread & IoThreadFactory.IoThread> EventLoopAwareNettyIoExecutor createIoExecutor(
-            IoThreadFactory<T> threadFactory) {
+    public static <T extends Thread & IoThread> EventLoopAwareNettyIoExecutor createIoExecutor(
+            io.servicetalk.transport.api.IoThreadFactory threadFactory) {
         return createIoExecutor(getRuntime().availableProcessors() * 2, threadFactory);
     }
 
@@ -58,8 +58,9 @@ public final class NettyIoExecutors {
      *
      * @param threadFactory the {@link ThreadFactory} to use.
      * @return The created {@link IoExecutor}
-     * @deprecated Future versions of ServiceTalk will require a {@link IoThreadFactory} for creating
-     * {@link IoExecutor} threads.
+     * @deprecated Future versions of ServiceTalk will require a
+     * {@link io.servicetalk.transport.api.IoThreadFactory} for creating {@link IoExecutor} threads, use
+     * {@link #createEventLoopGroup(int, io.servicetalk.transport.api.IoThreadFactory)} instead.
      */
     @Deprecated
     public static EventLoopAwareNettyIoExecutor createIoExecutor(ThreadFactory threadFactory) {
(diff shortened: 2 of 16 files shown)


bondolo:
Note carefully that this PR is for the sustaining **0.41** branch.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoExecutors.java
```diff
@@ -39,35 +40,89 @@ private NettyIoExecutors() {
...
 
+    /**
+     * Create a new {@link NettyIoExecutor} with the default number of {@code ioThreads}.
+     *
+     * @param <T> Type of the IO thread instances created by factory.
+     * @param threadFactory the {@link ThreadFactory} to use. If possible you should use an instance of
+     * {@link NettyIoThreadFactory} as it allows internal optimizations.
+     * @return The created {@link IoExecutor}
+     */
+    public static <T extends Thread & IoThreadFactory.IoThread> EventLoopAwareNettyIoExecutor createIoExecutor(
```
idelpivnitskiy, line 51: Consider importing `IoThread`

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoExecutors.java
```diff
@@ -39,35 +40,89 @@ private NettyIoExecutors() {
...
+        return createIoExecutor(getRuntime().availableProcessors() * 2, threadFactory);
+    }
+
     /**
      * Create a new {@link NettyIoExecutor} with the default number of {@code ioThreads}.
      *
      * @param threadFactory the {@link ThreadFactory} to use.
      * @return The created {@link IoExecutor}
+     * @deprecated Future versions of ServiceTalk will require a {@link IoThreadFactory} for creating
+     * {@link IoExecutor} threads.
```
idelpivnitskiy, line 62: Consider referencing an appropriate non-deprecated overload:
```
@deprecated Use {@link #createIoExecutor(IoThreadFactory)}. Future versions...
```

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoExecutors.java
```diff
@@ -39,35 +40,89 @@ private NettyIoExecutors() {
...
     /**
      * Create a new {@link EventLoopGroup}.
      *
      * @param ioThreads number of threads
      * @param threadFactory the {@link ThreadFactory} to use.
      * @return The created {@link IoExecutor}
+     * @deprecated Future versions of ServiceTalk will require a {@link IoThreadFactory} for creating
+     * {@link IoExecutor} threads.
      */
+    @Deprecated
```
idelpivnitskiy, line 125: Looking at `fromNettyEventLoop` and `fromNettyEventLoopGroup`. This is unfortunate for the use-case of detecting `IoThread` as we don't have control how that `EventLoop[Group]` is created. 
This is internal API, but some rare users potentially can use those methods. They are useful when you already have a netty-based application and you would like to share `EventLoop` between existing app/library and ST.
@bondolo any ideas how to make a predicate for conditional offloading operators more reliable?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoThreadFactory.java
```diff
@@ -17,20 +17,20 @@
...
 
 import static java.lang.Thread.NORM_PRIORITY;
 import static java.util.Objects.requireNonNull;
 
 /**
- * Default {@link ThreadFactory} to create IO {@link Thread}s.
+ * Default {@link NettyIoThreadFactory} to create IO {@link IoThread}s.
  */
-public final class IoThreadFactory implements ThreadFactory {
+public final class NettyIoThreadFactory implements IoThreadFactory<NettyIoThreadFactory.NettyIoThread> {
```
idelpivnitskiy, line 33: Even for internal modules, we should be extra cautious and maintain backward bytecode compatibility for 0.41, because some older ST version may require previous `IoThreadFactory`. Consider keeping both for now: `IoThreadFactory` and `NettyIoThreadFactory`.

servicetalk-transport-netty/src/main/java/io/servicetalk/transport/netty/NettyIoExecutors.java
```diff
@@ -35,13 +36,30 @@ private NettyIoExecutors() {
      *
      * @param ioThreads number of threads.
      * @param threadFactory the {@link ThreadFactory} to use. If possible you should use an instance
-     * of {@link IoThreadFactory} as it allows internal optimizations.
+     * of {@link NettyIoThreadFactory} as it allows internal optimizations.
```
idelpivnitskiy, line 39: Public module should not reference internal classes.

servicetalk-transport-netty/src/main/java/io/servicetalk/transport/netty/NettyIoExecutors.java
```diff
@@ -35,13 +36,30 @@ private NettyIoExecutors() {
...
     public static IoExecutor createIoExecutor(int ioThreads, ThreadFactory threadFactory) {
         return io.servicetalk.transport.netty.internal.NettyIoExecutors.createIoExecutor(ioThreads, threadFactory);
     }
 
+    /**
+     * Creates a new {@link IoExecutor} with the specified number of {@code ioThreads}.
+     *
+     * @param <T> Type of the IO thread instances created by factory.
+     * @param ioThreads number of threads.
+     * @param threadFactory the {@link ThreadFactory} to use. If possible you should use an instance
```
idelpivnitskiy, line 54: Consider linking to `IoThreadFactory`

servicetalk-transport-netty/src/main/java/io/servicetalk/transport/netty/NettyIoExecutors.java
```diff
@@ -35,13 +36,30 @@ private NettyIoExecutors() {
...
+     * of {@link NettyIoThreadFactory} as it allows internal optimizations.
+     * @return The created {@link IoExecutor}
+     */
+    public static <T extends Thread & IoThreadFactory.IoThread> IoExecutor createIoExecutor(int ioThreads,
```
idelpivnitskiy, line 58: Consider importing `IoThread`

idelpivnitskiy (changes requested):
It's good to maintain consistent history in the `main` branch. Prefer opening all PRs for the `main` and cherry-pick into `0.41`, unless a PR is specific to `0.41`. In this case, we need `IoThreadFactory` in the `main`.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoExecutors.java
```diff
@@ -39,35 +40,89 @@ private NettyIoExecutors() {
         // No instances.
     }
 
+    /**
+     * Create a new {@link NettyIoExecutor} with the default number of {@code ioThreads}.
+     *
+     * @param <T> Type of the IO thread instances created by factory.
+     * @param threadFactory the {@link ThreadFactory} to use. If possible you should use an instance of
```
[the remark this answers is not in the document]
bondolo: Ooops, I missed those.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoThreadFactory.java
```diff
@@ -0,0 +1,36 @@
...
+package io.servicetalk.transport.api;
+
+import java.util.concurrent.ThreadFactory;
+
+/**
+ * Thread factory for use with {@link IoExecutor}.
+ *
+ * @param <T> Type of threads created
+ */
+public interface IoThreadFactory<T extends Thread & IoThreadFactory.IoThread> extends ThreadFactory {
```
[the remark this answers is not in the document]
bondolo: > * …
> * …
> * Not necessary to change now, just thinking out loud: because we have a way to enforce  what else a `Thread` needs to implement, we can also enforce `AsyncContextMapHolder` interface to make sure users don't see perf degragation if they have a custom `ThreadFactory`. Would be glad to hear @Scottmitch opinion on that. The only hesitation I have: if we add it later, it will be API breaking change. Maybe we should defer cherry-picking this PR into 0.41 branch until after we have 0.41.4 release to give us some time to discuss.

Since `AsyncContextMapHolder` is a public API type it does not seem unreasonable to add to the required interface types for `IoThread`. For implementation of the interface we will have to provide sample code as the interface cannot be implemented via default methods (needs state).

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoExecutors.java
bondolo: Not really unfortunately. Other than the general `IoThread` mechanism we haven't come up with another good approach for identifying cases where we should offload. Solutions which identify which threads belong to which executors are complicated and would be annoying to retrofit back in to existing `ThreadFactory` and `Executor` implementations. The most universal conditional offloading predicate would be to ask the target `Executor` if the current thread is one of theirs and if it returns false, then offload. The current approach for #1724 of using the IoThread marker interface is not as complete but sufficient for protecting event loop threads.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoThreadFactory.java
bondolo: The name change is because spotbugs won't allow the new interface and the existing class to have the same name. We would need to add an exclusion rule.

servicetalk-transport-netty/src/main/java/io/servicetalk/transport/netty/NettyIoExecutors.java
bondolo: I have removed this reference in the public module.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoThreadFactory.java
bondolo: Name changed back to IoThreadFactory and spot bugs exclusion rule applied.

idelpivnitskiy:
It's hard to review two PRs, I will close this one and review it in #1738. We will cherry-pick into 0.41 after the other one is merged.

PR #1739, opened 2021-08-17
IoThreadFactory create daemon threads by default

bondolo:
Motivation:
The `IoThreadFactory` default constructor creates non-daemon threads
by default. For some types of servers a non-daemon thread is
advantageous where the initial thread starts the server and eventually
exits. The server thread then keeps the JVM running as long as The
server continues to accept connections. As ServiceTalk applications
don't rely on this behaviour all ServiceTalk threads should be daemon
threads.
Modifications:
`IoThreadFactory(String threadNamePrefix)` constructor changed to
create daemon threads.
Result:
ServiceTalk IO Threads will no longer keep JVM alive if application
threads have quit.


idelpivnitskiy (approved):
Good catch!
IIUC, all our code (`GlobalExecutionContext`, etc.) always uses `true`. So, it will be good if this ctor is consistent with what we use.

PR #1741, opened 2021-08-18
API to support start/stop accepting connections

tkountis:
Motivation:

Expose a way a user can yield accepting connection on server side, and resume on demand.?

Modifications:

ServerContext now supports an additional `acceptConnections(bool)` API that can be used to
hint to the server the need for start/stop accepting.

Result:

More ways to control the service on-demand.


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServerContext.java
```diff
@@ -50,6 +50,35 @@ default void awaitShutdown() {
...
 
+    /**
+     * Toggles the server's channel accepting connection ability.
+     * <p>
+     * Passing a {@code false} value, will stop accepting connections on the server channel, without affecting any other
+     * interaction to currently open child channels (i.e., reads / writes).
+     * <p>
+     * Depending on the transport, connections may still get ESTABLISHED, see.
+     * {@link ServiceTalkSocketOptions#SO_BACKLOG backlog} and
+     * <a href="https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt">SOMAXCONN</a>.
```
chemicL, line 61: Perhaps worth highlighting it's a Linux specific example, for instance by
```
    * (...) and (in case of Linux)
    * <a href="https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt">SOMAXCONN</a>.
```

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServerContext.java
```diff
@@ -50,6 +50,35 @@ default void awaitShutdown() {
...
+     * For instance, in case of TCP the 3-way handshake may finish, and the connection will await in the
+     * accept queue to be accepted. If the accept queue is full, connection SYNs will await in the
+     * SYN backlog (i.e., <a href="https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt">
```
chemicL, line 64: Same as above.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServerContext.java
```diff
@@ -50,6 +50,35 @@ default void awaitShutdown() {
...
+     * tcp_max_syn_backlog</a>). These additional parameters may affect the behavior of new flows when the service
+     * is not accepting.
+     * <p>
+     * Depending on how long this stays in the {@code false} state, it may affect other timeouts (i.e., connect-timeout
+     * or idleness) on the peer-side and/or the other flows to the peer (i.e., proxies).
+     * <p>
+     * Notes:
+     * <ul>
+     * <li>Upon resumption, {@code accept == true}, backlogged connection will be processed first,
```
chemicL, line 73: connection -> connections

chemicL (approved):
Minor comments and a suggestion for tests readability, but in general LGTM.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
```diff
@@ -74,6 +74,11 @@ private H2ServerParentConnectionContext(final Channel channel, final BufferAlloc
         this.listenAddress = requireNonNull(listenAddress);
     }
 
+    @Override
+    public void acceptConnections(final boolean accept) {
+        channel().config().setAutoRead(accept);
```
bondolo, line 79: Does this prevent new connections from being put in the accept queue? What happens to connections in the accept queue? Are they rejected or will they just timeout on the client side?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptingNettyHttpServerTest.java
```diff
@@ -0,0 +1,128 @@
...
+    private static final int CONNECT_TIMEOUT_MILLIS = (int) ofSeconds(5).toMillis();
+    private static final int VERIFY_REQUEST_AWAIT_SECS = 5;
+    private static final int TRY_REQUEST_AWAIT_SECS = 1;
+
+    private final StreamingHttpRequestResponseFactory reqRespFactory =
+            new DefaultStreamingHttpRequestResponseFactory(DEFAULT_ALLOCATOR, INSTANCE, HTTP_1_1);
+
+    @Override
+    protected void configureServerBuilder(final HttpServerBuilder serverBuilder) {
+        serverBuilder.listenSocketOption(ServiceTalkSocketOptions.SO_BACKLOG, TCP_BACKLOG);
```
bondolo, line 66: Question rather than a comment: Would the backlog ever be disabled in a real application?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServerContext.java
```diff
@@ -50,6 +50,35 @@ default void awaitShutdown() {
...
+     * Depending on the transport, connections may still get ESTABLISHED, see.
+     * {@link ServiceTalkSocketOptions#SO_BACKLOG backlog} and (in case of Linux)
+     * <a href="https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt">SOMAXCONN</a>.
+     * For instance, in case of TCP the 3-way handshake may finish, and the connection will await in the
+     * accept queue to be accepted. If the accept queue is full, connection SYNs will await in the
+     * SYN backlog (i.e., in case of Linux <a href="https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt">
+     * tcp_max_syn_backlog</a>). These additional parameters may affect the behavior of new flows when the service
+     * is not accepting.
+     * <p>
+     * Depending on how long this stays in the {@code false} state, it may affect other timeouts (i.e., connect-timeout
```
bondolo, line 68: It sounds like this may make the behaviour worse for clients since they might wait until connect timeout to learn that the server is not accepting connections. Is it possible to adjust the backlog queues to zero while not accepting connections and reject the queued connections as if the queue had been full?

Not being able to fully reject the inbound connection seems bad.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
tkountis: This stops the server socket from accepting a connection, the rest of the behavior observed is an indirect result.
Example:
- If the socket is not accepted, then new connections will be put in the accept-queue up-to the backlog allowance. These connections will appear as established on the client side, and if they have an idle timeout they will expire.
- If the accept queue is full, the SYN attempts will be backlogged, meaning that the client will repeat SYN attempts and the connection won't establish. In such cases if a connection timeout exists on the client side will be respected.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptingNettyHttpServerTest.java
tkountis: It could, I guess soft realtime applications that want to avoid any latencies throughout the flow may disable it. That's not necessary bad, because it will allow for connection attempts to timeout.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServerContext.java
tkountis: Good observation. That indeed may be a problem, and this is the reason I am trying to document all expectations in the javadoc so users can make an educated decision. I want to avoid making a global decision here, and give the tools to the user's to tune accordingly.

chemicL: Excellent point. I was wondering whether it makes more sense to not expose such capability if we can't provide an extensive guideline how to configure the TCP settings. Wouldn't a mechanism for accepting the connection and immediately replying with a 503 HTTP error code be more "by the book"? For gRPC that's also safe as the clients should honor that code according to this table: https://grpc.github.io/grpc/core/md_doc_http-grpc-status-mapping.html

tkountis:
Build failure due: https://github.com/apple/servicetalk/issues/1579

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServerListenContext.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Context for controlling listen behavior.
+ */
+public interface ServerListenContext {
+    /**
+     * Toggles the server's channel accepting connection ability.
+     * <p>
+     * Passing a {@code false} value, will stop accepting connections on the server channel, without affecting any other
+     * interaction to currently open child channels (i.e., reads / writes).
```
idelpivnitskiy, line 26: Since this is `transport-api` module, let's reduce amount of details that are relevant only to netty implementation. Consider something that Java users will understand easier:
```
Toggles the server's ability to accept new connections.

Passing a {@code false} value will signal the server to stop accepting new connections. It won't affecting any other
interaction to currently open connections (i.e., reads / writes).
```

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServerListenContext.java
```diff
@@ -0,0 +1,50 @@
...
+     * <p>
+     * Depending on the transport, connections may still get ESTABLISHED, see.
```
idelpivnitskiy, line 28: `see.` -> `see`?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServerListenContext.java
```diff
@@ -0,0 +1,50 @@
...
+     * {@link ServiceTalkSocketOptions#SO_BACKLOG backlog} and (in case of Linux)
+     * <a href="https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt">SOMAXCONN</a>.
```
idelpivnitskiy, line 30: Why "in case of Linux" only? macOS/BSD also has SOMAXCONN, see
```
sysctl -a | grep kern.ipc.somaxconn
```

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServerListenContext.java
```diff
@@ -0,0 +1,50 @@
...
+     * For instance, in case of TCP the 3-way handshake may finish, and the connection will await in the
+     * accept queue to be accepted. If the accept queue is full, connection SYNs will await in the
+     * SYN backlog (i.e., in case of Linux <a href="https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt">
+     * tcp_max_syn_backlog</a>). These additional parameters may affect the behavior of new flows when the service
+     * is not accepting.
+     * <p>
+     * Depending on how long this stays in the {@code false} state, it may affect other timeouts (i.e., connect-timeout
+     * or idleness) on the peer-side and/or the other flows to the peer (i.e., proxies).
```
idelpivnitskiy, line 38: Hm... interesting that TCP 3-way handshake completes before a server actually accepts the connection. Client will see the `channelActive` event and may start sending data.
1. Is there any way to avoid completing TCP handshake until after the connection is accepted?
2. If not, what happens with data that client sends to a server if the server hasn't accepted a connection yet?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServerListenContext.java
```diff
@@ -0,0 +1,50 @@
...
+     * <p>
+     * Considerations:
+     * <ul>
+     * <li>Upon resumption, {@code accept == true}, backlogged connections will be processed first,
+     * which may be inactive by that time.</li>
```
idelpivnitskiy, line 43: How server will react in this case? Will it go through initialization of the whole pipeline and then close it immediately?
Consider enhancing `ServerBootstrap#handle` callback in `TcpServerBinder`:
https://github.com/apple/servicetalk/blob/808b31cb4d01162578fec0d5fc9f06846b48bbea/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java#L98-L115

We can check the status if the new `Channel`. If we can detect from there that it's already closed/inactive, we may discard it before adding to `ChannelSet` and initializing throigh `childHandler`.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ServiceTalkSocketOptions.java
```diff
@@ -28,19 +28,19 @@
...
     /**
-     * The connect timeout in milliseconds.
+     * Connect timeout in milliseconds.
      */
     public static final SocketOption<Integer> CONNECT_TIMEOUT =
             new ServiceTalkSocketOption<>("CONNECT_TIMEOUT", Integer.class);
 
     /**
-     * The threshold after which the the Endpoint is not writable anymore.
+     * The threshold after which the Endpoint is not writable anymore.
```
idelpivnitskiy, line 37: 👁🦅

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
idelpivnitskiy, line 79: This `channel()` is the actual HTTP/2 connection, you should take its `parent()`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptingNettyHttpServerTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        // Restoring auto-read will resume accepting.
+        serverContext().acceptConnections(true);
+        assertConnectionRequestSucceeds(request);
+    }
+
+    @Test
+    void testIdleTimeout() throws Exception {
+        setUp(CACHED, CACHED);
+        final StreamingHttpRequest request = reqRespFactory.newRequest(GET, SVC_ECHO);
```
idelpivnitskiy, line 93: Instead of a custom `reqRespFactory`, you can use `streamingHttpClient()` or `streamingHttpConnection()`.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptingNettyHttpServerTest.java
```diff
@@ -0,0 +1,131 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class ConnectionAcceptingNettyHttpServerTest extends AbstractNettyHttpServerTest {
+
+    private static final boolean IS_LINUX = "linux".equals(normalizedOs());
+    // There is an off-by-one behavior difference between macOS & Linux.
+    // Linux has a greater-than check
+    // (see. https://github.com/torvalds/linux/blob/5bfc75d92efd494db37f5c4c173d3639d4772966/include/net/sock.h#L941)
+    private static final int TCP_BACKLOG = IS_LINUX ? 0 : 1;
+    private static final int CONNECT_TIMEOUT_MILLIS = (int) ofSeconds(5).toMillis();
```
idelpivnitskiy, line 60: Can we lower this value to 1 second or less? Not ideal that the test takes so long to complete.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptingNettyHttpServerTest.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private void assertConnectionRequestReceiveTimesOut(final StreamingHttpRequest request) {
+        final StreamingHttpClient httpClient = newClient();
+        assertThrows(TimeoutException.class, () -> await(httpClient.request(request), TRY_REQUEST_AWAIT_SECS, SECONDS));
+    }
+
+    private StreamingHttpClient newClient() {
+        return newClientBuilder()
+                .appendConnectionFactoryFilter(withMax(1))
```
idelpivnitskiy, line 117: We are leaking resources if we create new clients and never close them. `AbstractNettyHttpServerTest` already has a client. Instead of creating new clients, you can take the existing one using `streamingHttpClient()` and use its API to open new connections when you need them: `client.reserveConnection(...)`. That way, `AbstractNettyHttpServerTest` will take care of properly closing all resouces when a test completes.

To test the difference between `toFuture().get(...)` timeout and `CONNECT_TIMEOUT_MILLIS`, just use lower duration for `toFuture().get(...)`: half a sec vs second.

idelpivnitskiy (changes requested):

Issue #1684, opened 2021-07-21
jUnit5 will never stop a hanging test if it awaits uninterruptibly


idelpivnitskiy:
jUnit5 uses thread interrupt to stop hanging tests after the timeout expires. Some of our tests ignore InterruptedException by using:
- AwaitUtils
- netty’s syncUninterruptibly()
- awaitRequestNUninterruptibly
- awaitCancelledUninterruptibly

As the result, these tests may hang forever and jUnit5 won’t be able to stop them.
We should not ignore `InterruptedException` or run something uninterruptibly while testing. All these tests should propagate  `InterruptedException` and declare that on the method API. For netty, we should use `sync()` instead of `syncUninterruptibly()`.

Krupskis:
Hey @idelpivnitskiy, is this still available? I would love to explore it and try to fix it.

idelpivnitskiy:
Hi @Krupskis,
Thanks for the interest. Yes, the issue is still open and to my knowledge nobody started to work on it.

idelpivnitskiy:
@Krupskis I believe we are almost done here. As the last step, can you please go through all cases when we catch `InterruptedException` in tests to make sure we are consistent? I see a few places where we forgot to do `Thread.currentThread().interrupt()` and when we wrap `InterruptedException` as `RuntimeException` (which may prevent junit from recognizing it correctly).

idelpivnitskiy:
Great work @Krupskis! Thanks for your effort and consistent PRs!

PR #1744, opened 2021-08-19
Propogate InterruptedException to the NettyPipelinedConnectionTest tests

Krupskis:
Motivation:

The current usage of `awaitCancelledUninterruptibly()` catches
`InterruptedException` and silently discards it. jUnit5 uses
result, these tests may hang forever as the `InterruptedException`
is not propagated.

Modifications:

- use `awaitCancelled()`, which throws `InterruptedException`
- Declare `InterruptedException` on method declaration.

Result:

The `InterruptedException` is propagated to the test methods,
which allows jUnit5 to stop a hanging test when the test's
timeout expires.

This is related to #1684. The plan for the future is to propagate
`InterruptedException` by changing the usage of the other methods 
described in the issue. This is my first contribution and I am not
100% sure if I understood the issue correctly, so please correct me
if my assumptions on the issue are wrong.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/NettyPipelinedConnectionTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/NettyPipelinedConnectionTest.java
@@ -259,7 +259,7 @@ class NettyPipelinedConnectionTest {
     }
 
     @Test
-    void readCancelErrorsPendingReadCancelsPendingWrite() {
+    void readCancelErrorsPendingReadCancelsPendingWrite() throws Exception {
         TestSubscription writePublisher1Subscription = new TestSubscription();
         toSource(requester.write(writePublisher1
                 .afterSubscription(() -> writePublisher1Subscription))).subscribe(readSubscriber);
@@ -274,13 +274,13 @@ class NettyPipelinedConnectionTest {
         // should be terminated, cancelled, or not subscribed.
 
         assertThat(readSubscriber2.awaitOnError(), is(instanceOf(ClosedChannelException.class)));
-        writePublisher1Subscription.awaitCancelledUninterruptibly();
+        writePublisher1Subscription.awaitCancelled();
         assertFalse(writePublisher2.isSubscribed());
         assertFalse(channel.isOpen());
     }
 
     @Test
-    void channelCloseErrorsPendingReadCancelsPendingWrite() {
+    void channelCloseErrorsPendingReadCancelsPendingWrite() throws Exception {
         TestSubscription writePublisher1Subscription = new TestSubscription();
         toSource(requester.write(writePublisher1
                 .afterSubscription(() -> writePublisher1Subscription))).subscribe(readSubscriber);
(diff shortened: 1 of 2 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/NettyPipelinedConnectionTest.java
```diff
@@ -274,13 +274,13 @@ void readCancelErrorsPendingReadCancelsPendingWrite() {
         // should be terminated, cancelled, or not subscribed.
 
         assertThat(readSubscriber2.awaitOnError(), is(instanceOf(ClosedChannelException.class)));
-        writePublisher1Subscription.awaitCancelledUninterruptibly();
+        writePublisher1Subscription.awaitCancelled();
```
idelpivnitskiy, line 277: I would also suggest to mark `awaitCancelledUninterruptibly()` as `@Deprecated`,  and reference that `awaitCancelled()` should be used instead. Since `awaitCancelledUninterruptibly()` is problematic for jUnit5, it will be good to remove that method in future releases.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/NettyPipelinedConnectionTest.java
```diff
@@ -295,12 +295,12 @@ void channelCloseErrorsPendingReadCancelsPendingWrite() {
...
         assertThat(readSubscriber.awaitOnError(), is(instanceOf(ClosedChannelException.class)));
         assertThat(readSubscriber2.awaitOnError(), is(instanceOf(ClosedChannelException.class)));
-        writePublisher1Subscription.awaitCancelledUninterruptibly();
+        writePublisher1Subscription.awaitCancelled();
         assertFalse(writePublisher2.isSubscribed());
     }
 
     @Test
-    void readCancelClosesConnectionThenWriteDoesNotSubscribe() {
+    void readCancelClosesConnectionThenWriteDoesNotSubscribe() throws InterruptedException {
```
idelpivnitskiy, line 303: Not a problem, but because you will anyway address the previous comment, I will ask you to change `InterruptedException` to `Exception` here. For tests it's common that they may throw multiple checked exceptions. It's easier to use `Exception` instead of declaring each one independently. Makes tests easier to read.

idelpivnitskiy (commented):
Thank you for the PR @Krupskis!

Krupskis:
@idelpivnitskiy Thanks for the review. I pushed the new changes regarding your comments. Is it better for you if I amend changes to an existing commit or create a new one?

For next PR regarding the issue, should I create a pull request for each method or try to fix as much as possible in one PR?

idelpivnitskiy (approved):
Thank you so much @Krupskis!

> Is it better for you if I amend changes to an existing commit or create a new one?

Prefer to add new commits in the same branch and push to the existing PR. New commits help to review new changes only. If you squash and force-push, I have to re-review the whole PR (pain if the PR is big) to make sure you didn't change anything else that might be concerning.

> For next PR regarding the issue, should I create a pull request for each method or try to fix as much as possible in one PR?

Whatever approach works better for you. If you like more small PRs, you are welcome to do so. If you prefer to have the whole picture in one PR, that will also work for us. Any contribution is more than welcome ❤️

Krupskis:
@idelpivnitskiy is it possible to get write access to the repository so that I can merge the pull request or how is this handled here?

tkountis:
Thanks for the contribution @Krupskis 🎉 
We can merge the PRs once they are reviewed no worries about that.

PR #1425, opened 2021-03-11
Verify that `HttpHeaders` impl for HTTP/2 does not throw `NullPointerException`

idelpivnitskiy:
Motivation:

Netty had a bug that was fixed in https://github.com/netty/netty/pull/11028.
It could produce `NullPointerException` if two different headers end up in the
same hash bucket and you use the iterator to remove the first element returned
from the iterator. Our `H2ToStH1Utils.h1HeadersSplitCookieCrumbs` used this
scenario. Adding tests to guarantee our utility works correctly.

Modifications:

- Test `H2ToStH1Utils.h1HeadersSplitCookieCrumbs` when two headers are in
the same hash bucket;

Result:

Confidence in `H2ToStH1Utils.h1HeadersSplitCookieCrumbs`.

PR #1748, opened 2021-08-23
Improve split cookie crumbs test in `H2ToStH1UtilsTest`

idelpivnitskiy:
Motivation:

We used `ConstantHashCharSequence` to make sure that two headers fall
into the same bucket in internal `HttpHeaders` data structure. However,
the `hashCode()` is taken only if `CharSequence` implements
`AsciiBuffer` or `AsciiString`. Otherwise, a static utility is used
to compute case-insensitive `hashCode`. As the result,
`ConstantHashCharSequence` does not provide what we expected.

Modifications:

- Remove `ConstantHashCharSequence` and use `caseInsensitiveHashCode` /
`hashCodeAscii` utilities to make sure that two header names result
in the same bucket based on their `hashCode` value and `ARRAY_SIZE_HINT`

Result:

Test verifies the behavior it was created for.

This test was introduced in #1425.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ToStH1UtilsTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ToStH1UtilsTest.java
@@ -26,32 +26,49 @@ import org.junit.jupiter.api.Test;
 import java.util.ArrayList;
 import java.util.List;
 
+import static io.netty.util.internal.PlatformDependent.hashCodeAscii;
+import static io.servicetalk.buffer.api.CharSequences.caseInsensitiveHashCode;
 import static io.servicetalk.http.api.HttpHeaderNames.ACCEPT_PATCH;
 import static io.servicetalk.http.api.HttpHeaderNames.COOKIE;
-import static io.servicetalk.http.api.HttpHeaderValues.TEXT_PLAIN;
+import static io.servicetalk.http.api.HttpHeaderNames.EXPIRES;
 import static io.servicetalk.http.netty.H2ToStH1Utils.h1HeadersSplitCookieCrumbs;
-import static java.util.Objects.requireNonNull;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasSize;
 
 class H2ToStH1UtilsTest {
 
+    private static final int ARRAY_SIZE_HINT = 16;
+    private static final HttpHeadersFactory H1_FACTORY = new DefaultHttpHeadersFactory(true, true,
+            ARRAY_SIZE_HINT, 0);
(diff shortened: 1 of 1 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ToStH1UtilsTest.java
```diff
@@ -26,32 +26,49 @@
...
-    void testH1HeadersSplitCookieCrumbs(HttpHeadersFactory headersFactory) {
+    void testH1HeadersSplitCookieCrumbs(HttpHeadersFactory headersFactory, CharSequence secondHeaderName) {
         HttpHeaders headers = headersFactory.newHeaders();
         // Add two headers which will be saved in the same entries[index]:
-        headers.add(new ConstantHashCharSequence(COOKIE), "a=b; c=d; e=f");
-        headers.add(new ConstantHashCharSequence(ACCEPT_PATCH), TEXT_PLAIN);
+        headers.add(COOKIE, "a=b; c=d; e=f");
+        String secondHeaderValue = "some-value";
+        headers.add(secondHeaderName, secondHeaderValue);
         h1HeadersSplitCookieCrumbs(headers);
```
bondolo, line 73: Could also ensure that the value of `secondHeaderName` is still `secondHeaderValue`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ToStH1UtilsTest.java
```diff
@@ -26,32 +26,49 @@
...
+        assertThat(bucketIndex(caseInsensitiveHashCode(COOKIE)),
+                equalTo(bucketIndex(caseInsensitiveHashCode(secondHeaderName))));
+        testH1HeadersSplitCookieCrumbs(H1_FACTORY, secondHeaderName);
     }
 
     @Test
     void testH1HeadersSplitCookieCrumbsForH2Headers() {
-        testH1HeadersSplitCookieCrumbs(H2HeadersFactory.INSTANCE);
+        CharSequence secondHeaderName = ACCEPT_PATCH;
+        assertThat(bucketIndex(hashCodeAscii(COOKIE)), equalTo(bucketIndex(hashCodeAscii(secondHeaderName))));
```
bondolo, line 62: This assertion could be moved in to the `testH1HeadersSplitCookieCrumbs` as the test is invalid if `secondHeaderName` does not match `COOKIE`.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ToStH1UtilsTest.java
idelpivnitskiy: It's ensured in line 82, there are no changes for headers between 72 and 82

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ToStH1UtilsTest.java
idelpivnitskiy: `secondHeaderName` and the hashcode function are different between h1 and h2, therefore they are validated before the shared logic can be applied

PR #1750, opened 2021-08-25
Deprecate one-way boolean settings on builders

chemicL:
Motivation:

Some builder APIs allow disabling some feature without the option to
enable it. To keep the builders' APIs consistent all boolean settings
should accept a flag that can trigger the feature to on or off
configuration. This change deprecates the undesired methods and adds
methods that accept a boolean.

Modifications:

- Deprecated `GrpcClientBuilder.disableHostHeaderFallback` in favor of
  `GrpcClientBuilder.hostHeaderFallback(boolean)`,
- Deprecated `GrpcClientBuilder.disableDrainingRequestPayloadBody` in
  favor of `GrpcClientBuilder.drainRequestPayloadBody(boolean)`,
- Deprecated `HttpReporter.Builder.disableSpanBatching` in favor of
  `HttpReporter.Builder.spansBatchingEnabled(boolean)`,
- Deprecated `HttpServerBuilder.disableDrainingRequestPayloadBody` in
  favor of `HttpServerBuilder.drainRequestPayloadBody(boolean)`,
- Deprecated `SingleAddressHttpClientBuilder.disableHostHeaderFallback`
  in favor of
  `SingleAddressHttpClientBuilder.hostHeaderFallback(boolean)`.

Result:

More consistent builders' APIs.


servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -246,11 +246,11 @@ public Builder maxConcurrentReports(final int maxConcurrentReports) {
          * @param maxBatchDuration {@link Duration} of time to wait for {@code batchSizeHint} spans in a batch.
          * @return {@code this}.
          */
-        public Builder batchSpans(final int batchSizeHint, final Duration maxBatchDuration) {
```
Scottmitch, line 249: can we first deprecate the `batchSpans(int, Duration)` method, and add a new method `spansBatchingEnabled(int, Duration)` to reduce impacts of API breakage?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
```diff
@@ -270,11 +270,25 @@
      * <p>
      * This setting disables the default filter such that no {@code Host} header will be manipulated.
      *
+     * @deprecated Use {@link #hostHeaderFallback(boolean)}.
```
Scottmitch, line 273: note with updated Checkstyle rules the `@deprecated` javadoc tag must be the last tag.

Scottmitch (approved):
lgtm after javadoc fix and rebase

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
```diff
@@ -270,11 +270,25 @@
...
+     * Configures automatically setting {@code Host} headers by inferring from the address or {@link HttpMetaData}.
+     * <p>
+     * When {@code false} is passed, this setting disables the default filter such that no {@code Host} header will be
+     * manipulated.
+     *
+     * @param enable Whether a default filter for inferring the {@code Host} headers should be added.
+     * @return {@code this}
+     * @see #unresolvedAddressToHost(Function)
+     */
+    public abstract GrpcClientBuilder<U, R> hostHeaderFallback(boolean enable);
```
idelpivnitskiy, line 290: Adding a new abstract method will be a breaking change for 0.41 branch. Consider adding the default implementation that throws `UnsupportedOperationException` in this PR. After cherry-picking this into 0.41, please submit a follow-up for the `main` branch to remove reprecated methods and make new methods `abstract`. We followed the similar approach here: https://github.com/apple/servicetalk/pull/1781#discussion_r701419264

idelpivnitskiy (changes requested):

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
```diff
@@ -281,9 +281,26 @@
      *
      * @return {@code this}
      * @see #unresolvedAddressToHost(Function)
+     * @deprecated Use {@link #hostHeaderFallback(boolean)}.
      */
+    @Deprecated
     public abstract GrpcClientBuilder<U, R> disableHostHeaderFallback();
```
idelpivnitskiy, line 287: As we discussed offline, let's also add default implementations for the abstract methods we deprecate?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
```diff
@@ -281,8 +281,27 @@
...
+     * When {@code false} is passed, this setting disables the default filter such that no {@code Host} header will be
+     * manipulated.
+     *
+     * @param enable Whether a default filter for inferring the {@code Host} headers should be added.
+     * @return {@code this}
+     * @see #unresolvedAddressToHost(Function)
+     */
+    public GrpcClientBuilder<U, R> hostHeaderFallback(boolean enable) {
+        throw new UnsupportedOperationException("Setting automatic host header fallback using this method" +
+                " is not yet supported.");
```
idelpivnitskiy, line 303: In the other PR @bondolo used `getClass().getSimpleName()`: https://github.com/apple/servicetalk/pull/1781/files#diff-fa20d1d978b59de34bb9649cc6cc21ba5321886ccaa29636bbc285eea14eb07aR332
Consider using `getClass().getName()` moving forward. It will help users understand who they should contact if this method is not implemented.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -494,8 +494,8 @@ private static StreamingHttpConnectionFilterFactory appendConnectionFilter(
     }
 
     @Override
-    public DefaultSingleAddressHttpClientBuilder<U, R> disableHostHeaderFallback() {
-        addHostHeaderFallbackFilter = false;
+    public SingleAddressHttpClientBuilder<U, R> hostHeaderFallback(final boolean enable) {
```
idelpivnitskiy, line 497: Why the return type have changed? Consider being consistent with all other methods that return `DefaultSingleAddressHttpClientBuilder`. If you prefer to keep `SingleAddressHttpClientBuilder`, consider doing that in a follow-up (when you convert abstract classes into interfaces) for all methods.

idelpivnitskiy (approved):
Last comments than LGTM:

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
chemicL: I didn't intend to change anything. As a matter of fact, this is a different, newly added method. However, the previous method now retains the parent return type, as it's not overridden here. This implementation of the new method was auto generated by IDE, I didn't put much thought into it. Your comment inspired me to look at other methods in this class. Unfortunately I don't see any consistency, e.g.:

- `public SingleAddressHttpClientBuilder<U, R> executionStrategy(final HttpExecutionStrategy strategy)`
- `public DefaultSingleAddressHttpClientBuilder<U, R> executor(final Executor executor)`

I believe that:
- because the class is package private (so users can't see the actual return types as declared here),
- and we might want to provide pre-baked builders with calls to methods that could be specific to this particular builder,

the return type should be the actual protected class, not the public superclass (soon to be interface) to avoid the need to cast in such case. Perhaps after the PRs for getting rid of abstract classes in favor of interface we could clean it up or leave for when needed, as it is in fact internal.

PR #1751, opened 2021-08-25
Proxy address client using `MultiAddressHttpClientBuilder`

chemicL:
Motivation:

Using `MultiAddressHttpClientBuilder` it's currently not possible to set a proxy. Users of this API should also be able to specify a proxy for each address that the client connects to.

Modifications:

- Added `SingleAddressHttpClientBuilder#proxyAddress` to allow specifying proxy on the builder,
- Simplified internal logic of mapping specified `host` to a `CharSequence` in `DefaultSingleAddressHttpClientBuilder`,
- Introduced a test for the feature: `HttpMultiProxyTest`.

Result:

It's possible to specify a proxy for each individual address that clients created using `MultiAddressHttpClientBuilder` can reach.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -240,7 +236,9 @@ private DefaultSingleAddressHttpClientBuilder(@Nullable final U address,
 
         U address() {
             assert builder.address != null : "Attempted to buildStreaming with an unknown address";
-            return proxyAddress != null ? proxyAddress : builder.address;
+            return proxyAddress != null ? proxyAddress :
+                    // the builder can be modified post-context creation, therefore proxy can be set
+                    (builder.proxyAddress != null ? builder.proxyAddress : builder.address);
```
Scottmitch, line 241: Referencing `builder.proxyAddress` is unexpected. The motivation of `HttpClientBuildContext` is to capture mutable the state of the builder, so it can be re-used in subsequent `build()` without state unexpectedly changing. Note that `builder.address` is `final` so it will not change, and IIUC that is why it is referenced here.

chemicL: From my observations it looks like the assumption here is that the `builder` is **already a copy** owned by the context ([source](https://github.com/apple/servicetalk/blob/main/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java#L412)). The codebase uses that fact with the concept of `MultiAddressHttpClientBuilder.SingleAddressInitializer` where the cloned builder is saved during the `ClientFactory` creation ([source](https://github.com/apple/servicetalk/blob/main/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java#L96)) and later modified for each individually built client ([source](https://github.com/apple/servicetalk/blob/main/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java#L225-L244)).

Finally, those changes on the copied builder which is embedded as part of the captured `HttpClientBuildContext` (and may, as encouraged by current design, be later modified) are used to build the actual underlying client ([source](https://github.com/apple/servicetalk/blob/main/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java#L271-L354)).

Perhaps the design isn't ideal due to the entangled nature of the context:
- immutable: created from a mutable builder using a copy method,
- contains a mutable copied builder that is expected to be modified,
- needs to be reconciled during build time with the mutable state in order to achieve the flexibility required by the multi/partitioned builders.

This issue could be subject of a different PR, but all in all it feels that the end goal of preventing concurrent access is taken care of.

bondolo: > preventing concurrent access 

It looks like our current usage is safe, but do we need to be more clear throughout the ST API about which instances can be shared between threads and which are non-threadsafe? Are users likely to be unsure which objects they can share? Builders are typically not thread safe as they are not generally meant to be shared. The "copy-on-write" solution used here for a partially complete builder seems OK.

chemicL: I would love the APIs to expose something like https://jcip.net/annotations/doc/net/jcip/annotations/NotThreadSafe.html, but I think it's out of scope of this PR.

With regards to our case: as long as ` io.servicetalk.client.api.DefaultClientGroup#get` is thread safe, we do not have an issue. Users also need to make sure they don't reuse the builder from multiple threads, but being super explicit about it in the documentation would be beneficial in my opinion.

chemicL: Correct, I added a test for it.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
```diff
@@ -51,6 +51,13 @@
  */
 public abstract class SingleAddressHttpClientBuilder<U, R>
         implements HttpClientBuilder<U, R, ServiceDiscovererEvent<R>> {
+    /**
+     * Configure proxy to serve as an intermediary for requests.
+     * @param proxyAddress Unresolved address of the proxy.
+     * @return {@code this}.
+     */
+    public abstract SingleAddressHttpClientBuilder<U, R> proxyAddress(U proxyAddress);
```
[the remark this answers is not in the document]
chemicL: That makes sense. I followed up with another commit for these suggestions.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
```diff
@@ -56,7 +56,10 @@
      * @param proxyAddress Unresolved address of the proxy.
      * @return {@code this}.
      */
-    public abstract SingleAddressHttpClientBuilder<U, R> proxyAddress(U proxyAddress);
+    public SingleAddressHttpClientBuilder<U, R> proxyAddress(U proxyAddress) {
+        throw new UnsupportedOperationException("Setting proxy address is not yet supported by "
+                + getClass().getSimpleName());
```
idelpivnitskiy, line 61: Consider using `getName()` to let users know who is the owner of the class

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpsProxyTest.java
```diff
@@ -101,7 +101,8 @@ void startServer() throws Exception {
     void createClient() {
         assert serverAddress != null && proxyAddress != null;
         client = HttpClients
-                .forSingleAddressViaProxy(serverAddress, proxyAddress)
+                .forSingleAddress(serverAddress)
```
idelpivnitskiy, line 104: For `HttpsProxyTest` and `HttpProxyTest`, can we parametrize at least one test to verify that `forResolvedAddress` with proxy also works?

idelpivnitskiy (approved):
Nice!

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpsProxyTest.java
chemicL: Interesting that you ask. I started parametrizing the tests and was surprised to discover the resolved version did not work:

```
Unexpected address resolution request: 127.0.0.1:55937
java.lang.IllegalArgumentException: Unexpected address resolution request: 127.0.0.1:55937
	at io.servicetalk.http.netty.DefaultSingleAddressHttpClientBuilder$NoopServiceDiscoverer.discover(DefaultSingleAddressHttpClientBuilder.java:667)
```

As a result, I tried to parameterize the test on the `main` branch and it also fails when using `forResolvedAddressViaProxy` in place of `forSingleAddressViaProxy`. The reason is that the `NoopServiceDiscoverer` expects the discovered address to be the same as the address configured originally, while the discovered address is in fact the address of the proxy and not the eventual target.

Do we really need the verification that causes the failure? https://github.com/apple/servicetalk/blob/main/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java#L665-L667 <- Perhaps we can get rid of this validation?

chemicL: Please see my latest commit. I was really surprised the proxy doesn't work with resolved addresses even without my changes. So I made an attempt to fix it. I got rid of an optimization for caching the `Publisher` in the `NoopServiceDiscoverer` in the constructor and return a new instance upon each `discover`, but I don't think it would cause any performance hit as it's supposed to be executed once upon client initialization. Internal APIs needed to be change to facilitate the feature, but now it's possible to use a proxy in any case (except for an unresolved proxy address when the client was created for a resolved address -> for that I added a javadoc note).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
```diff
@@ -206,11 +217,16 @@ private HttpClients() {
      * {@link HttpHeaderNames#HOST}. Use {@link SingleAddressHttpClientBuilder#unresolvedAddressToHost(Function)}
      * if you want to override that value or {@link SingleAddressHttpClientBuilder#hostHeaderFallback(boolean)} if you
      * want to disable this behavior.
+     * <p>
+     * Note, if {@link SingleAddressHttpClientBuilder#proxyAddress(Object) a proxy} is configured for this client,
```
chemicL, line 221: This information was present in the deprecated methods. Unfortunately, they didn't work even for resolved proxy addresses. This is an attempt to warn the users about the expected usage of the returned builder.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpsProxyTest.java
idelpivnitskiy: Agreed, it should not affect perf. Thanks for fixing it!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
```diff
@@ -50,6 +50,16 @@
  * @param <R> the type of address after resolution (resolved address)
  */
 public interface SingleAddressHttpClientBuilder<U, R> extends HttpClientBuilder<U, R, ServiceDiscovererEvent<R>> {
+    /**
+     * Configure proxy to serve as an intermediary for requests.
+     * @param proxyAddress Unresolved address of the proxy.
```
idelpivnitskiy, line 55: Should we clarify here what should be used when a builder is created for resolved address?

chemicL: Makes sense. Although it's an implementation choice so the wording will include "may" instead of "will" for the note about runtime exceptions.

chemicL:
0.41: 349df1e316a52de037b90b5211182bf6ca9f19d2

PR #1756, opened 2021-08-26
Deprecate syncUninterruptibly() and awaitRequestNUninterruptibly() in tests

Krupskis:
Motivation:
Currently netty's syncUninterruptibly() and
awaitRequestNUninterruptibly() do not
propogate `InterruptedException`, meaning that jUnit5 is not
able to stop timeouted tests.

Modifications:
- change the uses of `syncUninterruptibly()` to `sync()`
- on `safeSync()` use `Callable` instead of `Runnable`, because
`Runnable` is unable to throw exception;
- Change all usages of `awaitRequestNUninterruptibly()`
to `awaitRequestN()`;
- add `@Deprecated` to `awaitRequestNUninterruptibly()`;

Result:
`InterruptedException` is no longer silently discarded,
meaning jUnit5 will be able to abort the timeouted test.

Related to #1684.

Questions from my side:
1. Why the N in awaitRequestN()?
2. Doesn’t safeClose() in HttpsProxyTest also not 
Propagate the InterruptedException, which might 
result in timeouts as well?
3. I couldn't figure out what the function of 
safeSync(), so I changed it to use Callable,
could someone explain what it does in 
a bit more in depth?

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamObserverTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamObserverTest.java
@@ -47,6 +47,7 @@ import org.junit.jupiter.api.Test;
 
 import java.net.InetSocketAddress;
 import java.nio.channels.ClosedChannelException;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 
@@ -83,7 +84,7 @@ class StreamObserverTest {
     private final HttpClient client;
     private final CountDownLatch requestReceived = new CountDownLatch(1);
 
-    StreamObserverTest() {
+    StreamObserverTest() throws Exception {
         clientTransportObserver = mock(TransportObserver.class, "clientTransportObserver");
         clientConnectionObserver = mock(ConnectionObserver.class, "clientConnectionObserver");
         clientMultiplexedObserver = mock(MultiplexedObserver.class, "clientMultiplexedObserver");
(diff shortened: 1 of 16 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamObserverTest.java
```diff
@@ -122,15 +123,17 @@ protected void channelRead0(final ChannelHandlerContext ctx, final Object msg) {
...
         safeClose(client);
     }
 
-    static void safeSync(Runnable runnable) {
+    static void safeSync(Callable<Object> callable) throws Exception {
         try {
-            runnable.run();
+            callable.call();
+        } catch (InterruptedException e) {
+            throw e;
```
idelpivnitskiy, line 136: > 2. Doesn’t safeClose() in HttpsProxyTest also not Propagate the InterruptedException, which might result in timeouts as well?

Good catch! You are right. Also, before re-throwing the `InterruptedException`, we should reset `interrupted` flag:
```java
throw e;
```

idelpivnitskiy (commented):
Great change! Thank you @Krupskis!
One comment, then LGTM.

Remaining answers:
> 1. Why the N in awaitRequestN()?

Reactive Streams specification defines a method `void request(long n)` for [Subscription](https://github.com/reactive-streams/reactive-streams-jvm#3-subscription-code).
The word "request" is used in different contexts, we decided to add N at the end to highlight it's related to requesting N elements for Reactive Streams. Just found this convenient, no other reasons.


> 3. I couldn't figure out what the function of `safeSync()`, so I changed it to use `Callable`, could someone explain what it does in a bit more in depth?

Good question! Let's take an example from `StreamObserverTest`:
```java
    @AfterEach
    void teardown() throws Exception {
        safeSync(() -> serverAcceptorChannel.close().sync());
        safeSync(() -> serverEventLoopGroup.shutdownGracefully(0, 0, MILLISECONDS).sync());
        safeClose(client);
    }
```
Without `safeSync`, if the first line throws, we will never close `serverEventLoopGroup` and `client`. `safeSync` helps to make sure we continue teardown even if any of the lines throw an exeption.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamObserverTest.java
Krupskis: Good catch, forgot that it resets the value :)

Krupskis:
@idelpivnitskiy thanks for the answers, all is clear now.

idelpivnitskiy:
Thank you @Krupskis!

Issue #1759, opened 2021-08-27
ConnectionAcceptingNettyHttpServerTest test failure


Scottmitch:
https://github.com/apple/servicetalk/runs/3440129511

```
ConnectionAcceptingNettyHttpServerTest.testStopAcceptingAndResume()

java.util.concurrent.TimeoutException: Timed out waiting for the result
	at io.servicetalk.concurrent.api.SourceToFuture.get(SourceToFuture.java:107)
	at io.servicetalk.concurrent.api.BlockingTestUtils.await(BlockingTestUtils.java:171)
	at io.servicetalk.http.netty.ConnectionAcceptingNettyHttpServerTest.assertConnectionRequestSucceeds(ConnectionAcceptingNettyHttpServerTest.java:126)
	at io.servicetalk.http.netty.ConnectionAcceptingNettyHttpServerTest.testStopAcceptingAndResume(ConnectionAcceptingNettyHttpServerTest.java:93)
```

Scottmitch:
https://github.com/apple/servicetalk/runs/3467797640

```
ConnectionAcceptingNettyHttpServerTest.testStopAcceptingAndResume()

java.util.concurrent.TimeoutException: Timed out waiting for the result
	at io.servicetalk.concurrent.api.SourceToFuture.get(SourceToFuture.java:107)
	at io.servicetalk.concurrent.api.BlockingTestUtils.await(BlockingTestUtils.java:171)
	at io.servicetalk.http.netty.ConnectionAcceptingNettyHttpServerTest.assertConnectionRequestSucceeds(ConnectionAcceptingNettyHttpServerTest.java:126)
	at io.servicetalk.http.netty.ConnectionAcceptingNettyHttpServerTest.testStopAcceptingAndResume(ConnectionAcceptingNettyHttpServerTest.java:93)
```

Scottmitch:
https://github.com/apple/servicetalk/pull/1776/checks?check_run_id=3478774871

```
ConnectionAcceptingNettyHttpServerTest > testStopAcceptingAndResume() FAILED

    java.util.concurrent.TimeoutException: Timed out waiting for the result
        at io.servicetalk.concurrent.api.SourceToFuture.get(SourceToFuture.java:107)
        at io.servicetalk.concurrent.api.BlockingTestUtils.await(BlockingTestUtils.java:171)
[...]

PR #1767, opened 2021-08-31
ZipkinPublisher remove unnecessary data

Scottmitch:
Motivation:
We currently publish a time stamp for begin, a duration, and a timestamp
for end. We can remove the end timestamp which can be derived by doing
begin+end. We also publish a tag for the type of span, but there is also
an explicit field for the type.

Modifications:
- Remove the tag from the span before publishing, remove end timestamp
  in ZipkinPublisher.

Result:
Less unnecessary data published from ZipkinPublisher.


Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1299

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1759

tkountis (approved):

idelpivnitskiy (approved):

Scottmitch:
since there is a change of behavior, I will skip backport to 0.41.x

PR #1772, opened 2021-08-31
Move isIoThreadSupported to IoExecutor

bondolo:
Motivation:
The `isIoThreadSupported` method was recently added to the
`NettyIoExecutor` as a mechanism for determining if the threads of the
`IoExecutor` were marked using the `IoThreadfactory.IoThread`
interface. It is useful to know this for all `IoExecutors`.
Modifications:
Move `isIoThreadSupported` from `NettyIoExecutor` the parent interface
`IoExecutor`
Result:
More situations can use conditional offloading by querying
`IoExecutor`s whether their threads support IoThread identification.


idelpivnitskiy (approved):
Consider looking at places that already use this flag to see if we can avoid `instanceof NettyIoExecutor` and cast. Feel free to merge after that.

PR #1778, opened 2021-09-01
Increase memory for quality build tasks

Scottmitch:
Motivation:
checkstyle 9 on JDK 8 has been observed to exhaust memory and fail
builds. We don't tune the default JVM arguments and should increase
memory to give quality build tasks more memory headroom.


Scottmitch:
I will merge this to restore the build to healthy.

PR #1781, opened 2021-09-02
`ExecutionStrategy` simplification deprecations

bondolo:
Motivation:
removed methods should be deprecated before they are removed.
Modifications:
Marks methods to be removed in #1695 as deprecated and, where possible,
suggests alternatives.
Result:
Warnings for future method removals are provided.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
@@ -45,7 +45,8 @@ public final class GrpcExecutionStrategies {
      *
      * @param executor {@link Executor} to use.
      * @return Default {@link GrpcExecutionStrategy}.
-     * @deprecated Set the executor to use on the {@link io.servicetalk.transport.api.ExecutionContext}.
+     * @deprecated Set the executor to use on the {@link io.servicetalk.transport.api.ExecutionContext} using
+     * {@code executor(Executor)} methods on client/server builders.
      */
     @Deprecated
     public static GrpcExecutionStrategy defaultStrategy(final Executor executor) {
@@ -135,7 +136,8 @@ public final class GrpcExecutionStrategies {
          *
          * @param executor {@link Executor} to use.
          * @return {@code this}.
-         * @deprecated Set the executor to use on the {@link io.servicetalk.transport.api.ExecutionContext}.
+         * @deprecated Set the executor to use on the {@link io.servicetalk.transport.api.ExecutionContext} using
+         * {@code executor(Executor)} methods on client/server builders.
          */
         @Deprecated
         public Builder executor(Executor executor) {
@@ -147,8 +149,8 @@ public final class GrpcExecutionStrategies {
          * Enable thread affinity while offloading. When enabled, offloading implementation will favor using a
          * single thread per subscribe of a source.
          * @return {@code this}.
-         * @deprecated Use a single threaded executor set on {@link io.servicetalk.transport.api.ExecutionContext} to
-         * ensure affinity.
+         * @deprecated Use a single threaded executor set on {@link io.servicetalk.transport.api.ExecutionContext} using
+         * {@code executor(Executor)} methods on client/server builders to ensure affinity.
          */
         @Deprecated
         public Builder offloadWithThreadAffinity() {
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
@@ -45,7 +45,9 @@ public final class GrpcExecutionStrategies {
      *
      * @param executor {@link Executor} to use.
      * @return Default {@link GrpcExecutionStrategy}.
+     * @deprecated Set the executor to use on the {@link io.servicetalk.transport.api.ExecutionContext}.
      */
+    @Deprecated
     public static GrpcExecutionStrategy defaultStrategy(final Executor executor) {
         return new DefaultGrpcExecutionStrategy(HttpExecutionStrategies.defaultStrategy(executor));
     }
@@ -133,7 +135,9 @@ public final class GrpcExecutionStrategies {
          *
          * @param executor {@link Executor} to use.
          * @return {@code this}.
+         * @deprecated Set the executor to use on the {@link io.servicetalk.transport.api.ExecutionContext}.
          */
+        @Deprecated
         public Builder executor(Executor executor) {
             httpBuilder.executor(executor);
             return this;
@@ -143,7 +147,8 @@ public final class GrpcExecutionStrategies {
          * Enable thread affinity while offloading. When enabled, offloading implementation will favor using a
          * single thread per subscribe of a source.
          * @return {@code this}.
-         * @deprecated Use a single threaded executor with {@link #executor(Executor)} to ensure affinity.
+         * @deprecated Use a single threaded executor set on {@link io.servicetalk.transport.api.ExecutionContext} to
+         * ensure affinity.
          */
         @Deprecated
         public Builder offloadWithThreadAffinity() {
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuilder.java
@@ -17,6 +17,7 @@ package io.servicetalk.http.api;
 
 import io.servicetalk.buffer.api.BufferAllocator;
 import io.servicetalk.client.api.ServiceDiscovererEvent;
+import io.servicetalk.concurrent.api.Executor;
 import io.servicetalk.transport.api.IoExecutor;
 
 /**
@@ -35,6 +36,14 @@ interface HttpClientBuilder<U, R, SDE extends ServiceDiscovererEvent<R>> {
      * @return {@code this}.
      */
     HttpClientBuilder<U, R, SDE> ioExecutor(IoExecutor ioExecutor);
+
+    /**
+     * Sets the {@link Executor} for all connections created from this builder.
+     *
+     * @param executor {@link Executor} to use.
+     * @return {@code this}.
+     */
+    HttpClientBuilder<U, R, SDE> executor(Executor executor);
 
     /**
      * Sets the {@link BufferAllocator} for all connections created from this builder.
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategies.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategies.java
@@ -66,7 +66,8 @@ public final class HttpExecutionStrategies {
      *
      * @param executor {@link Executor} to use.
      * @return Default {@link HttpExecutionStrategy}.
-     * @deprecated Set the {@link Executor} to use in the {@link io.servicetalk.transport.api.ExecutionContext}.
+     * @deprecated Set the executor to use on the {@link io.servicetalk.transport.api.ExecutionContext} using
+     * {@code executor(Executor)} methods on client/server builders.
      */
     @Deprecated
     public static HttpExecutionStrategy defaultStrategy(Executor executor) {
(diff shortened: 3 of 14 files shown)


bondolo:
This change will need to be integrated before #1695 and will be back ported to 0.41 branch.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategies.java
```diff
@@ -246,8 +250,10 @@ public Builder offloadWithThreadAffinity() {
          *
          * @param mergeStrategy {@link MergeStrategy} to use.
          * @return {@code this}.
+         * @deprecated Will be removed
          */
         // Intentionally package-private, API is not required to be public for the lack of use cases.
+        @Deprecated
```
idelpivnitskiy, line 256: This is pkg-private API. Ok to deprecate it, but won't matter for users either way

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategy.java
```diff
@@ -40,7 +40,10 @@
      * trailers, for the passed {@link StreamingHttpRequest} returns a {@link Single}.
      * @param <FS> The {@code FlushStrategy} type to use.
      * @return {@link Single} which is offloaded as required.
+     * @deprecated Copy the implementation from
+     * {@link DefaultHttpExecutionStrategy#invokeClient(Executor, Publisher, Object, ClientInvoker)}.
```
idelpivnitskiy, line 44:
```suggestion
     * @deprecated This method will be removed in future releases. If you depend on it, copy the implementation from
     * {@link DefaultHttpExecutionStrategy#invokeClient(Executor, Publisher, Object, ClientInvoker)}.
```

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategy.java
```diff
@@ -57,7 +60,9 @@
      * @param service {@link Function} representing a service.
      * @return A {@link Single} that invokes the passed {@link Function} and returns the result asynchronously.
      * Invocation of {@link Function} will be offloaded if configured.
+     * @deprecated Copy the implementation from {@link DefaultHttpExecutionStrategy#invokeService(Executor, Function)}.
```
idelpivnitskiy, line 63:
```suggestion
     * @deprecated This method will be removed in future releases. If you depend on it, copy the implementation from
     * {@link DefaultHttpExecutionStrategy#invokeService(Executor, Function)}.
```

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
```diff
@@ -133,7 +135,9 @@ public Builder offloadNone() {
          *
          * @param executor {@link Executor} to use.
          * @return {@code this}.
+         * @deprecated Set the executor to use on the {@link io.servicetalk.transport.api.ExecutionContext}.
```
idelpivnitskiy, line 138: `ExecutionContextBuilder` is an internal API. How users can set an executor for the client or server?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ExecutionStrategy.java
```diff
@@ -71,10 +71,12 @@
     <T> Publisher<T> offloadReceive(Executor fallback, Publisher<T> original);
 
     /**
-     * Returns the {@link Executor}, if any for this {@link ExecutionStrategy}.
+     * Returns the {@link Executor}, if any, for this {@link ExecutionStrategy}.
      *
      * @return {@link Executor} for this {@link ExecutionStrategy}. {@code null} if none specified.
+     * @deprecated Set the {@link Executor} to use in the {@link io.servicetalk.transport.api.ExecutionContext}.
```
idelpivnitskiy, line 77:
```suggestion
     * @deprecated Take the {@link Executor} from the {@link io.servicetalk.transport.api.ExecutionContext}.
```

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
bondolo: The will set it on the `HttpServerBuilder `or `*ClientBuilder`. Perhaps add the missing Executor setters in this same change? I had intended to keep this PR only deprecations but they can be added.

idelpivnitskiy: It's hard to deprecate without providing a working alternative for users to migrate to. If this is too complicated to provide executor setters on client/builder without the whole change of #1695, we can skip back-porting it into 0.41 branch. 0.42 will anyway have some deprecated API for serialization.

idelpivnitskiy: I mean not just a setter, but a setter that will work as expected.

idelpivnitskiy: We can keep deprecation of methods that are going away without replacement.

bondolo: Added `executor(Executor executor)` to builders

bondolo: The setter will work as expected. The existing implementation will use the Executor from the strategy in preference to the Executor set in the builder. In the #1695 implementation only the builder Executor will be used.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuilder.java
```diff
@@ -36,6 +37,14 @@
      */
     HttpClientBuilder<U, R, SDE> ioExecutor(IoExecutor ioExecutor);
 
+    /**
+     * Sets the {@link Executor} for all connections created from this builder.
+     *
+     * @param executor {@link Executor} to use.
+     * @return {@code this}.
+     */
+    HttpClientBuilder<U, R, SDE> executor(Executor executor);
```
Scottmitch, line 46: new methods on interfaces is API breaking, so we shouldn't merge into 0.41. if that is the case does it need to be part of this PR or can it come in #1695 for 0.42?

bondolo: The executor additions are already currently part of #1695. This change would simply add the methods sooner so that users can move from the deprecated methods now. The interfaces aren't ones that users implement or they extend only base classes that include the necessary implementation so I don't think there is need to worry about binary incompatibility.

idelpivnitskiy: For 0.41, we should be cautious. This interface is pkg-private. For public abstract builders we can add a non-abstract method with noop default implementation. @chemicL is going to convert abstract classes into interfaces in 0.42. So, it will be cleaned and refactored in 0.42 anyway.

bondolo: I have added no-op implementations to the abstract builders so that any class which extends them will not fail with `AbstractMethodError`. If invoked the methods will generate a log warning.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -322,6 +322,14 @@ public final HttpServerBuilder appendServiceFilter(final Predicate<StreamingHttp
      */
     public abstract HttpServerBuilder ioExecutor(IoExecutor ioExecutor);
 
+    /**
+     * Sets the {@link Executor} to use.
+     *
+     * @param executor {@link Executor} to use.
+     * @return {@code this}.
+     */
+    public abstract HttpServerBuilder executor(Executor executor);
```
idelpivnitskiy, line 331: This one should also have a default implementation

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
```diff
@@ -45,7 +45,9 @@ public static GrpcExecutionStrategy defaultStrategy() {
      *
      * @param executor {@link Executor} to use.
      * @return Default {@link GrpcExecutionStrategy}.
+     * @deprecated Set the executor to use on the {@link io.servicetalk.transport.api.ExecutionContext}.
```
idelpivnitskiy, line 48: Let's clarify new methods here (general comment for all `@deprecated` annotations):
```
* @deprecated Set the executor to use on the {@link io.servicetalk.transport.api.ExecutionContext} using {@code executor(Executor)} methods on client/server builders.
```

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategy.java
```diff
@@ -67,7 +72,9 @@
      * {@link Executor}.
      * @param handler {@link StreamingHttpService} to wrap.
      * @return Wrapped {@link StreamingHttpService}.
+     * @deprecated Will be replaced with a new class {@code StreamingHttpServiceToOffloadedStreamingHttpService}.
```
idelpivnitskiy, line 75: Because this class is not available now, let's leave the same suggestion as for other methods:
```
 * @deprecated This method will be removed in future releases. If you depend on it, copy the implementation from
 * {@link DefaultHttpExecutionStrategy#offloadService(Executor, StreamingHttpService)}.
```

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
```diff
@@ -164,6 +170,12 @@
     @Override
     public abstract SingleAddressHttpClientBuilder<U, R> ioExecutor(IoExecutor ioExecutor);
 
+    @Override
+    public SingleAddressHttpClientBuilder<U, R> executor(Executor executor) {
+        LOGGER.warn("Unimplemented method SingleAddressHttpClientBuilder::executor() called");
```
idelpivnitskiy, line 175: Alternative for logging will be to throw `UnsupportedOperationException`. That way, if users have a custom implementation of the builder, they won't miss this warning after migration to the new method.

bondolo: I considered that but wanted to "do no harm". If they do use the new `executor(Executor)` method though then perhaps it should fail explicitly rather than implicitly fail.

bondolo: I am going to switch to `UnsupportedOperationException`.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClientBuilder.java
bondolo: I have replaced the no-op implementations with versions that throw `UnsupportedOperationException` if used.

idelpivnitskiy (approved):
Thank you!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -322,6 +322,16 @@ public final HttpServerBuilder appendServiceFilter(final Predicate<StreamingHttp
...
     public abstract HttpServerBuilder ioExecutor(IoExecutor ioExecutor);
 
+    /**
+     * Sets the {@link Executor} to use.
+     *
+     * @param executor {@link Executor} to use.
+     * @return {@code this}.
+     */
+    public HttpServerBuilder executor(Executor executor) {
+        throw new UnsupportedOperationException("Setting Executor not yet supported by " + getClass().getSimpleName());
```
idelpivnitskiy, line 332: After looking at #1695, I've realized that we also need similar methods for gRPC builder. Can you please add those for 0.41 @bondolo?

bondolo:
**0.41** (9a605c294fe9092d602084313c1efd64bf31ef93)

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
Scottmitch: https://github.com/apple/servicetalk/pull/1804

PR #1793, opened 2021-09-03
`ExecutionStrategy` simplification deprecations (#1781)

bondolo:
Motivation:
removed methods should be deprecated before they are removed.
Modifications:
Marks methods to be removed in #1695 as deprecated and, where possible,
provides and suggests alternatives.
Result:
Warnings for future method removals are provided.


bondolo:
Created a PR for CI testing, no need to review--same as integrated in to `main`

PR #1799, opened 2021-09-03
Update mockito 2.28.2 -> 3.12.4

Scottmitch:


servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
```diff
@@ -84,15 +85,17 @@
...
-    @Context
-    private Provider<RouteStrategiesConfig> routeStrategiesConfigProvider;
-
-    @Context
-    private RequestScope requestScope;
+    private final Provider<Ref<ConnectionContext>> ctxRefProvider;
+    private final Provider<RouteStrategiesConfig> routeStrategiesConfigProvider;
+    private final RequestScope requestScope;
+
+    EndpointEnhancingRequestFilter(@Context final Provider<Ref<ConnectionContext>> ctxRefProvider,
```
Scottmitch, line 92: @ddossot - I'm assuming the constructor injection is equivalent to filed injection in this case in terms of injection and state (doesn't need to be re-assigned). wdyt?

idelpivnitskiy (approved):

Issue #1489, opened 2021-04-15
FlakyTest: ProtocolCompatibilityTest.serviceTalkToGrpcJavaTimeout 


tkountis:
```
org.junit.experimental.theories.internal.ParameterizedAssertionError: serviceTalkToGrpcJavaTimeout("false" <from ssl[0]>, "false" <from streaming[0]>, "gzip" <from compression[0]>)
	at org.junit.experimental.theories.Theories$TheoryAnchor.reportParameterizedError(Theories.java:293)
	at org.junit.experimental.theories.Theories$TheoryAnchor$1$1.evaluate(Theories.java:242)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithCompleteAssignment(Theories.java:223)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithAssignment(Theories.java:209)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithIncompleteAssignment(Theories.java:217)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithAssignment(Theories.java:207)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithIncompleteAssignment(Theories.java:217)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithAssignment(Theories.java:207)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithIncompleteAssignment(Theories.java:217)
	at org.junit.experimental.theories.Theories$TheoryAnchor.runWithAssignment(Theories.java:207)
	at org.junit.experimental.theories.Theories$TheoryAnchor.evaluate(Theories.java:192)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
[...]

PR #1801, opened 2021-09-04
gRPC send RST_STREAM on client deadline expire

Scottmitch:
Motivation:
In the event that gRPC client deadine expires it should send a
RST_STREAM to the peer service. There is a scenario where this may not
happen due to common transport code cleaning up internal state before
propagating the error.

Modifications:
- gRPC utiles to propage the TimeoutException so users can see the
  timeout duration.
- DefaultNettyConnection should still propagate channel close events in
  the event that a write is cancelled.
- WriteStreamSubscriber should still let the close handler know that the
  channel outbound close event occurred even if nothing is pending to
  write when a ChannelOutboundListener#channelClosed event happens. This
  method may not always be the result of a transport error and may be a
  local timeout, in which case we should close the outbound side.

Result:
gRPC will send a RST_STREAM on client timeout. HTTP will close the
channel.
Fixes https://github.com/apple/servicetalk/issues/1489.

### DIFF
--- a/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
+++ b/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
@@ -635,54 +635,27 @@ class ProtocolCompatibilityTest {
     }
 
     @ParameterizedTest
-    @CsvSource({"false,false", "false,true", "true,false", "true,true"})
-    void serverTimeout(boolean stClient, boolean stServer) throws Exception {
+    @CsvSource({"false,false,false", "false,false,true", "false,true,false", "false,true,true",
+                "true,false,false", "true,false,true", "true,true,false", "true,true,true"})
+    void timeoutMidRequest(boolean stClient, boolean stServer, boolean clientInitiatedTimeout) throws Exception {
+        Duration clientTimeout = clientInitiatedTimeout ? DEFAULT_DEADLINE : null;
+        Duration serverTimeout = clientInitiatedTimeout ? null : DEFAULT_DEADLINE;
         BlockingQueue<Throwable> serverErrorQueue = new ArrayBlockingQueue<>(16);
         final TestServerContext server = stServer ?
                 serviceTalkServer(ErrorMode.NONE, false, noOffloadsStrategy(), null, null, serverErrorQueue) :
                 grpcJavaServer(ErrorMode.NONE, false, null);
-        try (ServerContext proxyCtx = buildTimeoutProxy(server.listenAddress(), DEFAULT_DEADLINE, false)) {
+        try (ServerContext proxyCtx = buildTimeoutProxy(server.listenAddress(), serverTimeout, false)) {
             final CompatClient client = stClient ?
-                    serviceTalkClient(proxyCtx.listenAddress(), false, null, null) :
-                    grpcJavaClient(proxyCtx.listenAddress(), null, false, null);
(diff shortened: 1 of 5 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -140,6 +139,28 @@
     private volatile DataObserver dataObserver;
     private final boolean isClient;
     private final UnaryOperator<Throwable> enrichProtocolError;
+    private final TerminalSignalConsumer cleanupStateConsumer = new TerminalSignalConsumer() {
```
Scottmitch, line 142: I debated having `DefaultNettyConnection implement TerminalSignalConsumer` but was hesitant to expose  this on public api

Scottmitch, line 142: I debated having `DefaultNettyConnection implement TerminalSignalConsumer` but was hesitant to expose  this on public api

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
```diff
@@ -569,53 +589,131 @@ void serviceTalkToServiceTalkErrorWithStatusViaServerFilter(
 
     @ParameterizedTest
     @MethodSource("sslStreamingAndCompressionParams")
-    @Disabled("https://github.com/apple/servicetalk/issues/1489")
-    void grpcJavaToGrpcJavaTimeout(final boolean ssl,
-                                   final boolean streaming,
-                                   final String compression) throws Exception {
-        Duration timeout = Duration.ofNanos(1);
+    void grpcJavaToGrpcJavaTimeout(final boolean ssl, final boolean streaming, final String compression)
```
idelpivnitskiy, line 592: For all 4 variants: consider clarifying in the test name that timeout/deadline is applied on the client-side

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
```diff
@@ -569,53 +589,131 @@ void serviceTalkToServiceTalkErrorWithStatusViaServerFilter(
...
-        final CompatClient client = serviceTalkClient(server.listenAddress(), ssl, compression, timeout);
-        testGrpcError(client, server, false, streaming, compression, GrpcStatusCode.DEADLINE_EXCEEDED, null);
+        try (ServerContext proxyCtx = buildTimeoutProxy(server.listenAddress(), ssl)) {
+            final CompatClient client = serviceTalkClient(proxyCtx.listenAddress(), ssl, compression, DEFAULT_DEADLINE);
+            testGrpcError(client, server, false, streaming, compression, DEADLINE_EXCEEDED, null);
+        }
+    }
+
+    @ParameterizedTest
+    @MethodSource("sslAndStreamingParams")
```
idelpivnitskiy, line 635: While `sslAndStreamingParams` produces 4 pairs of booleans, the method name of the source does not match with what params the test defines. Consider using:
```java
@CsvSource({"false,false", "false,true", "true,false", "true,true"})
```

idelpivnitskiy (approved):
🎖️

Scottmitch:

Scottmitch:
main (860101ca9)

PR #1803, opened 2021-09-06
Add Http2Exception, convert to GrpcStatuCode

Scottmitch:
Motivation:
Http2Exception currently exists as an internal class inside of the
http.netty class. It is therefore not accessible to the grpc packages
which limits our ability to convert h2 exceptions to GrpcStatusCode.

Modifications:
- Add Http2Exception and HttpErrorCode to http.api
- Convert from Http2Exception to GrpcStatusCode in grpc.api

Result:
Users have more visibility into h2 exceptions, and gRPC can convert
these exception into gRPC error codes.


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatusCode.java
```diff
@@ -63,18 +61,31 @@
...
 
     static {
-        final Map<Integer, GrpcStatusCode> intToStatusCodeMap = new HashMap<>(GrpcStatusCode.values().length);
-        for (GrpcStatusCode code : GrpcStatusCode.values()) {
-            GrpcStatusCode replaced = intToStatusCodeMap.put(code.value(), code);
-            if (replaced != null) {
-                throw new IllegalStateException(String.format("GrpcStatusCode value %d used by both %s and %s",
-                        code.value(), replaced, code));
-            }
+        final GrpcStatusCode[] statusCodes = GrpcStatusCode.values();
```
idelpivnitskiy, line 68: As part of API stability, we should consider using the final class approach for `GrpcStatusCode` either. The code is in integer range. There is a change a new codes may be added in newer gRPC protocol versions.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatusCode.java
```diff
@@ -122,6 +134,17 @@ public int value() {
...
-        return GrpcStatus.CACHED_INSTANCES.get(value);
+        return GrpcStatus.fromCodeValue(value);
+    }
+
+    /**
+     * Convert from {@link Http2ErrorCode} to {@link GrpcStatus}.
+     * @param errorCode the {@link Http2ErrorCode} to convert.
+     * @return the result of the conversion.
+     */
+    static GrpcStatusCode fromHttp2Exception(Http2ErrorCode errorCode) {
```
idelpivnitskiy, line 145: `fromHttp2Exception` -> `fromHttp2ErrorCode`

idelpivnitskiy (changes requested):

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2ErrorCode.java
```diff
@@ -0,0 +1,74 @@
...
+    FRAME_SIZE_ERROR(0x6),
+    REFUSED_STREAM(0x7),
+    CANCEL(0x8),
+    COMPRESSION_ERROR(0x9),
+    CONNECT_ERROR(0xA),
+    ENHANCE_YOUR_CALM(0xB),
+    INADEQUATE_SECURITY(0xC),
+    HTTP_1_1_REQUIRED(0xD);
+
+    private final long errorCode;
```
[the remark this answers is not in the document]
Scottmitch: https://datatracker.ietf.org/doc/html/rfc7540#section-6.4
> The RST_STREAM frame contains a single unsigned, 32-bit integer
   identifying the error code (Section 7).  The error code indicates why
   the stream is being terminated.

The value is unsigned, so we used `long` in Netty. We could use `int` here as the high values that would make `int` negative are not used and the value is unlikely to be used in arithmetic and/or logical comparisons.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttp2ExceptionUtils.java
```diff
@@ -15,43 +15,62 @@
...
+final class NettyHttp2ExceptionUtils {
 
-    Http2Exception(final Throwable cause) {
-        super(cause);
+    private NettyHttp2ExceptionUtils() {
     }
 
     static Throwable wrapIfNecessary(final Throwable cause) {
         if (cause instanceof io.netty.handler.codec.http2.Http2Exception) {
+            final int streamId = (cause instanceof StreamException) ? ((StreamException) cause).streamId() : 0;
```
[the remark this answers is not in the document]
Scottmitch: I'll adjust the constructor to restrict to `>=0`. There are some cases where the streamId may not be valid in Netty (e.g. [CompositeStreamException](https://github.com/netty/netty/blob/netty-4.1.66.Final/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Exception.java#L295-L312)), but we can handle that in other ways and relax the value in the future.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatusCode.java
Scottmitch: I think enum is OK in this case bcz `UNKNOWN` accounts for this.

https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/Status.java#L73-L80
```java
/**
     * Unknown error.  An example of where this error may be returned is
     * if a Status value received from another address space belongs to
     * an error-space that is not known in this address space.  Also
     * errors raised by APIs that do not return enough error information
     * may be converted to this error.
     */
    UNKNOWN(2),
```

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Exception.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Create a new instance.
+     * @param streamId {@code 0} for the connection stream, {@code > 0} for a non-connection stream, and {@code < 0} if
+     * unknown.
+     * @param error The error code.
+     * @param cause The original cause which lead to this exception.
+     */
+    public Http2Exception(final int streamId, final Http2ErrorCode error, final Throwable cause) {
```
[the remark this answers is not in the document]
Scottmitch: lets just add a new constructor if we need the 4 params in the future

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttp2ExceptionUtils.java
```diff
@@ -107,29 +134,19 @@ public Throwable fillInStackTrace() {
...
-        assert stream != null;
-        if (resetFrame.errorCode() == REFUSED_STREAM.code()) {
-            return new H2StreamRefusedException("RST_STREAM received for streamId=" + stream.id() + ", stream refused");
-        } else {
-            return new H2StreamResetException("RST_STREAM received for streamId=" + stream.id() + " with error code: " +
-                    resetFrame.errorCode());
-        }
+    private static Http2ErrorCode stErrorCode(long code) {
+        Http2ErrorCode errorCode = Http2ErrorCode.valueOf(code);
+        return errorCode == null ? Http2ErrorCode.INTERNAL_ERROR : errorCode;
```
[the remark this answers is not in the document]
Scottmitch: I decided not to add logging here for now ... we create a new `Http2ErrorCode` now due to the class changes. we can add more visibility in a followup if necessary.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2ErrorCode.java
```diff
@@ -0,0 +1,169 @@
...
+ * limitations under the License.
+ */
+package io.servicetalk.http.api;
+
+import javax.annotation.Nullable;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * An enum containing http/2 <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-7">error codes</a>.
```
idelpivnitskiy, line 23: This is not an enum anymore.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2ErrorCode.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return errorCode;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        return this == o || (o instanceof Http2ErrorCode && errorCode == ((Http2ErrorCode) o).errorCode);
```
idelpivnitskiy, line 144: For the final classes, it's sligtly cheaper to use `getClass()` instead of `instanceof`. Up to you.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Exception.java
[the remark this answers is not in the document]
idelpivnitskiy: Ok, we can add 4 params later, but would be nice to give users `(streamId, error, message, cause)` overload because it's frequently used ctor.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatusCode.java
idelpivnitskiy: I'm less concerned with inability to map to something, more with something like gRPC2 is released in future with 10 more status codes defined. Adding new enum values may be a breaking change for users.

idelpivnitskiy (approved):
Last comments and LGTM,
thank you!

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatusCode.java
Scottmitch: yip fair enough, we can add it to the list. also note that gRPC-java represent this as an enum.

idelpivnitskiy: I saw that, but they can release grpc-java-2, for us it will be harder

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2ErrorCode.java
Scottmitch: Then I also have to check against `null` I'll just leave as is for now we can optimize/followup for similar `final` "enum" classes wholesale if this is found to be beneficial.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Exception.java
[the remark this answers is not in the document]
Scottmitch: added

Scottmitch:
main (7f13f340b)

PR #1805, opened 2021-09-07
Http executor builder methods abstract

Scottmitch:
Motivation:
0.42 release will break some APIs, and the builder methods that throw
UnsupportedOperationException can be made abstract to force
implementing.

Modifications:
- Remove UnsupportedOperationException and make methods abstract

Result:
Http builder executor methods are now abstract.


idelpivnitskiy (approved):
Do you want to do the same for grpc builders in this PR or a follow-up?

Scottmitch:
@idelpivnitskiy - the gRPC builders on `main` didn't use the `UnsupportedOperationException` approach.

PR #1806, opened 2021-09-07
Make NettyIoThread package private

Scottmitch:
Motivation:
NettyIoThreadFactory has an internal public class NettyIoThread factory,
however this type doesn't need to be exposed publically.

Modifications:
- Make NettyIoThread package private

Result:
Less public types.


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoThreadFactory.java
```diff
@@ -29,7 +26,7 @@
 /**
  * Default {@link IoThreadFactory} to create IO {@link NettyIoThread}s.
  */
-public final class NettyIoThreadFactory implements IoThreadFactory<NettyIoThreadFactory.NettyIoThread> {
+public final class NettyIoThreadFactory implements IoThreadFactory<NettyIoThread> {
```
Scottmitch, line 29: maybe awkward to have a non-public type for the generic argument but we don't actually need to expose this type.

idelpivnitskiy: if it passes all tests 🚀

idelpivnitskiy (approved):

chemicL (approved):

bondolo (approved):

bondolo (approved):

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoThread.java
```diff
@@ -0,0 +1,43 @@
...
+package io.servicetalk.transport.netty.internal;
+
+import io.servicetalk.concurrent.api.AsyncContextMap;
+import io.servicetalk.transport.api.IoThreadFactory.IoThread;
+
+import io.netty.util.concurrent.FastThreadLocalThread;
+
+import javax.annotation.Nullable;
+
+final class NettyIoThread extends FastThreadLocalThread implements IoThread {
```
idelpivnitskiy, line 25: This class is not used outside of `NettyIoThreadFactory`, consider keeping it where it was as a pkg-private.

PR #1810, opened 2021-09-07
Extend `japicmp.sh` to compare versus local build

bondolo:
Motivation:
It is useful to be able to compare the Java API compatibility of a
local build against a prior release.
Modifications:
Extend `japicmp.sh` to provide capability to compare an old version
of ServiceTalk against a local build.
Result:
Additional API compatibility comparison capability.

### DIFF
--- a/scripts/japicmp.sh
+++ b/scripts/japicmp.sh
@@ -18,11 +18,11 @@
 SCRIPT=$(basename "${BASH_SOURCE:-stidn}")
 
 if [ $# -lt 1 ] || [ $# -gt 3 ]; then
-	echo "# This script compares versions for binary backward compatibility."
-	echo "# Usage: ${SCRIPT} <old_version> (<new_version> (<group_id>))"
-	echo "# if <new_version> unspecified then comparison will be made againt local build"
-	echo "# if <group_id> unspecified then local dir gradle 'group' property will be used"
-	exit 1
+  echo "# This script compares versions for binary backward compatibility."
+  echo "# Usage: ${SCRIPT} <old_version> (<new_version> (<group_id>))"
+  echo "# if <new_version> unspecified then comparison will be made againt local build"
+  echo "# if <group_id> unspecified then local dir gradle 'group' property will be used"
+  exit 1
 fi
 
 MVN_REPO="$(mvn help:evaluate -Dexpression=settings.localRepository -q -DforceStdout)"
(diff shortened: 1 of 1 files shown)


bondolo:
Most of the rest of the changes are due either to [`shellcheck`](https://github.com/koalaman/shellcheck) or [`shfmt`](https://github.com/mvdan/sh).

scripts/japicmp.sh
```diff
@@ -15,43 +15,54 @@
...
-      -Dtransitive=false >/dev/null
-  java -jar "$JAR_FILE" -b --ignore-missing-classes \
-      --old $BASEPATH/$ARTIFACT_ID/$OLD_ST_VERSION/$ARTIFACT_ID-$OLD_ST_VERSION.jar \
-      --new $BASEPATH/$ARTIFACT_ID/$NEW_ST_VERSION/$ARTIFACT_ID-$NEW_ST_VERSION.jar
+ARTIFACTS="$(find servicetalk-* -type d -maxdepth 0 |
+  grep -v -- '-\(benchmarks\|bom\|examples\|gradle-plugin-internal\)$')"
+
+for ARTIFACT_ID in ${ARTIFACTS}; do
+  mvn -N -U dependency:get -DgroupId="${GROUP_ID}" -DartifactId="${ARTIFACT_ID}" \
+    -Dversion="${OLD_ST_VERSION}" -Dtransitive=false >/dev/null || continue
```
chemicL, line 52: in case of error, would we want anything reported at the end instead of skipping the error and going forward?

bondolo: The most likely cause of the error is a module that exists in the new version but not in the old version. The only way to do this "right" would be to process the bom lists from both the old and the new and the only diff the intersection but that seemed to be going too far. (We should look at the japicmp plugin instead for fancier use).

scripts/japicmp.sh
```diff
@@ -15,43 +15,54 @@
...
+
+  if [[ ${NEW_ST_VERSION} =~ .+-SNAPSHOT$ ]]; then
```
idelpivnitskiy, line 54: `NEW_ST_VERSION` can be a snapshot or a release, like `0.41.5`, we should keep supporting both.

scripts/japicmp.sh
bondolo: I enhanced the script to now output a warning that it is skipping.

scripts/japicmp.sh
```diff
@@ -20,8 +20,8 @@ SCRIPT=$(basename "${BASH_SOURCE:-stidn}")
 if [ $# -lt 1 ] || [ $# -gt 3 ]; then
   echo "# This script compares versions for binary backward compatibility."
   echo "# Usage: ${SCRIPT} <old_version> (<new_version> (<group_id>))"
-  echo "# if <new_version> unspecified then comparison will be made againt local build"
-  echo "# if <group_id> unspecified then local dir gradle 'group' property will be used"
+  echo "# if optional <new_version> unspecified then comparison will be made againt local build"
+  echo "# if optional <group_id> unspecified then local dir gradle 'group' property will be used"
```
idelpivnitskiy, line 24: It might be useful to be able to run a local build with `group_id`. Consider using "local" as a keyword for the local build to allow using 3rd parameter. We can also clarify in this message that if "local" is used, users should run `./gradlew build` before running this script.

idelpivnitskiy (approved):
Nice improvements, thank you! 🎖️

PR #1807, opened 2021-09-07
Revert "Deprecate `IoExecutor` create w/ `ThreadFactory` (#1738)"

Scottmitch:
Motivation:
japicmp.sh revealed some API breakage [1].

[1]
```
***! MODIFIED CLASS: PUBLIC NON_FINAL (<- FINAL) io.servicetalk.transport.netty.internal.IoThreadFactory  (not serializable)
	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
	---! REMOVED INTERFACE: java.util.concurrent.ThreadFactory
	===  UNCHANGED SUPERCLASS: java.lang.Object (<- java.lang.Object)
```

Modifications:
- Revert 441f907bf32e2f0d2086067f75ecf118664fcf79

Result:
Avoid API breakage, which can be corrected in a followup PR.

PR #1811, opened 2021-09-07
Add `ThreadFactory` interface to `IoThreadFactory`

bondolo:
Motivation:
For greatest potential compatibility the Netty `IoThreadFactory` should
implement the `ThreadFactory` interface explicitly even though this is
redundant.
Modifications:
Reverts the revert PR #1807 and Netty `IoThreadFactory` implements
the `ThreadFactory` interface explicitly which it did not in the original (#1738) PR.
Result:
Greatest backwards compatibility with existing apps compiled against
earlier versions of ServiceTalk 0.41

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/IoThreadFactory.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/IoThreadFactory.java
@@ -47,6 +47,7 @@ public class IoThreadFactory implements java.util.concurrent.ThreadFactory,
      * @param threadNamePrefix the name prefix used for the created {@link Thread}s.
      * @deprecated Consider using {@code io.servicetalk.transport.netty.NettyIoExecutors#createIoExecutor(String)}.
      */
+    @Deprecated
     public IoThreadFactory(String threadNamePrefix) {
         this(threadNamePrefix, true);
     }
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/IoThreadFactory.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/IoThreadFactory.java
@@ -78,7 +78,7 @@ public class IoThreadFactory implements java.util.concurrent.ThreadFactory,
         return t;
     }
 
-    public static final class NettyIoThread extends FastThreadLocalThread
+    static final class NettyIoThread extends FastThreadLocalThread
             implements io.servicetalk.transport.api.IoThreadFactory.IoThread {
 
         @Nullable
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/IoThreadFactory.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/IoThreadFactory.java
@@ -45,6 +45,7 @@ public class IoThreadFactory implements java.util.concurrent.ThreadFactory,
     /**
      * Create a new instance.
      * @param threadNamePrefix the name prefix used for the created {@link Thread}s.
+     * @deprecated Consider using {@code io.servicetalk.transport.netty.NettyIoExecutors#createIoExecutor(String)}.
      */
     public IoThreadFactory(String threadNamePrefix) {
         this(threadNamePrefix, true);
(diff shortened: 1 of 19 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/IoThreadFactory.java
```diff
@@ -72,12 +78,13 @@ public Thread newThread(Runnable r) {
         return t;
     }
 
-    private static final class AsyncContextHolderNettyThread extends FastThreadLocalThread
-            implements AsyncContextMapHolder {
+    public static final class NettyIoThread extends FastThreadLocalThread
```
Scottmitch, line 81: I don't think this needs to be public, and I made it package private in https://github.com/apple/servicetalk/pull/1806. wdyt?

bondolo: OK

bondolo:
Build failure due to #1507

servicetalk-transport-netty/src/main/java/io/servicetalk/transport/netty/NettyIoExecutors.java
```diff
@@ -91,12 +91,35 @@ public static IoExecutor createIoExecutor(ThreadFactory threadFactory) {
         return io.servicetalk.transport.netty.internal.NettyIoExecutors.createIoExecutor(threadFactory);
     }
 
+    public static IoExecutor createIoExecutor(String threadNamePrefix) {
```
idelpivnitskiy, line 94: This public method misses javadoc

servicetalk-transport-netty/src/main/java/io/servicetalk/transport/netty/NettyIoExecutors.java
```diff
@@ -91,12 +91,35 @@ public static IoExecutor createIoExecutor(ThreadFactory threadFactory) {
...
     public static IoExecutor createIoExecutor() {
-        return createIoExecutor(new NettyIoThreadFactory(NettyIoExecutor.class.getSimpleName()));
+        return createIoExecutor(newIoThreadFactory());
+    }
+
+    private static NettyIoThreadFactory newIoThreadFactory() {
+        return newIoThreadFactory(NettyIoExecutor.class.getSimpleName());
+    }
+
+    private static NettyIoThreadFactory newIoThreadFactory(String prefix) {
```
idelpivnitskiy, line 122: These two methods are not required because you have a corresponding overloads in `internal.NettyIoExecutors`.

bondolo: The expectation is that not everyone would have the `servicetalk-transport-netty-internal` module imported (or only transitively).

servicetalk-transport-netty/src/main/java/io/servicetalk/transport/netty/NettyIoExecutors.java
bondolo: I will add it at.

servicetalk-transport-netty/src/main/java/io/servicetalk/transport/netty/NettyIoExecutors.java
idelpivnitskiy: If they don't have `servicetalk-transport-netty-internal` module imported, they won't be able to use `new NettyIoThreadFactory(prefix)` the line below.

bondolo: I would expect that `servicetalk-transports-netty` is imported as API and `servicetalk-transports-netty-internal` is imported as implementation or is only imported transitively by one of their other imports.

If both are typically imported as API then this makes me wonder why then why there are two `NettyIoExecutors` utility classes.

idelpivnitskiy: I think we are talking about different things. You introduced two `private` utilities inside `netty.NettyIoExecutors` class, and new overloads in `netty.internal.NettyIoExecutors`. I'm saying that these new private static methods are unnecessary, because you can just invoke corresponding methods from `netty.internal.NettyIoExecutors`.

Reason why we have two `NettyIoExecutors` is that the internal one has access to `EventLoopGroupIoExecutor` and also exposes API to wrap pre-existing `EventLoop` with `IoExecutor` interface. We don't need to make that API public. As the result, `netty.NettyIoExecutors` is a limited variant of `netty.internal.NettyIoExecutors`, that makes public only what is required for users, but hides what is required for us.

bondolo: OK, got it now. It is as I expected, only `netty.NettyIoExecutors` is public.

bondolo: I will have to create a PR for main to keep it in sync.

PR #1812, opened 2021-09-08
Add missing subscribeShareContext to missing static sources

Scottmitch:
Motivation:
The subscribeShareContext was added in a few cases when defer was used
but not added to all return paths. It should also be added to static
sources like succeed and failed for consistency.

Modifications:
- Update deprecated ContentCoding filters to use subscribeShareContext
  on static succeeded operators.
- Move subscribeShareContext in tracing filters to the usage of defer,
  to make the need for subscribeShareContext more clear.

Result:
More consistent usage of subscribeShareContext on static sources when
control flow otherwise expects sharing context.


tkountis (approved):

chemicL (approved):

idelpivnitskiy:
Let's hold this PR until after we discuss this on team sync. I'm playing with your example and think it may not represent what is actually happens in this filters.

idelpivnitskiy (approved):
I had another look there, and realized what you were talking about. Effectively, there will be no difference for the AsyncContext state for users, because there is no user code involved when we return a static source. But I'm agreed that we should be consistent in case of errors as well, and not make a copy of the context. It will also remove one more allocation, which is nice. Thank you for the PR!

PR #1813, opened 2021-09-08
Introduce a helper predicate for delayed retries

tkountis:
Motivation:

We recently introduced a DelayedRetry marker interface that supports retries with certain fixed delays.

Modifications:

Yet another default predicate to help users easier adopt the change.

Result:

Friendlier API for adoption

### DIFF
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/RetryingHttpRequesterFilter.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/RetryingHttpRequesterFilter.java
@@ -148,15 +148,31 @@ public final class RetryingHttpRequesterFilter implements StreamingHttpClientFil
          * <a href="https://tools.ietf.org/html/rfc7231#section-4.2.2">idempotent</a> requests when applicable.
          * <p>
          * <b>Note:</b> This predicate expects that the retried {@link StreamingHttpRequest requests} have a
-         * {@link StreamingHttpRequest#payloadBody() payload body} that is replayable, i.e. multiple subscribes to the
+         * {@link StreamingHttpRequest#payloadBody() payload body} that is re-playable, i.e. multiple subscribes to the
          * payload {@link Publisher} emit the same data. {@link Publisher}s that do not emit any data or which are
-         * created from in-memory data are typically replayable.
+         * created from in-memory data are typically re-playable.
          *
          * @return a {@link BiPredicate} for {@link #retryFor(BiPredicate)} builder method
          */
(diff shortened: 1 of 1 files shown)


servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/RetryingHttpRequesterFilter.java
```diff
@@ -148,16 +148,32 @@ protected RetryingHttpRequesterFilter build(
...
+         * as such.
+         * <p>
+         * <b>Note:</b> This predicate expects that the retried {@link StreamingHttpRequest requests} have a
+         * {@link StreamingHttpRequest#payloadBody() payload body} that is re-playable, i.e. multiple subscribes to the
+         * payload {@link Publisher} emit the same data. {@link Publisher}s that do not emit any data or which are
+         * created from in-memory data are typically re-playable.
+         *
+         * @return a {@link BiPredicate} for {@link #retryFor(BiPredicate)} builder method
+         */
+        public BiPredicate<HttpRequestMetaData, Throwable> retryForDelayedRetryPredicate() {
```
chemicL, line 173: Have you considered adding a test case for this utility?

tkountis: I could but if the operator works (which is a tested in various places) my test would only test if the predicate and `instanceof` works, which doesn't add any value.

idelpivnitskiy: Thank you for the PR! That gave me a chance to think about it one more time:
1. It was easier to use `defaultRetryForPredicate().or(...)` pattern when we had only two predicates. With 3 predicates it's more challenging. How would you configure all: default one + idempotent requests + delayed retry predicates? I would propose to start using builder options, like:
- `retryOnRetryableExceptions(boolean)` - true by default
- `retryDelayedRetries(boolean)` - false by default
- `retryIdempotentRequests(boolean)` - false by default
We can clarify in javadoc that all these options are additive to whatever users pass in `retryFor`. That way they can provide a custom predicate + use what we provide. After the proposed change, there will be no need for public static predicates (those are currently public only to allow users to reuse pre-defined predicates if they build a custom predicate themselves).
2. `DelayedRetry` looks like a protocol-agnostic feature. Consider moving it to `AbstractRetryingFilterBuilder`.

tkountis: Makes sense, will follow up

PR #1816, opened 2021-09-08
Test utility ProxyTunnel wait for all input to copy

Scottmitch:
Motivation:
ProxyTunnel is a test utility written using java.io primitives. It
creates a proxy and copies the corresponding InputStreams to
OutputStreams. However this process may prematurely exit before the
proxy client's InputStream is finished copying to the Server socket's
OutputStream. This may lead to test failures if the output isn't
completely copied and premature termination happens.

Modifications:
- ProxyTunnel#handleRequest should block until the executed task to copy
  proxy client's InputStream to the server socket's OutputStream.

Result:
Tests which use ProxyTunnel less likely to fail due to premature
closures.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
@@ -27,7 +27,9 @@ import java.io.OutputStream;
 import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 import javax.annotation.Nullable;
 
@@ -117,7 +119,8 @@ final class ProxyTunnel implements AutoCloseable {
         return new String(bytes, 0, i, UTF_8);
     }
 
-    private void handleRequest(final Socket socket, final InputStream in, final String initialLine) throws IOException {
+    private void handleRequest(final Socket socket, final InputStream in, final String initialLine) throws IOException,
+            ExecutionException, InterruptedException {
         if (initialLine.startsWith(CONNECT_PREFIX)) {
             final int end = initialLine.indexOf(' ', CONNECT_PREFIX.length());
             final String authority = initialLine.substring(CONNECT_PREFIX.length(), end);
@@ -133,8 +136,9 @@ final class ProxyTunnel implements AutoCloseable {
             out.flush();
 
             final InputStream cin = clientSocket.getInputStream();
-            executor.submit(() -> copyStream(out, cin));
+            Future<?> f = executor.submit(() -> copyStream(out, cin));
             copyStream(clientSocket.getOutputStream(), in);
+            f.get(); // wait for the copy of proxy client input to server output to finish copying.
         } else {
             throw new RuntimeException("Unrecognized initial line: " + initialLine);
         }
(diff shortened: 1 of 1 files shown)


Scottmitch:
Related test failures:
- https://github.com/apple/servicetalk/issues/1507

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
```diff
@@ -126,45 +129,44 @@ private void handleRequest(final Socket socket, final InputStream in, final Stri
...
     }
 
-    private static void copyStream(final OutputStream out, final InputStream cin) {
+    private static void copyStream(final OutputStream out, final InputStream cin) throws IOException {
         try {
             int b;
             while ((b = cin.read()) >= 0) {
                 out.write(b);
             }
             out.flush();
```
bondolo, line 157: probably not needed since close() is imminent and implies flush()

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
```diff
@@ -126,45 +129,44 @@ private void handleRequest(final Socket socket, final InputStream in, final Stri
...
         } else {
             throw new RuntimeException("Unrecognized initial line: " + initialLine);
         }
     }
 
-    private static void copyStream(final OutputStream out, final InputStream cin) {
+    private static void copyStream(final OutputStream out, final InputStream cin) throws IOException {
         try {
             int b;
             while ((b = cin.read()) >= 0) {
```
bondolo, line 154: Unless the output is always tiny, even a small buffer would be orders of magnitude more efficient.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
```diff
@@ -126,45 +129,44 @@ private void handleRequest(final Socket socket, final InputStream in, final Stri
...
-            executor.submit(() -> copyStream(out, cin));
-            copyStream(clientSocket.getOutputStream(), in);
+                final InputStream cin = clientSocket.getInputStream();
+                Future<Void> f = executor.submit(() -> {
+                    copyStream(out, cin);
+                    return null;
+                });
+                copyStream(clientSocket.getOutputStream(), in);
+                f.get(); // wait for the copy of proxy client input to server output to finish copying.
+            }
```
bondolo, line 145: perhaps `finally { in.close(); }` for completeness

bondolo (approved):

idelpivnitskiy (approved):

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
idelpivnitskiy: IIRC the goal here is to simulate a slow simple proxy that fragments data.

Scottmitch: I do recall a side-goal of maximizing the fragmenting of data being discussed at some point. lets re-asses in a followup PR though.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
Scottmitch: I'll leave the flush for now. the `close()` javadoc doesn't explicitly say anything about flushing it just says `Closes this output stream and releases any system resources associated with this stream.`.

Scottmitch:
0.41 (eb246be1c)

idelpivnitskiy:
After thinking about this change more and seeing your #1817 PR, I realized this change violates what tests are designed to verify. We need a fool proxy that closes connection as soon as it observes FIN from one of the sides. It helps to verify that ST server closes gracefully only after it receives the full request payload body. So, the errors you saw by a proxy logger were intentional. I know that tests are flaky, but we need to search for another reason.

Scottmitch:
make sense. created https://github.com/apple/servicetalk/pull/1822 as a followup.

PR #1817, opened 2021-09-09
Test utility ProxyTunnel full duplex

Scottmitch:
Motivation:
The ProxyTunnel test utility uses java.io Socket APIs to proxy traffic.
It also closes each InputStream and OutputStream, but this will close
the underlying Socket which leads to test failures due to premature
closures.

Motivation:
- Use Socket.shutdownInput() and Socket.shutdownInput() so the proxy is
  full duplex.

Result:
Less likely that the ProxyTunnel will prematurely close.

PR #1818, opened 2021-09-09
ConnectionCloseHeaderHandlingTest don't write empty payload

Scottmitch:
Motivation:
ConnectionCloseHeaderHandlingTest$NonPipelinedRequestsTest with
noResponseContent=true has been observed to fail [1]. This is because
the test assumes there will be no payload content however the serializer
always appends a fixed 4 byte length onto every write (even if the
payload is empty). This may result in the payload not being empty.

[1]
```
java.lang.AssertionError:
Expected: "0"
     but: was "15"
```

Motiviations:
- ConnectionCloseHeaderHandlingTest server will not write empty content
  to the PayloadWriter for serialization.

Result:
Less test failures in ConnectionCloseHeaderHandlingTest.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionCloseHeaderHandlingTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionCloseHeaderHandlingTest.java
@@ -200,7 +200,9 @@ final class ConnectionCloseHeaderHandlingTest {
                                     }
                                 } while (!done);
                             }
-                            writer.write(content);
+                            if (!content.isEmpty()) {
+                                writer.write(content);
+                            }
                         }
                     });
 


Scottmitch:
Related issues:

Scottmitch:
I was able to reproduce the specific failure locally, and wasn't happening after this fix. I will merge this PR to debug the more frequently occurring variation related to https://github.com/apple/servicetalk/pull/1817

Scottmitch:
0.41 (537a72615)

idelpivnitskiy:
Imho we should improve `HttpStringSerializer` to not write anything if the string was empty

idelpivnitskiy:
Can we establish a contract that if `FixedLengthStreamingSerializer#bytesEstimator` returns 0, we skip an item?

idelpivnitskiy:
Nvm, it will break compatibility with `MessageLite#writeDelimitedTo(proto)`. Thanks for fixing the test!

PR #1821, opened 2021-09-09
Use bom to determine modules to compare

bondolo:
Motivation:
The japicmp script used the local directory to determine the modules to
compare, even if that local directory had nothing to do with the
versions being compared.
Modifications:
Use the bom files from the versions being compared to determine which
modules to compare. The local directory will be used rather than bom
file if doing a local compare. The modules compared is the intersection
of the modules present in both versions.
Result:
Smarter comparison

### DIFF
--- a/scripts/japicmp.sh
+++ b/scripts/japicmp.sh
@@ -29,6 +29,40 @@ if [ $# -lt 1 ] || [ $# -gt 3 ]; then
   exit 1
 fi
 
+function bom_artifacts() {
+  local GROUP_ID=${1}
+  local VERSION=${2}
+  mvn -N -U dependency:get -DgroupId="${GROUP_ID}" -DartifactId="servicetalk-bom" \
+    -Dversion="${VERSION}" -Dpackaging=pom -Dtransitive=false >/dev/null
+  (xsltproc - \
+    "${BASEPATH}/servicetalk-bom/${VERSION}/servicetalk-bom-${VERSION}.pom" |
+    grep '^servicetalk-' |
+    sort -) <<"XSLTDOC"
+<?xml version="1.0"?>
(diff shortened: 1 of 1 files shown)


bondolo:
I remembered something similar I did long ago and it was pretty easy to put this together. The xslt is actually generated by `xmlstarlet` but the processing is done by `xsltproc` since it is widely installed.

scripts/japicmp.sh
```diff
@@ -52,9 +86,19 @@ if [ -z "${OLD_ST_VERSION}" ]; then
...
+else
+  NEW_ARTIFACTS="$(bom_artifacts "${GROUP_ID}" "${NEW_ST_VERSION}")"
+fi
+
 # All servicetalk modules except:
 # servicetalk-benchmarks, servicetalk-bom, servicetalk-examples, servicetalk-gradle-plugin-internal
-ARTIFACTS="$(find servicetalk-* -type d -maxdepth 0 |
+ARTIFACTS="$(comm -1 -2 \
+  <(echo "${OLD_ARTIFACTS}" | tr ' ' '\n') \
+  <(echo "${NEW_ARTIFACTS}" | tr ' ' '\n') |
```
idelpivnitskiy, line 101: Do we need `NEW_ARTIFACTS` in the final `ARTIFACTS`? We should only make sure that old artifacts are compatible. If new artifacts were added, there is nothing we can compare with from `OLD_ARTIFACTS`

bondolo: `comm` is generating a list that is the intersection of the old and new artifacts. If either old or new contains artifacts that the other does not then they will not be compared.

idelpivnitskiy: Oh, I missed `comm` at the beginning, thanks for clarifying!

idelpivnitskiy (approved):
🚢

PR #1822, opened 2021-09-09
Avoid appSerializerUtf8FixLen in tests if content-length is byte length

Scottmitch:
Motivation:
Some tests use appSerializerUtf8FixLen but also rely upon content-length
header being the same quantity as the number of bytes to terminate
gracefully. In these cases we may get spurious errors due to
content-length being larger than the actual payload length because the
http protocol doesn't think we are done, but the socket closes anyways.

Motivation:
- Restore ProxyTunnel behavior which was intended to simulate proxies
  that don't support half closure.
- ConnectionCloseHeaderHandlingTest and
  GracefulConnectionClosureHandlingTest just count total bytes, and
  don't need to restore the exact String contents. In this case we can
  just use a raw string serializer and count bytes manually.

Result:
ConnectionCloseHeaderHandlingTest and
GracefulConnectionClosureHandlingTest should fail less frequently due to
premature closures. ProxyTunnel again replicates behavior of a proxy
that doesn't do half closure.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
@@ -21,15 +21,14 @@ import io.servicetalk.transport.api.HostAndPort;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 import javax.annotation.Nullable;
 
@@ -40,7 +39,6 @@ import static java.util.concurrent.Executors.newCachedThreadPool;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 final class ProxyTunnel implements AutoCloseable {
-
     private static final Logger LOGGER = LoggerFactory.getLogger(ProxyTunnel.class);
     private static final String CONNECT_PREFIX = "CONNECT ";
 
@@ -51,6 +49,7 @@ final class ProxyTunnel implements AutoCloseable {
     private ServerSocket serverSocket;
     private ProxyRequestHandler handler = this::handleRequest;
 
+    @SuppressWarnings("ResultOfMethodCallIgnored")
     @Override
     public void close() throws Exception {
         try {
@@ -61,6 +60,7 @@ final class ProxyTunnel implements AutoCloseable {
         }
     }
 
+    @SuppressWarnings("StatementWithEmptyBody")
     HostAndPort startProxy() throws IOException {
         serverSocket = new ServerSocket(0, 50, getLoopbackAddress());
         final InetSocketAddress serverSocketAddress = (InetSocketAddress) serverSocket.getLocalSocketAddress();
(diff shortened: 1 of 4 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
```diff
@@ -105,72 +106,65 @@ int connectCount() {
...
+                out.write((protocol + " 200 Connection established\r\n\r\n").getBytes(UTF_8));
+                out.flush();
+
+                executor.submit(() -> {
+                    try {
+                        copyStream(out, clientSocket.getInputStream());
+                    } finally {
+                        // We are simulating a proxy that doesn't do half closure. The proxy should close the server
+                        // socket as soon as the server read is done.
+                        socket.close();
```
idelpivnitskiy, line 143: 1. Thanks for leaving a comment here. We can even make it more visible with something like:
```
// ! NEVER CHANGE:
``` 
😄 

Consider also referencing which tests require this behavior: `GracefulConnectionClosureHandlingTest` and `ConnectionCloseHeaderHandlingTest`.

2. Do you mind reverting to how the closure worked before #1816? Because tests verify graceful closure triggered by both sides, we also need to close if client sends FIN. Better to keep closure logic inside `copyStream(...)`.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
```diff
@@ -105,72 +106,65 @@ int connectCount() {
...
     }
 
     private static void copyStream(final OutputStream out, final InputStream cin) throws IOException {
-        try {
-            int b;
-            while ((b = cin.read()) >= 0) {
-                out.write(b);
-            }
+        int read;
+        final byte[] bytes = new byte[256];
```
idelpivnitskiy, line 157: As described in https://github.com/apple/servicetalk/pull/1817#discussion_r705563782, we need to fragment data as much as possible. Consider keeping it as is (read by one byte) or use a small buffer, around 8 bytes. The overall size of transferred data in tests is small. I don't think it's a perf concern or something that can make CI slower.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
```diff
@@ -105,72 +106,65 @@ int connectCount() {
...
             out.flush();
-        } finally {
-            try {
-                cin.close();
-            } finally {
-                out.close();
-            }
         }
+        // Don't close either Stream! We are simulating a proxy that doesn't do half closure, and close the socket
+        // outside the scope of this method.
```
idelpivnitskiy, line 163: Closing Input and Output streams doesn't do half-closure. Only `shutdownInput/Output` does.

Scottmitch: Agreed. this comment isn't meant to imply closing a stream does half closure. I'll remove this part of the comment.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
Scottmitch: I think it is more clear to close the socket instead of each individual stream (which maybe confused to mean half-closure is desired). The client and server socket will always be closed at the conclusion of copying (either in try/finally for `clientSocket`, or in a `finally` outside the `handleRequest` method).

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
Scottmitch: used 8, and added a comment

Scottmitch:
Related test failures:
- https://github.com/apple/servicetalk/issues/1507

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ProxyTunnel.java
```diff
@@ -105,72 +106,76 @@ int connectCount() {
...
+                        } catch (IOException e) {
+                            LOGGER.debug("Error closing serverSocket", e);
+                        }
+                    }
+                });
+                copyStream(clientSocket.getOutputStream(), serverSocket.getInputStream());
+
+                // Don't wait on the clientSocket input to serverSocket output copy to complete. We want to simulate a
+                // proxy that doesn't do half closure and that means we should close as soon as possible. ServiceTalk's
+                // CloseHandler should handle this gracefully (and delay FIN/RST until requests/responses complete).
```
idelpivnitskiy, line 158: Consider referencing test classes that depend on this expectation: GracefulConnectionClosureHandlingTest and ConnectionCloseHeaderHandlingTest

idelpivnitskiy (approved):
Even if it doesn't solve all flaky tests, it's certainly an improvement, thank you!

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1630

Scottmitch:
0.41 (1d22cecb3)

PR #1823, opened 2021-09-10
Deprecate one-way boolean settings on builders (#1750)

chemicL:
Cherry-picking https://github.com/apple/servicetalk/pull/1750 onto 0.41 branch.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
@@ -66,7 +66,56 @@ public abstract class SingleAddressHttpClientBuilder<U, R>
     @Override
     public abstract SingleAddressHttpClientBuilder<U, R> protocols(HttpProtocolConfig... protocols);
 
+    /**
+     * Disables automatically setting {@code Host} headers by inferring from the address or {@link HttpMetaData}.
+     * <p>
+     * This setting disables the default filter such that no {@code Host} header will be manipulated.
+     *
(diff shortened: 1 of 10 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
```diff
@@ -66,7 +66,56 @@
...
+     *     builder.appendConnectionFilter(filter1).appendConnectionFilter(filter2).appendConnectionFilter(filter3)
+     * </pre>
+     * making a request to a connection wrapped by this filter chain the order of invocation of these filters will be:
+     * <pre>
+     *     filter1 ⇒ filter2 ⇒ filter3 ⇒ connection
+     * </pre>
+     * @param factory {@link StreamingHttpConnectionFilterFactory} to decorate a {@link StreamingHttpConnection} for the
+     * purpose of filtering.
+     * @return {@code this}
+     */
```
Scottmitch, line 118: I think this should just use `@Override` instead of duplicating javadocs.

Scottmitch (changes requested):
lgtm after comment resolved

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
chemicL: Not duplicated though, as this got retrofitted during the cherry-pick due to conflict in edited lines. This actual method is not present in the parent class in the main branch. I will remove the Javadoc, leave only the changes related to the deprecation :)

PR #1827, opened 2021-09-13
Use Netty BOM in module dependencies

bondolo:
Motivation:
Netty provides a BOM module which can be used for managing the versions
of dependencies.
Modifications:
Use `platform` to import Netty BOM and eliminate explicit version from
Netty components.
Result:
Less boilerplate for version management of Netty components.


servicetalk-dns-discovery-netty/build.gradle
```diff
@@ -26,10 +26,11 @@ dependencies {
   implementation project(":servicetalk-transport-netty")
   implementation project(":servicetalk-transport-netty-internal")
   implementation "com.google.code.findbugs:jsr305:$jsr305Version"
-  implementation "io.netty:netty-resolver-dns:$nettyVersion"
+  implementation(platform("io.netty:netty-bom:$nettyVersion"))
+  implementation "io.netty:netty-resolver-dns"
   implementation "org.slf4j:slf4j-api:$slf4jVersion"
 
-  runtimeOnly "io.netty:netty-resolver-dns-native-macos:$nettyVersion:osx-x86_64"
+  runtimeOnly( group:"io.netty", name:"netty-resolver-dns-native-macos", classifier:"osx-x86_64")
```
bondolo, line 33: Switching to the "map" syntax is necessary for omitting version as empty version in `foo:bar::classifier` is not allowed.

servicetalk-http-netty/build.gradle
```diff
@@ -34,8 +34,9 @@ dependencies {
   implementation project(":servicetalk-transport-netty-internal")
   implementation project(":servicetalk-utils-internal")
   implementation "com.google.code.findbugs:jsr305:$jsr305Version"
-  implementation "io.netty:netty-codec-http:$nettyVersion"
-  implementation "io.netty:netty-codec-http2:$nettyVersion"
+  implementation(platform("io.netty:netty-bom:$nettyVersion"))
```
chemicL, line 37: I don't know if there is any best practice here, but perhaps putting the platform dependency at the top would provide a cleaner outcome?

chemicL:
LGTM, only a minor suggestion about structure.

servicetalk-http-netty/build.gradle
bondolo: That is the way I started but then opted to move it to the other Netty lines to produce smaller diffs.

Now does seem like a good time to choose a convention. I would be fine with either "platforms at the top" or "platforms before first use".

bondolo: Now placing all BOM declarations at top of dependencies section.

bondolo:
This will remain draft until Netty `4.1.69.FINAL` is released so that we can include tcnative version references.

chemicL (approved):

servicetalk-buffer-api/build.gradle
```diff
@@ -17,12 +17,14 @@
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-library"
 
 dependencies {
+  testImplementation enforcedPlatform("org.junit:junit-bom:$junit5Version")
```
bondolo, line 20: The tighter `enforcedPlatform` import is used only for test dependencies.

idelpivnitskiy:
The snapshot build failed with https://github.com/apple/servicetalk/runs/3863471492 after this was merged:
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':servicetalk-concurrent-api:generateMetadataFileForMavenJavaPublication'.
> Invalid publication 'mavenJava':
    - Variant 'testFixturesRuntimeElements' contains a dependency on enforced platform 'org.junit:junit-bom'
  In general publishing dependencies to enforced platforms is a mistake: enforced platforms shouldn't be used for published components because they behave like forced dependencies and leak to consumers. This can result in hard to diagnose dependency resolution errors. If you did this intentionally you can disable this check by adding 'enforced-platform' to the suppressed validations of the :servicetalk-concurrent-api:generateMetadataFileForMavenJavaPublication task.https://docs.gradle.org/7.2/userguide/publishing_setup.html#sec:suppressing_validation_errors

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 43s
```

bondolo:
**0.41** - 8e9d0e631bab8a43810345416debf1e9779275d2

PR #1830, opened 2021-09-14
Deprecating `HttpServerBuilder` protected methods

chemicL:
Motivation:

Upcoming refactoring of `HttpServerBuilder` to become an interface will
remove some methods from public API. Therefore, those methods should be
deprecated.

Modifications:

- Deprecated `HttpServerBuilder#doListen` - both variants,
- Deprecated `HttpServerBuilder#buildExecutionContext`.

Result:

Preparing potential users to the removal of these methods in upcoming
release.


bondolo (approved):
Seems sufficient since it is very unlikely that either method has implementations other than `DefaultHttpServerBuilder`

chemicL:
Cherry-pick for 0.41: https://github.com/apple/servicetalk/commit/f3261db3db571d2718b5d7e2bdcc873e1eaaf6fd

PR #1831, opened 2021-09-15
Identify `LoadBalancer` instance using a `targetResource(String)`

chemicL:
Motivation:

An instance of `LoadBalancer` will do connection selection over a
collection of hosts provided by a `Publisher` of resolved addresses.
While debugging issues it helps to identify the intended resource that
the `LoadBalancer` is helping to reach. Unfortunately, at it's level,
the unresolved address is unknown and it may be the case that the
collection of hosts represents more than a single unresolved address. To
make the debugging process more insightful, a `String` parameter has
been added, named `targetResource` to identify a particular instance.

Modifications:

- Added a `String` parameter to `LoadBalancerFactory`'s
  `newLoadBalancer` method: `targetResource` in the form of a new
  default method in the interface to avoid breaking the API and the
  contract of `FunctionalInterface`,
- Using the address from
  `io.servicetalk.http.netty.DefaultSingleAddressHttpClientBuilder.HttpClientBuildContext`
  when creating instances of `LoadBalancer` in the
  DefaultSingleAddressHttpClientBuilder`.

Result:

Debugging connectivity issues originating from unhealthy state of
hosts behind the `LoadBalancer` should be easier as it is now possible
to identify the target resource that the `LoadBalancer` is connecting
to.


servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
```diff
@@ -35,8 +35,36 @@
...
             ConnectionFactory<ResolvedAddress, T> connectionFactory);
+
+    /**
+     * Create a new {@link LoadBalancer}.
+     * <p>
+     * Note this method has a default implementation to not break the {@link FunctionalInterface} contract, however
+     * in a future release the other deprecated {@link #newLoadBalancer(Publisher, ConnectionFactory) method}
+     * will be removed and this method will become the functional interface.
+     * @param targetResource A {@link String} representation of the target resource for which the created instance
+     * will perform load balancing. Bear in mind load balancing is performed over the a collection of hosts provided
```
tkountis, line 53: Consider comma after "Bear in mind"

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -323,7 +323,9 @@ public StreamingHttpClient buildStreaming() {
             }
 
             final LoadBalancer<LoadBalancedStreamingHttpConnection> lb =
-                    closeOnException.prepend(ctx.builder.loadBalancerFactory.newLoadBalancer(sdEvents,
+                    closeOnException.prepend(ctx.builder.loadBalancerFactory.newLoadBalancer(
+                            ctx.address().toString(),
```
idelpivnitskiy, line 327: 1. How users can override this value if they are not satisfied with what we pass automatically?
2. For a scenario when users use the same proxy to connect to any address (think of 100s different targets), the value returned from `ctx.address()` will be useless as it always indicates a proxy. Consider writing a method that will mix `ctx.proxyAddress` and `ctx.builder.address` together. It will provide a full picture of what is the target resource and how we connect (directly or via proxy).

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
```diff
@@ -35,8 +35,36 @@
      * and hence will call {@link ConnectionFactory#closeAsync()} when {@link LoadBalancer#closeAsync()} is called.
      * @param <T> Type of connections created by the passed {@link ConnectionFactory}.
      * @return a new {@link LoadBalancer}.
+     * @deprecated In the future only {@link #newLoadBalancer(String, Publisher, ConnectionFactory)} will remain,
+     * please use that method instead.
```
idelpivnitskiy, line 39: Sad that we can not add a `default` impl for the deprecated method because it's a functional interface. Looks like we will need to swap `default` in 0.42 and will be able to remove this method only in 0.43. Do we need to provide any recommendations how users can migrate (what to do with the old method)?

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
```diff
@@ -35,8 +35,36 @@
...
+     * connect to physical hosts. Typically generated from a {@link ServiceDiscoverer}.
+     * @param connectionFactory {@link ConnectionFactory} that the returned {@link LoadBalancer} will use to generate
+     * new connections. Returned {@link LoadBalancer} will own the responsibility for this {@link ConnectionFactory}
+     * and hence will call {@link ConnectionFactory#closeAsync()} when {@link LoadBalancer#closeAsync()} is called.
+     * @param <T> Type of connections created by the passed {@link ConnectionFactory}.
+     * @return a new {@link LoadBalancer}.
+     */
+    default <T extends C> LoadBalancer<T> newLoadBalancer(
+            String targetResource,
+            Publisher<? extends ServiceDiscovererEvent<ResolvedAddress>> eventPublisher,
```
idelpivnitskiy, line 66: We have another task on the short roadmap to change this input parameter from `Publisher<SDE>` to `Publisher<Collection<SDE>>` to make it consistent with `SD#discover(...)` API. It will force us to deprecate this method one more time. To minimize deprecations in 0.41 we should not cherry-pick this change into 0.41 until both PRs are ready.

idelpivnitskiy (commented):
We should also include the `targetResource` to log messages. Most of them automatically get it through `this.toString()`, but some of them don't:
1. line 584: `LOGGER.debug("Closing {} connection(s) gracefully to closed address: {}", toRemove.length, address);`
2. line 643: `LOGGER.debug("Active host for address {} failed to open {} connections"...`
3. line 655: `LOGGER.debug("Triggering health check for address {} after {} failed attempts"`
4. line 772: `LOGGER.debug("Health check for address {} cancelled.", healthCheck.host.address);`
5. 3 log statements inside `HealthCheck#reconnect()` method.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
chemicL: It seems because the functional interface's method is generic on its' own, the users can't use lambdas with it. Only method references. They should get compiler warnings and will be able to change their signatures to accept the String parameter and use the reference to the default method. WDYT?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
chemicL: 1. Fair point. They can't. I'm not sure we need to provide that flexibility at this point however.
2. Thank you for pointing that out. I just added a utility to combine the two.

Having both points in mind - I'm not necessarily in favor of over-engineering this feature at this time. It's a simple string at this moment, meant for debugging purposes. If we need users to have more flexibility, we could consider more API changes, but my intention was to bring useful functionality with absolutely minimum level of inconvenience. Most users should not notice the change. I'm open to suggestions though - perhaps we can do something without too many changes? I can't see it at the moment, but if there are ways - let's go for it.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -417,14 +454,16 @@ public void onComplete() {
...
                     }
                     if (host.addConnection(newCnx)) {
                         return succeeded(newCnx);
                     }
-                    return newCnx.closeAsync().concat(this.usedHosts == CLOSED_LIST ? failedLBClosed() :
+                    return newCnx.closeAsync().concat(this.usedHosts == CLOSED_LIST ? failedLBClosed(targetResource) :
                             failed(new ConnectionRejectedException(
-                                    "Failed to add newly created connection for host: " + host)));
+                                    "Failed to add newly created connection for " + targetResource
+                                            + " for host: " + host)));
```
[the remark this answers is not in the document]
chemicL: Please note this part of the exception has not changed in this PR.

Answering the question though, I believe it's essential to capture the state of the host and current number of connections upon the rejection. Not including that information might leave the users in the dark while debugging. Especially as we should only see rejections happen when the host is in closed state. Not including the state would leave some questions unanswered.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -417,14 +454,16 @@ public void onComplete() {
...
+                                        + " was rejected by the selection filter.")));
                     }
                     if (host.addConnection(newCnx)) {
                         return succeeded(newCnx);
                     }
-                    return newCnx.closeAsync().concat(this.usedHosts == CLOSED_LIST ? failedLBClosed() :
+                    return newCnx.closeAsync().concat(this.usedHosts == CLOSED_LIST ? failedLBClosed(targetResource) :
                             failed(new ConnectionRejectedException(
-                                    "Failed to add newly created connection for host: " + host)));
+                                    "Failed to add newly created connection for " + targetResource
```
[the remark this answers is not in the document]
chemicL: I'm not sure which one to choose: `newCnx` or `host`? Or both?
In fairness, the fact that the actual `newCnx` would include the resolved address is an implementation detail of a particular subclass. On `host` we can rely, as it's a private static class here.
Having those stated, it seems to me we should include both the `host` and the `newCnx` in the log. WDYT?

chemicL:
> We should also include the `targetResource` to log messages. Most of them automatically get it through `this.toString()`, but some of them don't:
> 
> 1. line 584: `LOGGER.debug("Closing {} connection(s) gracefully to closed address: {}", toRemove.length, address);`
> 2. line 643: `LOGGER.debug("Active host for address {} failed to open {} connections"...`
> 3. line 655: `LOGGER.debug("Triggering health check for address {} after {} failed attempts"`
> 4. line 772: `LOGGER.debug("Health check for address {} cancelled.", healthCheck.host.address);`
> 5. 3 log statements inside `HealthCheck#reconnect()` method.

I intentionally didn't include this information here. Note all of those logs come from the private static classes. They don't see the owner's `targetResource`. That would need to be passed to them. It's feasible of course, but I noticed that every time we actually log the particular `address` – which we assume (as an implementation detail) will identify both the resolved and unresolved address. Also, these are not the exceptions that escape the scope that we control - they are debug logs meant for deeper analysis. I'm ok treating the underlying assumption that `address` contains the unresolved address in its details in case of debug logs as binding. I just prefer that assumption not to surface the receiver of exceptions and hence only enriched the exceptions that are part of the API.

All in all, I'm happy to enrich those debug logs with the `targetResource` for full context and consistency's sake :-)

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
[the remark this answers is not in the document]
idelpivnitskiy: Both will be too much because they duplicate the address, and `Host#toString()` contains too much info that is not required for an exception message. Let's pick `newCnx`. We have a test that makes sure its `toString()` is consistent: #1746.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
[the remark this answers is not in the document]
idelpivnitskiy: Fair, let's keep both then (`newCtx` and `host`) in this exception message because it's also important to know what connection id was rejected.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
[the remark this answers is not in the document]
idelpivnitskiy: After reading your comment in the next discussion and thinking about it one more time, doesn't look like we need to make these two exception messages consistent. The "Newly created connection rejected" is rejected by the selector and doesn't need to tell anything about the host. "Failed to add newly created connection for host" is rejected by the host, it's better to include both (`newCtx` and `host`) in this case to let users know the connection id and the `host` state.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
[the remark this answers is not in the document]
idelpivnitskiy: Because `Host#toString()` already says that it's the `Host`, consider:
```
"Failed to add newly created connection " + newCnx + " for " + targetResource + " for " + host
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
idelpivnitskiy: +1, let's not overcomplicate API now. If necessary, users can wrap `LoadBalancerFactory` and then `LoadBalancer` interface to intercept the passed `targetResource` and change it.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
idelpivnitskiy: No, they won't be able to use lambda for a new method while it still has `default` impl. No worries, we will eventually clean it up, it will just take one more release to do so.

idelpivnitskiy:
@chemicL thanks for clarifying! Agreed, let's keep these logs as is.

idelpivnitskiy (approved):
This PR is good to go after the "Failed to add a newly created connection for" exception message includes `newCnx`. Thanks for a nice improvement!
Just don't cherry-pick to 0.41 until we have a second change for SD publisher.

chemicL:
Cherry-picked for 0.41: 29d4c4c7302a572804c682396fc2c3af25548d40

PR #1832, opened 2021-09-17
Add `ExecutionStrategyInfluencer`

bondolo:
Motivation:
To allow for simpler determination of the required execution strategy
for collections of client and server code.
Modifications:
Introduce `ExecutionStrategyInfluencer` and `requiredOffloads()` as
a replacement for the `influenceStrategy` method of
`HttpExecutionStrategyInfluencer` which is also deprecated.
Result:
A path towards simplified and more transparent execution
strategy computations.


bondolo:
Draft as there are still two test failures I have been unable to resolve.

bondolo:
A little more context to this PR. 

This work is being done as a prerequisite to adding additional `ExecutionStrategyInfluencer` sub-types beyond the existing `HttpExecutionStrategyInfluencer` type.

Some of the contortions about converting many implements to `ExecutionStrategyInfluencer<HttpExecutionStrategy>` are due to the difficulty of deprecating a functional interface. It is expected that for `requiredOffloads` either a bound method reference, a static method reference or a constant is used in almost all cases. Usage of the `makeInfluencer` utility should be rare as it is undesirable to remove the context of the object making the influencing claim. ie. if you are an influencer, implement the interface, implement the `requiredOffloads` method. Use of the "bare" influencer implementations produced by `makeInfluencer` should be restricted to non-influencers or special cases.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategy.java
```diff
@@ -82,7 +82,7 @@ public HttpExecutionStrategy merge(final HttpExecutionStrategy other) {
             return this;
         }
 
-        if (other instanceof SpecialHttpExecutionStrategy) {
+        if (!(other instanceof DefaultHttpExecutionStrategy)) {
```
Scottmitch, line 85: One of the big motivators for introducing this change was to unify the merge/difference process (remove special case instanceof and static equality checks). are any other changes required to make the simplification?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategy.java
```diff
@@ -60,4 +59,32 @@ default boolean hasOffloads() {
...
+
+    /**
+     * Return An execution strategy which contains only the additional offloads present in other that were not present
+     * in this execution strategy.
+     *
+     * @param other An execution strategy
+     * @return An execution strategy which contains only the additional offloads present in other that were not present
+     * in this execution strategy.
+     */
+    default HttpExecutionStrategy missing(HttpExecutionStrategy other) {
```
Scottmitch, line 71: This method isn't used, what is the intended use case for this method? does this need to be an instance method? If not can we move this into a static factory method to keep this interface focused on state?

Ideally we could do the same with `merge` after we remove special case combinations.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
```diff
@@ -40,4 +56,36 @@
     static HttpExecutionStrategyInfluencer defaultStreamingInfluencer() {
         return DEFAULT_STREAMING_STRATEGY_INFLUENCER;
     }
+
+    static HttpExecutionStrategy noInfluence() {
+        return DefaultHttpExecutionStrategy.OFFLOAD_NONE_STRATEGY;
+    }
+
+    static HttpExecutionStrategyInfluencer nonInfluencer() {
```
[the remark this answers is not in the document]
Scottmitch: also add javadocs for all new public methods.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -619,29 +623,18 @@ private static StreamingHttpService buildService(Stream<StreamingHttpServiceFilt
     }
 
     private static HttpExecutionStrategyInfluencer buildInfluencer(List<StreamingHttpServiceFilterFactory> filters,
```
Scottmitch, line 625: add `@SuppressWarnings("unchecked")` for the cast to `ExecutionStrategyInfluencer<HttpExecutionStrategy>`.

also is it the plan to have `StreamingHttpServiceFilterFactory extend ExecutionStrategyInfluencer<HttpExecutionStrategy>`? making what is an influencer more explicit is one of the original goals IIRC.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StrategyInfluencerAwareConversions.java
```diff
@@ -30,76 +33,98 @@ static StreamingHttpServiceFilterFactory toConditionalServiceFilterFactory(
...
+                    public HttpExecutionStrategy requiredOffloads() {
+                        return requiredOffloads;
+                    }
+
+                    @Override
+                    public StreamingHttpServiceFilter create(final StreamingHttpService service) {
+                        return new ConditionalHttpServiceFilter(predicate, original.create(service), service);
+                    }
+                } :
+                service -> new ConditionalHttpServiceFilter(predicate, original.create(service), service);
```
Scottmitch, line 53: if the returned value isn't a strategy influencer, couldn't it be treated as default strategy (offload all) later on?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StrategyInfluencerChainBuilder.java
```diff
@@ -40,54 +43,72 @@ public StrategyInfluencerChainBuilder() {
...
         }
         return false;
     }
 
+    /**
+     * Append another {@link ExecutionStrategyInfluencer} to this chain.
+     *
+     * @param next {@link ExecutionStrategyInfluencer} to append.
+     */
+    public void append(ExecutionStrategyInfluencer next) {
```
Scottmitch, line 90: `ExecutionStrategyInfluencer` -> `ExecutionStrategyInfluencer<?>` will avoid some class cast warnings (general comment in this class)

Scottmitch: since this class is in `http.api` can it be type safe on `ExecutionStrategyInfluencer<HttpExecutionStrategy>`?

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/JerseyRouteExecutionStrategy.java
```diff
@@ -74,14 +74,11 @@ public boolean isSendOffloaded() {
 
     @Override
     public HttpExecutionStrategy merge(final HttpExecutionStrategy other) {
-        return delegate.merge(other);
+        HttpExecutionStrategy merged = delegate.merge(other);
+        return merged == delegate ? this : new JerseyRouteExecutionStrategy(merged, executor);
```
Scottmitch, line 78: seems we are getting more specialized in `merge`, does this mean we won't be able to make `merge` a non-instance method?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
```diff
@@ -40,4 +56,36 @@
     static HttpExecutionStrategyInfluencer defaultStreamingInfluencer() {
         return DEFAULT_STREAMING_STRATEGY_INFLUENCER;
     }
+
+    static HttpExecutionStrategy noInfluence() {
```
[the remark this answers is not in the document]
bondolo: The two constants that are there are both destined to be removed. In the interface seemed like a more discoverable location for constants.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategy.java
Scottmitch: why does it need to be an instance method? can we defer adding it until it is used to have a more clear picture on intended use and how it will replace `diff`?

bondolo:
CI failure attributed to #1299

PR #1833, opened 2021-09-18
Add support for nonOffloading HTTP Server Filters (#1710)

bondolo:
Motivation:
Some filters may wish to run early in the request process before the
request has been offloaded.
Modifications:
A new API,
`HttpServerBuilder.appendNonOffloadingServiceFilter(StreamingHttpServiceFilterFactory)`
is provided for registering filters which will not be offloaded.
Offloading will be handled in the filter chain by the new
`OffloadingFilter` which is logically inserted between the
non-offloading and offloading filters. The new offloading filter also replaces
one of the `HttpExecutionStrategy.invokeService` methods.
Result:
New options for filters.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
@@ -16,7 +16,6 @@
 package io.servicetalk.http.api;
 
 import io.servicetalk.buffer.api.BufferAllocator;
-import io.servicetalk.concurrent.api.AsyncContext;
 import io.servicetalk.concurrent.api.Executor;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.http.api.HttpApiConversions.ServiceAdapterHolder;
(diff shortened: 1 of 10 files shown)


bondolo:
It builds and passes tests. I would prefer if it contained fewer changes to files outside of the `HttpServerBuilder`, some of which can be easily trimmed, but other minimization changes would require more work.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/OffloadingFilter.java
```diff
@@ -0,0 +1,81 @@
...
+                                                        final StreamingHttpResponseFactory responseFactory) {
+                Executor se = strategy.executor();
+                Executor e = null != se ? se : ctx.executionContext().executor();
+
+                // The service should see our ExecutionStrategy inside the ExecutionContext:
+                final HttpServiceContext wrappedCtx =
+                        new ExecutionContextOverridingServiceContext(ctx, strategy, e);
+
+                if (strategy.isDataReceiveOffloaded()) {
+                    request = request.transformMessageBody(p -> p.publishOn(e));
```
idelpivnitskiy, line 56: IIRC, with the 0.41 behavior of `publishOn` operator, the move of offloading logic from `NettyHttpServer` to this filter won't change the fact that offloading is applied at the beginning of this publisher chain. If this is the case, let's keep the code here to simplify `NettyHttpServer`, but hide the `appendNonOffloadingServiceFilter` methods on `HttpServerBuilder` to prevent users from using them. We can clarify in the PR title/message that this change doesn't add a user-facing feature.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -481,23 +542,106 @@ public final ServerContext listenBlockingStreamingAndAwait(
...
+            boolean anyOffloads = strategy.isSendOffloaded() ||
+                    strategy.isMetadataReceiveOffloaded() ||
+                    strategy.isDataReceiveOffloaded();
+
+            Stream<StreamingHttpServiceFilterFactory> nonOffloadingFilters = noOffloadServiceFilters.stream();
+
+            if (anyOffloads) {
+                // We are going to have to offload, even if just to the raw service
+                nonOffloadingFilters = Stream.concat(nonOffloadingFilters,
+                        Stream.of(new OffloadingFilter(strategy, buildFactory(serviceFilters))));
```
idelpivnitskiy, line 582: After adding offloading here, this PR should remove it from NettyHttpServer, like we did in the original PR: https://github.com/apple/servicetalk/pull/1710/files#diff-b6703aba79e18b2b8d87947010c3580135eb31bf596c832beb9e30cb2b52203f

idelpivnitskiy (commented):
@bondolo thank you for taking care of it ❤️ 
Let's discuss what to do with public methods. Even without exposing this feature to users, I still think this is an improvement for 0.41 if we can move offloading to a filter. IIUC, the behavior won't change, right?

PR #1834, opened 2021-09-20
Deprecate `AwaitUtils#awaitUninterruptibly`

Krupskis:
Motivation:

The current usage of `awaitUtils` catches
`InterruptedException` and silently discards it. jUnit5 uses
result, these tests may hang forever as the `InterruptedException`
is not propagated.

Modifications:
- Rewrite `awaitUninterruptibly` to await in `awaitUtils`

Result:
Deprecated awaitUninterruptibly. await now throws unchecked
Exception with the help of PlatformDependent.throwException.

It has been verified that a test that takes more than 10 seconds
and uses `awaitUtils.await` will fail with `java.util.concurrent.TimeoutException.`

### DIFF
--- a/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
+++ b/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
@@ -84,9 +84,8 @@ public final class AwaitUtils {
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             throwException(e);
+            return false;
         }
-
-        return false;
     }
 
     /**
--- a/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
+++ b/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
@@ -20,6 +20,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 
+import static io.servicetalk.utils.internal.PlatformDependent.throwException;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 
 /**
@@ -31,9 +32,24 @@ public final class AwaitUtils {
     }
 
     /**
-     * Await a {@link CountDownLatch} while suppressing {@link InterruptedException} until the latch is counted down.
+     * Await a {@link CountDownLatch} until the latch is counted down, throws unchecked InterruptedException.
      * @param latch the {@link CountDownLatch} to await.
      */
+    public static void await(CountDownLatch latch) {
+        try {
+            latch.await();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throwException(e);
+        }
+    }
+
+    /**
+     * Await a {@link CountDownLatch} while suppressing {@link InterruptedException} until the latch is counted down.
+     * @param latch the {@link CountDownLatch} to await.
+     * @deprecated use {@link #await(CountDownLatch)} instead
+     */
+    @Deprecated
     public static void awaitUninterruptibly(CountDownLatch latch) {
         boolean interrupted = false;
         try {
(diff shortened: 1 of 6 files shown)


servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
```diff
@@ -30,10 +30,22 @@ private AwaitUtils() {
...
     }
 
+    /**
+     * Await a {@link CountDownLatch} until the latch is counted down.
+     * @param latch the {@link CountDownLatch} to await.
+     * @throws InterruptedException if the current thread is interrupted while waiting or has its interrupted status set on
+     * entry to this method
+     */
+    public static void await(CountDownLatch latch) throws InterruptedException {
+        latch.await();
```
idelpivnitskiy, line 40: I see the pain with adding `throws Exception` to all tests that don't have it. Thanks for highlighting it and opening a PR before you finished to change all files.

Instead of changing gazillion test files, we can re-throw it bypassing compiler checks for checked exceptions:
```java
    public static void await(CountDownLatch latch) {
        try {
            latch.await();
            throwException(e);
        }
```
Where `throwException` is statically imported from `io.servicetalk.utils.internal.PlatformDependent` (you will need to add an `implementation` dependency for `servicetalk-utils-internal`).
This is a dirty hack, but if you can verify this approach work for jUnit5, I'm fine with it. We can improve it over time. More important is to make sure jUnit5 can terminate tests when they exceed timeout. 

To verify, try to write a test that does something that never completes:
```java
AwaitUtils.await(new CountDownLatch(1));
```
and make sure jUnit5 fails this test with a timeout exception.

idelpivnitskiy (commented):
Thank you for opening a PR @Krupskis!

> I tried manually rewriting test method signatures, but it is too tedious, as so many
methods depend on each other and the Exception just gets propagated through
all of them. Is there a way I could automate this or is there any way
to make this easier?

I'm not aware of any automated way. I hoped Intellij IDEA refactoring tool can manage it, but it doesn't it in other way we don't want. It just adds `try-catch-printStackTrace` to all methods that invoke this method without asking if this is what is expected 😞 

> How would I handle anonymous functions, which do not have
method signatures? How I handle them right now can be seen in
`ConnectableBufferOutputStreamTest.java`.

Good question. Sometimes it's possible to use an interface that throws (like using `Callable` instead of `Runnable`), but this is not possible if test has expectations of a specific interface behind the lambda. In these cases, bypassing compiler checks (see my other comment below) is the only option I'm aware of. @bondolo did I miss anything?

Krupskis:
@idelpivnitskiy sorry for a late reply, had other stuff going on, but finally had some time to sit down to this. Your suggestion does seem like a great workaround! I have verified that the test that never completes indeed timeouts after 10 seconds with a `TimeoutException`. Thank you :)

I left several comments on my code which I think could be improved or changed.

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
```diff
@@ -30,10 +31,25 @@ private AwaitUtils() {
         // no instances
     }
 
+    /**
+     * Await a {@link CountDownLatch} until the latch is counted down, throws unchecked InterruptedException.
```
Krupskis, line 35: Since the exception now is unchecked I could not include it in the method signature. This means that Javadoc cannot have `@throws`, otherwise, the build fails. I thought this could be a way to specify that the method now throws an unchecked exception, but I am not 100% sure.

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
```diff
@@ -52,14 +68,37 @@ public static void awaitUninterruptibly(CountDownLatch latch) {
...
+        final long timeoutNanos = NANOSECONDS.convert(timeout, unit);
+
+        try {
+            return latch.await(timeoutNanos, NANOSECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throwException(e);
+        }
+
+        return false;
```
Krupskis, line 89: This should return false if the exception is thrown, right?

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
idelpivnitskiy: Since this class is in "internal" package, don't worry about documentation. Let's leave it without `throws` in javadoc.

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
idelpivnitskiy: In reality, this point will never be reached. So, return value doesn't matter much, but it's nice to keep it as "false".

idelpivnitskiy, line 89:
```suggestion
            throwException(e);
            return false;
        }
```
We can move it under the `catch` block just to make it clear.

idelpivnitskiy (approved):

idelpivnitskiy:
Thank you @Krupskis!

idelpivnitskiy:
Related to #1684

PR #1835, opened 2021-09-20
`BeforeFinallyHttpOperator`: add option to discard signals after cancelation

idelpivnitskiy:
Motivation:

After `cancel`, async sources can stop propagating more signals. Therefore,
"beforeFinally" operators consider cancellation as a terminal event. However,
cancellation is the best effort and more signals can be delivered later or it may
race with other signals. In some scenarios, like observability, it may be confusing
to receive more signals after a terminal event. For those use-cases, a new option
is added for `BeforeFinallyHttpOperator` to discard further signals after cancel.

Modifications:

- Add a new ctor for `BeforeFinallyHttpOperator` that takes `discardEventsAfterCancel`;
- Enhance internal state machine to accommodate the new features;
- Add tests to verify new behavior;

Result:

Users can configure `BeforeFinallyHttpOperator` to discard further signals after
cancellation.

### DIFF
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/BeforeFinallyHttpOperator.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/BeforeFinallyHttpOperator.java
@@ -357,9 +357,9 @@ public final class BeforeFinallyHttpOperator implements SingleOperator<Streaming
                                                         PROCESSING_PAYLOAD, TERMINATED)) {
                                                     return state;
                                                 }
-                                            } else {
+                                            } else if (stateUpdater.compareAndSet(ResponseCompletionSubscriber.this,
+                                                    state, TERMINATED)) {
                                                 // re-entry, but we can terminate because this is a final event:
-                                                stateUpdater.set(ResponseCompletionSubscriber.this, TERMINATED);
                                                 return state;
                                             }
                                         }
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/BeforeFinallyHttpOperator.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/BeforeFinallyHttpOperator.java
@@ -62,6 +62,7 @@ import static java.util.concurrent.atomic.AtomicIntegerFieldUpdater.newUpdater;
 public final class BeforeFinallyHttpOperator implements SingleOperator<StreamingHttpResponse, StreamingHttpResponse> {
 
     private final TerminalSignalConsumer beforeFinally;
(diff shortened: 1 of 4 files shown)


servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/BeforeFinallyHttpOperator.java
```diff
@@ -142,44 +166,209 @@ public void request(final long n) {
...
+                                                    } finally {
+                                                        subscription.cancel();
+                                                    }
+                                                    return;
+                                                }
+
+                                                for (;;) {
+                                                    final int state = ResponseCompletionSubscriber.this.state;
+                                                    assert state != IDLE;
+                                                    if (state == PROCESSING_PAYLOAD) {
```
bondolo, line 185: Could use a switch(state) but then you would have to use named break to distinguish between case/break and for/break

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/BeforeFinallyHttpOperator.java
```diff
@@ -142,44 +166,209 @@ public void request(final long n) {
...
+                                                return state;
+                                            }
+                                            if (state == PROCESSING_PAYLOAD) {
+                                                if (stateUpdater.compareAndSet(ResponseCompletionSubscriber.this,
+                                                        PROCESSING_PAYLOAD, TERMINATED)) {
+                                                    return state;
+                                                }
+                                            } else {
+                                                // re-entry, but we can terminate because this is a final event:
+                                                stateUpdater.set(ResponseCompletionSubscriber.this, TERMINATED);
```
bondolo, line 362: Perhaps slightly safer to compare and set against the value we previously read? At worst it got updated to terminated and we do an extra volatile read to discover that.

servicetalk-http-utils/src/test/java/io/servicetalk/http/utils/BeforeFinallyHttpOperatorTest.java
```diff
@@ -1,5 +1,5 @@
 /*
```
bondolo, line 1: I really like these test improvements.

bondolo (approved):

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/BeforeFinallyHttpOperator.java
idelpivnitskiy: Tried to be consistent with pattern we use at any other operator

PR #1836, opened 2021-09-21
Revert "Add support for nonOffloading HTTP Server Filters (#1710)"

bondolo:
Reverts apple/servicetalk#1833


bondolo:
revert because #1833 merge was premature.

PR #1838, opened 2021-09-22
Prevent response publisher from being cancelled before `onComplete`

idelpivnitskiy:
If netty `Channel` closes after writing the last item of the outbound
publisher (like HTTP/2 stream closes after sending `endStream=true` flag),
`channelInactive` event may race with `onComplete`. If it comes first,
`WriteStreamSubscriber#channelClosed(Throwable)` will be invoked,
triggering cancellation on the subscription that is already effectively
terminated (because we saw `OutboundDataEndEvent` ->
`WriteStreamSubscriber#channelOutboundClosed` -> `SOURCE_TERMINATED`
flag is already set).
This may result in observing response cancellation from filters instead
of successful completion.

Modifications:

- Swap current `subscription` with `CANCELLED` after
`SOURCE_TERMINATED` flag is set in `WriteStreamSubscriber#AllPromise`;

Result:

Cancel is not propagated and filters see the successful completion of the
outbound publisher instead of a cancellation.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
@@ -366,6 +366,8 @@ final class WriteStreamSubscriber implements PublisherSource.Subscriber<Object>,
             }
             this.failureCause = cause;
             setFlag(SOURCE_TERMINATED);
+            // Mark the subscription as CANCELLED to prevent propagating cancel from channelClosed
+            subscriptionUpdater.getAndSet(WriteStreamSubscriber.this, CANCELLED);
             if (activeWrites == 0) {
                 try {
                     setFlag(SUBSCRIBER_TERMINATED);
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
@@ -366,8 +366,9 @@ final class WriteStreamSubscriber implements PublisherSource.Subscriber<Object>,
             }
             this.failureCause = cause;
             setFlag(SOURCE_TERMINATED);
-            // Mark the subscription as CANCELLED to prevent propagating cancel from channelClosed
-            subscriptionUpdater.getAndSet(WriteStreamSubscriber.this, CANCELLED);
+            // Mark the subscription as CANCELLED to prevent propagating cancel from channelClosed. At this point we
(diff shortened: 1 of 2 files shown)


idelpivnitskiy:
Real example: https://github.com/apple/servicetalk/runs/3669989669
Search for the logs related to:
```
GrpcLifecycleObserverTest > testResponseStream(boolean) > io.servicetalk.grpc.netty.GrpcLifecycleObserverTest.testResponseStream(boolean)[1]
```

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -366,6 +366,8 @@ void sourceTerminated(@Nullable Throwable cause) {
             }
             this.failureCause = cause;
             setFlag(SOURCE_TERMINATED);
+            // Mark the subscription as CANCELLED to prevent propagating cancel from channelClosed
+            subscriptionUpdater.getAndSet(WriteStreamSubscriber.this, CANCELLED);
```
Scottmitch, line 370: can this just be a regular volatile write since we don't need the return value? if not plz add a comment.

idelpivnitskiy: sure, we don't need the old value. how do you think, should we prevent from changing `null` to `CANCELLED`?

idelpivnitskiy: probably not, because we can not reach this state without emitting any signals from the publisher => we always have a subscription at this point

bondolo: Or just a `set()` if the intention is to avoid mixing use of field and updater.

idelpivnitskiy: `set` does additional work to validate access and check the value. We are not consistent in other places anyway.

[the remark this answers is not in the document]
idelpivnitskiy: No, all these changes to the internal state of `AllPromise` are happening on a single event loop thread.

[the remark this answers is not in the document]
idelpivnitskiy: Also, `channelOutboundClosed` and `channelClosed` are always sequential because they are on the same event loop thread. `channelOutboundClosed` comes from `userEventTriggered`, `channelClosed` comes from `channelInactive`.

bondolo (approved):

PR #1845, opened 2021-09-24
Fix race condition in SpliceFlatStreamToMetaSingle

Scottmitch:
Motivation:
SpliceFlatStreamToMetaSingle has a race condition which may result in
concurrency invocation of a Subscriber which is prohibited by the
Reactive Streams specification [1].

[1] https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#1.3

Modifications:
- Use a DelayedSubscription so onSubscribe is invoked before the
  Subscriber reference is made visible to other threads which may
  invoke terminal methods on the Subscriber.

Result:
SpliceFlatStreamToMetaSingle is more compliant with RS spec and provides
more predictable behavior.


idelpivnitskiy (approved):
🚀

Scottmitch:
0.41 (66eb87574)

Issue #245, opened 2019-01-11
NettyHttpServerTest.testErrorBeforeRead


Scottmitch:
reproduced locally on macOS during Gradle build.

```
org.junit.runners.model.TestTimedOutException: test timed out after 10000 milliseconds
	at sun.misc.Unsafe.park(Native Method)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
	at java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:442)
	at io.servicetalk.concurrent.api.PublisherAsBlockingIterable$SubscriberAndIterator.hasNext(PublisherAsBlockingIterable.java:162)
	at io.servicetalk.concurrent.api.PublisherAsBlockingIterable$SubscriberAndIterator.next(PublisherAsBlockingIterable.java:224)
	at io.servicetalk.http.netty.NettyHttpServerTest.testErrorBeforeRead(NettyHttpServerTest.java:452)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.mockito.internal.junit.JUnitRule$1.evaluateSafely(JUnitRule.java:52)
	at org.mockito.internal.junit.JUnitRule$1.evaluate(JUnitRule.java:43)
[...]

PR #1848, opened 2021-09-24
NettyHttpServerTest.testErrorBeforeRead handle exception

Scottmitch:
Motivation:
NettyHttpServerTest.testErrorBeforeRead has the server intentionally
write an error without first consuming the request content. The server
may close the connection before the client is done writting and if that
is the case a ClosedChannelException may be observed and fail the test.

Modifications:
- Allow ClosedChannelException when making a request

Result:
Fixes https://github.com/apple/servicetalk/issues/245

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/NettyHttpServerTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/NettyHttpServerTest.java
@@ -606,7 +606,11 @@ class NettyHttpServerTest extends AbstractNettyHttpServerTest {
         } catch (Throwable cause) {
             // The server intentionally triggers an error when it writes, if this happens before all content is read
             // the client may fail to write the request due to premature connection closure.
-            assertThat(cause, instanceOf(ClosedChannelException.class));
+            if (cause instanceof ExecutionException) {
+                assertThat(cause.getCause(), instanceOf(IOException.class));
+            } else {
+                assertThat(cause, instanceOf(IOException.class));
+            }
         } finally {
             assertConnectionClosed();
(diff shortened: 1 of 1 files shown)


Scottmitch:

chemicL (approved):

tkountis (approved):

Scottmitch:
0.41 (a1fb483)

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/NettyHttpServerTest.java
```diff
@@ -601,14 +595,22 @@ void testErrorBeforeRead(ExecutorSupplier clientExecutorSupplier,
...
             // Due to a race condition, the exception cause here can vary.
             // If the socket closure is delayed slightly
             // (for example, by delaying the Publisher.error(...) on the server)
             // then the client throws ClosedChannelException. However if the socket closure happens quickly enough,
             // the client throws NativeIoException (KQueue) or IOException (NIO).
             assertThat(e.getCause(), instanceOf(IOException.class));
+        } catch (Throwable cause) {
+            // The server intentionally triggers an error when it writes, if this happens before all content is read
+            // the client may fail to write the request due to premature connection closure.
+            if (cause instanceof ExecutionException) {
```
idelpivnitskiy, line 609: Alternative approach can be to use "flushOnEnd" strategy by converting `StreamingHttpRequest` to an aggregated one and back. Then we don't need to worry about write failures on the client-side.
This approach also good. Either way is great to fix this test.

idelpivnitskiy (approved):
Interesting observation

PR #1846, opened 2021-09-24
Avoid mock abstract class creation in test

Scottmitch:
Motivation:
OverlappingCapacityAwareEstimatorTest leverages a Mockito feature which
instantiates an abstract class as a mock object. This has been observed
to interact poorly with Mockito's `mock-maker-inline` feature and
results in false positive test failures.

Modifications:
- Use traditional mock objects injected into a concrete class that
  extends OverlappingCapacityAwareEstimator for test purposes.

Result:
OverlappingCapacityAwareEstimatorTest runs successfully regardless of
Mockito's `mock-maker-inline` feature being enabled.

PR #1850, opened 2021-09-24
Avoid error from connection closure when signals should be queued

Scottmitch:
Motivation:
The HTTP protocol knows the request/response processing is over in some
scenarios before the read or write Publisher terminates. For example if
content-length is 0 we know there is no payload, and if content-length
is present for HTTP/1.x we know there will be no trailers. Internally we
use an EmptyTrailers object as a marker to indicate "end of Publisher",
but this may result in premature errors in the event the protocol knows
we are done before internal state is updated, and socket close events
occur. This problem is exacerbated with offloading enabled and events
are processed on different threads.

Modifications:
- Remove the requirement that an EmptyTrailers object is always present
  to signify "end of Publisher". CloseHandler knows when the stream has
  ended and generates an event that ensures NettyChannelPublisher state
  has been updated.
- Increase robustness of HttpObjectEncoder to verify if content-length
  is present that the exact size must be written, and no non-empty
  trailers are allowed.

Result:
Fixes https://github.com/apple/servicetalk/issues/1507
Fixes https://github.com/apple/servicetalk/issues/1579


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/NettyHttpServerTest.java
```diff
@@ -584,12 +584,6 @@ void testSingleError(ExecutorSupplier clientExecutorSupplier,
     void testErrorBeforeRead(ExecutorSupplier clientExecutorSupplier,
                              ExecutorSupplier serverExecutorSupplier) throws Exception {
         setUp(clientExecutorSupplier, serverExecutorSupplier);
-        // Flaky test: https://github.com/apple/servicetalk/issues/245
```
Scottmitch, line 587: extracted into https://github.com/apple/servicetalk/pull/1848

servicetalk-transport-netty-internal/src/test/java/io/servicetalk/transport/netty/internal/OverlappingCapacityAwareEstimatorTest.java
```diff
@@ -18,137 +18,172 @@
 import org.junit.jupiter.api.Test;
 
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.function.LongSupplier;
+import java.util.function.ObjLongConsumer;
```
Scottmitch, line 21: extracted into https://github.com/apple/servicetalk/pull/1846

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/SpliceFlatStreamToMetaSingle.java
```diff
@@ -24,6 +24,7 @@
 import io.servicetalk.concurrent.api.PublisherToSingleOperator;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.concurrent.api.internal.SubscribablePublisher;
+import io.servicetalk.concurrent.internal.DelayedSubscription;
```
Scottmitch, line 27: extracted into https://github.com/apple/servicetalk/pull/1845

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverTest.java
```diff
@@ -250,7 +250,7 @@ void testRequestResponse(StreamingHttpRequest request, HttpResponseStatus expect
     @EnumSource(HttpProtocol.class)
     void serverFailsResponsePayloadBodyBeforeRead(HttpProtocol httpProtocol) throws Exception {
         setUp(httpProtocol);
-        testServerFailsResponsePayloadBody(SVC_ERROR_BEFORE_READ, false);
+        testServerFailsResponsePayloadBody(SVC_ERROR_BEFORE_READ, true);
```
Scottmitch, line 253: @idelpivnitskiy - inbound events are queued now so the server still reads complete even though write errors out. wdyt?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/LoggingCloseHandler.java
```diff
@@ -0,0 +1,125 @@
...
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPromise;
+
+import java.util.function.Consumer;
+
+import static io.servicetalk.logging.api.LogLevel.DEBUG;
+import static io.servicetalk.logging.slf4j.internal.Slf4jFixedLevelLoggers.newLogger;
+
+final class LoggingCloseHandler extends CloseHandler {
```
Scottmitch, line 30: this isn't currently used but was useful for debugging consistently between the two `CloseHandler` implementations so I figured I would leave it ... but I can also kill it.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -194,6 +197,15 @@ public void channelWritable() {
...
     public void channelOutboundClosed() {
         assert eventLoop.inEventLoop();
+        final Subscription sub = this.subscription;
+        if (sub != null) {
+            // Some protocols may know the source should terminate before the user is done writing all the content. For
+            // example if HTTP headers includes content-length: 0, the protocol knows it is done after writing the
+            // metadata, but the user may have also written empty buffer and/or empty trailers that still need to be
+            // consumed. If the FlushStrategy waits until the stream terminates we must request the remainder of content
+            // we may deadlock if we don't request enough onNext signals to see the terminal signal.
+            sub.request(Long.MAX_VALUE);
```
Scottmitch, line 207: @idelpivnitskiy - note this, not ideal but happy to discuss alternatives if you have ideas...

servicetalk-http-netty/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
Scottmitch, line 1: this makes it so Mockito can `spy` on `final` classes.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
```diff
@@ -81,7 +81,7 @@ static void writeBuffer(ChannelHandlerContext ctx, Object msg, ChannelPromise pr
         if (buffer.readableBytes() > 0) {
             ctx.write(new DefaultHttp2DataFrame(encodeAndRetain(buffer), false), promise);
         } else {
-            promise.setSuccess();
+            ctx.write(EMPTY_BUFFER, promise);
```
Scottmitch, line 84: @idelpivnitskiy - this was done so we can preserve order of promise completion relative to writes. in some cases we may not have a reference to the "last write" and may do something like `write(empty buffer).whenComplete(close)`.

I saw in a commit message related to this change mention of a race between cancel and write on the write stream subscriber, but you made a recent fix which seems related https://github.com/apple/servicetalk/commit/4cb71f59afecc3c1cd1d6fc0f050cd98c0134eb9 ... let me know if you still feel we have a race with this approach as that maybe concerning in other scenarios.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
```diff
@@ -76,13 +75,10 @@ final void writeMetaData(ChannelHandlerContext ctx, HttpMetaData metaData, Http2
...
-        Buffer buffer = (Buffer) msg;
-        if (buffer.readableBytes() > 0) {
-            ctx.write(new DefaultHttp2DataFrame(encodeAndRetain(buffer), false), promise);
-        } else {
-            promise.setSuccess();
-        }
+    static void writeBuffer(ChannelHandlerContext ctx, Buffer buffer, ChannelPromise promise) {
+        // Even if the Buffer is empty we still want to write it. This is to preserve order of notification for
+        // when writing empty payload in order to know when prior content is written (and then close the channel).
+        ctx.write(new DefaultHttp2DataFrame(encodeAndRetain(buffer), false), promise);
```
Scottmitch, line 81: original discussion: https://github.com/apple/servicetalk/pull/1850#discussion_r717129692

@idelpivnitskiy - this was done so we can preserve order of promise completion relative to writes. in some cases we may not have a reference to the "last write" and may do something like `write(empty buffer).whenComplete(close)`.

looks like netty shouldn't [write empty data frames to the wire](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java#L484-L488), was there a scenario that lead to writing the empty data frames to the wire?

I saw in a commit message related to this change mention of a race between cancel and write on the write stream subscriber, but you made a recent fix which seems related https://github.com/apple/servicetalk/commit/4cb71f59afecc3c1cd1d6fc0f050cd98c0134eb9 ... let me know if you still feel we have a race with this approach as that maybe concerning in other scenarios.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
Scottmitch: discussion moved to https://github.com/apple/servicetalk/pull/1850#discussion_r717859989

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpTransportObserverTest.java
idelpivnitskiy: IIUC this happens only when an incoming msg is already read from the socket and we fail a write later. If we can still complete reading this is a nice improvement!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractH2DuplexHandler.java
idelpivnitskiy: Thanks for clarifying. Consider creating a static constant as `new DefaultHttp2DataFrame(false)` to avoid allocation on empty buffers.
We can also use a constant for `new DefaultHttp2DataFrame(true)` in line 100.

Scottmitch: I considered this but `DefaultHttp2DataFrame` has state (`Http2FrameStream` state & id) that Netty uses internally.

idelpivnitskiy: True, forgot to check that `AbstractHttp2StreamFrame` has a non-final field.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/RequestResponseCloseHandler.java
```diff
@@ -95,26 +98,13 @@
 
         byte ALL_CLOSED = CLOSED | IN_CLOSED | OUT_CLOSED;
         byte IN_OUT_CLOSED = IN_CLOSED | OUT_CLOSED;
-        byte MASK_IDLE = READ | WRITE;
+        byte READ_WRITE = READ | WRITE;
+        byte DISCARDING_SERVER_IN_CLOSED = DISCARDING_SERVER_INPUT | IN_CLOSED;
+        byte GRACEFUL_SERVER_OUT_CLOSED = OUT_CLOSED | CLOSING_SERVER_GRACEFULLY;
+        byte GRACEFUL_SERVER_IN_CLOSED = IN_CLOSED | CLOSING_SERVER_GRACEFULLY;
```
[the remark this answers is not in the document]
Scottmitch: I'll add `_OR_` so we can use common bit set utilities.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HeaderUtils.java
```diff
@@ -107,6 +104,16 @@ private static boolean canAddContentLength(final HttpMetaData metadata) {
...
     }
 
+    private static boolean shouldAppendTrailers(final HttpProtocolVersion protocolVersion) {
+        // Always include trailers for h2 because trailers are allowed even if content-length is present, so we use
+        // trailers as a token to know the stream is done (even if they are empty).
+        return protocolVersion.major() != 1;
+    }
+
+    static boolean shouldAppendTrailers(final HttpProtocolVersion protocolVersion, final HttpHeaders headers) {
+        return shouldAppendTrailers(protocolVersion) || !headers.contains(CONTENT_LENGTH);
```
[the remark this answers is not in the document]
Scottmitch: I'll enhance the above conditional which focused on version check

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1630

PR #1851, opened 2021-09-27
Deprecate `AwaitUtils#takeUninterruptibly`

Krupskis:
Motivation:
The current usage of AwaitUtils catches
InterruptedException and silently discards it. jUnit5 uses
result, these tests may hang forever as the InterruptedException
is not propagated.

Modifications:
- Rewrite `takeUninterruptibly` to `take` in `AwaitUtils`

Result:
Deprecated `takeUninterruptibly`. `take` now throws unchecked
exception in the case when `queue.take()` throws `InterruptedException`.

The only problem here is that  I use `return null` to compile
successfully. `spotBugs` does not like that because 
I return null for a method that has `@NonNullByDefault`.

I don't understand where it is coming from since there is no
`@NonNullByDefault` for the method, could someone explain?

I tried putting `@Nullable`, but that did not fix the build, so
the workaround was to use `@SupressFBWarnings`, which is
not ideal, because it introduces new dependencies, is there a
smarter way to handle this?

### DIFF
--- a/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
+++ b/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
@@ -123,11 +123,34 @@ public final class AwaitUtils {
     }
 
     /**
-     * {@link BlockingQueue#take()} from the queue while suppressing {@link InterruptedException}s.
+     * {@link BlockingQueue#take()} from the queue or throws unchecked exception in the case
+     * of InterruptedException.
      * @param queue The queue to take from.
      * @param <T> The types of objects in the queue.
      * @return see {@link BlockingQueue#take()}.
      */
+    @edu.umd.cs.findbugs.annotations.SuppressFBWarnings(
+            value = "NP_NONNULL_RETURN_VIOLATION",
+            justification = "return null statement is never reached"
+    )
+    public static <T> T take(BlockingQueue<T> queue) {
+        try {
+            return queue.take();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throwException(e);
+            return null;
+        }
+    }
+
+    /**
+     * {@link BlockingQueue#take()} from the queue while suppressing {@link InterruptedException}s.
+     * @param queue The queue to take from.
+     * @param <T> The types of objects in the queue.
+     * @return see {@link BlockingQueue#take()}.
+     * @deprecated use {@link #take(BlockingQueue)} instead.
+     */
+    @Deprecated
     public static <T> T takeUninterruptibly(BlockingQueue<T> queue) {
         boolean interrupted = false;
         try {
--- a/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
+++ b/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
@@ -129,17 +129,16 @@ public final class AwaitUtils {
      * @param <T> The types of objects in the queue.
      * @return see {@link BlockingQueue#take()}.
      */
-    @edu.umd.cs.findbugs.annotations.SuppressFBWarnings(
-            value = "NP_NONNULL_RETURN_VIOLATION",
-            justification = "return null statement is never reached"
-    )
+    @SuppressWarnings("unchecked")
     public static <T> T take(BlockingQueue<T> queue) {
         try {
             return queue.take();
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             throwException(e);
-            return null;
+            // Returning Object cast to type T is not necessary as the method will either return or throw exception
+            // before even reaching this statement. However, this is necessary to compile successfully.
+            return (T) new Object();
         }
     }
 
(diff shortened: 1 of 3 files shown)


servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
```diff
@@ -122,12 +122,35 @@ public static boolean awaitUninterruptibly(CountDownLatch latch, long timeout, T
...
+            value = "NP_NONNULL_RETURN_VIOLATION",
+            justification = "return null statement is never reached"
+    )
+    public static <T> T take(BlockingQueue<T> queue) {
+        try {
+            return queue.take();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throwException(e);
+            return null;
```
idelpivnitskiy, line 142: This is how we workarounded this issue in `PlatformDependent` class:
https://github.com/apple/servicetalk/blob/464554ad984bc4897e7ba72e54a2ac0f12a9ecd9/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent.java#L163-L165

Consider following a similar approach and add a comment to clarify why this is necessary.

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
```diff
@@ -122,12 +122,35 @@ public static boolean awaitUninterruptibly(CountDownLatch latch, long timeout, T
...
     }
 
+    /**
+     * {@link BlockingQueue#take()} from the queue or throws unchecked exception in the case
+     * of InterruptedException.
+     * @param queue The queue to take from.
+     * @param <T> The types of objects in the queue.
+     * @return see {@link BlockingQueue#take()}.
+     */
+    @edu.umd.cs.findbugs.annotations.SuppressFBWarnings(
```
idelpivnitskiy, line 132: After the suggestion above, this won't be necessary. FYI for future:

Instead of annotations, we use `main-exclusions.xml` files in `gradle` subfolder to suppress SpotBugs warnings we consider acceptable. See example: https://github.com/apple/servicetalk/blob/main/servicetalk-grpc-api/gradle/spotbugs/main-exclusions.xml

idelpivnitskiy (commented):
Glad to see more PRs from you @Krupskis!

> I don't understand where it is coming from since there is no `@NonNullByDefault` for the method, could someone explain?

We use `@ElementsAreNonnullByDefault` annotation at `package-info.java` file to make all methods "non-null" by default:
https://github.com/apple/servicetalk/blob/f9119a7910031986750586cae118a572b4f11f1e/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/package-info.java#L16-L22

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
Krupskis: My other solution was to return Object cast to T, but never got to the point to also include `@SuppressWarnings("unchecked")`. This indeed makes it more straightforward and readable, thank you :)

Krupskis:
@idelpivnitskiy Thanks for your time to review and answer my questions! I fixed the PR. Weird stuff, but locally my build was failing with this, which for me seemed unreasonable as I think those tests don't even use `AwaitUtils#take`:
```
> Task :servicetalk-http-api:test

DefaultHttpExecutionStrategyTest > offloadSendSingle(Params) > io.servicetalk.http.api.DefaultHttpExecutionStrategyTest.offloadSendSingle(Params)[2] FAILED
    java.lang.AssertionError: Async errors occurred. See suppressed!
        at io.servicetalk.test.resources.TestUtils.assertNoAsyncErrors(TestUtils.java:57)
        at io.servicetalk.test.resources.TestUtils.assertNoAsyncErrors(TestUtils.java:42)
        at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest$ThreadAnalyzer.verifyNoErrors(DefaultHttpExecutionStrategyTest.java:379)
        at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest$ThreadAnalyzer.verifySend(DefaultHttpExecutionStrategyTest.java:363)
        at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest.offloadSendSingle(DefaultHttpExecutionStrategyTest.java:186)

DefaultHttpExecutionStrategyTest > offloadSendSingle(Params) > io.servicetalk.http.api.DefaultHttpExecutionStrategyTest.offloadSendSingle(Params)[6] FAILED
    java.lang.AssertionError: Async errors occurred. See suppressed!
        at io.servicetalk.test.resources.TestUtils.assertNoAsyncErrors(TestUtils.java:57)
        at io.servicetalk.test.resources.TestUtils.assertNoAsyncErrors(TestUtils.java:42)
        at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest$ThreadAnalyzer.verifyNoErrors(DefaultHttpExecutionStrategyTest.java:379)
        at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest$ThreadAnalyzer.verifySend(DefaultHttpExecutionStrategyTest.java:363)
        at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest.offloadSendSingle(DefaultHttpExecutionStrategyTest.java:186)

784 tests completed, 2 failed, 90 skipped
```

idelpivnitskiy (approved):

idelpivnitskiy:
Some tests a flaky due to concurrency or the way a test is written. We are tracking those: https://github.com/apple/servicetalk/issues?q=is%3Aopen+is%3Aissue+label%3A%22flaky+tests%22
No worries, thank you for contribution!

idelpivnitskiy:
Related to #1684

PR #1852, opened 2021-09-28
Unifying `ServiceDiscoverer` and `LoadBalancerFactory` APIs

chemicL:
Motivation:

`ServiceDiscoverer#discover` method returns a `Publisher<Collection<E>>`
while `LoadBalancerFactory#newLoadBalancer` takes `Publisher<E>` as an
argument. Those interfaces should be aligned.

Modifications:

Recently introduced `LoadBalancerFactory#newLoadBalancer` default method
now takes a `Publisher<Collection<E>>` as an argument. Implementations
in `DefaultHttpLoadBalancerFactory` and `RoundRobinLoadBalancerFactory`
implement that properly and `RoundRobinLoadBalancer` iterates over the
events.

Result:

A more aligned API that can allow acting upon a fixed collection
returned from the `ServiceDiscoverer` allowing more sophisticated usages
of the fact that these items come as one update.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
@@ -33,7 +33,8 @@ public interface LoadBalancerFactory<ResolvedAddress, C extends LoadBalancedConn
     /**
      * Create a new {@link LoadBalancer}.
      * @param eventPublisher A stream of {@link ServiceDiscovererEvent}s which the {@link LoadBalancer} can use to
-     * connect to physical hosts. Typically generated from a {@link ServiceDiscoverer}.
+     * connect to physical hosts. Typically generated from a
+     * {@link ServiceDiscoverer#discover(Object) ServiceDiscoverer}.
      * @param connectionFactory {@link ConnectionFactory} that the returned {@link LoadBalancer} will use to generate
      * new connections. Returned {@link LoadBalancer} will own the responsibility for this {@link ConnectionFactory}
      * and hence will call {@link ConnectionFactory#closeAsync()} when {@link LoadBalancer#closeAsync()} is called.
@@ -59,7 +60,7 @@ public interface LoadBalancerFactory<ResolvedAddress, C extends LoadBalancedConn
      * a merged collection.
      * @param eventPublisher A stream of {@link Collection}&lt;{@link ServiceDiscovererEvent}&gt;
      * which the {@link LoadBalancer} can use to connect to physical hosts. Typically generated
-     * from a {@link ServiceDiscoverer}.
+     * from {@link ServiceDiscoverer#discover(Object) ServiceDiscoverer}.
      * @param connectionFactory {@link ConnectionFactory} that the returned {@link LoadBalancer} will use to generate
      * new connections. Returned {@link LoadBalancer} will own the responsibility for this {@link ConnectionFactory}
      * and hence will call {@link ConnectionFactory#closeAsync()} when {@link LoadBalancer#closeAsync()} is called.
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
@@ -17,6 +17,10 @@ package io.servicetalk.client.api;
 
 import io.servicetalk.concurrent.api.Publisher;
 
+import java.util.Collection;
+
+import static java.util.function.Function.identity;
+
 /**
  * A factory for creating {@link LoadBalancer} instances.
  *
@@ -53,8 +57,9 @@ public interface LoadBalancerFactory<ResolvedAddress, C extends LoadBalancedConn
      * will perform load balancing. Bear in mind, load balancing is performed over the a collection of hosts provided
      * via the {@code eventPublisher} which may not correspond directly to a single unresolved address, but potentially
      * a merged collection.
-     * @param eventPublisher A stream of {@link ServiceDiscovererEvent}s which the {@link LoadBalancer} can use to
-     * connect to physical hosts. Typically generated from a {@link ServiceDiscoverer}.
+     * @param eventPublisher A stream of {@link Collection}&lt;{@link ServiceDiscovererEvent}&gt;
+     * which the {@link LoadBalancer} can use to connect to physical hosts. Typically generated
+     * from a {@link ServiceDiscoverer}.
      * @param connectionFactory {@link ConnectionFactory} that the returned {@link LoadBalancer} will use to generate
      * new connections. Returned {@link LoadBalancer} will own the responsibility for this {@link ConnectionFactory}
      * and hence will call {@link ConnectionFactory#closeAsync()} when {@link LoadBalancer#closeAsync()} is called.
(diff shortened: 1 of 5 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -278,8 +277,9 @@ public StreamingHttpClient buildStreaming() {
         // Track resources that potentially need to be closed when an exception is thrown during buildStreaming
         final CompositeCloseable closeOnException = newCompositeCloseable();
         try {
-            final Publisher<ServiceDiscovererEvent<R>> sdEvents =
-                    ctx.serviceDiscoverer(executionContext).discover(ctx.address()).flatMapConcatIterable(identity());
+            // final Publisher<ServiceDiscovererEvent<R>> sdEvents =
```
tkountis, line 280: leftover?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -183,17 +186,19 @@ public RoundRobinLoadBalancer(final Publisher<? extends ServiceDiscovererEvent<R
      * continues being eligible for connecting on the request path).
      * @see io.servicetalk.loadbalancer.RoundRobinLoadBalancerFactory
      */
-    RoundRobinLoadBalancer(String targetResource,
-                           final Publisher<? extends ServiceDiscovererEvent<ResolvedAddress>> eventPublisher,
-                           final ConnectionFactory<ResolvedAddress, ? extends C> connectionFactory,
-                           final boolean eagerConnectionShutdown,
-                           @Nullable final HealthCheckConfig healthCheckConfig) {
+    RoundRobinLoadBalancer(
+            String targetResource,
```
tkountis, line 190: final

tkountis (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
chemicL: yep, absolutely

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
```diff
@@ -53,8 +57,9 @@
      * will perform load balancing. Bear in mind, load balancing is performed over the a collection of hosts provided
      * via the {@code eventPublisher} which may not correspond directly to a single unresolved address, but potentially
      * a merged collection.
-     * @param eventPublisher A stream of {@link ServiceDiscovererEvent}s which the {@link LoadBalancer} can use to
-     * connect to physical hosts. Typically generated from a {@link ServiceDiscoverer}.
+     * @param eventPublisher A stream of {@link Collection}&lt;{@link ServiceDiscovererEvent}&gt;
+     * which the {@link LoadBalancer} can use to connect to physical hosts. Typically generated
```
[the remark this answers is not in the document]
chemicL: Note there's more mentions of "physical" around this file. I'd open a separate PR for that if you don't mind :)

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
```diff
@@ -63,8 +68,8 @@
      */
     default <T extends C> LoadBalancer<T> newLoadBalancer(
             String targetResource,
-            Publisher<? extends ServiceDiscovererEvent<ResolvedAddress>> eventPublisher,
+            Publisher<? extends Collection<? extends ServiceDiscovererEvent<ResolvedAddress>>> eventPublisher,
```
idelpivnitskiy, line 71: Do we need `? extends` for `Collection`? `discovery` method returns `Collection`:
https://github.com/apple/servicetalk/blob/f9119a7910031986750586cae118a572b4f11f1e/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java#L52

I can see the value of giving more flexibility to express a more specific `Collection` type, but only if `ServiceDiscoverer` API also aligned. Not sure if we have to go that path though. WDYT?

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
```diff
@@ -53,8 +57,9 @@
      * will perform load balancing. Bear in mind, load balancing is performed over the a collection of hosts provided
      * via the {@code eventPublisher} which may not correspond directly to a single unresolved address, but potentially
      * a merged collection.
-     * @param eventPublisher A stream of {@link ServiceDiscovererEvent}s which the {@link LoadBalancer} can use to
-     * connect to physical hosts. Typically generated from a {@link ServiceDiscoverer}.
+     * @param eventPublisher A stream of {@link Collection}&lt;{@link ServiceDiscovererEvent}&gt;
+     * which the {@link LoadBalancer} can use to connect to physical hosts. Typically generated
+     * from a {@link ServiceDiscoverer}.
```
idelpivnitskiy, line 62: Consider `{@link ServiceDiscoverer#discover(Object) ServiceDiscoverer}`

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
chemicL: Consider the following example:

```java
    private static <T> Publisher<? extends T> extract(Publisher<Collection<? extends T>> pub) {
        return pub.flatMapConcatIterable(Function.identity());
    }
    
    final Publisher<Collection<String>> from = Publisher.from(List.of("hello"), List.of("world"));
    extract(from).toFuture().get(); // failure: reason: Incompatible equality constraint: ? extends T and String
```

vs

```java
    private static <T> Publisher<? extends T> extract(Publisher<? extends Collection<? extends T>> pub) {
        return pub.flatMapConcatIterable(Function.identity());
    }
    
    final Publisher<Collection<String>> from = Publisher.from(List.of("hello"), List.of("world"));
    extract(from).toFuture().get(); // compiles fine
```

Full output of the compiler:

```
error: method extract in class (...) cannot be applied to given types;
        extract(from).toFuture().get();
        ^
  required: Publisher<Collection<? extends T>>
  found: Publisher<Collection<String>>
  reason: cannot infer type-variable(s) T
    (argument mismatch; Publisher<Collection<String>> cannot be converted to Publisher<Collection<? extends T>>)
  where T is a type-variable:
    T extends Object declared in method <T>extract(Publisher<Collection<? extends T>>)
```

The reason is that `Publisher<Collection<String>` is not a child of `Publisher<Collection<? extends String>`, but is a child of `Publisher<? extends Collection<? extends String>>` in the same way that `Collection<String>` is not a child of `Collection<Object>` but is a child of  `Collection<? extends Object>`:

```java
Collection<Object> l = List.<String>of("hello"); // doesn't compile
Collection<? extends Object> l2 = List.<String>of("world"); // compiles fine
```

chemicL:
Cherry-picked to 0.41: c15d825c9f3272908394e1efed54b6b3449fbb8f

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
idelpivnitskiy: Thanks for explaining, sgtm!

PR #1854, opened 2021-09-29
Log the cause of failed attempt to open a new connection

idelpivnitskiy:
Motivation:

Get visibility for reasons that trigger health checking. Useful to
debug unstable hosts.

Modifications:

- Log each failed attempt to open a new connection with the failure
cause;
- Make all debug logs inside `markUnhealthy` consistent;

Result:

Users can debug the cause for failed attempts to open a new connection.

### DIFF
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -631,25 +631,25 @@ public final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalance
         }
 
         void markUnhealthy(final Throwable cause, final ConnectionFactory<Addr, ? extends C> connectionFactory) {
-            LOGGER.debug("Load balancer for {}: failed to open a new connection to the host on address {}.",
-                    targetResource, address, cause);
             assert healthCheckConfig != null;
             for (;;) {
                 ConnState previous = connStateUpdater.get(this);
 
                 if (!ActiveState.class.equals(previous.state.getClass()) || previous.connections.length > 0) {
+                    LOGGER.debug("Load balancer for {}: failed to open a new connection to the host on address {}. {}",
+                            targetResource, address, previous, cause);
                     break;
                 }
 
                 ActiveState previousState = (ActiveState) previous.state;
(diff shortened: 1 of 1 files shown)


servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -630,7 +630,9 @@ void markHealthy() {
             cancelIfHealthCheck(oldState);
         }
 
-        void markUnhealthy(ConnectionFactory<Addr, ? extends C> connectionFactory) {
+        void markUnhealthy(final Throwable cause, final ConnectionFactory<Addr, ? extends C> connectionFactory) {
+            LOGGER.debug("Load balancer for {}: failed to open a new connection to the host on address {}.",
+                    targetResource, address, cause);
```
idelpivnitskiy, line 635: I planned to include the cause in pre-existing log message at line 649, but then decided to log it for all cases (even if the host became non-active at the time we tried to open a new connection) for better visibility.

chemicL: It feels the log is duplicated now. Can you please adjust the other log to be more about the number of times failure happened and skip the `failed to open a new connection to the host on address {}`?

E.g.
`"Load balancer for {}: failure #{} ({} consecutive failures will trigger health check)."`

chemicL (changes requested):
Please revisit the logs produced. I have a feeling we can adjust the log about going into health check mode or not doing so yet so that we don't confuse the reader with regards to the amount of times a connection failure happened. Right now a single failure would produce two log entries with the line `failed to open a new connection to the host on address {}` which I believe is confusing.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
idelpivnitskiy: Agreed, but we always have to include the host address in the log. Otherwise, we won't be able to track errors per IP. I made a separate log statement for each case inside `for` loop with consistent prefix to allow easier search of the logs related to the same LB and address.

chemicL (approved):
Great improvement, thank you :-)

PR #1856, opened 2021-09-29
Deprecate `AwaitUtils#pollUninterruptibly`

Krupskis:
Motivation:
The current usage of AwaitUtils catches
InterruptedException and silently discards it. jUnit5 uses
result, these tests may hang forever as the InterruptedException
is not propagated.

Modifications:

- introduce `AwaitUtils#poll`

Result:
`AwaitUtils#poll` now throws unchecked exception
in the case that the thread is interrupted. Meaning that
JUnit5 will be able to stop the tests by interrupting
when it timeouts.

### DIFF
--- a/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
+++ b/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
@@ -168,13 +168,35 @@ public final class AwaitUtils {
     }
 
     /**
-     * {@link BlockingQueue#poll(long, TimeUnit)} from the queue while suppressing {@link InterruptedException}s.
+     * {@link BlockingQueue#poll(long, TimeUnit)} from the queue, throws unchecked Exception if
+     * thread is interrupted while waiting.
      * @param queue the queue to poll from.
      * @param timeout the timeout duration to poll for.
      * @param unit the units applied to {@code timeout}.
      * @param <T> The types of objects in the queue.
      * @return see {@link BlockingQueue#poll(long, TimeUnit)}.
      */
+    @Nullable
+    public static <T> T poll(BlockingQueue<T> queue, long timeout, TimeUnit unit) {
+        try {
+            return queue.poll(timeout, NANOSECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throwException(e);
+            return null;
+        }
+    }
+
+    /**
+     * {@link BlockingQueue#poll(long, TimeUnit)} from the queue while suppressing {@link InterruptedException}s.
+     * @param queue the queue to poll from.
+     * @param timeout the timeout duration to poll for.
+     * @param unit the units applied to {@code timeout}.
+     * @param <T> The types of objects in the queue.
+     * @return see {@link BlockingQueue#poll(long, TimeUnit)}.
+     * @deprecated use {@link #poll(BlockingQueue, long, TimeUnit)} instead.
+     */
+    @Deprecated
     @Nullable
     public static <T> T pollUninterruptibly(BlockingQueue<T> queue, long timeout, TimeUnit unit) {
         final long startTime = System.nanoTime();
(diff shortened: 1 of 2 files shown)


servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
```diff
@@ -167,14 +167,36 @@ public static boolean awaitUninterruptibly(CountDownLatch latch, long timeout, T
...
+     * {@link BlockingQueue#poll(long, TimeUnit)} from the queue, throws unchecked Exception if
+     * thread is interrupted while waiting.
+     * @param queue the queue to poll from.
+     * @param timeout the timeout duration to poll for.
+     * @param unit the units applied to {@code timeout}.
+     * @param <T> The types of objects in the queue.
+     * @return see {@link BlockingQueue#poll(long, TimeUnit)}.
+     */
+    @Nullable
+    public static <T> T poll(BlockingQueue<T> queue, long timeout, TimeUnit unit) {
```
bondolo, line 180: I am not a fan of the sneaky-throws but this seems like a reasonable compromise if we really don't want the `throws InterruptedException` on the signature.

bondolo (approved):

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
```diff
@@ -167,14 +167,36 @@ public static boolean awaitUninterruptibly(CountDownLatch latch, long timeout, T
...
+        try {
+            return queue.poll(timeout, NANOSECONDS);
```
chemicL, line 182: The `unit` is unused and `NANOSECONDS` is enforced regardless of the intention of the caller. I believe that's unintended.

chemicL (changes requested):
Thank you for the PR! Noticed one detail, please see my comment in the introduced method.

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
Krupskis: @chemicL Good catch, fixed it, thanks. `pollUninterruptibly` uses `NANOSECONDS` by default, is that because the timeout is manually controlled by the `waitTime`? Here is the code of the `pollUninterruptibly`.
```
public static <T> T pollUninterruptibly(BlockingQueue<T> queue, long timeout, TimeUnit unit) {
        final long startTime = System.nanoTime();
        final long timeoutNanos = NANOSECONDS.convert(timeout, unit);
        long waitTime = timeout;
        boolean interrupted = false;
        try {
            do {
                try {
                    return queue.poll(waitTime, NANOSECONDS);
                    interrupted = true;
                }
                waitTime = timeoutNanos - (System.nanoTime() - startTime);
                if (waitTime <= 0) {
                    return null;
                }
            } while (true);
        } finally {
            if (interrupted) {
            }
        }
    }

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
idelpivnitskiy: @bondolo, agreed. We discussed it here: https://github.com/apple/servicetalk/pull/1834#discussion_r712622932

idelpivnitskiy (approved):
Thank you @Krupskis!

idelpivnitskiy:
Related to #1684

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/AwaitUtils.java
chemicL: Apologies for missing the notification. I think the only reason there is the `System.nanoTime()` call, which provides nanos, so the calculations are based around that unit. Any JDK method accepting a `long` with the specifier for the `TimeUnit` that the value represents will deal fine with any unit. Please note that the `BlockingQueue#poll(long, TimeUnit)` method internally transforms the values to nanos, so passing nanos here is the most straightforward and doesn't require any additional computation. Hope that helps.

PR #1860, opened 2021-10-01
Handle `InterruptedException` in tests consistently

Krupskis:
Motivation:
There are some test files where thread's interrupted
flag is not reset after InterruptedException is caught,
as well as where InterruptedException is being wrapped
into RuntimeException, which JUnit5 may not recognize,
when stopping the tests after a timeout.

This does not concern any non-test files.

Modifications:

- Use `Thread.currentThread().interrupt();` where it was forgotten
- Use unchecked exception with `throwException` instead of `RuntimeException`
- Where possibly simply rethrow the caught exception.

Result:
That did not work out as nicely as expected. Some tests are failing
locally, which I am still trying to figure out how to fix.

some questions:
- how to handle this in HttpOffloadingTest.java:345?
I don't really see how the errors in the errors queue are handled, 
so had no idea if there is anything I could fix here.

- Is it okay how I handle InterruptedException in `TestPublisherSubscriber#takeOnNext(int n)`?
as I understand if the exception was thrown we should propagate it at
some point, but wouldn't that break the current method?

That's about it, if I forgot something I will add it as a comment
to the PR.

### DIFF
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/ScanWithPublisherTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/ScanWithPublisherTest.java
@@ -390,13 +390,14 @@ class ScanWithPublisherTest {
                         }
 
                         @Override
-                        public boolean mapTerminal() {
+                        public boolean mapTerminal() throws InterruptedException {
                             if (interleaveCancellation) {
                                 checkpoint.countDown();
                                 try {
                                     resume.await();
                                 } catch (InterruptedException e) {
-                                    e.printStackTrace();
+                                    Thread.currentThread().interrupt();
+                                    throw e;
                                 }
                             }
                             return true;
--- a/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/TestPublisherSubscriber.java
+++ b/servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/TestPublisherSubscriber.java
@@ -37,6 +37,7 @@ import static io.servicetalk.concurrent.internal.TerminalNotification.error;
 import static io.servicetalk.concurrent.test.internal.AwaitUtils.await;
 import static io.servicetalk.concurrent.test.internal.AwaitUtils.poll;
 import static io.servicetalk.concurrent.test.internal.AwaitUtils.take;
+import static io.servicetalk.utils.internal.PlatformDependent.throwException;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toList;
 
@@ -198,6 +199,7 @@ public final class TestPublisherSubscriber<T> implements Subscriber<T> {
         } finally {
             if (interrupted) {
                 Thread.currentThread().interrupt();
+                throwException(new InterruptedException());
             }
         }
         return list;
--- a/servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultHttpExecutionStrategyTest.java
+++ b/servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultHttpExecutionStrategyTest.java
@@ -46,10 +46,10 @@ import static io.servicetalk.http.api.StreamingHttpRequests.newRequest;
 import static io.servicetalk.http.api.StreamingHttpResponses.newResponse;
 import static io.servicetalk.test.resources.TestUtils.assertNoAsyncErrors;
 import static io.servicetalk.transport.netty.internal.NettyIoExecutors.createIoExecutor;
+import static io.servicetalk.utils.internal.PlatformDependent.throwException;
 import static java.lang.Thread.currentThread;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
-import static org.junit.jupiter.api.Assertions.fail;
 
 class DefaultHttpExecutionStrategyTest {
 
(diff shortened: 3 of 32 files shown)


Krupskis:
One thing: the only place where I was unable to change `throw new RuntimeException(e)` to `throwException(e)` is the `awaitFor(CountDownLatch latch)` in `DefaultExecutorTest.java`. It fails with:
```
> Task :servicetalk-concurrent-api:test

DefaultExecutorTest > cancelExecute(ExecutorParam) > io.servicetalk.concurrent.api.DefaultExecutorTest.cancelExecute(ExecutorParam)[1] FAILED
    java.lang.AssertionError: 
    Expected: an instance of java.lang.InterruptedException
         but: null
        at io.servicetalk.concurrent.api.DefaultExecutorTest.cancelExecute(DefaultExecutorTest.java:226)

DefaultExecutorTest > cancelExecute(ExecutorParam) > io.servicetalk.concurrent.api.DefaultExecutorTest.cancelExecute(ExecutorParam)[2] FAILED
    java.lang.AssertionError: 
    Expected: an instance of java.lang.InterruptedException
         but: null
        at io.servicetalk.concurrent.api.DefaultExecutorTest.cancelExecute(DefaultExecutorTest.java:226)

DefaultExecutorTest > cancelExecute(ExecutorParam) > io.servicetalk.concurrent.api.DefaultExecutorTest.cancelExecute(ExecutorParam)[4] FAILED
    java.lang.AssertionError: 
    Expected: an instance of java.lang.InterruptedException
         but: null
        at io.servicetalk.concurrent.api.DefaultExecutorTest.cancelExecute(DefaultExecutorTest.java:226)

DefaultExecutorTest > cancelExecute(ExecutorParam) > io.servicetalk.concurrent.api.DefaultExecutorTest.cancelExecute(ExecutorParam)[5] FAILED
    java.lang.AssertionError: 
    Expected: an instance of java.lang.InterruptedException
         but: null
        at io.servicetalk.concurrent.api.DefaultExecutorTest.cancelExecute(DefaultExecutorTest.java:226)
```

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/ScanWithPublisherTest.java
```diff
@@ -390,13 +390,14 @@ public Integer mapOnComplete() {
...
-                        public boolean mapTerminal() {
+                        public boolean mapTerminal() throws InterruptedException {
                             if (interleaveCancellation) {
                                 checkpoint.countDown();
                                 try {
                                     resume.await();
                                 } catch (InterruptedException e) {
-                                    e.printStackTrace();
+                                    Thread.currentThread().interrupt();
+                                    throw e;
```
idelpivnitskiy, line 400: Can we use `PlatformDependent.throwException(e)` instead of adding `throws InterruptedException` on API?

servicetalk-concurrent-test-internal/src/main/java/io/servicetalk/concurrent/test/internal/TestPublisherSubscriber.java
```diff
@@ -198,6 +199,7 @@ public T takeOnNext() {
         } finally {
             if (interrupted) {
                 Thread.currentThread().interrupt();
+                throwException(new InterruptedException());
```
idelpivnitskiy, line 202: This method suppresses `InterruptedException` in line 192, but we try to avoid doing that. Consider changing this method to rethrow immediately after catching the first `InterruptedException`.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultHttpExecutionStrategyTest.java
```diff
@@ -161,8 +161,11 @@ void wrapServiceThatWasNotOffloaded(final Params params) throws Exception {
                 analyzer.instrumentedResponseForServer(svc.handle(ctx, req, ctx.streamingResponseFactory()))
                         .flatMapPublisher(StreamingHttpResponse::payloadBody)
                     .toFuture().get();
-            } catch (InterruptedException | ExecutionException e) {
-                fail("Unexepected exception", e);
+            } catch (InterruptedException e) {
```
idelpivnitskiy, line 164: Consider changing `Runnable` to `Callable` instead of catching and rethrowing exceptions.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionCloseHeaderHandlingTest.java
```diff
@@ -189,8 +189,10 @@ public Completable accept(final ConnectionContext context) {
                                     try {
                                         responseReceived.await();
                                         done = true;
-                                    } catch (InterruptedException interruptedException) {
+                                    } catch (InterruptedException e) {
                                         // ignored
```
idelpivnitskiy, line 193: This is not ignored anymore, consider removing this comment.

servicetalk-opentracing-zipkin-publisher/src/test/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporterTest.java
```diff
@@ -170,7 +171,10 @@ private void verifySpanSendFailuresAreRecovered(final Codec codec, final HttpRep
         responseGenerator = (httpServiceContext, factory) -> {
             try {
                 httpServiceContext.closeAsync().toFuture().get();
-            } catch (InterruptedException | ExecutionException e) {
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throwException(e);
+            } catch (ExecutionException e) {
                 throw new RuntimeException(e);
```
idelpivnitskiy, line 178: In all other places you rethrow `ExecutionException` using `throwException`. Can we be consistent here?

idelpivnitskiy (changes requested):
Thank you for the PR!

> how to handle this in `HttpOffloadingTest.java:345`?

You can follow the same pattern as everywhere else after `errors.add(e)`:
```java
            try {
                latch.await();
                errors.add(e);
                throwException(e);
            }
```

Krupskis:
@idelpivnitskiy thanks for the review! I considered the stuff you mentioned and also spotted one more case I missed, so added that as well.

Do you have any input on how I could fix `DefaultExecutorTest:506`?  I assume the reason why tests fail has something to do with exception type, but I was unable to find a solution. You can see the error messages and more details in the first comment of this PR.

PR #1861, opened 2021-10-04
Untangling `GrpcServerBuilder` from `HttpServerBuilder`

chemicL:
Motivation:

The methods in `GrpcServerBuilder` are often only delegating to the
underlying HTTP transport builder. It will be beneficial if the API can
be simplified and have only gRPC specific settings. The first step is to
deprecate the methods which can be called directly on the underlying
`HttpServerBuilder` and to introduce a method for configuring that
builder.

Modifications:

- Introduce a `GrpcServerBuilder.HttpInitializer` interface and a
  corresponding `GrpcServerBuilder#initialize` method which accepts the
  initializer for the underlying `HttpServerBuilder` instance,
- Deprecate methods which have only served the purpose of delegating to
  the underlying HTTP transport builder,
- Refactor tests to use the new approach to showcase it works.

Result:

The API of `GrpcServerBuilder` is prepared for removal of the deprecated
methods in 0.42 release and the API is more specific, while maintaining
the same level of configurability.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
@@ -69,6 +69,12 @@ public abstract class GrpcServerBuilder {
          */
         void initialize(HttpServerBuilder builder);
 
+        /**
+         * Appends the passed {@link HttpInitializer} to this {@link HttpInitializer} such that this instance is
+         * applied first and then the argument's {@link HttpInitializer}.
+         * @param toAppend {@link HttpInitializer} to append.
+         * @return A composite {@link HttpInitializer} after the append operation.
+         */
         default HttpInitializer append(HttpInitializer toAppend) {
             return builder -> {
                 initialize(builder);


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
```diff
@@ -55,8 +56,15 @@
...
 public abstract class GrpcServerBuilder {
 
+    @FunctionalInterface
+    public interface HttpInitializer {
+        void initialize(HttpServerBuilder builder);
+    }
+
     private boolean appendedCatchAllFilter;
 
+    public abstract GrpcServerBuilder initializer(HttpInitializer initializer);
```
tkountis, line 66: Does this affect the order of applied configurations? In other words, when we are in this transient state with the deprecations, if users call `appendHttpServiceFilter` directly (deprecated methods) and they also start using the `initializer` where they also might append filters, is the order of the filters consistent to what it would have been if that new API wasn't there? If thats the case, can we prevent bad configuration, e.g. allowing exclusively one way or the other.

tkountis (commented):
The deprecations will be removed for 0.42 correct?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
chemicL: Great observation! To answer – the order would not be preserved as the `initializer` is applied at build time, while the deprecated methods have an immediate effect on the underlying builder. We could add documentation, but that would not prevent users from possible issues. I will try to implement a mutually exclusive mechanism, bear with me.

chemicL:
> The deprecations will be removed for 0.42 correct?

Yes, that is the intention to remove the deprecated methods.

tkountis (approved):
LGTM 🚢

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
```diff
@@ -361,7 +442,10 @@ public final ServerContext listenAndAwait(GrpcBindableService<?, ?, ?>... servic
     protected abstract void doAppendHttpServiceFilter(Predicate<StreamingHttpRequest> predicate,
                                                       StreamingHttpServiceFilterFactory factory);
 
-    private void appendCatchAllFilterIfRequired() {
+    protected void appendCatchAllFilterIfRequired() {
```
idelpivnitskiy, line 445: Can not comment there, but can we also deprecate `executor/ioExecutor/bufferAllocator` top level methods? Server builder doesn't have a problem like multi-address client. We can take execution context from HttpServer.

Also not sure if we really need `GrpcExecutionStrategy` for now. Maybe we can deprecate that one as well and add later when we have a need? @bondolo wdyt?

idelpivnitskiy (changes requested):

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
idelpivnitskiy: I'm fine to do that in a follow-up

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
idelpivnitskiy: Ok, I found where this comes from :) 
We should not be extra caution in this scenario. Users either migrate all `appendHttpServiceFilter` methods or nothing. If they use a mixed approach - this is their problem. We can clarify it in javadoc, but no need to add extra guards in the code. Neither multi-address nor partitioned client builders have similar guards.

idelpivnitskiy:
@bondolo this is not a breaking change, we expect to cherry-pick thin into 0.41

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
```diff
@@ -197,6 +206,11 @@ protected void doAppendHttpServiceFilter(final Predicate<StreamingHttpRequest> p
 
     private HttpServerBuilder preBuild() {
         if (!invokedBuild) {
+            // TODO(dj): Consider logging a warning in every method that changes the state of the builder once it has
+            // been used to build a client.
```
idelpivnitskiy, line 210: Don't understand this comment, can you clarify what you mean and why a client is involved here?

idelpivnitskiy (commented):
LGTM after the last comments addressed.
Nice change! Thanks @tkountis for a proposal ❤️

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcServerBuilder.java
```diff
@@ -134,7 +192,12 @@ public abstract GrpcServerBuilder enableWireLogging(String loggerName, LogLevel
      *
      * @param transportObserver A {@link TransportObserver} that provides visibility into transport events.
      * @return {@code this}.
+     * @deprecated Call {@link #initializeHttp(HttpInitializer)} and use
+     * {@link HttpServerBuilder#transportObserver(TransportObserver)}
+     * on the {@code builder} instance by implementing {@link HttpInitializer#initialize(HttpServerBuilder)}
+     * functional interface.
      */
+    @Deprecated
     public abstract GrpcServerBuilder transportObserver(TransportObserver transportObserver);
```
idelpivnitskiy, line 201: Can not comment there, will leave here:
`lifecycleObserver` is the only method that also exists at `HttpServerBuilder`. While you are here, please clarify in javadoc that if users use `lifecycleObserver(GrpcLifecycleObserver)`, using a similar method `HttpServerBuilder#lifecycleObserver(HttpLifecycleObserver)` will override the value.

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
```diff
@@ -197,6 +206,11 @@ protected void doAppendHttpServiceFilter(final Predicate<StreamingHttpRequest> p
...
+            if (initializer != null) {
```
idelpivnitskiy, line 211: It will be unexpected to users if they reuse the builder and change `HttpInitializer` between different builds to apply different socket options or filters. We should preserve the typical builder pattern.

Consider changing the ctor argument from `HttpServerBuilder` to `Supplier<HttpServerBuilder>`. Then you will be able to create a fresh new HTTP builder every time users try to build a gRPC server.
For deprecated methods, consider having a separate "deprecatedHttpInitializer" variable and compose a new `HttpInitializer` every time users invoke deprecated methods by appending a new configuration to the end. At build time, invoke `deprecatedHttpInitializer` first, then `httpInitializer`.

idelpivnitskiy (changes requested):
Discovered when I reviewed the client-side:

PR #1864, opened 2021-10-04
`japicmp.sh`: skip jdkflow on JDK8 and adjust error msg format

idelpivnitskiy:
Motivation:

`servicetalk-concurrent-jdkflow` requires JDK9+. Validation of the
`local` build fails when JDK8 is used. Also, old artifact name format
is a bit confusing, better to have it as a jar name.

Modifications:

- If JDK8 is detected for the `local` build, skip verifying
`servicetalk-concurrent-jdkflow` module;
- Change artifact name for error messages to match jar file;

Result:

`japicmp.sh` doesn't fail on JDK8.

### DIFF
--- a/scripts/japicmp.sh
+++ b/scripts/japicmp.sh
@@ -65,6 +65,7 @@ XSLTDOC
 
 MVN_REPO="$(mvn help:evaluate -Dexpression=settings.localRepository -q -DforceStdout)"
 
+JAVA_VERSION=$(./gradlew --no-daemon -version | grep ^JVM: | awk -F\. '{gsub(/^JVM:[ \t]*/,"",$1); print $1"."$2}')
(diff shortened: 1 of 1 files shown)


scripts/japicmp.sh
```diff
@@ -65,6 +65,7 @@ XSLTDOC
 
 MVN_REPO="$(mvn help:evaluate -Dexpression=settings.localRepository -q -DforceStdout)"
 
+JAVA_VERSION=$(./gradlew --no-daemon -version | grep ^JVM: | awk -F\. '{gsub(/^JVM:[ \t]*/,"",$1); print $1"."$2}')
```
idelpivnitskiy, line 68: This is taken from `publish-docs.sh` as is

bondolo (approved):

PR #1867, opened 2021-10-06
Untangling `GrpcClientBuilder` from `SingleAddressHttpClientBuilder`

chemicL:
Motivation:

The methods in `GrpcClientBuilder` are often only delegating to the
underlying HTTP transport builder. It will be beneficial if the API can
be simplified and have only gRPC specific settings. The first step is to
deprecate the methods which can be called directly on the underlying
`SingleAddressHttpClientBuilder` and to introduce a method for configuring that
builder.

Modifications:

- Introduce a `GrpcClientBuilder.HttpInitializer` interface and a
  corresponding `GrpcClientBuilder#initializeHttp` method which accepts the
  initializer for the underlying `SingleAddressHttpClientBuilder` instance,
- Deprecate methods which have only served the purpose of delegating to
  the underlying HTTP transport builder,
- Refactor tests to use the new approach to showcase it works.

Result:

The API of `GrpcClientBuilder` is prepared for removal of the deprecated
methods in 0.42 release and the API is more specific, while maintaining
the same level of configurability.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
@@ -68,6 +68,19 @@ public abstract class GrpcClientBuilder<U, R> {
          * @param builder The builder to customize the HTTP layer.
          */
         void initialize(SingleAddressHttpClientBuilder<U, R> builder);
+
+        /**
+         * Appends the passed {@link HttpInitializer} to this {@link HttpInitializer} such that this instance is
+         * applied first and then the argument's {@link HttpInitializer}.
+         * @param toAppend {@link HttpInitializer} to append.
+         * @return A composite {@link HttpInitializer} after the append operation.
+         */
+        default HttpInitializer<U, R> append(HttpInitializer<U, R> toAppend) {
+            return builder -> {
+                initialize(builder);
+                toAppend.initialize(builder);
+            };
+        }
     }
 
     /**
@@ -95,7 +108,9 @@ public abstract class GrpcClientBuilder<U, R> {
      * by implementing {@link HttpInitializer#initialize(SingleAddressHttpClientBuilder)} functional interface.
      */
     @Deprecated
-    public abstract GrpcClientBuilder<U, R> executor(Executor executor);
+    public GrpcClientBuilder<U, R> executor(Executor executor) {
+        throw new UnsupportedOperationException("Method executor is not supported by " + getClass().getName());
+    }
 
     /**
      * Sets the {@link IoExecutor} for all clients created from this builder.
@@ -107,7 +122,9 @@ public abstract class GrpcClientBuilder<U, R> {
      * by implementing {@link HttpInitializer#initialize(SingleAddressHttpClientBuilder)} functional interface.
      */
     @Deprecated
-    public abstract GrpcClientBuilder<U, R> ioExecutor(IoExecutor ioExecutor);
+    public GrpcClientBuilder<U, R> ioExecutor(IoExecutor ioExecutor) {
+        throw new UnsupportedOperationException("Method ioExecutor is not supported by " + getClass().getName());
+    }
 
     /**
      * Sets the {@link BufferAllocator} for all clients created from this builder.
@@ -119,7 +136,9 @@ public abstract class GrpcClientBuilder<U, R> {
      * by implementing {@link HttpInitializer#initialize(SingleAddressHttpClientBuilder)} functional interface.
      */
     @Deprecated
-    public abstract GrpcClientBuilder<U, R> bufferAllocator(BufferAllocator allocator);
+    public GrpcClientBuilder<U, R> bufferAllocator(BufferAllocator allocator) {
+        throw new UnsupportedOperationException("Method bufferAllocator is not supported by " + getClass().getName());
+    }
 
     /**
      * Sets the {@link GrpcExecutionStrategy} for all clients created from this builder.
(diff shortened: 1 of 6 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
```diff
@@ -61,30 +56,69 @@
...
      * Sets the {@link Executor} for all clients created from this builder.
      *
      * @param executor {@link Executor} to use.
      * @return {@code this}.
+     * @deprecated Call {@link #initializeHttp(HttpInitializer)} and use
+     * {@link SingleAddressHttpClientBuilder#executor(Executor)} on the {@code builder} instance
+     * by implementing {@link HttpInitializer#initialize(SingleAddressHttpClientBuilder)} functional interface.
      */
+    @Deprecated
     public abstract GrpcClientBuilder<U, R> executor(Executor executor);
```
idelpivnitskiy, line 98: We should add default implementations for all `@Deprecated` methods

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
```diff
@@ -79,11 +87,20 @@ public HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strat
     @Nullable
     private Duration defaultTimeout;
     private boolean invokedBuild;
+    @Nullable
+    private HttpInitializer<U, R> httpInitializer;
```
idelpivnitskiy, line 91: Consider assigning a noop impl, then we should not worry about `@Nullable`

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ErrorHandlingTest.java
```diff
@@ -263,9 +263,12 @@ private void setUp(TestMode testMode, GrpcExecutionStrategy serverStrategy,
...
         serverContext = GrpcServers.forAddress(localAddress(0)).appendHttpServiceFilter(serviceFilterFactory)
                 .executionStrategy(serverStrategy).listenAndAwait(serviceFactory);
+        final StreamingHttpClientFilterFactory pickedClientFilterFactory = clientFilterFactory;
         GrpcClientBuilder<HostAndPort, InetSocketAddress> clientBuilder =
-                GrpcClients.forAddress(serverHostAndPort(serverContext)).appendHttpClientFilter(clientFilterFactory)
-                .executionStrategy(clientStrategy);
+                GrpcClients.forAddress(serverHostAndPort(serverContext))
+                        .initializeHttp(builder -> builder
+                                .appendClientFilter(pickedClientFilterFactory)
+                                .executionStrategy(clientStrategy));
```
idelpivnitskiy, line 271: If we migrate to `SingleAddressHttpClientBuilder#executionStrategy`, we should also migrate to `HttpExecutionStrategy` instead of `GrpcExecutionStrategy`.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
```diff
@@ -93,7 +127,11 @@
      * @param strategy {@link GrpcExecutionStrategy} to use.
      * @return {@code this}.
      * @see GrpcExecutionStrategies
+     * @deprecated Call {@link #initializeHttp(HttpInitializer)} and use
+     * {@link SingleAddressHttpClientBuilder#executionStrategy(HttpExecutionStrategy)} on the {@code builder} instance
+     * by implementing {@link HttpInitializer#initialize(SingleAddressHttpClientBuilder)} functional interface.
      */
+    @Deprecated
```
idelpivnitskiy, line 134: If we deprecate this method, we should also deprecate `GrpcExecutionStrategy` and `GrpcExecutionStrategies`.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcClientRequiresTrailersTest.java
```diff
@@ -105,7 +105,7 @@ protected HttpHeaders payloadComplete(final HttpHeaders trailers) {
                 });
 
         client = GrpcClients.forAddress(serverHostAndPort(serverContext))
-                .executionStrategy(noOffloadsStrategy())
+                .initializeHttp(builder -> builder.executionStrategy(noOffloadsStrategy()))
```
idelpivnitskiy, line 108: Migrate to `HttpExecutionStrategies.noOffloadsStrategy()`

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
```diff
@@ -237,4 +256,25 @@ public void doAppendHttpClientFilter(final Predicate<StreamingHttpRequest> predi
                                          final StreamingHttpClientFilterFactory factory) {
         httpClientBuilder.appendClientFilter(predicate, factory);
     }
+
+    private void appendCatchAllFilter() {
```
idelpivnitskiy, line 260: Hm, I didn't know that we have "CatchAll" filter for the client. Not sure why it's necessary because `DefaultGrpcClientCallFactory` will map all exceptions correctly. Any thoughts on why it can be useful?

idelpivnitskiy (changes requested):

PR #1884, opened 2021-10-12
gRPC catch-all filters should implement HttpExecutionStrategyInfluencer

idelpivnitskiy:
Motivation:

gRPC client and server unconditionally adds a "catch-all" filter, which
does not implement `HttpExecutionStrategyInfluencer`, leading to
unexpected offloading for users when they use aggregated and blocking
api.

Modifications:

- Implement `HttpExecutionStrategyInfluencer` for both:
`CatchAllHttpServiceFilter` and `CatchAllHttpClientFilter`;

Result:

No unexpected offloading when gRPC users use aggregated or blocking API.


idelpivnitskiy:
Will be cherry-picked after #1867 and #1861.

chemicL:
> Will be cherry-picked after #1867 and #1861.

Both are now cherry-picked: f249032a4009f321ac9316187e5807d65b38b815 and f8aaa8dd5cb1d72bb3f342d3293eb4beb1b036b7

PR #1905, opened 2021-10-22
Unified retry / auto-retry filter support

tkountis:
Motivation:

Use of `AutoRetry` alongside a `RetryingHttpRequesterFilter` creates issues with the retrying logic exponentially increasing to the product of the two max-retry settings. The conflict arises when both solutions are configured to retry `RetryableExceptions` which is the case by default.

Modifications:

`AutoRetry` logic is deprecated, to be removed in the future.
The logic of the `AutoRetry` is merged with the `RetryingHttpRequesterFilter` which now allows retryable flows. Different exceptions can have different retry, as well as support for retries of responses.

Results:

Unified API


servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/RetryingHttpRequesterFilter.java
```diff
@@ -17,56 +17,134 @@
...
+ * {@link SingleAddressHttpClientBuilder#appendClientFilter(StreamingHttpClientFilterFactory)}.
  * @see RetryStrategies
  */
-public final class RetryingHttpRequesterFilter implements StreamingHttpClientFilterFactory,
-                                                          StreamingHttpConnectionFilterFactory,
-                                                          HttpExecutionStrategyInfluencer {
+public final class RetryingHttpRequesterFilter
+        implements StreamingHttpClientFilterFactory, StreamingHttpConnectionFilterFactory,
+                   HttpExecutionStrategyInfluencer, InjectableStreamingClientFilterFactory,
+                   LoadBalancerReadinessAware, ServiceDiscoveryStatusAware {
```
idelpivnitskiy, line 95: Main issues with the approach proposed in this PR:
1. There are no retries by default, which breaks the client for all users because it has to wait for LB to be ready. 
2. If a filter is applied as `StreamingHttpConnectionFilterFactory`, there are multiple issues:
   a. wait for LB and SD error visibility won't work because we can not create a connection without having LB ready.
   b. `RetryableException` has to be handled at a client level, because LB may throw it.
3. SD and LB are protocol agnostic abstractions and expected to be reused outside HTTP. This is why they are defined in client-api. This is why `AutoRetryStrategyProvider` also lives in client-api.
4. `InjectableStreamingClientFilterFactory`, `LoadBalancerReadinessAware`, and `ServiceDiscoveryStatusAware` interfaces look like hacks in API to workaround a specific case. I don't see any other cases when this API will be useful for users and how they will discover that these hooks exist. Can you provide examples?

idelpivnitskiy (changes requested):

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/RetryingHttpRequesterFilter.java
tkountis: Discussed offline.

1. By default auto-retries are enabled (if they weren't disabled), but using the new filter. If the user provides a custom Retrying Filter, then auto-retry will be skipped.
2. Decided to remove support for connection based filters. Code was adjusted.
3. Decided to merge this together since there is no immediate need for supporting multiple protocols.
4. Agreed, removed.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilterBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Overrides the default criterion for determining which responses should be retried.
+     * <strong>It's important that this {@link Function} doesn't block to avoid performance impacts.</strong>
+     *
+     * @param mapper {@link Function} that checks whether a given {@link HttpResponseMetaData meta-data} should
+     * be retried, producing a {@link BackOffPolicy} in such cases.
+     * @return {@code this}
+     */
+    public RetryingHttpRequesterFilterBuilder retryResponses(
+            final Function<HttpResponseMetaData, BackOffPolicy> mapper) {
```
idelpivnitskiy, line 97: Today, if users want to retry a response, they can map a status code into some exception type and handle it using a function that take `HttpRequestMetaData` and a `Throwable`. Do we have to add this new API?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -0,0 +1,590 @@
...
+                                    maxRetries, t -> true, initialDelay, jitter, effectiveExecutor);
+                }
+            }
+        }
+    }
+
+    /**
+     * Default request retry policy builder.
+     */
+    public static final class DefaultRequestRetryPolicyBuilder {
```
idelpivnitskiy, line 489: It's hard to discover that this class exists. Did you consider moving these methods to the filter builder? All boolean flags + user-provided `BiFunction` can be joined using logical "or".

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilterBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+     * Overrides the default criterion for determining which requests or errors should be retried.
+     * <strong>It's important that this {@link Function} doesn't block to avoid performance impacts.</strong>
+     *
+     * @param mapper {@link BiFunction} that checks whether a given combination of
+     * {@link HttpRequestMetaData meta-data} and {@link Throwable cause} should be retried, producing a
+     * {@link BackOffPolicy} in such cases.
+     * @return {@code this}
+     */
+    public RetryingHttpRequesterFilterBuilder retryRequests(
+            final BiFunction<HttpRequestMetaData, Throwable, BackOffPolicy> mapper) {
```
idelpivnitskiy, line 112: Always requiring to return a `BackOffPolicy` looks like an overkill. Did you consider keeping both: a `BiPredicate` and `BiFunction`? Then users may choose `BiPredicate` if they don't need to worry about different `BackOffPolicies`. A builder method that takes `BiPredicate` may take `BackOffPolicy` as a 2nd argument.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -520,6 +549,18 @@ private static StreamingHttpConnectionFilterFactory appendConnectionFilter(
     public DefaultSingleAddressHttpClientBuilder<U, R> appendClientFilter(
             final StreamingHttpClientFilterFactory factory) {
         requireNonNull(factory);
+        // TODO 0.42 remove this extra test with class-name
+        if (factory instanceof RetryingHttpRequesterFilter ||
+                factory.getClass().getSimpleName().contains("RetryingHttpRequesterFilter")) {
```
idelpivnitskiy, line 554: Can you clarify why the class-name check is also necessary? If it's for `RetryingHttpRequesterFilter` from utils, consider using the full name or check the instance/class, http-utils are visible in http-netty.

idelpivnitskiy (changes requested):
I like where we are moving, looks good to combine two retry spots together. 
What will be your strategy for testing?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilterBuilder.java
tkountis: I think having everything in one place is more user friendly + results in one filter vs having more to do the mapping.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -0,0 +1,590 @@
...
+        @Override
+        public String toString() {
+            return "StacklessRetryResponseException{ metaData=" + metaData.toString(DEFAULT_HEADER_FILTER) + '}';
+        }
+    }
+
+    /**
+     * Definition and presets of retry backoff policies.
+     */
+    public static class BackOffPolicy {
```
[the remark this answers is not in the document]
tkountis: It's just a wrapper that allows support for a runtime provided 'Executor` + helpers to allow instant executions.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
tkountis: The reason this was introduced was that we have different APIs for retrying requests & responses. Having request flavored flags on the builder, where response retrying abilities are also provided is misleading.
I can advertise this class further on the request retrying function API. WDYT?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilterBuilder.java
idelpivnitskiy: We have 2 parallel discussions that are related:
1. `retryResponses`
2. `DefaultRequestRetryPolicyBuilder`
Let me continue here touching both of these points:

As you can see, it gets challenging to expose `retryResponses` and `retryRequests` on one filter. Not only because those are different. There are other related methods, like `maxTotalRetries`, which users may want to be different for request error or response status.

WDYT, if instead of two different retry points we still keep only one that handles `HttpRequestMetaData` and `Throwable`, but instead of `retryResponses(Function<HttpResponseMetaData, BackOffPolicy>)` we introduce something like `mapResponse(Function<HttpResponseMetaData, HttpResponseException>)`? `HttpResponseException` will have a reference to `HttpResponseMetaData`. That way users can configure mapping and keep using a single retry function (rename `retryRequests` to `retryOn/retryWhen`) to handle everything.

tkountis: Yup that's a good suggestion. Will do that now

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -0,0 +1,693 @@
...
+        private BiFunction<HttpRequestMetaData, IOException, BackOffPolicy>
+                retryIdempotentRequests;
+
+        @Nullable
+        private BiFunction<HttpRequestMetaData, DelayedRetry, BackOffPolicy>
+                retryDelayedRetries;
+
+        @Nullable
+        private BiFunction<HttpRequestMetaData, Throwable, BackOffPolicy>
+                retryOther;
```
idelpivnitskiy, line 535: How do you feel, should we have another `BiFunction` for `HttpResponseException` or we expect users to handle it by `retryOther`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -0,0 +1,693 @@
...
+
+                        if (retryDelayedRetries != null && throwable instanceof DelayedRetry) {
+                            final BackOffPolicy backOffPolicy =
+                                    retryDelayedRetries.apply(requestMetaData, (DelayedRetry) throwable);
+                            if (backOffPolicy != NO_RETRIES) {
+                                return backOffPolicy;
+                            }
+                        }
+
+                        if (retryRetryableExceptions != null && throwable instanceof RetryableException) {
```
idelpivnitskiy, line 676: Should this be the first criteria (even before `retryIdempotentRequests`)?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -0,0 +1,693 @@
...
+            return this;
+        }
+
+        public Builder responseMapper(final Function<HttpResponseMetaData, HttpResponseException> mapper) {
+            this.responseMapper = requireNonNull(mapper);
+            return this;
+        }
+
+        /**
+         * The retrying-filter will evaluate for {@link RetryableException}s in the request flow.
```
idelpivnitskiy, line 584: For all functions, consider clarifying that `ofNoRetries()` can be used to avoid retries

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -0,0 +1,693 @@
...
+        @Override
+        public String toString() {
+            return "HttpResponseException{ metaData=" + metaData.toString(DEFAULT_HEADER_FILTER) + '}';
+        }
+    }
+
+    /**
+     * Definition and presets of retry backoff policies.
+     */
+    public static class BackOffPolicy {
```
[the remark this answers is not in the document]
tkountis: That was intentional, in case users want to extend it. Its also mentioned in the root doc; moved it here now.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -0,0 +1,693 @@
...
+        }
+
+        @Override
+        public synchronized Throwable fillInStackTrace() {
+            return this;
+        }
+
+        @Override
+        public String toString() {
+            return "HttpResponseException{ metaData=" + metaData.toString(DEFAULT_HEADER_FILTER) + '}';
```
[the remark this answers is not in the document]
tkountis: Since we made it final, yes that makes sense.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
tkountis: I was 50-50 about that, since you mentioned it, I will add.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -238,13 +234,19 @@ public Completable closeAsyncGracefully() {
...
+        private static final long serialVersionUID = -7182949760823647710L;
 
-        private final HttpResponseMetaData metaData;
+        public final HttpResponseMetaData metaData;
+        public final String message;
 
-        public HttpResponseException(final HttpResponseMetaData metaData) {
+        public HttpResponseException(final String message, final HttpResponseMetaData metaData) {
             this.metaData = requireNonNull(metaData);
+            this.message = requireNonNull(message);
```
idelpivnitskiy, line 249: The message should be passed to the super class

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -254,12 +256,18 @@ public synchronized Throwable fillInStackTrace() {
 
         @Override
         public String toString() {
-            return "HttpResponseException{ metaData=" + metaData.toString(DEFAULT_HEADER_FILTER) + '}';
+            return "HttpResponseException{" +
+                    "metaData=" + metaData +
+                    ", message='" + message + '\'' +
```
idelpivnitskiy, line 261: Consider using `super.toString() + ", metaData=" + metaData` to preserve the original exception formatting

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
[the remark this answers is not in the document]
idelpivnitskiy: Can you provide an example why users may need to extend it? It anyway does not expose its internal state publicly. What can they do if they extend?
If extensibility is important, should we make it an interface?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
idelpivnitskiy: Don't see a change for this

idelpivnitskiy (approved):
Last comments then LGTM, thank you!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
tkountis: It's missing from one function (added), but it's there for the rest.
> To disable retries you can return {@link BackOffPolicy#NO_RETRIES} from the {@code mapper}.

tkountis:
>Can you provide an example why users may need to extend it? It anyway does not expose its internal state publicly. What can they do if they extend?
If extensibility is important, should we make it an interface?

You are right, I had in mind to make it extensible (thus the doc) but never made the fields public. I agree we can keep final for now.

PR #1921, opened 2021-10-29
`ConnectionAcceptor` `ConnectionFactory` offloading control

bondolo:
Motivation:
Currently `ConnectionAcceptor::accept` and
`ConnectionFactory::newConnection` are always offloaded. This may
be unnecessary.
Modifications:
By implementing `ExecutionStrategyInfluencer` and conditionally
offloading only when the required strategy of the acceptor or factory
is not `anyStrategy()` unnecessary offloading is avoided.
Result:
Potentially less offloading for connection acceptors and factories.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionFactoryFilter.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionFactoryFilter.java
@@ -16,6 +16,7 @@
 package io.servicetalk.client.api;
 
 import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
+import io.servicetalk.transport.api.ConnectExecutionStrategy;
 import io.servicetalk.transport.api.ExecutionStrategy;
 import io.servicetalk.transport.api.ExecutionStrategyInfluencer;
 
@@ -76,6 +77,8 @@ public interface ConnectionFactoryFilter<ResolvedAddress, C extends ListenableAs
     /**
      * {@inheritDoc}
      *
+     * <p>If the returned strategy extends {@link ConnectExecutionStrategy} then the connection creation or accept may
+     * be offloaded.
      * <p>If the returned strategy extends {@code HttpExecutionStrategy} then the HTTP execution strategy will be
      * applied to the connections created.</p>
      * <p>A utility class provides the ability to combine connect and HTTP execution strategies,
@@ -83,8 +86,8 @@ public interface ConnectionFactoryFilter<ResolvedAddress, C extends ListenableAs
      */
     @Override
     default ExecutionStrategy requiredOffloads() {
-        // safe default--implementations are expected to override
-        return ExecutionStrategy.offloadAll();
+        // safe default--implementations are expected to override if offloading is required.
+        return ConnectExecutionStrategy.anyStrategy();
     }
 
     /**
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ConnectAndHttpExecutionStrategy.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ConnectAndHttpExecutionStrategy.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.api;
+
+import io.servicetalk.transport.api.ConnectExecutionStrategy;
+import io.servicetalk.transport.api.ExecutionStrategy;
+
+import java.util.Objects;
+
+/**
+ * Combines a {@link ConnectExecutionStrategy} and an {@link HttpExecutionStrategy}.
+ */
+public final class ConnectAndHttpExecutionStrategy implements ConnectExecutionStrategy, HttpExecutionStrategy {
+
+    final ConnectExecutionStrategy connectStrategy;
+    final HttpExecutionStrategy httpStrategy;
+
+    public ConnectAndHttpExecutionStrategy(ConnectExecutionStrategy connectStrategy) {
+        this(connectStrategy, HttpExecutionStrategies.anyStrategy());
+    }
+
+    public ConnectAndHttpExecutionStrategy(HttpExecutionStrategy httpStrategy) {
+        this(ConnectExecutionStrategy.anyStrategy(), httpStrategy);
+    }
+
+    public ConnectAndHttpExecutionStrategy(ConnectExecutionStrategy connect, HttpExecutionStrategy http) {
+        this.connectStrategy = connect;
+        this.httpStrategy = http;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        final ConnectAndHttpExecutionStrategy that = (ConnectAndHttpExecutionStrategy) o;
+        return connectStrategy.equals(that.connectStrategy) && httpStrategy.equals(that.httpStrategy);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(connectStrategy, httpStrategy);
+    }
+
+    @Override
+    public String toString() {
+        return connectStrategy + " " + httpStrategy;
+    }
+
+    @Override
+    public boolean hasOffloads() {
+        return connectStrategy.hasOffloads() || httpStrategy.hasOffloads();
+    }
+
+    @Override
+    public boolean isMetadataReceiveOffloaded() {
+        return httpStrategy.isMetadataReceiveOffloaded();
+    }
+
+    @Override
+    public boolean isDataReceiveOffloaded() {
+        return httpStrategy.isDataReceiveOffloaded();
+    }
+
+    @Override
+    public boolean isSendOffloaded() {
+        return httpStrategy.isSendOffloaded();
+    }
+
+    @Override
+    public boolean isConnectOffloaded() {
+        return connectStrategy.isConnectOffloaded();
+    }
+
+    @Override
+    public ConnectAndHttpExecutionStrategy merge(final ExecutionStrategy other) {
+        if (other instanceof ConnectAndHttpExecutionStrategy) {
+            return merge((ConnectAndHttpExecutionStrategy) other);
+        } else if (other instanceof HttpExecutionStrategy) {
+            return merge((HttpExecutionStrategy) other);
+        } else if (other instanceof ConnectExecutionStrategy) {
+            return merge((ConnectExecutionStrategy) other);
+        } else {
+            return other.hasOffloads() ?
+                    new ConnectAndHttpExecutionStrategy(
+                            ConnectExecutionStrategy.offload(),
+                            HttpExecutionStrategies.offloadAll())
+                    : this;
+        }
+    }
+
+    private ConnectAndHttpExecutionStrategy merge(final ConnectExecutionStrategy other) {
+        ConnectExecutionStrategy merged = connectStrategy.merge(other);
+        return merged == connectStrategy ? this : new ConnectAndHttpExecutionStrategy(merged, httpStrategy);
+    }
+
+    @Override
+    public ConnectAndHttpExecutionStrategy merge(final HttpExecutionStrategy other) {
+        HttpExecutionStrategy merged = httpStrategy.merge(other);
+        return merged == httpStrategy ? this : new ConnectAndHttpExecutionStrategy(connectStrategy, merged);
+    }
+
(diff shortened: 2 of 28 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SpecialConnectExecutionStrategy.java
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.transport.api;
+
+/**
+ * package private constants for special Execution
+ */
+enum SpecialConnectExecutionStrategy implements ConnectExecutionStrategy {
```
Scottmitch, line 21: `SpecialConnectExecutionStrategy` -> "Special" as it is used here doesn't quite mean the same thing as it does when it is used in other contexts (e.g. `SpecialHttpExecutionStrategy` has "default" / "never" which are treated "special" all over the code base ... and we plan to remove the special treatment eventually). Consider using different terminology (e.g. `DefaultHttpExecutionStrategy`)

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/InfluencerConnectionAcceptor.java
```diff
@@ -0,0 +1,76 @@
...
+import static io.servicetalk.concurrent.api.Completable.completed;
+import static io.servicetalk.transport.api.ConnectExecutionStrategy.anyStrategy;
+
+/**
+ * A contract that defines the connection acceptance criterion.
+ *
+ * @see ConnectionAcceptorFactory
+ */
+@FunctionalInterface
+public interface InfluencerConnectionAcceptor extends ConnectionAcceptor,
```
Scottmitch, line 34: Any reason we can't use the same pattern established with `HttpExecutionStrategyInfluencer` (an influencer interface, and the filter/factory extends the influencer)?

suggested:
```java
// rename InfluencerConnectionAcceptor to ConnectExecutionStrategyInfluencer
interface ConnectExecutionStrategyInfluencer extends ExecutionStrategyInfluencer<ConnectExecutionStrategy> {}


interface ConnectionAcceptor extends ConnectExecutionStrategyInfluencer {}
```

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
```diff
@@ -131,7 +132,8 @@ protected void initChannel(Channel channel) {
                             // of connection processing.
                             defer(() -> connectionAcceptor.accept(conn).concat(succeeded(conn)))
                                     // subscribeOn is required to offload calls to connectionAcceptor#accept
-                                    .subscribeOn(executionContext.executor()));
+                                    .subscribeOn(executionContext.executor(),
+                                            connectionAcceptor.requiredOffloads()::hasOffloads));
```
Scottmitch, line 136: `hasOffloads` -> using this method in instances where we are applying targeted offloading makes extensibility of interface limited. For example if we add another influencer method later (e.g. for closing, etc.) then this case will start offloading when it shouldn't have to. Seems like we should use `isConnectOffloaded` here.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/PipelinedLBHttpConnectionFactory.java
```diff
@@ -40,12 +40,13 @@
...
             final StreamingHttpRequestResponseFactory reqRespFactory,
-            final HttpExecutionStrategy chainStrategy,
+            final ExecutionStrategy connectStrategy,
             final ConnectionFactoryFilter<ResolvedAddress, FilterableStreamingHttpConnection> connectionFactoryFilter,
             final Function<FilterableStreamingHttpConnection,
                     FilterableStreamingHttpLoadBalancedConnection> protocolBinding) {
-        super(config, executionContext, connectionFilterFunction, version -> reqRespFactory, chainStrategy,
-                connectionFactoryFilter, protocolBinding);
+        super(config, executionContext, version -> reqRespFactory,
+                connectStrategy, connectionFactoryFilter, connectionFilterFunction,
```
Scottmitch, line 48: nit: we generally try to fill the 120 chars up on a line unless it impacts readability. the additional newline here doesn't seem to help readability.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionFactoryFilter.java
```diff
@@ -76,6 +77,8 @@
     /**
      * {@inheritDoc}
      *
+     * <p>If the returned strategy extends {@link ConnectExecutionStrategy} then the connection creation or accept may
+     * be offloaded.
      * <p>If the returned strategy extends {@code HttpExecutionStrategy} then the HTTP execution strategy will be
      * applied to the connections created.</p>
      * <p>A utility class provides the ability to combine connect and HTTP execution strategies,
```
Scottmitch, line 84: can you explain scenarios where users are expected to combine the different strategies? I would expect that is only something we do internally while computing the execution strategies.

If this can be limited to an internal implementation detail (e.g. via `merge`) then we can make `ConnectAndHttpExecutionStrategy` internal (and remove the reference in this comment).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
```diff
@@ -50,21 +54,24 @@ void add(HttpLoadBalancerFactory<?> lb) {
...
     void add(StreamingHttpConnectionFilterFactory connectionFilter) {
-        connFilterChain =
-                connFilterChain.merge(HttpExecutionStrategy.from(connectionFilter.requiredOffloads()));
+        connFilterChain = connFilterChain.merge(connectionFilter.requiredOffloads());
     }
 
     HttpExecutionStrategy buildForClient(HttpExecutionStrategy transportStrategy) {
-        return clientChain.merge(buildForConnectionFactory(transportStrategy));
+        return transportStrategy.merge(clientChain.merge(connFilterChain.merge(
+                HttpExecutionStrategy.from(buildForConnectionFactory()))));
```
Scottmitch, line 66: #### How offloading is applied/combined
Currently the strategy for the connection factory and transport are being combined together. This means that if the connection factory blocks, then each request will be offloaded. However the connection factory is unlikely to be invoked on each request bcz the LB is assumed to pool connections.  Is it the intention to always combine the strategies into a HttpExecutionStrategy, and apply it at the transport level? If so, this `buildForClient` method is only used when converting to other API types in `FilterableClientToClient` so I'm not sure the offloading will be used in the expected spot.

Can you describe how all offloading strategies are expected to combine/interact on the client side, and how API optimizations are applied if default offloading is used? (`ClientFilter -> LoadBalancer -> ConnectionFactoryFilter -> ConnectionFilter -> Transport`)

#### Fallback method for combining
`HttpExecutionStrategy.from` fallback if a `HttpExecutionStrategy` is not returned will offload more aggressively than necessary (assuming we need to combine request-path offloading with connection creation offloading). For example for `ConnectFactory` only needs to offload when creating a connection. I wonder when we expect this condition to happen, and if we can rely upon the `merge` method of the runtime types to return the correct merge (or fallback to offload all if not implemented correctly, this is the solution we previous discussed).

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ConnectAndHttpExecutionStrategy.java
```diff
@@ -0,0 +1,132 @@
...
+        return merged == connectStrategy ? this : new ConnectAndHttpExecutionStrategy(merged, httpStrategy);
+    }
+
+    @Override
+    public ConnectAndHttpExecutionStrategy merge(final HttpExecutionStrategy other) {
+        HttpExecutionStrategy merged = httpStrategy.merge(other);
+        return merged == httpStrategy ? this : new ConnectAndHttpExecutionStrategy(connectStrategy, merged);
+    }
+
+    public ConnectAndHttpExecutionStrategy merge(final ConnectAndHttpExecutionStrategy other) {
```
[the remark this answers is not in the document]
bondolo: I ended up changing the type to `ConnectAndHttpExecutionStrategy merge(final  ExecutionStrategy other)` for better consistency and to include the conversion from other ExecutionStrategy types.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
bondolo: Briefly, the current combination of strategies offered is because connection filters can alter the http strategy required. The is independent of their needs for offloading of the connection establishment step.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/InfluencerConnectionAcceptor.java
Scottmitch: discussed offline. 👍 my misunderstanding on how this was intended to be used. this interface is internal only.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionFactoryOffloadingTest.java
```diff
@@ -0,0 +1,108 @@
...
+                HttpResponse response = single.toFuture().get();
+                assertThat("unexpected status", response.status(), is(HttpResponseStatus.OK));
+            }
+        }
+        assertTrue((offload && !IoThreadFactory.IoThread.isIoThread(factoryThread.get())) ||
+                        (!offload && factoryThread.get() == appThread),
+                "incorrect offloading, offload=" + offload + " thread=" + factoryThread.get());
+    }
+
+    private Single<HttpResponse> helloWorld(HttpServiceContext ctx,
```
Scottmitch, line 103: do we have a test that tests `ConnectionFactoryFilter` ability to return `HttpExecutionStrategy` and `ConnectExecutionStrategy + HttpExecutionStrategy`? if not can you plz add one as the merging/combining is relatively complex internally so good to verify we don't break it in the future.

Scottmitch (approved):

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectExecutionStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+    default boolean hasOffloads() {
+        return isConnectOffloaded();
+    }
+
+    /**
+     * Returns true if connection creation or accept requires offloading.
+     *
+     * @return true if connection creation or accept requires offloading
+     */
+    boolean isConnectOffloaded();
```
idelpivnitskiy, line 38: Did you consider promoting this flag to `ExecutionStrategy`. Understand that `HttpExecutionStrategy` will also inherit it and it will have no influence on HTTP filters. However, it might be useful for users to specify it on the client/server builder to explicitly tell if they need offloading for connect or not.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionFactoryFilter.java
bondolo: In the simplest case we would only need the `ConnectExecutionStrategy` to specify whether the connection factory, filters or acceptor requires offloading.

The additional capability `ConnectAndHttpExecutionStrategy` is needed for cases where the connection filter modifies the handling of request or otherwise changes the behaviour of the connection. It is equivalent to the connection filter installing a streaming filter per connection.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
```diff
@@ -131,7 +132,8 @@ protected void initChannel(Channel channel) {
                             // of connection processing.
                             defer(() -> connectionAcceptor.accept(conn).concat(succeeded(conn)))
                                     // subscribeOn is required to offload calls to connectionAcceptor#accept
-                                    .subscribeOn(executionContext.executor()));
+                                    .subscribeOn(executionContext.executor(),
+                                            connectionAcceptor.requiredOffloads()::isConnectOffloaded));
```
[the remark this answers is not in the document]
bondolo: It would not be dynamic so could be applied only as needed. I will adjust this.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectExecutionStrategy.java
bondolo: I have a lot of concerns about flags being present in situations where they aren't used and similarly about the implications of adding flags in the future that existing usages don't contemplate. It is not always clear that a new flag with appropriate default behavior can be cleanly added. It seems that in a lot of cases the default for the new flag, and the absence of providing an appropriate response for the flag by existing code could result in changes in behaviour (resulting from the application of the default) just by introducing a new flag.

I have tried to keep the flags very specific to the context in which they are used so that new flags can be safely added and implementors don't have to worry about affecting behaviour for features they are not using.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionFactoryOffloadingTest.java
bondolo: Now added. The `ConnectionFactoryOffloadingTest` now exercises the use of `ConnectAndHttpExecutionStrategy`.

PR #1923, opened 2021-10-29
Enhance GrpcStatus.fromCodeValue static factories to include actual status code when it maps to UNKNOWN

cancecen:
Motivation:
Preserving the original code value in the description when the code is not recognized. This is also consistent with grpc-java.

Modifications:
Pass in the original error code to `GrpcStatus` in the static factories.

Result:
GrpcStatus with UNKNOWN status code preserves the ogirinal code value.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatus.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatus.java
@@ -17,7 +17,6 @@ package io.servicetalk.grpc.api;
 
 import com.google.rpc.Status;
 
-import java.util.Objects;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatus.java
```diff
@@ -178,6 +179,13 @@ public String description() {
         return description;
     }
 
+    public GrpcStatus withDescription(String description) {
```
idelpivnitskiy, line 182: We try to avoid adding new public API if it's not required. Consider using existing public ctor `GrpcStatus(GrpcStatusCode, null, description)` instead.

idelpivnitskiy (changes requested):
Thank you for the PR @cancecen!

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatus.java
cancecen: sounds good, fixed.

idelpivnitskiy (approved):
Thank you!

PR #1924, opened 2021-10-29
Adjust parent types for exceptions

idelpivnitskiy:
Motivation:

After revisiting all existing exception types, we found that some of
them use incorrect parent type.

Modifications:

- Use more appropriate parent type for existing exceptions:
  - `ConnectionRejectedException` extend `RuntimeException` -> `SocketException`;
  - `NoAvailableHostException` extend `RuntimeException` -> `IOException`;
  - `DuplicateAttributeException` extend `RuntimeException` -> `IllegalStateException`;
  - `TerminateRepeatException` extend `Exception` -> `RuntimeException`;
  - `ClosedServiceDiscovererException` extend `RuntimeException` -> `ClosedChannelException`;
  - `UnsupportedHttpChunkException` extend `SerializationException` -> `IllegalArgumentException`;
- Regenerate `serialVersionUID` for updated exception classes;
- Use `StacklessConnectionRejectedException` for
`RoundRobinLoadBalancer`;
- Mark `MessageEncodingException` as `@Deprecated` (was missed during rework of 
serialization API);
- Remove unused ctor from `ResponseTimeoutTest`;

Result:

Exceptions have more appropriate parent types.


servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionRejectedException.java
```diff
@@ -17,11 +17,13 @@
...
 import io.servicetalk.transport.api.RetryableException;
 
+import java.net.SocketException;
+
 /**
  * Thrown when a newly created connection was rejected.
  */
-public final class ConnectionRejectedException extends RuntimeException implements RetryableException {
-    private static final long serialVersionUID = -2573006397795678870L;
+public class ConnectionRejectedException extends SocketException implements RetryableException {
```
idelpivnitskiy, line 25: Debated between `ConnectException` and `SocketException`. Based on their javadoc, decided to go with `SocketException` because of technically the connection was already established, but there was a problem "accessing" it due to a selector or a host becoming unavailable.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RepeatStrategies.java
```diff
@@ -40,8 +40,8 @@
     /**
      * An {@link Exception} instance used to indicate termination of repeats.
      */
-    public static final class TerminateRepeatException extends Exception {
-        private static final long serialVersionUID = -1725458427890873886L;
+    public static final class TerminateRepeatException extends RuntimeException {
```
idelpivnitskiy, line 43: The decision to throw this exception happens at runtime

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/MessageEncodingException.java
```diff
@@ -15,10 +15,18 @@
...
+
 /**
  * Exception thrown when a message was encoded with an unsupported encoder.
+ *
+ * @deprecated This exception type was thrown only by {@code ProtoBufSerializationProviderBuilder} that was deprecated
+ * and will be removed in future releases. Use {@link SerializationException} instead that can be thrown by a new
+ * implementation.
  */
+@Deprecated
 public final class MessageEncodingException extends RuntimeException {
```
idelpivnitskiy, line 28: `ProtoBufSerializationProviderBuilder` was deprecated only in the main branch, we don't need to cherry-pick this into 0.41

servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProviderBuilder.java
```diff
@@ -172,7 +172,7 @@ public GrpcSerializationProvider build() {
             @SuppressWarnings("unchecked")
             HttpSerializer<T> httpSerializer = serializersForType.get(codec);
             if (httpSerializer == null) {
-                throw new MessageEncodingException("Unknown encoding: " + codec.name());
+                throw new MessageEncodingException(codec.name().toString());
```
idelpivnitskiy, line 175: Full message is generated inside a ctor

servicetalk-http-api/src/main/java/io/servicetalk/http/api/UnsupportedHttpChunkException.java
```diff
@@ -15,11 +15,9 @@
  */
 package io.servicetalk.http.api;
 
-import io.servicetalk.serializer.api.SerializationException;
-
 import javax.annotation.Nullable;
 
-final class UnsupportedHttpChunkException extends SerializationException {
+final class UnsupportedHttpChunkException extends IllegalArgumentException {
```
idelpivnitskiy, line 20: The change from `IllegalArgumentException` to `SerializationException` was made only in the main branch in #1673. This is not related to serialization. Reverting back to `IllegalArgumentException`.
cc @Scottmitch

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -263,21 +265,32 @@ public StreamingHttpClient apply(final UrlKey urlKey) {
...
         }
 
         @Override
         public Single<StreamingHttpResponse> request(final HttpExecutionStrategy strategy,
                                                      final StreamingHttpRequest request) {
-            return defer(() -> selectClient(request).request(strategy, request).subscribeShareContext());
+            return defer(() -> {
+                try {
+                    return selectClient(request).request(strategy, request).subscribeShareContext();
+                } catch (Throwable t) {
```
idelpivnitskiy, line 290: The `try-catch` here is necessary anyway, because methods that return async sources are not expected to throw.

idelpivnitskiy (commented):
I don't expect it to be an impactful breaking change because most of the users use our types instead of parent types. Should be an appropriate change for 0.42 release, wdyt?

servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProviderBuilder.java
idelpivnitskiy: Decided to open a separate PR #1940 because we should also cherry-pick this into 0.41

servicetalk-client-api/src/main/java/io/servicetalk/client/api/NoAvailableHostException.java
```diff
@@ -17,11 +17,13 @@
...
 import io.servicetalk.transport.api.RetryableException;
 
+import java.net.SocketException;
+
 /**
  * Thrown when no host is available but at least one is required.
  */
-public class NoAvailableHostException extends RuntimeException implements RetryableException {
-    private static final long serialVersionUID = 5340791072245425967L;
+public class NoAvailableHostException extends SocketException implements RetryableException {
```
tkountis, line 25: Is it a "SocketException"? I would rather keep it a runtime one or associate with a LB/SD parent type.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -63,6 +64,9 @@
  * A builder of {@link StreamingHttpClient} instances which have a capacity to call any server based on the parsed
  * absolute-form URL address information from each {@link StreamingHttpRequest}.
  * <p>
+ * If {@link HttpRequestMetaData#requestTarget()} is not an absolute-form URL, a {@link MalformedURLException} will be
```
bondolo, line 67: Possibly just a personal preference but I usually prefer `URISyntaxException` to `MalformedURLException` because the later often gets handled as a generic `IOException` by lazy people who hate checked exceptions.

bondolo (approved):

servicetalk-client-api/src/main/java/io/servicetalk/client/api/NoAvailableHostException.java
idelpivnitskiy: The intend was to make it `IOException` because it prevents a request execution due to unavailability to execute an IO operation. During offline discussion, @bondolo recommended a more specific subtype - `SocketException`. I'm agreed with that proposal because it notifies that there are no sockets available at this point of time.

tkountis: Don't see a reason to block the change but I don't agree that this is a good correlation. I agree that `IOException` is a valid super type, but `SocketException` is clearly doc'ed to do with failure to access or create a socket. In our case this is not even related to a connection but rather to host-finding.

idelpivnitskiy: Ok, I was reading "failure to access a socket" as "no hosts -> no access" 😄 
Changing to `IOException`: https://github.com/apple/servicetalk/pull/1924/commits/c2b41cf0f43e5aaf3ce8e7e4e4cbd8c796ef530f

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
idelpivnitskiy: I've opened a separate PR #1955 for this change to make it move visible in release notes

Issue #1889, opened 2021-10-15
Flaky test: MalformedDataAfterHttpMessageTest


chemicL:
https://github.com/apple/servicetalk/pull/1887/checks?check_run_id=3905793736

```
MalformedDataAfterHttpMessageTest > afterResponseNextClientRequestSucceeds() FAILED
java.lang.AssertionError: 
Expected: not <[id: 0x574fe329, L:/127.0.0.1:50732 ! R:127.0.0.1/127.0.0.1:39629]>
     but: was <[id: 0x574fe329, L:/127.0.0.1:50732 ! R:127.0.0.1/127.0.0.1:39629]>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at io.servicetalk.http.netty.MalformedDataAfterHttpMessageTest.afterResponseNextClientRequestSucceeds(MalformedDataAfterHttpMessageTest.java:142)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutInvocation.proceed(TimeoutInvocation.java:46)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:149)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:140)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:84)
[...]

PR #1925, opened 2021-10-29
Enhance javadoc to clarify StackOverflowError possibility

Scottmitch:
Motivation:
Publisher [repeat | retry] operators may cause StackOverflowError if too
many recovery operations are done consecutively on the same thread. This
condition can be difficult to detect as the onError recovery path is
unlikely to be called. We should highlight this possibility in the
javadocs for these methods.

Modifications:
- Publisher, Single, Completable repeat* and retry* operators have
  javadoc that clarifies the risk of StackOverflowError and mitigation
  technique.

Result:
Javadocs highlight risks of StackOverflowError and how it can be
avoided.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
@@ -867,7 +867,6 @@ public abstract class Completable {
      * avoided by trampolining the call stack onto an Executor. For example:
      *   {@code retryWhen((i, cause) -> i % 10 == 0 ? executor.submit(() -> { }) : Completable.completed())}
      * </pre>
-     * <p>
      * This method provides a means to retry an operation under certain failure conditions and in sequential programming
      * is similar to:
      * <pre>{@code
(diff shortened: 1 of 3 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -862,7 +862,11 @@ public final Completable mergeDelayError(Iterable<? extends Completable> other)
     /**
      * Re-subscribes to this {@link Completable} if an error is emitted and the passed {@link BiIntPredicate} returns
      * {@code true}.
-     * <p>
+     * <pre>
+     * This method may result in a {@link StackOverflowError} if too many consecutive calls are made. This can be
+     * avoided by trampolining the call stack onto an Executor. For example:
```
idelpivnitskiy, line 867: Consider `{@link Executor}`

idelpivnitskiy (approved):

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1889

Scottmitch:
0.41 (e300d27a5)

PR #1926, opened 2021-10-30
MalformedDataAfterHttpMessageTest.afterResponseNextClientRequestSucceeds fix

Scottmitch:
Motivation:
afterResponseNextClientRequestSucceeds attempts to validate that
NettyChannelPublisher preemptively closes the channel after request
processing concluded if there is an error pending to prevent subsequent
requests from using this connection. This condition may not be hit in
all scenarios due to sequencing of offloading as well as when bytes are
read from the network and the error condition is observed. This results
in the test being flaky and failing.

Modifications:
- Make afterResponseNextClientRequestSucceeds more robust by inserting a
  RetryingHttpRequesterFilter and adjusting the assertions to expect
  retry may happen.

Result:
Fixes https://github.com/apple/servicetalk/issues/1889.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MalformedDataAfterHttpMessageTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MalformedDataAfterHttpMessageTest.java
@@ -53,8 +53,8 @@ import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.nio.channels.ClosedChannelException;
 import java.util.Queue;
-import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.LinkedBlockingQueue;
 
 import static io.netty.buffer.ByteBufUtil.writeAscii;
 import static io.netty.util.ReferenceCountUtil.release;
@@ -126,7 +126,7 @@ class MalformedDataAfterHttpMessageTest {
     @ParameterizedTest
     @ValueSource(booleans = {true, false})
     void afterResponseNextClientRequestSucceeds(boolean doOffloading) throws Exception {
-        Queue<ConnectionContext> contextQueue = new ArrayBlockingQueue<>(4);
+        Queue<ConnectionContext> contextQueue = new LinkedBlockingQueue<>();
         ServerSocketChannel server = nettyServer(RESPONSE_MSG);
         try (BlockingHttpClient client = stClientBuilder(server.localAddress())
                 .executionStrategy(doOffloading ? defaultStrategy() : noOffloadsStrategy())
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MalformedDataAfterHttpMessageTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MalformedDataAfterHttpMessageTest.java
@@ -26,6 +26,7 @@ import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
 import io.servicetalk.http.api.StreamingHttpConnectionFilter;
 import io.servicetalk.http.api.StreamingHttpRequest;
 import io.servicetalk.http.api.StreamingHttpResponse;
+import io.servicetalk.http.utils.RetryingHttpRequesterFilter;
 import io.servicetalk.transport.api.ConnectionContext;
 import io.servicetalk.transport.api.HostAndPort;
 import io.servicetalk.transport.api.ServerContext;
@@ -44,6 +45,8 @@ import io.netty.handler.codec.http.HttpObjectAggregator;
 import io.netty.handler.codec.http.HttpRequestDecoder;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 import java.io.IOException;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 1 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MalformedDataAfterHttpMessageTest.java
```diff
@@ -118,11 +123,20 @@ void afterResponse() throws Exception {
...
+    @ValueSource(booleans = {true, false})
+    void afterResponseNextClientRequestSucceeds(boolean doOffloading) throws Exception {
         Queue<ConnectionContext> contextQueue = new ArrayBlockingQueue<>(4);
         ServerSocketChannel server = nettyServer(RESPONSE_MSG);
         try (BlockingHttpClient client = stClientBuilder(server.localAddress())
+                .executionStrategy(doOffloading ? defaultStrategy() : noOffloadsStrategy())
+                // ClosedChannelException maybe observed on the second request if write is done before read of the
+                // garbage data, which won't be a RetryableExcepton. We may also see an exception from flush if the read
+                // closed the connection and then attempt to write on the same connection.
+                .appendClientFilter(new RetryingHttpRequesterFilter.Builder()
```
Scottmitch, line 136: @idelpivnitskiy - inserting retries kinda goes against the spirit of what we originally wanted to verify, but I don't think it can be done reliably with out a retry in place. wdyt?

idelpivnitskiy: Interesting, thanks for looking at it. Agreed, the behavior can be racy and it's expected that we may race between connection selection and closure on the IO thread. We can keep as you proposed, or sequence requests without the race, waiting for the ctx1 to be closed:
```patch
Index: servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MalformedDataAfterHttpMessageTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MalformedDataAfterHttpMessageTest.java b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MalformedDataAfterHttpMessageTest.java
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MalformedDataAfterHttpMessageTest.java	(revision 58cb933d24f2c71d6cad99bb091b89fb80364f84)
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MalformedDataAfterHttpMessageTest.java	(date 1635803916686)
@@ -26,7 +26,6 @@
 import io.servicetalk.http.api.StreamingHttpConnectionFilter;
 import io.servicetalk.http.api.StreamingHttpRequest;
 import io.servicetalk.http.api.StreamingHttpResponse;
-import io.servicetalk.http.utils.RetryingHttpRequesterFilter;
 import io.servicetalk.transport.api.ConnectionContext;
 import io.servicetalk.transport.api.HostAndPort;
 import io.servicetalk.transport.api.ServerContext;
@@ -45,8 +44,6 @@
 import io.netty.handler.codec.http.HttpRequestDecoder;
 import org.junit.jupiter.api.extension.RegisterExtension;
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.ValueSource;
 
 import java.io.IOException;
@@ -60,7 +57,6 @@
 import static io.netty.util.ReferenceCountUtil.release;
 import static io.servicetalk.buffer.api.Matchers.contentEqualTo;
 import static io.servicetalk.http.api.HttpExecutionStrategies.defaultStrategy;
-import static io.servicetalk.http.api.HttpExecutionStrategies.noOffloadsStrategy;
 import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_LENGTH;
 import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_TYPE;
 import static io.servicetalk.http.api.HttpHeaderValues.TEXT_PLAIN;
@@ -75,7 +71,6 @@
 import static io.servicetalk.transport.netty.internal.EventLoopAwareNettyIoExecutors.toEventLoopAwareNettyIoExecutor;
 import static java.lang.Integer.MAX_VALUE;
 import static java.lang.String.valueOf;
-import static java.time.Duration.ofNanos;
 import static org.hamcrest.Matchers.anyOf;
 import static org.hamcrest.Matchers.empty;
@@ -123,20 +118,11 @@
         }
     }
 
-    @ParameterizedTest
-    @ValueSource(booleans = {true, false})
-    void afterResponseNextClientRequestSucceeds(boolean doOffloading) throws Exception {
+    @Test
+    void afterResponseNextClientRequestSucceeds() throws Exception {
         Queue<ConnectionContext> contextQueue = new ArrayBlockingQueue<>(4);
         ServerSocketChannel server = nettyServer(RESPONSE_MSG);
         try (BlockingHttpClient client = stClientBuilder(server.localAddress())
-                .executionStrategy(doOffloading ? defaultStrategy() : noOffloadsStrategy())
-                // ClosedChannelException maybe observed on the second request if write is done before read of the
-                // garbage data, which won't be a RetryableExcepton. We may also see an exception from flush if the read
-                // closed the connection and then attempt to write on the same connection.
-                .appendClientFilter(new RetryingHttpRequesterFilter.Builder()
-                        .maxRetries(MAX_VALUE)
-                        .retryFor((req, cause) -> true)
-                        .buildWithConstantBackoffFullJitter(ofNanos(1)))
                 .appendConnectionFilter(connection -> new StreamingHttpConnectionFilter(connection) {
                     @Override
                     public Single<StreamingHttpResponse> request(final HttpExecutionStrategy strategy,
@@ -147,19 +133,15 @@
                 })
                 .buildBlocking()) {
             validateClientResponse(client.request(client.get("/1")));
-            validateClientResponse(client.request(client.get("/2")));
-
             ConnectionContext ctx1 = contextQueue.poll();
             assertThat(ctx1, not(nullValue()));
-            // RetryingHttpRequesterFilter or AutoRetry may re-issue the request if a failure is seen locally. Verify
-            // the last connection (used for second request) is different from the first.
-            ConnectionContext ctx2 = null;
-            ConnectionContext tmp;
-            while ((tmp = contextQueue.poll()) != null) {
-                ctx2 = tmp;
-            }
+            ctx1.onClose().toFuture().get();
+
+            validateClientResponse(client.request(client.get("/2")));
+            ConnectionContext ctx2 = contextQueue.poll();
             assertThat(ctx2, not(nullValue()));
-            assertThat(ctx1, not(equalTo(ctx2)));
+            assertThat(ctx2, not(equalTo(ctx1)));
+            ctx2.onClose().toFuture().get();
             assertThat(contextQueue, empty());
         } finally {
             server.close().sync();
```

idelpivnitskiy (approved):

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MalformedDataAfterHttpMessageTest.java
Scottmitch: the read may not happen until the next request is issued so I'll stick with the retry approach for now.

PR #1928, opened 2021-11-02
Make static final buffers unreleasable and read-only

Scottmitch:
Motivation:
`HttpObjectEncoder` writes static final ByteBufs to encode HTTP chunks.
Some handlers (e.g. SslHandler) may attempt to aggregate data into
existing buffers if they can be expanded. If static buffers are not
marked as ReadOnly then data maybe appended to these static buffers over
time which may lead to data corruption.

Modification:
- Make static ByteBuf/Buffer objects ReadOnly in `HttpObjectEncoder` and
  elsewhere.

Result:
Static buffers will not be used to aggregate data.


normanmaurer (approved):

Scottmitch:
0.41 (70ce4bb22)

PR #1930, opened 2021-11-02
Remove `grpc` and `http` modules from BOM

bondolo:
Motivation:
The ServiceTalk BOM file currently contains references to two modules,
`grpc` and `http` which do not exist. These modules are generated from
the examples parent directories but have no artifacts.
Modifications:
Exclude the `grpc` and `http` examples parent modules when building
BOM.
Result:
Non-existant modules no longer included in BOM.

### DIFF
--- a/servicetalk-bom/build.gradle
+++ b/servicetalk-bom/build.gradle
@@ -17,7 +17,8 @@
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-core"
 apply plugin: "java-platform"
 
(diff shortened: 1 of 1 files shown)


bondolo:
Minor, but this one has been bothering me since working on the `japicmp.sh` script.

idelpivnitskiy (approved):
Interesting, didn't know those folders are considered as subprojects. I though only what is listed in `settings.gradle` is a subproject. Good catch!

bondolo:
I did not know where they were coming from either! It took a while to figure out what was causing them to be generated.

bondolo:
**0.41** → 6433e3227ae694a08acc938962c8264148917a01

PR #1934, opened 2021-11-03
`RoundRobinLoadBalancerFactory#EAGER_CONNECTION_SHUTDOWN_ENABLED` false by default

chemicL:
Motivation:

`RoundRobinLoadBalancer` can maintain connections to hosts which
`ServiceDiscoverer` marks as `UNAVAILABLE` via a boolean flag. This
improves overall performance by reusing existing connections if the
backend is still available. We should make it the default behaviour.

Modifications:

- `RoundRobinLoadBalancerFactory#EAGER_CONNECTION_SHUTDOWN_ENABLED` is
  `false` by default.

Result:

The default behaviour of `RoundRobinLoadBalancer` will provide better
connection reuse for most cases.


idelpivnitskiy (approved):
🚢

PR #1935, opened 2021-11-03
`HttpLoadBalancerFactory` should override `newLoadBalancer` with 3 args

idelpivnitskiy:
Motivation:

#1831 added a new
`LoadBalancerFactory#newLoadBalancer(String, Publisher, ConnectionFactory)`
overload, but didn't override it in `HttpLoadBalancerFactory`.

Modifications:

- Add `HttpLoadBalancerFactory.newLoadBalancer(...)` overload with 3
arguments;

Result:

`HttpLoadBalancerFactory` follows `LoadBalancerFactory` contract with
http-specific types.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
@@ -21,6 +21,10 @@ import io.servicetalk.client.api.LoadBalancerFactory;
 import io.servicetalk.client.api.ServiceDiscovererEvent;
 import io.servicetalk.concurrent.api.Publisher;
 
+import java.util.Collection;
+
+import static java.util.function.Function.identity;
+
 /**
  * A {@link LoadBalancerFactory} for HTTP clients.
  *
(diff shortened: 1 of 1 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
```diff
@@ -29,11 +33,20 @@
 public interface HttpLoadBalancerFactory<ResolvedAddress>
         extends LoadBalancerFactory<ResolvedAddress, FilterableStreamingHttpLoadBalancedConnection> {
 
+    @Deprecated
```
chemicL, line 36: I realized it's impossible to get rid of the deprecated implementation. We should add default implementations to base interface and this one too to be able to remove the deprecated variant.

chemicL (changes requested):

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
idelpivnitskiy: We discussed this when you originally introduced the change in https://github.com/apple/servicetalk/pull/1831#discussion_r712703689
Because the parent interface implements `@FunctionalInterface`, we can not make both methods `default`. We will make swap `default` impl in 0.42 (I will submit a follow-up PR, because this PR is going to 0.41 as well), and remove the old method in 0.43.

chemicL: Ah, right. Thanks for reminding me.

chemicL (approved):

PR #1936, opened 2021-11-04
`SingleThreadedScheduler` should fallback execution upon offloading failure

chemicL:
Motivation:

When the `Executor` provided by the user fails to accept the task
executed by the `SingleThreadedScheduler`, the exception is ignored and
the scheduler is no longer usable. The exceptions from the user land
should not surface to the global scheduler and the implementation should
do its' best to run the task even at the cost of skipping offloading.

Modification:

`DefaultExecutor.SingleThreadedScheduler#schedule` is guarded against
rejections from the user's `offloadExecutor` and triest to fallback
and execute the task on its' own thread.

Result:

User error does not prevent the `SingleThreadedScheduler` from executing
future tasks.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
@@ -258,8 +258,9 @@ final class DefaultExecutor extends AbstractExecutor implements Consumer<Runnabl
                         try {
                             offloadExecutor.execute(task);
                         } catch (Throwable t) {
-                            LOGGER.error("Unexpected exception while offloading scheduled task: {} to executor: {}.",
-                                    task, offloadExecutor, t);
+                            LOGGER.error("Unexpected exception while offloading scheduled task: {} to executor: {}. " +
+                                            "Will execute the task on {} thread.",
+                                    task, offloadExecutor, Thread.currentThread().getName(), t);
                             // If the offloadExecutor failed to execute, we fallback to the scheduler thread for
                             // best-effort attempt at actually executing the task.
                             try {
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
@@ -22,6 +22,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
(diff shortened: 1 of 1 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
```diff
@@ -249,11 +254,27 @@ public Cancellable schedule(final Runnable task, final long delay, final TimeUni
...
-            // Schedulers are only used to generate a tick and do not execute any user code. This means they will never
-            // run any blocking code and hence it does not matter whether we use the interruptOnCancel as sent by the
-            // user upon creation in the scheduler. User code (completion of Completable on tick) will be executed on
-            // the configured executor and not the Scheduler thread.
+                    () -> {
+                        try {
+                            offloadExecutor.execute(task);
+                        } catch (Throwable t) {
+                            LOGGER.error("Unexpected exception while offloading scheduled task: {} to executor: {}.",
+                                    task, offloadExecutor, t);
```
idelpivnitskiy, line 262:
```suggestion
                            LOGGER.error("Unexpected exception while offloading scheduled task: {} to executor: {}. " +
                                            "Will execute the task on {} thread.",
                                    task, offloadExecutor, Thread.currentThread().getName(), t);
```

WDYT? You can commit from the UI, I verified locally.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
```diff
@@ -249,11 +254,27 @@ public Cancellable schedule(final Runnable task, final long delay, final TimeUni
...
+                            try {
+                                task.run();
+                            } catch (Throwable taskFailure) {
+                                LOGGER.error("Scheduled task {} threw exception during fallback to scheduler executor.",
+                                        task, taskFailure);
+                            }
+                        }
+                    }, delay, unit);
+            // Schedulers are only used to generate a tick and should not execute any user code (unless the
+            // offloadExecutor throws). This means they will never run any blocking code and hence it does not matter
```
idelpivnitskiy, line 274: Yes, this is not ideal that we can still run a blocking code on this thread, but should be ok as a workaround for now. In future, we probably need to expend `schedule` API to also take a consumer of an error.

idelpivnitskiy (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
```diff
@@ -249,11 +254,27 @@ public Cancellable schedule(final Runnable task, final long delay, final TimeUni
...
             ScheduledFuture<?> future = GLOBAL_SINGLE_THREADED_SCHEDULED_EXECUTOR.schedule(
-                    () -> offloadExecutor.execute(task), delay, unit);
-            // Schedulers are only used to generate a tick and do not execute any user code. This means they will never
-            // run any blocking code and hence it does not matter whether we use the interruptOnCancel as sent by the
-            // user upon creation in the scheduler. User code (completion of Completable on tick) will be executed on
-            // the configured executor and not the Scheduler thread.
+                    () -> {
+                        try {
+                            offloadExecutor.execute(task);
+                        } catch (Throwable t) {
```
bondolo, line 260: I wonder if the attempt to re-execute the task should only happen for `RejectedExecutionException` (ignoring `NullPointerException` for null task) as all other cases should be presumed to be caused by the task and re-running it would either cause the same exception or other problems.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
```diff
@@ -249,11 +254,27 @@ public Cancellable schedule(final Runnable task, final long delay, final TimeUni
             // When using the global scheduler, offload timer ticks to the user specified Executor since user code
             // executed on the timer tick can block.
             ScheduledFuture<?> future = GLOBAL_SINGLE_THREADED_SCHEDULED_EXECUTOR.schedule(
```
bondolo, line 256: if `offloadExecutor` is a `ScheduledExecutorService` instance then perhaps schedule with it directly?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
idelpivnitskiy: We should still log any `Throwable`. Otherwise, it's a hidden error with zero visibility. 
But I'm agreed, that we should run the `Runnable` only for `RejectedExecutionException`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
bondolo: Fair enough. This class appears to be used only for non-`ScheduledExecutorServices` instances anyway as elsewhere in `Executors` the `from(ScheduledExecutorService scheduled)` method handles executors that support scheduling.

bondolo (approved):

PR #1937, opened 2021-11-04
`RoundRobinLoadBalancer` default `Executor` with unbounded queue

chemicL:
Motivation:

The default `Executor` used by `RoundRobinLoadBalancer` for health
checking uses a single `Thread` combined with a `SynchronousQueue`.
If multiple health checks are executed simultaneously, that can lead to
task rejections and health checks failing. The proper implementation can
still use a single thread, but should queue scheduled tasks.

Modification:

Default `RoundRobinLoadBalancerFactory.SharedExecutor` uses
a single-threaded, unbounded `ThreadPoolExecutor` to schedule health checks.

Result:

Health checks can be safely rescheduled using the default executor used
by `RoundRobinLoadBalancer`.


idelpivnitskiy (approved):

bondolo (approved):

idelpivnitskiy:
Verified locally, it fixes the issue. We will add tests in a follow-up

PR #1938, opened 2021-11-04
RoundRobinLoadBalancer: reduce number of operators in health check chain

idelpivnitskiy:
Motivation:

`Completable.retryWhen` actually does
`toSingle().retryWhen(...).ignoreElement()`. We can avoid it by retrying
the `Single` directly.
We can also skip `onErrorMap` and log inside a retry predicate.

Modifications:

- Avoid premature conversion of a `Single` to `Completable`;
- Use retry predicate instead of `onErrorMap` to log an attempt failure;
- Use the original cause instead of `RESCHEDULE_SIGNAL`;

Result:

Less operators in the health check execution chain.


bondolo (approved):

chemicL:
Late LGTM! 👍

PR #1940, opened 2021-11-04
`MessageEncodingException` should take an encoding name only

idelpivnitskiy:
Motivation:

`MessageEncodingException` expects an encoding name only, not a cause.
It already has a prepared cause inside a ctor.

Modifications:

- Pass encoding to the `MessageEncodingException` ctor instead of a
cause;

Result:

Users see a correct cause message for `MessageEncodingException`.

### DIFF
--- a/servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProviderBuilder.java
+++ b/servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProviderBuilder.java
@@ -172,7 +172,7 @@ public final class ProtoBufSerializationProviderBuilder {
             @SuppressWarnings("unchecked")
             HttpSerializer<T> httpSerializer = serializersForType.get(codec);
             if (httpSerializer == null) {
-                throw new MessageEncodingException("Unknown encoding: " + codec.name());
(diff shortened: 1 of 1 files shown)


tkountis (approved):

servicetalk-grpc-protobuf/src/main/java/io/servicetalk/grpc/protobuf/ProtoBufSerializationProviderBuilder.java
```diff
@@ -172,7 +172,7 @@ public GrpcSerializationProvider build() {
             @SuppressWarnings("unchecked")
             HttpSerializer<T> httpSerializer = serializersForType.get(codec);
             if (httpSerializer == null) {
-                throw new MessageEncodingException("Unknown encoding: " + codec.name());
```
bondolo, line 175: Perhaps instead of the encoding name, which may be non-canonical, the exception could be created with the `ContentCodec` instance responsible.

idelpivnitskiy: This PR mostly targets 0.41 branch. In the main branch, old serialization is already deprecated. This exception type was missed, but will be deprecated by #1924. Will minimize changes for the code that is going away soon.

PR #1941, opened 2021-11-05
Completable#concat(Completable...)

Scottmitch:
Motivation:
There are some use cases (DefaultCompositeCloseable) that dynamically
build a large chain of Completable#concat(Completable) operations. These
large chains may result in a StackOverflowException.

Modifications:
- Introduce a `Completable#concat(Completable...)` operator that has stack
  overflow protection.

Result:
Completable concat operator exists with stack overflow protection.

PR #1942, opened 2021-11-05
DefaultCompositeCloseable StackOverflowException

Scottmitch:
Motivation:
DefaultCompositeCloseable uses operator composition when merging or
appending AsyncClosables. When a close operation is invoked on the
closable that means the stack must have enough depth in order to
cascade invocations of subscribe and signal notification. There are
scenarios where the number of operations is proportional to the number
of channels (ChannelSet, RRLB) which are more likely to result in
StackOverflowException as the number of channels grow.

Modifications:
- Update DefaultCompositeCloseable to avoid composing individual
  operators and instead use bulk array based operators.

Result:
Less likely to get SOOE from DefaultCompositeCloseable during
client/server shutdown.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
@@ -34,6 +34,10 @@ import static java.util.stream.StreamSupport.stream;
 final class DefaultCompositeCloseable implements CompositeCloseable {
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultCompositeCloseable.class);
 
+    /**
+     * Initial size of 2 because we don't expect a large cardinality of {@link Operand}s. Usually only 1 is present
+     * (e.g. either merge or concat).
+     */
     private final Deque<Operand> operands = new ArrayDeque<>(2);
     @Nullable
     private Completable closeAsync;
@@ -60,21 +64,21 @@ final class DefaultCompositeCloseable implements CompositeCloseable {
 
     @Override
     public <T extends AsyncCloseable> T append(final T closeable) {
-        concatCloseableDelayError(closeable);
+        appendCloseableDelayError(closeable);
         return closeable;
     }
 
     @Override
     public CompositeCloseable appendAll(final AsyncCloseable... asyncCloseables) {
         for (AsyncCloseable closeable : asyncCloseables) {
-            concatCloseableDelayError(closeable);
+            appendCloseableDelayError(closeable);
         }
         return this;
     }
 
     @Override
     public CompositeCloseable appendAll(final Iterable<? extends AsyncCloseable> asyncCloseables) {
-        asyncCloseables.forEach(this::concatCloseableDelayError);
+        asyncCloseables.forEach(this::appendCloseableDelayError);
         return this;
     }
 
@@ -160,7 +164,7 @@ final class DefaultCompositeCloseable implements CompositeCloseable {
             } else {
                 operand = new Operand(isMerge);
                 if (append) {
-                    operands.add(operand);
+                    operands.addLast(operand);
                 } else {
                     operands.addFirst(operand);
                 }
@@ -169,7 +173,7 @@ final class DefaultCompositeCloseable implements CompositeCloseable {
         return operand;
     }
 
-    private void concatCloseableDelayError(final AsyncCloseable closeable) {
+    private void appendCloseableDelayError(final AsyncCloseable closeable) {
         final Operand operand = getOrAddConcatOperand(true);
         operand.closables.add(closeable);
         resetState();
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
@@ -18,7 +18,12 @@ package io.servicetalk.concurrent.api;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Deque;
 import java.util.List;
+import java.util.function.Function;
+import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.api.Completable.completed;
 import static io.servicetalk.concurrent.internal.FutureUtils.awaitTermination;
@@ -27,11 +32,13 @@ import static java.util.stream.Collectors.toList;
 import static java.util.stream.StreamSupport.stream;
 
 final class DefaultCompositeCloseable implements CompositeCloseable {
-
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultCompositeCloseable.class);
 
-    private Completable closeAsync = completed();
-    private Completable closeAsyncGracefully = completed();
+    private final Deque<Operand> operands = new ArrayDeque<>(2);
+    @Nullable
+    private Completable closeAsync;
+    @Nullable
+    private Completable closeAsyncGracefully;
 
     @Override
     public <T extends AsyncCloseable> T merge(final T closeable) {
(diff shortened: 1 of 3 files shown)


Scottmitch:
depends upon PR https://github.com/apple/servicetalk/pull/1941

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
```diff
@@ -112,39 +125,85 @@ public void closeGracefully() {
...
+                    operands.add(operand);
+                } else {
+                    operands.addFirst(operand);
+                }
+            }
+        }
+        return operand;
     }
 
     private void concatCloseableDelayError(final AsyncCloseable closeable) {
```
idelpivnitskiy, line 172: While you are here, consider renaming `concatCloseableDelayError` -> `appendCloseableDelayError`. We used `concat` before we changed it to append/prepend.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
```diff
@@ -112,39 +125,85 @@ public void closeGracefully() {
...
+        final Operand operand = getOrAddConcatOperand(false);
+        operand.closables.add(closeable);
+        resetState();
+    }
+
+    private Completable buildCompletable(Function<AsyncCloseable, Completable> closerFunc) {
+        Completable result = completed();
+        for (Operand operand : operands) {
+            if (operand.isMerge) {
+                result = result.mergeDelayError(operand.closables.stream().map(closerFunc).toArray(Completable[]::new));
```
idelpivnitskiy, line 188: Any pros/cons of using `completed().mergeDelayError(array)` vs static `mergeAllDelayError(array)`? If there are any, have you considered to add a static `concatAll(array)` utility instead of `Completable.concat(array)`?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
```diff
@@ -112,39 +125,85 @@ public void closeGracefully() {
...
+            } else {
+                result = result.concat(operand.closables.stream().map(closerFunc).map(completable ->
+                        completable.onErrorComplete(th -> {
+                            // TODO: This should use concatDelayError when available.
+                            LOGGER.debug("Ignored failure to close", th);
+                            return true;
```
idelpivnitskiy, line 194: That was pre-existing, but looks like we hide all errors instead of delaying them. Should we try to save them in `AsyncContext` and verify its key at the end, as the last completable?

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/CompositeClosableTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import static io.servicetalk.concurrent.api.AsyncCloseables.newCompositeCloseable;
+import static io.servicetalk.concurrent.api.Completable.completed;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+final class CompositeClosableTest {
+    @ParameterizedTest
+    @CsvSource(value = {"true,true", "true,false", "false, true", "false,false"})
```
idelpivnitskiy, line 30: nit: extra space in 3rd case

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/CompositeClosableTest.java
```diff
@@ -0,0 +1,78 @@
...
+    void sameOperationDoesNotSOOE(boolean merge, boolean gracefully) throws Exception {
```
idelpivnitskiy, line 31: SOOE -> SOE?

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/CompositeClosableTest.java
```diff
@@ -0,0 +1,78 @@
...
+        if (gracefully) {
+            compositeCloseable.closeGracefully();
+        } else {
+            compositeCloseable.close();
+        }
+    }
+
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void alternatingOperationSOOE(boolean gracefully) {
```
idelpivnitskiy, line 54: SOOE -> SOE?

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/CompositeClosableTest.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        if (gracefully) {
+            compositeCloseable.closeGracefully();
+        } else {
+            compositeCloseable.close();
+        }
+    }
+
+    @ParameterizedTest
```
idelpivnitskiy, line 52: Consider adding `name` format

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
```diff
@@ -27,11 +32,13 @@
 import static java.util.stream.StreamSupport.stream;
 
 final class DefaultCompositeCloseable implements CompositeCloseable {
-
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultCompositeCloseable.class);
 
-    private Completable closeAsync = completed();
-    private Completable closeAsyncGracefully = completed();
+    private final Deque<Operand> operands = new ArrayDeque<>(2);
```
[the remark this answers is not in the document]
Scottmitch: `ArrayDeque` allows for smaller arrays, default size of the underlying array will be `3`. I also don't think we need to couple internal behavior of `ArrayDeque` with here, basically we don't expect a large cardinality of mixed usage so we start with a small number (I will add a comment).

```java
        elements =
            new Object[(numElements < 1) ? 1 :
                       (numElements == Integer.MAX_VALUE) ? Integer.MAX_VALUE :
                       numElements + 1];
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
Scottmitch: `concatAll` may have a small advantage if you just have an array bcz no need to bootstrap with `completed()` (initialization of subscribe a bit more streamlined) however the control flow here mixes the two operations so it makes it cleaner to just compose each operator individually.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
Scottmitch: Lets defer for `concatDelayError` in a followup PR when it is available.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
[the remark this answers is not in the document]
idelpivnitskiy: Where is this code from? `ArrayDeque` has `MIN_INITIAL_CAPACITY = 8` internally and discards any values less than that.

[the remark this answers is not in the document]
Scottmitch: [ArrayDeque JDK11](https://github.com/openjdk/jdk11u-dev/blob/master/src/java.base/share/classes/java/util/ArrayDeque.java#L186-L197) and [ArrayDeque JDK17](https://github.com/openjdk/jdk17u/blob/master/src/java.base/share/classes/java/util/ArrayDeque.java#L185-L196)

Scottmitch:
0.41 (https://github.com/apple/servicetalk/pull/1946)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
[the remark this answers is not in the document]
idelpivnitskiy: Ok, I'm still on 8 😛

PR #1944, opened 2021-11-05
Adjust changes to http-api related to `ExecutionStrategyInfluencer`

idelpivnitskiy:
Motivation:

1. There is no much value in making `HttpFilterFactory` public,
hard to think of a use-case for users.
2. Inconsistencies in default `requiredOffloads()` implementations.
3. Incorrect `requiredOffloads()` impl for some service API.
4. No need to implement `ExecutionStrategyInfluencer` for `HttpClient`,
it already has a strategy in `HttpExecutionContext` associated with the
client. No other client API implement `ExecutionStrategyInfluencer`.

Modifications:

- Rename `HttpFilterFactory` to `DefaultHttpExecutionStrategyInfluencer`
and make it pkg-private;
- Make all default `requiredOffloads()` implementations consistent;
- Use `HttpExecutionStrategies.offloadAll()` instead of a pkg-private
`OFFLOAD_ALL_STRATEGY` constant because users will read the default
impl, better to have something that has javadoc;
- `BlockingHttpService#requiredOffloads()` have to be
`OFFLOAD_RECEIVE_DATA_STRATEGY` as it's defined in
`BlockingToStreamingService`;
- `BlockingStreamingHttpService#requiredOffloads()` have to be
`OFFLOAD_RECEIVE_META_STRATEGY` as it's defined in
`BlockingStreamingToStreamingService`;
- `StreamingHttpServiceFilterFactory` should not override the default
`offloadAll()` strategy;
- Remove `ExecutionStrategyInfluencer` interface from `HttpClient`;

Result:

Less public API, more consistency, correct default implementations.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingHttpService.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingHttpService.java
@@ -19,7 +19,7 @@ import io.servicetalk.concurrent.GracefulAutoCloseable;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.transport.api.ExecutionStrategyInfluencer;
 
-import static io.servicetalk.http.api.DefaultHttpExecutionStrategy.OFFLOAD_NONE_STRATEGY;
+import static io.servicetalk.http.api.DefaultHttpExecutionStrategy.OFFLOAD_RECEIVE_DATA_STRATEGY;
 
 /**
  * The equivalent of {@link HttpService} but with synchronous/blocking APIs instead of asynchronous APIs.
@@ -40,7 +40,8 @@ public interface BlockingHttpService extends ExecutionStrategyInfluencer<HttpExe
 
     @Override
     default HttpExecutionStrategy requiredOffloads() {
-        return OFFLOAD_NONE_STRATEGY;
+        // safe default--implementations are expected to override
+        return OFFLOAD_RECEIVE_DATA_STRATEGY;
     }
 
     @Override
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategyInfluencer.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategyInfluencer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018-2019, 2021 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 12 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingHttpService.java
```diff
@@ -40,7 +40,8 @@ HttpResponse handle(HttpServiceContext ctx, HttpRequest request, HttpResponseFac
 
     @Override
     default HttpExecutionStrategy requiredOffloads() {
-        return OFFLOAD_NONE_STRATEGY;
+        // safe default--implementations are expected to override
+        return OFFLOAD_RECEIVE_DATA_STRATEGY;
```
idelpivnitskiy, line 44: @bondolo I'm surprised that there are no tests that could catch cases like this one. Can you please have a look how this value is used? Is it used for computation of the final strategy?

bondolo: Indeed, I would have expected that the existing tests would have caught this.

bondolo: It is referenced by the some of the Jersey tests but seems to have no effect. I will continue to investigate why.

idelpivnitskiy: We can work on tests in a follow-up. I would like to pull in changes for interfaces, bcz I'm working on filters API

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionStrategyInfluencer.java
```diff
@@ -17,16 +17,11 @@
 
 import io.servicetalk.transport.api.ExecutionStrategyInfluencer;
 
-import static io.servicetalk.http.api.DefaultHttpExecutionStrategy.OFFLOAD_ALL_STRATEGY;
-
-/**
- * A factory for HTTP filters.
- */
-public interface HttpFilterFactory extends ExecutionStrategyInfluencer<HttpExecutionStrategy> {
+interface DefaultHttpExecutionStrategyInfluencer extends ExecutionStrategyInfluencer<HttpExecutionStrategy> {
```
bondolo, line 20: Many of the classes that implement this interface will be public and any similar user classes will also want to use this interface. A similar example would be the Java `StringBuilder` and `StringBuffer` classes which extend the non-public `AbstractStringBuilder` class which receives constant complaints.

In general, public classes and interfaces should have only public ancestors.

I am also not convinced by the name change. It seems too broad.

bondolo: On further thought it seems reasonable as long as users won't need to recreate the same interface themselves.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpClient.java
```diff
@@ -17,17 +17,14 @@
...
 import static io.servicetalk.concurrent.internal.FutureUtils.awaitTermination;
-import static io.servicetalk.http.api.DefaultHttpExecutionStrategy.OFFLOAD_ALL_STRATEGY;
 
 /**
  * Provides a means to issue requests against HTTP service. The implementation is free to maintain a collection of
  * {@link HttpConnection} instances and distribute calls to {@link #request(HttpRequest)} amongst this collection.
  */
-public interface HttpClient extends
-                            HttpRequester, GracefulAutoCloseable, ExecutionStrategyInfluencer<HttpExecutionStrategy> {
+public interface HttpClient extends HttpRequester, GracefulAutoCloseable {
```
bondolo, line 27: I will need to check if this is getting used. Removal may change offloading.

bondolo: It is never used. Remove it.

bondolo (approved):

PR #1945, opened 2021-11-06
Completable#concat(Completable...)

Scottmitch:
Motivation:
There are some use cases (DefaultCompositeCloseable) that dynamically
build a large chain of Completable#concat(Completable) operations. These
large chains main result in a StackOverflowException.

Modifications:
- Introduce a Completable#concat(Completable...) operator that has stack
  overflow protection.

Result:
Completable concat operator exists with stack overflow protection.


Scottmitch:
Merging bcz https://github.com/apple/servicetalk/pull/1941 was approved and this is just a back port.

PR #1946, opened 2021-11-06
DefaultCompositeCloseable StackOverflowException

Scottmitch:
Motivation:
DefaultCompositeCloseable uses operator composition when merging or
appending AsyncClosables. When a close operation is invoked on the
closable that means the stack must have enough depth in order to
cascade invocations of subscribe and signal notification. There are
scenarios where the number of operations is proportional to the number
of channels (ChannelSet, RRLB) which are more likely to result in
StackOverflowException as the number of channels grow.

Modifications:
- Update DefaultCompositeCloseable to avoid composing individual
  operators and instead use bulk array based operators.

Result:
Less likely to get SOOE from DefaultCompositeCloseable during
client/server shutdown.


Scottmitch:
back port of https://github.com/apple/servicetalk/pull/1942

PR #1948, opened 2021-11-09
Configurable offloading for async close

bondolo:
Motivation:
Currently publishing of terminal signals for the `Completable` returned by `AsyncClose` methods and `onClose` methods is always offloaded. For some cases this adds latency or unnecessary offloading.
Modifications:
`HttpServerBuilder` and `SingleAddressClientBuilder` are extended to offer configuration of async close offloading. The behaviour can be configured specifically or automatically as a result of specifying  the `noOffloadsStrategy` execution strategy.
Result:
Offloading of async close completion is configurable for use cases that do not require offloading.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -250,8 +251,14 @@ public StreamingHttpClient buildStreaming() {
     }
 
     private static <U, R> StreamingHttpClient buildStreaming(final HttpClientBuildContext<U, R> ctx) {
-        final ReadOnlyHttpClientConfig roConfig = ctx.httpConfig().asReadOnly();
         final HttpExecutionContext executionContext = ctx.builder.executionContextBuilder.build();
+        // XXX This should probably be somewhere else
```
bondolo, line 255: Suggestions welcome as to how this should be handled. It seems that copying the tcp config, applying the configuration adjustment and then building the read-only config would be the most user-friendly approach.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ParentConnectionContext.java
```diff
@@ -97,6 +100,10 @@ public FlushStrategy defaultFlushStrategy() {
         return flushStrategyHolder.currentStrategy();
     }
 
+    public boolean isAsyncCloseOffload() {
```
Scottmitch, line 103: package private?

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
```diff
@@ -154,7 +156,7 @@ protected void handleSubscribe(Subscriber<? super ServerContext> subscriber) {
                     Throwable cause = f.cause();
                     if (cause == null) {
                         subscriber.onSuccess(NettyServerContext.wrap(channel, channelSet,
-                                connectionAcceptor, executionContext));
+                                connectionAcceptor, executionContext, config.isAsyncCloseOffloaded()));
```
Scottmitch, line 159: just thinking out loud to make sure I follow  ... having a single flag will make it so offloading is excessive for consumers of `ServerContext` (e.g. only need to offload the `ChannelSet`), however this approach is "safe" bcz a service could close the connection via the `HttpServiceContext`. If we wanted to we could add another flag in the future so users can offload just the `ChannelSet` in a non-breaking fashion.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -250,8 +251,14 @@ public StreamingHttpClient buildStreaming() {
...
+        if (noOffloadsStrategy() == executionContext.executionStrategy()) {
```
Scottmitch, line 256: I was originally assuming `HttpExecutionStrategy` would have something like `isCloseOffloaded`, which would solve this issue but I think you mentioned this would complicate merging/combination. What is the expected behavior if the user says `noOffloadsStrategy()` but offloads `asyncCloseOffload`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -322,35 +308,59 @@ private HttpExecutionContext buildExecutionContext(final HttpExecutionStrategy s
...
                 // All the filters can be appended.
                 nonOffloadingFilters = Stream.concat(nonOffloadingFilters, serviceFilters.stream());
-                serviceContext = buildExecutionContext(strategy);
+                executionContext = buildExecutionContext(strategy);
             }
             filteredService = buildService(nonOffloadingFilters, rawService);
         }
 
-        return doListen(connectionAcceptor, serviceContext, filteredService, drainRequestPayloadBody)
+        if (noOffloadsStrategy() == strategy) {
```
Scottmitch, line 334: reminder: pending discussion in `DefaultSingleAddressHttpClientBuilder` we should be consistent here.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
idelpivnitskiy: We had an offline discussion with Mike about having this as a builder option or promoting to `ExecutionStrategy` interface as an additional flag. Rational was that almost everything extends `AsyncCloseable` interface, including `ConnectionFactory`, `ConnectionAcceptor`, HTTP filters, etc. All may apply blocking code for `closeAsync` or `onClose` processing and should be able to influence its behavior. It's also useful for setting as an `HttpExecutionStrategy` at the builder level, because `onClose()` can be used outside of the filters, and will eliminate question "What is the expected behavior if the user says `noOffloadsStrategy()` but offloads asyncCloseOffload?".

@bondolo have you had a chance to consider this approach?

bondolo: I think it is worth considering as control of `close()` offloading is a common enough behaviour. I haven't had time to try implementing it as an `ExecutionStrategy` flag.

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
```diff
@@ -319,6 +321,11 @@ public HttpServerBuilder allowDropRequestTrailers(final boolean allowDrop) {
             return this;
         }
 
+        @Override
+        public HttpServerBuilder asyncCloseOffload(final boolean offload) {
+            return delegate.asyncCloseOffload(offload);
+        }
```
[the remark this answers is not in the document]
bondolo: Oops. Thank you.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
Scottmitch: any methods on ExecutionStrategy require thought bcz it is so general. even for things which may be on lots of APIs (like close) it may not always apply in the expected context (e.g. service filters are closable but don't need to be offloaded IIUC bcz they aren't invoked from the event loop but are driven by user code). If possible it would be nice to have a targeted approach on APIs (e.g. extending an influencer allows the user to control only what needs to be offloaded), however maybe we can make exceptions if this would overly complicate APIs/usage.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ExecutionStrategy.java
```diff
@@ -31,6 +31,15 @@
      */
     boolean hasOffloads();
 
+    /**
+     * Returns {@code true} if the instance has offloading for any operation.
```
chemicL, line 35: Looks like copy-paste. Please adjust the javadoc.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -241,8 +242,16 @@ public StreamingHttpClient buildStreaming() {
     }
 
     private static <U, R> StreamingHttpClient buildStreaming(final HttpClientBuildContext<U, R> ctx) {
-        final ReadOnlyHttpClientConfig roConfig = ctx.httpConfig().asReadOnly();
         final HttpExecutionContext executionContext = ctx.builder.executionContextBuilder.build();
+        // XXX This modifies the builder state. Suggestions for allowing execution strategy to override but not
```
chemicL, line 246: Wouldn't noOffloadsStrategy dictate that the `asyncCloseOffload` setting is ineffective?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -163,11 +164,11 @@ private void cleanupOnWriteTerminated() {
 
     private DefaultNettyConnection(
             Channel channel, ExecutionContext<?> executionContext,
-            CloseHandler closeHandler, FlushStrategy flushStrategy,
+            CloseHandler closeHandler, FlushStrategy flushStrategy, final boolean offloadAsyncClose,
             @Nullable Long idleTimeoutMs, Protocol protocol, @Nullable SSLSession sslSession,
             @Nullable ChannelConfig parentChannelConfig, DataObserver dataObserver, boolean isClient,
             UnaryOperator<Throwable> enrichProtocolError) {
-        super(channel, executionContext.executor());
+        super(channel, offloadAsyncClose ? executionContext.executor() : immediate());
```
chemicL, line 171: What prevents the usage of `executionContext.executionStrategy()` and it's to-be-added new method: `asyncCloseOffloaded()`? I suppose that would simplify things greatly. Especially having in mind that `AsyncCloseable` is a core concept in `concurrent-api`.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ExecutionStrategy.java
idelpivnitskiy: +1, and add reference to `ListenableAsyncCloseable` methods to clarify what this flag contrils

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionContext.java
```diff
@@ -17,17 +17,14 @@
...
 /**
  * Default implementation of {@link HttpExecutionContext}.
  */
-public final class DefaultHttpExecutionContext implements HttpExecutionContext {
-
-    private final BufferAllocator bufferAllocator;
-    private final IoExecutor ioExecutor;
-    private final Executor executor;
-    private final HttpExecutionStrategy executionStrategy;
+public final class DefaultHttpExecutionContext extends DefaultExecutionContext<HttpExecutionStrategy>
```
idelpivnitskiy, line 26: Was this change made only to reduce duplication or for some other reason?

servicetalk-http-api/src/testFixtures/java/io/servicetalk/http/api/DelegatingHttpExecutionStrategy.java
```diff
@@ -33,6 +33,21 @@ public DelegatingHttpExecutionStrategy(final HttpExecutionStrategy delegate) {
...
     }
 
+    @Override
+    public boolean hasOffloads() {
+        return delegate.hasOffloads();
+    }
+
+    @Override
+    public boolean isRequestResponseOffloaded() {
+        return HttpExecutionStrategy.super.isRequestResponseOffloaded();
```
idelpivnitskiy, line 43: Should this method call `delegate`? If not, please leave a comment explaining why

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpOffloadingTest.java
```diff
@@ -232,50 +262,63 @@ void clientSettingsStreamIsOffloaded() throws Exception {
...
         terminated.await();
         assertNoAsyncErrors(errors);
     }
 
-    @Test
-    void clientOnCloseIsOffloaded() throws Exception {
+    @ParameterizedTest(name = "offloadClose={0}")
+    @ValueSource(booleans = {false, true})
+    void clientOnCloseIsOffloaded(boolean noOffloads) throws Exception {
+        setup(noOffloads);
```
idelpivnitskiy, line 286: Would be nice to add a test that verifies adding a filter or connection factory that influences the close adds offloading. Could be done in a follow-up, not necessary to block this PR

idelpivnitskiy (approved):
Looks great!

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/DefaultConnectExecutionStrategy.java
```diff
@@ -22,11 +22,30 @@
...
             return false;
         }
+
+        @Override
+        public boolean isCloseOffloaded() {
+            return false;
+        }
+    },
+    /**
+     * Offload all connect invocations
```
chemicL, line 37: the comment doesn't seem to be true

servicetalk-http-api/src/testFixtures/java/io/servicetalk/http/api/DelegatingHttpExecutionStrategy.java
bondolo: Appears to be a bad default method implementation, fixed.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpExecutionContext.java
bondolo: Yes, to reduce duplication. I'd actually prefer that `DefaultHttpExecutionContext` not even exist since it is just a non-generic version of `DefaultExecutionContext<HttpExecutionStrategy>`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
bondolo: yes, it was confusing. Hopefully better now with the new approach

tkountis (approved):
late lgtm too

PR #1951, opened 2021-11-10
Use connection execution context in proxy connect

bondolo:
Motivation:
Currently default execution strategy is always used for the connect
request but the connection/client may be using a different strategy.
Modifications:
Use the connection execution strategy so as not to introduce unexpected
offloading.
Result:
Proxy connect uses same strategy as the rest of the client operations.


idelpivnitskiy (approved):
🚢

PR #1954, opened 2021-11-12
Minor `ExecutionStrategy` cleanups

bondolo:
Motivation:
Several small issues with `ExecutionStrategy` were discovered.
Modifications:
1. Use default identity `equals()`` and `hashCode()` for
`ExecutionStrategy` instances. Document expected method behaviour.
2. Remove obsolete, temporary
`SpecialExecutionStrategy::asInfluencer()` and
`DefaultHttpExecutionStrategy::asInfluencer()`
3. Replace redundant class `NoopExecutionStrategy`
Result:
More consistent `ExecutionStrategy` usage.


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ExecutionStrategy.java
```diff
@@ -17,6 +17,9 @@
 
 /**
  * An execution strategy for all transports.
+ *
+ * <p>Implementations should not override the default {@link Object#equals(Object)} and {@link Object#hashCode()} method
+ * implementations. Default instance equality and hash-code behavior should be consistent across all instances.
```
idelpivnitskiy, line 22: IIUC, this is because we anyway treat some strategies as "special", right?
Is it worth, adding `missing` method for `ExecutionStrategy`? Even if it has one (or two) method-flags, it will be useful for users to get the difference between two strategies.

idelpivnitskiy (approved):

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ExecutionStrategy.java
bondolo: The main reason would be that there is no guarantee that all implementations provide consistent and reflexive `equals()` and `hashCode()`. Especially with the cases which use anonymous inner classes to override a single method to override behaviour. I started this patch with trying to fix the `equals()` method of `DelegatingHttpExecutionStrategy` thinking that it the implementation should just be `delegate.equals(obj)` but realized that contradicted how it was used and that only identity `equals()` made sense. 

We could try to define `equals()` and `hashCode()` to be consistent per sub-type, such as if the instance is an `HttpExecutionStrategy` the `equals()` must be defined as having the same offloads and `hashCode()` similarly defined. This still runs afoul of the "special" strategies where "no offloads" is different than "never offload" even though the flags are the same or for "offload all" having the same flags as "default offloads".

So the conclusion is that only identity `equals()` and `hashCode()` make sense and, as a "reward", we get to use `==` for comparing strategies.

PR #1955, opened 2021-11-15
Multi-address client: throw  `MalformedURLException` instead of `IllegalArgumentException`

idelpivnitskiy:
Motivation:

`MalformedURLException` is more suitable for a use-case when an URL is
not in absolute-form.

Modifications:

- Throw  `MalformedURLException` instead of `IllegalArgumentException`;
- Try-catch `selectClient` operation and return failed `Single` in case
of an error;

Result:

`StreamingUrlHttpClient` try-catches all exception during client
selection and throws `MalformedURLException` when a non-absolute form
URL is provided.


idelpivnitskiy:
This code was reviewed and approved in #1924.

PR #1958, opened 2021-11-18
Reduce SSLSession cache misses

Scottmitch:
Motivation:
The key for SSLSession is `peerHost+peerPort`. Our builders current
set the peerHost before resolution is complete. This may result in
cache misses if the hostname resolves to multiple IPs.

Modifications:
- The peerHost should be set after resolution, when the IP address
  is known. The cache key will then be specific to a host and more
  likely to have a cache hit. The peerHost is not authoritative so
  overriding should be safe.

Result:
SSLSession cache hits are more likely to happen in practice.

### DIFF
--- a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/ReadOnlyTcpClientConfig.java
+++ b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/ReadOnlyTcpClientConfig.java
@@ -16,6 +16,7 @@
 package io.servicetalk.tcp.netty.internal;
 
 import io.servicetalk.transport.api.ClientSslConfig;
+import io.servicetalk.transport.api.DelegatingClientSslConfig;
 
 import io.netty.handler.ssl.SslContext;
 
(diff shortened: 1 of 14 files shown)


servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/ReadOnlyTcpClientConfig.java
```diff
@@ -68,4 +80,58 @@ public String preferredAlpnProtocol() {
...
+            return peerHost;
+        }
+
+        @Override
+        public int peerPort() {
+            return peerPort;
+        }
+
+        @Override
+        public String sniHostname() {
```
idelpivnitskiy, line 128: Not related to this PR, but something to consider: is it worth changing a return type of this method to `javax.net.ssl.SNIHostName` instead of a `String`? It will eliminate a case when users passed a `String` that won't pass `SNIHostName` validation.

idelpivnitskiy (approved):

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/ReadOnlyTcpClientConfig.java
Scottmitch: good point. will consider for followup PR.

Scottmitch:
0.41 (back port PR https://github.com/apple/servicetalk/pull/1966)

PR #1962, opened 2021-11-19
Removing deprecated `ServiceDiscovererEvent` methods

chemicL:
Motivation:

Recently, `ServiceDiscovererEvent`'s (and consequently
`DnsServiceDiscovererObserver`'s) API has changed and the deprecations
can be removed in `main` branch.

Modifications:

Removed deprecated methods and reverted default implementations for new
methods.

Result:

Cleaner API with less deprecations.


idelpivnitskiy (approved):
`ServiceDiscovererUtils` is used only by `DefaultDnsClient`. Consider moving it to `servicetalk-dns-discovery-netty` module.

chemicL:
> `ServiceDiscovererUtils` is used only by `DefaultDnsClient`. Consider moving it to `servicetalk-dns-discovery-netty` module.

I moved it now.

PR #1963, opened 2021-11-19
Clarify `ServiceDiscovererEvent` contract related to address and status

idelpivnitskiy:
Motivation:

Add a note that clarifies how subsequent events are treated by SD
contract.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscovererEvent.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscovererEvent.java
@@ -27,6 +27,10 @@ import static io.servicetalk.client.api.ServiceDiscovererEvent.Status.UNAVAILABL
 public interface ServiceDiscovererEvent<ResolvedAddress> {
     /**
      * Get the resolved address which is the subject of this event.
+     * <p>
+     * Note: all subsequent events for the same address override its {@link #status()} or any additional meta-data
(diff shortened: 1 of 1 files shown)


servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscovererEvent.java
```diff
@@ -27,6 +27,10 @@
 public interface ServiceDiscovererEvent<ResolvedAddress> {
     /**
      * Get the resolved address which is the subject of this event.
+     * <p>
+     * Note: all subsequent events for the same address override its {@link #status()} or any additional meta-data
```
Scottmitch, line 31: maybe this type of comment is better placed on the `ServiceDiscoverer` where this constraint is expected to be honored (`ServiceDiscovererEvent` just carries the data)

Scottmitch (approved):

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscovererEvent.java
idelpivnitskiy: I don't see a good place in `ServiceDiscoverer` for this note without rewriting it to clarify all aspects of `ServiceDiscovererEvent`. I will leave it here for now and we can polish documentation as a separate effort across all modules.

PR #1966, opened 2021-11-19
Reduce SSLSession cache misses (#1958)

Scottmitch:
Motivation:
The key for SSLSession is `peerHost+peerPort`. Our builders current
set the peerHost before resolution is complete. This may result in
cache misses if the hostname resolves to multiple IPs.

Modifications:
- The peerHost should be set after resolution, when the IP address
  is known. The cache key will then be specific to a host and more
  likely to have a cache hit. The peerHost is not authoritative so
  overriding should be safe.

Result:
SSLSession cache hits are more likely to happen in practice.


Scottmitch:
back port of https://github.com/apple/servicetalk/pull/1958

PR #1969, opened 2021-11-29
Give visibility to the assigned HTTP/2 Stream ID

idelpivnitskiy:
Motivation:

`HttpServerContext` always has an HTTP/2 stream id as part of the
`ctx.toString()`. On the client-side, the identifier is assigned only
after the first write of `HEADERS` frame is made. We can give visibility
into the assigned stream id through the `StreamObserver`.

Modifications:

- Add `StreamObserver#streamIdAssigned(long)` callback;
- Invoke the new callback when the steam id is known;
- Test new behavior;

Result:

Users have visibility into the assigned HTTP/2 stream id and when it's
assigned.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
@@ -152,8 +152,6 @@ public interface ConnectionObserver {
          */
         default void streamIdAssigned(long streamId) {  // Use long to comply with HTTP/3 requirements
             // FIXME: 0.42 - remove default impl
-            throw new UnsupportedOperationException("Method streamIdAssigned(long) is not yet implemented by "
-                    + getClass().getName());
         }
 
         /**
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
@@ -156,7 +156,7 @@ public interface ConnectionObserver {
 
         /**
          * Callback when the stream is established and ready to be used. It may or may not have an already assigned
-         * {@code streamId} at this time.
+         * {@code streamId} at the time.
          *
          * @return a new {@link DataObserver} that provides visibility into read and write events
          * @see #streamIdAssigned(long)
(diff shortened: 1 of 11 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -144,9 +144,22 @@
...
 
         /**
-         * Callback when the stream is established and ready.
+         * Callback when a {@code streamId} is assigned.
+         * <p>
+         * Stream identifier may be deferred until after the first write is made on a newly established stream.
+         *
+         * @param streamId assigned stream identifier
+         */
+        default void streamIdAssigned(long streamId) {  // Use long to comply with HTTP/3 requirements
```
Scottmitch, line 153: for visibility and to confirm I understand the rational of this change ... if `StreamObserver` has access to some object representing the stream they could use `stream.getId()` instead of forcing the `long` parameter here, and they could also use this elsewhere. however the thought is since we don't have a good place to put such an abstraction on the base APIs enhancing the observer in this way is the lowest impact change?

idelpivnitskiy: > since we don't have a good place to put such an abstraction on the base APIs enhancing the observer in this way is the lowest impact change?

Correct.

The initial thought was to make `StreamObserver#streamEstablished()` similar to `ConnectionObserver#connectionEstablished(ConnectionInfo)` and give it some `StreamInfo` object with `#id()` method. However, on the client-side, the `streamEstablished()` is invoked before the first write => stream id is not assigned yet. We discussed this in the past and concluded that giving uses something like `StreamInfo#id()` that can return a different id if invoked before/after the first write is not convenient and complicates understanding.

Instead, this PR introduces a new callback, which is invoked exactly at the time when the id is assigned for a stream. It better represents what is happening at the protocol level.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -144,9 +144,22 @@
...
+            // FIXME: 0.42 - remove default impl
+        }
+
+        /**
+         * Callback when the stream is established and ready to be used. It may or may not have an already assigned
+         * {@code streamId} at this time.
```
tkountis, line 159: I believe that should be "at this time." -> "at the time"

tkountis (approved):

PR #1970, opened 2021-11-29
Correctly report TLS handshake when TCP Fast Open is enabled

idelpivnitskiy:
Motivation:

When TCP Fast Open socket option is enabled and available by the OS,
client sends `ClientHello` message with the initial TCP handshake. We
should trigger `onSecurityHandshake()` earlier in this case and give
users another callback to see when the TCP handshake completes.

Modifications:

- Add `ConnectionObserver#onTransportHandshakeComplete()`;
- Improve tests to verity new callback;
- Adjust `ConnectionObserverInitializer` to trigger
`onSecurityHandshake()` earlier when TCP Fast Open is enabled and
available;

Result:

Users can observe TCP and TLS handshake behavior correctly when
Fast Open is used.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
@@ -55,10 +55,14 @@ public interface ConnectionObserver {
      * Callback when a security handshake is initiated.
      * <p>
      * For the typical TCP connection, this callback is invoked after {@link #onTcpHandshakeComplete()}. When
-     * {@link ServiceTalkSocketOptions#TCP_FASTOPEN_CONNECT} option is used and the Fast Open feature is supported by
-     * the OS, this callback may be invoked earlier.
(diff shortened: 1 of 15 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -44,10 +44,25 @@
      */
     void onFlush();
 
+    /**
+     * Callback when a TCP handshake completes.
+     */
+    default void onTcpHandshakeComplete() {
```
Scottmitch, line 50: I would recommend avoid using `tcp` at this level bcz we could use the same abstractions for other protocols (udp, ..). Netty uses `channelActive` terminology and we could use something similar here.

idelpivnitskiy: Agreed, changed it to `onTransportHandshakeComplete()`. Debated with myself if this callback really necessary or not. Discussed with @tkountis and he convinced me that it's a lot easier to understand API and events sequencing of the `ConnectionObserver` if we have it. Thanks for discussion!

Scottmitch (approved):

PR #1971, opened 2021-11-30
Add support for io_uring

idelpivnitskiy:
Motivation:

io_uring is a new asynchronous I/O API for Linux that provides
significant performance improvements for I/O operations. Netty
implements it through `netty-incubator-transport-io_uring` module.

Modifications:

- Add `netty-incubator-transport-io_uring` dependencies;
- Add `io.servicetalk.transport.netty.tryIoUring` system property to
activate io_uring transport;
- Let ST use io_uring transport when the system property is set to
`true` and the feature is supported by the OS;
- Test the new feature;

Result:

Users can opt-in for usage of io_uring.


gradle.properties
```diff
@@ -30,6 +30,7 @@ ciManagementUrl=https://github.com/apple/servicetalk/actions
 
 # dependency versions
 nettyVersion=4.1.72.Final
+nettyIoUringVersion=0.0.10.Final
```
Scottmitch, line 33: `0.0.11.Final`

Scottmitch (changes requested):

servicetalk-test-resources/src/main/resources/log4j2.xml
```diff
@@ -27,6 +27,8 @@
     </Console>
   </Appenders>
   <Loggers>
+    <!-- FIXME: remove after verifying CI logs -->
```
Scottmitch, line 30: remove?

Scottmitch (approved):

PR #1972, opened 2021-12-01
Introduce `TimeSource` in `Executor`

chemicL:
Motivation:

Currently, time sensitive operations use `System.nanoTime()` and it is
difficult to control the passage of time indepdendently of the
underlying clock. That makes it more difficult to simulate the passing
of time in tests.

Modifications:

- `io.servicetalk.concurrent.TimeSource` interface was added, which
  includes `currentTime(TimeUnit)` method for extracting the current
  time for operations such as timeouts,
- `io.servicetalk.concurrent.Executor` interface extends `TimeSource`
  and provides a default implementation using `System.nanoTime()`,
- Time sensitive operations that have access to an `Executor` have been
  updated to use `TimeSource` instead of `nanoTime()`,
- `TimeoutPublisherTest` has been altered to incorporate the change,
- Deprecated `TimeoutFromRequest` and introduced `BiFunction<HttpRequestMetadata, TimeSource, Duration>` in constructors of timeout filters, deprecating the old variants as well.

Result:

The concurrent API is prepared for introducing the concept of global
time mocking in the future. Time sensitive operations use a unified
concept of time with the characteristics of `System.nanoTime()`.
Areas which have not been addressed are `BlockingIterable` and related
blocking classes, and Netty related scheduling on the event loop, which
at present we have no control over.

### DIFF
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
@@ -15,14 +15,19 @@
  */
 package io.servicetalk.http.utils;
 
+import io.servicetalk.concurrent.TimeSource;
 import io.servicetalk.http.api.HttpExecutionStrategies;
 import io.servicetalk.http.api.HttpExecutionStrategy;
 import io.servicetalk.http.api.HttpRequestMetaData;
 import io.servicetalk.transport.api.ExecutionStrategyInfluencer;
 
 import java.time.Duration;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 import javax.annotation.Nullable;
+
+import static java.lang.System.nanoTime;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
 
 /**
  * A function to determine the appropriate timeout to be used for a given {@link HttpRequestMetaData HTTP request}.
@@ -37,10 +42,24 @@ public interface TimeoutFromRequest extends
      *
      * @param request the current request
      * @return The timeout or null for no timeout
+     * @deprecated Use {@link #apply(HttpRequestMetaData, TimeSource)}. This method will be removed in a future release.
      */
     @Override
+    @Deprecated
     @Nullable
-    Duration apply(HttpRequestMetaData request);
+    default Duration apply(HttpRequestMetaData request) {
+        return apply(request, (unit) -> unit.convert(nanoTime(), NANOSECONDS));
+    }
+
+    /**
+     * Determine timeout duration, if present, from a request and/or apply default timeout durations.
+     *
+     * @param request the current request
+     * @param timeSource {@link TimeSource} for calculating the timeout
+     * @return The timeout or null for no timeout
+     */
+    @Nullable
+    Duration apply(HttpRequestMetaData request, TimeSource timeSource);
 
     /**
      * {@inheritDoc}
@@ -61,13 +80,13 @@ public interface TimeoutFromRequest extends
      * @return {@link TimeoutFromRequest} instance which applies the provided function and requires the specified
      * strategy.
      */
-    static TimeoutFromRequest toTimeoutFromRequest(final Function<HttpRequestMetaData, Duration> function,
+    static TimeoutFromRequest toTimeoutFromRequest(final BiFunction<HttpRequestMetaData, TimeSource, Duration> function,
                                                    final HttpExecutionStrategy requiredStrategy) {
         return new TimeoutFromRequest() {
             @Nullable
             @Override
-            public Duration apply(final HttpRequestMetaData request) {
-                return function.apply(request);
+            public Duration apply(final HttpRequestMetaData request, final TimeSource timeSource) {
+                return function.apply(request, timeSource);
             }
 
             @Override
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
@@ -26,14 +26,17 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import javax.annotation.Nullable;
 
-import static java.lang.System.nanoTime;
-import static java.util.concurrent.TimeUnit.NANOSECONDS;
-
 /**
  * A function to determine the appropriate timeout to be used for a given {@link HttpRequestMetaData HTTP request}.
  * The result is a {@link Duration} which may be null if no timeout is to be applied. If the function blocks then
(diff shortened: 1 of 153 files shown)


servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
```diff
@@ -37,10 +42,24 @@
...
+    /**
+     * Determine timeout duration, if present, from a request and/or apply default timeout durations.
+     *
+     * @param request the current request
+     * @param timeSource {@link TimeSource} for calculating the timeout
+     * @return The timeout or null for no timeout
+     */
     @Nullable
-    Duration apply(HttpRequestMetaData request);
+    Duration apply(HttpRequestMetaData request, TimeSource timeSource);
```
chemicL, line 62: Note this is a breaking change and hard to avoid due to the nature of `@FunctionalInterface` from `java.util.function.Function`. This change is required in order to make the gRPC deadline use `TimeSource` instead of `System.nanoTime()`. We can discard the latest commit if we don't want to introduce this breaking change.

idelpivnitskiy: This is similar to the case we had with a new method on LB interface. We can still do that without breaking change, it will just require one more release:
1. add new default method, keep the pre-existing as-is in 0.41;
2. swap default methods in 0.42;
3. kill old method in 0.43;

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
```diff
@@ -37,10 +42,24 @@
      *
      * @param request the current request
      * @return The timeout or null for no timeout
+     * @deprecated Use {@link #apply(HttpRequestMetaData, TimeSource)}. This method will be removed in a future release.
      */
     @Override
+    @Deprecated
+    @Nullable
+    default Duration apply(HttpRequestMetaData request) {
```
idelpivnitskiy, line 50: There was a discussion to remove this interface because we agreed that all user-provided funtrions have to be non-blocking => we don't need `ExecutionStrategyInfluencer` here => we can use `java.util.Function`.
//cc @bondolo 

We can consider doing it at the same time you need to change it for gRPC in this PR or in a follow-up to use `BiFunction`.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/EventLoopAwareNettyIoExecutor.java
```diff
@@ -24,9 +24,9 @@
 public interface EventLoopAwareNettyIoExecutor extends NettyIoExecutor {
 
     /**
-     * Checks if the calling thread is an I/O thread manager by this {@link NettyIoExecutor}.
+     * Checks if the calling thread is an I/O thread managed by this {@link NettyIoExecutor}.
```
idelpivnitskiy, line 27: 👁️ 🦅

idelpivnitskiy (commented):
Overall LGTM. Waiting for decision of `TimeoutFromRequest` and `FIXME` comments

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
chemicL: I considered it. The drawbacks are the following:
* Users of 0.41 who currently provide a lambda, in order to get rid of the deprecated method, would need to swap the lambda with an actual implementation of the interface, and would still need to implement the non-default method,
* If a user ignores the deprecation, upon bumping to 0.42 the lambda won't work as the new method takes 2 arguments, and the 1-argument variant no longer is subject to the semantics of a functional interface. They will be forced to migrate.

As a result, that sounds like a breaking change in 0.42 is either the same amount of work for the user or less, depending on user's strategy. WDYT?

chemicL: Also, I'm not sure how frequently this interface is incorporated into codebases by end users.

The only surfacing area I see is:

`TimeoutHttpRequesterFilter` and `TimeoutHttpServiceFilter` constructors with `TimeoutFromRequest` as an argument -> that we could deprecate and provide overrides whose arguments are:

```
Function<HttpRequestMetaData, Duration> // deprecated, applies default strategy
Function<HttpRequestMetaData, Duration>, HttpExecutionStrategy strategy // deprecated
BiFunction<HttpRequestMetadata, TimeSource, Duration> // applies default strategy
BiFunction<HttpRequestMetadata, TimeSource, Duration>, HttpExecutionStrategy strategy
```

but this strategy is also problematic, as once we change the signatures in `TimeoutFromRequest` later we will introduce binary incompatibility even though source compatibility would be maintained. I'm just trying to come up with a smooth transition, but this also sounds as inconvenient.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
chemicL: We can't bring the `TimeSource` change to 0.41, so anything that has to do with `BiFunction` in this context would not make sense in 0.41. Therefore, the plan would look like the following:

1. Deprecate `TimeoutFromRequest` and usages and introduce variants accepting `Function` instead. Merge to main and cherry-pick to 0.41.
2. Pick up that change in this branch and remove `TimeoutFromRequest`, introduce variants for `BiFunction` and deprecate all `Function` variants.
3. In 0.43 we can remove `Function` variants.

Would that accomplish what you had in mind?

bondolo: The proposed plan sounds good to me.

idelpivnitskiy: During the offline discussion we agreed to deprecate `TimeoutFromRequest` in 0.42 only, deprecate ctros that take `TimeoutFromRequest`. Instead, we will add ctors that take `BiFunction<HttpRequestMetaData, TimeSource, Duration>`. The `TimeoutFromRequest` will be removed in 0.43.

chemicL: And now that's been implemented.

chemicL:
> Overall LGTM. Waiting for decision of `TimeoutFromRequest` and `FIXME` comments

Both were addressed in the latest commit. Let's 🚢 if there are no more concerns.

idelpivnitskiy (approved):
Highlight changes for `TimeoutFromRequest` in the PR description and 🚢

PR #1973, opened 2021-12-01
Make `GlobalExecutor` part of `concurrent-api` module

chemicL:
Motivation:

`Executors.immediate()` was used in time sensitive operators, which
don't accept an `Executor` instance. When the `ImmediateExecutor` was
used to offload a scheduled task, such as a timeout, it would have been
executed on the global
`io.servicetalk.concurrent.api.DefaultExecutor.SingleThreadedScheduler`
which is used for all future executions internally. When the actual
termination happened, if it blocked, it would have stalled all following
scheduled delayed tasks.

Modifications:

- Moved `GlobalExecutor` from `transport-netty-internal` module to
  `concurrent-api`,
- Introduced `Executors.global()` `Executor`, which uses a singleton of
  `GlobalExecutor`,
- Using `Executors.global()` in all `timeout(...)` operators that do not
  accept `Executor` as an argument.

Result:

The offloading of terminating a `Timeout*` operators happens on the
`GlobalExecutor`, which is the default concurrent instance also used for
reactive chain processing in all transport related operations.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
@@ -42,7 +42,7 @@ import java.util.function.Supplier;
 import static io.servicetalk.concurrent.api.CompletableDoOnUtils.doOnCompleteSupplier;
 import static io.servicetalk.concurrent.api.CompletableDoOnUtils.doOnErrorSupplier;
 import static io.servicetalk.concurrent.api.CompletableDoOnUtils.doOnSubscribeSupplier;
-import static io.servicetalk.concurrent.api.Executors.immediate;
+import static io.servicetalk.concurrent.api.Executors.global;
 import static io.servicetalk.concurrent.api.Publisher.from;
 import static io.servicetalk.concurrent.api.Publisher.fromIterable;
 import static io.servicetalk.concurrent.internal.SubscriberUtils.deliverErrorFromSource;
@@ -468,7 +468,7 @@ public abstract class Completable {
      * @see <a href="http://reactivex.io/documentation/operators/timeout.html">ReactiveX timeout operator.</a>
      */
     public final Completable timeout(long duration, TimeUnit unit) {
-        return timeout(duration, unit, immediate());
+        return timeout(duration, unit, global());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


bondolo (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Executors.java
```diff
@@ -48,6 +49,22 @@ public static Executor immediate() {
...
     }
 
+    /**
+     * Returns a global {@link Executor} instance, which should be used for operations that do not involve I/O and
+     * are expected to happen concurrently.
+     * It creates as many threads as required but reuses threads when possible. It is therefore 'safe to block' when
+     * using it.
+     * <p>
+     * <strong>Be cautious</strong> though, and prefer creating dedicated {@link Executor executors}
+     * with proper limits that match the profile of your application and use case
```
idelpivnitskiy, line 59: Consider removing this suggestion for now. Executors with limits (queues) may lead to a deadlock if one callback depends on another one queued for later execution.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -468,7 +468,7 @@ public final Completable whenCancel(Runnable onCancel) {
      * @see <a href="http://reactivex.io/documentation/operators/timeout.html">ReactiveX timeout operator.</a>
      */
     public final Completable timeout(long duration, TimeUnit unit) {
-        return timeout(duration, unit, immediate());
+        return timeout(duration, unit, global());
```
idelpivnitskiy, line 471: Ufff... an overload that does not take an explicit `Executor` makes everything harder for users who want to disable offloading. `immediate()` is also bad in this case. In future, we should consider `IoExecutor implements io.servicetalk.concurrent.Executor`

idelpivnitskiy (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
chemicL: `NettyIoExecutor` has `Executor asExecutor()` method BTW.

PR #1977, opened 2021-12-03
Fix client expected strategies regression

bondolo:
Motivation:
The computed execution strategy used by the client was different in
some cases than 0.41
Modifications:
Improve `ClientEffectiveStrategyTest` to cover more cases and ensure
that the strategy used matches, where reasonable, the behaviour from
0.41. Some cases, where the strategy from filters or connection factory
was `defaultStrategy()` or `noOffloadsStrategy()` are now ignored with
a warning. Influencers must use a computed strategy rather one of the
special strategies.
Result:
More compatible and consistent client behavior with computed execution
strategy.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ServiceToStreamingService.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ServiceToStreamingService.java
@@ -19,6 +19,7 @@ import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.Single;
 
 import static io.servicetalk.http.api.DefaultHttpExecutionStrategy.OFFLOAD_RECEIVE_DATA_AND_SEND_STRATEGY;
+import static io.servicetalk.http.api.HttpExecutionStrategies.defaultStrategy;
 import static java.util.Objects.requireNonNull;
 
 final class ServiceToStreamingService extends AbstractServiceAdapterHolder {
@@ -29,7 +30,7 @@ final class ServiceToStreamingService extends AbstractServiceAdapterHolder {
     private final HttpService original;
 
     ServiceToStreamingService(final HttpService original, HttpExecutionStrategy strategy) {
-        super(DEFAULT_STRATEGY.merge(strategy));
+        super(defaultStrategy() == strategy ? DEFAULT_STRATEGY : strategy);
         this.original = requireNonNull(original);
     }
 
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
@@ -24,6 +24,7 @@ import io.servicetalk.http.api.StreamingHttpClientFilterFactory;
 import io.servicetalk.http.api.StreamingHttpConnectionFilterFactory;
 import io.servicetalk.transport.api.ConnectExecutionStrategy;
 import io.servicetalk.transport.api.ExecutionStrategy;
+import io.servicetalk.transport.api.ExecutionStrategyInfluencer;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -58,24 +59,20 @@ final class ClientStrategyInfluencerChainBuilder {
     }
 
     void add(StreamingHttpClientFilterFactory clientFilter) {
-        HttpExecutionStrategy filterOffloads = clientFilter.requiredOffloads();
-        if (defaultStrategy() == filterOffloads || noOffloadsStrategy() == filterOffloads) {
-            LOGGER.warn("Ignoring illegal filter required strategy ({}) for {}", filterOffloads, clientFilter);
-            filterOffloads = anyStrategy();
-        }
-        if (defaultStrategy() != filterOffloads && filterOffloads.hasOffloads()) {
-            clientChain = null != clientChain ? clientChain.merge(filterOffloads) : filterOffloads;
-        }
+        add("filter", clientFilter, clientFilter.requiredOffloads());
     }
 
     void add(HttpLoadBalancerFactory<?> lb) {
-        HttpExecutionStrategy filterOffloads = lb.requiredOffloads();
-        if (defaultStrategy() == filterOffloads || noOffloadsStrategy() == filterOffloads) {
-            LOGGER.warn("Ignoring illegal load balancer required strategy ({}) for {}", filterOffloads, lb);
-            filterOffloads = anyStrategy();
+        add("load balancer", lb, lb.requiredOffloads());
+    }
+
+    private void add(String purpose, ExecutionStrategyInfluencer<?> influencer, HttpExecutionStrategy strategy) {
+        if (defaultStrategy() == strategy || noOffloadsStrategy() == strategy) {
+            LOGGER.warn("Ignoring illegal {} required strategy ({}) for {}", purpose, strategy, influencer);
+            strategy = anyStrategy();
         }
-        if (defaultStrategy() != filterOffloads && filterOffloads.hasOffloads()) {
-            clientChain = null != clientChain ? clientChain.merge(filterOffloads) : filterOffloads;
+        if (strategy.hasOffloads()) {
+            clientChain = null != clientChain ? clientChain.merge(strategy) : strategy;
         }
     }
 
@@ -99,17 +96,21 @@ final class ClientStrategyInfluencerChainBuilder {
                     filterOffloads, connectionFilter);
             filterOffloads = anyStrategy();
         }
-        if (defaultStrategy() != filterOffloads && filterOffloads.hasOffloads()) {
+        if (filterOffloads.hasOffloads()) {
             connFilterChain = null != connFilterChain ? connFilterChain.merge(filterOffloads) : filterOffloads;
         }
     }
 
     HttpExecutionStrategy buildForClient(HttpExecutionStrategy transportStrategy) {
-        // We merge against default in case transport strategy is default.
-        return transportStrategy.merge(null != clientChain ? clientChain : defaultStrategy())
-                .merge(null != connFilterChain ? connFilterChain : defaultStrategy())
-                .merge(null != connFactoryChain ?
-                        HttpExecutionStrategy.from(buildForConnectionFactory()) : defaultStrategy());
+        HttpExecutionStrategy clientStrategy =
+                null != clientChain ? transportStrategy.merge(clientChain) : transportStrategy;
+        if (null != connFilterChain) {
+            clientStrategy = clientStrategy.merge(connFilterChain);
+        }
+        if (null != connFactoryChain) {
+            clientStrategy = clientStrategy.merge(HttpExecutionStrategy.from(buildForConnectionFactory()));
+        }
+        return clientStrategy;
     }
 
     ExecutionStrategy buildForConnectionFactory() {
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
@@ -67,7 +67,7 @@ final class ClientStrategyInfluencerChainBuilder {
     }
 
     private void add(String purpose, ExecutionStrategyInfluencer<?> influencer, HttpExecutionStrategy strategy) {
-        if (defaultStrategy() == strategy || noOffloadsStrategy() == strategy) {
+        if (noOffloadsStrategy() == strategy) {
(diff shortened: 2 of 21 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/ServiceToStreamingService.java
```diff
@@ -29,7 +30,7 @@
     private final HttpService original;
 
     ServiceToStreamingService(final HttpService original, HttpExecutionStrategy strategy) {
-        super(DEFAULT_STRATEGY.merge(strategy));
+        super(defaultStrategy() == strategy ? DEFAULT_STRATEGY : strategy);
```
Scottmitch, line 33: consider putting this in a package private utility method to make it easier to detect common special cases.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
```diff
@@ -46,32 +58,68 @@ private ClientStrategyInfluencerChainBuilder(ClientStrategyInfluencerChainBuilde
...
+        if (defaultStrategy() != filterOffloads && filterOffloads.hasOffloads()) {
+            connFactoryChain = null != connFactoryChain ?
+                    connFactoryChain.merge(filterOffloads) : ConnectAndHttpExecutionStrategy.from(filterOffloads);
+        }
     }
 
     void add(StreamingHttpConnectionFilterFactory connectionFilter) {
-        connFilterChain = connFilterChain.merge(connectionFilter.requiredOffloads());
+        HttpExecutionStrategy filterOffloads = connectionFilter.requiredOffloads();
+        if (defaultStrategy() == filterOffloads || noOffloadsStrategy() == filterOffloads) {
```
Scottmitch, line 97: consider sharing code for the above methods. IIUC the code is the same except for the log statement?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ClientEffectiveStrategyTest.java
```diff
@@ -44,277 +43,284 @@
...
+            HttpExecutionStrategies.customStrategyBuilder().offloadSend().build(),
+            offloadAll(),
+    };
+
+    static final ServerContext context;
+
+    static {
+        try {
+            HttpServerBuilder serverBuilder = HttpServers.forAddress(localAddress(0));
+            context = serverBuilder.listenBlocking((ctx, request, responseFactory) ->
```
Scottmitch, line 137: should this be created in a `@BeforeAll` since it is cleaned up in a `@AfterAll`?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ClientEffectiveStrategyTest.java
```diff
@@ -44,277 +43,284 @@
...
+                    offloadPoints.add(ResponseMeta);
+                }
+                if (effectiveStrategy.isDataReceiveOffloaded()) {
+                    offloadPoints.add(ResponseData);
+                }
+            }
 
-        ClientInvokingThreadRecorder(final InvokingThreadsRecorder<ClientOffloadPoint> holder) {
-            this.holder = holder;
+            System.out.printf("API=%s effectiveStrategy=%s offloads=%s%n",
```
Scottmitch, line 322: use logger or remove debug statement plz

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ClientEffectiveStrategyTest.java
```diff
@@ -44,277 +43,284 @@
...
+                  @Nullable final HttpExecutionStrategy builderStrategy,
+                  @Nullable final HttpExecutionStrategy filterStrategy,
+                  @Nullable final HttpExecutionStrategy lbStrategy,
+                  @Nullable final HttpExecutionStrategy cfStrategy) throws Exception {
+        HttpExecutionStrategy effectiveStrategy = null == builderStrategy ?
+                defaultStrategy() : builderStrategy;
+        effectiveStrategy = null == filterStrategy || anyStrategy() == filterStrategy ||
+                defaultStrategy() == filterStrategy || noOffloadsStrategy() == filterStrategy ?
+                effectiveStrategy : effectiveStrategy.merge(filterStrategy);
+        effectiveStrategy = null == lbStrategy || anyStrategy() == lbStrategy ||
```
Scottmitch, line 180: can you put this logic in a function to share code and provide the operation a name for ease of understanding?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ClientEffectiveStrategyTest.java
```diff
@@ -44,277 +43,284 @@
...
+            case Blocking:
+                BlockingHttpClient blockingClient = client.asBlockingClient();
+                buffers = Collections.singleton(blockingClient.request(blockingClient.get("/")).payloadBody());
+                break;
+            case BlockingStreaming:
+                BlockingStreamingHttpClient blockingStreamingClient = client.asBlockingStreamingClient();
+                buffers = blockingStreamingClient.request(blockingStreamingClient.get("/")).payloadBody();
+                break;
+            case AsyncStreaming:
+                buffers = client.request(client.get("/")).flatMapPublisher(StreamingHttpResponse::payloadBody)
```
Scottmitch, line 258: we should be able to avoid the custom `buffersToString` conversion here by collecting into a `CompositeBuffer` and using `toString`.

```java
                String result = client.request(client.get("/"))
                                      .flatMap(resp -> resp.payloadBody().collect(() ->
                                        client.executionContext().bufferAllocator().newCompositeBuffer(),
                                        CompositeBuffer::addBuffer))
                                      .toFuture().get()
                                      .toString(StandardCharsets.US_ASCII)
```

Scottmitch (commented):
did a quick pass (assuming the validation covers the expected cases).

idelpivnitskiy (commented):
Overall LGTM, only one comment:

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ClientEffectiveStrategyTest.java
bondolo: Using `@BeforeAll` would require that the static be mutable. I (we?, everyone?) prefer `static  final` when possible so we end up with a wart of having to use `@AfterAll` without before all. Making a `ServerContext` extension as we have done with `ExecutionContextExtension` and `ExecutorExtension` would be the next step to take if we use `ServerContext` in this way in multiple tests.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ClientEffectiveStrategyTest.java
bondolo: removed

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ClientEffectiveStrategyTest.java
bondolo: Nice!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
```diff
@@ -46,32 +58,68 @@ private ClientStrategyInfluencerChainBuilder(ClientStrategyInfluencerChainBuilde
     }
 
     void add(StreamingHttpClientFilterFactory clientFilter) {
-        clientChain = clientChain.merge(clientFilter.requiredOffloads());
+        HttpExecutionStrategy filterOffloads = clientFilter.requiredOffloads();
+        if (defaultStrategy() == filterOffloads || noOffloadsStrategy() == filterOffloads) {
```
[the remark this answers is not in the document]
bondolo: It is really late in the game but I am tempted to redefine `noOffloadsStrategy` so that it is consistent with what you suggest and define another "neverOffloadStrategy()" constant for the "force no offloading". I am concerned that the inconsistent behavior between `noOffloadsStrategy` for builder (which means **disable offloading**) and filters (which means "no offloading needed").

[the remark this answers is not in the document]
idelpivnitskiy: Sounds reasonable. Only one worry: what will happen with 0.41 users who already use `noOffloadsStrategy()` on the builder after they upgrade to 0.42? Will that work as in 0.41? If not, we can consider doing the following:
1. Deprecate `noOffloadsStrategy()` and introduce `neverOffloadStrategy()` as a replacement they should use.
2. Rename `anyStrategy()` to `noOffloadingStrategy()` (slightly different name just to avoid changing behavior for the pre-existing name)
More hustle for users to rename what they use, but should keep the expected behavior IIUC. WDYT?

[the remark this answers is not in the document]
bondolo: OK, I will propose this as a separate PR.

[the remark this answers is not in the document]
bondolo: > So, `defaultStrategy()` can be replaced with `offloadAll()`

This is problematic because the API conversion step applies different offloads for the default and converting default to offloadAll at the factory addition stage would force additional unnecessary offloading. I am still tempted to have using default be an error.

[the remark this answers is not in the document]
idelpivnitskiy: The `DefaultHttpExecutionStrategyInfluencer` default impl is `offloadAll()`. If users override it and return `defaultStrategy()` instead, does it change anything? IIRC, filter chain computation is always additive.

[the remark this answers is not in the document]
bondolo: Returning `defaultStrategy()`probably doesn't have any effect except that is more ambiguous than `offloadAll()` and if we substitute `offloadAll()` we lose the ability of API conversions to substitute less offloading when default is presented. I'd prefer if influencers only returned non-special execution strategies and I don't see any useful situation for them to return special strategies.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
bondolo: slightly improved. Differing types prevent more sharing.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
```diff
@@ -46,32 +59,68 @@ private ClientStrategyInfluencerChainBuilder(ClientStrategyInfluencerChainBuilde
...
+        add("filter", clientFilter, clientFilter.requiredOffloads());
     }
 
     void add(HttpLoadBalancerFactory<?> lb) {
-        clientChain = clientChain.merge(lb.requiredOffloads());
+        add("load balancer", lb, lb.requiredOffloads());
+    }
+
+    private void add(String purpose, ExecutionStrategyInfluencer<?> influencer, HttpExecutionStrategy strategy) {
+        if (defaultStrategy() == strategy || noOffloadsStrategy() == strategy) {
```
idelpivnitskiy, line 70: > if we substitute `offloadAll()` we lose the ability of API conversions to substitute less offloading when default is presented.
 
Agreed, replacing it with `offloadAll()` will have other side-effects. WDYT if we keep it as-is? Merging `defaultStrategy()` with `defaultStrategy()` will result in `defaultStrategy()`, heloing us to preserve all the flags set without side-effects for merging with API conversions?

idelpivnitskiy (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
bondolo: I have removed `defaultStrategy()` as an error case for now.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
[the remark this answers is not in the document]
bondolo: `defaultStrategy()` is now allowed without generating a warning. I think the docs should still suggest that it is not good to return it from `requiredOffloads()`

PR #1978, opened 2021-12-03
Provide more context when native transport is unavailable

idelpivnitskiy:
Motivation:

Users who fail to load native transport may experience hard to debug
issues because Java NIO is less reactive to the network events and
does not proactively notify pipeline when the channel becomes inactive.

Modifications:

- Provide more context in the warn message and notify ST will fail to
start in future releases unless native transport was explicitly disabled;

Result:

More clear warning when native transport can not be loaded.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NativeTransportUtils.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NativeTransportUtils.java
@@ -29,6 +29,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static io.netty.util.internal.PlatformDependent.normalizedArch;
+import static java.lang.Boolean.getBoolean;
 
 /**
  * Utility to check availability of Netty <a href="https://netty.io/wiki/native-transports.html">native transports</a>.
@@ -39,6 +40,8 @@ import static io.netty.util.internal.PlatformDependent.normalizedArch;
 final class NativeTransportUtils {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NativeTransportUtils.class);
+    private static final String NETTY_NO_NATIVE_NAME = "io.netty.transport.noNative";
+    private static final boolean NETTY_NO_NATIVE = getBoolean(NETTY_NO_NATIVE_NAME);
 
     private static final boolean IS_LINUX;
     private static final boolean IS_OSX_OR_BSD;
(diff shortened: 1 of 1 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NativeTransportUtils.java
```diff
@@ -39,6 +40,8 @@
 final class NativeTransportUtils {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NativeTransportUtils.class);
+    private static final String NETTY_NO_NATIVE_NAME = "io.netty.transport.noNative";
```
Scottmitch, line 43: `noNative` -> can we avoid `no` in the name which leads to double negatives and can be confusing? What about `forceNio`?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NativeTransportUtils.java
```diff
@@ -60,9 +63,19 @@ private NativeTransportUtils() {
     }
 
     private static void logUnavailability(final String transport, final String os, final Throwable cause) {
+        if (NETTY_NO_NATIVE) {
+            LOGGER.warn("io.netty:netty-transport-native-{} is explicitly disabled with \"-D{}=true\". Note that it " +
```
Scottmitch, line 67: `LOGGER.warn` -> should this be `info` or something lower as they are explicitly opting in?

Scottmitch (approved):
lgtm after comments resolved

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NativeTransportUtils.java
idelpivnitskiy: This is the system property defined by netty, we can not change it: 
https://github.com/netty/netty/blob/1e189654f913b13be3616c552a04459465b752b3/transport-classes-epoll/src/main/java/io/netty/channel/epoll/Epoll.java#L33-L35

Scottmitch: Do we have to use the same property as Netty? We can decide at the ST level if we should load the libraries or not, and may want different levels of control (e.g. use io_uring vs epoll, etc.).

idelpivnitskiy: `-Dio.netty.transport.noNative=true` is the only way today to enforce Java NIO for users who really need it. In order for us to be backward compatible with this use-case, we should read the same property to decide when to throw an exception. If I introduce a new property, existing users will be broken after #1979 until they find out what is the new property they have to use. We can always add our own property if we find a use-case for that.

For io_uring, a new property is added in #1971.

idelpivnitskiy: I will start with netty property for now, we can add another one later if necessary

PR #1979, opened 2021-12-03
Fail to start without native transport

idelpivnitskiy:
Motivation:

Users who fail to load native transport may experience hard to debug
issues because Java NIO is less reactive to the network events and
does not proactively notify pipeline when the channel becomes inactive.
In most cases, it's unexpected to run without native transport and
represents a bug with packaging or deployment.

Modifications:

- Throw an exception if the native transport can not be loaded without
being explicitly disabled;

Result:

ST does not start without native transport unless it's explicitly
disabled.

Depends on #1978.


tkountis (approved):
Looking forward to see how many complain about it. 😨

PR #1983, opened 2021-12-03
`publishOn/subscribeOn` take `io.servicetalk.concurrent.Executor`

idelpivnitskiy:
Motivation:

`publishOn/subscribeOn` does not need any API that
`io.servicetalk.concurrent.api.Executor` define and can work with
`io.servicetalk.concurrent.Executor`.

Modifications:

- Change argument to `io.servicetalk.concurrent.Executor` for
`publishOn/subscribeOn` in `Completable/Single/Publisher`;
- Update javadoc for modified methods;

Result:

`publishOn/subscribeOn` API is less restrictive for the passed executor.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
@@ -481,7 +481,8 @@ public abstract class Completable {
      * {@link Subscriber#onError(Throwable) terminated}.
      * @param duration The time duration which is allowed to elapse before {@link Subscriber#onComplete()}.
      * @param unit The units for {@code duration}.
-     * @param timeoutExecutor The {@link Executor} to use for managing the timer notifications.
+     * @param timeoutExecutor The {@link io.servicetalk.concurrent.Executor} to use for managing the timer
+     * notifications.
      * @return a new {@link Completable} that will mimic the signals of this {@link Completable} but will terminate with
      * a {@link TimeoutException} if time {@code duration} elapses before {@link Subscriber#onComplete()}.
      * @see <a href="http://reactivex.io/documentation/operators/timeout.html">ReactiveX timeout operator.</a>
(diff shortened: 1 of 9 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1419,88 +1421,91 @@ public final Completable liftAsync(CompletableOperator operator) {
...
-     * @return A new {@link Completable} that will use the passed {@link Executor} to invoke all {@link Subscriber}
-     * methods.
-     * @see #publishOn(Executor, BooleanSupplier)
+     * @param executor {@link io.servicetalk.concurrent.Executor} to use.
+     * @return A new {@link Completable} that will use the passed {@link io.servicetalk.concurrent.Executor} to invoke
+     * all {@link Subscriber} methods.
+     * @see #publishOn(io.servicetalk.concurrent.Executor, BooleanSupplier)
      */
-    public final Completable publishOn(Executor executor) {
+    public final Completable publishOn(io.servicetalk.concurrent.Executor executor) {
```
idelpivnitskiy, line 1438: Decided that we don't need to go through the deprecated approach because these methods are backward compatible for the compilation. We do not need ABI compatibility for 0.42.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -481,7 +481,8 @@ public final Completable timeout(long duration, TimeUnit unit) {
      * {@link Subscriber#onError(Throwable) terminated}.
      * @param duration The time duration which is allowed to elapse before {@link Subscriber#onComplete()}.
      * @param unit The units for {@code duration}.
-     * @param timeoutExecutor The {@link Executor} to use for managing the timer notifications.
+     * @param timeoutExecutor The {@link io.servicetalk.concurrent.Executor} to use for managing the timer
```
chemicL, line 484: Thanks for paying attention here, too.

chemicL (approved):

PR #1986, opened 2021-12-04
Convert `BlockingStreamingHttpServerResponse` to an interface

idelpivnitskiy:
Motivation:

`BlockingStreamingHttpServerResponse` is the only response type that is
an abstract class, not an interface. It limits users ability to wrap it
to intercept the methods because its ctor is pkg-private.

Modifications:

- Convert `BlockingStreamingHttpServerResponse` to an interface;
- Update `DefaultBlockingStreamingHttpServerResponse` to take full
implementation responsibility;

Result:

`BlockingStreamingHttpServerResponse` is consistent with other response
types and users can wrap it now.


bondolo:
This seems like a reasonable change. Do we know if any users ever extended it?

idelpivnitskiy:
I know that some users wrap other types. Would be nice to support the same ability for this API variant.

bondolo (approved):

PR #1994, opened 2021-12-07
Update proto-google-common-protos 2.5.1 -> 2.7.0

idelpivnitskiy:
Depends on #1992


bondolo (approved):

bondolo:
I seem to recall some temporary hack that was in place until we used a newer protobuf version. I don't know where it was, ring any bells?

idelpivnitskiy:
There was something related to protobuf-gradle-plugin & caching

PR #2000, opened 2021-12-08
`ExecutionStrategyInfluencer` cleanup

idelpivnitskiy:
Motivation:

Do the final cleanup of `ExecutionStrategyInfluencer` API.

Modifications:

- Remove `newInfluencer` methods from `ExecutionStrategyInfluencer` and
`HttpExecutionStrategyInfluencer`;
- Default offloading for `ConnectionFactoryFilter` have to be
`offloadAll()`;
- Remove deprecated `noOffloadsStrategy()` from `GrpcExecutionStrategies`
and `HttpExecutionStrategies`;
- `InfluencerConnectionAcceptor` does not need to override `closeAsync()`;
- All HTTP classes/interfaces should use `HttpExecutionStrategyInfluencer`
instead of `ExecutionStrategyInfluencer`;
- Remove `DefaultHttpExecutionStrategyInfluencer` because it does the
same as public `HttpExecutionStrategyInfluencer`;

Result:

Removed deprecations, less public API, consistent use of influencer
interfaces in HTTP.


servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingHttpService.java
```diff
@@ -17,15 +17,14 @@
...
-import io.servicetalk.transport.api.ExecutionStrategyInfluencer;
 
 import static io.servicetalk.http.api.DefaultHttpExecutionStrategy.OFFLOAD_RECEIVE_DATA_STRATEGY;
 
 /**
  * The equivalent of {@link HttpService} but with synchronous/blocking APIs instead of asynchronous APIs.
  */
 @FunctionalInterface
-public interface BlockingHttpService extends ExecutionStrategyInfluencer<HttpExecutionStrategy>, GracefulAutoCloseable {
+public interface BlockingHttpService extends HttpExecutionStrategyInfluencer, GracefulAutoCloseable {
```
idelpivnitskiy, line 27: Should services implement influencer? They don't in 0.41. If a different strategy is necessary, users can use a router to specify one. 
On the other hand, better to keep a definition of the strategy close to the implementation of the service.

bondolo: In 0.41 and earlier individual instances of service could implement `HttpExecutionStrategyInfluencer` and it would influence the execution strategy used. 0.42 just makes the need to implement the interface a requirement and adds a default for each API model. The intention was that if something can be an influencer then the parent interface should be explicit about that.

idelpivnitskiy: Thanks for clarifying, sounds good!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpService.java
```diff
@@ -17,16 +17,14 @@
...
 
 import static io.servicetalk.http.api.DefaultHttpExecutionStrategy.OFFLOAD_RECEIVE_META_STRATEGY;
 
 /**
  * The equivalent of {@link StreamingHttpService} but with synchronous/blocking APIs instead of asynchronous APIs.
  */
 @FunctionalInterface
-public interface BlockingStreamingHttpService extends
-            ExecutionStrategyInfluencer<HttpExecutionStrategy>, GracefulAutoCloseable {
+public interface BlockingStreamingHttpService extends HttpExecutionStrategyInfluencer, GracefulAutoCloseable {
```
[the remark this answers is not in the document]
idelpivnitskiy: Discussed offline, decided to keep `HttpExecutionStrategyInfluencer` for now and use it for all HTTP cases. It gives a safe default for users and migration from 0.41 will be easier if the interface name doesn't change. We can try to remove it later if we see the need.

bondolo (approved):

PR #2005, opened 2021-12-09
Retry unification deprecations removal

tkountis:


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -120,22 +118,12 @@
...
-    boolean customAutoRetry;
-
-    //FIXME: 0.42 - to be removed when auto-retry is removed.
-    @Deprecated
-    boolean retryingFilterAppended;
-
-    @Nullable
-    private RetryingHttpRequesterFilter retryingHttpRequesterFilter;
+    private RetryingHttpRequesterFilter retryingHttpRequesterFilter =
+            new RetryingHttpRequesterFilter.Builder().build();
```
idelpivnitskiy, line 126: Consider using a private static constant for this default

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ClientClosureRaceTest.java
```diff
@@ -175,13 +176,13 @@ private void runIterations(Callable<Single<?>> test) throws Exception {
...
-                        .retryFor((md, t) ->
+                .appendClientFilter(retryBuilder
+                        .retryOther((md, t) ->
                             // This test has the server intentionally hard-close the connection after responding
                             // to the first request, however some tests use pipelining and may write multiple requests
                             // on the same connection which would result in a non-retryable exception. Since this test
                             // doesn't care about idempotency it should always retry.
-                            true
-                        ).buildWithConstantBackoffFullJitter(ofNanos(1)));
+                           ofConstantBackoffFullJitter(ofNanos(1), MAX_VALUE)
```
idelpivnitskiy, line 185: The snapshot build is currently failing with: https://github.com/apple/servicetalk/runs/4476095925
```
Check failure on line 184 in servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ClientClosureRaceTest.java 
Comment has incorrect indentation level 28, expected is 27, indentation should be the same level as line 185.
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -83,7 +83,7 @@
 
     public static final RetryingHttpRequesterFilter DISABLE_RETRIES =
             new RetryingHttpRequesterFilter(false, true, 0, null,
-                (__, ___) -> NO_RETRIES);
+                    (__, ___) -> NO_RETRIES);
```
idelpivnitskiy, line 86: 1. This public constant misses javadoc.
2. Did you consider a public static method instead of a constant?

servicetalk-client-api/src/test/java/io/servicetalk/client/api/DefaultAutoRetryStrategyProviderTest.java
```diff
@@ -1,190 +0,0 @@
...
-    private final TestCompletableSubscriber retrySubscriber;
-
-    DefaultAutoRetryStrategyProviderTest() {
-        lbEvents = new TestPublisher<>();
-        sdStatus = new TestCompletable();
-        retrySubscriber = new TestCompletableSubscriber();
-    }
-
-    @Test
-    void disableWaitForLb() {
```
idelpivnitskiy, line 59: We shouldn't lose these test scenarios when we migrate users to the new filter

idelpivnitskiy (changes requested):

servicetalk-client-api/src/test/java/io/servicetalk/client/api/DefaultAutoRetryStrategyProviderTest.java
tkountis: Good catch, re-added.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
idelpivnitskiy: With the constant, you have less flexibility to adjust it. With a static method it's easier to change what it returns.

idelpivnitskiy (approved):
LGTM after build passes

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -103,10 +104,16 @@
...
     void inject(final Publisher<Object> lbEventStream) {
+        if (this.lbEventStream != null) {
+            throw new ConcurrentModificationException();
+        }
         this.lbEventStream = lbEventStream;
     }
 
     void inject(final Completable sdStatus) {
+        if (this.sdStatus != null) {
+            throw new ConcurrentModificationException();
```
idelpivnitskiy, line 115: Is this even possible? a `create` method is invoked by the client builder, meaning nobody else has a reference to the filter to inject something

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -235,6 +226,16 @@ public Completable closeAsyncGracefully() {
         }
     }
 
+    /**
+     * Retrying filter that disables any form of retry behaviour. All types of failures will not be re-attempted.
+     * @return a retrying filter that disables any form of retry behaviour. All types of failures will not be
+     * re-attempted.
+     */
+    public static RetryingHttpRequesterFilter disableAutoRetries() {
```
idelpivnitskiy, line 241: IIUC, you plan to cherry-pick the new approach into 0.41. Make sure that `AutoRetryStrategyProvider#DISABLE_AUTO_RETRIES` references this new method as a replacement

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
tkountis: I will remove these, it was only for debugging. Regardless, the problem is that the shared DEFAULT is re-injected state in test environment and concurrent testing and this is causing the problem. The factory is holding state. This shouldn't be a problem for real use-cases but nevertheless let's discuss about it before we merge changes.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -407,7 +386,10 @@ private static StreamingHttpClientFilterFactory appendFilter(
         if (currClientFilterFactory == null) {
             return appendClientFilterFactory;
         } else {
-            return client -> currClientFilterFactory.create(appendClientFilterFactory.create(client));
+            return client -> currClientFilterFactory.create(
+                    new ContextAwareDelegateStreamingHttpClient(appendClientFilterFactory.create(client),
+                            ((ContextAwareDelegateStreamingHttpClient) client).lbEventStream(),
+                            ((ContextAwareDelegateStreamingHttpClient) client).sdStatus()));
```
idelpivnitskiy, line 392: With this approach I'm a bit concerned about number of wrapping layers created for each client. Because it will always be on request path, there is a risk it may affect JIT or some other JVM optimizations.

What I had in mind was to extend `StreamingHttpClientFilterFactory` API only for internal purposes:
```java
    @FunctionalInterface
    interface ExtendedStreamingHttpClientFilterFactory extends StreamingHttpClientFilterFactory {

        StreamingHttpClientFilter create(FilterableStreamingHttpClient client, Publisher<Object> lbEventStream, SdStatusCompletable sdStatus);
    }
```
Then we can use it to pass 2 extra arguments:
```java
    private static ExtendedStreamingHttpClientFilterFactory appendExtendedFilter(
            @Nullable final ExtendedStreamingHttpClientFilterFactory currClientFilterFactory,
            final StreamingHttpClientFilterFactory appendClientFilterFactory) {
        if (appendClientFilterFactory instanceof RetryingHttpRequesterFilter) {
            if (currClientFilterFactory == null) {
                return (client, lbEventStream, sdError) -> {
                    ContextAwareRetryingHttpRequesterFilter filter = (ContextAwareRetryingHttpRequesterFilter) ((RetryingHttpRequesterFilter) appendClientFilterFactory).create(client);
                    filter.inject(lbEventStream, sdError);
                    return filter;
                };
            } else {
                return (client, lbEventStream, sdError) -> {
                    ContextAwareRetryingHttpRequesterFilter filter = (ContextAwareRetryingHttpRequesterFilter) ((RetryingHttpRequesterFilter) appendClientFilterFactory).create(client);
                    filter.inject(lbEventStream, sdError);
                    return currClientFilterFactory.create(filter, lbEventStream, sdError);
                };
            }

        } else {
            if (currClientFilterFactory == null) {
                return (client, lbEventStream, sdError) -> appendClientFilterFactory.create(client);
            } else {
                return (client, lbEventStream, sdError) -> currClientFilterFactory.create(appendClientFilterFactory.create(client), lbEventStream, sdError);
            }
        }
    }
```
^ it also creates extra wrappers for `StreamingHttpClientFilterFactory`, but those will be only at the build time.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ContextAwareStreamingHttpClientFilterFactory.java
```diff
@@ -0,0 +1,36 @@
...
+import io.servicetalk.concurrent.api.Publisher;
+import io.servicetalk.http.api.FilterableStreamingHttpClient;
+import io.servicetalk.http.api.StreamingHttpClientFilter;
+import io.servicetalk.http.api.StreamingHttpClientFilterFactory;
+import io.servicetalk.http.netty.DefaultSingleAddressHttpClientBuilder.SdStatusCompletable;
+
+import javax.annotation.Nullable;
+
+@FunctionalInterface
+interface ContextAwareStreamingHttpClientFilterFactory extends StreamingHttpClientFilterFactory {
```
idelpivnitskiy, line 27: If this interface doesn't leak outside `DefaultSingleAddressHttpClientBuilder`, consider defining it there instead of the top level interface

idelpivnitskiy (approved):
Cherry-pick will be tricky :)

tkountis:
Closed in favor of https://github.com/apple/servicetalk/pull/2019

PR #2006, opened 2021-12-10
Let users opt-in for app initialization failures without native libs

idelpivnitskiy:
Motivation:

#1979 fails app initialization if native transport libs can not be
loaded. This is too aggressive. Instead, keep warning users and let
them opt-in for the enforced behavior.

Modifications:

- Fail initialization only if users explicitly opt-it for this using
`-Dio.servicetalk.transport.netty.requireNativeLibs=true`;
- Improve exception and log messages;

Result:

Users can enforce requirement for native libs if they need.


bondolo (approved):

idelpivnitskiy:
Because it's not an enforced behavior anymore, I'm going to cherry-pick this into 0.41

PR #2011, opened 2021-12-10
Give more context on CHANNEL_CLOSED_INBOUND when no native transport

idelpivnitskiy:
Motivation:

When native transport is not used, users can see increased number of
`CHANNEL_CLOSED_INBOUND` exceptions because Java NIO does not notify
us about a closed connection until we make an attempt to write data.

Modifications:

- Provide more context for users when `CHANNEL_CLOSED_INBOUND` happens
with no native transport available;

Result:

Users can get an idea where to look for the root cause for this
use-case.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
@@ -80,6 +80,7 @@ import static io.servicetalk.concurrent.api.SourceAdapters.toSource;
 import static io.servicetalk.concurrent.internal.SubscriberUtils.deliverErrorFromSource;
 import static io.servicetalk.transport.netty.internal.ChannelCloseUtils.close;
 import static io.servicetalk.transport.netty.internal.ChannelSet.CHANNEL_CLOSEABLE_KEY;
+import static io.servicetalk.transport.netty.internal.CloseHandler.CloseEvent.CHANNEL_CLOSED_INBOUND;
 import static io.servicetalk.transport.netty.internal.CloseHandler.UNSUPPORTED_PROTOCOL_CLOSE_HANDLER;
 import static io.servicetalk.transport.netty.internal.Flush.composeFlushes;
 import static io.servicetalk.transport.netty.internal.NettyIoExecutors.fromNettyEventLoop;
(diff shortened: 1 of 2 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -376,6 +377,12 @@ private Throwable enrichError(final Throwable t) {
         } else {
             if ((closeReason = this.closeReason) != null) {
                 throwable = wrapWithCloseReason(closeReason, t);
+                if (closeReason == CHANNEL_CLOSED_INBOUND && t instanceof StacklessClosedChannelException &&
+                        !NativeTransportUtils.isEpollAvailable() && !NativeTransportUtils.isKQueueAvailable()) {
+                    throwable = new StacklessClosedChannelException("Observed closed channel after an attempt to " +
```
idelpivnitskiy, line 382: @Scottmitch wdyt about this approach?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -376,6 +377,12 @@ private Throwable enrichError(final Throwable t) {
         } else {
             if ((closeReason = this.closeReason) != null) {
                 throwable = wrapWithCloseReason(closeReason, t);
+                if (closeReason == CHANNEL_CLOSED_INBOUND && t instanceof StacklessClosedChannelException &&
+                        !NativeTransportUtils.isEpollAvailable() && !NativeTransportUtils.isKQueueAvailable()) {
```
Scottmitch, line 381: Instead of checking transports in general here, consider checking the `EventLoop`/`Channel` type and adding a utility method in `NativeTransportUtils` for this type of check (e.g. `channelNotifiesCloseIfIdel`). this would be similar to the approach in `isUnixDomainSocketSupported` and helps consolidate the "transport conditional logic" in a single place.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
Scottmitch: I like the idea of providing more context in exceptions. this specific implementation seems somewhat indirect and potentially fragile (e.g. exceptions types are pinned, is the the only place the exception needs to be handled, can the exception be wrapped in anything else before this point), but I haven't thought about an alternative.

Scottmitch:
@idelpivnitskiy - should we close this and revisit later if necessary?

PR #2012, opened 2021-12-10
`TransportObserver#onNewConnection` give visibility into addresses

idelpivnitskiy:
Motivation:

Existing `TransportObserver#onNewConnection()` callback does not
provide visibility into an address. In case a connect attempt fails,
users don't know which remote was involved.

Modifications:

- Add a new callback `onNewConnection(Object, Object)` that gives
visibility to local/remote address;
- Deprecate `TransportObserver#onNewConnection()`;
- Update tests and implementation classes;

Result:

`TransportObserver` users have visibility into local/remote address
for each `onNewConnection` attempt.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/TransportObserver.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/TransportObserver.java
@@ -15,6 +15,8 @@
  */
 package io.servicetalk.transport.api;
 
+import javax.annotation.Nullable;
+
 /**
  * An observer interface that provides visibility into transport events.
  */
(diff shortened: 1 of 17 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/TransportObserver.java
```diff
@@ -24,6 +26,19 @@
...
     ConnectionObserver onNewConnection();
+
+    /**
+     * Callback when transport starts initializing a new network connection.
+     *
+     * @param localAddress a local address of a new connection, if known
+     * @param remoteAddress a remote address of a new connection
+     * @return a new {@link ConnectionObserver} that provides visibility into events associated with a new connection
+     */
+    default ConnectionObserver onNewConnection(@Nullable Object localAddress, Object remoteAddress) {
```
idelpivnitskiy, line 41: I tried to use `SocketAddress` as an upper bound for the client-side type, but it's too widespread across all API, making it impossible to be backward compatible or easy to migrate to. Went with an `Object` approach.

bondolo: Wouldn't the local address already be bound at this point? I'd expect at least `ANY_ADDR` with local port.

bondolo (approved):

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/TransportObserver.java
idelpivnitskiy: It will be available on the server-side, but on the client-side it's `null` until we have `connectionEstablished`

PR #2018, opened 2021-12-14
Fix sharing issue with Retrying requests filter factory

tkountis:
Motivation:

New retrying http request filter has bug when the factory is shared across client builders.

Modifications:

Internalized the state to the filter, where the lazy injection now happens rather than the factory as it was before.

Result:

Safer implementation


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -869,4 +887,16 @@ private static HttpHeadersFactory headersFactory(@Nullable HttpHeadersFactory fa
             return factory;
         }
     }
+
+    @FunctionalInterface
+    interface ContextAwareStreamingHttpClientFilterFactory extends StreamingHttpClientFilterFactory {
+        StreamingHttpClientFilter create(FilterableStreamingHttpClient client,
+                                         @Nullable Publisher<Object> lbEventStream,
+                                         @Nullable SdStatusCompletable sdStatus);
```
idelpivnitskiy, line 895: Is it enough to use `Completable` here?

idelpivnitskiy (approved):

PR #2025, opened 2021-12-16
Add Publisher#distinct

Scottmitch:
Motivation:
Publisher#distinct operator currently doesn't exist, but
maybe useful in select use cases with small cardinality
stream or high cardniality of duplicate signals.

Modifications:
- Add Publisher#distinct operator which provides
  https://reactivex.io/documentation/operators/distinct.html
  semantics.

Result:
Publisher#distinct operator exists.


Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/pull/2025

bondolo (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -137,7 +138,57 @@ protected Publisher() {
...
+     * }</pre>
+     *
+     * @param filterSupplier Used to obtain a new {@link Predicate} such that if {@link Predicate#test(Object)} returns
+     * {@code true} the item is emitted to the returned {@link Publisher}. This {@link Predicate} is unique per
+     * subscribe and may therefore keep state and can be used for example to track distinct signals.
+     * @return A {@link Publisher} emits filtered signals observed by this {@link Publisher}.
+     * @see <a href="http://reactivex.io/documentation/operators/filter.html">ReactiveX filter operator.</a>
+     * @see <a href="http://reactivex.io/documentation/operators/distinct.html">ReactiveX distinct operator.</a>
+     */
+    public final Publisher<T> filter(Supplier<? extends Predicate<? super T>> filterSupplier) {
```
idelpivnitskiy, line 166: With this approach, number of operators can easily double. People will come and ask for the same for map/flatMap/when*/etc. For the public API, can we just rely on users to use `defer` when they need a state per subscribe? We can keep this approach as pkg-private for use-cases like `distinct`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -137,7 +138,57 @@ protected Publisher() {
...
+        return new FilterPublisher<>(this, filterSupplier);
+    }
+
+    /**
+     * Only emits distinct signals observed by this {@link Publisher}.
+     * <p>
+     * {@link Publisher#filter(Supplier)} can be used to customize the distinctness criteria.
```
idelpivnitskiy, line 173: Consider clarifying what "distinct" means, like requirements for equals/hashCode.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -137,7 +138,57 @@ protected Publisher() {
...
+     * <p>
+     * This method provides a data transformation in sequential programming similar to:
+     * <pre>{@code
+     *     List<T> results = ...;
```
idelpivnitskiy, line 177: Set?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
Scottmitch: Sure this method can be made private for now. The filter API in this PR provides a simplification of `distinct` specializations in rxjava [1] and [reactor](https://github.com/reactor/reactor-core/blob/main/reactor-core/src/main/java/reactor/core/publisher/Flux.java#L4404-L4408). It is true that folks may ask for similar `Supplier` style APIs on other methods, but in this case there is precedence for it in other libraries.
[1]
```java
public final <@NonNull K> Flowable<T> distinct(@NonNull Function<? super T, K> keySelector,
            @NonNull Supplier<? extends Collection<? super K>> collectionSupplier)
```

PR #2031, opened 2021-12-20
Prevent retry logic modifications through the builder

idelpivnitskiy:
Motivation:

The final retry predicate depends on class variables of the builder in
`RetryingHttpRequesterFilter`. User can keep a reference to the builder
and modify functions to change the retry behavior in runtime.

Modifications:

- Capture the context of the builder state in `final` variables to
eliminate any runtime behavior change in `RetryingHttpRequesterFilter`;

Result:

`RetryingHttpRequesterFilter` behavior can not be altered through the
builder at runtime.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
@@ -705,6 +705,16 @@ public final class RetryingHttpRequesterFilter
          * @return A new retrying {@link RetryingHttpRequesterFilter}
          */
         public RetryingHttpRequesterFilter build() {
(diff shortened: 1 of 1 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -705,6 +705,16 @@ public Builder retryOther(
          * @return A new retrying {@link RetryingHttpRequesterFilter}
          */
         public RetryingHttpRequesterFilter build() {
+            final BiFunction<HttpRequestMetaData, RetryableException, BackOffPolicy> retryRetryableExceptions =
```
tkountis, line 708: Nice!

tkountis (approved):

PR #2032, opened 2021-12-20
`RetryingHttpRequesterFilter`: change max retries to 5

idelpivnitskiy:
Motivation:

Current default is to retry indefinitely. This may lead to an issues,
like retry storm, timeout exceptions, or inability for users to know
that request keeps failing.

Modifications:

- Change default value for `maxRetries` from `MAX_VALUE` to `5`;

Result:

Default version of `RetryingHttpRequesterFilter` propagates an error to
the user after 5 failed attempts.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -530,7 +529,7 @@ public static BackOffPolicy ofExponentialBackoffDeltaJitter(
         private boolean waitForLb = true;
         private boolean ignoreSdErrors;
 
-        private int maxRetries = MAX_VALUE;
+        private int maxRetries = 5;
```
tkountis, line 532: This is the default TOTAL max retries. The individual back-off policies have separate max-retries (eg. 3 for retryable exceptions)

idelpivnitskiy: 🤦 I forgot about that, nvm

PR #2033, opened 2021-12-21
Fix H2 Flush Strategy Aggregated API Optimization

Scottmitch:
Motivation:
72a0e64e9837d673d78f3bbb71499585d6fb0373 made it such that
flushOnEnd optimizations weren't always applied when using
HTTP/2 and gRPC. This may have around a 20% performance impact.

Modifications:
- NettyHttpServer should use SplittingFLushStrategy in all cases.
  This is necessary because the flush strategy needs to be applied
  in a lazy fashion after the write starts, which is a capability
  provided by SplittingFLushStrategy. The terminal conditions are
  also commonly enforced by SplittingFLushStrategy and should be
  reused.
- SplittingFLushStrategy should handle exceptions from the flush
  boundaries. This may happen if header parsing throws (duplicate
  content-length header).
- While aggregating an in memory request track the content-length
  to avoid multiple iterations.

Result:
20% rps performance improvement for aggregated h2/gRPC use cases.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
@@ -109,7 +109,6 @@ import static io.servicetalk.transport.netty.internal.SplittingFlushStrategy.Flu
 import static io.servicetalk.transport.netty.internal.SplittingFlushStrategy.FlushBoundaryProvider.FlushBoundary.Start;
 
 final class NettyHttpServer {
-
     private static final Logger LOGGER = LoggerFactory.getLogger(NettyHttpServer.class);
 
     private NettyHttpServer() {
@@ -249,8 +248,7 @@ final class NettyHttpServer {
         private final NettyConnection<Object, Object> connection;
         private final HttpHeadersFactory headersFactory;
         private final HttpExecutionContext executionContext;
-        @Nullable
-        private final SplittingFlushStrategy splittingFlushStrategy;
+        private final SplittingFlushStrategy flushStrategy;
         private final boolean drainRequestPayloadBody;
         private final boolean requireTrailerHeader;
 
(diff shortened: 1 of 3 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -273,41 +271,39 @@ public String toString() {
...
-                                if (itemWritten instanceof Buffer) {
-                                    return contentLength > 0 &&
-                                            (contentLength -= ((Buffer) itemWritten).readableBytes()) <= 0 ?
-                                            End : InProgress;
-                                }
-                                if (itemWritten instanceof HttpHeaders) {
-                                    return End;
-                                }
-                                return InProgress;
+            flushStrategy = new SplittingFlushStrategy(connection.defaultFlushStrategy(),
```
Scottmitch, line 274: I implemented a simplified version of `SplittingFlushStrategy` which doesn't care as much about switching, but the core concepts of lazy-set and boundaries are the same. figured it was best to avoid more complexity around flush strategy as the goal is to replace/remove `SplittingFlushStrategy` anyways.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -273,41 +271,39 @@ public String toString() {
...
+                            if (itemWritten instanceof HttpResponseMetaData) {
+                                final HttpResponseMetaData metadata = (HttpResponseMetaData) itemWritten;
+                                contentLength = getContentLength(metadata);
+                                // The content length maybe unknown at this point (e.g. 204 response) but then later
+                                // determined to be 0. In that case we should conservatively use End and rely upon
+                                // adjustForMissingBoundaries to accommodate if more data comes. Otherwise the
+                                // FlushStrategy may not trigger a flush (e.g. flushOnEnd) and if so the response
+                                // won't actually be written.
+                                return contentLength > 0 || (contentLength < 0 &&
+                                        (!emptyMessageBody(metadata) &&
```
Scottmitch, line 292: We tend to add `content-length: 0` for the empty body optimization in h2, but figured I would add `emptyMessageBody` to be safe.

idelpivnitskiy (approved):
Thanks!

Scottmitch:
0.42 (30550d518)

PR #2035, opened 2021-12-22
Add Publisher#cast, Single#cast, and Publisher#ofType

Scottmitch:
Motivation:
Filtering and casting based upon types maybe a common
use case but it currently requires knowledge of Class
methods and potentially multiple operators.

Modifications:
- Add Publisher#cast, Single#cast, and Publisher#ofType
  operators and associated tests.

Result:
Operators exist which make casting and filtering between
types easier.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -131,7 +131,8 @@ public abstract class Publisher<T> {
      * }</pre>
      * @param clazz The type to cast to.
      * @param <R> The resulting type of the cast operation.
-     * @return The cast of this {@link Publisher} to type {@link R}.
+     * @return The cast of this {@link Publisher} to type {@link R}. Terminate with a {@link ClassCastException} if
+     * signals cannot be cast to type {@link R}.
      * @see <a href="https://reactivex.io/documentation/operators/map.html">ReactiveX cast operator.</a>
      */
     public final <R> Publisher<R> cast(Class<R> clazz) {
@@ -161,7 +162,7 @@ public abstract class Publisher<T> {
     }
 
     /**
-     * Filters items so that only items of type {@link R} are emitted by the return value.
+     * Filters items so that only non-{@code null} items of type {@link R} are emitted by the return value.
      * <p>
      * This method provides a data transformation in sequential programming similar to:
      * <pre>{@code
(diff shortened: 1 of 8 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -118,6 +118,26 @@ protected Publisher() {
...
+     *         results.add(clazz.cast(t));
+     *     }
+     *     return results;
+     * }</pre>
+     * @param clazz The type to cast to.
+     * @param <R> The resulting type of the cast operation.
+     * @return The cast of this {@link Publisher} to type {@link R}.
+     * @see <a href="https://reactivex.io/documentation/operators/map.html">ReactiveX cast operator.</a>
+     */
+    public final <R> Publisher<R> cast(Class<R> clazz) {
```
Scottmitch, line 137: these operators are not impossible for users todo them selves but are commonly provided by other RS implementations and lowers the barrier to entry.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -134,13 +154,34 @@ protected Publisher() {
...
      * @return A {@link Publisher} that only emits the items that pass the {@code predicate}.
-     *
      * @see <a href="http://reactivex.io/documentation/operators/filter.html">ReactiveX filter operator.</a>
      */
     public final Publisher<T> filter(Predicate<? super T> predicate) {
         return filter(() -> predicate);
     }
 
+    /**
+     * Filters items so that only items of type {@link R} are emitted by the return value.
```
bondolo, line 164: …only non-null items of…

bondolo (commented):
I like these operators as using Java streams I frequently use the `filter(clazz::isInstance).map(clazz::cast)` pattern.

I wonder how we should document the effects of casting a bad item with `cast()` so that people can add their appropriate `onErrorResume` for the `ClassCastException`s which may result.

Also worth documenting that `ofType` will filter nulls.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherOfTypeTest.java
```diff
@@ -0,0 +1,59 @@
...
+        assertThat(subscriber.takeOnNext(), is(1));
+        source.onComplete();
+        subscriber.awaitOnComplete();
+    }
+
+    @Test
+    void noSignals() {
+        toSource(source.ofType(Integer.class)).subscribe(subscriber);
+        subscriber.awaitSubscription().request(2);
+        source.onNext("NotInteger1", "NotInteger2");
```
idelpivnitskiy, line 55: Consider adding a test case with `null` item

idelpivnitskiy (approved):
LGTM after @bondolo comments addressed

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/2036

bondolo (approved):

PR #2044, opened 2022-01-10
`RoundRobinLoadBalancer`: make `targetResource` logging less confusing

idelpivnitskiy:
Motivation:

Current version of `RoundRobinLoadBalancerFactory` adds `#incr_number`
to the `targetResource` as a way to distinguish two different RRLB
instances for the same `targetResource`. It results in something like
`localhost:8080#1` in logs, which is confusing for users because they
can think there is a typo in the address name.

Modifications:

- Remove `RoundRobinLoadBalancerFactory#FACTORY_COUNT`;
- Append `RoundRobinLoadBalancer#hashCode()` value to the passed
`targetResource`;

Result:

Less confusing address in logs:
`localhost:8080#1` -> `localhost:8080 (instance @cafebabe)`.


chemicL (approved):
🚢

PR #2048, opened 2022-01-10
Update proto-google-common-protos 2.7.0 -> 2.7.1

idelpivnitskiy:
~Depends on~ Merge after #2047 to avoid git conflicts


tkountis (approved):

Scottmitch:
why is there a dependency? will this break others if they don't update together? if any restrictions lets call them out in the release notes.

idelpivnitskiy:
@Scottmitch nvm, it's only because of the git history, if I open these PRs concurrently I will have to deal with merge conflicts

PR #2050, opened 2022-01-11
Make exception mapper filters public

idelpivnitskiy:
Motivation:

Some users want to map exceptions earlier in the request processing
pipeline in order to observe a real HTTP status code in other filters.

Modifications:

- Move `DefaultHttpServerBuilder.ExceptionMapperServiceFilter` ->
`io.servicetalk.http.api.HttpExceptionMapperServiceFilter`;
- Move `io.servicetalk.grpc.netty.CatchAllHttpServiceFilter` ->
`io.servicetalk.grpc.api.GrpcExceptionMapperServiceFilter`;
- Rename `CatchAllHttpServiceFilterTest` ->
`GrpcExceptionMapperServiceFilterTest`;
- Remove `io.servicetalk.grpc.netty.GrpcUtils` duplicate;

Result:

Users have access to our exception mapper implementations and can move
them around other filters in the processing pipeline.

PR #2054, opened 2022-01-12
`GrpcExceptionMapperServiceFilter`: set grpc-status if response payload fails

idelpivnitskiy:
Motivation:

Existing version of `GrpcExceptionMapperServiceFilter` does not set
`grpc-status` if an error happens during response payload body
processing.

Modifications:

- Apply `GrpcStatusUpdater` when `GrpcExceptionMapperServiceFilter` gets
a response;

Result:

If any filter throws an error during payload body processing,
`grpc-status` will be set correctly.

Depends on #2050.


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExceptionMapperServiceFilter.java
```diff
@@ -0,0 +1,141 @@
...
+
+            @Override
+            public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
+                                                        final StreamingHttpRequest request,
+                                                        final StreamingHttpResponseFactory responseFactory) {
+                final BufferAllocator allocator = ctx.executionContext().bufferAllocator();
+                final Single<StreamingHttpResponse> handle;
+                try {
+                    handle = delegate().handle(ctx, request, responseFactory).map(response -> {
+                        // FIXME: make this logic optional for users who want to be able to respond for HTTP requests
```
idelpivnitskiy, line 75: Actually, this change improves one thing (missed `grpc-status` if a filter throws) but breaks another (users won't be able to respond to HTTP requests in a filter). I will defer integrating this PR until we have a way to modify filters chain.

PR #2055, opened 2022-01-13
Add more logging for `GrpcRouter` and other transport classes

idelpivnitskiy:
Motivation:

Better visibility of the request-response processing.

Modifications:

- Add debug level logging when route fails with an error in `GrpcRouter`;
- Add debug level logging when `GrpcStatusUpdater` converts an exception
into a `grpc-status` code;
- Add more debug logging in `NettyChannelPublisher` and
`WriteStreamSubscriber`;

Result:

Easier debugging.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
@@ -506,7 +506,9 @@ final class WriteStreamSubscriber implements PublisherSource.Subscriber<Object>,
 
         private void terminateSubscriber(@Nullable Throwable cause) {
             if (cause == null) {
-                LOGGER.debug("{} Terminate subscriber, state: {}", channel, Integer.toString(state, 2));
+                if (LOGGER.isDebugEnabled()) {
+                    LOGGER.debug("{} Terminate subscriber, state: {}", channel, Integer.toString(state, 2));
+                }
                 try {
                     observer.writeComplete();
                     subscriber.onComplete();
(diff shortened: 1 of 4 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -519,6 +520,8 @@ private void terminateSubscriber(@Nullable Throwable cause) {
                 Throwable enrichedCause = enrichProtocolError.apply(cause);
                 assignConnectionError(channel, enrichedCause);
                 enrichedCause = !written ? new AbortedFirstWriteException(enrichedCause) : enrichedCause;
+                LOGGER.debug("{} Terminate subscriber with an error, state: {}", channel, Integer.toString(state, 2),
```
Scottmitch, line 523: error termination isn't expected to be common, but do we want todo `Integer.toString(state, 2)` unconditionally?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -506,6 +506,7 @@ private boolean nettySharedPromiseTryStatus() {
 
         private void terminateSubscriber(@Nullable Throwable cause) {
             if (cause == null) {
+                LOGGER.debug("{} Terminate subscriber, state: {}", channel, Integer.toString(state, 2));
```
Scottmitch, line 509: can we avoid `Integer.toString(state, 2)` unconditionally?

Scottmitch (approved):

bondolo (approved):

PR #2057, opened 2022-01-13
Update grpc example pom.xml

Scottmitch:
Motivation:
The pom.xml is not using servicetalk-bom and is
also missing a required dependecy.


Scottmitch:
discovered while testing https://github.com/servicetalk/examples

bondolo (approved):

PR #2058, opened 2022-01-14
Default `requiredOffloads` use `influenceStrategy`

bondolo:
Motivation:
The current default implementation of `requiredOffloads` does not
consider the case where `influenceStrategy` is implemented.
Modifications:
Call `influenceStrategy` within default `requiredOffloads` to determine
needed offloads. Default `influenceStrategy` may not call default
`requiredOffloads` to avoid infinite recursion.
Result:
Legacy influencers which do not yet implement `requiredOffloads` will
still be able to influence execution strategy.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
@@ -18,6 +18,8 @@ package io.servicetalk.http.api;
 import io.servicetalk.transport.api.ExecutionStrategyInfluencer;
 
 import static io.servicetalk.http.api.DefaultStreamingStrategyInfluencer.DEFAULT_STREAMING_STRATEGY_INFLUENCER;
+import static io.servicetalk.http.api.HttpExecutionStrategies.defaultStrategy;
+import static io.servicetalk.http.api.HttpExecutionStrategies.offloadAll;
 
 /**
  * An entity that wishes to influence {@link HttpExecutionStrategy} for an HTTP client or server.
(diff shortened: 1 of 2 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
```diff
@@ -47,7 +56,8 @@ default HttpExecutionStrategy influenceStrategy(HttpExecutionStrategy strategy)
     @Override
     default HttpExecutionStrategy requiredOffloads() {
         // safe default--implementations are expected to override
-        return HttpExecutionStrategies.offloadAll();
+        HttpExecutionStrategy result = influenceStrategy(defaultStrategy());
+        return defaultStrategy() == result ? offloadAll() : result;
```
idelpivnitskiy, line 60: 0.41.x users who want to say "my filter does not block" implement `influenceStrategy` this way:
```java

    @Override
    public HttpExecutionStrategy influenceStrategy(HttpExecutionStrategy strategy) {
        // No influence since we do not block.
        return strategy;
    }
```
it will always return `defaultStrategy()` and result in `offloadAll()` instead of `offloadNone()`

bondolo: You are correct. I've addressed this issue and added two more test cases.

idelpivnitskiy (approved):
Thank you!

bondolo:
I thought up an additional case that needed to be handled, if the default `requiredOffloads` method was implemented by a sub-interface of `HttpExecutionStrategy`. The recursion-avoidance check now ensures that the default `requiredOffloads` method is defined on `HttpExecutionStrategy` before avoiding calling it.

bondolo:
**0.42** → 886cbcd43462901ed0b4bf1c83169be497032a4c

PR #2060, opened 2022-01-20
Fix jersey-jackson not considering deprecated JacksonSerializationProvider

tkountis:
**Motivation**:

Upon deprecation of `JacksonSerializationProvider` our `servicetalk-data-jackson-jersey` stopped evaluating custom object-mappers offered through the `ServiceTalkJacksonSerializerFeature#contextResolverFor(ObjectMapper)`.

**Modifications**:

Restore old execution path if `JacksonSerializationProvider` is found as a resolver during runtime, otherwise fallback to the new flow.
Order is "old custom Jackson (if found)" -> "new custom Jackson (if found)" -> "new Jackson (fallback)"

**Result**:

Offer solid migration path while the deprecated `JacksonSerializationProvider` is still around.

### DIFF
--- a/servicetalk-data-jackson-jersey/src/test/java/io/servicetalk/data/jackson/jersey/CustomJacksonSerializationFactoryTest.java
+++ b/servicetalk-data-jackson-jersey/src/test/java/io/servicetalk/data/jackson/jersey/CustomJacksonSerializationFactoryTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2022 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2018 Apple Inc. and the ServiceTalk project authors
  *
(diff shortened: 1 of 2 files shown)


servicetalk-data-jackson-jersey/src/test/java/io/servicetalk/data/jackson/jersey/CustomJacksonSerializationFactoryTest.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright © 2018 Apple Inc. and the ServiceTalk project authors
```
idelpivnitskiy, line 2: 2022

idelpivnitskiy (approved):
Looks reasonable, thank you!

PR #2062, opened 2022-01-22
Fix ClientClosureRaceTest bring back max-retries

tkountis:
Motivation:

During the retrying http rework, the filter changed from having by default infinite max-retries to have 4. Some places that were modified during the refectory with the first assumption, weren't reverted. This can result in rare test failures.

Modifications:

Restore max-retries on the retrying http builder for retryable and other exceptions.

Result:

Restore old behavior.

(cherry picked from commit 5e8d4828a64594cb69409f82d67fad1cfe98d398)


Scottmitch:
in the commit message can you clarify why it was originally removed (was it just a mistake, or was behavior assumed to change by using the new method, etc.) and why it is being added back now?

chemicL (approved):

PR #2065, opened 2022-01-26
Publisher#multicast option to not cancel upstream

Scottmitch:
Motivation:
Some multicast use cases do not want cancellation
from downstream Subscribers to generate a cancel
to the upstream source. For example if the
upstream source doesn't support re-subscribes
or if there are expected Subscribers after
cancellation.

Modifications:
- Add a `boolean cancelUpstream` parameter to the
  Publisher#multicast operator. This makes it so
  no cancel is propagated upstream and internal
  state of the operator is preserved.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -2408,7 +2408,8 @@ public abstract class Publisher<T> {
      * @param minSubscribers The upstream subscribe operation will not happen until after this many {@link Subscriber}
      * subscribe to the return value.
      * @param cancelUpstream {@code true} if upstream should be {@link Subscription#cancel() cancelled} when all
-     * downstream {@link Subscriber}s cancel.
+     * downstream {@link Subscriber}s cancel. {@code false} means that cancel will not be propagated upstream even if
+     * all downstream {@link Subscriber}s cancel, and the upstream Subscription will stay valid until termination.
      * @return a {@link Publisher} that subscribes a single time upstream but allows for multiple downstream
      * {@link Subscriber}s. Signals from upstream will be multicast to each downstream {@link Subscriber}.
      * @see <a href="http://reactivex.io/documentation/operators/publish.html">ReactiveX multicast operator</a>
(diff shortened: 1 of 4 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2356,7 +2356,7 @@ public final Completable flatMapCompletableDelayError(Function<? super T, ? exte
      */
     @Deprecated
     public final Publisher<T> multicastToExactly(int expectedSubscribers, int queueLimit) {
-        return new MulticastPublisher<>(this, expectedSubscribers, true, queueLimit, t -> completed());
+        return new MulticastPublisher<>(this, expectedSubscribers, true, true, queueLimit, t -> completed());
```
NiteshKant, line 2359: I know this would be a backward incompatible change but using `false` as default would be better. Typically, multicast is used for sources that do not otherwise support multiple subscribes (as our processors do), so inadvertently subscribing multiple times (due to the lifetime of downstream subscribers) to upstream produces hard to debug issues.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2384,7 +2384,37 @@ public final Completable flatMapCompletableDelayError(Function<? super T, ? exte
...
+     *     List<T> results = resultOfThisPublisher();
+     *     List<List<T>> multiResults = ...; // simulating multiple Subscribers
+     *     for (int i = 0; i < expectedSubscribers; ++i) {
+     *         multiResults.add(results);
+     *     }
+     *     return multiResults;
+     * }</pre>
+     * @param minSubscribers The upstream subscribe operation will not happen until after this many {@link Subscriber}
+     * subscribe to the return value.
+     * @param cancelUpstream {@code true} if upstream should be {@link Subscription#cancel() cancelled} when all
```
NiteshKant, line 2410: Maybe provide some information around what is the lifecycle when this is set to `false`. eg: "If set to {@code false}, upstream subscription is active till termination"

NiteshKant (approved):
Couple comments then LGTM

idelpivnitskiy (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
Scottmitch: I would prefer to keep the default the same in this PR. It isn't clear which default behavior is better... I can imagine not propagating cancellation resulting in a hang which would also be hard to debug. At least propagating the cancel will ensure definitive termination.

NiteshKant: That is fine; IMHO preserving lifecycle of upstream is an appropriate default behavior. Multicast is providing just a way to support multiple subscribers. If upstream is an infinite stream, it will stay infinite with multicast else it will terminate with upstream.

PR #2066, opened 2022-01-28
`IoExecutor` implements `concurrent.Executor`

bondolo:
Motivation:
Implementing `Executor` allows `IoExecutor` to used in existing
interfaces. Usage of `IoExecutor`s for `Executor` will still be very
limited for specialized cases.
Modifications:
`IoExecutor` extends `io.servicetalk.concurrent.Executor`. No user
implementations of `IoExecutor` likely exist and `NettyIoExecutor`
already implements the interface.
Result:
`IoExecutor` instances can now be passed in interfaces which currently
only accept `io.servicetalk.concurrent.Executor`.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
@@ -15,15 +15,15 @@
  */
 package io.servicetalk.transport.api;
 
+import io.servicetalk.concurrent.Executor;
 import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
 
-import java.util.concurrent.Executor;
 import java.util.function.BooleanSupplier;
 
 /**
  * {@link Executor} that handles IO.
  */
-public interface IoExecutor extends ListenableAsyncCloseable {
+public interface IoExecutor extends Executor, ListenableAsyncCloseable {
 
     /**
      * Determine if <a href="https://en.wikipedia.org/wiki/Unix_domain_socket">Unix Domain Sockets</a> are supported.
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
@@ -15,15 +15,17 @@
  */
 package io.servicetalk.transport.api;
 
+import io.servicetalk.concurrent.Cancellable;
 import io.servicetalk.concurrent.api.Executor;
-import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
 
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.function.BooleanSupplier;
 
 /**
  * {@link Executor} that handles IO.
  */
-public interface IoExecutor extends Executor, ListenableAsyncCloseable {
+public interface IoExecutor extends Executor {
 
     /**
      * Determine if <a href="https://en.wikipedia.org/wiki/Unix_domain_socket">Unix Domain Sockets</a> are supported.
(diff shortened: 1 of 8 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
```diff
@@ -15,15 +15,15 @@
...
 import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
 
-import java.util.concurrent.Executor;
 import java.util.function.BooleanSupplier;
 
 /**
  * {@link Executor} that handles IO.
  */
-public interface IoExecutor extends ListenableAsyncCloseable {
+public interface IoExecutor extends Executor, ListenableAsyncCloseable {
```
idelpivnitskiy, line 26: 1. We can probably remove `Executor` declaration from `AbstractNettyIoExecutor` now.
2. This is ok change for the main branch (0.43 version). If you are going to cherry pick this into 0.41 and 0.42, consider adding default implementations that throw `UnsupportedOperationException`. As you mentioned in the commit description, nobody will probably be broken either way. Having default impls will help to avoid any ABI issues.

bondolo: 1. `AbstractNettyIoExecutor` is implementing the `io.servicetalk.concurrent.api.Executor` sub-interface so still needs to declare it. For the same reason `NettyIoExecutor` still needs to provide the `asExecutor()` for returning the casted instance. We could have `NettyIoExecutor` extend `io.servicetalk.concurrent.api.Executor` perhaps.
2. Providing defaults makes sense for 0.42 but backporting to 0.41 would be too much work for the benefit I suspect.

idelpivnitskiy: 1. 👍 
2. The change of this PR is small and defaults will be the same for 0.42 and 0.41. Why do you expect 0.41 to be much different?

bondolo: > The change of this PR is small and defaults will be the same for 0.42 and 0.41. Why do you expect 0.41 to be much different?

Sorry, I was thinking of the follow-on bug that will use this new feature. I have no objections to back-porting this specific change to both 0.41 and 0.42

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
```diff
@@ -15,15 +15,15 @@
...
```
idelpivnitskiy, line 26: 1. `io.servicetalk.concurrent.api.Executor` already implements `io.servicetalk.concurrent.api.ListenableAsyncCloseable`, explicit declaration can be removed.
2. This PR deprecates `NettyIoExecutor#asExecutor()`, meaning we will anyway need a follow-up PR to remove this deprecation for 0.43. In this case, consider adding `default` implementations for `IoExecutor#execute(...)` and `IoExecutor#schedule(...)` that will throw `UnsupportedOperationException` with a comment `FIXME: 0.43 - remove default methods`. It will make this PR ready for cherry-picking.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyIoExecutor.java
```diff
@@ -31,6 +31,8 @@
      * Implementation of this method assumes there would be no blocking code inside the submitted {@link Runnable}s.
      * If this assumption is violated, it will impact EventLoop responsiveness and hence should be avoided.
      * @return an {@link Executor} which will use an {@link IoExecutor} thread for execution.
+     * @deprecated IoExecutor now implements {@link Executor} so this method is redundant.
      */
+    @Deprecated
```
idelpivnitskiy, line 36: Consider adding a comment like `FIXME: 0.43 - remove deprecated method`, it makes easier to search what have to be cleaned in the main branch later.

idelpivnitskiy (approved):
LGTM after the last comments addressed:

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/IoExecutor.java
```diff
@@ -60,4 +62,16 @@ default BooleanSupplier shouldOffloadSupplier() {
...
+    // FIXME: 0.43 - remove default method
+    @Override
+    default Cancellable execute(Runnable task) throws RejectedExecutionException {
+        throw new UnsupportedOperationException("No existing IoExecutor implementations require this default");
+    }
+
+    // FIXME: 0.43 - remove default method
+    @Override
+    default Cancellable schedule(Runnable task, long delay, TimeUnit unit) throws RejectedExecutionException {
+        throw new UnsupportedOperationException("No existing IoExecutor implementations require this default");
```
idelpivnitskiy, line 75: When we added new methods in 0.41, we usually returned the current class name in the message to let users know which implementation lacks the method. Example: 
https://github.com/apple/servicetalk/blob/100c58bb5098dbcc74ee04778facd68c3f05ebfb/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java#L177-L180

PR #2068, opened 2022-02-01
Use `setClassLevel(true)` on static instances

bondolo:
Motivation:
The `ExecutionContextExtension` testing extension initializes the
contained execution context once per test unless `setClassLevel(true)`
is called. Some of our tests use a static class-instance of
`ExecutionContextExtension` but fail to call `setClassLevel(true)`
which results in many hundreds of extra invocations of an expensive
operation.
Modifications:
Apply `setClassLevel(true)` to all static instances of
`ExecutionContextExtension`.
Result:
Faster test execution. Saves about 40 seconds for `./gradlew --no-build-cache clean build`
for me.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/Tls13Test.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/Tls13Test.java
@@ -56,10 +56,12 @@ class Tls13Test {
 
     @RegisterExtension
     static final ExecutionContextExtension SERVER_CTX =
-        ExecutionContextExtension.cached("server-io", "server-executor");
+        ExecutionContextExtension.cached("server-io", "server-executor")
+                .setClassLevel(true);
     @RegisterExtension
     static final ExecutionContextExtension CLIENT_CTX =
-        ExecutionContextExtension.cached("client-io", "client-executor");
+        ExecutionContextExtension.cached("client-io", "client-executor")
+                .setClassLevel(true);
 
     private static final String TLS1_3 = "TLSv1.3";
     private static final String TLS1_3_REQUIRED_CIPHER = "TLS_AES_128_GCM_SHA256";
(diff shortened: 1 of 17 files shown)


bondolo:
There are additional cases where the `ExecutionContextExtension` could perhaps be made `static` but I did not investigate these in this PR.

servicetalk-transport-netty-internal/src/testFixtures/java/io/servicetalk/transport/netty/internal/ExecutionContextExtension.java
```diff
@@ -116,6 +116,13 @@ public static ExecutionContextExtension single(NettyIoThreadFactory nettyIoThrea
         return fixed(1, nettyIoThreadFactory);
     }
 
+    /**
+     * Set to true if the extension is being shared among all tests in a class.
+     *
+     * @param classLevel true if extension is shared between tests within test class otherwise false to create a new
```
tkountis, line 122: consider rephrasing the last part of the sentence. as it is, it reads as if it creates instances itself. 
e.g. `true if extension is shared between tests within test class otherwise false to indicate that the extension's lifecycle should be follow per-test behavior`

tkountis (approved):

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/Tls13Test.java
```diff
@@ -56,10 +56,12 @@
...
     @RegisterExtension
     static final ExecutionContextExtension SERVER_CTX =
-        ExecutionContextExtension.cached("server-io", "server-executor");
+        ExecutionContextExtension.cached("server-io", "server-executor")
+                .setClassLevel(true);
     @RegisterExtension
     static final ExecutionContextExtension CLIENT_CTX =
-        ExecutionContextExtension.cached("client-io", "client-executor");
+        ExecutionContextExtension.cached("client-io", "client-executor")
+                .setClassLevel(true);
```
idelpivnitskiy, line 64: Great catch! jUnit 4.x were able to recognize if a similar extension was a static field or not, but in jUnit5 it works differently. Thank you!

idelpivnitskiy (approved):

PR #2071, opened 2022-02-01
Fix filters for MultiAddressUrlHttpClientBuilder

chemicL:
Motivation:

Recently the API of filters changed to align better with the way
`ExecutionStrategy` should be set. A temporary additional filter
bridging calls via new API to old API was created to assure compatibility.
Unfortunately, the `MultiAddressUrlHttpClientBuilder` was not updated
and legacy filters may not be exercised.

Modifications:

- Properly appending the `NewToDeprecatedFilter` to filters added via
  `MultiAddressUrlHttpClientBuilder`,
- Legacy `StrategyInfluencerAwareConversions#toMultiAddressClientFactory`
  was also updated to include the bridge,
- `MixedFiltersTest` include a case validating the
  `MultiAddressUrlHttpClientBuilder` behavior.

Result:

Deprecated methods on filters are properly executed when using
`MultiAddressUrlHttpClientBuilder`.


idelpivnitskiy (approved):
Thank you!

PR #2075, opened 2022-02-02
Implement `setClassLevel()` for `ExecutorExtension`

bondolo:
Motivation:
The `ExecutorExtension` is often used as a class static with the
assumption that the instance will only be created once per test class
and reused (shared) for all tests. This was not the case and the
contained Executor instance was being created for each test case.
Modifications:
Provide `setClassLevel()` method similar to `ExecutionContextExtension`
and call it with `true` for shared instances.
Result:
Reduced test overhead.


idelpivnitskiy (approved):
Thank you!

bondolo:
**0.42** → 5fe4903562ae98d2f884638bd99402edb776b17e

PR #2084, opened 2022-02-09
Use gRPC-Java BOM

bondolo:
Motivation:
The gRPC Java project provides a BOM file for managing versions of
modules and dependencies. Using the BOM results in more consistent
component resolution.
Modifications:
Use BOM in `build.graddle` as an enforcedPlatform. Remove version
references from most other components to use BOM version.
Result:
Cleaner use of gRPC-Java dependencies.


idelpivnitskiy:
Let's merge it after the release, better to give it some time to be verified using a snapshot

Scottmitch (approved):

idelpivnitskiy (approved):

PR #2070, opened 2022-02-01
Schedule/execute no offload timeouts on IoExecutor

bondolo:
Motivation:
Currently the timeout filter executes timeout events on the execution
context executor even if the execution strategy requires no offloading.
The IoExecutor should instead be use for these situations to reduce
offloads.
Modifications:
Request `ExecutionContext.ioExecutor()` as the `contextExecutor` if no
offloads are configured.
Result:
Less offloading for some timeout events.

PR #2087, opened 2022-02-09
Revert API change involving `Executor`

bondolo:
Motivation:
A source compatible API change was made in #2070 that replaced
`io.servicetalk.concurrent.api.Executor` with
`io.servicetalk.concurrentExecutor`. This change, while source
compatible as `io.servicetalk.concurrentExecutor` is the super-type of
`io.servicetalk.concurrent.api.Executor`, is not binary compatible.
Modifications:
Revert change to use same `Executor` class as prior releases on these
branches.
Result:
Binary compatibility is retained.


idelpivnitskiy (approved):
Thank yoU!

PR #2088, opened 2022-02-11
Changing instances of ExecutorExtension to static

gabAlek:
Motivation:

To potentially reduce the overhead of instantiating the ExecutorExtension class repeatedly between test cases.

Modifications:
Changed instances of ExecutorExtension to static

Result:

All instances of the ExecutorExtension are now static inside servicetalk-concurrent-api module test classes.

### DIFF
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/BufferStrategiesTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/BufferStrategiesTest.java
@@ -77,7 +77,7 @@ import static org.mockito.Mockito.when;
 class BufferStrategiesTest {
 
     @RegisterExtension
-    static final ExecutorExtension<Executor> EXECUTOR_EXTENSION = withCachedExecutor().setClassLevel(true);
+    static final ExecutorExtension<Executor> EXEC = withCachedExecutor().setClassLevel(true);
 
     @RegisterExtension
     final ExecutorExtension<TestExecutor> testExecutorExtension = withTestExecutor();
@@ -110,7 +110,7 @@ class BufferStrategiesTest {
                 .ignoreElements().subscribe();
         CyclicBarrier accumulateAndTimerStarted = new CyclicBarrier(2);
 
-        Future<Object> timersFuture = EXECUTOR_EXTENSION.executor().submit(() -> {
+        Future<Object> timersFuture = EXEC.executor().submit(() -> {
             accumulateAndTimerStarted.await();
             while (boundariesDone.getCount() > 0) {
                 timers.take().onComplete();
@@ -118,7 +118,7 @@ class BufferStrategiesTest {
             return null;
         }).toFuture();
 
-        Future<Object> accumulateFuture = EXECUTOR_EXTENSION.executor().submit(() -> {
+        Future<Object> accumulateFuture = EXEC.executor().submit(() -> {
             accumulateAndTimerStarted.await();
             int boundariesReceived = 0;
             Iterator<Integer> itemsToPopulate = items.iterator();
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/completable/CompletableToPublisherTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/completable/CompletableToPublisherTest.java
@@ -40,7 +40,7 @@ import static org.hamcrest.Matchers.sameInstance;
 
 class CompletableToPublisherTest {
     @RegisterExtension
-    final ExecutorExtension<Executor> executorExtension = ExecutorExtension.withCachedExecutor();
+    static final ExecutorExtension<Executor> EXEC = ExecutorExtension.withCachedExecutor().setClassLevel(true);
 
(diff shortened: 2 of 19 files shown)


bondolo:
Thank you! In addition to marking the instances `static` the PR should investigate which of the instances can be made truly static by marking them with `setClassLevel(true)`. At least some of the cases fail with a static executor. I had left the instances as non-`static` to highlight that they still needed investigation as to whether one instance could be used for all tests. Can you review the tests to see which still work once `setClassLevel(true)` is applied?

gabAlek:
> 



> Thank you! In addition to marking the instances `static` the PR should investigate which of the instances can be made truly static by marking them with `setClassLevel(true)`. At least some of the cases fail with a static executor. I had left the instances as non-`static` to highlight that they still needed investigation as to whether one instance could be used for all tests. Can you review the tests to see which still work once `setClassLevel(true)` is applied?

Sure, I'm on it!

gabAlek:
> Thank you! In addition to marking the instances `static` the PR should investigate which of the instances can be made truly static by marking them with `setClassLevel(true)`. At least some of the cases fail with a static executor. I had left the instances as non-`static` to highlight that they still needed investigation as to whether one instance could be used for all tests. Can you review the tests to see which still work once `setClassLevel(true)` is applied?

Commit updated

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/BufferStrategiesTest.java
```diff
@@ -80,7 +80,7 @@
     static final ExecutorExtension<Executor> EXECUTOR_EXTENSION = withCachedExecutor().setClassLevel(true);
 
     @RegisterExtension
-    final ExecutorExtension<TestExecutor> testExecutorExtension = withTestExecutor();
+    static final ExecutorExtension<TestExecutor> testExecutorExtension = withTestExecutor();
```
idelpivnitskiy, line 83: If we can not use `.setClassLevel(true)` in this case or any other, it's better to keep it as non-static variable. Otherwise, it's harder to understand the test flow and when this extension is re-initialized.

idelpivnitskiy (commented):
Thank you for the PR @gabAlek!
Please check build failures, there are some violations of checkstyle rules. You can test it locally with `./gradlew checkstyle` or `./gradlew build` before pushing an update.

gabAlek:
> Thank you for the PR @gabAlek! Please check build failures, there are some violations of checkstyle rules. You can test it locally with `./gradlew checkstyle` or `./gradlew build` before pushing an update.

Commit updated!Thank you for the suggestions, @idelpivnitskiy

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/completable/CompletableToPublisherTest.java
```diff
@@ -40,7 +40,8 @@
 
 class CompletableToPublisherTest {
     @RegisterExtension
-    final ExecutorExtension<Executor> executorExtension = ExecutorExtension.withCachedExecutor();
+    static final ExecutorExtension<Executor> executorExtension = ExecutorExtension.withCachedExecutor()
```
bondolo, line 43: Likely requires changing the name to `EXEC` or `EXECUTOR_EXTENSION` to satisfy CheckStyle. Even it does not the convention should be applied.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/completable/CompletableToSingleTest.java
```diff
@@ -34,7 +34,8 @@
 
 class CompletableToSingleTest {
     @RegisterExtension
-    final ExecutorExtension<Executor> executorExtension = ExecutorExtension.withCachedExecutor();
+    static final ExecutorExtension<Executor> executorExtension = ExecutorExtension.withCachedExecutor()
```
bondolo, line 37: Another `executorExtension` → `EXEC` case

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleToPublisherTest.java
```diff
@@ -45,7 +45,8 @@
 class SingleToPublisherTest {
 
     @RegisterExtension
-    final ExecutorExtension<Executor> executorExtension = ExecutorExtension.withCachedExecutor();
+    static final ExecutorExtension<Executor> executorExtension = ExecutorExtension.withCachedExecutor()
```
bondolo, line 48: Another `executorExtension` → `EXEC` case

bondolo (commented):
Excellent work! This is getting very close to completion. Thank you!

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/completable/CompletableToPublisherTest.java
gabAlek: Decided for changing it all to EXEC, to make it shorter

bondolo (approved):

bondolo:
Thank you Gabriel!

PR #2090, opened 2022-02-12
Add Single.repeatWhen(BiIntFunction<T, Completable>)

Scottmitch:
Motivation:
The criteria to repeat a Single may depend upon the
termination value. Existing repeat operators in
Single do not expose the value.

Modifications:
- Add Single.repeatWhen(BiIntFunction<T, Completable>)
  operator.
- Change existing Single.repeat* operators to use this
  new implementation instead of converting to Publisher.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RepeatWhenSingle.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RepeatWhenSingle.java
@@ -35,11 +35,11 @@ import static java.util.Objects.requireNonNull;
 final class RepeatWhenSingle<T> extends AbstractNoHandleSubscribePublisher<T> {
     static final Exception END_REPEAT_EXCEPTION = unknownStackTrace(new Exception(), RepeatWhenSingle.class, "<init>");
     private final Single<T> original;
-    private final BiIntFunction<? super T, ? extends Completable> shouldRedo;
+    private final BiIntFunction<? super T, ? extends Completable> repeater;
 
-    RepeatWhenSingle(final Single<T> original, final BiIntFunction<? super T, ? extends Completable> shouldRedo) {
+    RepeatWhenSingle(final Single<T> original, final BiIntFunction<? super T, ? extends Completable> repeater) {
         this.original = original;
-        this.shouldRedo = requireNonNull(shouldRedo);
+        this.repeater = requireNonNull(repeater);
     }
 
     @Override
@@ -64,6 +64,7 @@ final class RepeatWhenSingle<T> extends AbstractNoHandleSubscribePublisher<T> {
         private final Subscriber<? super T> subscriber;
         private final ContextMap contextMap;
         private final AsyncContextProvider contextProvider;
+        private final RepeatSubscriber repeatSubscriber = new RepeatSubscriber();
         private volatile long outstandingDemand;
         private int redoCount;
 
@@ -81,7 +82,7 @@ final class RepeatWhenSingle<T> extends AbstractNoHandleSubscribePublisher<T> {
                 final long prev = outstandingDemandUpdater.getAndAccumulate(this, n,
                         FlowControlUtils::addWithOverflowProtectionIfNotNegative);
                 if (prev == 0) {
-                    outer.original.delegateSubscribe(new RedoSubscriber(), contextMap, contextProvider);
+                    outer.original.delegateSubscribe(repeatSubscriber, contextMap, contextProvider);
                 }
             } else {
                 requestNInvalid(n);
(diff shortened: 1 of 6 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RepeatWhenSingle.java
```diff
@@ -0,0 +1,192 @@
...
+            public void onSubscribe(final Cancellable cancellable) {
+                sequentialCancellable.nextCancellable(cancellable);
+            }
+
+            @Override
+            public void onSuccess(@Nullable final T result) {
+                final Completable completable;
+                try {
+                    subscriber.onNext(result);
+                    completable = requireNonNull(outer.repeater.apply(++redoCount, result));
```
idelpivnitskiy, line 172: nit: `redoCount` looks like a leftover from the publisher impl of the same operator, consider `repeatCount`

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/RepeatWhenSingleTest.java
```diff
@@ -0,0 +1,272 @@
...
+        assertThat(subscriber.takeOnNext(), equalTo(1));
+        s.request(-1);
+        assertThat(subscriber.awaitOnError(), instanceOf(IllegalArgumentException.class));
+    }
+
+    @Test
+    void singleRepeatAllDemandUpfront() {
+        AtomicInteger value = new AtomicInteger();
+        toSource(Single.defer(() -> Single.succeeded(value.incrementAndGet()))
+                .repeatWhen((repeatCount, v) -> repeatCount == 1 ? completed() : failed(new DeliberateException())))
```
idelpivnitskiy, line 136: In case users will look at these tests as an example, consider using constant `DELIBERATE_EXCEPTION` instead. Allocationg a new exception is costly if they follow the same pattern.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/RepeatWhenSingleTest.java
```diff
@@ -0,0 +1,272 @@
...
+        s.request(2);
+        assertThat(subscriber.takeOnNext(), equalTo(1));
+        assertThat(subscriber.pollOnNext(10, MILLISECONDS), nullValue());
+        latch.countDown();
+        assertThat(subscriber.takeOnNext(), equalTo(2));
+        subscriber.awaitOnComplete();
+    }
+
+    @Test
+    void singleDelayedRepeat() {
```
idelpivnitskiy, line 179: This test is the same as `singleRepeatAllDemandUpfront`

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/RepeatWhenSingleTest.java
```diff
@@ -0,0 +1,272 @@
...
+        AtomicInteger value = new AtomicInteger();
+        toSource(Single.defer(() -> Single.succeeded(value.incrementAndGet()))
+                .repeatWhen((repeatCount, v) -> repeatCount == 1 ? completed() : failed(new DeliberateException())))
+                .subscribe(subscriber);
+        Subscription s = subscriber.awaitSubscription();
+        s.request(2);
```
idelpivnitskiy, line 185: Consider also testing when `s.request(1)` is made multiple times in a row and from inside `onNext`.

idelpivnitskiy (approved):
LGTM after comments addressed:

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -988,7 +990,43 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
...
+     * @param repeatWhen {@link BiIntFunction} that given the repeat count and value from the current iteration returns
+     * a {@link Completable}. If this {@link Completable} emits an error repeat is terminated, otherwise,
+     * original {@link Single} is re-subscribed when this {@link Completable} completes.
+     *
+     * @return A {@link Publisher} that emits all items from this {@link Single} and from all re-subscriptions whenever
+     * the operation is repeated.
+     *
+     * @see <a href="http://reactivex.io/documentation/operators/retry.html">ReactiveX retry operator.</a>
+     */
+    public final Publisher<T> repeatWhen(BiIntFunction<? super T, ? extends Completable> repeatWhen) {
```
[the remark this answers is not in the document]
Scottmitch: @bondolo  - `Completable` lets you control when to do the repeat asynchronously. See [RepeatStrategies](https://github.com/apple/servicetalk/blob/main/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RepeatStrategies.java) for `repeat*` strategies and [RetryStrategies](https://github.com/apple/servicetalk/blob/main/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RetryStrategies.java) for `repeat*` strategies.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/RepeatWhenSingleTest.java
Scottmitch: sounds good, just want to differentiate for tests that actually expect `DELIBERATE_EXCEPTION` to propagate, but that set of tests is smaller so I can reduce allocation.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/RepeatWhenSingleTest.java
Scottmitch: good catch, need to actually delay the repeat :)

PR #2092, opened 2022-02-14
Always encode spans as list in `HttpReporter`

chemicL:
Motivation:

Zipkin Collector interface accepts spans always as a list using HTTP or
sequence for gRPC:

- https://zipkin.io/zipkin-api/#/default/post_spans
- https://github.com/openzipkin/zipkin-api/blob/master/zipkin.proto#L232

`HttpReporter` reports batched spans as a JSON list or a sequence with
gRPC. Unfortunately, `HttpReporter` behaves differently when batching is
not enabled and submits the encoded span without the enclosing
collection.

Modifications:

- Always encode a `List<Span>` in the `HttpReporter`,
- Removed test alternatives for single value reports and always
  validating a collection is reported.

Result:

The OpenTracing Zipkin Publisher is conformant to the Collector API.

### DIFF
--- a/servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
+++ b/servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
@@ -38,6 +38,7 @@ import zipkin2.reporter.Reporter;
 
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.function.Function;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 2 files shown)


bondolo:
Is there the possibility of services which depend up on the incorrect behaviour? Or was there just nobody using it in un-batched mode so the problem was not noticed?

Scottmitch:
If we are not following the spec and/or common conventions I think we can correct the behavior and call this out in release notes. We can consider a conditional system property to enable old behavior for some time if necessary (until they also have time to follow conventions). However if you have to accept a list, it seems unlikely an implementation would consider a list of 1 element an error.

chemicL:
My understanding is that accepting 1 item in a sequence is an expected behavior of a collector. I will merge this and we can explain the change in the release docs like @Scottmitch suggests. If there really is a collector that does not support a sequence of one item (or sequences in general) but expects items to be published via the non-compliant API we can introduce a setting for the old functionality upon request. Still, that's handling a broken implementation, it should not be a major concern in my opinion.

PR #2094, opened 2022-02-15
Revert breaking changes for `TimeoutFromRequest` introduced in #1832

idelpivnitskiy:
Motivation:

In #1972 we deprecated `TimeoutFromRequest` in favor on `BiFunction`.
Because users have to avoid using `TimeoutFromRequest` anyway, there is
no need to require them to deal with breaking changes when they migrate
from 0.41.X to 0.42.X.

Modifications:

- `TimeoutFromRequest` implements `HttpExecutionStrategyInfluencer`
again and provides a default implementation of `influenceStrategy`
identical to 0.41 branch;

Result:

Users who upgrade from 0.41.X to 0.42.X don't have to resolve
incompatible breaking API changes for `TimeoutFromRequest`. Instead,
they can migrate from `TimeoutFromRequest` to `BiFunction` after they
are already on 0.42.X release.

### DIFF
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
@@ -16,8 +16,8 @@
 package io.servicetalk.http.utils;
 
 import io.servicetalk.concurrent.TimeSource;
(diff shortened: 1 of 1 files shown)


bondolo (approved):

chemicL:
Late LGTM

PR #2095, opened 2022-02-16
Add client `RetryingHttpRequesterFilter` example

bondolo:
Motivation:
Users have requested a usage sample for client retry filter
Modifications:
Adds an HTTP example for the client `RetryingHttpRequesterFilter` retry
filter.
Result:
Additional usage example available for users.

### DIFF
--- a/servicetalk-examples/http/retry/src/main/java/io/servicetalk/examples/http/retry/RetryClient.java
+++ b/servicetalk-examples/http/retry/src/main/java/io/servicetalk/examples/http/retry/RetryClient.java
@@ -16,29 +16,27 @@
 package io.servicetalk.examples.http.retry;
 
 import io.servicetalk.http.api.HttpClient;
-import io.servicetalk.http.api.HttpResponseStatus;
 import io.servicetalk.http.netty.HttpClients;
 import io.servicetalk.http.netty.RetryingHttpRequesterFilter;
 
+import static io.servicetalk.http.api.HttpResponseStatus.StatusClass.SERVER_ERROR_5XX;
 import static io.servicetalk.http.api.HttpSerializers.textSerializerUtf8;
 
 /**
- * Extends the Async "Hello World" client to immediately retry requests that get a 5XX response. Two attempts will be
- * made before failure.
+ * Extends the Async "Hello World" client to immediately retry requests that get a 5XX response. Up to three attempts
+ * total, one initial attempt and up to two retries, will be made before failure.
  */
 public final class RetryClient {
     public static void main(String[] args) throws Exception {
         try (HttpClient client = HttpClients.forSingleAddress("localhost", 8080)
                 .appendClientFilter(new RetryingHttpRequesterFilter.Builder()
-                        .responseMapper((httpResponseMetaData -> {
-                            // If the response status is 500-599 then request a retry
-                            HttpResponseStatus.StatusClass status = httpResponseMetaData.status().statusClass();
-                            return HttpResponseStatus.StatusClass.SERVER_ERROR_5XX == status ?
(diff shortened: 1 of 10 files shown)


servicetalk-examples/http/retry/src/main/java/io/servicetalk/examples/http/retry/RetryClient.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.servicetalk.http.api.HttpClient;
+import io.servicetalk.http.api.HttpResponseStatus;
+import io.servicetalk.http.netty.HttpClients;
+import io.servicetalk.http.netty.RetryingHttpRequesterFilter;
+
+import static io.servicetalk.http.api.HttpSerializers.textSerializerUtf8;
+
+/**
+ * Extends the Async "Hello World" client to immediately retry requests that get a 5XX response. Two attempts will be
```
tkountis, line 26: "Two attempts" --> "up-to two attempts"

servicetalk-examples/http/retry/src/main/java/io/servicetalk/examples/http/retry/RetryClient.java
```diff
@@ -0,0 +1,56 @@
...
+ * made before failure.
+ */
+public final class RetryClient {
+    public static void main(String[] args) throws Exception {
+        try (HttpClient client = HttpClients.forSingleAddress("localhost", 8080)
+                .appendClientFilter(new RetryingHttpRequesterFilter.Builder()
+                        .responseMapper((httpResponseMetaData -> {
+                            // If the response status is 500-599 then request a retry
+                            HttpResponseStatus.StatusClass status = httpResponseMetaData.status().statusClass();
+                            return HttpResponseStatus.StatusClass.SERVER_ERROR_5XX == status ?
```
tkountis, line 36: consider `SERVER_ERROR_5XX.contains(meta.status())`
also consider static imports in this area, all these inner classes create unnecessarily code noise.

servicetalk-examples/http/retry/src/main/java/io/servicetalk/examples/http/retry/RetryClient.java
bondolo: I was definitely unsure whether there would be two attempts or two retries, additional attempts following an initial failure. I will correct this usage but I did find it ambiguous in the API docs as to the total number of attempts that would be made, often due to whether the first attempt counted as a retry or not.

tkountis (approved):
minor comments then lgtm

servicetalk-examples/http/retry/src/main/java/io/servicetalk/examples/http/retry/RetryClient.java
tkountis: good point, its up-to two retries, excluding the initial request.

servicetalk-examples/http/retry/src/main/java/io/servicetalk/examples/http/retry/RetryClient.java
```diff
@@ -0,0 +1,56 @@
...
+                                    new RetryingHttpRequesterFilter.HttpResponseException(
+                                            "Retry 5XX", httpResponseMetaData)
+                                    // Not a 5XX response, someone else can deal with it.
```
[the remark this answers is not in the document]
bondolo: My version of the comment is probably too informal, but the mapper is deciding less than "no need to retry" as other facilities could decide that retry is needed. It is an ambiguity of "do not retry" vs "I was unable to determine if retry is needed."

PR #2097, opened 2022-02-18
CompositeCloseable: prepend doesn't respect ordering

idelpivnitskiy:
Motivation:

`DefaultCompositeCloseable` does not respect ordering if `prepend`
operation is invoked after `append`. It causes issues when the ordering
matters for closeable resources that depend upon each other.

Modifications:

- Change `Operand.closables` to `ArrayDeque`;
- Use `addFirst` for `prepend` operation;
- Test ordiring for "prepend after append" and "prepend after merge";

Result:

Correct ordering of invoking `AsyncCloseable`s if `prepend` is used
after `append`.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
@@ -202,7 +202,7 @@ final class DefaultCompositeCloseable implements CompositeCloseable {
     }
 
     private static final class Operand {
-        final Deque<AsyncCloseable> closables = new ArrayDeque<>(2);
+        final Deque<AsyncCloseable> closables = new ArrayDeque<>(4);
         final boolean isMerge;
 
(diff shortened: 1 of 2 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultCompositeCloseable.java
```diff
@@ -203,8 +202,8 @@ private Completable buildCompletable(Function<AsyncCloseable, Completable> close
     }
 
     private static final class Operand {
-        private final List<AsyncCloseable> closables = new ArrayList<>(4);
-        private final boolean isMerge;
+        final Deque<AsyncCloseable> closables = new ArrayDeque<>(2);
```
bondolo, line 205: The default size of the collection changes. There is probably no reason not to still use 4 as that still typically fits within the alignment size of a heap allocation of a reference array. ie. The first resize to 4 will probably just reallocate to another array with less wasted space.

bondolo (approved):

Scottmitch (approved):
good catch!

PR #2098, opened 2022-02-19
Introduce public API FilterableReservedControlledStreamingHttpLoadBalancedConnection

tkountis:
Motivation:

At the lack of extra context available at connection creation time, it's useful to
allow for extentibility of LoadBalancedStreamingHttpConnection (implementation) to support extension points to mimic/wrap it.

Modifications:

Introduced `FilterableReservedControlledStreamingHttpLoadBalancedConnection` that `LoadBalancedStreamingHttpConnection` now implements, along with internal classes moved to support this new public API..

Result:

Netty LB connection has a public interface now, allowing additional override/delegate behavior in various flows.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancedStreamingHttpConnection.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancedStreamingHttpConnection.java
@@ -16,7 +16,7 @@
 package io.servicetalk.http.netty;
 
 import io.servicetalk.client.api.LoadBalancer;
-import io.servicetalk.client.api.internal.ReservableRequestConcurrencyController;
+import io.servicetalk.client.api.ReservableRequestConcurrencyController;
 import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.Publisher;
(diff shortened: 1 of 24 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancedStreamingHttpConnection.java
```diff
@@ -44,8 +44,9 @@
 /**
  * Makes the wrapped {@link StreamingHttpConnection} aware of the {@link LoadBalancer}.
  */
-final class LoadBalancedStreamingHttpConnection implements FilterableStreamingHttpLoadBalancedConnection,
-                   ReservedStreamingHttpConnection, ReservableRequestConcurrencyController,
+public class LoadBalancedStreamingHttpConnection
```
idelpivnitskiy, line 47: Instead of opening an implementation class with extra methods (like `unwrap/limiter`), have you considered adding `io.servicetalk.http.api.LoadBalancedStreamingHttpConnection` interface just to define a contract for the load-balanced connection. Implementations will be free to do whatever if they follow the contract without tightening to our impl. class.

tkountis:
This proved as not needed for now.

PR #2100, opened 2022-02-20
Add PayloadSizeLimitingHttpRequesterFilter

Scottmitch:
Motivation:
In some cases it is useful to be able to limit the
response payload size. For example when using aggregated
APIs the user may not want to allow larger response
payloads.

Modifications:
- Add PayloadSizeLimitingHttpRequesterFilter which applies
  a limit to the response payload size.

### DIFF
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
@@ -15,12 +15,12 @@
  */
 package io.servicetalk.http.utils;
 
+import io.servicetalk.buffer.api.Buffer;
 import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.http.api.FilterableStreamingHttpClient;
 import io.servicetalk.http.api.FilterableStreamingHttpConnection;
 import io.servicetalk.http.api.HttpExecutionStrategy;
-import io.servicetalk.http.api.HttpExecutionStrategyInfluencer;
 import io.servicetalk.http.api.StreamingHttpClientFilter;
 import io.servicetalk.http.api.StreamingHttpClientFilterFactory;
 import io.servicetalk.http.api.StreamingHttpConnectionFilter;
@@ -30,16 +30,16 @@ import io.servicetalk.http.api.StreamingHttpRequester;
 import io.servicetalk.http.api.StreamingHttpResponse;
 
 import java.util.function.Function;
+import java.util.function.UnaryOperator;
 
 import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
 
 /**
- * Limits the response payload size for a request. The filter will throw an exception which may result in
- * stream/connection closure.
+ * Limits the response payload size. The filter will throw an exception which may result in stream/connection closure.
+ * A {@link PayloadTooLargeException} will be thrown when the maximum payload size is exceeded.
  */
 public final class PayloadSizeLimitingHttpRequesterFilter implements
-                        StreamingHttpClientFilterFactory, StreamingHttpConnectionFilterFactory,
-                        HttpExecutionStrategyInfluencer {
+                        StreamingHttpClientFilterFactory, StreamingHttpConnectionFilterFactory {
     private final int maxResponsePayloadSize;
 
     /**
@@ -48,7 +48,7 @@ public final class PayloadSizeLimitingHttpRequesterFilter implements
      */
     public PayloadSizeLimitingHttpRequesterFilter(int maxResponsePayloadSize) {
         if (maxResponsePayloadSize < 0) {
-            throw new IllegalArgumentException("maxResponsePayloadSize: " + maxResponsePayloadSize + " (expected >0)");
+            throw new IllegalArgumentException("maxResponsePayloadSize: " + maxResponsePayloadSize + " (expected >=0)");
         }
         this.maxResponsePayloadSize = maxResponsePayloadSize;
     }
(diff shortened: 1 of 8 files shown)


servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
```diff
@@ -0,0 +1,109 @@
...
+                        StreamingHttpClientFilterFactory, StreamingHttpConnectionFilterFactory,
+                        HttpExecutionStrategyInfluencer {
+    private final int maxResponsePayloadSize;
+
+    /**
+     * Create a new instance.
+     * @param maxResponsePayloadSize The maximum response payload size allowed.
+     */
+    public PayloadSizeLimitingHttpRequesterFilter(int maxResponsePayloadSize) {
+        if (maxResponsePayloadSize < 0) {
```
tkountis, line 50: `<=`

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
```diff
@@ -0,0 +1,109 @@
...
+                            throw new PayloadTooLongException("Maximum payload size=" + maxResponsePayloadSize +
+                                    " current payload size=" + responsePayloadSize.value +
+                                    " new buffer size=" + buff.readableBytes());
+                        }
+                        responsePayloadSize.value += buff.readableBytes();
+                    }).shareContextOnSubscribe();
+                })));
+    }
+
+    static final class PayloadTooLongException extends IllegalStateException {
```
tkountis, line 98: do we need the stack-trace here?

tkountis (approved):
looks good.
did you consider a canned "413 Payload Too Large" response for the aggregated use-cases?

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
```diff
@@ -0,0 +1,109 @@
...
+
+import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
+
+/**
+ * Limits the response payload size for a request. The filter will throw an exception which may result in
+ * stream/connection closure.
+ */
+public final class PayloadSizeLimitingHttpRequesterFilter implements
+                        StreamingHttpClientFilterFactory, StreamingHttpConnectionFilterFactory,
+                        HttpExecutionStrategyInfluencer {
```
idelpivnitskiy, line 42: No need to add this interface explicitly starting from 0.42, `StreamingHttpClientFilterFactory` and `StreamingHttpConnectionFilterFactory` already extend it. It's enough just to override `requireOffloads()` method.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
```diff
@@ -0,0 +1,109 @@
...
+import io.servicetalk.http.api.StreamingHttpRequester;
+import io.servicetalk.http.api.StreamingHttpResponse;
+
+import java.util.function.Function;
+
+import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
+
+/**
+ * Limits the response payload size for a request. The filter will throw an exception which may result in
+ * stream/connection closure.
```
idelpivnitskiy, line 38: Consider clarifying what exception type or subtype is thrown, users likely need to write a logic that handles it.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
```diff
@@ -0,0 +1,109 @@
...
+                        HttpExecutionStrategyInfluencer {
+    private final int maxResponsePayloadSize;
+
+    /**
+     * Create a new instance.
+     * @param maxResponsePayloadSize The maximum response payload size allowed.
+     */
+    public PayloadSizeLimitingHttpRequesterFilter(int maxResponsePayloadSize) {
+        if (maxResponsePayloadSize < 0) {
+            throw new IllegalArgumentException("maxResponsePayloadSize: " + maxResponsePayloadSize + " (expected >0)");
```
idelpivnitskiy, line 51: Some users may want it to be 0 to accept only messages with empty body.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @Override
+    public HttpExecutionStrategy requiredOffloads() {
+        return offloadNone();
+    }
+
+    private Single<StreamingHttpResponse> applyLimit(
+            StreamingHttpRequest request, Function<StreamingHttpRequest, Single<StreamingHttpResponse>> delegator) {
+        return delegator.apply(request).map(response -> response.transformPayloadBody(publisher ->
```
idelpivnitskiy, line 84: Have you considered using `transformMessageBody`? While it's more advanced, it will result in less overhead and won't modify `PayloadInfo` flags. For a filter that we provide, it's justified to use more advanced transformation method.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
idelpivnitskiy: From javadoc for `IllegalStateException`:
> Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation.

Hard to say that `PayloadTooLongException` is about invoking at illegal or inappropriate time. Have you considered `SerializationException` or `IllegalArgumentException` instead?

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.util.function.Function;
+
+import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
+
+/**
+ * Limits the response payload size for a request. The filter will throw an exception which may result in
+ * stream/connection closure.
+ */
+public final class PayloadSizeLimitingHttpRequesterFilter implements
```
idelpivnitskiy, line 40: I filter like this will provide great value on the server-side. While you are here, consider adding `PayloadSizeLimitingHttpServiceFilter`. On the server side, +1 to @tkountis comment, we may need a ctor option or a public exception type to let users map it to 413.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
```diff
@@ -0,0 +1,109 @@
...
+import io.servicetalk.http.api.StreamingHttpRequest;
+import io.servicetalk.http.api.StreamingHttpRequester;
+import io.servicetalk.http.api.StreamingHttpResponse;
+
+import java.util.function.Function;
+
+import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
+
+/**
+ * Limits the response payload size for a request. The filter will throw an exception which may result in
```
idelpivnitskiy, line 37: request -> response.
On the client-side, it won't result in stream/connection closure for the client, but may affect a server-side connection as a side-effect if an exception is not handled properly.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
Scottmitch: Sure I can do this.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
Scottmitch: I will extend from `RuntimeException` for now

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
Scottmitch: Someone may want no payload I guess.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/PayloadSizeLimitingHttpRequesterFilter.java
Scottmitch: The exception should propagate up to the NettyChannelPublisher or be cancelled by intermediate operator which should result in cancellation of the source.

PR #2101, opened 2022-02-21
Content-Type for Zipkin Protobuf should be application/x-protobuf.

saturnism:
Motivation:

Zipkin receiver expects Content-Type of application/x-protobuf.

Why is this change being made?

Correct Content-Type is needed for receivers to correctly identity the payload format.

Modifications:

- HttpReporter's PROTO_CONTENT_TYPE is updated.

Result:

When using Codec.PROTO3, the outbound request to the trace collector will have the correct Content-Type.

### DIFF
--- a/servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
+++ b/servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
@@ -69,7 +69,7 @@ public final class HttpReporter extends Component implements Reporter<Span>, Asy
     static final String V1_PATH = "/api/v1/spans";
     static final String V2_PATH = "/api/v2/spans";
     static final CharSequence THRIFT_CONTENT_TYPE = newAsciiString("application/x-thrift");
-    static final CharSequence PROTO_CONTENT_TYPE = newAsciiString("application/protobuf");
+    static final CharSequence PROTO_CONTENT_TYPE = newAsciiString("application/x-protobuf");
 
     private final PublisherSource.Processor<Span, Span> buffer;
     private final CompositeCloseable closeable;


servicetalk-opentracing-zipkin-publisher/src/main/java/io/servicetalk/opentracing/zipkin/publisher/reporter/HttpReporter.java
```diff
@@ -69,7 +69,7 @@
     static final String V1_PATH = "/api/v1/spans";
     static final String V2_PATH = "/api/v2/spans";
     static final CharSequence THRIFT_CONTENT_TYPE = newAsciiString("application/x-thrift");
-    static final CharSequence PROTO_CONTENT_TYPE = newAsciiString("application/protobuf");
+    static final CharSequence PROTO_CONTENT_TYPE = newAsciiString("application/x-protobuf");
```
idelpivnitskiy, line 72: Consider enhancing `HttpReporterTest#verifyRequest` (btw, while you are there, please make that method `static`) to also validate content-type.

idelpivnitskiy (commented):
Thank you, @saturnism, good catch!
Just for the history of this PR, consider adding references to where you found it out.

Scottmitch (approved):

idelpivnitskiy:
References to Zipkin and OpenTelemetry that require `x-protobuf` header:
1. https://github.com/openzipkin/zipkin/blob/4a690eb1486eea8d5eaf10e47f1ea350b3ad0959/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java#L218
2. https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/zipkinreceiver/trace_receiver.go#L139

idelpivnitskiy (approved):
@saturnism we are planning to release today, I'm going to pull this in and we can enhance tests in a follow-up. Thank you for the great catch!

PR #2103, opened 2022-02-22
Validate HTTP method name matches the token format

idelpivnitskiy:
Motivation:

RFC7230, section 3.1.1 defines `method = token`. Currently, there is
no validation for the name format in `HttpRequestMethod`. HTTP/1.1
validation currently requires all uppercase letters. While the standard
HTTP methods are all uppercase, there are other standardized method that
contain tchar:
https://www.iana.org/assignments/http-methods/http-methods.xhtml

Modifications:

- Move validation of the method name from `HttpRequestDecoder` to
`HttpRequestMethod` ctor;
- Update tests to match new behavior;

Result:

HTTP/1.1 supports all method names from IANA registry.
All other protocols and use-cases get consistent method name validation.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestDecoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestDecoder.java
@@ -26,7 +26,6 @@ import io.servicetalk.utils.internal.IllegalCharacterException;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.ChannelHandlerContext;
-import io.netty.handler.codec.DecoderException;
 import io.netty.util.ByteProcessor;
 
 import java.util.Queue;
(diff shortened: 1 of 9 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestDecoder.java
```diff
@@ -113,11 +104,11 @@ private static HttpRequestMethod decodeHttpMethod(final ByteBuf buffer, final in
             return method;
         }
         try {
-            buffer.forEachByte(start, length, ENSURE_UPPER_CASE);
-        } catch (IllegalCharacterException cause) {
-            throw newInvalidMethodException(cause);
+            return HttpRequestMethod.of(methodName, NONE);
+        } catch (IllegalArgumentException cause) {
+            throw new StacklessDecoderException(
```
bondolo, line 109: The decoder exception is stackless but the IAE that is the `cause` is not. OK?

bondolo (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestDecoder.java
idelpivnitskiy: Yes, `HttpRequestMethod.of` may be invoked by users from any place they want, it might be useful to know the stack-trace there. But here we are coming from the network side, the stack-trace won't give users much info.

PR #2105, opened 2022-02-23
Increase log level when global resources are closed

idelpivnitskiy:
Motivation:

Closing global resources (`IoExecutor` and `Executor`) is unexpected.
Users should have visibility when it happens.

Modifications:

- Change log level from `debug` to `info`;
- Add stack-trace to see where subscribe comes from;

Result:

Users can notice if they unexpectedly closed any of the global
resources.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/GlobalExecutionContext.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/GlobalExecutionContext.java
@@ -154,6 +154,6 @@ public final class GlobalExecutionContext {
     private static void log(final Logger logger, final String name, final String methodName) {
         logger.info("Closure of \"{}\" was initiated using {} method. Closing the global instance before " +
                 "closing all resources that use it may result in unexpected behavior.", name, methodName,
-                new Exception("Stack trace"));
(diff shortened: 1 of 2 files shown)


tkountis (approved):

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/GlobalExecutionContext.java
```diff
@@ -152,7 +152,8 @@ public Cancellable schedule(final Runnable task, final long delay, final TimeUni
     }
 
     private static void log(final Logger logger, final String name, final String methodName) {
-        logger.debug("Closure of \"{}\" was initiated using {} method. Closing the global instance before " +
-                "closing all resources that use it may result in unexpected behavior.", name, methodName);
+        logger.info("Closure of \"{}\" was initiated using {} method. Closing the global instance before " +
+                "closing all resources that use it may result in unexpected behavior.", name, methodName,
+                new Exception("Stack trace"));
```
bondolo, line 157: I have always used `new Throwable` for stack dumps to make it clear that it is not actually an exception.

bondolo (approved):

idelpivnitskiy:
All tests passed for the previous run, I will merge without waiting for CI, the last commit is trivial

PR #2106, opened 2022-02-24
Move existing serialization example to `json` sub-folder

idelpivnitskiy:
Motivation:

Prepare `servicetalk-examples/http/serialization` for more examples for
different formats.

Modification:

- Move `servicetalk-examples:http:serialization` to
`servicetalk-examples:http:serialization:json`;
- Rename module to `servicetalk-examples-http-serialization-json`;
- Update references in docs;

Result:

`servicetalk-examples/http/serialization` can be used to add more
examples.

PR #2107, opened 2022-02-24
Add an HTTP example for protobuf serialization

idelpivnitskiy:
Motivation:

HTTP users who work with protobuf payloads need an example of how to use
`ProtobufSerializerFactory`.

Modifications:

- Add `servicetalk-examples-http-serialization-protobuf` module to
demonstrate use of `ProtobufSerializerFactory` with HTTP;
- Update docs to reference a new example;

Result:

Users can see how to use `ProtobufSerializerFactory` with HTTP.

Depends on #2106 to avoid merge conflicts, review the 2nd commit.


tkountis (approved):
looks good - consider removing the URL flavored ones, I don't think they offer any value in this example, these should be just part of the Hello one.

idelpivnitskiy:
Good point. Let me merge as is for consistency with json example and I will open a follow-up to remove URL clients from both.

idelpivnitskiy:
Removed in #2129

PR #2108, opened 2022-02-26
Remove special `defaultStrategy()` merge method

bondolo:
Motivation:
The `HttpExecutionStrategies.defaultStrategy()` strategy currently
provides a special `merge()` implementation that, rather than merge,
simply returns the other strategy. This behavior is often surprising
and requires specific handling elsewhere to ensure that the strategy
has consistent merging when referenced reflexively.
Modifications:
Replace use of `HttpExecutionStrategies.defaultStrategy().merge()` with
explicit substitution of `defaultStrategy()` with the appropriate
strategy.
Result:
Reduced special behavior in the `HttpExecutionStrategy` implementation
which clarifies the actual interaction and behavior of strategies.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
@@ -25,10 +25,20 @@ import io.servicetalk.http.api.HttpExecutionStrategyInfluencer;
 public final class GrpcExecutionStrategies {
 
     private static final GrpcExecutionStrategy NEVER_OFFLOAD_STRATEGY =
-            new DefaultGrpcExecutionStrategy(HttpExecutionStrategies.offloadNever());
+            new DefaultGrpcExecutionStrategy(HttpExecutionStrategies.offloadNever()) {
+                @Override
+                public HttpExecutionStrategy merge(final HttpExecutionStrategy other) {
+                    return this;
+                }
+            };
 
     private static final GrpcExecutionStrategy DEFAULT_GRPC_EXECUTION_STRATEGY =
-            new DefaultGrpcExecutionStrategy(HttpExecutionStrategies.defaultStrategy());
+            new DefaultGrpcExecutionStrategy(HttpExecutionStrategies.defaultStrategy()) {
+                @Override
+                public HttpExecutionStrategy merge(final HttpExecutionStrategy other) {
+                    return other;
+                }
+            };
 
     private GrpcExecutionStrategies() {
         // No instances
@@ -37,9 +47,8 @@ public final class GrpcExecutionStrategies {
     /**
      * A special default {@link GrpcExecutionStrategy} that offloads all actions unless merged with another strategy
      * that requires less offloading. The intention of this strategy is to provide a safe default if no strategy is
-     * specified; it should not be returned by
-     * {@link HttpExecutionStrategyInfluencer#requiredOffloads()}, which should return
-     * {@link HttpExecutionStrategy#offloadNone()} or {@link HttpExecutionStrategy#offloadAll()} instead.
+     * specified; it should not be returned by {@link HttpExecutionStrategyInfluencer#requiredOffloads()} which should
+     * return the actual required offloads.
      *
      * @return Default {@link GrpcExecutionStrategy}.
      */
--- a/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ErrorHandlingTest.java
+++ b/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ErrorHandlingTest.java
@@ -16,6 +16,7 @@
 package io.servicetalk.grpc.netty;
 
 import io.servicetalk.concurrent.BlockingIterator;
+import io.servicetalk.concurrent.api.AsyncCloseable;
 import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.concurrent.test.internal.TestPublisherSubscriber;
@@ -58,9 +59,11 @@ import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.stream.Stream;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.api.AsyncCloseables.newCompositeCloseable;
@@ -418,9 +421,8 @@ class ErrorHandlingTest {
     }
 
     static Collection<Arguments> data() {
-        GrpcExecutionStrategy noopStrategy = offloadNever();
         GrpcExecutionStrategy[] strategies =
-                new GrpcExecutionStrategy[]{noopStrategy, defaultStrategy()};
+                new GrpcExecutionStrategy[]{offloadNever(), defaultStrategy()};
         List<Arguments> data = new ArrayList<>(strategies.length * 2 * TestMode.values().length);
         for (GrpcExecutionStrategy serverStrategy : strategies) {
             for (GrpcExecutionStrategy clientStrategy : strategies) {
@@ -437,9 +439,14 @@ class ErrorHandlingTest {
     @AfterEach
     void tearDown() throws Exception {
         try {
-            blockingClient.close();
+            if (null != blockingClient) {
+                blockingClient.close();
+            }
         } finally {
-            newCompositeCloseable().appendAll(client, serverContext).close();
+            newCompositeCloseable().appendAll(Stream.of(client, serverContext)
+                    .filter(Objects::nonNull)
+                    .toArray(AsyncCloseable[]::new)
+            ).close();
         }
     }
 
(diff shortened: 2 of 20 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
```diff
@@ -25,10 +25,20 @@
...
+                    return this;
+                }
+            };
 
     private static final GrpcExecutionStrategy DEFAULT_GRPC_EXECUTION_STRATEGY =
-            new DefaultGrpcExecutionStrategy(HttpExecutionStrategies.defaultStrategy());
+            new DefaultGrpcExecutionStrategy(HttpExecutionStrategies.defaultStrategy()) {
+                @Override
+                public HttpExecutionStrategy merge(final HttpExecutionStrategy other) {
+                    return other;
```
bondolo, line 39: This is an unfortunate necessity currently. The alternative is to include explicit checks for this constant everywhere `HttpExecutionStrategies.defaultStrategy()` is being detected.

Given the ways in which `GrpcExecutionStrategy` is used the best option may be just to eliminate it entirely and use `HttpExecutionStrategy` everywhere.

idelpivnitskiy: From what I remember, last time we discussed possibility to remove `GrpcExecutionStrategy` we decided to keep it because it may be useful in future to control offloading for grpc-specific features. For example, users may prefer to configure offloading only for serialization/deserialization or compression/decompression but not for the request/response processing. Having it just gives us more flexibility in future.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SpecialHttpExecutionStrategy.java
```diff
@@ -112,7 +112,7 @@ public boolean isCloseOffloaded() {
          */
         @Override
         public HttpExecutionStrategy merge(final HttpExecutionStrategy other) {
-            return other;
+            throw new UnsupportedOperationException("must not be merged");
```
idelpivnitskiy, line 115: Can you please elaborate what is the benefit of throwing an exception from here? Because we change the behavior, it introduces a low risk of blowing up at runtime if users used `defaultStrategy().merge(...)` somewhere. Instead of throwing here, consider logging at info/warn level and keeping the pre-existing behavior. Understand that it may lead to unexpected results, but it will still work instead of failing. Offloading is just an optimization.

If you would like to verify that ST never invokes this method in tests, consider using `assert`.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
```diff
@@ -29,7 +29,7 @@
 /**
  * A function to determine the appropriate timeout to be used for a given {@link HttpRequestMetaData HTTP request}.
  * The result is a {@link Duration} which may be null if no timeout is to be applied. If the function blocks then
- * {@link #influenceStrategy(HttpExecutionStrategy)} should alter the execution strategy as required.
+ * {@link #requiredOffloads()} should specify the execution strategy as required.
```
idelpivnitskiy, line 32: This interface is deprecated, users should migrate from using it instead of migrating from `influenceStrategy` to `requiredOffloads()`. Consider keeping as-is.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
```diff
@@ -67,50 +68,70 @@ void add(HttpLoadBalancerFactory<?> lb) {
...
+        assert offloadNever() != strategy && defaultStrategy() != strategy :
+                "illegal " + purpose + " required strategy (" + strategy + ") for " + influencer;
         if (offloadNever() == strategy) {
             LOGGER.warn("Ignoring illegal {} required strategy ({}) for {}", purpose, strategy, influencer);
             strategy = offloadNone();
         }
-        if (strategy.hasOffloads()) {
-            clientChain = null != clientChain ? clientChain.merge(strategy) : strategy;
+        if (defaultStrategy() == strategy) {
+            LOGGER.warn("Ignoring illegal {} required strategy ({}) for {}", purpose, strategy, influencer);
```
idelpivnitskiy, line 78: Consider something like:
```java
LOGGER.warn("{}#requireOffloads() returns defaultStrategy(), which is unexpected. offloadAll() should be used instead. Making automatic adjustment, consider updating the filter.");
```
Feel free to reword. The main idea is to let users know what was done for them at runtime and what they have to do to avoid this warning.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
```diff
@@ -67,50 +68,70 @@ void add(HttpLoadBalancerFactory<?> lb) {
     }
 
     private void add(String purpose, ExecutionStrategyInfluencer<?> influencer, HttpExecutionStrategy strategy) {
+        assert offloadNever() != strategy && defaultStrategy() != strategy :
+                "illegal " + purpose + " required strategy (" + strategy + ") for " + influencer;
```
idelpivnitskiy, line 72: If we are doing automatic replacement, not sure assertion is worth being here. Better to test that we are doing replacement correctly.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ClientEffectiveStrategyTest.java
```diff
@@ -93,34 +93,28 @@
...
+            defaultStrategy(), // should be same as unspecified
             HttpExecutionStrategies.customStrategyBuilder().offloadSend().build(),
             offloadAll(),
     };
 
     private static final HttpExecutionStrategy[] FILTER_STRATEGIES = {
             null, // absent
-            offloadNever(), // treated as "offloadNoneStrategy"
             offloadNone(),
-            defaultStrategy(), // treated as "offloadNoneStrategy"
```
idelpivnitskiy, line 105: My 2c is that these use-cases are still worth to be tested

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpExecutionStrategyInfluencer.java
```diff
@@ -56,13 +58,12 @@ default HttpExecutionStrategy influenceStrategy(HttpExecutionStrategy strategy)
      *
      * <p>The provided default implementation requests offloading of all operations. Implementations that require no
      * offloading should be careful to return {@link HttpExecutionStrategies#offloadNone()} rather than
-     * {@link HttpExecutionStrategies#offloadNever()}.
+     * {@link HttpExecutionStrategies#offloadNever()}. Implementations should avoid returning
+     * {@link HttpExecutionStrategies#defaultStrategy()}, instead returning the strategy they require.
```
idelpivnitskiy, line 62: Give them a hint for the default use-case:
```
"the strategy they require." -> "the strategy they require or {@link HttpExecutionStrategies#offloadAll()} if offloading for all paths is required (safe default)."
```

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ErrorHandlingTest.java
```diff
@@ -437,9 +439,14 @@ private void setupForBlockingServiceWritesThenThrows(final BlockingTesterService
...
-            blockingClient.close();
+            if (null != blockingClient) {
+                blockingClient.close();
+            }
         } finally {
-            newCompositeCloseable().appendAll(client, serverContext).close();
+            newCompositeCloseable().appendAll(Stream.of(client, serverContext)
+                    .filter(Objects::nonNull)
+                    .toArray(AsyncCloseable[]::new)
+            ).close();
```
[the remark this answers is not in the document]
bondolo: The test was throwing NPE without actually closing the resources because some or all of the references were null. I only noticed because I had enabled a debugger breakpoint for uncaught exceptions.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutFromRequest.java
bondolo: This was a JavaDocs change that was required when I was contemplating that this change might require removal of `influenceStrategy`. As you suggest it is not critical, but it does avoid a potential problem depending upon the order we remove this interface and `influenceStrategy`

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
bondolo: Understood. For now the focus will be on removing the need for the special constants entirely and `GrpcExecutionStrategy` can continue to be used the same way it is currently.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SpecialHttpExecutionStrategy.java
bondolo: I will switch to `assert` for now.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
bondolo: The intention was that at least tests would fail and users would have some incentive to correct the problem. I suspect that when we eventually do disallow the incorrect strategies there will be failures due to client code not fixing this issue. We need to incrementally move towards disallowing the illegal strategies. I have removed the `assert` for now.

idelpivnitskiy (approved):

PR #2109, opened 2022-02-26
Publisher.multicast lack of demand on cancel

Scottmitch:
Motivation:
Publisher.multicast fails to request more from upstream
if the Subscription with the minimum demand cancels.
This may result in a live lock because the remaining
Subscribers have requested data but that request hasn't
been communicated upstream.

Modifications:
- Publisher.multicast checks when a Subscription is
  cancelled if it is has the minimum demand more
  must be requested from upstream.

Result:
No more live lock if the minimum demand Subscription cancels.


idelpivnitskiy (commented):
LGTM

PR #2110, opened 2022-02-26
Set EventLoopGroup ioRatio to 100

Scottmitch:
Motivation:
Netty's EventLoop implementations use ioRatio in
an attempt to balance the amount of time spent
executing I/O related tasks and executed/scheduled
tasks. However executing the executed tasks often
result in I/O (e.g. write operations) and scheduled
tasks are often time sensitive. Delaying execution
of these tasks may increase latency while offloading
is enabled and delay scheduled tasks on the EventLoop.
Waking up the event loop can be relatively expensive
so we amortize that cost by processing more events.
This comes with the trade-off that reading (or writing
to sockets where SND_BUF was full but now has space)
could potentially be delayed if the cardinality/cost
of executed/scheduled tasks is high. However by default
offloading doesn't execute user code so folks would
have to opt-in to run code on the EventLoop.

Modifications:
- Set ioRatio to 100 so that all executed/scheduled
  tasks are serviced on each wakeup.

Result:
Scheduled writes and timer tasks execute in more timely
fashion.


bondolo:
It should be possible to go ahead with this change now.

Scottmitch:
I will close this as it is no longer a priority.

Issue #1915, opened 2021-10-27
Flaky Test: ConnectionCloseHeaderHandlingTest


idelpivnitskiy:
https://github.com/apple/servicetalk/runs/4017719358

```
ConnectionCloseHeaderHandlingTest$PipelinedRequestsTest.3: useUds=false, viaProxy=true, awaitRequestPayload=false

java.util.concurrent.ExecutionException: io.servicetalk.transport.netty.internal.CloseHandler$CloseEventObservedException: PROTOCOL_CLOSING_INBOUND(The application protocol closed the read side of this connection. This maybe the result of sending an HTTP header such as Connection: close.) [id: 0xe60bfed2, L:/127.0.0.1:45560 ! R:127.0.0.1/127.0.0.1:39923]
	at io.servicetalk.concurrent.api.SourceToFuture.reportGet(SourceToFuture.java:121)
	at io.servicetalk.concurrent.api.SourceToFuture.get(SourceToFuture.java:92)
	at io.servicetalk.http.netty.ConnectionCloseHeaderHandlingTest$ConnectionSetup.assertResponsePayloadBody(ConnectionCloseHeaderHandlingTest.java:251)
	at io.servicetalk.http.netty.ConnectionCloseHeaderHandlingTest$PipelinedRequestsTest.serverCloseTwoPipelinedRequestsSentBeforeFirstResponse(ConnectionCloseHeaderHandlingTest.java:371)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
...
Caused by: io.servicetalk.transport.netty.internal.CloseHandler$CloseEventObservedException: PROTOCOL_CLOSING_INBOUND(The application protocol closed the read side of this connection. This maybe the result of sending an HTTP header such as Connection: close.) [id: 0xe60bfed2, L:/127.0.0.1:45560 ! R:127.0.0.1/127.0.0.1:39923]
Caused by: io.servicetalk.transport.netty.internal.StacklessClosedChannelException
	at io.servicetalk.transport.netty.internal.DefaultNettyConnection.userEventTriggered(ChannelInputShutdownReadComplete)(Unknown Source)
```

idelpivnitskiy:
https://github.com/apple/servicetalk/pull/2000/checks?check_run_id=4453922696
```
[...]

PR #2113, opened 2022-02-27
Add maxSubscriber limit to Completable/Single Processors

Scottmitch:
Motivation:
`Processors.new[Single|Completable]Processor` create a `Processor`s
that permit an unbounded number of `Subscriber`s. This may result
in memory pressure in some scenarios where the number of `Subscriber`s
is high.

Modifications:
- `ClosableConcurrentStack` to support `maxSize` to bound
  memory for its use cases.
- Add new methods for `Processors.new[Single|Completable]Processor`
  that allow configuring the maximum number of `Subscriber`s.

Result:
Completable/Single Processors and underlying data structure support
limiting the maximum amount of `Subscriber`s.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AsyncCloseables.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AsyncCloseables.java
@@ -23,6 +23,7 @@ import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
+import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.atomic.AtomicIntegerFieldUpdater.newUpdater;
 import static java.util.function.Function.identity;
 
@@ -57,27 +58,19 @@ public final class AsyncCloseables {
      * @return A new {@link ListenableAsyncCloseable}.
      */
     public static ListenableAsyncCloseable emptyAsyncCloseable() {
-        return new ListenableAsyncCloseable() {
+        return new EmptyAsyncClosable(new CompletableProcessor());
+    }
 
-            private final CompletableProcessor onClose = new CompletableProcessor();
-            private final Completable closeAsync = new SubscribableCompletable() {
-                @Override
-                protected void handleSubscribe(final Subscriber subscriber) {
-                    onClose.onComplete();
-                    onClose.subscribeInternal(subscriber);
-                }
-            };
-
-            @Override
-            public Completable closeAsync() {
-                return closeAsync;
-            }
-
-            @Override
-            public Completable onClose() {
-                return onClose;
-            }
-        };
+    /**
+     * Creates an empty {@link ListenableAsyncCloseable} that does nothing when
+     * {@link ListenableAsyncCloseable#closeAsync()} apart from completing the
+     * {@link ListenableAsyncCloseable#onClose()}.
+     * @param maxSubscribers The maximum amount of {@link CompletableSource.Subscriber} that can be subscribed to
+     * {@link ListenableAsyncCloseable#onClose()} on the return value.
+     * @return A new {@link ListenableAsyncCloseable}.
+     */
+    public static ListenableAsyncCloseable emptyAsyncCloseable(int maxSubscribers) {
+        return new EmptyAsyncClosable(new CompletableProcessor(maxSubscribers));
     }
 
     /**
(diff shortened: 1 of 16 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ClosableConcurrentStack.java
```diff
@@ -28,22 +29,56 @@
...
     @Nullable
     private volatile Object top;
+    private volatile int size;
+    private final int maxSize;
+
+    /**
+     * Create a new instance.
+     */
+    ClosableConcurrentStack() {
+        this(128);
```
Scottmitch, line 44: Rational for this number is most external use cases are to support `ListenableAsyncClosable` which tend to have a small set of fixed size `Subscriber`. Our internal repeat/retry use cases typically cancel which remove old `Subscriber` so the size doesn't need to be that big in those cases. Open to making this larger since it is a change in behavior (but likely will catch bugs/leaks, and force people to set a higher max value if required).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -737,11 +737,7 @@ protected void handleSubscribe(final Subscriber subscriber) {
 
         void nextError(final Throwable t) {
             seenError = true;
-            final CompletableSource.Processor oldProcessor = processor;
```
Scottmitch, line 740: @idelpivnitskiy - can you confirm I didn't miss anything here? This seems unnecessary now as the underlying `CompletableSource.Processor` will flush its stack when it completes and new listeners will be notified of the terminal state.

Scottmitch:
I extracted the memory leak fix into https://github.com/apple/servicetalk/pull/2114 to isolated additional features and potential change in behavior in this PR.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
Scottmitch: moved to https://github.com/apple/servicetalk/pull/2115

Scottmitch:
build failure caused by https://github.com/apple/servicetalk/issues/1915

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AsyncCloseables.java
```diff
@@ -57,27 +58,19 @@ public static Completable closeAsyncGracefully(AsyncCloseable closable, long tim
...
-        };
+    /**
+     * Creates an empty {@link ListenableAsyncCloseable} that does nothing when
+     * {@link ListenableAsyncCloseable#closeAsync()} apart from completing the
+     * {@link ListenableAsyncCloseable#onClose()}.
+     * @param maxSubscribers The maximum amount of {@link CompletableSource.Subscriber} that can be subscribed to
+     * {@link ListenableAsyncCloseable#onClose()} on the return value.
+     * @return A new {@link ListenableAsyncCloseable}.
+     */
+    public static ListenableAsyncCloseable emptyAsyncCloseable(int maxSubscribers) {
```
idelpivnitskiy, line 72: We or users may have special implementations of `CompletableSource.Processor`, consider taking the `Processor` instead of targeting specific features at this level.

Overall, I feel it will be hard for users to think about number of subscribers when they create a `ListenableAsyncCloseable`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CancellableStack.java
```diff
@@ -36,8 +44,9 @@ public void cancel() {
      * or be cancelled immediately if this object's {@link #cancel()} method has already been called.
      * @param toAdd The {@link Cancellable} to add.
      * @return {@code true} if the {@code toAdd} was added. If {@code false} {@link Cancellable#cancel()} is called.
+     * @throws IllegalStateException if the maximum size would be exceeded by inserting this element.
      */
-    boolean add(Cancellable toAdd) {
+    boolean add(Cancellable toAdd) throws IllegalStateException {
```
idelpivnitskiy, line 49: Javadoc should be enough, we try to use `throws` only for checked exceptions, like in Collection API.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ClosableConcurrentStack.java
```diff
@@ -28,22 +29,57 @@
...
      * Push an item onto the stack.
      * @param item the item to push onto the stack.
      * @return {@code true} if the operation was successful. {@code false} if {@link #close(Consumer)} has been called
      * and {@code item} has been consumed via {@link Consumer#accept(Object)} of the {@link #close(Consumer)} argument.
+     * @throws IllegalStateException if the maximum size would be exceeded by inserting this element.
      */
     @SuppressWarnings("unchecked")
     @Nullable
-    boolean push(T item) {
+    boolean push(T item) throws IllegalStateException {
```
idelpivnitskiy, line 68: Same

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/IterableMergeCompletable.java
```diff
@@ -37,7 +37,7 @@ public CompletableDynamicCountSubscriber apply(final Subscriber subscriber) {
 
     @Override
     void doMerge(final CompletableDynamicCountSubscriber subscriber) {
-        long count = 1;
```
idelpivnitskiy, line 40: Why it can not be `long` anymore?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CompletableMergeSubscriber.java
```diff
@@ -35,10 +35,10 @@
     private final CancellableStack dynamicCancellable;
     private final boolean delayError;
 
-    CompletableMergeSubscriber(Subscriber subscriber, boolean delayError) {
+    CompletableMergeSubscriber(Subscriber subscriber, boolean delayError, int maxCancellables) {
         this.subscriber = subscriber;
         this.delayError = delayError;
-        dynamicCancellable = new CancellableStack();
+        dynamicCancellable = new CancellableStack(maxCancellables);
```
idelpivnitskiy, line 41: Should we try-catch line 47 now?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ClosableConcurrentStack.java
idelpivnitskiy: Don't mind changing the default as long as users can understand what is wrong/changed from the popped up error and what they should do. Currently, the exception message from `ClosableConcurrentStack` is too abstract to understand the root cause.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/IterableMergeCompletable.java
Scottmitch: The limit of the stack is `int` so going beyond this range will no longer be allowed.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/CancellableStack.java
bondolo: There is no harm in documenting runtime exceptions and it can improve IDE hints and static analysis.

Scottmitch:
I'm going to close this as after the memory leak was fixed on close/cancel we don't need this feature. it adds additional complexity and memory when limits are rarely required.

PR #2114, opened 2022-02-27
Fix memory leak in Completable/Single Processors

Scottmitch:
Motivation:
`Processors.new[Single|Completable]Processor` create a `Processor`s
that permit multiple subscribers. The underlying datastructure is
`ClosableConcurrentStack` that makes a best effort to remove items
on cancellation by setting the `Subscriber` reference to `null`.
However it doesn't attempt to remove the `Node` which can lead
to a memory leak until `close(..)` is called. This may not happen
in scenarios that perpetually retry operations that are failing
continuously for prolonged periods of time.

Modifications:
- `ClosableConcurrentStack` makes a best effort to remove the
  `Node` object from the stack in `relaxedRemove`.

Result:
Memory leak fixed.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ClosableConcurrentStack.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ClosableConcurrentStack.java
@@ -70,15 +70,27 @@ final class ClosableConcurrentStack<T> {
             return false;
         }
         @SuppressWarnings("unchecked")
-        Node<T> currTop = (Node<T>) rawCurrTop;
-        while (currTop != null) {
-            if (item.equals(currTop.item)) {
-                currTop.item = null; // best effort null out the item. pop/close will discard the Node later.
+        Node<T> curr = (Node<T>) rawCurrTop;
+        Node<T> prev = null;
+        do {
+            if (item.equals(curr.item)) {
+                // Best effort Node removal. close will discard the Node if this attempt isn't visible to all threads.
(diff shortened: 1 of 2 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ClosableConcurrentStack.java
```diff
@@ -70,15 +70,27 @@ boolean relaxedRemove(T item) {
...
-                currTop.item = null; // best effort null out the item. pop/close will discard the Node later.
+        Node<T> curr = (Node<T>) rawCurrTop;
+        Node<T> prev = null;
+        do {
+            if (item.equals(curr.item)) {
+                // Best effort Node removal. close will discard the Node if this attempt isn't visible to all threads.
+                curr.item = null;
+                if (prev != null) {
+                    prev.next = curr.next;
+                    // Don't set curr.next to null! If multiple threads are removing items at the same time removed
```
Scottmitch, line 81: this case is already covered by existing unit tests `ClosableConcurrentStackTest.concurrentPushRemoveDifferentThread`

Scottmitch:
Will merge to absorb into release. Comments welcome and will address in followup.

Scottmitch:
followup PR for more robustness: https://github.com/apple/servicetalk/pull/2116

idelpivnitskiy (commented):
LGTM

Scottmitch:
0.41 (096d4d206)

PR #2115, opened 2022-02-27
Minor Processor usage cleanup

Scottmitch:
Modifications:
- DefaultSingleAddressHttpClientBuilder$SdStatusCompletable to
  avoid creating multiple Processors when an error occurs.
  The Processor will ensure the underlying data structure doesn't
  leak memory.
- LoadBalancerReadySubscriber to retain the Processor reference
  in a terminal state. Subsequent subscribers will therefore see
  the correct terminal state.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018-2021 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2018-2022 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -737,11 +737,7 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
 
         void nextError(final Throwable t) {
             seenError = true;
-            final CompletableSource.Processor oldProcessor = processor;
-            oldProcessor.onError(t);
-            final CompletableSource.Processor newProcessor = newCompletableProcessor();
-            newProcessor.onError(t);
-            processor = newProcessor;
+            processor.onError(t);
         }
 
         void resetError() {
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
@@ -737,7 +737,13 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
 
         void nextError(final Throwable t) {
             seenError = true;
-            processor.onError(t);
+            // This state is reused across multiple subscribes, and we reset the processor to deliver the latest error
+            // to new subscribers.
+            final CompletableSource.Processor oldProcessor = processor;
+            oldProcessor.onError(t);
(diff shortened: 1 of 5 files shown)


tkountis (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -737,11 +737,7 @@ protected void handleSubscribe(final Subscriber subscriber) {
 
         void nextError(final Throwable t) {
             seenError = true;
-            final CompletableSource.Processor oldProcessor = processor;
-            oldProcessor.onError(t);
-            final CompletableSource.Processor newProcessor = newCompletableProcessor();
-            newProcessor.onError(t);
-            processor = newProcessor;
```
idelpivnitskiy, line 744: We discussed this in the past: https://github.com/apple/servicetalk/pull/1070#discussion_r434170797

If we do not replace the processor with the new one, we will hide newer exceptions because the original processor is already terminated. While you are here, please leave a comment to clarify that.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancerReadySubscriber.java
```diff
@@ -58,31 +57,29 @@ public void onNext(final Object o) {
...
             }
         }
     }
 
     @Override
     public void onError(final Throwable t) {
-        Processor onHostsAvailable = LoadBalancerReadySubscriber.this.onHostsAvailable;
+        Processor onHostsAvailable = this.onHostsAvailable;
         if (onHostsAvailable != null) {
-            LoadBalancerReadySubscriber.this.onHostsAvailable = null;
```
idelpivnitskiy, line 76: Good catch!
Would be great to have a test for that

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancerReadySubscriber.java
```diff
@@ -58,31 +57,29 @@ public void onNext(final Object o) {
...
     }
 
     @Override
     public void onComplete() {
-        Processor onHostsAvailable = LoadBalancerReadySubscriber.this.onHostsAvailable;
+        Processor onHostsAvailable = this.onHostsAvailable;
         if (onHostsAvailable != null) {
-            LoadBalancerReadySubscriber.this.onHostsAvailable = null;
             // Let the load balancer or retry strategy fail any pending requests.
             onHostsAvailable.onComplete();
```
idelpivnitskiy, line 84: Looks like here is a bug. Javadoc for `onHostAvailable()` says:

> Get `Completable` that will complete when a `LoadBalancerReadyEvent` returns `true` from `LoadBalancerReadyEvent#isReady()`.

If `this.onHostsAvailable != null`, LB is not ready. Completing it here will be misleading. We should terminate `onHostsAvailable` with an error to signal that LB events stream terminated.

idelpivnitskiy (changes requested):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
Scottmitch: ahh thx for reminding me. yes +1 to comment

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancerReadySubscriber.java
Scottmitch: tests added

idelpivnitskiy (approved):
Thank you!

PR #2116, opened 2022-02-27
Completable/Single Processors more robust top of stack removal

Scottmitch:
Motivation:
Completable/Single Processors make a best effort to remove
nodes from the underlying data structure upon cancellation.
However if the node is the last added and there is concurrent
cancellation we may may fail to remove the node and give up
leading to increased memory consumption.

Modifications:
- `ClosableConcurrentStack` should iterate the stack to remove
  the root node if the first removal attempt fails.

Result:
More robust cleanup of memory on removal.


idelpivnitskiy (commented):
LGTM,
Consider cherry-picking this and #2114 into 0.41

Scottmitch:
0.41 (805a73ed2)

Scottmitch:
@idelpivnitskiy - +1 for cherry-pick. done.

PR #2118, opened 2022-03-01
Reduce initial queue size for h1 pipelined connection

Scottmitch:
Motivation:
NettyPipelinedConnection provides pipelined request/response
support on the client. It uses 2 queues: 1 for writes and 1
for reads. The initial queue sizes are not set and therefore
the default value of 1024 is selected. This number should be
correlated to the max number of pipelined requests expected
(which is often 1). This initial queue size is wasteful and
consumes extra memory per connection.

Modifications:
- Correlate queue size to max pipelined requests in
  NettyPipelinedConnection.

Result:
Less memory usage for h1 connections.


chemicL:
Late LGTM

idelpivnitskiy (commented):
LGTM

PR #2119, opened 2022-03-01
Reduce initial size of WriteStreamSubscriber queue

Scottmitch:
Motivation:
WriteStreamSubscriber initial queue size is 8 however
most messages have ~3 items in the stream
{headers, payload, trailers}, and h2 may have 2 items.
We can save memory in most cases by reducing the initial
size from 8 to 4.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
@@ -340,9 +340,9 @@ final class WriteStreamSubscriber implements PublisherSource.Subscriber<Object>,
          * </pre>
          * We assume that no listener for a write is added after that write is completed (a.k.a late listeners).
          * Most of the messages have 3 listeners (headers, one payload body chunk, trailers). Messages with large
-         * streaming payload body may have more listeners. However, the MIN_INITIAL_CAPACITY of ArrayDeque is 8.
+         * streaming payload body may have more listeners.
          */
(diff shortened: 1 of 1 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
```diff
@@ -340,9 +340,9 @@ private void requestMoreIfRequired(@Nullable Subscription subscription, long byt
          * </pre>
          * We assume that no listener for a write is added after that write is completed (a.k.a late listeners).
          * Most of the messages have 3 listeners (headers, one payload body chunk, trailers). Messages with large
-         * streaming payload body may have more listeners. However, the MIN_INITIAL_CAPACITY of ArrayDeque is 8.
```
Scottmitch, line 343: `MIN_INITIAL_CAPACITY` -> OpenJDK no longer has this restriction.

normanmaurer: nice! Do you know when this changed ?

chemicL:
Late LGTM

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/WriteStreamSubscriber.java
idelpivnitskiy: starting from JDK9, there is no `MIN_INITIAL_CAPACITY` anymore

idelpivnitskiy (commented):
LGTM

Issue #2069, opened 2022-02-01
FlakyTest: MalformedDataAfterHttpMessageTest


bondolo:
https://github.com/apple/servicetalk/pull/2068/checks?check_run_id=5015006266
```
MalformedDataAfterHttpMessageTest.[2] false

io.servicetalk.transport.netty.internal.StacklessClosedChannelException
	at io.servicetalk.transport.netty.internal.NettyChannelPublisher.channelOnError(Unknown Source)
Caused by: io.servicetalk.http.netty.HttpObjectDecoder$StacklessDecoderException: Invalid preface character before the start-line of the HTTP message
Caused by: io.servicetalk.utils.internal.IllegalCharacterException: '?' (0x00), expected [CR (0x0d), LF (0x0a)]
	at io.servicetalk.http.netty.HttpObjectDecoder.lambda$static$0(HttpObjectDecoder.java:99)
	at io.netty.buffer.AbstractByteBuf.forEachByteAsc0(AbstractByteBuf.java:1301)
	at io.netty.buffer.AbstractByteBuf.forEachByte(AbstractByteBuf.java:1292)
	at io.servicetalk.http.netty.HttpObjectDecoder.skipControlCharacters(HttpObjectDecoder.java:588)
	at io.servicetalk.http.netty.HttpObjectDecoder.decode(HttpObjectDecoder.java:239)
	at io.servicetalk.transport.netty.internal.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:365)
	at io.servicetalk.transport.netty.internal.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:321)
	at io.servicetalk.transport.netty.internal.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:185)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.servicetalk.transport.netty.internal.CopyByteBufHandlerChannelInitializer$CopyByteBufHandler.channelRead(CopyByteBufHandlerChannelInitializer.java:86)
[...]

Issue #1894, opened 2021-10-18
Flaky Test: FullDuplexAndSequentialModeTest


bondolo:
https://github.com/apple/servicetalk/pull/1892/checks?check_run_id=3921774002

```
FullDuplexAndSequentialModeTest.deferResponseUntilAfterRequestSent()
java.util.concurrent.TimeoutException: deferResponseUntilAfterRequestSent() timed out after 30 seconds

java.util.concurrent.TimeoutException: deferResponseUntilAfterRequestSent() timed out after 30 seconds
	at org.junit.jupiter.engine.extension.TimeoutInvocation.createTimeoutException(TimeoutInvocation.java:70)
	at org.junit.jupiter.engine.extension.TimeoutInvocation.proceed(TimeoutInvocation.java:59)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:149)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:140)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:84)
	at org.junit.jupiter.engine.execution.ExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(ExecutableInvoker.java:115)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.lambda$invoke$0(ExecutableInvoker.java:105)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.proceed(InvocationInterceptorChain.java:64)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.chainAndInvoke(InvocationInterceptorChain.java:45)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.invoke(InvocationInterceptorChain.java:37)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:104)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:98)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$7(TestMethodTestDescriptor.java:214)
[...]

PR #2120, opened 2022-03-01
Reduce Publisher/Single retry/repeat allocations

Scottmitch:
Motivation:
RepeatWhenSingle was recently introduced and avoids allocating
new Subscribers for each repeat operation. This helps save
object allocation when the operator is used in a loop. We
should use a consistent strategy for existing Publisher/Single
repeat/retry related operators.

Modifications:
- Redo[When]Publisher and Retry[When]Single no longer allocates
  a new Subscriber on each iteration.
- All the repeat/retry operators now have consistent behavior
  for AsyncContext map which is to re-use it. Previously a copy
  was made before resubscribing, however the old map is no longer
  used after a terminal event anyways so this shouldn't have any
  benefit. Either we copy the map up front before subscribe, or
  we just re-use the same map and let the async source at the top
  of the chain reset if necessary.

Result:
Less object allocation and consistent AsyncContext behavior for
repeat/retry operators.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RedoPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RedoPublisher.java
@@ -49,10 +49,9 @@ final class RedoPublisher<T> extends AbstractNoHandleSubscribePublisher<T> {
     }
 
     abstract static class AbstractRedoSubscriber<T> implements Subscriber<T> {
-
         final SequentialSubscription subscription;
-        final int redoCount;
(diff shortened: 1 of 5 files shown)


Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1894

Scottmitch:
> Could not GET 'https://plugins.gradle.org/m2/com/google/code/findbugs/jsr305/3.0.2/jsr305-3.0.2.pom'. Received status code 503 from server: Service Unavailable

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/2069

PR #2121, opened 2022-03-01
Add netty-tcnative-boringssl-static runtime dependency

Scottmitch:
Motivation:
netty-tcnative-boringssl-static accelerates TLS and is enabled
if included on the classpath. ServiceTalk include's Netty's bom
which now also includes the compatible tcnative version, so we
can include this dependency to simplify version alignment.

Modifications:
- Add netty-tcnative-boringssl-static runtimeOnly dependency for
  macOS/linux x86-64/aarch-64

Result:
Accelerated TLS without coordinating versions manually.

### DIFF
--- a/servicetalk-transport-netty-internal/build.gradle
+++ b/servicetalk-transport-netty-internal/build.gradle
@@ -42,6 +42,10 @@ dependencies {
   implementation "io.netty.incubator:netty-incubator-transport-native-io_uring:$nettyIoUringVersion"
   runtimeOnly "io.netty.incubator:netty-incubator-transport-native-io_uring:$nettyIoUringVersion:linux-x86_64"
   runtimeOnly "io.netty.incubator:netty-incubator-transport-native-io_uring:$nettyIoUringVersion:linux-aarch_64"
+  runtimeOnly( group:"io.netty", name:"netty-tcnative-boringssl-static", classifier:"linux-x86_64")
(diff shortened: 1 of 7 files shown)


servicetalk-transport-netty-internal/build.gradle
```diff
@@ -42,6 +42,10 @@ dependencies {
   implementation "io.netty.incubator:netty-incubator-transport-native-io_uring:$nettyIoUringVersion"
   runtimeOnly "io.netty.incubator:netty-incubator-transport-native-io_uring:$nettyIoUringVersion:linux-x86_64"
   runtimeOnly "io.netty.incubator:netty-incubator-transport-native-io_uring:$nettyIoUringVersion:linux-aarch_64"
+  runtimeOnly( group:"io.netty", name:"netty-tcnative-boringssl-static", classifier:"linux-x86_64")
```
Scottmitch, line 45: Note as of tcnative-2.0.49 Netty [no longer repackages into an uber jar](https://github.com/netty/netty-tcnative/commit/6fb5b01208671c007ad7f752e51ccc8b25dfca2f), so build tools will be able to dedup between  the uber jar and individual classifier artifacts.

thx @normanmaurer ❤️

chemicL (approved):

idelpivnitskiy (approved):

PR #2125, opened 2022-03-02
Explain wire/frame logging `logUserData`

bondolo:
Motivation:
The current debugging examples include a minimal explanation of the
`logUserData` feature of `enableWireLogging` and `enableFrameLogging`
which could be expanded to clarify the use.
Modifications:
Additional explanation of `logUserData` by provided by defining a
`LOG_USER_DATA` constant with JavaDoc
Result:
Improved examples.

### DIFF
--- a/servicetalk-examples/grpc/debugging/src/main/java/io/servicetalk/examples/grpc/debugging/DebuggingClient.java
+++ b/servicetalk-examples/grpc/debugging/src/main/java/io/servicetalk/examples/grpc/debugging/DebuggingClient.java
@@ -43,7 +43,9 @@ import static io.servicetalk.logging.api.LogLevel.TRACE;
  * </ol>
  * <p>The wire and frame logging features require that you configure a logger with an appropriate log level. For this
  * example {@code log4j2.xml} is used by both the client and server and configures the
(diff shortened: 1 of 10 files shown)


servicetalk-examples/grpc/debugging/src/main/java/io/servicetalk/examples/grpc/debugging/DebuggingClient.java
```diff
@@ -109,6 +111,15 @@
         // AsyncContext.disable();
     }
 
+    /**
+     * If {@link Boolean#TRUE} then all user data in the payload will be logged in addition to the protocol metadata.
+     * If {@link Boolean#FALSE} then protocol operations will be logged, but user data contents will be omitted.
+     *
+     * <p>This implementation uses a constant function to enable or disable logging of user data, your implementation
+     * could selectively choose to log user data based upon application state or context.</p>
```
idelpivnitskiy, line 119: Consider "selectively choose" -> "selectively choose at runtime"

idelpivnitskiy: Also, consider clarifying that logging user data may leak sensitive content (header values and payload body) to the logs and users should be careful enabling it in production environments.

idelpivnitskiy (approved):

PR #2126, opened 2022-03-04
Enhance URI parsing exceptions

idelpivnitskiy:
Motivation:

Provide more information in URI exceptions to help users understand the
issue.

Modifications:

- Enhance messages with more context;

Result:

Messages provide more context without leaking the full URI because it
may contain sensitive info in query or path params.

Example:
- before: duplicate/invalid host
- after: Invalid URI format: duplicate or invalid host. Already parsed host: localhost, but found ':' character at index 23. Total URI length: 45


tkountis (approved):
Change looks ok. Can you please include error samples (before / after) of the change in the description?

PR #2129, opened 2022-03-07
Remove examples for multi-address client that similar to single-address

idelpivnitskiy:
Motivation:

Some HTTP examples that use `HttpClients.forMultiAddressUrl()` are
identical to the other example that uses
`HttpClients.forSingleAddress(...)`. It doesn't provide much value for
users because the top level client API is the same for both client
factories. It's enough to demonstrate multi-address use-case in
hello-world example only.

Modifications:

- Remove classes that target `forMultiAddressUrl()` from json, protobuf,
and retry examples;
- Remove link to `RetryUrlClient` from docs;

Result:

Less classes in examples to maintain.


tkountis (approved):
ty

idelpivnitskiy:
@bondolo good point, I've created #2130

PR #2131, opened 2022-03-08
Deprecated `PartitionedHttpClientBuilder` and partition client APIs

idelpivnitskiy:
Motivation:

`PartitionedHttpClientBuilder` is hard to use because it requires
implementation of a special `ServiceDiscoverer` that we do not provide
OOTB. We are unaware of anyone using this feature, but if there are any
users they should consider `ClientGroup` as an alternative or reach out
to us to provide more context why the `PartitionedHttpClientBuilder` is
required.

Modifications:

- Deprecate `PartitionedHttpClientBuilder` and
`HttpClients.forPartitionedAddress`;
- Deprecate everything in `io.servicetalk.client.api.partition` package

Result:

`PartitionedHttpClientBuilder` is deprecated and can be removed in
future releases.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/partition/ClosedPartitionException.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/partition/ClosedPartitionException.java
@@ -15,12 +15,19 @@
  */
 package io.servicetalk.client.api.partition;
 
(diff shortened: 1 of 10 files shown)


tkountis (approved):
🔥

bondolo (approved):
I made minor editorial changes to title and commit message.

PR #2136, opened 2022-03-09
Decouple multi-address and partitioned client builders from `HttpClientBuildContext`

idelpivnitskiy:
Motivation:

`DefaultMultiAddressHttpClientBuilder` and
`DefaultPartitionedHttpClientBuilder` use
`DefaultSingleAddressHttpClientBuilder` as a template for underlying
single-address clients. It was helpful before we introduced a
`SingleAddressInitializer` in #1387. Before that, all builders had the
same set of methods and it was nice to reuse the single-address builder
instead of maintaining a similar state at higher level builders. Now,
`SingleAddressInitializer` helps to avoid duplication between builders.
We don't have to use `DefaultSingleAddressHttpClientBuilder` as a
template and can simplify its state by using a factory for
`SingleAddressHttpClientBuilder`.

Modifications:

- `DefaultMultiAddressUrlHttpClientBuilder` and
`DefaultPartitionedHttpClientBuilder` use a factory for
`SingleAddressHttpClientBuilder` instead of
`DefaultSingleAddressHttpClientBuilder` as a template and
`HttpExecutionContextBuilder` for context-related items;
- Remove and hide methods from `DefaultSingleAddressHttpClientBuilder`
that are not used by other builders anymore;
- Add a test to verify `DefaultMultiAddressUrlHttpClientBuilder` sets
execution context items for underlying single-address builders;

Result:

`DefaultSingleAddressHttpClientBuilder` has less pkg-private API,
`DefaultMultiAddressUrlHttpClientBuilder` and
`DefaultPartitionedHttpClientBuilder` create a new single-address
builder on demand, similar to grpc builders.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
@@ -35,12 +35,12 @@ import io.servicetalk.http.api.HttpRequestMetaData;
 import io.servicetalk.http.api.HttpRequestMethod;
 import io.servicetalk.http.api.MultiAddressHttpClientBuilder;
 import io.servicetalk.http.api.RedirectConfig;
+import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
 import io.servicetalk.http.api.StreamingHttpClient;
 import io.servicetalk.http.api.StreamingHttpRequest;
 import io.servicetalk.http.api.StreamingHttpRequestResponseFactory;
 import io.servicetalk.http.api.StreamingHttpResponse;
 import io.servicetalk.http.api.StreamingHttpResponseFactory;
-import io.servicetalk.http.netty.DefaultSingleAddressHttpClientBuilder.HttpClientBuildContext;
 import io.servicetalk.http.utils.RedirectingHttpRequesterFilter;
 import io.servicetalk.transport.api.ClientSslConfig;
 import io.servicetalk.transport.api.ClientSslConfigBuilder;
@@ -64,6 +64,7 @@ import static io.servicetalk.concurrent.api.AsyncCloseables.toListenableAsyncClo
 import static io.servicetalk.concurrent.api.Single.defer;
 import static io.servicetalk.concurrent.internal.SubscriberUtils.deliverCompleteFromSource;
 import static io.servicetalk.http.api.HttpProtocolVersion.HTTP_1_1;
+import static io.servicetalk.http.netty.HttpExecutionContextBuilder.setExecutionContext;
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -85,7 +86,8 @@ final class DefaultMultiAddressUrlHttpClientBuilder
 
     private static final String HTTPS_SCHEME = HTTPS.toString();
 
-    private final DefaultSingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> builderTemplate;
+    private final Function<HostAndPort, SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress>> builderFactory;
+    private final HttpExecutionContextBuilder executionContextBuilder = new HttpExecutionContextBuilder();
 
     @Nullable
     private HttpHeadersFactory headersFactory;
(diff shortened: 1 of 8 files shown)


idelpivnitskiy:
Another part of the motivation for this change is to be able to intercept when multi-address client initializes a new single-address client builder to make it work for providers in #2137.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -220,32 +215,36 @@ public int hashCode() {
...
             this.singleAddressInitializer = singleAddressInitializer;
         }
 
         @Override
         public StreamingHttpClient apply(final UrlKey urlKey) {
-            // Copy existing builder to prevent changes at runtime when concurrently creating clients for new addresses
-            final HttpClientBuildContext<HostAndPort, InetSocketAddress> buildContext =
-                    builderTemplate.copyBuildCtx(urlKey.hostAndPort);
+            final SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> builder =
+                    builderFactory.apply(urlKey.hostAndPort);
```
bondolo, line 236: Perhaps null check this result. I have noticed we have generally checked for provided Functions and Suppliers and think it is a good practice.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultPartitionedHttpClientBuilder.java
```diff
@@ -66,75 +66,73 @@
...
         }
-        ServiceDiscoverer<U, R, PartitionedServiceDiscovererEvent<R>> psd =
+        final ServiceDiscoverer<U, R, PartitionedServiceDiscovererEvent<R>> psd =
                 new DefaultSingleAddressHttpClientBuilder.RetryingServiceDiscoverer<>(serviceDiscoverer,
                         sdRetryStrategy);
 
         final PartitionedClientFactory<U, R, FilterableStreamingHttpClient> clientFactory = (pa, sd) -> {
             // build new context, user may have changed anything on the builder from the filter
-            DefaultSingleAddressHttpClientBuilder<U, R> builder = buildContext.builder.copyBuildCtx().builder;
+            final SingleAddressHttpClientBuilder<U, R> builder = builderFactory.get();
```
bondolo, line 115: Another case of null-check for Function/Supplier result

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpExecutionContextBuilder.java
```diff
@@ -71,4 +72,12 @@ public HttpExecutionContext build() {
                 executor == null ? defaultContextSupplier.get().executor() : executor,
                 strategy == null ? defaultStrategy() : strategy);
     }
+
+    static <U, R> SingleAddressHttpClientBuilder<U, R> setExecutionContext(
```
bondolo, line 76: It seems strange to have this as a method on `HttpExecutionContextBuilder` when none of the types involved is this type. Can we somehow put this on `SingleAddressHttpClientBuilder`?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
```diff
@@ -138,75 +137,98 @@ public HttpHeaders newHeaders() {
...
-            assertThat(actualInternalExecutionStrategy.get(), equalTo(internalExecutionStrategy));
-
-            streamingHttpClient.closeAsync().toFuture().get();
+            assertExecutionContext(internalCtx, actualInternalCtx.get());
         }
     }
 
+    private static void assertExecutionContext(HttpExecutionContext expected, HttpExecutionContext actual) {
+        assertThat(actual, is(notNullValue()));
+        assertThat(actual.ioExecutor(), equalTo(expected.ioExecutor()));
```
bondolo, line 226: I wonder if these should be slightly stronger `sameInstance` rather than `equalTo`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilderTest.java
```diff
@@ -85,8 +85,9 @@ private static void hostToCharSequenceFunction(String hostNamePrefix, String hos
                         .build())
                 .listenBlockingAndAwait((ctx, request, responseFactory) -> responseFactory.ok());
              BlockingHttpClient client =
-                     forResolvedAddress(hostNamePrefix + hostName + hostNameSuffix + (port == null ? "" : port),
-                             u -> serverCtx.listenAddress())
+                     forSingleAddress(
+                             GlobalDnsServiceDiscoverer.mappingServiceDiscoverer(u -> serverCtx.listenAddress()),
```
bondolo, line 89: Cute! I wondered what you were adding that for, it seemed unrelated.

bondolo (approved):

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilderTest.java
idelpivnitskiy: The goal was to remove all static factories from `DefaultSingleAddressHttpClientBuilder` to keep the only one ctor. To achieve that, I had to move SD for resolved addresses out or `DefaultSingleAddressHttpClientBuilder`. It required change in this test because there is no more `DefaultSingleAddressHttpClientBuilder.forResolvedAddress`

idelpivnitskiy: Actually, because this is `DefaultSingleAddressHttpClientBuilderTest`, let me just use `DefaultSingleAddressHttpClientBuilder` ctor here.

chemicL:
Nice! 👍

PR #2137, opened 2022-03-09
Introduce `HttpProviders` and `GrpcProviders`

idelpivnitskiy:
Allow users to intercept static factories that create client/server
builders to enhance builders' behavior. Users can use these providers
to override default values on the builders or to intercept builders for
metrics purposes.

Modifications:

- Add `HttpProviders` and `GrpcProviders` that define provider
interfaces for `ServiceLoader`;
- Test how users can intercept builders;

Result:

Users can change the default builders using providers that are
automatically loaded via `ServiceLoader`.


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DelegatingGrpcClientBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Override
+    public GrpcClientBuilder<U, R> initializeHttp(final HttpInitializer<U, R> initializer) {
+        delegate = delegate.initializeHttp(initializer);
+        return this;
+    }
+
+    @Override
+    public GrpcClientBuilder<U, R> defaultTimeout(final Duration defaultTimeout) {
+        delegate = delegate.defaultTimeout(defaultTimeout);
```
idelpivnitskiy, line 58: When I had all builder methods as
```java
    @Override
    public GrpcClientBuilder<U, R> defaultTimeout(final Duration defaultTimeout) {
        delegate.defaultTimeout(defaultTimeout);
        return this;
    }
```
Spotbugs complained that the return value is discarded. I though to suppress it because all our builder implementations always return `this`. But then considered a use-case when users may have an immutable provider or builder impl that always return a new instance. LMK what you think. Should we support that and reassign the `delegate` or is it better to keep `delegate` as `final`?

bondolo: I don't see any alternative to assigning `delegate` with each delegate invocation. Eventually someone will get clever and implement a builder which doesn't return `this` and we will have to fix it anyway. I've had to fix very non-obvious problems more than once where failing to reassign the builder at the end of the chain meant that the operations in the chain were lost. It is always a frustrating bug to locate.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcProviders.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    @FunctionalInterface
+    public interface GrpcServerBuilderProvider {
+
+        /**
+         * Returns a {@link GrpcServerBuilder} based on the address and pre-initialized {@link GrpcServerBuilder}.
+         *
+         * @param address a server address used to create a {@link GrpcServerBuilder}
+         * @param builder pre-initialized {@link GrpcServerBuilder}
+         * @return a {@link GrpcServerBuilder} based on the address and pre-initialized{@link GrpcServerBuilder}.
```
bondolo, line 70: For all of the provider interfaces we should be more clear about the implications for the source builder, specifically that a provider **CAN** return the source builder from this method and subsequent operations on the returned builder **MAY** (and also perhaps don't) mutate the source builder.

bondolo, line 70: For all of the provider interfaces we should be more clear about the implications for the source builder, specifically that a provider **CAN** return the source builder from this method and subsequent operations on the returned builder **MAY** (and also perhaps don't) mutate the source builder.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcProviders.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.net.SocketAddress;
+import java.util.ServiceLoader;
+
+/**
+ * A holder for all gRPC-specific providers that can be registered using {@link ServiceLoader}.
+ *
+ * @see HttpProviders
+ */
+public final class GrpcProviders {
```
bondolo, line 30: When I think of the name `Provider` it is akin to me to `Supplier`, but extended from no-args to act as an abstracted virtual constructor. These interfaces feel more like `Customizers` to use Spring terminology. I wonder if others might be confused by the name.

bondolo, line 30: When I think of the name `Provider` it is akin to me to `Supplier`, but extended from no-args to act as an abstracted virtual constructor. Providers are also generally not chaining interfaces, but more like a singleton. These interfaces feel more like `Customizers` to use Spring terminology. I wonder if others might be confused by the name.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcProvidersTest.java
```diff
@@ -0,0 +1,202 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static io.servicetalk.transport.netty.internal.AddressUtils.localAddress;
+import static io.servicetalk.transport.netty.internal.AddressUtils.serverHostAndPort;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+class GrpcProvidersTest {
```
bondolo, line 58: Please add `@Execution(ExecutionMode.SAME_THREAD)` since the `TestGrpcServerBuilderProvider` and `TestGrpcClientBuilderProvider` have static state.

bondolo, line 58: Please add `@Execution(ExecutionMode.SAME_THREAD)` since the `TestGrpcServerBuilderProvider` and `TestGrpcClientBuilderProvider` have static state.

bondolo: (I won't suggest you create new class loaders and define new classes for each test run)

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpProviders.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+package io.servicetalk.http.api;
+
+import java.net.SocketAddress;
+import java.util.ServiceLoader;
+
+/**
+ * A holder for all HTTP-specific providers that can be registered using {@link ServiceLoader}.
+ */
+public final class HttpProviders {
```
bondolo, line 24: Same comments as `GrpcProviders`

bondolo, line 24: Same comments as `GrpcProviders`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
```diff
@@ -36,27 +38,59 @@
...
 import static java.util.function.Function.identity;
 
 /**
  * Factory methods for building {@link HttpClient} (and other API variations) instances.
  */
 public final class HttpClients {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(HttpClients.class);
+
+    private static final List<SingleAddressHttpClientBuilderProvider> SINGLE_ADDRESS_PROVIDERS;
```
bondolo, line 65: Same comments as `GrpcClients`

bondolo, line 65: Same comments as `GrpcClients`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
```diff
@@ -36,27 +38,59 @@
...
 import static io.servicetalk.concurrent.api.Publisher.failed;
 import static io.servicetalk.http.netty.GlobalDnsServiceDiscoverer.globalDnsServiceDiscoverer;
 import static io.servicetalk.http.netty.GlobalDnsServiceDiscoverer.globalSrvDnsServiceDiscoverer;
 import static io.servicetalk.http.netty.GlobalDnsServiceDiscoverer.mappingServiceDiscoverer;
+import static io.servicetalk.utils.internal.ServiceLoaderUtils.loadProviders;
 import static java.util.function.Function.identity;
 
 /**
  * Factory methods for building {@link HttpClient} (and other API variations) instances.
  */
```
bondolo, line 60: You may want to references `HttpProviders` or the client provider interfaces here. "You can customize the creation of clients using the `SingleAddressHttpClientBuilderProvider` interface".

bondolo, line 60: You may want to references `HttpProviders` or the client provider interfaces here. "You can customize the creation of clients using the `SingleAddressHttpClientBuilderProvider` interface".

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpServers.java
```diff
@@ -15,38 +15,62 @@
...
+import org.slf4j.LoggerFactory;
+
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.util.List;
+
+import static io.servicetalk.utils.internal.ServiceLoaderUtils.loadProviders;
 
 /**
  * Factory methods for building HTTP Servers backed by {@link ServerContext}.
```
bondolo, line 32: The connection to `ServerContext` is not clear from reading this class.

bondolo, line 32: The connection to `ServerContext` is not clear from reading this class.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpServers.java
```diff
@@ -15,38 +15,62 @@
...
  */
 public final class HttpServers {
```
bondolo, line 34: Same comments as `GrpcServers` and `HttpClients`

bondolo, line 34: Same comments as `GrpcServers` and `HttpClients`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpProvidersTest.java
```diff
@@ -0,0 +1,246 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import static io.servicetalk.http.api.HttpResponseStatus.OK;
+import static io.servicetalk.http.netty.TestServiceStreaming.SVC_ECHO;
+import static io.servicetalk.transport.netty.internal.AddressUtils.localAddress;
+import static io.servicetalk.transport.netty.internal.AddressUtils.serverHostAndPort;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+class HttpProvidersTest {
```
bondolo, line 55: Same comments as `GrpcProvidersTest`.

bondolo, line 55: Same comments as `GrpcProvidersTest`.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcProviders.java
```diff
@@ -0,0 +1,74 @@
...
+
+import io.servicetalk.http.api.HttpProviders;
+import io.servicetalk.http.api.HttpProviders.HttpServerBuilderProvider;
+import io.servicetalk.http.api.HttpProviders.SingleAddressHttpClientBuilderProvider;
+
+import java.net.SocketAddress;
+import java.util.ServiceLoader;
+
+/**
+ * A holder for all gRPC-specific providers that can be registered using {@link ServiceLoader}.
```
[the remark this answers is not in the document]
idelpivnitskiy: Discussed offline. This class is a holder for whatever grpc-specific providers we will need in future. We won't list them in javadoc here, but I added a javadoc from each place where the provider applies for discoverability.

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/ServiceLoaderUtils.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Loads provider classes via {@link ServiceLoader}.
+     *
+     * @param clazz interface of abstract class which implementations should be loaded
+     * @param classLoader {@link ClassLoader} to use
+     * @param logger {@link Logger} to use
+     * @param <T> type of the provider
+     * @return a list of loaded providers for the specified class
+     */
+    public static <T> List<T> loadProviders(final Class<T> clazz, final ClassLoader classLoader, final Logger logger) {
```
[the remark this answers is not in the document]
idelpivnitskiy: Discussed offline, posting here for visibility:
This is an internal utility.
`ServiceLoader` implements `Iterable`. Will be more predictable to follow the result of the `ServiceLoader`, including the order. If it decides to load multiple equal instances (not sure if this is possible), we should preserve this behavior. Also, there are no requirements for providers to implement equals/hashCode.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcProviders.java
idelpivnitskiy: "provider" is a term defined by `ServiceLoader` javadoc. Since all "providers" are tight to `ServiceLoader`, using its terminology vs Spring terminology is more natural. Armeria also uses the name "provider".

idelpivnitskiy:
By some reason, comments were posted twice. I deleted duplicated comments

bondolo (approved):

PR #2141, opened 2022-03-10
Update Netty 4.1.74.Final -> 4.1.75.Final

chemicL:


Scottmitch:
blocked on: https://github.com/grpc/grpc-java/issues/8981. we can consider temporarily disabling these tests if necessary too.

PR #2148, opened 2022-03-15
Processors.newPublisherProcessor simplify duplicate terminal check

Scottmitch:
Motivation:
Processors.newPublisherProcessor checks in the thread generating
data if a terminal has already been observed, and discards duplicates.
However the consumer thread also has the same logic which is required
because we cannot deliver signals to the Subscriber after a terminal
signal is delivered.

Modifications:
- Keep the required check in the queue consumer thread which filters
  out duplicate terminal events, remove the check in the producer
  thread. If there is a strict limit applied to the queue size
  this may impact how many items the queue needs to hold.

Result:
Less atomic operations, 1 less class in the hierarchy, and less
memory consumed by Processors.newPublisherProcessor.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ProcessorBufferUtils.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ProcessorBufferUtils.java
@@ -17,22 +17,12 @@ package io.servicetalk.concurrent.api;
 
 import io.servicetalk.concurrent.internal.TerminalNotification;
 
-import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.api.SubscriberApiUtils.unwrapNullUnchecked;
-import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
 
-abstract class AbstractProcessorBuffer {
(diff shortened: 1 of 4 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ProcessorBufferUtils.java
```diff
@@ -17,22 +17,12 @@
...
 import static io.servicetalk.concurrent.api.SubscriberApiUtils.unwrapNullUnchecked;
-import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
 
-abstract class AbstractProcessorBuffer {
-    private static final AtomicReferenceFieldUpdater<AbstractProcessorBuffer,
-            TerminalNotification> terminalUpdater = newUpdater(AbstractProcessorBuffer.class,
-            TerminalNotification.class, "terminal");
-
-    @Nullable
-    private volatile TerminalNotification terminal;
```
Scottmitch, line 32: @NiteshKant - are there any hard requirements for filtering out duplicate terminals before the queue, or does this change make sense?

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1894

idelpivnitskiy (approved):
Good to have confirmation from @NiteshKant, otherwise LGTM

Scottmitch:
discussed with @NiteshKant offline. got lgtm and will merge.

PR #2150, opened 2022-03-16
Expose `SslConfig` through `ConnectionInfo`

idelpivnitskiy:
Motivation:

`SSLSession` represents the result of SSL negotiation. However, it's
useful to know what `SslConfig` was used to establish the session for
observability or runtime checks.

Modifications:

- Add `ConnectionInfo#sslConfig()` method;
- Propagate `SslConfig` to all `ConnectionInfo` implementations;
- Adjust tests;

Result:

Users have access to `SslConfig` from service or connection
factory/filter.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionInfo.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionInfo.java
@@ -40,6 +40,16 @@ public interface ConnectionInfo {
     SocketAddress remoteAddress();
 
     /**
+     * Get the {@link SslConfig} for this connection.
+     *
+     * @return The {@link SslConfig} if SSL/TLS is configured, or {@code null} otherwise.
+     */
+    @Nullable
+    default SslConfig sslConfig() { // FIXME: 0.43 - consider removing default impl
+        throw new UnsupportedOperationException(getClass().getSimpleName() + "does not implement sslConfig()");
+    }
+
+    /**
      * Get the {@link SSLSession} for this connection.
      *
      * @return The {@link SSLSession} if SSL/TLS is enabled, or {@code null} otherwise.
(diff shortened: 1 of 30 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionInfo.java
```diff
@@ -39,6 +39,16 @@
      */
     SocketAddress remoteAddress();
 
+    /**
+     * Get the {@link SslConfig} for this connection.
+     *
+     * @return The {@link SslConfig} if SSL/TLS is configured, or {@code null} otherwise.
+     */
+    @Nullable
+    default SslConfig sslConfig() { // FIXME: 0.43 - consider removing default impl
```
idelpivnitskiy, line 48: Added more reviewers for visibility of this message:

WDYT about exposing `sslConfig()` through `ExecutionContext` instead of `ConnectionInfo`? That way it will be automatically visible through `client.executionContext()` and `serverContext.executionContext()` too.

tkountis: We discussed this offline, and agreed that its more relevant as part of the `ConnectionInfo`.

tkountis (approved):
lean and clean. :shipit:

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionInfo.java
idelpivnitskiy: For visibility: client is an abstraction that does not know about the actual connections. There are implementations of the client (multi-address client) that can talk to HTTP and HTTPS targets. Therefore, having `SslConfig` at the client level does not make sense.
We can add `serverContext.sslConfig()` later, if we have a use-case for that. For now, will just keep it at `ConnectionInfo`.

Issue #1716, opened 2021-08-04
Flaky Test: DefaultHttpExecutionStrategyTest


Scottmitch:
`./gradlew clean, ./gradlew build`
`macOS 11.5, JDK 16.0.2+7`
local development of https://github.com/apple/servicetalk/pull/1673 based upon https://github.com/apple/servicetalk/commit/9ac4f9852d547fdc6939219b0e93d5f0367dda34 ... will rebase and re-evaluate.

```
[12] NO_EXEC_OFFLOAD_SEND

java.lang.AssertionError: Async errors occurred. See suppressed!
	at io.servicetalk.test.resources.TestUtils.assertNoAsyncErrors(TestUtils.java:57)
	at io.servicetalk.test.resources.TestUtils.assertNoAsyncErrors(TestUtils.java:42)
	at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest$ThreadAnalyzer.verifyNoErrors(DefaultHttpExecutionStrategyTest.java:424)
	at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest$ThreadAnalyzer.verifySend(DefaultHttpExecutionStrategyTest.java:408)
	at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest.offloadSendSingle(DefaultHttpExecutionStrategyTest.java:233)

...

Suppressed: java.lang.AssertionError: Unexpected thread requested from cancel. Thread: Thread[Test worker,5,main]
		at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest$ThreadAnalyzer.addError(DefaultHttpExecutionStrategyTest.java:428)
		at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest$ThreadAnalyzer.verifyThread(DefaultHttpExecutionStrategyTest.java:402)
		at io.servicetalk.http.api.DefaultHttpExecutionStrategyTest$ThreadAnalyzer.lambda$instrumentSend$11(DefaultHttpExecutionStrategyTest.java:320)
		at io.servicetalk.concurrent.api.ComposedCancellable.cancel(ComposedCancellable.java:33)
		at io.servicetalk.concurrent.api.ContextPreservingCancellable.slowPath(ContextPreservingCancellable.java:61)
		at io.servicetalk.concurrent.api.ContextPreservingCancellable.cancel(ContextPreservingCancellable.java:53)
[...]

PR #2151, opened 2022-03-17
Add Publisher.flatMapConcatSingle

Scottmitch:
Motivation:
Publisher.flatMapConcatSingle provides similar semantics to
Publisher.flatMapMergeSingle except the results are returned
in the same order as the original Publisher.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -886,13 +886,12 @@ public abstract class Publisher<T> {
 
     /**
      * Map each element of this {@link Publisher} into a {@link Single}&lt;{@link R}&gt; and flatten all signals
-     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned
-     * {@link Publisher}&lt;{@link R}&gt;.
+     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned {@link Publisher}&lt;{@link R}&gt;.
      * <p>
      * To control the amount of concurrent processing done by this operator see
      * {@link #flatMapMergeSingle(Function, int)}.
      * <p>
-     * This method is similar to {@link #map(Function)} but the result is asynchronous, and provides a data
+     * This method is similar to {@link #map(Function)} but the result is asynchronous, unordered, and provides a data
      * transformation in sequential programming similar to:
      * <pre>{@code
      *     ExecutorService e = ...;
@@ -925,10 +924,9 @@ public abstract class Publisher<T> {
 
     /**
      * Map each element of this {@link Publisher} into a {@link Single}&lt;{@link R}&gt; and flatten all signals
-     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned
-     * {@link Publisher}&lt;{@link R}&gt;.
+     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned {@link Publisher}&lt;{@link R}&gt;.
      * <p>
-     * This method is similar to {@link #map(Function)} but the result is asynchronous, and provides a data
+     * This method is similar to {@link #map(Function)} but the result is asynchronous, unordered, and provides a data
      * transformation in sequential programming similar to:
      * <pre>{@code
      *     ExecutorService e = ...;
@@ -964,51 +962,45 @@ public abstract class Publisher<T> {
 
     /**
      * Map each element of this {@link Publisher} into a {@link Single}&lt;{@link R}&gt; and flatten all signals
-     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned
-     * {@link Publisher}&lt;{@link R}&gt;.
+     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned {@link Publisher}&lt;{@link R}&gt;.
      * <p>
      * The behavior is the same as {@link #flatMapMergeSingle(Function, int)} with the exception that if any
      * {@link Single} returned by {@code mapper}, terminates with an error, the returned {@link Publisher} will not
      * immediately terminate. Instead, it will wait for this {@link Publisher} and all {@link Single}s to terminate and
-     * then terminate the returned {@link Publisher} with all errors emitted by the {@link Single}s produced by the
+     * then terminate the returned {@link Publisher} with errors emitted by the {@link Single}s produced by the
      * {@code mapper}.
      * <p>
      * To control the amount of concurrent processing done by this operator see
      * {@link #flatMapMergeSingleDelayError(Function, int)}.
      * <p>
-     * This method is similar to {@link #map(Function)} but the result is asynchronous, and provides a data
+     * This method is similar to {@link #map(Function)} but the result is asynchronous, unordered, and provides a data
      * transformation in sequential programming similar to:
      * <pre>{@code
-     *     Executor e = ...;
-     *     List<T> tResults = resultOfThisPublisher();
-     *     List<R> rResults = ...; // assume this is thread safe
-     *     List<Throwable> errors = ...;  // assume this is thread safe
-     *     CountDownLatch latch =  new CountDownLatch(tResults.size());
-     *     for (T t : tResults) {
+     *     ExecutorService e = ...;
+     *     List<Future<R>> futures = ...; // assume this is thread safe
+     *     for (T t : resultOfThisPublisher()) {
      *         // Note that flatMap process results in parallel.
-     *         e.execute(() -> {
-     *             try {
-     *                 R r = mapper.apply(t); // Asynchronous result is flatten into a value by this operator.
-     *                 rResults.add(r);
-     *             } catch (Throwable cause) {
-     *                 errors.add(cause);  // Asynchronous error is flatten into an error by this operator.
-     *             } finally {
-     *                 latch.countdown();
-     *             }
-     *         });
+     *         futures.add(e.submit(() -> {
+     *             processResult(mapper.apply(t));
+     *             return null;
+     *         }));
      *     }
-     *     latch.await();
-     *     if (errors.isEmpty()) {
-     *         return rResults;
+     *     List<Throwable> errors = ...;
+     *     for (Future<R> future : futures) {
+     *         try {
+     *           future.get(); // Throws if the processing for this item failed.
+     *         } catch (Throwable cause) {
+     *           errors.add(cause);
+     *         }
      *     }
-     *     createAndThrowACompositeException(errors);
+     *     throwExceptionIfNotEmpty(errors);
      * }</pre>
      *
      * @param mapper {@link Function} to convert each item emitted by this {@link Publisher} into a {@link Single}.
      * @param <R> Type of items emitted by the returned {@link Publisher}.
      * @return A new {@link Publisher} that emits all items emitted by each single produced by {@code mapper}.
      *
-     * @see <a href="http://reactivex.io/documentation/operators/merge.html">ReactiveX merge operator.</a>
+     * @see <a href="https://reactivex.io/documentation/operators/flatmap.html">ReactiveX flatMap operator.</a>
      * @see #flatMapMergeSingleDelayError(Function, int)
      */
     public final <R> Publisher<R> flatMapMergeSingleDelayError(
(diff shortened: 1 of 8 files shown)


Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/1716

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatUtils.java
```diff
@@ -0,0 +1,148 @@
...
+import static java.lang.Math.min;
+
+final class PublisherFlatMapConcatUtils {
+    private PublisherFlatMapConcatUtils() {
+    }
+
+    static <T, R> Publisher<R> flatMapConcatSingle(final Publisher<T> publisher,
+                                                   final Function<? super T, ? extends Single<? extends R>> mapper) {
+        return defer(() -> {
+            final Queue<Item<R>> results = newUnboundedSpscQueue(4);
```
Scottmitch, line 40: note there is a queue here, and in `flatMapMergeSingle`. we can optimize and combine in the future with a native operator implementation but for now reusing `flatMapMergeSingle` is good enough to get started and keeps complexity down.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -886,13 +886,12 @@ protected Publisher() {
...
      * Map each element of this {@link Publisher} into a {@link Single}&lt;{@link R}&gt; and flatten all signals
-     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned
-     * {@link Publisher}&lt;{@link R}&gt;.
+     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned {@link Publisher}&lt;{@link R}&gt;.
      * <p>
      * To control the amount of concurrent processing done by this operator see
      * {@link #flatMapMergeSingle(Function, int)}.
      * <p>
-     * This method is similar to {@link #map(Function)} but the result is asynchronous, and provides a data
+     * This method is similar to {@link #map(Function)} but the result is asynchronous, unordered, and provides a data
```
bondolo, line 894: The asynchronous and unordered characteristic should be covered in the first sentence to distinguish this operation from `map()`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -1344,6 +1321,166 @@ public final Completable flatMapCompletableDelayError(Function<? super T, ? exte
...
+    /**
+     * Map each element of this {@link Publisher} into a {@link Single}&lt;{@link R}&gt; and flatten all signals
+     * emitted from each mapped {@link Single}&lt;{@link R}&gt; into the returned {@link Publisher}&lt;{@link R}&gt;.
+     * Each mapped {@link Single}&lt;{@link R}&gt; maybe subscribed to concurrently but the results are emitted in
+     * the same order as this {@link Publisher}.
+     * <p>
+     * To control the amount of concurrent processing done by this operator see
+     * {@link #flatMapConcatSingle(Function, int)}.
+     * <p>
+     * This method is similar to {@link #map(Function)} but the result is asynchronous, and provides a data
```
bondolo, line 1333: Perhaps contrast with `flatMapMergeSingle` or `@see flatMapMergeSingle`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatUtils.java
```diff
@@ -0,0 +1,148 @@
...
+            final AtomicInteger consumerLock = new AtomicInteger();
+            return publisher.flatMapMergeSingle(orderedMapper(mapper, results, consumerLock))
+                    .shareContextOnSubscribe();
+        });
```
bondolo, line 44: perhaps just call `flatmapConcat(publisher, mapper, 4)` except for the lower bounds issue…

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatUtils.java
```diff
@@ -0,0 +1,148 @@
...
+            return publisher.flatMapMergeSingleDelayError(orderedMapper(mapper, results, consumerLock))
+                    .shareContextOnSubscribe();
+        });
+    }
+
+    static <T, R> Publisher<R> flatMapConcatSingle(final Publisher<T> publisher,
+                                                   final Function<? super T, ? extends Single<? extends R>> mapper,
+                                                   final int maxConcurrency) {
+        return defer(() -> {
+            final Queue<Item<R>> results = newUnboundedSpscQueue(min(8, maxConcurrency));
```
bondolo, line 61: this lower bounds is not mentioned in the Publisher docs.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/ConcurrentUtils.java
```diff
@@ -78,6 +79,36 @@ private ConcurrentUtils() {
...
 
+    /**
+     * Acquire a lock that is exclusively held with no re-entry, but attempts to acquire the lock while it is
+     * held can be detected by {@link #releaseLock(AtomicInteger)}.
+     * @param lock The {@link AtomicInteger} used to control the lock state.
+     * @return {@code true} if the lock was acquired, {@code false} otherwise.
+     */
+    public static boolean tryAcquireLock(AtomicInteger lock) {
+        for (;;) {
+            final int prevEmitting = lock.get();
```
bondolo, line 90: Perhaps`get()` → `intValue()`. When I see `get()` I always think "reference".

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -964,51 +962,45 @@ protected Publisher() {
...
+     *             return null;
+     *         }));
      *     }
-     *     latch.await();
-     *     if (errors.isEmpty()) {
-     *         return rResults;
+     *     List<Throwable> errors = ...;
+     *     for (Future<R> future : futures) {
+     *         try {
+     *           future.get(); // Throws if the processing for this item failed.
```
idelpivnitskiy, line 991: Because merge is unordered, consider using [CompletableFuture.allOf(...)](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#allOf-java.util.concurrent.CompletableFuture...-) to demonstrate that.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatSingleTest.java
```diff
@@ -0,0 +1,149 @@
...
+    static void beforeClass() {
+        executor = newCachedThreadExecutor();
+    }
+
+    @AfterAll
+    static void afterClass() throws Exception {
+        executor.closeAsync().toFuture().get();
+    }
+
+    @ParameterizedTest
```
idelpivnitskiy, line 51: Consider:
```
@ParameterizedTest(name = "{displayName} [{index}] begin={0} end={1} maxConcurrency={2} delayError={3}")
```

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatSingleTest.java
```diff
@@ -0,0 +1,149 @@
...
+    @CsvSource(value = {"0,5,10,false", "0,25,10,false", "0,5,10,true", "0,25,10,true"})
+    void orderPreservedInRangeWithConcurrency(int begin, int end, int maxConcurrency, boolean delayError) {
+        toSource(concatSingle(Publisher.range(begin, end),
+                i -> executor.timer(ofMillis(1)).toSingle().map(__ -> i + "x"), maxConcurrency, delayError)
```
idelpivnitskiy, line 55: For all tests: to make sure each `Single` completes at different time in different order, consider using a random delay <=5ms. Currently, there is a risk that ordering is guaranteed by having the same timer delay.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/ConcurrentUtils.java
```diff
@@ -78,6 +79,36 @@ private ConcurrentUtils() {
         return lockUpdater.getAndSet(owner, CONCURRENT_IDLE) == CONCURRENT_EMITTING;
     }
 
+    /**
+     * Acquire a lock that is exclusively held with no re-entry, but attempts to acquire the lock while it is
+     * held can be detected by {@link #releaseLock(AtomicInteger)}.
+     * @param lock The {@link AtomicInteger} used to control the lock state.
+     * @return {@code true} if the lock was acquired, {@code false} otherwise.
+     */
+    public static boolean tryAcquireLock(AtomicInteger lock) {
```
idelpivnitskiy, line 88: If we use named `Runnable`(s) (static inner classes) in `PublisherFlatMapConcatUtils` we can reuse existing utilities that operate with `AtomicIntegerFieldUpdater` and save some memory.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatSingleTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+import static io.servicetalk.concurrent.api.Executors.newCachedThreadExecutor;
+import static io.servicetalk.concurrent.api.SourceAdapters.toSource;
+import static io.servicetalk.concurrent.internal.DeliberateException.DELIBERATE_EXCEPTION;
+import static java.time.Duration.ofMillis;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.sameInstance;
+
+class PublisherFlatMapConcatSingleTest {
```
idelpivnitskiy, line 36: Should not block this PR, but will be nice to have a test that verifies AsyncContext is propagated correctly to make sure all `shareContextOnSubscribe()` operators placed correctly.

idelpivnitskiy (approved):
LGTM after comments :)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatUtils.java
Scottmitch: The goal is to inherit the limits/bounds/default from `flatMapMergeSingle` unless explicitly specified.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatUtils.java
Scottmitch: The queue is unbound, the parameter is just an estimate for initial size of the queue. This value is not appropriate for javadocs as it is not part of the API contract and can be adjusted if necessary.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
Scottmitch: `delayError` error propagation is different than `CompletableFuture.allOf(..)` because it may potentially return all exceptions (e.g. not bail on the first one).

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/ConcurrentUtils.java
Scottmitch: good catch! named `Function` will do the trick in this case based upon where state lives.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
Scottmitch: I'll add a `@see flatMapMergeSingle` for now to keep javadocs compressed.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/ConcurrentUtils.java
Scottmitch: I'll just leave it as `get()` bcz that is the accessor for `Atomic*` types

PR #2155, opened 2022-03-18
Add `publishOn`/`subscribeOn` `Executor` overload

bondolo:
Motivation:
The signature of `publishOn(Executor)` and `subscribeOn(Executor)`
methods changed between 0.41 and 0.42 from
`io.servicetalk.concurrent.api.Executor` to the more general parent
type `io.servicetalk.concurrent.api.Executor`. This is a source
compatible change since the new type is a parent of the old type, but
is not a binary compatible change; code compiled against 0.41 will not
run with 0.42 as it will not find the method it expects with the
`io.servicetalk.concurrent.api.Executor` signature.
Modifications:
Add overloads `publishOn(Executor)` and `subscribeOn(Executor)` for
`io.servicetalk.concurrent.Executor`. For binary compatibility with
0.42 it will also be necessary to cast executors to
`io.servicetalk.concurrent.api.Executor` to ensure that the correct
overload is used.
Result:
Improved compatibility with ServiceTalk 0.42

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
@@ -1544,6 +1544,23 @@ public abstract class Completable {
     /**
      * Creates a new {@link Completable} that will use the passed {@link Executor} to invoke all {@link Subscriber}
      * methods.
+     * This method does <strong>not</strong> override preceding {@link Executor}s, if any, specified for {@code this}
+     * {@link Completable}. Only subsequent operations, if any, added in this execution chain will use this
+     * {@link Executor}. If such an override is required, {@link #publishOnOverride(Executor)} can be used.
+     *
+     * @param executor {@link Executor} to use.
+     * @return A new {@link Completable} that will use the passed {@link Executor} to invoke all methods on the
+     * {@link Subscriber}.
+     * @deprecated This method is provided for 0.42 compatibility
+     */
+    @Deprecated
+    public final Completable publishOn(io.servicetalk.concurrent.Executor executor) {
+        return PublishAndSubscribeOnCompletables.publishOn(this, (Executor) executor);
+    }
+
+    /**
+     * Creates a new {@link Completable} that will use the passed {@link Executor} to invoke all {@link Subscriber}
+     * methods.
      * This method overrides preceding {@link Executor}s, if any, specified for {@code this} {@link Completable}.
      * That is to say preceding and subsequent operations for this execution chain will use this {@link Executor}.
      * If such an override is not required, {@link #publishOn(Executor)} can be used.
@@ -1557,6 +1574,26 @@ public abstract class Completable {
     @Deprecated
     public final Completable publishOnOverride(Executor executor) {
         return PublishAndSubscribeOnCompletables.publishOnOverride(this, executor);
+    }
+
+    /**
+     * Creates a new {@link Completable} that will use the passed {@link Executor} to invoke the following methods:
+     * <ul>
+     *     <li>All {@link Cancellable} methods.</li>
+     *     <li>The {@link #handleSubscribe(CompletableSource.Subscriber)} method.</li>
+     * </ul>
+     * This method does <strong>not</strong> override preceding {@link Executor}s, if any, specified for {@code this}
+     * {@link Completable}. Only subsequent operations, if any, added in this execution chain will use this
+     * {@link Executor}. If such an override is required, {@link #subscribeOnOverride(Executor)} can be used.
+     *
+     * @param executor {@link Executor} to use.
+     * @return A new {@link Completable} that will use the passed {@link Executor} to invoke all methods of
+     * {@link Cancellable} and {@link #handleSubscribe(CompletableSource.Subscriber)}.
+     * @deprecated This method is provided for 0.42 compatibility
+     */
+    @Deprecated
+    public final Completable subscribeOn(io.servicetalk.concurrent.Executor executor) {
+        return PublishAndSubscribeOnCompletables.subscribeOn(this, (Executor) executor);
     }
 
     /**
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -2874,6 +2874,23 @@ public abstract class Publisher<T> {
     /**
      * Creates a new {@link Publisher} that will use the passed {@link Executor} to invoke all {@link Subscriber}
      * methods.
+     * This method does <strong>not</strong> override preceding {@link Executor}s, if any, specified for {@code this}
+     * {@link Publisher}. Only subsequent operations, if any, added in this execution chain will use this
+     * {@link Executor}. If such an override is required, {@link #publishOnOverride(Executor)} can be used.
+     *
+     * @param executor {@link Executor} to use.
+     * @return A new {@link Publisher} that will use the passed {@link Executor} to invoke all methods of
+     * {@link Subscriber}.
+     * @deprecated This method is provided for 0.42 compatibility
+     */
+    @Deprecated
+    public final Publisher<T> publishOn(io.servicetalk.concurrent.Executor executor) {
+        return PublishAndSubscribeOnPublishers.publishOn(this, (Executor) executor);
+    }
+
+    /**
+     * Creates a new {@link Publisher} that will use the passed {@link Executor} to invoke all {@link Subscriber}
+     * methods.
      * This method overrides preceding {@link Executor}s, if any, specified for {@code this} {@link Publisher}.
      * That is to say preceding and subsequent operations for this execution chain will use this {@link Executor}.
      * If such an override is not required, {@link #publishOn(Executor)} can be used.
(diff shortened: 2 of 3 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1541,6 +1541,23 @@ public final Completable publishOn(Executor executor) {
...
+     * This method does <strong>not</strong> override preceding {@link Executor}s, if any, specified for {@code this}
+     * {@link Completable}. Only subsequent operations, if any, added in this execution chain will use this
+     * {@link Executor}. If such an override is required, {@link #publishOnOverride(Executor)} can be used.
+     *
+     * @param executor {@link Executor} to use.
+     * @return A new {@link Completable} that will use the passed {@link Executor} to invoke all methods on the
+     * {@link Subscriber}.
+     * @deprecated This method is provided for 0.42 compatibility
+     */
+    @Deprecated
```
Scottmitch, line 1556: this method shouldn't be deprecated IIUC bcz it still exists in 0.42. The `publishOn (io.servicetalk.concurrent.api.Executor)` variant above should be deprecated and point to this method.

Scottmitch: https://github.com/apple/servicetalk/blob/main/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java#L1423-L1440

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1559,6 +1576,26 @@ public final Completable publishOnOverride(Executor executor) {
...
+     *     <li>The {@link #handleSubscribe(CompletableSource.Subscriber)} method.</li>
+     * </ul>
+     * This method does <strong>not</strong> override preceding {@link Executor}s, if any, specified for {@code this}
+     * {@link Completable}. Only subsequent operations, if any, added in this execution chain will use this
+     * {@link Executor}. If such an override is required, {@link #subscribeOnOverride(Executor)} can be used.
+     *
+     * @param executor {@link Executor} to use.
+     * @return A new {@link Completable} that will use the passed {@link Executor} to invoke all methods of
+     * {@link Cancellable} and {@link #handleSubscribe(CompletableSource.Subscriber)}.
+     * @deprecated This method is provided for 0.42 compatibility
```
idelpivnitskiy, line 1592: Consider linking which method should be used instead

Scottmitch: +1 https://github.com/apple/servicetalk/pull/2155/files#r830220534 -> this deprecation go away, and the linking should be from `io.servicetalk.concurrent.api.Executor` methods to `io.servicetalk.concurrent.Executor` methods (general comment in this PR).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1573,6 +1592,28 @@ public final Completable publishOnOverride(Executor executor) {
...
+     *     <li>The {@link #handleSubscribe(CompletableSource.Subscriber)} method.</li>
+     * </ul>
+     * This method does <strong>not</strong> override preceding {@link Executor}s, if any, specified for {@code this}
+     * {@link Completable}. Only subsequent operations, if any, added in this execution chain will use this
+     * {@link Executor}. If such an override is required, {@link #subscribeOnOverride(Executor)} can be used.
+     *
+     * @param executor {@link Executor} to use.
+     * @return A new {@link Completable} that will use the passed {@link Executor} to invoke all methods of
+     * {@link Cancellable} and {@link #handleSubscribe(CompletableSource.Subscriber)}.
+     * @deprecated For compatibility with 0.42 convert to using {@link #publishOn(io.servicetalk.concurrent.Executor)}
```
Scottmitch, line 1612: `publishOn` -> `subscribeOn`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2902,6 +2921,28 @@ public final Cancellable forEach(Consumer<? super T> forEach) {
...
+     *     <li>The {@link #handleSubscribe(PublisherSource.Subscriber)} method.</li>
+     * </ul>
+     * This method does <strong>not</strong> override preceding {@link Executor}s, if any, specified for {@code this}
+     * {@link Publisher}. Only subsequent operations, if any, added in this execution chain will use this
+     * {@link Executor}. If such an override is required, {@link #subscribeOnOverride(Executor)} can be used.
+     *
+     * @param executor {@link Executor} to use.
+     * @return A new {@link Publisher} that will use the passed {@link Executor} to invoke all methods of
+     * {@link Subscription} and {@link #handleSubscribe(PublisherSource.Subscriber)}.
+     * @deprecated For compatibility with 0.42 convert to using {@link #publishOn(io.servicetalk.concurrent.Executor)}
```
Scottmitch, line 2941: `publishOn` -> `subscribeOn`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1545,6 +1563,28 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
...
+     *     <li>The {@link #handleSubscribe(SingleSource.Subscriber)} method.</li>
+     * </ul>
+     * This method does <strong>not</strong> override preceding {@link Executor}s, if any, specified for {@code this}
+     * {@link Single}. Only subsequent operations, if any, added in this execution chain will use this
+     * {@link Executor}. If such an override is required, {@link #subscribeOnOverride(Executor)} can be used.
+     *
+     * @param executor {@link Executor} to use.
+     * @return A new {@link Single} that will use the passed {@link Executor} to invoke all methods of
+     * {@link Cancellable} and {@link #handleSubscribe(SingleSource.Subscriber)}.
+     * @deprecated For compatibility with 0.42 convert to using {@link #publishOn(io.servicetalk.concurrent.Executor)}
```
Scottmitch, line 1583: `publishOn` -> `subscribeOn`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1537,6 +1537,25 @@ public final Completable liftAsync(CompletableOperator operator) {
...
+     * Creates a new {@link Completable} that will use the passed {@link Executor} to invoke all {@link Subscriber}
+     * methods.
+     * This method does <strong>not</strong> override preceding {@link Executor}s, if any, specified for {@code this}
+     * {@link Completable}. Only subsequent operations, if any, added in this execution chain will use this
+     * {@link Executor}. If such an override is required, {@link #publishOnOverride(Executor)} can be used.
+     *
+     * @param executor {@link Executor} to use.
+     * @return A new {@link Completable} that will use the passed {@link Executor} to invoke all methods on the
+     * {@link Subscriber}.
+     * @deprecated For compatibility with 0.42 convert to using {@link #publishOn(io.servicetalk.concurrent.Executor)}
```
Scottmitch, line 1554: `For compatibility with 0.42 convert to using {@link #publishOn(io.servicetalk.concurrent.Executor)} instead.` -> `Use {@link #publishOn(io.servicetalk.concurrent.Executor)}.`. Consider minimizing the dialog here, deprecation implies it is removed in future releases.

Scottmitch (changes requested):

Scottmitch:
feel free to merge after comments addressed.

Scottmitch (approved):
I didn't investigate all usages where casting would be required, but assume you covered them.

bondolo:
> I didn't investigate all usages where casting would be required, but assume you covered them.

I converted only one set of usages in the current tests to ensure that the new code was exercised, but otherwise left all of the other references in 0.41 using the deprecated method.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1537,6 +1537,24 @@ public final Completable liftAsync(CompletableOperator operator) {
      * @return A new {@link Completable} that will use the passed {@link Executor} to invoke all methods on the
      * {@link Subscriber}.
      */
+    public final Completable publishOn(io.servicetalk.concurrent.Executor executor) {
+        return publishOn((Executor) executor);
```
idelpivnitskiy, line 1541: I expected the opposite change: non-deprecated method should be the one that does the action, deprecated method should downcast from `api.Executor` to `io.servicetalk.concurrent.Executor` and invoke this method.

Otherwise, there is a risk of `ClassCastException` if users pass `io.servicetalk.concurrent.Executor` that does not implement `io.servicetalk.concurrent.api.Executor`

idelpivnitskiy (changes requested):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
idelpivnitskiy: Nvm, this is for 0.41 branch, not for main

idelpivnitskiy (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
bondolo: Using a downcast somewhere seems like the simplest approach and I don't think it is a particular risk. The alternative is to start changing all of the references in `PublishAndSubscribeOn` classes and all of the classes they touch. I looked at this briefly but decide it touched too many things and it does seem unlikely that there is a `io.servicetalk.concurrent.Executor` that is not a `io.servicetalk.concurrent.api.Executor` in current use.

idelpivnitskiy: Agreed, the risk is low

PR #2156, opened 2022-03-19
Use filter strategy for streaming async server

bondolo:
Motivation:
Currently the filter strategy is not considered in the computation of
the effective strategy for async streaming servers. This was done
because with a default builder strategy the result could be less
offloading which would be unexpected for that builder strategy.
Modifications:
Consider filter strategy in computing effective strategy for
async streaming servers.
Result:
More predictable offloading behavior.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
@@ -62,7 +62,6 @@ import javax.annotation.Nullable;
 
 import static io.servicetalk.http.api.HttpApiConversions.toStreamingHttpService;
 import static io.servicetalk.http.api.HttpExecutionStrategies.defaultStrategy;
-import static io.servicetalk.http.api.HttpExecutionStrategies.offloadAll;
 import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
 import static io.servicetalk.http.netty.StrategyInfluencerAwareConversions.toConditionalServiceFilterFactory;
 import static io.servicetalk.transport.api.ConnectionAcceptor.ACCEPT_ALL;
@@ -262,11 +261,7 @@ final class DefaultHttpServerBuilder implements HttpServerBuilder {
 
     @Override
     public Single<HttpServerContext> listenStreaming(final StreamingHttpService service) {
-        HttpExecutionStrategy serviceStrategy = requiredOffloads(service, defaultStrategy());
-        HttpExecutionStrategy filterStrategy = computeRequiredStrategy(serviceFilters, serviceStrategy);
-        HttpExecutionStrategy useStrategy = defaultStrategy() == strategy ? offloadAll() :
-                strategy.hasOffloads() ? strategy.merge(filterStrategy) : offloadNone();
-        return listenForService(service, useStrategy);
+        return listenForService(service, computeServiceStrategy(service));
     }
 
     @Override
@@ -365,7 +360,7 @@ final class DefaultHttpServerBuilder implements HttpServerBuilder {
         HttpExecutionStrategy serviceStrategy = requiredOffloads(service, defaultStrategy());
         HttpExecutionStrategy filterStrategy = computeRequiredStrategy(serviceFilters, serviceStrategy);
         return defaultStrategy() == strategy ? filterStrategy :
-                strategy.hasOffloads() ? strategy.merge(filterStrategy) : offloadNone();
+                strategy.hasOffloads() ? strategy.merge(filterStrategy) : strategy;
     }
 
     private static StreamingHttpService applyInternalFilters(StreamingHttpService service,
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
@@ -62,6 +62,7 @@ import javax.annotation.Nullable;
 
 import static io.servicetalk.http.api.HttpApiConversions.toStreamingHttpService;
 import static io.servicetalk.http.api.HttpExecutionStrategies.defaultStrategy;
+import static io.servicetalk.http.api.HttpExecutionStrategies.offloadAll;
 import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
 import static io.servicetalk.http.netty.StrategyInfluencerAwareConversions.toConditionalServiceFilterFactory;
 import static io.servicetalk.transport.api.ConnectionAcceptor.ACCEPT_ALL;
@@ -261,7 +262,11 @@ final class DefaultHttpServerBuilder implements HttpServerBuilder {
 
     @Override
     public Single<HttpServerContext> listenStreaming(final StreamingHttpService service) {
-        return listenForService(service, strategy);
+        HttpExecutionStrategy serviceStrategy = requiredOffloads(service, defaultStrategy());
+        HttpExecutionStrategy filterStrategy = computeRequiredStrategy(serviceFilters, serviceStrategy);
+        HttpExecutionStrategy useStrategy = defaultStrategy() == strategy ? offloadAll() :
+                strategy.hasOffloads() ? strategy.merge(filterStrategy) : offloadNone();
+        return listenForService(service, useStrategy);
     }
 
     @Override
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/BasicAuthStrategyInfluencerTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/BasicAuthStrategyInfluencerTest.java
@@ -31,7 +31,6 @@ import io.servicetalk.http.api.StreamingHttpResponseFactory;
 import io.servicetalk.http.api.StreamingHttpService;
 import io.servicetalk.http.utils.auth.BasicAuthHttpServiceFilter;
 import io.servicetalk.http.utils.auth.BasicAuthHttpServiceFilter.CredentialsVerifier;
-import io.servicetalk.transport.api.ExecutionStrategyInfluencer;
 import io.servicetalk.transport.api.HostAndPort;
 import io.servicetalk.transport.api.IoExecutor;
 import io.servicetalk.transport.api.ServerContext;
@@ -129,6 +128,7 @@ class BasicAuthStrategyInfluencerTest {
         when(credentialsVerifier.apply(anyString(), anyString())).thenReturn(succeeded("success"));
         when(credentialsVerifier.closeAsync()).thenReturn(completed());
         when(credentialsVerifier.closeAsyncGracefully()).thenReturn(completed());
+        when(credentialsVerifier.requiredOffloads()).thenCallRealMethod();
         CredentialsVerifier<String> verifier = credentialsVerifier;
         if (noOffloadsInfluence) {
             verifier = new InfluencingVerifier(verifier, offloadNone());
@@ -146,8 +146,7 @@ class BasicAuthStrategyInfluencerTest {
         return this.client;
     }
 
-    private static final class OffloadCheckingService implements StreamingHttpService,
-                                                                 ExecutionStrategyInfluencer<HttpExecutionStrategy> {
+    private static final class OffloadCheckingService implements StreamingHttpService {
 
         private enum OffloadPoint {
             ServiceHandle,
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ExecutionStrategyInContextTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ExecutionStrategyInContextTest.java
@@ -54,7 +54,6 @@ import javax.annotation.Nullable;
 import static io.servicetalk.concurrent.api.Single.succeeded;
 import static io.servicetalk.http.api.HttpExecutionStrategies.customStrategyBuilder;
 import static io.servicetalk.http.api.HttpExecutionStrategies.defaultStrategy;
-import static io.servicetalk.http.api.HttpExecutionStrategies.offloadAll;
 import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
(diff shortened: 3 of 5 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -356,11 +347,13 @@ private HttpExecutionContext buildExecutionContext(final HttpExecutionStrategy s
...
 
-    private HttpExecutionStrategy computeServiceStrategy(Object service) {
-        HttpExecutionStrategy serviceStrategy = requiredOffloads(service, defaultStrategy());
+    private HttpExecutionStrategy computeServiceStrategy(HttpExecutionStrategyInfluencer service) {
+        HttpExecutionStrategy serviceStrategy = service.requiredOffloads();
         HttpExecutionStrategy filterStrategy = computeRequiredStrategy(serviceFilters, serviceStrategy);
-        return defaultStrategy() == strategy ? filterStrategy :
-                strategy.hasOffloads() ? strategy.merge(filterStrategy) : offloadNone();
+        return defaultStrategy() == strategy ?
+            filterStrategy.hasOffloads() ? strategy.merge(filterStrategy) : strategy :
```
bondolo, line 354: I am tempted to replace the return of `strategy` when `filterStrategy` has no offloads with `offloadAll()`. Currently this results in `defaultStrategy` appearing the context which will merge differently with non-default execution strategy than `offloadAll()` and potentially result in additional offloading.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -261,7 +253,8 @@ public HttpServerBuilder bufferAllocator(final BufferAllocator allocator) {
 
     @Override
     public Single<HttpServerContext> listenStreaming(final StreamingHttpService service) {
```
idelpivnitskiy, line 255: Consider adding a test when users specify a custom strategy with one flag on the builder and another flag on the filter. Assert that the resulting strategy has both offloading flags set for any `listen*` method.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -356,11 +349,13 @@ private HttpExecutionContext buildExecutionContext(final HttpExecutionStrategy s
...
 
-    private HttpExecutionStrategy computeServiceStrategy(Object service) {
-        HttpExecutionStrategy serviceStrategy = requiredOffloads(service, defaultStrategy());
+    private HttpExecutionStrategy computeServiceStrategy(HttpExecutionStrategyInfluencer service) {
+        HttpExecutionStrategy serviceStrategy = service.requiredOffloads();
         HttpExecutionStrategy filterStrategy = computeRequiredStrategy(serviceFilters, serviceStrategy);
-        return defaultStrategy() == strategy ? filterStrategy :
-                strategy.hasOffloads() ? strategy.merge(filterStrategy) : offloadNone();
+        return defaultStrategy() == strategy ?
+            filterStrategy.hasOffloads() ? strategy.merge(filterStrategy) : strategy :
```
idelpivnitskiy, line 356: `strategy.merge(filterStrategy)` - this will result in a warning log printed by `DEFAULT_HTTP_EXECUTION_STRATEGY`. Should it be `filterStrategy.hasOffloads() ? filterStrategy : strategy`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -356,11 +349,13 @@ private HttpExecutionContext buildExecutionContext(final HttpExecutionStrategy s
...
+            // We currently treat the builder strategy specially, allowing it to override the filter strategy.
+            strategy.hasOffloads() ? strategy : offloadNone();
```
idelpivnitskiy, line 358: In this case, looks like by some reasons users passed a custom strategy to the builder that does not have offloading. As a user, I would expect to see my custom strategy on the `ServerContext`. `offloadNone()` will be surprising to me. Maybe I have a special subtype that carries more information/state.

bondolo: The strategy that is exposed is in the `ExecutionContext` currently almost always `OFFLOAD_EVENT_CLOSE_STRATEGY`  because there is always a non-offloading filter (`ClearAsyncContext`) and the real offloading is done by the `OffloadingFilter`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
bondolo: yes, well spotted. this can be just `filterStrategy`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
idelpivnitskiy: I mean that users can implement their own `HttpExecutionStrategy`, keep `hasOffloads()` as `false`, but have some other flags or state. Don't think replacing it with `offloadNone()` is reasonable. I would expect this to be:
```java
return defaultStrategy() == strategy ?
            (filterStrategy.hasOffloads() ? filterStrategy : strategy) :
            (filterStrategy.hasOffloads() ? strategy.merge(filterStrategy) : strategy);
```

idelpivnitskiy: Same but in different way:
```java
return filterStrategy.hasOffloads() ?
            (defaultStrategy() == strategy ? filterStrategy : strategy.merge(filterStrategy)) : strategy;
```

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -57,6 +57,11 @@
         this.original = requireNonNull(original);
     }
 
+    @Override
+    public HttpExecutionStrategy requiredOffloads() {
+        return original.requiredOffloads();
```
idelpivnitskiy, line 62: Hm... might be wrong, but it looks a bit inconsistent. Check out the ctor. It can optimize the strategy and pass the result to super class to be used for `serviceInvocationStrategy()`. The value for `requiredOffloads()` will be different. We may need to reconsider the whole approach with `serviceInvocationStrategy()` later.

Is this change required for this PR? If not, let's revert and work on it separately.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -289,12 +283,16 @@ private HttpExecutionContext buildExecutionContext(final HttpExecutionStrategy s
...
+     * @param apiStrategy The {@link HttpExecutionStrategy} provided by the user/API
      * @return A {@link Single} that completes when the server is successfully started or terminates with an error if
      * the server could not be started.
      */
     private Single<HttpServerContext> listenForService(final StreamingHttpService rawService,
-                                                       final HttpExecutionStrategy serviceStrategy) {
+                                                       final HttpExecutionStrategy apiStrategy) {
+        HttpExecutionStrategy serviceStrategy = computeServiceStrategy(rawService);
+        HttpExecutionStrategy serverStrategy =
+                apiStrategy.hasOffloads() ? apiStrategy.merge(serviceStrategy) : apiStrategy;
```
idelpivnitskiy, line 287: 1. Consider adding an assertion that `apiStrategy` is never `defaultStrategy()`
2. When `apiStrategy.hasOffloads() == false`, why do we ignore the result of `computeServiceStrategy`?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
bondolo: This change is a simplification to better handle the builder `defaultStrategy()` with the API conversions and each having their own default. The builder strategy is now passed to the API conversion and the default, if appropriate, is applied in the conversion. The conversion then returns either the appropriate default strategy or the builder strategy. The builder strategy overrides the default.

This is then later combined with the filter and service strategy in the `listenForService` which allows the service and filters to adjust the strategy and add additional offloading if needed.

idelpivnitskiy: Ok, can you please add some clarification comments in code to describe why `requiredOffloads()` is different from `serviceInvocationStrategy()`? 
We introduced `serviceInvocationStrategy()` in the past because `influenceStrategy` required an input strategy. With `requiredOffloads()` there is a change to remove `serviceInvocationStrategy()` in future.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -307,9 +307,28 @@ HttpServerBuilder appendServiceFilter(Predicate<StreamingHttpRequest> predicate,
...
+     *  builder:                service:                  filters:             Effective:
+     *  defaultStrategy()
+     *  (if service or filter needs additional offloading, add it)
+     *                          offloadNone()             offloadNone()        apiDefaultStrategy()
+     *                          customStrategy()          offloadNone()        apiDefaultStrategy() + customStrategy()
+     *                          offloadNone()             customStrategy()     apiDefaultStrategy() + customStrategy()
+     *  customStrategy()
+     *  (if service or filter needs additional offloading, add it)
+     *                          offloadNone()             offloadNone()        customStrategy()
+     *                          customStrategy()          offloadNone()        customStrategy() + customStrategy()
```
idelpivnitskiy, line 324: This change looks unrelated to the original problem this PR is trying to solve and increases review iterations. Consider taking it out to a separate PR.
These abbreviations are still hard to understand for new users.

bondolo: Sorry, I should have moved the PR in to draft before pushing those changes. I have replaced the PR with a different more minimal one.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
bondolo: The behaviour of the builder `executionStrategy()` is effectively to override the filter and service strategy. In this case the builder strategy was likely either `offloadNone()` or `offloadNever()` indicating that the user desired no offloading.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/BasicAuthStrategyInfluencerTest.java
```diff
@@ -146,8 +146,7 @@ private BlockingHttpClient setup(boolean noOffloadsInfluence) throws Exception {
         return this.client;
     }
 
-    private static final class OffloadCheckingService implements StreamingHttpService,
-                                                                 ExecutionStrategyInfluencer<HttpExecutionStrategy> {
```
bondolo, line 150: `ExecutionStrategyInfluencer<HttpExecutionStrategy>` is now redundant.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
bondolo: no longer applicable

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -261,7 +262,11 @@ public HttpServerBuilder bufferAllocator(final BufferAllocator allocator) {
 
     @Override
     public Single<HttpServerContext> listenStreaming(final StreamingHttpService service) {
-        return listenForService(service, strategy);
+        HttpExecutionStrategy serviceStrategy = requiredOffloads(service, defaultStrategy());
+        HttpExecutionStrategy filterStrategy = computeRequiredStrategy(serviceFilters, serviceStrategy);
+        HttpExecutionStrategy useStrategy = defaultStrategy() == strategy ? offloadAll() :
+                strategy.hasOffloads() ? strategy.merge(filterStrategy) : offloadNone();
```
idelpivnitskiy, line 268: Why does this have to be different from all other API variants? I mean, why don't we use `computeServiceStrategy` method here? Let's discuss and we should at least leave a comment in the code to clarify the difference.

My main concerning part is `defaultStrategy() == strategy ? offloadAll()`. If I'm a user who has 0 filters and passed `StreamingHttpService` impl that override `requireOffloads()` to `offloadNone()`, I won't expect `offloadAll()` by default.

bondolo: We can't use `computeServiceStrategy()` because it maps `defaultStrategy()` which we want to handle differently. The other three variants with adapters have their only handling for `defaultStrategy()` and won't ever trigger the `defaultStrategy() == strategy` in `computeServiceStrategy()`

> My main concerning part is `defaultStrategy() == strategy ? offloadAll()`. If I'm a user who has 0 filters and passed `StreamingHttpService` impl that override `requiredOffloads()` to `offloadNone()`, I won't expect `offloadAll()` by default.

Fair point. If the service does not override `requiredOffloads()` then the default `offloadAll()` it provides is sufficient to ensure safety otherwise builder `defaultStrategy()` allows whatever strategy the filter and service require.

With this change, and the additional change of returning `strategy` when `strategy.hasNoOffloads()` is true the result is that we can use `computeServiceStrategy()` everywhere.

idelpivnitskiy (approved):
Thank you!

PR #2157, opened 2022-03-19
servicetalk-grpc-protoc name conflict fixes

Scottmitch:
Motivation:
servicetalk-grpc-protoc attempts to detect naming conflicts
and appends numbers to the end of the class names to avoid
the conflict. However in the event `java_outer_classname` isn't
specified and the service type name conflicts with the file name
a conflict is incorrectly detected which leads to forcing a `0`
at the end of the service name.

Detection of implicit class name isn't consistent with
protobuf-java which may result in not generating code in the
same file when `java_multiple_files = false`.

Types (messages, enums) that start with a lower case value
would fail to compile due to CLassName.bestGuess restrictions.

Potential User Changes:
If you are in this scenario you must regenerate code and your
class names will change, requiring code changes.


servicetalk-grpc-protoc/src/test/proto/test_conflict_enum3.proto
```diff
@@ -0,0 +1,36 @@
...
+
+option java_multiple_files = true;
+option java_package = "io.servicetalk.grpc.protoc.test.conflict.enums";
+
+service TestConflictEnum3Service {
+  rpc DoNothing (TestConflictEnum3Req) returns (TestConflictEnum3Resp);
+}
+
+// todo(scott): make first char lowercase
+// https://github.com/protocolbuffers/protobuf/issues/9653
```
Scottmitch, line 27: once https://github.com/protocolbuffers/protobuf/issues/9653 is fixed, these tests can be updated to validate behavior.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/FileDescriptor.java
```diff
@@ -47,23 +48,24 @@
...
-    @Nullable
     private final String javaPackageName;
-    @Nullable
     private final String outerClassName;
     @Nullable
     private final String typeNameSuffix;
     private final List<TypeSpec.Builder> serviceClassBuilders;
-    @Nullable
-    private Set<String> reservedJavaTypeName;
+    private final Set<String> reservedJavaTypeName;
```
bondolo, line 68: Can do the assignment right here.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/FileDescriptor.java
```diff
@@ -227,13 +229,63 @@ private static void insertSingleFileContent(final String content, String fileNam
...
+    }
+
+    private static String inferOuterClassName(String sanitizedProtoFileName, FileDescriptorProto protoFile) {
+        final String sanitizeClassName = sanitizeClassName(sanitizedProtoFileName);
+        return hasConflictingClassName(sanitizeClassName, protoFile) ?
+                sanitizeClassName + OUTER_CLASS_SUFFIX : sanitizeClassName;
+    }
+
+    /**
+     * See <a href="https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/java/java_name_resolver.cc#L192-L214">java_name_resolver.cc</a>.
```
bondolo, line 246: same problem with line numbers and non-specific version of the referenced file.

bondolo (approved):

servicetalk-grpc-protoc/src/test/proto/test_conflict_enum.proto
```diff
@@ -0,0 +1,41 @@
...
+  rpc DoNothing (TestConflictEnumReq) returns (TestConflictEnumResp);
+  // This name will conflict with conflict with internal generated interface.
+  rpc DoNothingRpc (TestConflictEnumReq) returns (TestConflictEnumResp);
+}
+
+enum TestConflictEnum { // This name conflicts with the file name!
+  A = 0;
+}
+
+message DoNothingRpc {} // This name will conflict with conflict with internal generated interface.
```
idelpivnitskiy, line 35: "conflict with conflict with" -> "conflict with"

servicetalk-grpc-protoc/src/test/proto/test_conflict_enum.proto
```diff
@@ -0,0 +1,41 @@
...
+
+option java_multiple_files = true;
+option java_package = "io.servicetalk.grpc.protoc.test.conflict.enums";
+
+// No java_outer_classname provided to make sure the inferred outer classname doesn't clash
+// with the generated TestConflictMultiService inner interface
+
+service TestConflictEnumService {
+  rpc DoNothing (TestConflictEnumReq) returns (TestConflictEnumResp);
+  // This name will conflict with conflict with internal generated interface.
```
idelpivnitskiy, line 27: "conflict with conflict with" -> "conflict with"

idelpivnitskiy (commented):
Code changes LGTM.

> If you are in this scenario you must regenerate code and your class names will change, requiring code changes.

IIUC the risk is minimal: (1) not many users are in this situation; (2) if they are, they likely already use `java_outer_classname`; (3) if they don't, they can use `java_outer_classname` before upgrading to the next version to avoid any code changes as part of the version bump. Right?

Trying to see if it's ok to merge into 0.42.x or defer until 0.43

Scottmitch:
@idelpivnitskiy - yes I agree with your assessment that the risk is minimal. Relying on default behavior by not specifying the name seem risky and you can even see that protobuf compiler has some uncovered bugs in this department.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/FileDescriptor.java
```diff
@@ -47,23 +48,24 @@
  * A single protoc file for which we will be generating classes
  */
 final class FileDescriptor implements GenerationContext {
-    private static final String GENERATED_BY_COMMENT = "Generated by ServiceTalk proto compiler";
-
+    private static final String GENERATED_BY_COMMENT = "Generated by ServiceTalk gRPC protoc plugin";
+    /**
+     * Inferred behavior from protobuf-java is that if no suffix is explicitly provided the root file name will have
+     * this suffix. See
+     * <a href="https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/java/java_name_resolver.cc#L49">java_name_resolver.cc</a>
```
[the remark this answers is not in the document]
Scottmitch: good catch! link now points to a tag.

PR #2159, opened 2022-03-19
Jersey Protobuf Serialization Support

Scottmitch:
Motivation:
ServiceTalk has protobuf serializers but they
aren't integrated into Jersey. This PR provides
that integration.

Modifications:
- Introduce `servicetalk-data-protobuf-jersey` module.
  Put this on your classpath to get protobuf serialization.
- Introduce VarInt serialization header types so that
  VarInt encoding doesn't get confused with non-framed
  protobuf encoding.

Result:
Jersey users can now enjoy ServiceTalk protobuf serialization.


servicetalk-data-protobuf-jersey/docs/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,27 @@
+= JSON (Jackson)
```
Scottmitch, line 1: reminder: this needs to be updated

servicetalk-examples/http/serialization/protobuf/src/main/java/io/servicetalk/examples/http/serialization/protobuf/SerializerUtils.java
```diff
@@ -40,28 +41,30 @@
...
+            headers -> hasContentType(headers, APPLICATION_PROTOBUF_VAR_INT, null);
 
     public static final HttpSerializerDeserializer<RequestMessage> REQ_SERIALIZER =
             serializer(PROTOBUF.serializerDeserializer(RequestMessage.parser()),
                     CONTENT_TYPE_SETTER, CONTENT_TYPE_VALIDATOR);
 
     public static final HttpStreamingSerializerDeserializer<RequestMessage> REQ_STREAMING_SERIALIZER =
             streamingSerializer(PROTOBUF.streamingSerializerDeserializer(RequestMessage.parser()),
-                    CONTENT_TYPE_SETTER, CONTENT_TYPE_VALIDATOR);
+                    VAR_INT_CONTENT_TYPE_SETTER, VAR_INT_CONTENT_TYPE_VALIDATOR);
```
Scottmitch, line 59: @idelpivnitskiy - note this change. Streaming serialization with VarInt encoding is not the same as encoding/decoding single objects. Having a unique content type makes the exception more clear if you mix the examples (streaming client, scalar server).

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufMediaTypes.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public static final String APPLICATION_PROTOBUF = "application/x-protobuf";
+    /**
+     * {@code "application/x-protobuf-var-int"}
+     * <a href="https://developers.google.com/protocol-buffers/docs/encoding">base 128 VarInt protobuf encoding</a>.
+     * Should be used with {@link Publisher} to stream data.
+     * @see #APPLICATION_PROTOBUF_VAR_INT_TYPE
+     * @see HttpHeaderValues#APPLICATION_PROTOBUF_VAR_INT
+     */
+    public static final String APPLICATION_PROTOBUF_VAR_INT = "application/x-protobuf-var-int";
```
Scottmitch, line 40: I couldn't find a canonical content type to use in this scenario, the motivation is when using streaming the encoding (var-int) is different than scalar (raw single object). the different maybe subtle for people, but at least we should have exception messages to indicate the content type isn't compatible if people try to mix them.

@chemicL @bondolo - have you come across this scenario in spring use cases?

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufSerializerMessageBodyReaderWriter.java
```diff
@@ -0,0 +1,240 @@
...
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    @Override
+    public void writeTo(final Object o, final Class<?> type, final Type genericType,
+                        final Annotation[] annotations,
+                        final MediaType mediaType, final MultivaluedMap<String, Object> httpHeaders,
+                        final OutputStream entityStream) throws WebApplicationException {
+        final BufferAllocator allocator = ctxRefProvider.get().get().executionContext().bufferAllocator();
+        final Publisher<Buffer> bufferPublisher;
+        if (o instanceof Single) {
+            final Class<? extends MessageLite> clazz = genericType instanceof Class ? (Class) genericType :
```
bondolo, line 141: Why can it be assumed that the result will extend `MessageLite`? Is the `isWritable` sufficient? Perhaps worth an `assert class.isAssignableFrom(MessageLite.class) : "non-protobuf class class";`?

servicetalk-data-protobuf/build.gradle
```diff
@@ -31,14 +31,14 @@ dependencies {
   api project(":servicetalk-buffer-api")
   api project(":servicetalk-concurrent-api")
   api project(":servicetalk-serializer-api")
+  api "com.google.protobuf:protobuf-java:$protobufVersion"
```
bondolo, line 34: Consider using the [protobuf-bom](https://mvnrepository.com/artifact/com.google.protobuf/protobuf-bom)

servicetalk-data-protobuf/src/main/java/io/servicetalk/data/protobuf/ProtobufSerializerFactory.java
```diff
@@ -54,6 +60,18 @@ private ProtobufSerializerFactory() {
...
+     */
+    public <T extends MessageLite> SerializerDeserializer<T> serializerDeserializer(Class<T> clazz) {
+        @SuppressWarnings("unchecked")
+        final Parser<T> parser = (Parser<T>) parserMap.get(clazz);
+        return parser == null ? serializerDeserializer(newParser(clazz)) : serializerDeserializer(parser);
+    }
+
     /**
      * Get a {@link StreamingSerializerDeserializer} which supports &lt;VarInt length, value&gt; encoding as described
      * in <a href="https://developers.google.com/protocol-buffers/docs/techniques">Protobuf Streaming</a>.
```
bondolo, line 77: This explanation would also be helpful to accompany the top-level mime-type docs.

servicetalk-data-protobuf/src/main/java/io/servicetalk/data/protobuf/ProtobufSerializerFactory.java
```diff
@@ -70,4 +88,38 @@ private ProtobufSerializerFactory() {
...
+     * in <a href="https://developers.google.com/protocol-buffers/docs/techniques">Protobuf Streaming</a>.
+     * @param clazz Used to obtain a {@link Parser} which is used to serialize and deserialize.
+     * @param <T> The type to serialize and deserialize.
+     * @return a {@link StreamingSerializerDeserializer} which supports &lt;VarInt length, value&gt; encoding as
+     * described in <a href="https://developers.google.com/protocol-buffers/docs/techniques">Protobuf Streaming</a>.
+     * @see VarIntLengthStreamingSerializer
+     */
+    public <T extends MessageLite> StreamingSerializerDeserializer<T> streamingSerializerDeserializer(Class<T> clazz) {
+        @SuppressWarnings("unchecked")
+        final Parser<T> parser = (Parser<T>) parserMap.get(clazz);
```
bondolo, line 103: Perhaps encapsulate the `parserMap.get` in `newParser` and use `computeIfAbsent`?

servicetalk-data-protobuf/src/main/java/io/servicetalk/data/protobuf/ProtobufSerializerFactory.java
```diff
@@ -54,6 +60,18 @@ private ProtobufSerializerFactory() {
...
 
+    /**
+     * Get a {@link SerializerDeserializer}.
+     * @param clazz Used to obtain a {@link Parser} which is used to serialize and deserialize.
+     * @param <T> The type to serialize and deserialize.
+     * @return a {@link SerializerDeserializer}.
+     */
+    public <T extends MessageLite> SerializerDeserializer<T> serializerDeserializer(Class<T> clazz) {
+        @SuppressWarnings("unchecked")
+        final Parser<T> parser = (Parser<T>) parserMap.get(clazz);
```
bondolo, line 71: There is a race here than can result in more than on instance of `Parser<T>` for a given class, but normally just the one cached instance.

servicetalk-examples/http/jaxrs/build.gradle
```diff
@@ -25,11 +32,24 @@ dependencies {
   // input stream adaptation that kicks in with out-of-the-box body readers and also allows accepting/returning
   // Single<Pojo> and Publisher<Pojo> from resource methods.
   implementation project(":servicetalk-data-jackson-jersey")
+  implementation project(":servicetalk-data-protobuf-jersey")
   implementation project(":servicetalk-http-netty")
   implementation project(":servicetalk-http-router-jersey")
+  implementation project(":servicetalk-data-protobuf") // only for ProtobufClient
+  implementation "com.google.protobuf:protobuf-java:$protobufVersion"
```
bondolo, line 39: Perhaps use protobuf-pom (see above)

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ServiceTalkProtobufSerializerFeature.java
```diff
@@ -0,0 +1,76 @@
...
+                ST_PROTOBUF_FEATURE, PROTOBUF_FEATURE, String.class);
+
+        // Do not register our Protobuf feature if another one is already registered
+        if (!PROTOBUF_FEATURE.equalsIgnoreCase(protobufFeature)) {
+            LOGGER.warn("Skipping registration of: {} as Protobuf support is already provided by: {}",
+                    ST_PROTOBUF_FEATURE, protobufFeature);
+            return false;
+        }
+
+        // Prevent other not yet registered JSON features to register themselves
```
idelpivnitskiy, line 57: JSON -> protobuf

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufSerializerMessageBodyReaderWriter.java
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    private static boolean isSse(ContainerRequestContext requestCtx) {
+        final ResourceMethod method = ((ExtendedUriInfo) requestCtx.getUriInfo()).getMatchedResourceMethod();
+        return method != null && method.isSse();
+    }
+
+    private static boolean isSupportedMediaType(final Class<?> type, final MediaType mediaType) {
+        return (!Publisher.class.equals(type) &&
+                (mediaType.getType().equalsIgnoreCase(APPLICATION_PROTOBUF_TYPE.getType()) &&
```
idelpivnitskiy, line 209: IIUC, this method is required only because annotation on this class use `WILDCARD`. If we specify media types explicitly, we likely don't need this logic.

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufSerializerMessageBodyReaderWriter.java
```diff
@@ -0,0 +1,240 @@
...
+                    (mediaType.getSubtype().equalsIgnoreCase(APPLICATION_PROTOBUF_TYPE.getSubtype()) ||
+                    mediaType.getSubtype().toLowerCase().endsWith('+' + APPLICATION_PROTOBUF_TYPE.getSubtype())))) ||
+                (Publisher.class.equals(type) &&
```
idelpivnitskiy, line 212: I can see users who need to use delimited format for a single object. Publisher is not required for that.

servicetalk-data-protobuf-jersey/build.gradle
```diff
@@ -0,0 +1,75 @@
...
+  implementation project(":servicetalk-concurrent-internal")
+  implementation project(":servicetalk-data-protobuf")
+  implementation project(":servicetalk-http-api")
+  implementation project(":servicetalk-http-router-jersey")
+  implementation project(":servicetalk-http-router-jersey-internal")
+  implementation project(":servicetalk-http-utils")
+  implementation project(":servicetalk-transport-netty")
+  implementation "com.google.protobuf:protobuf-java:$protobufVersion"
+  implementation "com.google.code.findbugs:jsr305:$jsr305Version"
+  implementation "org.glassfish.jersey.core:jersey-server"
```
idelpivnitskiy, line 44: This module exposes `jersey-common` and `jakarta.ws.rs-api` to the public API. We have to define them as `api` dependencies.
`servicetalk-data-jackson-jersey` has the same issue.

idelpivnitskiy (changes requested):

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufSerializerMessageBodyReaderWriter.java
Scottmitch: the object maybe of type `javax.ws.rs.core.Response` which allows the user to customize headers and response code.

Scottmitch: `MessageLite` is the super class for all protobuf generated objects. The generics on the jersey serializer API doesn't easily allow for being more specific the types but if we encounter a class cast error a 5xx style error will be returned.

servicetalk-data-protobuf/build.gradle
Scottmitch: lets consider this as a followup PR and do it consistently throughout the project. `protobufVersion` is also used in plugins which I'm not sure can use the bom.

```gradle
protobuf {
  protoc {
    artifact = "com.google.protobuf:protoc:$protobufVersion"
  }
}
```

servicetalk-data-protobuf/src/main/java/io/servicetalk/data/protobuf/ProtobufSerializerFactory.java
Scottmitch: Can you clarify which specific mime types you are referring to? The new types include links to the specific encoding they refer to:

```java
     * {@code "application/x-protobuf-var-int"}
     * <a href="https://developers.google.com/protocol-buffers/docs/encoding">base 128 VarInt protobuf encoding</a>.
     * Should be used with {@link Publisher} to stream data.
```

servicetalk-data-protobuf/src/main/java/io/servicetalk/data/protobuf/ProtobufSerializerFactory.java
Scottmitch: The race is benign but I can use `computeIfAbsent`. I was previously throwing the checked exceptions which made this more painful. The `computeIfAbsent` will just require more casting (on generator and return value) but should be fine.

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufMediaTypes.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Media type definitions for Protobuf.
+ */
+public final class ProtobufMediaTypes {
+    /**
+     * {@code "application/x-protobuf"}.
+     * @see #APPLICATION_PROTOBUF_TYPE
+     * @see HttpHeaderValues#APPLICATION_PROTOBUF
+     */
+    public static final String APPLICATION_PROTOBUF = "application/x-protobuf";
```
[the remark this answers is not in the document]
Scottmitch: I say we stick with `x-protobuf` which seems to be the most popular option and we can add more over time if necessary. It will require additional runtime checks for content type matching that we can avoid if not necessary.

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufSerializerMessageBodyReaderWriter.java
Scottmitch: We still need to match the media type against `Publisher` for `var-int` vs "other" types for regular content type.

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufSerializerMessageBodyReaderWriter.java
Scottmitch: The encoding for var-int is different than the no-delimiter case (e.g. single object) and is not interchangeable from a serialization perspective. If you know you always have a single object you wouldn't need delimited encoding. The var-int delimited encoding is only implemented for the streaming serializers and provides the framing to send multiple objects over a single stream, if a single object is needed with var-int decoding `Publisher.from(pojo)` is the supported way.

servicetalk-data-protobuf-jersey/build.gradle
Scottmitch: I see `public class ServiceTalkProtobufSerializerFeature implements Feature` which is where `jakarta.ws.rs-api` is exposed, where is `jersey-common` exposed?

Scottmitch: Ah, I see `public final class ServiceTalkProtobufSerializerAutoDiscoverable implements AutoDiscoverable`

Scottmitch: however these types are linked via class path discovery so I'm not sure they are required to put in the public api ... would `runtime` be sufficient in this case? https://search.maven.org/artifact/io.servicetalk/servicetalk-data-jackson-jersey/0.42.8/jar

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaderValues.java
```diff
@@ -37,6 +37,15 @@
      */
     public static final CharSequence APPLICATION_X_WWW_FORM_URLENCODED_UTF_8 =
             newAsciiString(APPLICATION_X_WWW_FORM_URLENCODED + "; charset=UTF-8");
+    /**
+     * {@code "application/x-protobuf"}
+     */
+    public static final CharSequence APPLICATION_PROTOBUF = newAsciiString("application/x-protobuf");
```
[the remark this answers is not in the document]
Scottmitch: lets discuss on https://github.com/apple/servicetalk/pull/2159#discussion_r831759458

servicetalk-data-protobuf/build.gradle
bondolo: Sure. I tried previously and it is still necessary to use `$protobufVersion` with the plugin.

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufMediaTypes.java
bondolo: I believe that what's done here is the correct approach, a different mime-type. There doesn't appear to be any standard for the delimited format. I find it kind of strange that Google recommends the approach but doesn't attempt to standardize it since presumably it will be commonly used.

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufSerializerMessageBodyReaderWriter.java
bondolo: OK

servicetalk-data-protobuf-jersey/docs/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,11 @@
...
+This module provides Protobuf serialization and deserialization for ServiceTalk Jersey router. This module bypasses
+thee `InputStream` adaptation that kicks in with out-of-the-box body readers and also allows accepting/returning
+`Single<MessageLite>` and `Publisher<MessageLite>` from resource methods.
+
+CAUTION: This serializer can not currently be used with Server-Sent Events (SSE).
+
+== VarInt encoding
+`Publisher` encoding currently uses
+link:https://developers.google.com/protocol-buffers/docs/encoding[base 128 VarInt protobuf encoding] and is currently
+associated with `application/x-protobuf-var-int` content encoding.
```
idelpivnitskiy, line 11: Consider documenting what is the media-type for a single proto

servicetalk-data-protobuf-jersey/docs/antora.yml
```diff
@@ -0,0 +1,21 @@
+#
+# Copyright © 2020 Apple Inc. and the ServiceTalk project authors
```
idelpivnitskiy, line 2: 2022

servicetalk-data-protobuf-jersey/docs/antora.yml
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: servicetalk-data-protobuf-jersey
+title: JSON (Jackson)
```
idelpivnitskiy, line 18: Protobuf

servicetalk-data-protobuf-jersey/README.adoc
```diff
@@ -0,0 +1,8 @@
+= ServiceTalk Data Protobuf Jersey
+
+This module provides Protobuf serialization and deserialization for ServiceTalk Jersey router. This module
+allows skipping `InputStream` conversions and also serializes `Single<MessageLite>` and `Publisher<MessageLite>`
+from resource methods. Note that `Publisher<MessageLite>` will use
+link:https://developers.google.com/protocol-buffers/docs/encoding[Base 128 VarInt Encoding]
```
idelpivnitskiy, line 6: Missing dot at the end

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufMediaTypes.java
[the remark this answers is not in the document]
idelpivnitskiy: Consider renaming APPLICATION_PROTOBUF -> APPLICATION_X_PROTOBUF then in case we will need to add `application/protobuf` later

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaderValues.java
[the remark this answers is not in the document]
idelpivnitskiy: APPLICATION_PROTOBUF -> APPLICATION_X_PROTOBUF?

servicetalk-data-protobuf-jersey/build.gradle
idelpivnitskiy: `FeatureContext` and `MediaType` from `jakarta.ws.rs-api` are part of the public API -> it have to be an `api` dependency.
Not 100% sure about `implements AutoDiscoverable`. Something reminds me that a user had an issue when a parent abstract class was from another module defined as `implementation` instead of `api`. I would make both `api` just to be safe. 

jackson module has the same issue.

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufSerializerMessageBodyReaderWriter.java
```diff
@@ -0,0 +1,243 @@
...
+import static io.servicetalk.data.protobuf.jersey.ProtobufSerializerMessageBodyReaderWriter.PRODUCES_CONSUMES_STR;
+import static io.servicetalk.http.router.jersey.internal.BufferPublisherInputStream.handleEntityStream;
+import static io.servicetalk.http.router.jersey.internal.RequestProperties.setResponseBufferPublisher;
+import static javax.ws.rs.Priorities.ENTITY_CODER;
+
+// Less priority than the *MessageBodyReaderWriters provided by the Jersey Router itself to avoid attempting
+// Protobuf (de)serialization of core types like Buffer.
+@Priority(ENTITY_CODER + 100)
+@Consumes(PRODUCES_CONSUMES_STR)
+@Produces(PRODUCES_CONSUMES_STR)
```
idelpivnitskiy, line 74: These annotations take an array of values, consider:
```java
@Consumes({APPLICATION_PROTOBUF, APPLICATION_PROTOBUF_VAR_INT})
@Produces({APPLICATION_PROTOBUF, APPLICATION_PROTOBUF_VAR_INT})
```

servicetalk-data-protobuf-jersey/src/main/java/io/servicetalk/data/protobuf/jersey/ProtobufSerializerMessageBodyReaderWriter.java
```diff
@@ -0,0 +1,243 @@
...
+import static io.servicetalk.data.protobuf.jersey.ProtobufMediaTypes.APPLICATION_PROTOBUF_VAR_INT;
+import static io.servicetalk.data.protobuf.jersey.ProtobufMediaTypes.APPLICATION_PROTOBUF_VAR_INT_TYPE;
+import static io.servicetalk.data.protobuf.jersey.ProtobufSerializerMessageBodyReaderWriter.PRODUCES_CONSUMES_STR;
+import static io.servicetalk.http.router.jersey.internal.BufferPublisherInputStream.handleEntityStream;
+import static io.servicetalk.http.router.jersey.internal.RequestProperties.setResponseBufferPublisher;
+import static javax.ws.rs.Priorities.ENTITY_CODER;
+
+// Less priority than the *MessageBodyReaderWriters provided by the Jersey Router itself to avoid attempting
+// Protobuf (de)serialization of core types like Buffer.
+@Priority(ENTITY_CODER + 100)
```
idelpivnitskiy, line 72: How do you think, should we make this priority slightly different than jackson? Not sure how that works when both features are registered.

Scottmitch: I think starting with the same priority is fine, we can adjust later if necessary.

servicetalk-data-protobuf-jersey/build.gradle
Scottmitch: alright I'll add them in both for now ... harder to remove them later though.

servicetalk-data-protobuf-jersey/src/test/java/io/servicetalk/data/protobuf/jersey/resources/SingleProtobufResources.java
```diff
@@ -52,17 +52,17 @@
...
                                                final Single<HelloRequest> single) {
         return postSingle(fail, single).map(m -> accepted(m).build());
     }
 
     @POST
     @Path("map-publisher")
-    @Consumes(APPLICATION_PROTOBUF)
-    @Produces(APPLICATION_PROTOBUF_VAR_INT)
+    @Consumes(APPLICATION_X_PROTOBUF)
+    @Produces(APPLICATION_X_PROTOBUF_VAR_INT)
```
idelpivnitskiy, line 65: Do we need an opposite test?

```java
@Consumes(APPLICATION_X_PROTOBUF_VAR_INT)
@Produces(APPLICATION_X_PROTOBUF)
```

idelpivnitskiy (approved):
Thank you!

servicetalk-data-protobuf-jersey/src/test/java/io/servicetalk/data/protobuf/jersey/resources/SingleProtobufResources.java
Scottmitch: This test lives in `PublisherProtobufResources`

PR #2160, opened 2022-03-22
Make ServiceTalk protoc plugin extensible

suman-ganta:
### Summary of change in one line.
Adds plugin insertion point (comment) into servicetalk generated service classes.
### Motivation:
When we want to enrich ServiceTalk generated service classes with additional functionality, this insertion point enables to add such code.
### Why is this change being made?
When we want to enrich ServiceTalk generated service classes with additional functionality, this insertion point enables to add such code.
### Modifications:

- List the changes
Since JavaPoet has no way to add a comment within a class, Added an empty type as place holder and replaced it with the comment before adding the content to CodeGeneratorResponse

### Result:

What is the result of this change?
Every ServiceTalk generated service class will have following comment towards the end of the class.
```
// @@protoc_insertion_point(service_scope:<fully qualified name of service>)
```

### DIFF
--- a/servicetalk-grpc-protoc/build.gradle
+++ b/servicetalk-grpc-protoc/build.gradle
@@ -127,5 +127,5 @@ task testJavadoc(type: Javadoc) {
 test.finalizedBy(testJavadoc)
 
 test {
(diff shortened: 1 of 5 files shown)


idelpivnitskiy (approved):
Thank you @suman-ganta!

PR #2161, opened 2022-03-23
Publisher.multicast duplicate cancel IndexOutOfBounds exception

Scottmitch:
Motivation:
The Publisher.multicast operator may throw IndexOutOfBounds
exception if a Subscription is cancelled multiple times and
there are other active Subscriptions.


idelpivnitskiy (approved):
🏅

bondolo (commented):
late LGTM

PR #2162, opened 2022-03-23
Add [Single | Completable].cache operators

Scottmitch:
Motivation:
Some scenarios require that a `Single` or `Completable`
can be subscribed to multiple times. Using `Processors`
requires additional state management outside of the
subscriber/operators chain which must be done with care
to preserve backpressure, cancellation, and lazy semantics.
`Publisher.multicast` isn't directly applicable either
because the terminal value isn't cached and retained for
late subscribers as is often required.

Modifications:
- Add `Single.cache` and `Completable.cache` operators.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
@@ -1071,7 +1071,7 @@ public abstract class Single<T> {
      *     return multiResults;
      * }</pre>
      * @return a {@link Single} that subscribes a single time upstream but allows for multiple downstream
-     * {@link Subscriber}s. Signals from upstream will be multicast to each downstream {@link Subscriber}.
+     * {@link Subscriber}s. The terminal signal will be cached and delivered to each downstream {@link Subscriber}.
      * @see <a href="https://reactivex.io/documentation/operators/replay.html">ReactiveX cache operator</a>
      * @see #cache(int)
      */
@@ -1095,7 +1095,7 @@ public abstract class Single<T> {
      * @param minSubscribers The upstream subscribe operation will not happen until after this many {@link Subscriber}
      * subscribe to the return value.
      * @return a {@link Single} that subscribes a single time upstream but allows for multiple downstream
-     * {@link Subscriber}s. Signals from upstream will be multicast to each downstream {@link Subscriber}.
+     * {@link Subscriber}s. The terminal signal will be cached and delivered to each downstream {@link Subscriber}.
      * @see <a href="https://reactivex.io/documentation/operators/replay.html">ReactiveX cache operator</a>
      * @see #cache(int, boolean)
      */
@@ -1122,7 +1122,7 @@ public abstract class Single<T> {
      * downstream {@link Subscriber}s cancel. {@code false} means that cancel will not be propagated upstream even if
      * all downstream {@link Subscriber}s cancel, and the upstream Subscription will stay valid until termination.
      * @return a {@link Single} that subscribes a single time upstream but allows for multiple downstream
-     * {@link Subscriber}s. Signals from upstream will be multicast to each downstream {@link Subscriber}.
+     * {@link Subscriber}s. The terminal signal will be cached and delivered to each downstream {@link Subscriber}.
      * @see <a href="https://reactivex.io/documentation/operators/replay.html">ReactiveX cache operator</a>
      * @see #cache(int, boolean, BiFunction)
      */
@@ -1152,7 +1152,7 @@ public abstract class Single<T> {
      * returns a {@link Completable} whose termination resets the state of the returned {@link Single} and allows
      * for downstream resubscribing.
      * @return a {@link Single} that subscribes a single time upstream but allows for multiple downstream
-     * {@link Subscriber}s. Signals from upstream will be multicast to each downstream {@link Subscriber}.
+     * {@link Subscriber}s. The terminal signal will be cached and delivered to each downstream {@link Subscriber}.
      * @see <a href="https://reactivex.io/documentation/operators/replay.html">ReactiveX cache operator</a>
      */
     public final Single<T> cache(int minSubscribers, boolean cancelUpstream,
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
@@ -1058,6 +1058,109 @@ public abstract class Single<T> {
     }
 
     /**
+     * Create a {@link Single} that subscribes a single time upstream but allows for multiple downstream
+     * {@link Subscriber}s. The terminal signal will be cached and delivered to each downstream {@link Subscriber}.
+     * <p>
+     * In sequential programming this is similar to the following:
+     * <pre>{@code
+     *     T result = resultOfThisSingle();
+     *     List<T> multiResults = ...; // simulating multiple Subscribers
+     *     for (int i = 0; i < expectedSubscribers; ++i) {
+     *         multiResults.add(result);
+     *     }
+     *     return multiResults;
+     * }</pre>
+     * @return a {@link Single} that subscribes a single time upstream but allows for multiple downstream
+     * {@link Subscriber}s. Signals from upstream will be multicast to each downstream {@link Subscriber}.
+     * @see <a href="https://reactivex.io/documentation/operators/replay.html">ReactiveX cache operator</a>
+     * @see #cache(int)
+     */
+    public final Single<T> cache() {
+        return cache(1);
+    }
+
+    /**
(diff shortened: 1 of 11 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1057,6 +1057,96 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
...
+     * @param cancelUpstream {@code true} if upstream should be {@link Cancellable#cancel() cancelled} when all
+     * downstream {@link Subscriber}s cancel. {@code false} means that cancel will not be propagated upstream even if
+     * all downstream {@link Subscriber}s cancel, and the upstream Subscription will stay valid until termination.
+     * @return a {@link Single} that subscribes a single time upstream but allows for multiple downstream
+     * {@link Subscriber}s. Signals from upstream will be multicast to each downstream {@link Subscriber}.
+     * @see <a href="http://reactivex.io/documentation/operators/publish.html">ReactiveX multicast operator</a>
+     * @see #multicast(int, boolean, BiFunction)
+     */
+    public final Single<T> multicast(int minSubscribers, boolean cancelUpstream) {
+        return multicast(minSubscribers, cancelUpstream, (v, t) -> Completable.never());
```
Scottmitch, line 1114: `(v, t) -> Completable.never()` -> the default here is different than what is used for `Publisher` (which uses `Completable.completed()` which clears state to allow for immediate re-subscribes by default). The difference is that we don't have to worry about data being dropped bcz for Single the terminal event includes all the data, and re-subscribes can get access to the complete state. Open to adjusting for consistency though.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MulticastSingle.java
```diff
@@ -0,0 +1,371 @@
...
+                    System.arraycopy(currSubs, 1, newSubs, 0, newSubs.length);
+                } else {
+                    System.arraycopy(currSubs, 0, newSubs, 0, i);
+                    System.arraycopy(currSubs, i + 1, newSubs, i, newSubs.length - i);
+                }
+                if (newSubscribersUpdater.compareAndSet(this, currSubs, newSubs)) {
+                    if (cancelUpstream && newSubs.length == 0) {
+                        // Reset the state when all subscribers have cancelled to allow for re-subscribe.
+                        state = new State();
+                        delayedCancellable.cancel();
```
idelpivnitskiy, line 138: Consider adding a test when a new subscriber re-subscribes after all previous subscribers canceled

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MulticastSingle.java
```diff
@@ -0,0 +1,371 @@
...
+
+        private void safeTerminalStateReset(@Nullable T value, @Nullable Throwable t) {
+            Completable completable;
+            try {
+                completable = terminalResubscribe.apply(value, t);
+            } catch (Throwable cause) {
+                LOGGER.warn("terminalStateReset {} threw", terminalResubscribe, cause);
+                completable = Completable.never();
+            }
+            completable.afterFinally(() -> state = new State()).subscribe();
```
idelpivnitskiy, line 193: IIUC, after/before doesn't matter in this case. Consider using `whenFinally`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MulticastSingle.java
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    private static final class ConcurrentOnSubscribeSubscriber<T> implements Subscriber<T> {
+        @SuppressWarnings("rawtypes")
+        private static final AtomicReferenceFieldUpdater<ConcurrentOnSubscribeSubscriber, Object> stateUpdater =
+                AtomicReferenceFieldUpdater.newUpdater(ConcurrentOnSubscribeSubscriber.class, Object.class, "state");
+        private static final Object INIT = new Object();
+        private static final Object INVOKING_ON_SUBSCRIBE = new Object();
+        private static final Object WAITING_FOR_TERMINAL = new Object();
+        private static final Object TERMINATED = new Object();
```
idelpivnitskiy, line 204: Consider using enum for these states, it's easier to read them in the heap dump then

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MulticastSingle.java
```diff
@@ -0,0 +1,371 @@
...
+        public void onSuccess(@Nullable final T result) {
+            for (;;) {
+                final Object currState = state;
+                if (currState == WAITING_FOR_TERMINAL) {
+                    state = TERMINATED; // onSubscribe already invoked, no need for atomic as no concurrency allowed.
+                    delegate.onSuccess(result);
+                    break;
+                } else if (currState == INIT) {
+                    if (stateUpdater.compareAndSet(this, INIT, TERMINATED)) {
+                        deliverSuccessFromSource(delegate, result);
```
idelpivnitskiy, line 286: Let's consider a case when Thread-onSubscribe and Thread-onSuccess both enter `currState == INIT`:
1. Thread-onSuccess wins cas -> swaps the state to `TERMINATED` -> invokes `onSubscribe` -> delivers terminal
2. Thread-onSubscribe goes to the second for loop iteration -> sees `currState == TERMINATED` -> invokes `duplicateOnSubscribe`

We should be able to distinguish this situation from the real duplicate onSubscribe.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MulticastSingle.java
Scottmitch: This should be covered in `singleSubscriberCancelStillDeliversData`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MulticastSingle.java
Scottmitch: good catch, I will adjust. we don't need to differentiate between duplicate onSubscribe and race between terminal/onSubscribe (`logDuplicateTerminal` is best effort) but we just need to make sure we terminate form the loop.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1057,6 +1057,96 @@ public final Completable flatMapCompletable(Function<? super T, ? extends Comple
...
+     *     return multiResults;
+     * }</pre>
+     * @param minSubscribers The upstream subscribe operation will not happen until after this many {@link Subscriber}
+     * subscribe to the return value.
+     * @return a {@link Single} that subscribes a single time upstream but allows for multiple downstream
+     * {@link Subscriber}s. Signals from upstream will be multicast to each downstream {@link Subscriber}.
+     * @see <a href="http://reactivex.io/documentation/operators/publish.html">ReactiveX multicast operator</a>
+     * @see #multicast(int, boolean)
+     */
+    public final Single<T> multicast(int minSubscribers) {
```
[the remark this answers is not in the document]
Scottmitch: rxjava uses [cache](https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/core/Single.java#L2640-L2659) reactor also has a [cache](https://github.com/reactor/reactor-core/blob/main/reactor-core/src/main/java/reactor/core/publisher/Mono.java#L1763) but also translates the source to hot (which we never do). we can go with `cache`.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/MulticastSingleTest.java
```diff
@@ -0,0 +1,305 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+class MulticastSingleTest {
```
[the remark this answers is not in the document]
Scottmitch: lets duplicate the tests if/when we specialize the operator implementation for `Completable`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
Scottmitch: this is now not an issue after renaming to `cache` as the difference in behavior is expected.

PR #2169, opened 2022-03-31
Let `ConnectionFactory` see difference between proxied and direct connections

idelpivnitskiy:
Motivation:
The connection initialization sequence is different for secure HTTP proxy
tunnels.  It's necessary for the `ConnectionFactory` and
`TransportObserver` to know if the `newConnection(...)` is being invoked for
a regular connection or a proxy tunnel in order to correctly count timing
of events.

Modifications:

- Introduce new `HttpContextKeys#HTTP_TARGET_ADDRESS_BEHIND_PROXY` key;
- Enhance `ProxyConnectConnectionFactoryFilter` and
`AbsoluteAddressHttpRequesterFilter` to fill in the value for
`HTTP_TARGET_ADDRESS_BEHIND_PROXY`;
- Enhance proxy tests to verify that the key value always present;

Result:

Users can distinguish attempts to open direct connections and
connections through a proxy.

Depends on #2168, review starting from the 2nd commit.

### DIFF
--- a/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DefaultContextMap.java
+++ b/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DefaultContextMap.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2021-2022 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,9 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.servicetalk.http.api;
+package io.servicetalk.concurrent.internal;
 
-import io.servicetalk.concurrent.internal.ContextMapUtils;
 import io.servicetalk.context.api.ContextMap;
 
 import java.util.HashMap;
@@ -26,11 +25,19 @@ import javax.annotation.Nullable;
 
 import static java.util.Objects.requireNonNull;
 
-final class DefaultContextMap implements ContextMap {
+/**
+ * Default implementation of {@link ContextMap}.
+ * <p>
+ * Note: it's not concurrent!
+ */
+public final class DefaultContextMap implements ContextMap {
 
     private final HashMap<Key<?>, Object> theMap;
 
-    DefaultContextMap() {
+    /**
+     * Creates a new instance.
+     */
+    public DefaultContextMap() {
         theMap = new HashMap<>(4); // start with a smaller table
     }
 
--- a/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DefaultContextMap.java
+++ b/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DefaultContextMap.java
@@ -28,7 +28,7 @@ import static java.util.Objects.requireNonNull;
 /**
  * Default implementation of {@link ContextMap}.
  * <p>
- * Note: it's not concurrent!
+ * Note: it's not thread-safe!
  */
 public final class DefaultContextMap implements ContextMap {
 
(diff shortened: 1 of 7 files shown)


bondolo (approved):

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionFactory.java
```diff
@@ -38,6 +42,32 @@
      * @param address to connect.
      * @param observer {@link TransportObserver} for the newly created connection.
      * @return {@link Single} that emits the created connection.
+     * @deprecated Use {@link #newConnection(Object, ContextMap, TransportObserver)}.
+     */
+    @Deprecated // FIXME: 0.43 - remove deprecated method
+    default Single<C> newConnection(ResolvedAddress address, @Nullable TransportObserver observer) {
```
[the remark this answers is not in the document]
idelpivnitskiy: As of item [13.5.6. of the JLS](https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.5.6):

> Adding a default method, or changing a method from abstract to default, does not break compatibility with pre-existing binaries, but may cause an `IncompatibleClassChangeError` if a pre-existing binary attempts to invoke the method. This error occurs if the qualifying type, T, is a subtype of two interfaces, I and J, where both I and J declare a default method with the same signature and result, and neither I nor J is a subinterface of the other.

Taking into account this use-case, considering risk of `IncompatibleClassChangeError` to be really low here. We did a similar trick in 0.41.x and it worked fine.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancer.java
```diff
@@ -42,8 +44,31 @@
...
      */
-    Single<C> selectConnection(Predicate<C> selector);
+    @Deprecated
+    default Single<C> selectConnection(Predicate<C> selector) { // FIXME: 0.43 - remove deprecated method
+        throw new UnsupportedOperationException("LoadBalancer#selectConnection(Predicate) is not supported by " +
+                getClass());
+    }
+
+    /**
+     * Select the most appropriate connection for a request. Returned connection may be used concurrently for other
```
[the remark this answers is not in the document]
idelpivnitskiy: Depends on what side you look at it :) 
Even if a connection is technically "available", selector can still reject it and continue to search for a better one.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionFactory.java
```diff
@@ -38,6 +42,32 @@
...
+     * provided. {@code null} context may also mean that a connection is created outside the normal request processing
+     * (e.g. health-checking).
+     * @param observer {@link TransportObserver} for the newly created connection or {@code null} if no observer
+     * provided.
+     * @return {@link Single} that emits the created connection.
      */
-    Single<C> newConnection(ResolvedAddress address, @Nullable TransportObserver observer);
+    default Single<C> newConnection(ResolvedAddress address, @Nullable ContextMap context,
+                                    @Nullable TransportObserver observer) { // FIXME: 0.43 - remove default impl
+        return Single.defer(() -> {
```
bondolo, line 66: What will the implementation be when Connection factory implementations override this method? Are they to implement the defer and the addition of the key to the context?

idelpivnitskiy: No, the `defer` here is only necessary to "glue" the old overload with `DeprecatedToNewConnectionFactoryFilter`. If users implement this method, it's expected that they will invoke the same method in delegate chain, skipping the AsyncContext dance and directly propagating the `context` to the next filter.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DefaultContextMap.java
```diff
@@ -26,11 +25,19 @@
 
 import static java.util.Objects.requireNonNull;
 
-final class DefaultContextMap implements ContextMap {
+/**
+ * Default implementation of {@link ContextMap}.
+ * <p>
+ * Note: it's not concurrent!
```
chemicL, line 31: I suppose better wording would be "it's not thread-safe".

PR #2170, opened 2022-04-01
Add jitter for the subsequent DNS queries

idelpivnitskiy:
Motivation:

Current DNS implementation of the `ServiceDiscoverer` sends new queries
after each TTL expires. This may lead to spikes in load on DNS servers
if many clients align their requests based on the returned TTL value.

Modifications:

- Add jitter for the subsequent DNS queries;
- Allow users to customize jitter value;

Result:

Subsequent DNS queries are spread out in time using jitter.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -118,10 +118,10 @@ final class DefaultDnsClient implements DnsClient {
     private final int srvConcurrency;
     private final boolean srvFilterDuplicateEvents;
     private final boolean inactiveEventsOnError;
-    private final long scheduleJitterNanos;
+    private final long ttlJitterNanos;
     private boolean closed;
 
-    DefaultDnsClient(final IoExecutor ioExecutor, final int minTTL, final long scheduleJitterNanos,
+    DefaultDnsClient(final IoExecutor ioExecutor, final int minTTL, final long ttlJitterNanos,
                      final int srvConcurrency, final boolean inactiveEventsOnError,
                      final boolean completeOncePreferredResolved, final boolean srvFilterDuplicateEvents,
                      Duration srvHostNameRepeatInitialDelay, Duration srvHostNameRepeatJitter,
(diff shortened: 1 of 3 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -117,9 +118,11 @@
     private final int srvConcurrency;
     private final boolean srvFilterDuplicateEvents;
     private final boolean inactiveEventsOnError;
+    private final long scheduleJitterNanos;
     private boolean closed;
 
-    DefaultDnsClient(final IoExecutor ioExecutor, final int minTTL, int srvConcurrency, boolean inactiveEventsOnError,
+    DefaultDnsClient(final IoExecutor ioExecutor, final int minTTL, final long scheduleJitterNanos,
```
tkountis, line 124: Consider naming it `ttlJitterNanos`

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -629,12 +633,15 @@ private void cancelAndTerminate0(Throwable cause) {
...
+                final long delay = ThreadLocalRandom.current()
+                        .nextLong(nanos, addWithOverflowProtection(nanos, scheduleJitterNanos));
+                LOGGER.debug("DnsClient {}, scheduling DNS query for {} after {}ms, original TTL: {}ms.",
+                        DefaultDnsClient.this, AbstractDnsPublisher.this, NANOSECONDS.toMillis(delay),
+                        NANOSECONDS.toMillis(nanos));
 
                 // This value is coming from DNS TTL for which the unit is seconds and the minimum value we accept
                 // in the builder is 1 second.
-                cancellableForQuery = nettyIoExecutor.schedule(this::doQuery0, nanos, NANOSECONDS);
+                cancellableForQuery = nettyIoExecutor.schedule(this::doQuery0, delay, NANOSECONDS);
```
tkountis, line 644: It would have been nice to support `retry strategies` here too, but can't think of an easy way to do so.

tkountis (approved):

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
idelpivnitskiy: SD retries are handled at client level, bcz it needs visibility into SD errors, see `SingleAddressHttpClientBuilder#retryServiceDiscoveryErrors(...)`.
Don't think we need retry `schedule`, `nettyIoExecutor` won't reject unless it's closed.

PR #2171, opened 2022-04-01
Let users move `GrpcLifecycleObserver` in the filter chain

idelpivnitskiy:
Motivation:

gRPC users can not use `HttpLifecycleObserverServiceFilter` to install
`GrpcLifecycleObserver` because `onGrpcStatus` won't be invoked.
Sometimes, it's necessary to see the state of other filters from inside
the observer.

Modifications:

- Make `HttpLifecycleObserverServiceFilter` extensible as a class, but
keep all its methods `final`;
- Add `GrpcLifecycleObserverServiceFilter`;

Result:

Users can install `GrpcLifecycleObserver` anywhere in a filter chain.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
@@ -161,6 +161,12 @@ abstract class AbstractLifecycleObserverHttpFilter implements HttpExecutionStrat
         return HttpExecutionStrategies.offloadNone();
     }
 
+    @Override
+    public final HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strategy) {
+        // no influence since we do not block and the observer is not expected to block either
+        return strategy;
+    }
+
     private static final class ExchangeContext implements TerminalSignalConsumer {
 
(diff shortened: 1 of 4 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
```diff
@@ -161,6 +161,12 @@ public final HttpExecutionStrategy requiredOffloads() {
         return HttpExecutionStrategies.offloadNone();
     }
 
+    @Override
+    public final HttpExecutionStrategy influenceStrategy(final HttpExecutionStrategy strategy) {
```
Scottmitch, line 165: is it necessary to implement both methods? I thought @bondolo set this up so that if `requiredOffloads` is implemented we should take the strategy from there?

Scottmitch (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
idelpivnitskiy: The goal was to make all existing methods `final` even of they are deprecated. It won't be invoked at runtime

PR #2173, opened 2022-04-02
`RoundRobinLoadBalancer`: better identify excess connections

idelpivnitskiy:
Motivation:

Cold start or traffic spikes can result in opening excess connections to
handle unexpected load. Current selection algorithm in
`RoundRobinLoadBalancer` attempts to pick a random connection. This
approach does not let to decrease number of open connections when
traffic stabilizes.

Modifications:

Introduce `linearSearchSpace` configuration option for
`RoundRobinLoadBalancerFactory` to start a linear search before falling
back to a random selection.

Result:

Allows configuring more predictable selection of the connection. When
configured with `IDLE_TIMEOUT` socket option, helps to reduce the number
of connections that can serve multiplexed or pipelined requests.
Default value of 16 has no performance impact on non-pipelined HTTP/1.1
connections, but significantly reduces latencies for HTTP/2 requests and
number of open HTTP/2 connections. For 1024 concurrent HTTP/2 requests
p99 latency drops down 2.5x times without offloading and 4x with offloading,
number of open connections goes down from ~256 to <10.


tkountis (approved):
lgtm - nice trick

chemicL:
Great improvement, congratulations!

PR #2174, opened 2022-04-02
Always set default `ServiceTalkSocketOptions#IDLE_TIMEOUT`

idelpivnitskiy:
Motivation:

It's a good practice to close excess connections if they become idle.

Modifications:

- Set default `IDLE_TIMEOUT` to 5 min;
- Clarify in javadoc that 0 value disabled idle timeout;
- Remove `@Nullable` annotation for `idleTimeoutMs` in internal classes;

Result:

Users always have `ServiceTalkSocketOptions#IDLE_TIMEOUT` set by
default.

Improves default behavior in conjunctions with #2173.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentConnectionContext.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentConnectionContext.java
@@ -93,7 +93,7 @@ import static java.util.Objects.requireNonNull;
 
 final class H2ClientParentConnectionContext extends H2ParentConnectionContext {
     private H2ClientParentConnectionContext(Channel channel, HttpExecutionContext executionContext,
-                                            FlushStrategy flushStrategy, Long idleTimeoutMs,
+                                            FlushStrategy flushStrategy, long idleTimeoutMs,
                                             @Nullable final SslConfig sslConfig,
                                             final KeepAliveManager keepAliveManager) {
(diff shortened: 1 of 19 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentConnectionContext.java
```diff
@@ -93,7 +93,7 @@
 
 final class H2ClientParentConnectionContext extends H2ParentConnectionContext {
     private H2ClientParentConnectionContext(Channel channel, HttpExecutionContext executionContext,
-                                            FlushStrategy flushStrategy, @Nullable Long idleTimeoutMs,
+                                            FlushStrategy flushStrategy, Long idleTimeoutMs,
```
bondolo, line 96: At 0.43 you may want to switch the signature to `long` since users will have to recompile. (Or add a deprecate now and a new `long` overload)

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ParentConnectionContext.java
```diff
@@ -71,13 +71,12 @@
     private final KeepAliveManager keepAliveManager;
     @Nullable
     private final SslConfig sslConfig;
-    @Nullable
     final Long idleTimeoutMs;
```
bondolo, line 74: can be `long`

tkountis (approved):

PR #2185, opened 2022-04-11
Shared BlockingUtils in concurrent api internal

tkountis:
Motivation:

BlockingUtils offer useful functionality that could be used outside the http package.

Modifications:

Moved BlockingUtils to concurrent-api-internal

Result:

Shared utilities among modules.

### DIFF
--- a/servicetalk-concurrent-api-internal/src/main/java/io/servicetalk/concurrent/api/internal/BlockingUtils.java
+++ b/servicetalk-concurrent-api-internal/src/main/java/io/servicetalk/concurrent/api/internal/BlockingUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2022 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.servicetalk.http.api;
+package io.servicetalk.concurrent.api.internal;
 
 import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.Single;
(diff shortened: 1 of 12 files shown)


servicetalk-concurrent-api-internal/src/main/java/io/servicetalk/concurrent/api/internal/BlockingUtils.java
```diff
@@ -24,13 +24,19 @@
...
 
     private BlockingUtils() {
         // no instances
     }
 
-    interface RunnableCheckedException {
+    /**
+     * Wrapper {@link Runnable} that captures and converts exceptions to unchecked.
+     */
+    public interface RunnableCheckedException {
```
idelpivnitskiy, line 39: We discussed offline that to avoid a custom runnable type in our internal classes, we can add `Completable.fromCallbable(Callable<Void>)` static method. Consider submitting a PR for that first, then we won't need to make this type public.

Scottmitch: I also wonder what methods we need from this class externally. IIRC this class pre-dates the `Future` and other blocking conversion methods on our async sources.

tkountis: @Scottmitch the interesting parts is the `blockingInvocation` APIs. The public surface is small, nevertheless is nice not to repeat this code-block in various spots.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -1793,7 +1794,27 @@ public static Completable completed() {
...
+     * <p>
+     * Blocking inside {@link Callable#call()} will in turn block the subscribe call to the returned
+     * {@link Completable}.
+     * If this behavior is undesirable then the returned {@link Completable} should be offloaded using
+     * {@link #subscribeOn(io.servicetalk.concurrent.Executor)} which offloads the subscribe call.
+     *
+     * @param callable {@link Callable} which is invoked before completion.
+     * @return A new {@code Completable}.
+     */
+    public static Completable fromCallable(final Callable<Void> callable) {
```
idelpivnitskiy, line 1816: For visibility in release notes and git history, can you please open a separate PR for `Completable.fromCallable`?

idelpivnitskiy (approved):

PR #2166, opened 2022-03-29
`MultiAddressClientBuilder` execution strategy control

bondolo:
Motivation:
The `MultiAddressClientBuilder` does not currently always respect
overrides of execution strategy made by the single address
initializer
Modification:
The computed execution strategy is a combination of the multiaddress
client builder strategy, the api strategy and the client strategy.
Result:
More predictable and controllable behavior for `MultiAddressClientBuilder`

PR #2192, opened 2022-04-15
Fix regression of client `offloadNone()` strategy

bondolo:
Motivation:
The changes in #2108 introduced a regression in the handling of
computation of the effective strategy of a client when a
strategy `offloadNone()` (or the deprecated `offloadNever()`) was used
with the  `HttpClientBuilder`. Rather than preventing all offloading,
the computed offloading was used.
Modifications:
If the strategy specified on the `HttpClientBuilder` does not specify
any offloads then no offloading will be used.
Result:
The offloading behavior matches the requested behavior specified on
the  `HttpClientBuilder`.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ClientStrategyInfluencerChainBuilder.java
@@ -120,7 +120,7 @@ final class ClientStrategyInfluencerChainBuilder {
         }
     }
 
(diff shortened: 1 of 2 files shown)


Scottmitch:
@bondolo - test failure occurred, is it related to these changes?

bondolo:
The test's assumptions appear to be incorrect. I will take a look

bondolo:
The fix to this regression was independently fixed in #2166. This PR extracts the only the necessary fix along with required test changes.

idelpivnitskiy (approved):
Thank you!

PR #2193, opened 2022-04-16
Call japicmp.sh from release.sh

Scottmitch:
Motivation:
japicmp.sh is a standalone script that must be
manually run which is easy to forget.

Modifications:
- add branch name support to release.sh
- build locally
- call japicmp.sh by default

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -21,16 +21,26 @@ cd ..
 set -eu
 
 function usage() {
-  echo "Usage: $0 next_version"
+  echo "Usage: $0 old_version next_version [branch_name]"
+  echo "old_version - the previous version to run japicmp against. 'skip' to skip japicmp"
   echo "next_version - the next version to update gradle.properties, expected -SNAPSHOT suffix"
+  echo "branch_name - the branch name to release from (default is 'main')"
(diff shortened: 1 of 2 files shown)


bondolo (approved):

Scottmitch:
0.41 (45f7e07f3)

scripts/release.sh
```diff
@@ -54,6 +64,23 @@ if ( echo "$version" | grep -q "SNAPSHOT" ); then
     exit 1
 fi
 
+if [ -z "${DRYRUN:-}" ]; then
+    gradle_build_args="--no-build-cache --warning-mode all --refresh-dependencies clean build publishToMavenLocal"
```
idelpivnitskiy, line 68: `build` takes quite a long time to run all the tests. When I want to run `japicmp.sh`, I usually use `assemble` task. It's enough to compile and generate jars.
`publishToMavenLocal` is not required, japicmp.sh will take artifacts from the `build` directory

Scottmitch: `publishToMavenLocal` was done bcz some of the related release tasks (for related projects) involve validating with local builds. I can remove though.

PR #2194, opened 2022-04-16
release.sh fixes to build on non-main branch

Scottmitch:
Motivation:
release.sh has a bug that doesn't allow specifying the branch,
also generating docs from non-main branch may overwrite antora
docs bundle which includes main branch content, so disable
publishing in this case.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -25,9 +25,9 @@ JAPICMP_SKIP_VERSION="skip"
 
 function usage() {
   echo "Usage: $0 old_version next_version [branch_name]"
-  echo "old_version - the previous version to run japicmp against. ${JAPICMP_SKIP_VERSION} to skip japicmp"
-  echo "next_version - the next version to update gradle.properties, expected -SNAPSHOT suffix"
-  echo "branch_name - the branch name to release from (default is ${DEFAULT_BRANCH})"
+  echo "old_version - the previous version to run japicmp against. \"${JAPICMP_SKIP_VERSION}\" to skip japicmp"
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -20,15 +20,18 @@ cd ..
 
 set -eu
 
+DEFAULT_BRANCH="main"
+JAPICMP_SKIP_VERSION="skip"
+
 function usage() {
   echo "Usage: $0 old_version next_version [branch_name]"
-  echo "old_version - the previous version to run japicmp against. 'skip' to skip japicmp"
+  echo "old_version - the previous version to run japicmp against. ${JAPICMP_SKIP_VERSION} to skip japicmp"
```
bondolo, line 28: Some form of quote would help here.

scripts/release.sh
```diff
@@ -144,18 +147,25 @@ done
...
+fi
 
 $git commit -a -m "Preparing for $nextVersion development"
 $git push -u ${remote_name} "$branchName"
 # Push tag after branch otherwise, CodeQL GH Action will fail.
 $git push ${remote_name} "$version"
 
-./scripts/publish-docs.sh "$version_majorminor"
+# Antora docs are published as a single bundle which includes versions, and can only be published from main branch
+# (or else we main drop docs for newer versions). If docs are updated on minor release update the tag version in
```
bondolo, line 167: > (or else we main drop docs for newer versions)

 is unclear to me.

bondolo (approved):

scripts/release.sh
Scottmitch: clarified

Scottmitch:
main (eaa9cfdf0)

PR #2195, opened 2022-04-16
Resolve javadocAll errors

Scottmitch:
Motivation:
The release.sh script runs `javadocAll` task on JDK17 which
found a few errors.

Motivation:
- JDK8 javadoc generation has an erroneous
  `warning - @link can't find` from generated code.
  Workaround the issue by using canonical type name in the link.
- Remove `h3` usage in Test sources
- Remove exclude of generated code that is linked in javadocs
  from non-generated code.

PR #2196, opened 2022-04-16
Combine publish-docs.sh and release.sh

Scottmitch:
Motivation:
publish-docs.sh expects to only run on the main branch.
However we also want to generate javadocs on branches
other than main, for example to publish updated deprecation
notices.

Modifications:
- Remove publish-docs.sh, and move its logic into release.sh

Result:
Non-main branches will have javadocs published during release.


Scottmitch:
Depends upon https://github.com/apple/servicetalk/pull/2195

scripts/release.sh
```diff
@@ -67,6 +85,9 @@ if ( echo "$version" | grep -q "SNAPSHOT" ); then
     exit 1
 fi
 
+# Clean up the state at the beginning in case the previous run did not finish successfully
+clean_up_gh_pages
+
 if [ -z "${DRYRUN:-}" ]; then
     gradle_build_args="--no-build-cache --warning-mode all --refresh-dependencies clean build publishToMavenLocal"
```
idelpivnitskiy, line 92: I've left a comment how to simplify this command on the PR that is already merged, we can update it here: https://github.com/apple/servicetalk/pull/2193#discussion_r854283209

idelpivnitskiy (approved):

scripts/release.sh
Scottmitch: used `assemble` as recommended.

Scottmitch:
0.41 (7f4e63349)

PR #2201, opened 2022-04-26
Exclude internal modules from Maven BOM

bondolo:
Motivation:
The ServiceTalk Maven BOM currently includes all modules including
internal modules. The suggested practice for libraries is to include
only modules which are part of the public API.
Modifications:
Filter internal modules from inclusion in BOM
Result:
Published BOM contains only public API modules. Internal modules are
referenced explicitly by modules which depend upon them.

### DIFF
--- a/servicetalk-bom/build.gradle
+++ b/servicetalk-bom/build.gradle
@@ -17,9 +17,12 @@
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-core"
 apply plugin: "java-platform"
 
-rootProject.subprojects.findAll { !it.name.contains("bom") && !it.name.contains("examples") &&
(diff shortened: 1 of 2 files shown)


Scottmitch (approved):
lgtm assuming the `-internal` project dependencies are setup correctly and folks get them transitively from the "public module".

bondolo:
> lgtm assuming the `-internal` project dependencies are setup correctly and folks get them transitively from the "public module".

That is the intention. There is a possible second step to use an internal BOM to lock the versions of internal components, but it is not required for this change.

bondolo:
After discussion with @saturnism this proposal is being abandoned. Listing versions for internal modules improves resolution. The exclusion of internal modules from Javadoc will be completed in a new PR.

PR #2203, opened 2022-05-05
Add example for h2 keep alive

Scottmitch:
Motivation:
Add an example demonstrating how the HTTP/2 transport keep alive feature can be used for gRPC clients and servers.

### DIFF
--- a/servicetalk-examples/docs/modules/ROOT/pages/grpc/index.adoc
+++ b/servicetalk-examples/docs/modules/ROOT/pages/grpc/index.adoc
@@ -102,7 +102,7 @@ within that time or cancels the request.
 Demonstrates how to use HTTP/2 keep alive for gRPC
 link:{source-root}/servicetalk-examples/grpc/keepalive/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveServer.java[server]
 and
-link:{source-root}/servicetalk-examples/grpc/keepalive/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveeClient.java[client].
+link:{source-root}/servicetalk-examples/grpc/keepalive/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveClient.java[client].
 Keep alive uses transport control frames to ensure the peer is still able to read and write to open connections. If the
 peer is not able to respond to the control frame within the configured amount of time, the connection is closed. This
 is useful if your environment doesn't provide other forms of connection keep alive (e.g.
--- a/servicetalk-examples/docs/modules/ROOT/pages/grpc/index.adoc
+++ b/servicetalk-examples/docs/modules/ROOT/pages/grpc/index.adoc
@@ -100,9 +100,9 @@ within that time or cancels the request.
 == Keep Alive
 
 Demonstrates how to use HTTP/2 keep alive for gRPC
-link:{source-root}/servicetalk-examples/grpc/helloworld/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveServer.java[server]
+link:{source-root}/servicetalk-examples/grpc/keepalive/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveServer.java[server]
 and
-link:{source-root}/servicetalk-examples/grpc/helloworld/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveeClient.java[client].
+link:{source-root}/servicetalk-examples/grpc/keepalive/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveeClient.java[client].
 Keep alive uses transport control frames to ensure the peer is still able to read and write to open connections. If the
 peer is not able to respond to the control frame within the configured amount of time, the connection is closed. This
 is useful if your environment doesn't provide other forms of connection keep alive (e.g.
--- a/servicetalk-examples/docs/modules/ROOT/pages/grpc/index.adoc
+++ b/servicetalk-examples/docs/modules/ROOT/pages/grpc/index.adoc
@@ -96,6 +96,14 @@ https://grpc.io/docs/what-is-grpc/core-concepts/#deadlines[gRPC deadlines] (aka 
 – Sends hello requests to the server with 1 minute deadline and 3 second deadline and receives a greeting response
 within that time or cancels the request.
 
+[#KeepAlive]
+== Keep Alive
+
+Demonstrates how to use HTTP/2 keep alive for gRPC
+link:{source-root}/servicetalk-examples/grpc/helloworld/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveServer.java[server]
+and
+link:{source-root}/servicetalk-examples/grpc/helloworld/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveeClient.java[client].
+
 [#Debugging]
 == Debugging
 
--- a/servicetalk-examples/docs/modules/ROOT/pages/grpc/index.adoc
+++ b/servicetalk-examples/docs/modules/ROOT/pages/grpc/index.adoc
@@ -103,6 +103,13 @@ Demonstrates how to use HTTP/2 keep alive for gRPC
 link:{source-root}/servicetalk-examples/grpc/helloworld/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveServer.java[server]
 and
 link:{source-root}/servicetalk-examples/grpc/helloworld/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveeClient.java[client].
+Keep alive uses transport control frames to ensure the peer is still able to read and write to open connections. If the
+peer is not able to respond to the control frame within the configured amount of time, the connection is closed. This
+is useful if your environment doesn't provide other forms of connection keep alive (e.g.
+link:https://docs.oracle.com/javase/8/docs/api/java/net/StandardSocketOptions.html#SO_KEEPALIVE[SO_KEEPALIVE], and maybe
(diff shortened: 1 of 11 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromIterablePublisher.java
```diff
@@ -33,7 +33,15 @@
 
     private final Iterable<? extends T> iterable;
 
-    FromIterablePublisher(Iterable<? extends T> iterable) {
+    @SuppressWarnings("unchecked")
+    static <T> Publisher<T> fromIterable0(Iterable<? extends T> iterable) {
```
Scottmitch, line 37: The example uses a request streaming and uncovered the following scenario worth discussion:

- PublisherAsBlockingIterable.onSubscribe does a request(16)
- FromIterablePublisher.request(..) loops calling hasNext+next until demand is exhausted. Each loop iteration calls subscriber.onNext(next()).
- The same thread looping in FromIterablePublisher.request is also responsible for draining the queue in PublisherAsBlockingIterable. This means that no data will be sent downstream until FromIterablePublisher exits the loop. This maybe perceived as “dead lock” because no data is sent even if the Iterable returns an element. 

Options / Workarounds
- Improvement: FromIterablePublisher can unwrap PublisherAsBlockingIterable and directly return underlying Publisher
    - Skips blocking conversion, but may not solve the general problem (if not of type PublisherAsBlockingIterable)
    - Followup: If not able to unwrap, force subscribeOn. This breaks the dependency between Subscription/Subscriber but may not always be necessary, what level is this applied at, how do we infer/accept an executor, and are there other areas in the blocking APIs with similar issues?
- Workaround: Reduce PublisherAsBlockingIterable demand to 1
    - batch sizes of 1 means we won’t have to wait for batch size to be exhausted before delivering downstream

servicetalk-examples/grpc/keepalive/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveClient.java
```diff
@@ -0,0 +1,88 @@
...
+import static java.time.Duration.ofSeconds;
+
+/**
+ * Example that demonstrates how to enable HTTP/2 keep alive for a gRPC client.
+ */
+public final class KeepAliveClient {
+    public static void main(String... args) throws Exception {
+        try (BlockingStreamingGreeterClient client = GrpcClients.forAddress("localhost", 8080)
+                .initializeHttp(httpBuilder -> httpBuilder.protocols(
+                    // 4 second timeout is typically much shorter than necessary, but demonstrates PING frame traffic.
```
bondolo, line 41: Some hints as to more typical values and the rational for those choices would help.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromIterablePublisher.java
bondolo: If it is not possible to recognize the source as potentially blocking on `hasNext()` then perhaps demand should be reduced to 1.

There doesn't seem to be a way to insert a `subscribeOn` here without using global executor which would be unacceptable for at least some customers. 

> are there other areas in the blocking APIs with similar issues?

I am concerned that there probably are and curious that none of our existing tests uncovered them. This problem reminds me of the problems users encounter with [`PipedOutputStream`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/PipedOutputStream.html)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherAsBlockingIterable.java
```diff
@@ -49,7 +49,7 @@
  * @param <T> Type of items emitted by the {@link Publisher} from which this {@link BlockingIterable} is created.
  */
 final class PublisherAsBlockingIterable<T> implements BlockingIterable<T> {
```
bondolo, line 51: A more detailed comment on `itemsToRequestNext` describing problem with blocking `Iterator` might be helpful.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromIterablePublisher.java
Scottmitch: moved to https://github.com/apple/servicetalk/pull/2204

Scottmitch: captured this comment/discussion in https://github.com/apple/servicetalk/issues/2205 so we can make progress with the example 👍

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherAsBlockingIterable.java
Scottmitch: I'll move this comment to https://github.com/apple/servicetalk/issues/2205.

servicetalk-examples/docs/modules/ROOT/pages/grpc/index.adoc
```diff
@@ -96,6 +96,14 @@ https://grpc.io/docs/what-is-grpc/core-concepts/#deadlines[gRPC deadlines] (aka
 – Sends hello requests to the server with 1 minute deadline and 3 second deadline and receives a greeting response
 within that time or cancels the request.
 
+[#KeepAlive]
+== Keep Alive
+
+Demonstrates how to use HTTP/2 keep alive for gRPC
```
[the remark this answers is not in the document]
Scottmitch: I added some comments. let me know if this is what you had in mind.

Scottmitch:
This PR depends upon https://github.com/apple/servicetalk/pull/2204

servicetalk-examples/grpc/keepalive/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveClient.java
```diff
@@ -0,0 +1,91 @@
...
+                    // Using the default value is suitable in most scenarios, but if you want to customize the value
```
bondolo, line 42: The improved comments will really help.

`h2().keepAlivePolicy(new KeepAlivePolicyBuilder().build())`

is needed to configure the default. Perhaps a constant `defaultKeepAlive()` or `useDefaults()` for configuring the default.

servicetalk-examples/grpc/keepalive/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveClient.java
```diff
@@ -0,0 +1,91 @@
...
+                    // consider how many resources (network traffic, CPU for local timer management) vs time to detect
+                    // bad connection.
+                    // By default, keep alive is only sent when no traffic is detected, so if both peers have keep alive
```
bondolo, line 45: > "By default, "

may be misleading since no keep alive is the default h2 configuration.

servicetalk-examples/grpc/keepalive/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveServer.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public final class KeepAliveServer {
+    public static void main(String... args) throws Exception {
+        GrpcServers.forPort(8080)
+                .initializeHttp(httpBuilder -> httpBuilder.protocols(
+                    // 6 second timeout is typically much shorter than necessary, but demonstrates PING frame traffic.
+                    // Using the default value is suitable in most scenarios, but if you want to customize the value
+                    // consider how many resources (network traffic, CPU for local timer management) vs time to detect
+                    // bad connection.
+                    // By default, keep alive is only sent when no traffic is detected, so if both peers have keep alive
```
bondolo, line 39: Same comments as for client

bondolo (approved):

servicetalk-examples/grpc/keepalive/src/main/java/io/servicetalk/examples/grpc/keepalive/KeepAliveClient.java
Scottmitch: Since we expose the builder type I don't think the static factory helper methods provide much value and add duplication.

bondolo: True, but I had to spend a couple minutes figuring out that there was no default object and how to make one (the Builder docs don't mention that default values will be use if not supplied). Oddly, we provide a constant for disabling keep alive which is the default behaviour (not clear why you would ever needed it except to be explicit). Some mechanism or doc suggestion assistance of how to get a default keep alive policy would help IMO.

Scottmitch: Lets consider this for a followup PR to make any changes more explicit. I would prefer to remove the `disabled()` API and just use `null` if we need a default (one less API, no special "by reference" comparison required).

bondolo: All are fine approaches, whatever ensures users don't have to read the source to figure out how to get the default.

idelpivnitskiy (commented):
Late LGTM

PR #2206, opened 2022-05-06
Remove pom.xml from gRPC KeepAlive example

Scottmitch:
Motivation:
The gRPC KeepAlive example was copy/pasted from helloworld
which brought a pom.xml that isn't necessary and not updated
for this project.


Scottmitch:
I missed this before merge...

PR #2209, opened 2022-05-10
Tighter dependency for protoc plugin

bondolo:
Motivation:
Currently the `servicetalk-grpc-protoc` module exposes javapoet and
protobuf-java as dependencies though they are used only in the protoc
plugin and are not accurately dependencies of the module.

Modifications:
Use `compileOnly` for protoc plugin dependencies and change shadowJar
to build using `compileClasspath` configuration. The dependencies are
also needed for `testImplementation`.

Result:
Fewer exposed dependencies, especially dependency which is not
actually used by module outside of protoc plugin.

### DIFF
--- a/servicetalk-grpc-protoc/build.gradle
+++ b/servicetalk-grpc-protoc/build.gradle
@@ -30,7 +30,6 @@ dependencies {
   api project(":servicetalk-data-protobuf")
 
   compileOnly "com.google.code.findbugs:jsr305:$jsr305Version"
-  compileOnly "com.google.protobuf:protobuf-java:$protobufVersion"
   compileOnly "com.squareup:javapoet:$javaPoetVersion"
 
   testImplementation project(":servicetalk-grpc-api")
--- a/servicetalk-grpc-protoc/build.gradle
+++ b/servicetalk-grpc-protoc/build.gradle
@@ -26,7 +26,7 @@ dependencies {
 
   implementation project(":servicetalk-annotations")
 
-  // runtimeOnly is enough for this module in isolation, but we want transitive modules to include at compile time too.
+  // Needed for the generated classes
   api project(":servicetalk-data-protobuf")
 
   compileOnly "com.google.code.findbugs:jsr305:$jsr305Version"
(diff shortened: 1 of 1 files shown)


servicetalk-grpc-protoc/build.gradle
```diff
@@ -29,12 +29,15 @@ dependencies {
   // runtimeOnly is enough for this module in isolation, but we want transitive modules to include at compile time too.
   api project(":servicetalk-data-protobuf")
 
-  implementation "com.google.code.findbugs:jsr305:$jsr305Version"
-  implementation "com.google.protobuf:protobuf-java:$protobufVersion"
-  implementation "com.squareup:javapoet:$javaPoetVersion"
+  compileOnly "com.google.code.findbugs:jsr305:$jsr305Version"
+  compileOnly "com.google.protobuf:protobuf-java:$protobufVersion"
+  compileOnly "com.squareup:javapoet:$javaPoetVersion"
```
idelpivnitskiy, line 34: Agreed that the javapoet is completely internal dependency for this module and there is no need to leak that into users' runtime classpath. However, I'm looking at `api project(":servicetalk-data-protobuf")` and not sure why it is required to be an `api` dependency. This module itself doesn't leak `servicetalk-data-protobuf` into the public API. There is barely no public API in this module. However, it may be required (I'm not sure) for generated classes. If that's true, I think `protobuf-java` is also required, bcz generated classes reference types from `protobuf-java`. It's harder to test because we always add `servicetalk-grpc-protobuf` and `servicetalk-grpc-netty` in all projects. Those bring `protobuf-java`.

bondolo: > However, it may be required (I'm not sure) for generated classes. 

`servicetalk-data-protobuf` is required for generated classes and `servicetalk-data-protobuf` does pull in `protobuf-java` as an API dependency so the changes to `compileOnly` in `servicetalk-grpc-protoc` do not cause any problems.

idelpivnitskiy: Makes sense, thanks for clarifying!
Can we remove `protobuf-java` dependency from here then and rely on `servicetalk-data-protobuf` to pull it in?

bondolo: Yes! Now removed.

idelpivnitskiy (approved):

PR #2211, opened 2022-05-11
Avoid self-suppression for exceptions

idelpivnitskiy:
Motivation:

In case a processing function re-throws the original exception, we risk
to face `IllegalArgumentException("Self-suppression not permitted)` when
operator adds a re-thrown exception as suppressed.

Modifications:

- Add utility that verifies the suppressed exception is not the same as
the original one;

Result:

No `IllegalArgumentException("Self-suppression not permitted)`.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
@@ -289,8 +289,8 @@ abstract class AbstractLifecycleObserverHttpFilter implements HttpExecutionStrat
         try {
             onError.accept(t);
         } catch (Throwable unexpected) {
-            LOGGER.warn("Unexpected exception from {} while reporting a '{}' event",
-                    observer, eventName, addSuppressed(unexpected, t));
+            addSuppressed(unexpected, t);
+            LOGGER.warn("Unexpected exception from {} while reporting a '{}' event", observer, eventName, unexpected);
         }
     }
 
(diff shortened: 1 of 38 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
```diff
@@ -288,8 +289,8 @@ private static void safeReport(final Consumer<Throwable> onError, final Throwabl
         try {
             onError.accept(t);
         } catch (Throwable unexpected) {
-            unexpected.addSuppressed(t);
-            LOGGER.warn("Unexpected exception from {} while reporting a '{}' event", observer, eventName, unexpected);
+            LOGGER.warn("Unexpected exception from {} while reporting a '{}' event",
+                    observer, eventName, addSuppressed(unexpected, t));
```
bondolo, line 293: I preferred the version that has the function with side-effects as a separate statement. It is too easy to miss it here as an argument to logger. Pure functions are of less concern.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/CatchAllTransportObserver.java
```diff
@@ -307,8 +308,8 @@ private static void safeReport(final Runnable runnable, final Object observer, f
         try {
             runnable.run();
         } catch (Throwable unexpected) {
-            unexpected.addSuppressed(original);
-            LOGGER.warn("Unexpected exception from {} while reporting a {} event", observer, eventName, unexpected);
+            LOGGER.warn("Unexpected exception from {} while reporting a {} event",
+                    observer, eventName, addSuppressed(unexpected, original));
```
bondolo, line 312: same comment about hiding function with side-effects as method arg.

bondolo (approved):

bondolo (approved):

Scottmitch (commented):
late lgtm

PR #2212, opened 2022-05-12
Make `HostHeaderHttpRequesterFilter` public

idelpivnitskiy:
Motivation:

Users may have a need to set the `Host` header with a different value
than they used to create a client. For these cases, it's convenient to
reuse existing filter instead of writing their own.

Modifications:

- Move `HostHeaderHttpRequesterFilter` from `http-netty` to `http-utils`
and make it `public`;
- Disable
`DefaultSingleAddressHttpClientBuilder.addHostHeaderFallbackFilter` when
`HostHeaderHttpRequesterFilter` is added manually;
- Move `NetUtils` from `http-api` to `utils-internal` module to share
with other modules outside `http-api`, rename it to `NetworkUtils`;
- Add `NetworkUtilsTests`;
- Replace all usages of `io.netty.util.NetUtil.isValidIpV*Address` with
`io.servicetalk.utils.internal.NetworkUtils.isValidIpV*Address`;

Result:

Users can use `HostHeaderHttpRequesterFilter` to set the `Host` header
and move it anywhere in the filter chain.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -448,13 +449,22 @@ public DefaultSingleAddressHttpClientBuilder<U, R> appendConnectionFilter(
...
     @Override
     public DefaultSingleAddressHttpClientBuilder<U, R> appendConnectionFilter(
             final Predicate<StreamingHttpRequest> predicate, final StreamingHttpConnectionFilterFactory factory) {
-        return appendConnectionFilter(toConditionalConnectionFilterFactory(predicate, factory));
+        appendConnectionFilter(toConditionalConnectionFilterFactory(predicate, factory));
+        ifHostHeaderHttpRequesterFilter(factory);
+        return this;
+    }
+
+    private void ifHostHeaderHttpRequesterFilter(final Object filter) {
```
bondolo, line 464: do you mean `isHostHeaderHttpRequesterFilter`? I would normally write this as returning a `boolean` and do the assignment to a final field in constructor.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -448,13 +449,22 @@ public DefaultSingleAddressHttpClientBuilder<U, R> appendConnectionFilter(
         requireNonNull(factory);
         connectionFilterFactory = appendConnectionFilter(connectionFilterFactory, factory);
         strategyComputation.add(factory);
+        ifHostHeaderHttpRequesterFilter(factory);
         return this;
     }
 
     @Override
     public DefaultSingleAddressHttpClientBuilder<U, R> appendConnectionFilter(
             final Predicate<StreamingHttpRequest> predicate, final StreamingHttpConnectionFilterFactory factory) {
```
bondolo, line 458: Is `HostHeaderHttpRequesterFilter` always guaranteed to be the outermost filter in the chain?

bondolo (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
idelpivnitskiy: It's used in 4 different places, the goal was to write a single method that will take care of `addHostHeaderFallbackFilter` variable. Open for other name suggestions.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
idelpivnitskiy: There is no such requirement. The default variant that we add is always the last one, because we never know if users have another filter that manipulates with the `host` header or not. When users append it manually, they are free to place it at any position.

bondolo: OK, understood. It may be important to document that behavior so that users, who can now add the filter themselves, know to add it last.

idelpivnitskiy: They can add it at any place as they may have other filters later in the chain that rely on the `host` header reading

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
bondolo: I'm always a sucker for pure functions rather than side-effect functions and if a field can be made final, I'll go to extra trouble. Maybe not terribly important here.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
bondolo: So there is no downside for us adding another instance of this filter? We have behavior associated with this filter, it seems like it should be described.

idelpivnitskiy: From behavior point of view - no, you may have 2+ `HostHeaderHttpRequesterFilter` in the chain and result will be the same (first wins). Hence we don't try to prevent users from appending it multiple times.
For our internals, we tend to remove any unnecessary logic from runtime of request processing if we can at build time. So, I'm trying to avoid adding the default `HostHeaderHttpRequesterFilter` when there is no need for that.

bondolo: OK, sorry for the back and forth. When you say "first wins" do you mean outermost or innermost? If you mean outermost then there is there the possibility of the user added filter being ignored or not having the opportunity to have the effect they want? (calling back to the point that they may need to care about addition order).

idelpivnitskiy: The first in a request processing chain:
```
client.request(req) -> HostHeaderHttpRequesterFilter(first.servicetalk.io) -> ... -> HostHeaderHttpRequesterFilter(last.servicetalk.io) -> LB -> connection filters -> netty
```
The serialized result will have `host: first.servicetalk.io`.

idelpivnitskiy: This is clarified in javadoc:
> A filter which will set a HttpHeaderNames#HOST header with the fallback value if the header is not already present in HttpRequestMetaData.

bondolo: OK, the filter behaviour and order is understood. If the user adds their own `HostHeaderHttpRequesterFilter` instance how do they ensure that it will have effect and not be overridden by our adding another instance at the head of the chain if they did not place their instance at the head of the chain?

idelpivnitskiy: We don't add to the head, our default `HostHeaderHttpRequesterFilter` always added last. This PR adds extra validation: if it sees that user appended `HostHeaderHttpRequesterFilter`, client builder won't add our default at all. Doesn't look like people should worry about anything in this case.

PR #2208, opened 2022-05-10
Update netty-incubator-transport-native-io_uring 0.0.13 -> 0.0.14

idelpivnitskiy:

PR #2215, opened 2022-05-13
Add `mavenCentral()` repository if no repositories are configured

idelpivnitskiy:
Motivation:

After some environmental changes (possibly #2184), some PRs fail to
resolve dependencies if `mavenCentral()` is not defined in
`build.gradle`.

Modifications:

- Add `mavenCentral()` if no repositories are configured;

Result:

Successful respolution of dependencies, see #2208.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,6 +22,7 @@ if (!repositories) {
       }
     }
     repositories {
+      mavenCentral()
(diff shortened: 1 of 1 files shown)


idelpivnitskiy:
Since we already tested it in #2208, will merge

PR #2216, opened 2022-05-14
Add unit test for gRPC over h1, negotiated by TLS+ALPN

Scottmitch:
Motivation:
ServiceTalk can do gRPC over HTTP/1.x, and it can negotiate
protocols with TLS via ALPN. This isn't expected to be commonly
used in practice but is possible given the layering of transports
below protocol.

### DIFF
--- a/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcOverH1.java
+++ b/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcOverH1.java
@@ -43,6 +43,7 @@ final class GrpcOverH1 {
     private static final HttpProtocolConfig[] H2H1 = new HttpProtocolConfig[] {h2Default(), h1Default()};
 
     private enum ProtocolTestMode {
+        ServerH1_ClientH1(H1, H1),
         ServerH1_ClientH1H2(H1, H1H2),
         ServerH1_ClientH2H1(H1, H2H1),
         ServerH1H2_ClientH1(H1H2, H1),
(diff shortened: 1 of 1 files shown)


servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcOverH1.java
```diff
@@ -0,0 +1,84 @@
...
+import static io.servicetalk.transport.netty.internal.AddressUtils.localAddress;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+final class GrpcOverH1 {
+    private static final HttpProtocolConfig[] H1 = new HttpProtocolConfig[] {h1Default()};
+    private static final HttpProtocolConfig[] H1H2 = new HttpProtocolConfig[] {h1Default(), h2Default()};
+    private static final HttpProtocolConfig[] H2H1 = new HttpProtocolConfig[] {h2Default(), h1Default()};
+
+    private enum ProtocolTestMode {
+        ServerH1_ClientH1H2(H1, H1H2),
```
bondolo, line 46: Is it perhaps worth testing the ServerH1_ClientH1 case?

bondolo (approved):

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcOverH1.java
idelpivnitskiy: With h1-h1 there will be no ALPN involved, but should still work. Testing it won't harm for sure

idelpivnitskiy (approved):

PR #2217, opened 2022-05-16
Use `servicetalk-dependencies` BOM internally

bondolo:
Motivation:
The new `dependencies` BOM introduced in #2210 can be used to improve
the management of external dependencies by reducing use of explicit
versions within the module `build.gradle` files.
Modification:
Use `dependencies` BOM for every module which includes `api` or
`implementation` external dependencies. The BOM is imported as `api`
platform only in cases where the external dependency is also imported
`api`. `implementation` platform is used in cases where the external
dependencies are only `implementation` dependencies themselves.
Result:
Fewer explicit version specifications and better managed dependency
versioning.

### DIFF
--- a/servicetalk-dependencies/build.gradle
+++ b/servicetalk-dependencies/build.gradle
@@ -29,7 +29,6 @@ dependencies {
   api platform("org.glassfish.jersey:jersey-bom:$jerseyVersion")
 
   constraints {
-    // Use `api` only for dependencies whose types appear in ServiceTalk API.
     api "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
     api "com.google.api.grpc:proto-google-common-protos:$protoGoogleCommonProtosVersion"
     api "com.google.code.findbugs:jsr305:$jsr305Version"
@@ -39,7 +38,7 @@ dependencies {
     api "com.sun.xml.bind:jaxb-core:$javaxJaxbCoreVersion"
     api "io.opentracing:opentracing-api:$openTracingVersion"
     api "io.zipkin.reporter2:zipkin-reporter:$zipkinReporterVersion"
-    // `spi` and `core` types exposed in `log4j2-mdc-utils`
+    api "org.apache.logging.log4j:log4j-api:$log4jVersion"
     api "org.apache.logging.log4j:log4j-core:$log4jVersion"
     api "jakarta.ws.rs:jakarta.ws.rs-api:$jaxRsVersion"
     api "jakarta.xml.bind:jakarta.xml.bind-api:$javaxJaxbApiVersion"
@@ -50,6 +49,7 @@ dependencies {
     api "org.reactivestreams:reactive-streams:$reactiveStreamsVersion"
     api "org.slf4j:slf4j-api:$slf4jVersion"
 
+    // Use `runtime` for dependencies which are used ONLY at runtime
     runtime "com.sun.xml.bind:jaxb-impl:$javaxJaxbImplVersion"
     runtime "org.apache.logging.log4j:log4j-slf4j-impl:$log4jVersion"
   }
--- a/servicetalk-dependencies/build.gradle
+++ b/servicetalk-dependencies/build.gradle
@@ -26,12 +26,13 @@ javaPlatform {
 dependencies {
   api platform(project(":servicetalk-bom"))
   api platform("io.netty:netty-bom:${nettyVersion}")
-  api platform("com.fasterxml.jackson:jackson-bom:2.13.2.20220328")
   api platform("com.google.protobuf:protobuf-bom:${protobufVersion}")
   api platform("org.apache.logging.log4j:log4j-bom:${log4jVersion}")
   api platform("org.glassfish.jersey:jersey-bom:${jerseyVersion}")
 
   constraints {
+    // We don't use the jackson-bom because we want to target specifically jackson-databind
+    api "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
     api "com.google.api.grpc:proto-google-common-protos:$protoGoogleCommonProtosVersion"
     api "com.google.code.findbugs:jsr305:$jsr305Version"
     api "com.sun.activation:jakarta.activation:$javaxActivationVersion"
--- a/servicetalk-dependencies/build.gradle
+++ b/servicetalk-dependencies/build.gradle
@@ -26,7 +26,7 @@ javaPlatform {
 dependencies {
   api platform(project(":servicetalk-bom"))
   api platform("io.netty:netty-bom:${nettyVersion}")
-  api platform("com.fasterxml.jackson:jackson-bom:${jacksonVersion}")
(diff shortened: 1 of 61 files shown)


servicetalk-dependencies/build.gradle
```diff
@@ -32,26 +32,26 @@ dependencies {
...
     // `spi` and `core` types exposed in `log4j2-mdc-utils`
     api "org.apache.logging.log4j:log4j-core:$log4jVersion"
     api "jakarta.ws.rs:jakarta.ws.rs-api:$jaxRsVersion"
+    api "jakarta.xml.bind:jakarta.xml.bind-api:$javaxJaxbApiVersion"
     // Matchers are exposed by `servicetalk-test-resources`
     api "org.hamcrest:hamcrest:$hamcrestVersion"
+    api "org.jctools:jctools-core:$jcToolsVersion"
+    api "org.openjdk.jmh:jmh-core:$jmhCoreVersion"
     api "org.reactivestreams:reactive-streams:$reactiveStreamsVersion"
-    // Use `runtime` for dependencies which are used for implementation
```
bondolo, line 44: The Maven POM produced is identical regardless of the use of `api` or `runtime` but when imported via `platform(project(":servicetalk-dependencies"))` then `runtime` dependencies are not in scope for compile class path, ie. they are truly runtime only. Accordingly, only a few components are eligible for `runtime` status.

bondolo:
This PR includes only changes for artifact producing configurations, it does not include use of the `servicetalk-dependencies` BOM for `testImplementation` or `testImplementationFixtures` which are separate scopes–the BOM will need to be imported specifically for those scopes to be used, the changes here have no effect on those configurations.

servicetalk-log4j2-mdc-utils/build.gradle
```diff
@@ -17,14 +17,15 @@
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-library"
 
 dependencies {
+  api platform(project(":servicetalk-dependencies"))
   testImplementation enforcedPlatform("org.junit:junit-bom:$junit5Version")
 
-  api "org.apache.logging.log4j:log4j-core:$log4jVersion"
+  api "org.apache.logging.log4j:log4j-core"
```
idelpivnitskiy, line 23: Let me continue https://github.com/apple/servicetalk/pull/2210#discussion_r870547622 discussion here:

> `servicetalk-log4j2-mdc-utils` requires `log4j-core` for SPI types it exposes. I have tweaked the dependencies it uses and the dependencies referenced in the `-dependencies` BOM.

can you please clarify what requires `log4j-core`? I see only `log4j-api` classes in `servicetalk-log4j2-mdc-utils`

servicetalk-test-resources/build.gradle
```diff
@@ -17,15 +17,16 @@
...
 
   implementation project(":servicetalk-annotations")
   implementation project(":servicetalk-utils-internal")
 
-  api "org.hamcrest:hamcrest:$hamcrestVersion"
+  api "org.hamcrest:hamcrest"
 
-  implementation "org.apache.logging.log4j:log4j-core:$log4jVersion"
-  implementation "org.apache.logging.log4j:log4j-slf4j-impl:$log4jVersion"
+  implementation "org.apache.logging.log4j:log4j-core"
```
idelpivnitskiy, line 28: IIUC, this line can be removed bcz `log4j-slf4j-impl` will bring it transitively

servicetalk-dependencies/build.gradle
idelpivnitskiy: Thanks for clarifying!
Consider sequencing this PR after #2218

servicetalk-log4j2-mdc-utils/build.gradle
bondolo: I have tuned what is required more closely in the updated PR. Only the minimum necessary should be expressed now.

servicetalk-test-resources/build.gradle
bondolo: `log4j-slf4j-impl` is runtime only so the`core` is still needed at compile time.

servicetalk-dependencies/build.gradle
```diff
@@ -26,28 +26,29 @@ javaPlatform {
...
+    api "org.jctools:jctools-core:$jcToolsVersion"
+    api "org.openjdk.jmh:jmh-core:$jmhCoreVersion"
     api "org.reactivestreams:reactive-streams:$reactiveStreamsVersion"
+    api "org.slf4j:slf4j-api:$slf4jVersion"
+
     // Use `runtime` for dependencies which are used ONLY at runtime
-    runtime "com.google.code.findbugs:jsr305:$jsr305Version"
-    runtime "com.sun.activation:jakarta.activation:$javaxActivationVersion"
-    runtime "com.sun.xml.bind:jaxb-core:$javaxJaxbCoreVersion"
     runtime "com.sun.xml.bind:jaxb-impl:$javaxJaxbImplVersion"
```
bondolo, line 51: Only modules which are not needed for compile should be classified as `runtime`, `api` in BOM is used for both `api` and `implementation` in module `build.gradle` files.

servicetalk-dependencies/build.gradle
```diff
@@ -26,28 +26,29 @@ javaPlatform {
 dependencies {
   api platform(project(":servicetalk-bom"))
   api platform("io.netty:netty-bom:${nettyVersion}")
-  api platform("com.fasterxml.jackson:jackson-bom:${jacksonVersion}")
+  api platform("com.fasterxml.jackson:jackson-bom:2.13.2.20220328")
```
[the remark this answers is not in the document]
idelpivnitskiy: Thanks for researching this!

We use only one module (`jackson-databind`) from `com.fasterxml.jackson`. We can skip this bom if it creates more pain instead of value.

For us it will be easier to track only `jackson-databind` updates instead of the `jackson-bom`. With the bom, we will have to always keep the bom version the latest, even if `jackson-databind` remains the same version when they hot fix some other module. 

Protobuf and log4j don't have the same issue bcz they always keep version number consistent across modules.

servicetalk-test-resources/build.gradle
idelpivnitskiy: Can you clarify why? There are no java code in this module that depends on `log4j-core`. We only add `log4j2.xml` file that is evaluated at runtime.

bondolo: I believed it was needed for logging API but can try removing it.

servicetalk-dependencies/build.gradle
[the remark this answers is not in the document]
bondolo: Agreed. I will remove the Jackson-bom

servicetalk-test-resources/build.gradle
idelpivnitskiy: Agreed, it was not there before #2210 and everything worked fine

idelpivnitskiy (approved):

PR #2219, opened 2022-05-18
Use `servicetalk-dependencies` for tests

bondolo:
Motivation:
Some of the dependencies imported for tests can use the
`servicetalk-dependencies` BOM introduced in #2210.
Modifications:
Import `servicetalk-dependencies` for test configurations
Result:
Better management of test dependencies.


bondolo:
This is, hopefully, the last in the sequence of PRs for improving ServiceTalk dependency management via introduction and use of a `dependencies` BOM.

PR #2220, opened 2022-05-20
Clear async-context inside RRLB health checking

tkountis:
Motivation:
Health checking is an indefinite process (for as long as the host is present), and relying on async-context can lead to misbehavior with code storing stuff in it.

Modification:
Upon subscribing for the health-check process, we clear async-context.

Result:
No state maintained in async-context for the lifetime of health-checking.

### DIFF
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -25,6 +25,7 @@ import io.servicetalk.concurrent.PublisherSource.Processor;
 import io.servicetalk.concurrent.PublisherSource.Subscriber;
 import io.servicetalk.concurrent.PublisherSource.Subscription;
 import io.servicetalk.concurrent.api.AsyncCloseable;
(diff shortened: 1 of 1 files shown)


servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -778,6 +779,8 @@ public void schedule(final Throwable originalCause) {
                                 host.healthCheckConfig.executor)
                                 .apply(0, originalCause)
                                 .concat(connectionFactory.newConnection(host.address, null, null)
+                                        // Remove state from async context
+                                        .afterOnSubscribe(__ -> AsyncContext.clear())
```
idelpivnitskiy, line 783: `afterOnSubscribe` is too late in case users access `AsyncContext` from `beforeSubscriber` or `beforeOnSubscribe`. 

Consider either `beforeOnSubscribe` placed before `concat` or wrap the whole thing with `defer` and clear the context there.

idelpivnitskiy (changes requested):

idelpivnitskiy (approved):

Issue #2117, opened 2022-02-28
GracefulConnectionClosureHandlingTest failure


Scottmitch:
https://github.com/apple/servicetalk/runs/5347779658

```
GracefulConnectionClosureHandlingTest.5: protocol=HTTP_1 initiateClosureFromClient=false useUds=false viaProxy=true

java.util.concurrent.ExecutionException: io.servicetalk.transport.netty.internal.CloseHandler$CloseEventObservedException: CHANNEL_CLOSED_OUTBOUND(The transport backing this connection has been shutdown (write)) [id: 0xac0bf9d7, L:/127.0.0.1:39712 ! R:127.0.0.1/127.0.0.1:42991]
	at io.servicetalk.concurrent.api.SourceToFuture.reportGet(SourceToFuture.java:121)
	at io.servicetalk.concurrent.api.SourceToFuture.get(SourceToFuture.java:92)
	at io.servicetalk.http.netty.GracefulConnectionClosureHandlingTest.closePipelinedAfterTwoRequestsSentBeforeAnyResponseReceived(GracefulConnectionClosureHandlingTest.java:470)


Caused by: io.servicetalk.transport.netty.internal.CloseHandler$CloseEventObservedException: CHANNEL_CLOSED_OUTBOUND(The transport backing this connection has been shutdown (write)) [id: 0xac0bf9d7, L:/127.0.0.1:39712 ! R:127.0.0.1/127.0.0.1:42991]
Caused by: io.servicetalk.transport.netty.internal.StacklessClosedChannelException
	at io.servicetalk.transport.netty.internal.DefaultNettyConnection.channelInactive(...)(Unknown Source)
```

idelpivnitskiy:
https://github.com/apple/servicetalk/pull/2174/checks?check_run_id=5796849625

```
GracefulConnectionClosureHandlingTest.5: protocol=HTTP_1 initiateClosureFromClient=false useUds=false viaProxy=true

java.util.concurrent.ExecutionException: io.servicetalk.transport.netty.internal.CloseHandler$CloseEventObservedException: CHANNEL_CLOSED_OUTBOUND(The transport backing this connection has been shutdown (write)) [id: 0xb2f977e3, L:/127.0.0.1:60616 ! R:127.0.0.1/127.0.0.1:33911]
	at io.servicetalk.concurrent.api.SourceToFuture.reportGet(SourceToFuture.java:121)
[...]

PR #2221, opened 2022-05-21
gRPC improve error status for non-200 h2 response status

Scottmitch:
Motivation:
ServiceTalk currently assumes the response status is 200
when decoding into gRPC. This means we will fail during
checking the content-type or during deserialization if
an error is return from a middle proxy with a different
encoding type. We can provide more informative gRPC
error codes if we validate the h2 response status.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatusCode.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatusCode.java
@@ -67,7 +67,7 @@ public enum GrpcStatusCode {
      * when they are done.
      */
     OUT_OF_RANGE(11),
-    /** The method/operation is not implemented/supported/implemented. */
+    /** The method/operation is not implemented/supported. */
     UNIMPLEMENTED(12),
     /** Internal invariant violated. */
     INTERNAL(13),
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatusCode.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatusCode.java
@@ -40,17 +40,34 @@ public enum GrpcStatusCode {
     NOT_FOUND(5),
     /** Some entity that we attempted to create already exists. */
     ALREADY_EXISTS(6),
-    /** Permission denied for a particular client. Different from {@link #UNAUTHENTICATED}. */
+    /**
+     * Permission denied for a particular client. Must not be used for the following cases:
+     * <ul>
+     *     <li>rejections caused by exhausting some resource (use {@link #RESOURCE_EXHAUSTED} instead)</li>
+     *     <li>the caller cannot be identified (use {@link #UNAUTHENTICATED} instead)</li>
+     * </ul>
+     */
     PERMISSION_DENIED(7),
     /** Resource exhausted. */
     RESOURCE_EXHAUSTED(8),
-    /** The action cannot be executed on the current system state. Client should not retry.. */
+    /** The action cannot be executed on the current system state. Client should not retry. */
     FAILED_PRECONDITION(9),
-    /** Aborted, typically due to a concurrency issue (think CAS). Client may retry the whole sequence.. */
+    /** Aborted, typically due to a concurrency issue (think CAS). Client may retry the whole sequence. */
     ABORTED(10),
-    /** Used for range errors. */
+    /**
+     * Used for range errors (e.g. seeking or reading past end of file.)
+     * <p>
+     * Unlike {@link #INVALID_ARGUMENT}, this error indicates a problem that may be fixed if the system state changes.
+     * For example, a 32-bit file system will generate {@link #INVALID_ARGUMENT} if asked to read at an offset that is
+     * not in the range [0,2^32-1], but it will generate OUT_OF_RANGE if asked to read from an offset past the current
+     * file size.
+     * <p>
+     * There is a fair bit of overlap with {@link #FAILED_PRECONDITION}. This error is more specific and recommended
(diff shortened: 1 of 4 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -263,10 +277,35 @@ static GrpcStatusException toGrpcException(Throwable cause) {
...
+        final int statusCode = status.code();
+        if (statusCode == OK.code()) {
+            return;
+        }
+        final GrpcStatusCode grpcStatusCode;
+        if (statusCode == BAD_GATEWAY.code() || statusCode == SERVICE_UNAVAILABLE.code() ||
+                statusCode == GATEWAY_TIMEOUT.code() || statusCode == TOO_MANY_REQUESTS.code()) {
+            grpcStatusCode = UNAVAILABLE;
+        } else if (statusCode == UNAUTHORIZED.code()) {
+            grpcStatusCode = UNAUTHENTICATED;
```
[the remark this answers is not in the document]
Scottmitch: I think `UNAUTHENTICATED` make sense in this case.

https://datatracker.ietf.org/doc/html/rfc7235#section-3.1
> The 401 (Unauthorized) status code indicates that the request has not
   been applied because it lacks valid authentication credentials for
   the target resource.

Where `403 Forbidden` means we authenticated the client, but permission was denied.

The unit test also confirms we use a consistent strategy to `grpc-java`.

Scottmitch:
test failure due to https://github.com/apple/servicetalk/issues/2117

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatusCode.java
```diff
@@ -40,17 +40,34 @@ public enum GrpcStatusCode {
...
+     * not in the range [0,2^32-1], but it will generate OUT_OF_RANGE if asked to read from an offset past the current
+     * file size.
+     * <p>
+     * There is a fair bit of overlap with {@link #FAILED_PRECONDITION}. This error is more specific and recommended
+     * in scenarios when callers who are iterating through a space can easily look for an OUT_OF_RANGE error to detect
+     * when they are done.
+     */
     OUT_OF_RANGE(11),
-    /** Unimplemented action. */
+    /** The method/operation is not implemented/supported/implemented. */
```
idelpivnitskiy, line 70: `implemented` appears twice

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -263,10 +277,35 @@ static GrpcStatusException toGrpcException(Throwable cause) {
...
+        } else if (statusCode == FORBIDDEN.code()) {
+            grpcStatusCode = PERMISSION_DENIED;
+        } else if (statusCode == NOT_FOUND.code()) {
+            grpcStatusCode = UNIMPLEMENTED;
+        } else if (statusCode == BAD_REQUEST.code() || statusCode == REQUEST_HEADER_FIELDS_TOO_LARGE.code()) {
+            grpcStatusCode = INTERNAL;
```
idelpivnitskiy, line 296: I'm surprised grpc-java maps these cases to `INTERNAL`.
In HTTP, bad request usually means that client sent something incorrectly rather than a server-side error occurred. I would expect `INVALID_ARGUMENT` in both of these cases. 

Also surprised that `412 Precondition Failed` is not mapped to `FAILED_PRECONDITION` and many other 4xx/5xx mappings may have a better grpc-status code.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
```diff
@@ -485,6 +493,52 @@ void grpcJavaToServiceTalkBlocking(
...
+                try {
+                    stClient.scalarCall(request).toFuture().get();
+                } catch (ExecutionException e) {
+                    stCode = ((GrpcStatusException) e.getCause()).status().code().value();
+                }
+                if (httpCode < 200) {
+                    // grpc-java maps 1xx responses to error code INTERNAL, we currently map to UNKNOWN. The test server
+                    // isn't following the http protocol by returning only a 1xx response and each framework catches
+                    // this exception differently internally.
+                    assertThat("mismatch for h2 response code: " + httpCode, stCode, equalTo(UNKNOWN.value()));
```
idelpivnitskiy, line 534: 1. Consider adding assertion for grpc status code too:
```asciidoc
assertThat("mismatch for h2 response code: " + httpCode, grpcJavaCode, equalTo(INTERNAL.value()));
```
2. Can you please clarify what prevents us from mapping it to `INTERNAL` too?

idelpivnitskiy (approved):

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
Scottmitch: I took a stab at providing better mappings. I would suggest we can refine the http response code mapping in followup as well.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
Scottmitch: (1) -> done

(2) -> when there is a 1xx response w/out an expect request our client ignore the initial response headers, splicer sees a NettyBuffer and tries to cast it to HttpResponseMetaData (fails with ClassCastException) which gets mapped to unknown via `DefaultGrpcClientBuilder$CatchAllHttpClientFilter`.

PR #2222, opened 2022-05-21
Update spotbugs 4.6.0 -> 4.7.0

Scottmitch:


Scottmitch:
more work required, scope was larger than I anticipated. posting WIP PR to avoid duplicate effort.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
```diff
@@ -157,7 +158,7 @@ private static void shutdownExecutor(java.util.concurrent.Executor jdkExecutor)
             try {
                 ((AutoCloseable) jdkExecutor).close();
             } catch (Exception e) {
-                throw new RuntimeException("unexpected exception while closing executor: " + jdkExecutor, e);
+                throwException(e);
```
Scottmitch, line 161: most cases I added an exception, but some cases were either expected to be rare or otherwise seemed to make sense to re-throw or choose a use some other unchecked exception

bondolo: This will potentially change the type of the exception thrown. Wrapping the exception as a `RuntimeException` is gross, but it is at least consistent.

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/TestCompletable.java
```diff
@@ -148,7 +148,7 @@ public void onError(final Throwable t) {
 
     private Subscriber checkSubscriberAndExceptions() {
         if (!exceptions.isEmpty()) {
-            final RuntimeException exception = new RuntimeException("Unexpected exception(s) encountered",
+            final IllegalStateException exception = new IllegalStateException("Unexpected exception(s) encountered",
```
idelpivnitskiy, line 151: If it's a test utility, should we use `AssertionError` in this case for `TestCompletable`, `TestSingle`, and `TestPublisher`?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DefaultExecutor.java
idelpivnitskiy: both ways are not ideal, don't have any strong preference

Scottmitch: I will revert the change for this PR and submit in a followup for discussion and to make it more clear in release notes if we pull it in.

Scottmitch: moved to https://github.com/apple/servicetalk/pull/2228

PR #2229, opened 2022-05-27
Add `ConnectionObserver#connectionWritabilityChanged(boolean)`

idelpivnitskiy:
Motivation:

Callback helps users to see when channel's writability state changes.

Modifications:

- Add `ConnectionObserver#connectionWritabilityChanged(boolean)`;

Result:

Users can detect inefficient network configuration (too small outbound
socket buffer) or when the remote peer reads slower than expected.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
@@ -90,6 +90,16 @@ public interface ConnectionObserver {
     MultiplexedObserver multiplexedConnectionEstablished(ConnectionInfo info);
 
     /**
+     * Callback when a writable state of the connection changes.
+     *
+     * @param isWritable describes the current state of the connection: {@code true} when the I/O thread will perform
+     * the requested write operation immediately. If {@code false}, write requests will be queued until the I/O thread
+     * is ready to process the queued items and the transport will start applying backpressure.
+     */
+    default void connectionWritabilityChanged(boolean isWritable) { // FIXME: 0.43 - consider removing default impl
+    }
+
+    /**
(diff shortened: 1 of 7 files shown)


bondolo (approved):

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -89,6 +89,16 @@ public interface ConnectionObserver {
...
     MultiplexedObserver multiplexedConnectionEstablished(ConnectionInfo info);
 
+    /**
+     * Callback when a writable state of the connection changes.
+     *
+     * @param isWritable describes the current state of the connection: {@code true} when the I/O thread will perform
+     * the requested write operation immediately. If {@code false}, write requests will be queued until the I/O thread
+     * is ready to process the queued items and the transport will start applying backpressure.
+     */
+    default void connectionWritabilityChanged(boolean isWritable) { // FIXME: 0.43 - consider removing default impl
```
tkountis, line 99: This to the end-user looks like, it may/may-not be invoked by I/O thread. Worth commenting about the execution flow?

tkountis (approved):

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
idelpivnitskiy: For `TransportObserver` and `HttpLifecycleObserver` we expect most of the callbacks to be invoked from the IO thread. While `HttpLifecycleObserver` clarifies that in javadoc, this note is missed for `TransportObserver`. I will add that in a separate PR for visibility. Thanks for suggestion!

PR #2235, opened 2022-06-02
Add default implementations for deprecated methods

idelpivnitskiy:
Motivation:

Without providing a `default` implementation for methods that are
deprecated, users can not avoid using those in the codebase.

Modifications:

- Enhance gRPC code generator to add `default` impl for deprecated
client methods;
- Add all missed `default` implementations for deprecated methods of
`HttpMetaData` and its subclasses;

Result:

Users can avoid depending on deprecated methods in their codebase and
prepare for the next version.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpRequest.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpRequest.java
@@ -133,7 +133,12 @@ public interface BlockingStreamingHttpRequest extends HttpRequestMetaData {
      * @deprecated Use {@link #payloadBody(Iterable, HttpStreamingSerializer)}.
      */
     @Deprecated
-    <T> BlockingStreamingHttpRequest payloadBody(Iterable<T> payloadBody, HttpSerializer<T> serializer);
+    default <T> BlockingStreamingHttpRequest payloadBody(Iterable<T> payloadBody, HttpSerializer<T> serializer) {
+        throw new UnsupportedOperationException("BlockingStreamingHttpRequest#payloadBody(Iterable, HttpSerializer) " +
+                "is not supported by " + getClass() + ". This method is deprecated, consider migrating to " +
+                "BlockingStreamingHttpRequest#payloadBody(Iterable, HttpStreamingSerializer) or implement this " +
+                "method if it's required temporarily.");
+    }
 
     /**
      * Returns a {@link BlockingStreamingHttpRequest} with its underlying payload set to the result of serialization.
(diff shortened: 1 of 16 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingHttpRequest.java
```diff
@@ -133,7 +133,12 @@ default <T> BlockingIterable<T> payloadBody(HttpDeserializer<T> deserializer) {
      * @deprecated Use {@link #payloadBody(Iterable, HttpStreamingSerializer)}.
      */
     @Deprecated
-    <T> BlockingStreamingHttpRequest payloadBody(Iterable<T> payloadBody, HttpSerializer<T> serializer);
+    default <T> BlockingStreamingHttpRequest payloadBody(Iterable<T> payloadBody, HttpSerializer<T> serializer) {
+        throw new UnsupportedOperationException("BlockingStreamingHttpRequest#payloadBody(Iterable, HttpSerializer) " +
```
bondolo, line 137: If the method is throwing a UOE then it is too late for them to "consider" migrating. I would expect that deprecated methods typically don't throw UOE.  My concern is would be that implementations of this method might be removed prematurely to silence the deprecation warning and instead produce a UOE in the application. The method is still part of the public API (though deprecated) and it is reasonable for applications to expect it to still work. There is a decoupling between implementations deciding to remove the method and applications moving to the recommended method. 

This change makes it "easier" for implementations by providing a default. Are we telling them that it is OK to not provide an implementation or remove existing implementations? Doing so is potentially worse for applications though because they have to convert **now** before the deprecation removal to avoid potentially encountering the UOE.

idelpivnitskiy: > My concern is would be that implementations of this method might be removed prematurely to silence the deprecation warning and instead produce a UOE in the application.

The current problem is the opposite. Users don't have a way to stop depending on deprecated code because they are forced to implement these methods. By providing the default implementation, we open a way for them to migrate. It's their responsibility to make sure that the methods they remove are not used anymore by their business logic.

> The method is still part of the public API (though deprecated) and it is reasonable for applications to expect it to still work.

All these methods still work for our implementations, this PR does not remove methods from our implementation classes, like `AbstractHttpMetaData`. Also, ST does not invoke any of these methods. 
This PR covers cases when users implement a request/response wrapper.

> There is a decoupling between implementations deciding to remove the method and applications moving to the recommended method.

In case those wrappers leak outside of their codebase as a library on top of ST, it's the responsibility of maintainers to make sure they can remove these deprecations. In case they missed something, they can always add it back. Our UOE includes a full class name that should be helpful to understand who is the owner of that particular implementation.

> Are we telling them that it is OK to not provide an implementation or remove existing implementations?

We are giving them a way to avoid depending on deprecated code and prepare their codebase for the next version. In order to make their codebase compatible with 0.42 and the next version at the same time, all deprecations have to be removed.

> Doing so is potentially worse for applications though because they have to convert now before the deprecation removal to avoid potentially encountering the UOE.

There is no one size fits all. There are users who ready to remove deprecations and some who don't. If someone is not ready, they have a way to avoid UOE. We should not block others.

bondolo: After offline discussion we are going to go ahead with this change with the understanding that if a new implementation did not provide implementations of these method then older apps using the legacy encoding may fail. This is unlikely since most (all existing) implementations just extend the built-in implementations.

bondolo (approved):

PR #2237, opened 2022-06-03
`InOrderRouter` does not require offloading

bondolo:
Motivation:
The `InOrderRouter` did not formerly implement
`HttpExecutionStrategyInfluencer` and as a result it received full
offloading. It requires no offloading and will manually offload for
routes which specify requiring offloading.
Modifications:
Specify `offloadsNone()` as `requiredStrategy()` and adjust test
expectations that no offloading of predicate will be expected by
default. Update documentation to indicate that predicate
implementations must not block.
Result:
Potentially more efficient `InOrderRouter` which does less offloading
by default.

### DIFF
--- a/servicetalk-http-router-predicate/src/test/java/io/servicetalk/http/router/predicate/PredicateRouterOffloadingTest.java
+++ b/servicetalk-http-router-predicate/src/test/java/io/servicetalk/http/router/predicate/PredicateRouterOffloadingTest.java
@@ -61,6 +61,7 @@ import static io.servicetalk.concurrent.api.Executors.newCachedThreadExecutor;
 import static io.servicetalk.concurrent.api.Single.succeeded;
 import static io.servicetalk.http.api.HttpExecutionStrategies.customStrategyBuilder;
 import static io.servicetalk.http.api.HttpExecutionStrategies.defaultStrategy;
+import static io.servicetalk.http.api.HttpExecutionStrategies.offloadAll;
 import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
 import static io.servicetalk.http.api.HttpResponseStatus.OK;
 import static io.servicetalk.http.router.predicate.PredicateRouterOffloadingTest.RouteServiceType.ASYNC_AGGREGATED;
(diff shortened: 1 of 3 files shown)


servicetalk-http-router-predicate/src/test/java/io/servicetalk/http/router/predicate/PredicateRouterOffloadingTest.java
```diff
@@ -132,7 +133,7 @@ void predicateOffloadedAndNotRoute(RouteServiceType routeServiceType) throws Exc
         this.routeServiceType = routeServiceType;
         assumeSafeToDisableOffloading(routeServiceType);
         final HttpPredicateRouterBuilder routerBuilder = newRouterBuilder();
-        serverBuilder.executor(executionContextRule.executor());
+        serverBuilder.executionStrategy(offloadAll()).executor(executionContextRule.executor());
```
idelpivnitskiy, line 136: Hm, after noticing this change I'm not sure if we are ready to override `requireOffloads()` now because it changes the default behavior.
By returning `offloadNone()`, the default behavior will result in no offloading for all routes if the strategy is not specified at the builder. This is not what we need for the default behavior.
If a custom strategy is configured at the builder (like `offloadNone()`), it will take priority.

bondolo: > By returning offloadNone(), the default behavior will result in no offloading for all routes if the strategy is not specified at the builder. This is not what we need for the default behavior.

I am fine with leaving the behaviour as it currently is. A note on the class to specify `offloadNone()` to change the default behaviour may be sufficient.

bondolo: The other option would be for the service to require no offloads but the default "unspecified" behavior for routes would be to offload. This would achieve the same effect but not require use of the builder `offloadNone()`

bondolo: After discussion we don't want to change the behavior. The suggested behavior of having the router require no offloads and offloading happen by default for routes that do not specify offloading would be better but it is not appropriate to change the existing behavior.

PR #2238, opened 2022-06-03
Clarify expectations for `TransportObserver` implementations

idelpivnitskiy:
Motivation:

`TransportObserver` does not clarify if users can block or not inside
their implementations.

Modifications:

- Copy offloading requirements from `HttpLifecycleObserver` to
`TransportObserver`;

Result:

`TransportObserver` javadoc clarifies expectations for non-blocking
implementations and suggests to do manual offloading if necessary.


tkountis (approved):
♥️

Issue #1847, opened 2021-09-24
Flaky test: HttpLifecycleObserverTest


idelpivnitskiy:
https://github.com/apple/servicetalk/runs/3703697490 (0.41 branch)

```
HttpLifecycleObserverTest.testClientCancelsRequestBeforeResponse(HttpProtocol) [2] protocol=HTTP_2

org.mockito.exceptions.verification.VerificationInOrderFailure: 
Verification in order failure
Wanted but not invoked:
serverRequestObserver.onRequestError(
    <any java.io.IOException>
);
-> at io.servicetalk.http.netty.HttpLifecycleObserverTest.testClientCancelsRequestBeforeResponse(HttpLifecycleObserverTest.java:281)
Wanted anywhere AFTER following interaction:
serverRequestObserver.onRequestData(
    NettyBuffer{buffer=UnreleasableNoZeroingHeapByteBuf(ridx: 0, widx: 7, cap: 7)}
);
-> at io.servicetalk.http.utils.BiHttpLifecycleObserver$BiHttpRequestObserver.onRequestData(BiHttpLifecycleObserver.java:133)

	at io.servicetalk.http.netty.HttpLifecycleObserverTest.testClientCancelsRequestBeforeResponse(HttpLifecycleObserverTest.java:281)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
```

idelpivnitskiy:
Seeing this only for 0.41 branch:
https://github.com/apple/servicetalk/runs/3795514597

```
HttpLifecycleObserverTest.testClientCancelsRequestBeforeResponse(HttpProtocol) [2] protocol=HTTP_2
java.util.concurrent.TimeoutException: testClientCancelsRequestBeforeResponse(io.servicetalk.http.netty.HttpProtocol) timed out after 30 seconds

java.util.concurrent.TimeoutException: testClientCancelsRequestBeforeResponse(io.servicetalk.http.netty.HttpProtocol) timed out after 30 seconds
	at org.junit.jupiter.engine.extension.TimeoutInvocation.createTimeoutException(TimeoutInvocation.java:70)
[...]

PR #2241, opened 2022-06-07
Make routers explicit about offloading behavior

bondolo:
Motivation:
The `InOrderRouter` (aka predicate router), GRPC and Jersey routers
inherit default `requiredOffloads()` of `offloadAll()` from
`HttpStreamingService`. This behavior should be explicit.
Modifications:
Add implementations of `requiredOffloads()` returning `offloadAll()` with
documentation explaining the behavior and suggestions for handling
non-offloading requirements.
Result:
Explicit rather than implicit behavior.


idelpivnitskiy (approved):
Thank you!

bondolo:
Test failure attributed to #1847

PR #2244, opened 2022-06-08
Return `ExecutionContext` with correct strategy

bondolo:
Motivation:
`FilterableClientToClient` uses a provided `HttpExecutionStrategy` with
the underlying client but the `executionContext()` method returns an
`ExecutionContext` of the underlying client rather than the context and
strategy it is using.
Modifications:
Accept an `ExecutionContext` rather than an `ExecutionStrategy` and
return the context for `executionContext()`.
Result:
Accurate `executionContext()` result.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/FilterableClientToClient.java
```diff
@@ -159,7 +159,7 @@ public StreamingHttpRequest newRequest(final HttpRequestMethod method, final Str
 
     @Override
     public HttpExecutionContext executionContext() {
-        return client.executionContext();
+        return executionContext;
```
bondolo, line 162: This is the critical difference. The returned execution context contains the correct execution strategy.

idelpivnitskiy (approved):
Thank you, good catch!

PR #2246, opened 2022-06-08
Don't disable HTTP 1.1 in debugging example

bondolo:
Motivation:
The debugging examples show logging configuration of HTTP1 but do not
enable the protocol in the `protocols()` invocation.
Modifications:
Ensure that HTTP 1.1 is enabled.
Result:
Example is less likely to cause problems if used as cut-and-paste code

### DIFF
--- a/servicetalk-examples/grpc/debugging/src/main/java/io/servicetalk/examples/grpc/debugging/DebuggingClient.java
+++ b/servicetalk-examples/grpc/debugging/src/main/java/io/servicetalk/examples/grpc/debugging/DebuggingClient.java
@@ -30,6 +30,7 @@ import io.grpc.examples.debugging.HelloRequest;
 
 import java.util.function.BooleanSupplier;
 
+import static io.servicetalk.http.netty.HttpProtocolConfigs.h1Default;
 import static io.servicetalk.logging.api.LogLevel.TRACE;
 
 /**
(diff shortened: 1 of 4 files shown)


servicetalk-examples/grpc/debugging/src/main/java/io/servicetalk/examples/grpc/debugging/DebuggingClient.java
```diff
@@ -149,7 +150,7 @@ public static void main(String[] args) throws Exception {
                          * Be sure to also enable the logger in your logging config file,
                          * {@code log4j2.xml} for this example.
                          */
-                        .protocols(HttpProtocolConfigs.h2()
+                        .protocols(h1Default(), HttpProtocolConfigs.h2()
```
idelpivnitskiy, line 153: This will result in a runtime exception. 2 protocols are supported only when there is SSL configuration provided, bcz it requires ALPN.

idelpivnitskiy (changes requested):

servicetalk-examples/grpc/debugging/src/main/java/io/servicetalk/examples/grpc/debugging/DebuggingClient.java
bondolo: OK, any suggestion what we should do with this example? It is misleading in the current form in that it demonstrates something that is disabled.

idelpivnitskiy: This is gRPC example, it's expected to work only over HTTP/2. Can you clarify what exactly is misleading?

bondolo:
For some reason I misremembered that the general wireLogging configuration applied only to h1

PR #2247, opened 2022-06-09
Allow no SP after status-code for HTTP/1.x responses

idelpivnitskiy:
Motivation:

While RFC7230, section 3.1.2
(https://datatracker.ietf.org/doc/html/rfc7230#section-3.1.2) requires a
`SP` before a `reason-phrase`, even if it's an empty phrase, many
implementations don't include that `SP` character. Many clients parse a
response without trailing `SP` character with no error. Those
implementations are:
 - Apache HTTP Client
 - JDK HttpClient
 - Netty
 - Tomcat (https://bz.apache.org/bugzilla/show_bug.cgi?id=60362)
 - Go HttpClient (https://github.com/golang/go/commit/d71d08af5ab15c7b166d92a31219c4c218438841#diff-0e6945dfde3e4dcaba7cfd394a85328b8137c42e152b4486dbfb6756ad69a779R145-R146)
 - Rust HttpClient (https://github.com/hyperium/http/issues/345)
 - httpd with Passenger (https://bugzilla.redhat.com/show_bug.cgi?id=1032733)

Moreover, neither HTTP/2 nor HTTP/3 support reason-phrase. For this
reasons, many HTTP/1.x implementations (Tomcat, Rust, etc.) decided to
drop including the reason-phrase in serialized response for performance
reasons. Instead, they use the reason-phrase defined by RFC for all
known status codes and use `Unknown` for non-standard status codes.

Modifications:

- `HttpObjectDecoder`: second `SP` is optional for `HttpResponseDecoder`;
- Enhance tests to verify new behavior;

Result:

ServiceTalk HttpClient parses responses with no `SP` after status code.
Example: `HTTP/1.1 200\r\n`.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpResponseDecoderTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpResponseDecoderTest.java
@@ -42,6 +42,8 @@ import io.servicetalk.http.netty.HttpResponseDecoder.Signal;
 
 import io.netty.channel.embedded.EmbeddedChannel;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -105,6 +107,11 @@ class HttpResponseDecoderTest extends HttpObjectDecoderTest {
     }
 
     @Override
+    boolean isDecodingRequest() {
+        return false;
+    }
+
+    @Override
     String startLine() {
         return "HTTP/1.1 204 No Content";
     }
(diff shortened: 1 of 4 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpResponseDecoderTest.java
```diff
@@ -131,19 +138,14 @@ void illegalPrefaceCharacter() {
 
     @Test
     void noVersion() {
-        assertDecoderException("200 OK" + "\r\n", "Invalid start-line");
+        assertDecoderException("200 OK" + "\r\n", "Invalid HTTP version");
```
tkountis, line 141: 🥇

tkountis (approved):
lg

PR #2262, opened 2022-07-07
Enhance NPE in operators that invoke functions

Scottmitch:
Motivation:
Some operators require a Functional interface that provide a
non-null result (e.g. flatMap). When this scenario occurs
as NPE is propagated however the control flow is async so
it is challenging to determine where the code is that caused
the NPE.

Modifications:
- Enhance the NPE to include a reference to the Functional
  interface which provides more context as to where the NPE
  originated from.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutCompletable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutCompletable.java
@@ -102,7 +102,8 @@ final class TimeoutCompletable extends AbstractNoHandleSubscribeCompletable {
                 // it enabled for the Subscriber, however the user explicitly specifies the Executor with this operator
                 // so they can wrap the Executor in this case.
                 localTimerCancellable = requireNonNull(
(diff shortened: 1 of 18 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutCompletable.java
```diff
@@ -102,7 +102,8 @@ static TimeoutSubscriber newInstance(TimeoutCompletable parent, Subscriber targe
                 // it enabled for the Subscriber, however the user explicitly specifies the Executor with this operator
                 // so they can wrap the Executor in this case.
                 localTimerCancellable = requireNonNull(
-                        parent.timeoutExecutor.schedule(s::timerFires, parent.durationNs, NANOSECONDS));
+                        parent.timeoutExecutor.schedule(s::timerFires, parent.durationNs, NANOSECONDS),
+                        () -> "Executor " + parent.timeoutExecutor + " returned null");
```
idelpivnitskiy, line 106: In all similar cases, consider including an executor method name

idelpivnitskiy (approved):

Scottmitch:
test failure due to https://github.com/apple/servicetalk/issues/2265

PR #2263, opened 2022-07-07
`TimeoutHttpRequesterFilter` as a connection filter causes `ClosedChannelException`

idelpivnitskiy:
Motivation:

If applied as a connection-level filter, it skips the logic defined in
`LoadBalancedStreamingHttpClient` for the cancellation. Cancel will be
scheduled on the event-loop, then `onError` is propagated.
`LoadBalancedStreamingHttpClient` will mark the request as finished, it
won’t see cancel. As the result, the same connection can be selected for
another request, and users see `ClosedChannelException`.

Modifications:

- For `StreamingHttpConnectionFilterFactory` variant used for HTTP/1.X
connections, apply `BeforeFinallyHttpOperator` with `cancel` callback
similar to `LoadBalancedStreamingHttpClient`;
- Enhance mocks to account for new behavior;
- Move the common logic of extracting a fallback executor as a static
method in `AbstractTimeoutHttpFilter`;

Result:

HTTP/1.X requests don't fail with `ClosedChannelException` if a previous
request on the same connection times out.

### DIFF
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019, 2021 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019, 2021-2022 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package io.servicetalk.http.utils;
 import io.servicetalk.concurrent.TimeSource;
 import io.servicetalk.concurrent.api.Executor;
 import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.concurrent.api.TerminalSignalConsumer;
 import io.servicetalk.http.api.FilterableStreamingHttpClient;
 import io.servicetalk.http.api.FilterableStreamingHttpConnection;
 import io.servicetalk.http.api.HttpExecutionContext;
@@ -161,10 +162,7 @@ public final class TimeoutHttpRequesterFilter extends AbstractTimeoutHttpFilter
             @Override
             protected Single<StreamingHttpResponse> request(final StreamingHttpRequester delegate,
                                                             final StreamingHttpRequest request) {
-                HttpExecutionContext executionContext = client.executionContext();
-                return TimeoutHttpRequesterFilter.this.withTimeout(request, delegate::request,
-                        executionContext.executionStrategy().hasOffloads() ?
-                                executionContext.executor() : executionContext.ioExecutor());
+                return TimeoutHttpRequesterFilter.this.withTimeout(request, delegate::request, executionContext());
             }
         };
     }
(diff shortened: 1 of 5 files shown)


bondolo:
> If applied as a connection-level filter, it skips the logic defined in
LoadBalancedStreamingHttpClient for the cancellation. 

Perhaps instead:
>  If applied as a connection-level filter, then timeouts will skip the logic defined in
LoadBalancedStreamingHttpClient for the request cancellation.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -174,10 +172,37 @@ public StreamingHttpConnectionFilter create(final FilterableStreamingHttpConnect
         return new StreamingHttpConnectionFilter(connection) {
             @Override
             public Single<StreamingHttpResponse> request(final StreamingHttpRequest request) {
-                HttpExecutionContext executionContext = connection.executionContext();
-                return TimeoutHttpRequesterFilter.this.withTimeout(request, r -> delegate().request(r),
-                        executionContext.executionStrategy().hasOffloads() ?
-                                executionContext.executor() : executionContext.ioExecutor());
+                final FilterableStreamingHttpConnection delegate = delegate();
+                return TimeoutHttpRequesterFilter.this.withTimeout(request,
+                        delegate.connectionContext().protocol().major() > 1 ? delegate::request :
```
bondolo, line 177: `>= 2` feels more explicit about defining the condition.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -174,10 +172,37 @@ public StreamingHttpConnectionFilter create(final FilterableStreamingHttpConnect
...
+                                            }
+
+                                            @Override
+                                            public void onError(final Throwable throwable) {
+                                                // noop
+                                            }
+
+                                            @Override
+                                            public void cancel() {
+                                                // In alignment with cancellation processing in
```
bondolo, line 192: Consider moving the comments up to the conditional operator.

bondolo (approved):

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
idelpivnitskiy: Having the comment here helps to quickly find a similar callback inside `LoadBalancedStreamingHttpClient` that has more clarification around the same topic. Let me keep it here.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -174,10 +172,37 @@ public StreamingHttpConnectionFilter create(final FilterableStreamingHttpConnect
...
+                                                // LoadBalancedStreamingHttpClient, we need to close the HTTP/1.x
+                                                // connection when this filter is applied at connection level.
+                                                // Otherwise, a connection will be marked as "free" and can be selected
+                                                // for another request, racing with closure.
+                                                delegate.closeAsync().subscribe();
```
Scottmitch, line 197: Aims of the biggest challenges is identifying who initiates connection closure and why. I know our api doesn’t support exception/rational propagation but should we add a debug log statement and an internal utility method to share code with the other spot that does this?

servicetalk-http-utils/src/test/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilterTest.java
```diff
@@ -27,14 +27,19 @@
...
 import java.util.function.BiFunction;
 
 import static io.servicetalk.buffer.netty.BufferAllocators.DEFAULT_ALLOCATOR;
+import static io.servicetalk.concurrent.api.Completable.completed;
+import static io.servicetalk.http.api.HttpProtocolVersion.HTTP_1_1;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+@Timeout(3)
```
Scottmitch, line 42: Do we need this? If so will 3 seconds be aggressive on different hardware (overloaded CI)? (I’m on my phone so limited access to code).

Scottmitch (approved):

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
idelpivnitskiy: Agreed, it's a bit unfortunate that `cancel` doesn't carry any context with it. 
There is no good place for a utility (and the line is really simple) bcz this filter and `LoadBalancedStreamingHttpClient` are in different modules. I will add debug logging.

idelpivnitskiy:
I will merge this for now and can partially revert later if #2266 will be more preferable approach.

PR #2264, opened 2022-07-08
Cancellation can leave an HTTP/1.x connection in half-closed state

idelpivnitskiy:
Motivation:

If a response is cancelled at the connection level (skipping explicit
closure of the connection in `LoadBalancedStreamingHttpClient`), it will
hang in half-closed state (inbound closed) if the server doesn't proceed
with sending more bytes on that connection (outbound will never be
closed). Also, cancellation marks the connection as
`CHANNEL_CLOSED_INBOUND`, which is misleading because the closure is
initiated by the client.

Modifications:

- Enhance `ResponseCancelTest` to verify the described scenario;
- Adjust `NettyChannelPublisher` to close both outbound (first, to
correctly mark the connection close reason) and inbound;

Result:

Cancelled HTTP/1.x connection is always closed correctly.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseCancelTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseCancelTest.java
@@ -54,13 +54,11 @@ import java.nio.channels.ClosedChannelException;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.atomic.AtomicInteger;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.api.AsyncCloseables.newCompositeCloseable;
 import static io.servicetalk.concurrent.api.Completable.completed;
 import static io.servicetalk.concurrent.api.Processors.newSingleProcessor;
-import static io.servicetalk.concurrent.api.Single.defer;
 import static io.servicetalk.concurrent.api.SourceAdapters.fromSource;
 import static io.servicetalk.http.netty.HttpClients.forSingleAddress;
 import static io.servicetalk.http.netty.HttpServers.forAddress;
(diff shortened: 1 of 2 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseCancelTest.java
```diff
@@ -126,54 +164,108 @@ void cancel() throws Throwable {
         // and hence fail the response.
         ClientTerminationSignal.resumeExpectFailure(delayedClientTermination, latch1,
                 instanceOf(ClosedChannelException.class));
+        clientConnectionClosed.await();
+        // serverConnectionClosed.await();  server doesn't close the connection bcz it hasn't finished the response yet
```
Scottmitch, line 168: consider modifying the test to verify the server cleanups too:

```diff
diff --git a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseCancelTest.java b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseCancelTest.java
index 97f0da293..46ddb14d6 100644
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseCancelTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseCancelTest.java
@@ -154,7 +154,7 @@ class ResponseCancelTest {
         CountDownLatch latch1 = new CountDownLatch(1);
         Cancellable cancellable = sendRequest(client, latch1);
         // wait for server to receive request.
-        serverResponses.take();
+        Processor<StreamingHttpResponse, StreamingHttpResponse> serverResp = serverResponses.take();
         assertThat("Unexpected connections count.", connectionCount.get(), is(1));
         cancellable.cancel();
         // wait for cancel to be observed but don't send cancel to the transport so that transport does not close the
@@ -165,9 +165,11 @@ class ResponseCancelTest {
         ClientTerminationSignal.resumeExpectFailure(delayedClientTermination, latch1,
                 instanceOf(ClosedChannelException.class));
         clientConnectionClosed.await();
-        // serverConnectionClosed.await();  server doesn't close the connection bcz it hasn't finished the response yet
 
         sendSecondRequest();
+
+        serverResp.onSuccess(client.asStreamingClient().httpResponseFactory().ok());
+        serverConnectionClosed.await();
     }
 
     @Test

```

Scottmitch: similar comment on other new tests

Scottmitch (approved):

PR #2266, opened 2022-07-11
Handle HTTP/1.1 response cancelation same way at all levels

idelpivnitskiy:
Motivation:

When response cancel is generated at the client level,
`LoadBalancedStreamingHttpClient` has a logic in place to close the
connection to mitigate a race between connection availability and
closure. However, this logic is skipped if cancellation is originated from
a connection-level filter.

Modifications:

- Move cancellation handling from `LoadBalancedStreamingHttpClient` to
`AbstractStreamingHttpConnection`;
- Propagate `OnStreamClosedRunnable` using `request.context()`;
- Remove `StreamingHttpRequestWithContext`;
- Remove "cancel handling" logic from `TimeoutHttpRequesterFilter`
introduced in #2263;
- Add more comments and javadoc;
- Adjust tests and comments;

Result:

Request to cancel is intercepted after all user-defined filters to make sure
we catch all sources of cancellation.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/PipelinedStreamingHttpConnection.java
```diff
@@ -40,16 +41,40 @@
...
                                              @Nullable final FlushStrategy flushStrategy) {
-        if (flushStrategy == null) {
-            return connection.write(requestStream);
-        } else {
-            // TODO(scott): if we can remove the flush state on the connection we can simplify the control flow here.
-            return Publisher.defer(() -> {
+        assert connectionContext().protocol().major() <= 1 : "Unexpected protocol version";
+        return (flushStrategy == null ? connection.write(requestStream) :
+            Publisher.defer(() -> {
                 final Cancellable resetFlushStrategy = connection.updateFlushStrategy(
```
Scottmitch, line 47: `// TODO(scott): if we can remove the flush state on the connection we can simplify the control flow here.` -> not necessary to retain but we will be reworking this in the next few months so I assume this is understood :)

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/PipelinedStreamingHttpConnection.java
```diff
@@ -40,16 +41,40 @@
...
+
+                @Override
+                public void cancel() {
+                    // If the HTTP/1.x request gets cancelled, we pessimistically assume that the transport will close
+                    // the connection since the Subscriber did not read the entire response and cancelled. This reduces
+                    // the time window during which a connection is eligible for selection by the load balancer post
+                    // cancel and the connection being closed by the transport.
+                    // Transport MAY not close the connection if cancel raced with completion and completion was seen by
+                    // the transport before cancel. We have no way of knowing at this layer if this indeed happen.
+                    closeAsync().subscribe();
```
Scottmitch, line 70: identifying the rational for closure is one of the most common support questions. can we add an internal utility method shared by other cases that force closure to provide additional visibility?

```
static void doClose(AsyncClosable closable, String reason) {
  closable.closeAsync();
  LOGGER.debug("closing {} because {}", closable, reason):
}

File: servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NonPipelinedStreamingHttpConnection.java [hunk 3]
```diff
@@ -39,6 +39,7 @@
     @Override
     protected Publisher<Object> writeAndRead(final Publisher<Object> requestStream,
                                              @Nullable final FlushStrategy flushStrategy) {
+        assert connectionContext().protocol().major() >= 2 : "Unexpected protocol version";
```
Scottmitch, line 42: the cancel+close is the only reason why which may not be the case in the future (e.g. for non-pipelined h1 connections we could use this utility, IIRC we don't bcz folks may do `reserveConnection` which doesn't queue requests currently). consider adding comments as to why this restriction exists.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseCancelTest.java
```diff
@@ -54,9 +53,9 @@
 import static io.servicetalk.transport.netty.internal.AddressUtils.localAddress;
 import static io.servicetalk.transport.netty.internal.AddressUtils.serverHostAndPort;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
 
+// @Timeout(3)
```
Scottmitch, line 58: remove this?

Scottmitch (approved):

idelpivnitskiy:
@Scottmitch I've changed the approach. Instead of splitting logic for pipelined/non-pipelined cases, cancellation is handled in `AbstractStreamingHttpConnection` for any protocol. Please, have another look.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractStreamingHttpConnection.java
```diff
@@ -100,10 +107,64 @@ public final <T> Publisher<? extends T> transportEventStream(final HttpEventKey<
...
     }
 
-    private Single<StreamingHttpResponse> makeRequest(final Publisher<Object> flattenedRequest,
+    private Single<StreamingHttpResponse> makeRequest(final HttpRequestMetaData requestMetaData,
+                                                      final Publisher<Object> flattenedRequest,
                                                       @Nullable final FlushStrategy flushStrategy) {
-        return writeAndRead(flattenedRequest, flushStrategy).liftSyncToSingle(new SpliceFlatStreamToMetaSingle<>(
-                this::newSplicedResponse));
+        @Nullable
+        final OnStreamClosedRunnable onStreamClosed = requestMetaData.context().get(OnStreamClosedRunnable.KEY);
```
Scottmitch, line 114: does this need to be lazy? if not plz add a comment clarifying why.

also consider clarifying why we can't lazy-fetch this value if needed in `canel()` below (then it doesn't look as awkward keeping state outside the operator chain).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractStreamingHttpConnection.java
```diff
@@ -100,10 +107,64 @@ public final <T> Publisher<? extends T> transportEventStream(final HttpEventKey<
...
+                //     logic) helps to capture all possible sources of cancellation.
+                //  2. Intercepting cancel on the caller thread before jumping to the event-loop thread helps to notify
+                //     concurrency controller that the channel is going to close before potentially delivering a
+                //     terminal event back to the response Subscriber (e.g. TimeoutHttpRequesterFilter emits "onError"
+                //     right after propagating "cancel").
+                //  3. Doing it before SpliceFlatStreamToMetaSingle helps to avoid the need for
+                //     BeforeFinallyHttpOperator.
+                //  4. Doing it before offloading of terminal signals helps to reduce the risk of closing a connection
+                //     after response terminates.
+                // We use beforeFinally instead of beforeCancel to avoid closing connection after response terminates.
```
Scottmitch, line 127: thank you for these comments 🔥

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilter.java
```diff
@@ -178,38 +177,7 @@ public StreamingHttpConnectionFilter create(final FilterableStreamingHttpConnect
...
-                                            }
-
-                                            @Override
-                                            public void cancel() {
-                                                // In alignment with cancellation processing in
-                                                // LoadBalancedStreamingHttpClient, we need to close the HTTP/1.x
-                                                // connection when this filter is applied at connection level.
-                                                // Otherwise, a connection will be marked as "free" and can be selected
-                                                // for another request, racing with closure.
-                                                LOGGER.debug("{} closing this {} connection due to cancellation",
```
Scottmitch, line 202: love the code removal here ❤️ 

LOGGER is no longer used in this class -> remove?

Scottmitch (approved):
lgtm after comments addressed

PR #2271, opened 2022-07-12
Grpc upgrade

bondolo:
Contains several changes as separate commits which will be merged using rebase-merge

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -54,7 +54,7 @@ openTracingVersion=0.33.0
 zipkinReporterVersion=2.16.3
 
 # gRPC
-protobufGradlePluginVersion=0.8.18
+protobufGradlePluginVersion=0.8.19
 protobufVersion=3.21.0
 protoGoogleCommonProtosVersion=2.8.1
 javaPoetVersion=1.13.0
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -54,9 +54,9 @@ openTracingVersion=0.33.0
 zipkinReporterVersion=2.16.3
 
 # gRPC
-protobufGradlePluginVersion=0.8.18
-protobufVersion=3.21.0
-protoGoogleCommonProtosVersion=2.8.1
+protobufGradlePluginVersion=0.8.19
+protobufVersion=3.21.2
```
idelpivnitskiy, line 58: Protobuf aligns the version for all programming languages. It's a frequent occasion that their new bug-fix release doesn't have any changes for Java: https://github.com/protocolbuffers/protobuf/releases

In cases like that, it's preferable to avoid upgrading the version as for some users it's important to keep all versions consistent, but that work has overhead with no actual benefits when there are no changes in protobuf.

idelpivnitskiy (approved):
LGTM, consider reverting `protobufVersion`

PR #2273, opened 2022-07-13
Enhance testing of `HttpProviders`

idelpivnitskiy:
Motivation:

Some features, like `build*`, `listen*`, `append*` have multiple
overloads. Our implementation may invoke one from another. We should
make sure this behavior does not cause confusion for `HttpProviders`.

Modifications:

- Verify that provider intercepts only one invocation of the mentioned
methods even when delegating impl invokes one from another;

Result:

More confidence how providers work.


bondolo (approved):
Nice test improvements!

PR #2274, opened 2022-07-14
`LoadBalancedStreamingHttpConnection` doesn't need connect strategy

idelpivnitskiy:
Motivation:

The `connectStrategy` propagated to `AbstractLBHttpConnectionFactory`
should be used only for `newConnection` operation, but not for the
conversions inside `LoadBalancedStreamingHttpConnection`. That strategy
has to be taken from the `ExecutionContext` that takes into account
all filters that are applied to a reserved connection.

Modifications:

- `LoadBalancedStreamingHttpConnection` uses a strategy from execution
context instead of a `connectStrategy`;
- Validate `protocolBinding` in `AbstractLBHttpConnectionFactory` before
allocating `filterableConnectionFactory`;
- Skip the merge of `connectionFactoryStrategy` with
`ProxyConnectConnectionFactoryFilter` bcz we know the internal filters
never block or modify the strategy;

Result:

`LoadBalancedStreamingHttpConnection` uses the correct strategy for
conversions.


bondolo (commented):
I have given this one round of review and it looks good but I want to take another look in a couple of hours.

bondolo (approved):

bondolo (approved):

PR #2277, opened 2022-07-15
Upgrade Gradle wrapper 7.4.2 → 7.5

bondolo:


idelpivnitskiy (commented):
We should also upd the version in docs generation subproject

bondolo:
> We should also upd the version in docs generation subproject

Oops, yes, I forgot that one. I will update the PR

PR #2278, opened 2022-07-18
Make `RequestConcurrencyController` public

idelpivnitskiy:
Motivation:

Users can not wrap connections inside the `LoadBalancer` or
`HttpLoadBalancerFactory` because `RequestConcurrencyController` API is
internal.

Modifications:

- Move `RequestConcurrencyController` and
`ReservableRequestConcurrencyController` from `client-api-internal` to
`client-api`;
- `FilterableStreamingHttpLoadBalancedConnection` now extends
`ReservableRequestConcurrencyController` and
`ReservedStreamingHttpConnection`;
- Introduce `HttpLoadBalancerFactory#toLoadBalancedConnection` overload
that also accepts `ReservableRequestConcurrencyController`, deprecate
pre-existing overload;
- Add `DefaultFilterableStreamingHttpLoadBalancedConnection` that helps
users combine `FilterableStreamingHttpConnection`,
`ReservableRequestConcurrencyController` and `ScoreSupplier` or do
wrapping;
- Remove `LoadBalancedStreamingHttpConnection` as it's not required
anymore;
- Update internals to use new API;

Result:

Users can wrap the `FilterableStreamingHttpLoadBalancedConnection`
inside their `LoadBalancer` or `HttpLoadBalancerFactory` implementations.


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
```diff
@@ -29,16 +42,180 @@
...
+     * @param concurrencyController {@link ReservableRequestConcurrencyController} to control access to the connection
      * @return {@link FilterableStreamingHttpLoadBalancedConnection} for the passed
-     * {@link FilterableStreamingHttpConnection}.
+     * {@link FilterableStreamingHttpConnection}
      */
-    FilterableStreamingHttpLoadBalancedConnection toLoadBalancedConnection(
-            FilterableStreamingHttpConnection connection);
+    default FilterableStreamingHttpLoadBalancedConnection toLoadBalancedConnection(
+            FilterableStreamingHttpConnection connection,
+            ReservableRequestConcurrencyController concurrencyController) {
```
idelpivnitskiy, line 72: I'm open to discussions here. For now, I tried to keep pre-existing assumption that the `HttpLoadBalancerFactory` knows how to implement `score()` for the passed connection. In practice, the `score()` implementation may not be self-sufficient and can rely on external information from the LB or CF. Since we already pass the `ContextMap` through `request -> selectConnection -> newConnection` chain, we can add it here as a 3rd nullable argument too to preserve that context info. LMK WDYT.

idelpivnitskiy: Discussed with @tkountis offline and agreed to add the context as the 3rd arg. It may be helpful to propagate information between layers without API changes in the future. Also, consistency with `selectConnection` and `newConnection`.

idelpivnitskiy:
@tkountis your offline comments addressed, please have another look

tkountis (approved):

tkountis:
Nice work.

PR #2283, opened 2022-07-20
servicetalk-bom to include internal depdencies as API

Scottmitch:
Motivation:
servicetalk-bom currently has `-internal` dependencies as `runtime`. This means
that projects which do depend upon these artifacts for implementation purposes
don't inherit the version from the bom. This can be problematic if the version
of ServiceTalk isn't directly specified by the project but instead inherited
through another bom (multi-level bom). In this case without `api` depdencies
the application must again manually track versions which the bom is intended
to avoid.

### DIFF
--- a/servicetalk-bom/build.gradle
+++ b/servicetalk-bom/build.gradle
@@ -19,10 +19,12 @@ apply plugin: "java-platform"
 
 description="ServiceTalk BOM that includes all modules"
 
-rootProject.subprojects.findAll { !it.name.endsWith("-bom") && !it.name.endsWith("-dependencies") &&
-                                            !it.name.contains("examples") &&
-                                          !it.name.equals("grpc") && !it.name.equals("http") }.each {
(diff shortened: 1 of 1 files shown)


idelpivnitskiy (approved):

bondolo:
It is unfortunate that Gradle requires explicit version specification for internal modules.

I am not saying we shouldn't make this change or not have internal modules but we may need some code reorganization if indeed interfaces currently in internal need to be exposed as public API.
 
We should consider, over time, exposing the interfaces/classes currently in `internal` modules in API modules if the interfaces are really part of our public API and needed in external code.

Where the interfaces are not needed by external code, treating internal and external modules differently is an important and useful property because we do not offer API stability guarantees on our internal modules and want to preserve that capability.

bondolo (approved):

Scottmitch:
ack. this PR is intended to decouple "find publish API equivalents for `-internal` features" (which may take a long time) with "consumers using consistent versions" (needed immediately).

PR #2286, opened 2022-07-22
Add `.p12` stores for tests which require format

bondolo:
Motivation:
Some new tests require certificates and keys to be presented in PKCS#12
keystore format.
Modifications:
Add PKCS#12 stores for client, server and truststore. Certs are
regenerated as server CA cert was expired and could not be used for
signing of localhost_server server cert. Key size is updated to 3072
bits, the current FIPS 140-2 SP800 requirement for RSA keys as this is
what will typically be required for commercial use.
Result:
More test resources


servicetalk-test-resources/src/main/java/io/servicetalk/test/resources/DefaultTestCerts.java
```diff
@@ -44,6 +44,15 @@ public static InputStream loadServerPem() {
         return DefaultTestCerts.class.getResourceAsStream("localhost_server.pem");
     }
 
+    /**
+     * Load the server certificate PKCS#12 keystore file.
+     *
+     * @return an {@link InputStream} from the server key/certificate keystore file.
+     */
+    public static InputStream loadServerP12() {
```
idelpivnitskiy, line 52: Will be great to add an example for PKCS#12 later as it might be a bit trickier for users.

idelpivnitskiy (approved):

servicetalk-test-resources/src/main/java/io/servicetalk/test/resources/DefaultTestCerts.java
bondolo: I agree. I ended up copying the very weak test pattern used for the other resources but believe that more complete tests should be created for each resource.

bondolo: I will create an RFE issue. This would be a good "first contribution" issue.

bondolo: https://github.com/apple/servicetalk/issues/2290

idelpivnitskiy: I was talking about examples, not tests. Creating an `SslConfig` with pre-existing pem certs is straightforward, users just need input stream. We already do that in examples: https://github.com/apple/servicetalk/blob/14e9fa7e16bbf2512dfb194bc075c917f5293e8c/servicetalk-examples/http/mutual-tls/src/main/java/io/servicetalk/examples/http/mutualtls/HttpClientMutualTLS.java#L37

This approach won't work for PKCS#12. Users have to load them manually and go through the trust/key factory. It's not obvious how to do that.

PR #2301, opened 2022-08-01
Use `compileJava` `options.release`

bondolo:
Motivation:
Using `sourceCompatibility` and `targetCompatibility` for specifying
Java version does not always produce correct module metadata because
the setting is quite sensitive to order of operations.
Modifications:
Use `options.release` whenever build JDK is >= Java 9. To support `javac`
compilation using the `--release` option it is necessary to hide direct use
of the `sun.misc.Unsafe` class as this class is not provided in the symbols
used for compiling.
Result:
Correct Java variant versions exported in module files

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -18,6 +18,7 @@
 org.gradle.parallel=true
 org.gradle.caching=true
 org.gradle.configureondemand=true
+org.gradle.java.installations.auto-download=false
 org.gradle.jvmargs=-Xms2g -Xmx4g -dsa -da -ea:io.servicetalk... -XX:+HeapDumpOnOutOfMemoryError
 
 # project metadata used for publications
--- a/servicetalk-buffer-api/build.gradle
+++ b/servicetalk-buffer-api/build.gradle
@@ -28,5 +28,6 @@ dependencies {
   testImplementation "org.junit.jupiter:junit-jupiter-params"
   testImplementation "org.hamcrest:hamcrest:$hamcrestVersion"
   testImplementation "org.mockito:mockito-core:$mockitoCoreVersion"
+
   testFixturesImplementation "org.hamcrest:hamcrest:$hamcrestVersion"
 }
--- a/servicetalk-concurrent-jdkflow/build.gradle
+++ b/servicetalk-concurrent-jdkflow/build.gradle
@@ -16,17 +16,24 @@
 
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-library"
 
-if (!JavaVersion.current().isJava9Compatible()) {
+// Required version for module
+def javaLanguageVersion = JavaVersion.VERSION_1_9
+
+if (!JavaVersion.current().isCompatibleWith(javaLanguageVersion)) {
     project.tasks.all { task -> task.enabled = false }
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_9
-    targetCompatibility = JavaVersion.VERSION_1_9
+    sourceCompatibility = javaLanguageVersion
+    targetCompatibility = javaLanguageVersion
 }
 
 compileJava {
-    options.compilerArgs.addAll(['--release', '9'])
+    options.release = Integer.parseInt(javaLanguageVersion.getMajorVersion())
+}
+
+compileTestJava {
+    options.release = compileJava.options.release
 }
 
 dependencies {
--- a/servicetalk-gradle-plugin-internal/src/main/groovy/io/servicetalk/gradle/plugin/internal/ServiceTalkLibraryPlugin.groovy
+++ b/servicetalk-gradle-plugin-internal/src/main/groovy/io/servicetalk/gradle/plugin/internal/ServiceTalkLibraryPlugin.groovy
@@ -19,7 +19,6 @@ import com.github.spotbugs.snom.SpotBugsTask
 import info.solidsoft.gradle.pitest.PitestTask
 import org.gradle.api.JavaVersion
 import org.gradle.api.Project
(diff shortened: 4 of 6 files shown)


gradle.properties
```diff
@@ -18,6 +18,7 @@
 org.gradle.parallel=true
 org.gradle.caching=true
 org.gradle.configureondemand=true
+org.gradle.java.installations.auto-download=false
```
bondolo, line 21: This change is not strictly required.

It disables Gradle from automatically downloading of JDK versions for use for building. When disabled all of the required Java toolchain must be present in the environment, which is what we expect.

bondolo:
I am willing to move the `sun.misc.Unsafe` work out to a separate PR that would be completed first.

servicetalk-buffer-api/build.gradle
```diff
@@ -28,5 +28,12 @@ dependencies {
   testImplementation "org.junit.jupiter:junit-jupiter-params"
   testImplementation "org.hamcrest:hamcrest:$hamcrestVersion"
   testImplementation "org.mockito:mockito-core:$mockitoCoreVersion"
+
   testFixturesImplementation "org.hamcrest:hamcrest:$hamcrestVersion"
+
+  if (JavaVersion.current().isJava9Compatible()) {
+    compileTestFixturesJava {
+      options.release = compileJava.options.release
```
idelpivnitskiy, line 36: Instead of adding this code snippet in every `build.gradle`, we should rely on our gradle plugin to set it. The plugin is a single source of truth for consistent gradle configuration across all modules.
This can be repeated for submodules that do not apply our plugin.

servicetalk-concurrent-jdkflow/build.gradle
```diff
@@ -20,13 +20,12 @@ if (!JavaVersion.current().isJava9Compatible()) {
     project.tasks.all { task -> task.enabled = false }
 }
 
-java {
-    sourceCompatibility = JavaVersion.VERSION_1_9
-    targetCompatibility = JavaVersion.VERSION_1_9
+compileJava {
+    options.release = 9
```
idelpivnitskiy, line 24: Can we keep `sourceCompatibility` and `targetCompatibility` too? IIRC, some other tools, like IDE, look at them to generate project meta-data correctly.

servicetalk-gradle-plugin-internal/src/main/groovy/io/servicetalk/gradle/plugin/internal/ServiceTalkLibraryPlugin.groovy
```diff
@@ -56,8 +57,16 @@ final class ServiceTalkLibraryPlugin extends ServiceTalkCorePlugin {
     project.configure(project) {
       pluginManager.apply("java-library")
 
-      sourceCompatibility = TARGET_VERSION
-      targetCompatibility = TARGET_VERSION
```
idelpivnitskiy, line 60: Same questions here. I'm hesitant to remove these properties. Your change doesn't look conflicting.

idelpivnitskiy (changes requested):

servicetalk-buffer-api/build.gradle
bondolo: This snippet is only in the `build.gradle` of modules which add `testFixtures`. Unless the test fixtures are added then the `compileTestFixturesJava` task is undefined which means this snippet can't be added to the plugin. I was unable, thus far, to find a way to make the behaviour conditional so that happened at the right time and only for the projects with a `compileTestFixturesJava` task.

servicetalk-concurrent-jdkflow/build.gradle
bondolo: Sure, as long as it does not cause problems. My reading of https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_cross_compilation

was that these had been superseded by `options.release` and were no longer required.

bondolo: > Using a toolchain makes it illegal to configure the `sourceCompatibility` or `targetCompatibility` at the java { } extension level.

It appears that these properties are partially obsolete and may eventually be deprecated. I had looked at using `toolchain` rather than `options.release` but it doesn't quite do what we need.

servicetalk-buffer-api/build.gradle
idelpivnitskiy: Gradle allows searching for tasks by their string name: `tasks.findByName("someName")`. It helps to apply logic conditionally. There are a few examples inside our plugin.

servicetalk-concurrent-jdkflow/build.gradle
idelpivnitskiy: Until they are officially deprecated, let's keep them. It doesn't hurt to have some redundancy. Those options have 1-1 translation to pom.xml properties and may be useful for IDE.

bondolo: I am fine with keeping them as long as they don't cause problems.

bondolo:
This will have to wait for PR #2303 to merge.

bondolo:
Marked as draft until PR #2303 is merged.

gradle.properties
idelpivnitskiy: Should we set the same property for `docs` subproject?

idelpivnitskiy (approved):

gradle.properties
bondolo: yes, I would argue for setting it in all of our gradle.properties files to prevent downloading and building with random JDK installations.

PR #2303, opened 2022-08-02
Hide use of `sun.misc.Unsafe` from `javac`

bondolo:
Motivation:
Attempting to build using a Java 9+ `javac` with the `--release`
option fails because the ServicetTalk source contains references to
the `sun.misc.Unsafe` class. This class is not required by ServiceTalk
for correct execution but the appearance of the class in the source
currently requires that source be compiled with Java 8 or without
the `--release` option.
Modifications:
Use reflection and method handles rather than direct class use for
`sun.misc.Unsafe` to avoid [JDK-8214165](https://bugs.openjdk.org/browse/JDK-8214165).
Result:
ServiceTalk may be compiled using the `javac --release` option.

### DIFF
--- a/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
+++ b/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
@@ -32,26 +32,29 @@ package io.servicetalk.utils.internal;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import sun.misc.Unsafe;
 
 import java.io.FileDescriptor;
+import java.lang.invoke.CallSite;
+import java.lang.invoke.LambdaMetafactory;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.nio.ByteBuffer;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.util.Objects;
+import java.util.function.Consumer;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.utils.internal.ReflectionUtils.extractNioBitsMethod;
 import static io.servicetalk.utils.internal.ReflectionUtils.lookupAccessibleObject;
 import static java.lang.Boolean.getBoolean;
-import static java.util.Objects.requireNonNull;
 
 /**
  * {@link PlatformDependent} operations that require access to {@code sun.misc.*}.
- *
+ * <p/>
  * This class is forked from the netty project and modified to suit our needs.
  */
 final class PlatformDependent0 {
@@ -63,7 +66,7 @@ final class PlatformDependent0 {
     private static final String DEALLOCATOR_CLASS_NAME = "java.nio.DirectByteBuffer$Deallocator";
 
     @Nullable
-    private static final Unsafe UNSAFE;
+    private static final Object UNSAFE;
     @Nullable
     private static final MethodHandle DIRECT_BUFFER_CONSTRUCTOR;
     @Nullable
@@ -73,10 +76,19 @@ final class PlatformDependent0 {
     @Nullable
     private static final MethodHandle UNRESERVE_MEMORY;
 
+    @Nullable
+    private static final MethodHandle ALLOCATE_MEMORY;
+
+    @Nullable
+    private static final MethodHandle FREE_MEMORY;
+
+    @Nullable
+    private static final Consumer<Throwable> THROW_EXCEPTION;
+
     private static final boolean USE_DIRECT_BUFFER_WITHOUT_ZEROING;
 
     static {
-        Unsafe unsafe;
+        Object unsafe;
 
         if (IS_EXPLICIT_NO_UNSAFE) {
             unsafe = null;
@@ -84,14 +96,15 @@ final class PlatformDependent0 {
             // attempt to access field Unsafe#theUnsafe
             final Object maybeUnsafe = AccessController.doPrivileged((PrivilegedAction<Object>) () -> {
                 try {
-                    final Field unsafeField = Unsafe.class.getDeclaredField("theUnsafe");
+                    Class<?> unsafeClass = Class.forName("sun.misc.Unsafe");
+                    final Field unsafeField = unsafeClass.getDeclaredField("theUnsafe");
                     final Throwable cause = ReflectionUtils.trySetAccessible(unsafeField, false);
                     if (cause != null) {
                         return cause;
                     }
                     // the unsafe instance
                     return unsafeField.get(null);
-                } catch (NoSuchFieldException | SecurityException | IllegalAccessException e) {
+                } catch (ClassNotFoundException | NoSuchFieldException | SecurityException | IllegalAccessException e) {
                     return e;
                 }
             });
@@ -100,15 +113,55 @@ final class PlatformDependent0 {
             // is an instanceof Unsafe and reversing the if and else blocks; this is because an
             // instanceof check against Unsafe will trigger a class load and we might not have
             // the runtime permission accessClassInPackage.sun.misc
-            if (maybeUnsafe instanceof Exception) {
+            if (maybeUnsafe instanceof Throwable) {
                 unsafe = null;
-                LOGGER.debug("sun.misc.Unsafe.theUnsafe: unavailable", (Exception) maybeUnsafe);
+                LOGGER.debug("sun.misc.Unsafe.theUnsafe: unavailable", (Throwable) maybeUnsafe);
             } else {
-                unsafe = (Unsafe) maybeUnsafe;
-                LOGGER.debug("sun.misc.Unsafe.theUnsafe: available");
+                unsafe = maybeUnsafe;
+                LOGGER.debug("sun.misc.Unsafe.theUnsafe: {}", null == unsafe ? "unavailable" : "available");
             }
         }
         UNSAFE = unsafe;
+
+        if (null == unsafe) {
+            THROW_EXCEPTION = null;
+            ALLOCATE_MEMORY = null;
+            FREE_MEMORY = null;
+        } else {
+            Consumer<Throwable> throwConsumer = null;
+            MethodHandle allocateMH = null;
+            MethodHandle freeMH = null;
+            MethodHandles.Lookup lookup = MethodHandles.lookup();
+            MethodHandle throwExceptionMH = null;
+            try {
+                throwExceptionMH = lookup.findVirtual(unsafe.getClass(), "throwException",
+                        MethodType.methodType(void.class, Throwable.class));
+                CallSite throwExceptionCallSite = LambdaMetafactory.metafactory(
+                        lookup, "accept",
+                        MethodType.methodType(Consumer.class, unsafe.getClass()),
+                        MethodType.methodType(void.class, Object.class),
+                        throwExceptionMH,
+                        MethodType.methodType(void.class, Throwable.class));
+                throwConsumer = (Consumer<Throwable>) throwExceptionCallSite.getTarget().bindTo(unsafe).invoke();
+            } catch (Throwable all) {
+                LOGGER.warn("failed computing THROW_EXCEPTION", all);
+            }
+            try {
+                allocateMH = lookup.findVirtual(unsafe.getClass(), "allocateMemory",
+                        MethodType.methodType(long.class, long.class)).bindTo(unsafe);
+            } catch (Throwable all) {
+                LOGGER.warn("failed computing THROW_EXCEPTION", all);
+            }
+            try {
+                freeMH = lookup.findVirtual(unsafe.getClass(), "freeMemory",
+                        MethodType.methodType(void.class, long.class)).bindTo(unsafe);
+            } catch (Throwable all) {
+                LOGGER.warn("failed computing THROW_EXCEPTION", all);
+            }
+            THROW_EXCEPTION = throwConsumer;
+            ALLOCATE_MEMORY = allocateMH;
+            FREE_MEMORY = freeMH;
+        }
 
         final ByteBuffer direct;
         final MethodHandles.Lookup lookup;
@@ -215,12 +268,21 @@ final class PlatformDependent0 {
 
     static long allocateMemory(final long size) {
         assert UNSAFE != null;
-        return UNSAFE.allocateMemory(size);
+        try {
+            return (long) ALLOCATE_MEMORY.invokeExact(size);
+        } catch (Throwable e) {
+            throwException(e);
+            return Long.MIN_VALUE; // never reached
+        }
     }
 
     static void freeMemory(final long address) {
         assert UNSAFE != null;
-        UNSAFE.freeMemory(address);
+        try {
+            FREE_MEMORY.invokeExact(address);
+        } catch (Throwable e) {
+            throwException(e);
+        }
     }
 
     static ByteBuffer newDirectBuffer(final long address, final long size, final int capacity) {
@@ -241,6 +303,14 @@ final class PlatformDependent0 {
     static void throwException(Throwable cause) {
         assert UNSAFE != null;
         // JVM has been observed to crash when passing a null argument. See https://github.com/netty/netty/issues/4131.
-        UNSAFE.throwException(requireNonNull(cause));
+        try {
+            THROW_EXCEPTION.accept(Objects.requireNonNull(cause));
+        } catch (Throwable all) {
+            sneakyThrows(cause);
+        }
+    }
+
+    private static <E extends Throwable> void sneakyThrows(final Throwable t) throws E {
+        throw (E) t;
     }
 }
--- a/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
+++ b/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
@@ -150,13 +150,9 @@ final class PlatformDependent0 {
                 Consumer<Throwable> unsafeThrowConsumer =
                         (Consumer<Throwable>) throwExceptionCallSite.getTarget().bindTo(unsafe).invoke();
                 throwConsumer = (t) -> {
-                    try {
-                        // JVM has been observed to crash when passing a null argument.
-                        // See https://github.com/netty/netty/issues/4131.
-                        unsafeThrowConsumer.accept(null != t ? t : new NullPointerException("Throwable was null"));
-                    } catch (Throwable all) {
-                        throwException0(t);
-                    }
+                    // JVM has been observed to crash when passing a null argument.
+                    // See https://github.com/netty/netty/issues/4131.
+                    unsafeThrowConsumer.accept(null != t ? t : new NullPointerException("Throwable was null"));
                 };
                 LOGGER.debug("sun.misc.Unsafe#throwException(Throwable): available");
             } catch (Throwable t) {
--- a/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
+++ b/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
@@ -44,7 +44,6 @@ import java.lang.reflect.Field;
 import java.nio.ByteBuffer;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.Objects;
 import java.util.function.Consumer;
 import javax.annotation.Nullable;
 
@@ -54,7 +53,7 @@ import static java.lang.Boolean.getBoolean;
 
 /**
  * {@link PlatformDependent} operations that require access to {@code sun.misc.*}.
- * <p/>
+ * <p>
  * This class is forked from the netty project and modified to suit our needs.
  */
 final class PlatformDependent0 {
@@ -65,8 +64,9 @@ final class PlatformDependent0 {
 
     private static final String DEALLOCATOR_CLASS_NAME = "java.nio.DirectByteBuffer$Deallocator";
 
+    @Deprecated
     @Nullable
-    private static final Object UNSAFE;
+    private static final Object UNSAFE; // FIXME: 0.43 - remove deprecated constant
     @Nullable
     private static final MethodHandle DIRECT_BUFFER_CONSTRUCTOR;
     @Nullable
@@ -86,6 +86,7 @@ final class PlatformDependent0 {
     private static final Consumer<Throwable> THROW_EXCEPTION;
 
     private static final boolean USE_DIRECT_BUFFER_WITHOUT_ZEROING;
+    private static final Object DUMMY = new Object();
 
     static {
         Object unsafe;
@@ -96,8 +97,7 @@ final class PlatformDependent0 {
             // attempt to access field Unsafe#theUnsafe
             final Object maybeUnsafe = AccessController.doPrivileged((PrivilegedAction<Object>) () -> {
                 try {
-                    Class<?> unsafeClass = Class.forName("sun.misc.Unsafe");
-                    final Field unsafeField = unsafeClass.getDeclaredField("theUnsafe");
+                    final Field unsafeField = Class.forName("sun.misc.Unsafe").getDeclaredField("theUnsafe");
                     final Throwable cause = ReflectionUtils.trySetAccessible(unsafeField, false);
                     if (cause != null) {
                         return cause;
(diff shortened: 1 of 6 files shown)


bondolo:
This change has been extracted from PR #2301 and is now a pre-requisite of that PR.

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
```diff
@@ -100,16 +113,56 @@ final class PlatformDependent0 {
...
+        } else {
+            Consumer<Throwable> throwConsumer = null;
+            MethodHandle allocateMH = null;
+            MethodHandle freeMH = null;
+            MethodHandles.Lookup lookup = MethodHandles.lookup();
+            MethodHandle throwExceptionMH = null;
+            try {
+                throwExceptionMH = lookup.findVirtual(unsafe.getClass(), "throwException",
+                        MethodType.methodType(void.class, Throwable.class));
+                CallSite throwExceptionCallSite = LambdaMetafactory.metafactory(
```
idelpivnitskiy, line 139: What is the benefit of using a lambda here instead of a direct use of a `MethodHandle` like in all other cases?

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
```diff
@@ -73,25 +76,35 @@ final class PlatformDependent0 {
...
 
         if (IS_EXPLICIT_NO_UNSAFE) {
             unsafe = null;
         } else {
             // attempt to access field Unsafe#theUnsafe
             final Object maybeUnsafe = AccessController.doPrivileged((PrivilegedAction<Object>) () -> {
                 try {
-                    final Field unsafeField = Unsafe.class.getDeclaredField("theUnsafe");
+                    Class<?> unsafeClass = Class.forName("sun.misc.Unsafe");
+                    final Field unsafeField = unsafeClass.getDeclaredField("theUnsafe");
```
idelpivnitskiy, line 100: Consider the following changes:
```patch
Index: servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent.java b/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent.java
--- a/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent.java	(revision a277cc5937f38e850170daf7cf722c310f391a14)
+++ b/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent.java	(date 1659587537325)
@@ -50,7 +50,6 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.lang.reflect.Field;
 import java.nio.ByteBuffer;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
@@ -76,7 +75,6 @@
     private static final int MAX_ALLOWED_QUEUE_CAPACITY = Pow2.MAX_POW2;
     private static final int MIN_ALLOWED_SPSC_CHUNK_SIZE = 8; // JCTools does not allow lower initial capacity.
     private static final int MIN_ALLOWED_MPSC_CHUNK_SIZE = 2; // JCTools does not allow lower initial capacity.
-    private static final Object DUMMY = new Object();
 
     private PlatformDependent() {
         // no instantiation
@@ -86,8 +84,10 @@
      * Checks if {@code sun.misc.Unsafe} is available and has not been disabled.
      *
      * @return {@code true} if {@code sun.misc.Unsafe} is available.
+     * @deprecated This method is not used by internal ServiceTalk code anymore and will be removed in future releases.
      */
-    public static boolean hasUnsafe() {
+    @Deprecated
+    public static boolean hasUnsafe() { // FIXME: 0.43 - remove deprecated method
         return PlatformDependent0.hasUnsafe();
     }
 
@@ -160,26 +160,11 @@
     * @return nothing actually will be returned from this method because it rethrows the specified exception. Making
     * this method return an arbitrary type makes the caller method easier as they do not have to add a return statement
     * after calling this method.
+    * @deprecated Use {@link ThrowableUtils#throwException(Throwable)}.
     */
-    public static <T> T throwException(final Throwable t) {
-        if (hasUnsafe()) {
-            PlatformDependent0.throwException(t);
-        } else {
-            PlatformDependent.<RuntimeException>throwException0(t);
-        }
-        // This will never be invoked at runtime because each branch above with rethrow the passed Throwable.
-        // However, this is necessary to fool the compiler.
-        return uncheckedCast();
-    }
-
-    @SuppressWarnings("unchecked")
-    private static <E extends Throwable> void throwException0(final Throwable t) throws E {
-        throw (E) t;
-    }
-
-    @SuppressWarnings("unchecked")
-    private static <T> T uncheckedCast() {
-        return (T) DUMMY;
+    @Deprecated
+    public static <T> T throwException(final Throwable t) { // FIXME: 0.43 - remove deprecated method
+        return ThrowableUtils.throwException(t);
     }
 
     /**
@@ -312,25 +297,20 @@
         }
 
         static {
-            Object unsafe = null;
-            if (hasUnsafe()) {
-                // jctools goes through its own process of initializing unsafe; of
-                // course, this requires permissions which might not be granted to calling code, so we
-                // must mark this block as privileged too
-                unsafe = AccessController.doPrivileged((PrivilegedAction<Object>) () -> {
-                    // force JCTools to initialize unsafe
-                    try {
-                        Field unsafeStatic = UnsafeAccess.class.getDeclaredField("UNSAFE");
-                        return unsafeStatic.get(null);
-                    } catch (IllegalAccessException | NoSuchFieldException e) {
-                        LOGGER.debug("jctools unsafe failed: ", e);
-                        return null;
-                    }
-                });
-            }
+            // jctools goes through its own process of initializing unsafe; of
+            // course, this requires permissions which might not be granted to calling code, so we
+            // must mark this block as privileged too
+            Object maybeUnsafe = AccessController.doPrivileged((PrivilegedAction<Object>) () -> {
+                // force JCTools to initialize unsafe
+                try {
+                    return UnsafeAccess.class.getDeclaredField("UNSAFE").get(null);
+                } catch (IllegalAccessException | NoSuchFieldException e) {
+                    return e;
+                }
+            });
 
-            if (unsafe == null) {
-                LOGGER.debug("jctools Unbounded/ChunkedArrayQueue: unavailable.");
+            if (maybeUnsafe instanceof Throwable) {
+                LOGGER.debug("jctools Unbounded/ChunkedArrayQueue: unavailable.", (Throwable) maybeUnsafe);
                 USE_UNSAFE_QUEUES = false;
             } else {
                 LOGGER.debug("jctools Unbounded/ChunkedArrayQueue: available.");
Index: servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/ThrowableUtilTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/ThrowableUtilTest.java b/servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/ThrowableUtilsTest.java
rename from servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/ThrowableUtilTest.java
rename to servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/ThrowableUtilsTest.java
--- a/servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/ThrowableUtilTest.java	(revision a277cc5937f38e850170daf7cf722c310f391a14)
+++ b/servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/ThrowableUtilsTest.java	(date 1659634527083)
@@ -24,8 +24,15 @@
 import static org.hamcrest.Matchers.nullValue;
 import static org.hamcrest.Matchers.sameInstance;
 import static org.hamcrest.core.Is.is;
 
-class ThrowableUtilTest {
+class ThrowableUtilsTest {
+
+    @Test
+    void throwException() {
+        Exception customException = new Exception("Deliberate Exception");
+        assertThrows(customException.getClass(), () -> ThrowableUtils.throwException(customException));
+    }
 
     @Test
     void combineNonThrowable() {
Index: servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/PlatformDependentTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/PlatformDependentTest.java b/servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/PlatformDependentTest.java
--- a/servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/PlatformDependentTest.java	(revision a277cc5937f38e850170daf7cf722c310f391a14)
+++ b/servicetalk-utils-internal/src/test/java/io/servicetalk/utils/internal/PlatformDependentTest.java	(date 1659634527089)
@@ -41,12 +41,4 @@
         assertThat(allocated, is(Matchers.not(0)));
         PlatformDependent.freeMemory(allocated);
     }
-
-    @Test
-    void throwException() {
-        assumeTrue(PlatformDependent.hasUnsafe(), "Unsafe absent or disabled");
-        Exception customException = new Exception() {
-        };
-        assertThrows(customException.getClass(), () -> PlatformDependent.throwException(customException));
-    }
 }
Index: servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java b/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
--- a/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java	(revision a277cc5937f38e850170daf7cf722c310f391a14)
+++ b/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java	(date 1659634527094)
@@ -44,17 +44,17 @@
 import java.nio.ByteBuffer;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.Objects;
 import java.util.function.Consumer;
 
 import static io.servicetalk.utils.internal.ReflectionUtils.extractNioBitsMethod;
 import static io.servicetalk.utils.internal.ReflectionUtils.lookupAccessibleObject;
 import static java.lang.Boolean.getBoolean;
 
 /**
  * {@link PlatformDependent} operations that require access to {@code sun.misc.*}.
- * <p/>
+ * <p>
  * This class is forked from the netty project and modified to suit our needs.
  */
 final class PlatformDependent0 {
@@ -65,8 +65,9 @@
 
     private static final String DEALLOCATOR_CLASS_NAME = "java.nio.DirectByteBuffer$Deallocator";
 
+    @Deprecated
     @Nullable
-    private static final Object UNSAFE;
+    private static final Object UNSAFE; // FIXME: 0.43 - remove deprecated constant
     @Nullable
     private static final MethodHandle DIRECT_BUFFER_CONSTRUCTOR;
     @Nullable
@@ -86,6 +87,7 @@
     private static final Consumer<Throwable> THROW_EXCEPTION;
 
     private static final boolean USE_DIRECT_BUFFER_WITHOUT_ZEROING;
+    private static final Object DUMMY = new Object();
 
     static {
         Object unsafe;
@@ -96,8 +98,7 @@
             // attempt to access field Unsafe#theUnsafe
             final Object maybeUnsafe = AccessController.doPrivileged((PrivilegedAction<Object>) () -> {
                 try {
-                    Class<?> unsafeClass = Class.forName("sun.misc.Unsafe");
-                    final Field unsafeField = unsafeClass.getDeclaredField("theUnsafe");
+                    final Field unsafeField = Class.forName("sun.misc.Unsafe").getDeclaredField("theUnsafe");
                     final Throwable cause = ReflectionUtils.trySetAccessible(unsafeField, false);
                     if (cause != null) {
                         return cause;
@@ -109,30 +110,34 @@
                 }
             });
 
-            // the conditional check here can not be replaced with checking that maybeUnsafe
+            // The conditional check here can not be replaced with checking that maybeUnsafe
             // is an instanceof Unsafe and reversing the if and else blocks; this is because an
             // instanceof check against Unsafe will trigger a class load and we might not have
-            // the runtime permission accessClassInPackage.sun.misc
+            // the runtime permission accessClassInPackage.sun.misc.
+            // We also try to avoid any direct reference to sun.misc.Unsafe to be able to compile with `--release 8`
+            // flag on JDK9+, see https://bugs.openjdk.org/browse/JDK-8214165.
             if (maybeUnsafe instanceof Throwable) {
                 unsafe = null;
                 LOGGER.debug("sun.misc.Unsafe.theUnsafe: unavailable", (Throwable) maybeUnsafe);
             } else {
                 unsafe = maybeUnsafe;
-                LOGGER.debug("sun.misc.Unsafe.theUnsafe: {}", null == unsafe ? "unavailable" : "available");
+                LOGGER.debug("sun.misc.Unsafe.theUnsafe: available");
             }
         }
-        UNSAFE = unsafe;
 
+        final MethodHandles.Lookup lookup;
         if (null == unsafe) {
+            lookup = null;
             THROW_EXCEPTION = null;
             ALLOCATE_MEMORY = null;
             FREE_MEMORY = null;
+            UNSAFE = null;
         } else {
+            lookup = MethodHandles.lookup();
             Consumer<Throwable> throwConsumer = null;
+            MethodHandle throwExceptionMH = null;
             MethodHandle allocateMH = null;
             MethodHandle freeMH = null;
-            MethodHandles.Lookup lookup = MethodHandles.lookup();
-            MethodHandle throwExceptionMH = null;
             try {
                 throwExceptionMH = lookup.findVirtual(unsafe.getClass(), "throwException",
                         MethodType.methodType(void.class, Throwable.class));
@@ -142,37 +147,46 @@
                         MethodType.methodType(void.class, Object.class),
                         throwExceptionMH,
                         MethodType.methodType(void.class, Throwable.class));
+                //noinspection unchecked
                 throwConsumer = (Consumer<Throwable>) throwExceptionCallSite.getTarget().bindTo(unsafe).invoke();
-            } catch (Throwable all) {
-                LOGGER.warn("failed computing THROW_EXCEPTION", all);
+                LOGGER.debug("sun.misc.Unsafe#throwException(Throwable): available");
+                LOGGER.debug("sun.misc.Unsafe#throwException(Throwable): unavailable", t);
             }
+            Long address = null;
             try {
                 allocateMH = lookup.findVirtual(unsafe.getClass(), "allocateMemory",
                         MethodType.methodType(long.class, long.class)).bindTo(unsafe);
-            } catch (Throwable all) {
-                LOGGER.warn("failed computing THROW_EXCEPTION", all);
+                address = (long) allocateMH.invokeExact(1L);
+                LOGGER.debug("sun.misc.Unsafe#allocateMemory(long): available");
+                LOGGER.debug("sun.misc.Unsafe#allocateMemory(long): unavailable", t);
             }
             try {
                 freeMH = lookup.findVirtual(unsafe.getClass(), "freeMemory",
                         MethodType.methodType(void.class, long.class)).bindTo(unsafe);
-            } catch (Throwable all) {
-                LOGGER.warn("failed computing THROW_EXCEPTION", all);
+                if (address != null) {
+                    freeMH.invokeExact((long) address);
+                }
+                LOGGER.debug("sun.misc.Unsafe#freeMemory(long): available");
+                LOGGER.debug("sun.misc.Unsafe#freeMemory(long): unavailable", t);
             }
             THROW_EXCEPTION = throwConsumer;
             ALLOCATE_MEMORY = allocateMH;
             FREE_MEMORY = freeMH;
+            // Mark Unsafe as available only if all required Unsafe methods are available too:
+            UNSAFE = THROW_EXCEPTION != null && ALLOCATE_MEMORY != null && FREE_MEMORY != null ? unsafe : null;
         }
+        LOGGER.debug("sun.misc.Unsafe: {}", UNSAFE != null ? "available" : "unavailable");
 
         final ByteBuffer direct;
-        final MethodHandles.Lookup lookup;
         // Define DIRECT_BUFFER_CONSTRUCTOR:
         if (UNSAFE == null) {
             direct = null;
-            lookup = null;
             DIRECT_BUFFER_CONSTRUCTOR = null;
         } else {
             direct = ByteBuffer.allocateDirect(1);
-            lookup = MethodHandles.lookup();
             DIRECT_BUFFER_CONSTRUCTOR = lookupAccessibleObject(() -> direct.getClass().getDeclaredConstructor(
                     int.class, long.class, FileDescriptor.class, Runnable.class), Constructor.class, constructor -> {
 
@@ -231,7 +245,8 @@
 
         // Define USE_DIRECT_BUFFER_WITHOUT_ZEROING:
         USE_DIRECT_BUFFER_WITHOUT_ZEROING = UNSAFE != null && DIRECT_BUFFER_CONSTRUCTOR != null &&
-                DEALLOCATOR_CONSTRUCTOR != null && RESERVE_MEMORY != null && UNRESERVE_MEMORY != null;
+                DEALLOCATOR_CONSTRUCTOR != null && RESERVE_MEMORY != null && UNRESERVE_MEMORY != null &&
+                ALLOCATE_MEMORY != null && FREE_MEMORY != null;
         LOGGER.debug("Allocation of DirectByteBuffer without zeroing memory: {}",
                 USE_DIRECT_BUFFER_WITHOUT_ZEROING ? "available" : "unavailable");
     }
@@ -240,7 +255,8 @@
         // no instantiation
     }
 
-    static boolean hasUnsafe() {
+    @Deprecated
+    static boolean hasUnsafe() {    // FIXME: 0.43 - remove deprecated method
         return UNSAFE != null;
     }
 
@@ -253,7 +269,7 @@
         try {
             RESERVE_MEMORY.invoke(size, capacity);
         } catch (Throwable t) {
-            throw new Error(t);
+            throwException(t);
         }
     }
 
@@ -262,26 +278,26 @@
         try {
             UNRESERVE_MEMORY.invoke(size, capacity);
         } catch (Throwable t) {
-            throw new Error(t);
+            throwException(t);
         }
     }
 
     static long allocateMemory(final long size) {
-        assert UNSAFE != null;
+        assert ALLOCATE_MEMORY != null;
         try {
             return (long) ALLOCATE_MEMORY.invokeExact(size);
-        } catch (Throwable e) {
-            throwException(e);
+            throwException(t);
             return Long.MIN_VALUE; // never reached
         }
     }
 
     static void freeMemory(final long address) {
-        assert UNSAFE != null;
+        assert FREE_MEMORY != null;
         try {
             FREE_MEMORY.invokeExact(address);
-        } catch (Throwable e) {
-            throwException(e);
+            throwException(t);
         }
     }
 
@@ -300,17 +316,30 @@
         }
     }
 
-    static void throwException(Throwable cause) {
-        assert UNSAFE != null;
-        // JVM has been observed to crash when passing a null argument. See https://github.com/netty/netty/issues/4131.
-        try {
-            THROW_EXCEPTION.accept(Objects.requireNonNull(cause));
-        } catch (Throwable all) {
-            sneakyThrows(cause);
+    static <T> T throwException(Throwable t) {
+        if (THROW_EXCEPTION != null) {
+            try {
+                // JVM has been observed to crash when passing a null argument.
+                // See https://github.com/netty/netty/issues/4131.
+                THROW_EXCEPTION.accept(requireNonNull(t));
+            } catch (Throwable all) {
+                throwException0(t);
+            }
+        } else {
+            throwException0(t);
         }
+        // This will never be invoked at runtime because each branch above with rethrow the passed Throwable.
+        // However, this is necessary to fool the compiler.
+        return uncheckedCast();
     }
 
-    private static <E extends Throwable> void sneakyThrows(final Throwable t) throws E {
+    private static <E extends Throwable> void throwException0(final Throwable t) throws E {
         throw (E) t;
     }
+
+    private static <T> T uncheckedCast() {
+        return (T) DUMMY;
+    }
 }
Index: servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/UnreleasableUnsafeNoZeroingDirectByteBuf.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/UnreleasableUnsafeNoZeroingDirectByteBuf.java b/servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/UnreleasableUnsafeNoZeroingDirectByteBuf.java
--- a/servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/UnreleasableUnsafeNoZeroingDirectByteBuf.java	(revision a277cc5937f38e850170daf7cf722c310f391a14)
+++ b/servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/UnreleasableUnsafeNoZeroingDirectByteBuf.java	(date 1659587537300)
@@ -38,8 +38,8 @@
 import static io.servicetalk.utils.internal.PlatformDependent.freeMemory;
 import static io.servicetalk.utils.internal.PlatformDependent.newDirectBuffer;
 import static io.servicetalk.utils.internal.PlatformDependent.reserveMemory;
-import static io.servicetalk.utils.internal.PlatformDependent.throwException;
 import static io.servicetalk.utils.internal.PlatformDependent.unreserveMemory;
+import static io.servicetalk.utils.internal.ThrowableUtils.throwException;
 
 final class UnreleasableUnsafeNoZeroingDirectByteBuf extends UnreleasableUnsafeDirectByteBuf {
 
Index: servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/ThrowableUtils.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/ThrowableUtils.java b/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/ThrowableUtils.java
--- a/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/ThrowableUtils.java	(revision a277cc5937f38e850170daf7cf722c310f391a14)
+++ b/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/ThrowableUtils.java	(date 1659587537310)
@@ -26,6 +26,19 @@
         // No instances
     }
 
+    /**
+     * Raises an exception bypassing compiler checks for checked exceptions.
+     *
+     * @param t The {@link Throwable} to throw.
+     * @param <T> The expected type
+     * @return nothing actually will be returned from this method because it rethrows the specified exception. Making
+     * this method return an arbitrary type makes the caller method easier as they do not have to add a return statement
+     * after calling this method.
+     */
+    public static <T> T throwException(final Throwable t) {
+        return PlatformDependent0.throwException(t);
+    }
+
     /**
      * Combine two potential {@link Throwable}s into one.
      * If both parameters are {@link Throwable}, the {@code second} one will be
```

idelpivnitskiy (changes requested):

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
bondolo: My thought was to use it in `PlatformDependent` by having a static field that selected which implementation to use based on the availability of Unsafe. The sneakyThrows would be referenced using a method reference.

I didn't end up making that change though so this could be used as a reference handle. The JIT compiler is also reputedly better about optimizing/inlining `CallSite` functional interfaces than `MethodHandles`

idelpivnitskiy: Ok, thanks for explaining. Don't mind either way, was just curious.

bondolo: I did end up using this technique in the updated PR.

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
bondolo: Thank you for these improvements, this is turning in to nice cleanup.

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
```diff
@@ -142,37 +146,56 @@ final class PlatformDependent0 {
...
-            } catch (Throwable all) {
-                LOGGER.warn("failed computing THROW_EXCEPTION", all);
+                //noinspection unchecked
+                Consumer<Throwable> unsafeThrowConsumer =
+                        (Consumer<Throwable>) throwExceptionCallSite.getTarget().bindTo(unsafe).invoke();
+                throwConsumer = (t) -> {
+                    try {
+                        // JVM has been observed to crash when passing a null argument.
+                        // See https://github.com/netty/netty/issues/4131.
+                        unsafeThrowConsumer.accept(null != t ? t : new NullPointerException("Throwable was null"));
```
idelpivnitskiy, line 156: Can we use `Objects.requireNonNull` here?

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
```diff
@@ -142,37 +146,56 @@ final class PlatformDependent0 {
...
+                    } catch (Throwable all) {
+                        throwException0(t);
```
idelpivnitskiy, line 158: This is likely a problem. `unsafeThrowConsumer.accept(...)` always throws. Then we will catch it and rethrow again without unsafe. No benefit of using unsafe then.

bondolo: Hah, yes… What do you think about removing the catch?

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
bondolo: The intention was to have `unsafeThrowConsumer` throw rather the fall-back sneaky throws.

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
idelpivnitskiy: yes, we have to remove the `try-catch` block

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
idelpivnitskiy: Looks like after removing the catch block we can start using `Objects.requireNonNull`

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
bondolo: done.

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent0.java
bondolo: OK, I moved it to `throwException` so that it would be consistent on both paths.

idelpivnitskiy (approved):

PR #2304, opened 2022-08-02
Dependencies direct & jaxb-impl `api` declaration

bondolo:
Both change sets will be merged with rebase-and-merge

### DIFF
--- a/servicetalk-dependencies/build.gradle
+++ b/servicetalk-dependencies/build.gradle
@@ -17,14 +17,24 @@
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-core"
 apply plugin: "java-platform"
 
-description = "ServiceTalk BOM that includes direct dependencies"
+description = "ServiceTalk BOM that includes all modules and direct dependencies"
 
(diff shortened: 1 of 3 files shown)


servicetalk-dependencies/build.gradle
```diff
@@ -23,8 +23,17 @@ javaPlatform {
   allowDependencies()
 }
 
+// Include modules directly rather than by defining servicetalk bom as
+// a platform to work around [MNG-6772](https://issues.apache.org/jira/browse/MNG-6772).
+rootProject.subprojects.findAll { !it.name.endsWith("-bom") &&
```
idelpivnitskiy, line 28: Consider adding a comment that this part have to be consistent with what we have in `servicetalk-bom` and vice versa.

idelpivnitskiy (approved):
Make sure that the comment is added on the first commit. Thanks for the fix!

servicetalk-dependencies/build.gradle
bondolo: Comment added in both files.

PR #2307, opened 2022-08-04
Share bom inclusion closure to reduce code duplication

Scottmitch:


bondolo (approved):
I tried to see if there was a way to reference the `bomInclusionClosure` as a method reference but could not get the groovy syntax right.

PR #2309, opened 2022-08-05
gRPC add default method for asBlockingClient to simplify filter creation

Scottmitch:
Motivation:
gRPC filters were previously removed due to infrequent usage, large amount
of generated code, and users ability to manually wrap to provide the same
capabilities. However on the client side wrapping requires converting
from async to blocking which pushes common boilerplate code onto users.

Modifications:
- Use the existing AsyncToBlocking client generated code (which isn't
  used) as the default conversion from the async API.

Result:
Users can create an async wrapper for the client, and don't need
to worry about the blocking API conversion.


tkountis (approved):

idelpivnitskiy (approved):
🚀

idelpivnitskiy:
I will pull this in to rebase my coming PR. Thanks!

PR #2313, opened 2022-08-09
gRPC Trailers-Only response doesn't mark a streaming request as finished

idelpivnitskiy:
Motivation:

If server returns a `Trailers-Only` response for a streaming request, we
never drain the response message-body, because `concat` never subscribes
to the next source if the first source fails. As the result, concurrency
controller won't mark the request as finished. If the request publisher
never completes we will start leaking h2 connections because streams
won’t be closed.

Modifications:

- Enhance `TrailersOnlyErrorTest` to cover blocking client and make sure
the HTTP response terminates;
- In case of an error in `Trailers-Only` response, subscribe and cancel
the response message body to abort the request;
- Always check that `Trailers-Only` response doesn't have any further
frames on the wire;

Result:

gRPC response correctly terminates even if client receives a
`Trailers-Only` response.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
@@ -137,7 +137,7 @@ public abstract class Completable {
      *         }
      *     }
      * }</pre>
-     * @param predicate returns {@code true} if the {@link Throwable} should be transformed to and
+     * @param predicate returns {@code true} if the {@link Throwable} should be transformed to
      * {@link Subscriber#onComplete()} signal. Returns {@code false} to propagate the error.
      * @return A {@link Completable} which transform errors emitted on this {@link Completable} which match
      * {@code predicate} into a {@link Subscriber#onComplete()} signal (e.g. swallows the error).
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
@@ -137,7 +137,7 @@ public abstract class Completable {
      *         }
      *     }
      * }</pre>
-     * @param predicate returns {@code true} if the {@link Throwable} should be transformed to
+     * @param predicate returns {@code true} if the {@link Throwable} should be transformed to an
      * {@link Subscriber#onComplete()} signal. Returns {@code false} to propagate the error.
      * @return A {@link Completable} which transform errors emitted on this {@link Completable} which match
      * {@code predicate} into a {@link Subscriber#onComplete()} signal (e.g. swallows the error).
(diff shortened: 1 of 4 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Completable.java
```diff
@@ -137,7 +137,7 @@ public final <E extends Throwable> Completable onErrorComplete(Class<E> type) {
      *         }
      *     }
      * }</pre>
-     * @param predicate returns {@code true} if the {@link Throwable} should be transformed to and
+     * @param predicate returns {@code true} if the {@link Throwable} should be transformed to
```
tkountis, line 140: to "an" onComplete signal?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -312,13 +316,35 @@ static <Resp> Publisher<Resp> validateResponseAndGetPayload(final StreamingHttpR
...
         if (grpcStatusCode != null) {
+            // Drain the response messageBody to make sure concurrency controller marks the request as finished.
+            // In case the grpc-status is received in headers, we expect an empty messageBody, draining should not see
+            // any other frames. However, the messageBody won't complete until after the request stream completes too.
+            final Completable drainResponse = response.messageBody().beforeOnNext(frame -> {
+                throw new GrpcStatus(INTERNAL, null, "Violation of the protocol: received unexpected " +
+                        (frame instanceof HttpHeaders ? "Trailers" : "Data") +
+                        "frame after already receiving Trailers-Only response with grpc-status: " +
+                        grpcStatusCode.value() + '(' + grpcStatusCode + ')').asException();
+            }).ignoreElements();
```
tkountis, line 327: consider adding some more context in the error msg, eg. mentioning that it happened when trailers-only were received.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -312,13 +316,35 @@ static <Resp> Publisher<Resp> validateResponseAndGetPayload(final StreamingHttpR
...
+                // see the exception if the request publisher never terminates, or they may see a TimeoutException that
+                // will hide the original exception. Therefore, we have to return an error asap and then immediately
+                // subscribe & cancel the drainResponse. Cancellation is necessary to prevent sending large request
+                // payloads over network when server returns an error.
+                return Publisher.<Resp>failed(grpcStatusException).afterOnError(__ -> {
+                    // Because we subscribe asynchronously, users won't receive any further errors from drainResponse.
+                    // Instead, we log those errors for visibility. Use onErrorComplete instead of whenOnError to avoid
+                    // logging the same exception twice inside SimpleCompletableSubscriber.
+                    drainResponse.onErrorComplete(t -> {
+                        LOGGER.error("Unexpected error", t);
```
tkountis, line 340: consider adding some more context in the error msg, eg. mentioning that it happened when trailers-only were received.

tkountis:
excellent find 🎸

tkountis (approved):

PR #2317, opened 2022-08-10
Null check result of `HttpServerBuilder` supplier

bondolo:
Motivation:
`DefaultGrpcServerBuilder` invokes a provided Supplier without checking
the result for null. A `NullPointerException` will occur, but will not
include context about what was null.
Modifications:
Add `Objects.requireNonNull` check when invoking `Supplier`
Result:
Better diagnostic for failed Supplier.

### DIFF
--- a/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
+++ b/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
@@ -95,7 +95,7 @@ final class DefaultGrpcServerBuilder implements GrpcServerBuilder, ServerBinder 
     // Do not use this ctor directly, GrpcServers is the entry point for creating a new builder.
     DefaultGrpcServerBuilder(final Supplier<HttpServerBuilder> httpServerBuilderSupplier) {
         this.httpServerBuilderSupplier = () ->
-                Objects.requireNonNull(httpServerBuilderSupplier.get(), "supplier result was null")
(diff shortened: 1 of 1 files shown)


servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
```diff
@@ -93,8 +94,9 @@ final class DefaultGrpcServerBuilder implements GrpcServerBuilder, ServerBinder
 
     // Do not use this ctor directly, GrpcServers is the entry point for creating a new builder.
     DefaultGrpcServerBuilder(final Supplier<HttpServerBuilder> httpServerBuilderSupplier) {
-        this.httpServerBuilderSupplier = () -> httpServerBuilderSupplier.get()
-                .protocols(h2Default()).allowDropRequestTrailers(true);
+        this.httpServerBuilderSupplier = () ->
+                Objects.requireNonNull(httpServerBuilderSupplier.get(), "supplier result was null")
```
tkountis, line 98: consider static import

tkountis (approved):
lgtm

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
idelpivnitskiy: consider including in the msg that we expected an HttpServerBuilder, something like:
`Supplier<HttpServerBuilder> result was null` or `Supplier result was null instead of a valid HttpServerBuilder instance`

Issue #2298, opened 2022-07-29
GrpcLifecycleObserverTest


Scottmitch:
```
GrpcLifecycleObserverTest.testBiDiStream(boolean) [2] error=true

org.mockito.exceptions.verification.NoInteractionsWanted: 
No interactions wanted here:
-> at io.servicetalk.grpc.netty.GrpcLifecycleObserverTest.verifyObservers(GrpcLifecycleObserverTest.java:271)
But found this interaction on mock 'clientRequestObserver':
-> at io.servicetalk.grpc.utils.BiGrpcLifecycleObserver$BiGrpcRequestObserver.onRequestDataRequested(BiGrpcLifecycleObserver.java:134)
***
For your reference, here is the list of all invocations ([?] - means unverified).
1. -> at io.servicetalk.grpc.utils.BiGrpcLifecycleObserver$BiGrpcRequestObserver.onRequestDataRequested(BiGrpcLifecycleObserver.java:134)
2. -> at io.servicetalk.grpc.utils.BiGrpcLifecycleObserver$BiGrpcRequestObserver.onRequestData(BiGrpcLifecycleObserver.java:143)
3. -> at io.servicetalk.grpc.utils.BiGrpcLifecycleObserver$BiGrpcRequestObserver.onRequestComplete(BiGrpcLifecycleObserver.java:161)
4. [?]-> at io.servicetalk.grpc.utils.BiGrpcLifecycleObserver$BiGrpcRequestObserver.onRequestDataRequested(BiGrpcLifecycleObserver.java:134)

	at app//io.servicetalk.grpc.netty.GrpcLifecycleObserverTest.verifyObservers(GrpcLifecycleObserverTest.java:271)
	at app//io.servicetalk.grpc.netty.GrpcLifecycleObserverTest.runTest(GrpcLifecycleObserverTest.java:219)
	at app//io.servicetalk.grpc.netty.GrpcLifecycleObserverTest.testBiDiStream(GrpcLifecycleObserverTest.java:205)
	
```

tkountis:
```
GrpcLifecycleObserverTest.testBiDiStream(boolean) [1] error=false

org.mockito.exceptions.verification.NoInteractionsWanted: 
No interactions wanted here:
-> at io.servicetalk.grpc.netty.GrpcLifecycleObserverTest.verifyObservers(GrpcLifecycleObserverTest.java:271)
But found this interaction on mock 'clientRequestObserver':
[...]

PR #2318, opened 2022-08-10
Throw error when score is invoked on a connection that doesn't support it.

tkountis:
Motivation:

Different load-balancer implementations can support connection scores or not.
Due to common interfaces it would be easy for a user to mix scoring protocol agnostic load-balancer with a non scoring
protocol specific wrapper, without knowing that the configuration was wrong, either in built or run times.

Modifications:

Let default implementations of `score()` throw an error instead, thus making visible the misuse during runtime.

Result:

Better misconfiguration visibility

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
@@ -119,8 +119,11 @@ public interface HttpLoadBalancerFactory<ResolvedAddress>
                 final FilterableStreamingHttpConnection delegate,
                 final ReservableRequestConcurrencyController controller) {
             this(delegate, controller, () -> {
-                throw new UnsupportedOperationException("This type of connection doesn't support scoring. " +
-                        "Connection scoring is only available through scoring supported load balancers.");
+                throw new UnsupportedOperationException(
+                        DefaultFilterableStreamingHttpLoadBalancedConnection.class.getName() +
+                                " doesn't support scoring. " + ScoreSupplier.class.getName() +
+                                " is only available through " + HttpLoadBalancerFactory.class.getSimpleName() +
+                                " implementations that support scoring.");
             });
         }
 
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
@@ -119,7 +119,10 @@ public interface HttpLoadBalancerFactory<ResolvedAddress>
         public DefaultFilterableStreamingHttpLoadBalancedConnection(
                 final FilterableStreamingHttpConnection delegate,
                 final ReservableRequestConcurrencyController controller) {
-            this(delegate, controller, () -> MAX_VALUE);
+            this(delegate, controller, () -> {
+                throw new UnsupportedOperationException("This type of connection doesn't support scoring. " +
+                        "Connection scoring is only available through scoring supported load balancers.");
+            });
         }
 
         /**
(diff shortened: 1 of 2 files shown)


tkountis:
Failure due to https://github.com/apple/servicetalk/issues/2298

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
```diff
@@ -119,7 +118,10 @@ default HttpExecutionStrategy requiredOffloads() {
         public DefaultFilterableStreamingHttpLoadBalancedConnection(
                 final FilterableStreamingHttpConnection delegate,
                 final ReservableRequestConcurrencyController controller) {
-            this(delegate, controller, () -> MAX_VALUE);
+            this(delegate, controller, () -> {
+                throw new UnsupportedOperationException("This type of connection doesn't support scoring. " +
```
idelpivnitskiy, line 122: It will be useful to know the type of "this connection", consider:
```patch
Index: servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java	(revision f076f7e355de22f777c7c0d304b5c5c2475eee00)
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java	(date 1660162709813)
@@ -119,8 +119,11 @@
                 final FilterableStreamingHttpConnection delegate,
                 final ReservableRequestConcurrencyController controller) {
             this(delegate, controller, () -> {
-                throw new UnsupportedOperationException("This type of connection doesn't support scoring. " +
-                        "Connection scoring is only available through scoring supported load balancers.");
+                throw new UnsupportedOperationException(
+                        DefaultFilterableStreamingHttpLoadBalancedConnection.class.getName() +
+                                " doesn't support scoring. " + ScoreSupplier.class.getName() +
+                                " is only available through " + HttpLoadBalancerFactory.class.getSimpleName() +
+                                " implementations that support scoring.");
             });
         }
 
Index: servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java	(revision f076f7e355de22f777c7c0d304b5c5c2475eee00)
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java	(date 1660162856713)
@@ -18,6 +18,7 @@
 import io.servicetalk.client.api.ConnectionFactory;
 import io.servicetalk.client.api.LoadBalancer;
 import io.servicetalk.client.api.LoadBalancerFactory;
+import io.servicetalk.client.api.ScoreSupplier;
 import io.servicetalk.client.api.ServiceDiscovererEvent;
 import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.Publisher;
@@ -141,8 +142,11 @@
 
         @Override
         public int score() {
-            throw new UnsupportedOperationException("This type of connection doesn't support scoring. " +
-                    "Connection scoring is only available through scoring supported load balancers.");
+            throw new UnsupportedOperationException(
+                    DefaultFilterableStreamingHttpLoadBalancedConnection.class.getName() +
+                            " doesn't support scoring. " + ScoreSupplier.class.getName() +
+                            " is only available through " + HttpLoadBalancerFactory.class.getSimpleName() +
+                            " implementations that support scoring.");
         }
 
         @Override

```

idelpivnitskiy (approved):
The change makes sense, consider adding more information in the exception message.

PR #2319, opened 2022-08-10
Give access to the underlying TCP ConnectionContext from HTTP/2 service

idelpivnitskiy:
Motivation:

Server-side users have access only to HTTP/2 stream `ConnectionContext`
from the `HttpServiceContext`. In some cases, it's important to access
a `ConnectionContext` of the underlying TCP connection.

Modifications:

- Add `@Nullable ConnectionContext#parent()` method;
- Implement support for a new method;
- Add tests;

Result:

Users can access the underlying TCP `ConnectionContext` from the HTTP/2
service endpoint.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpConnectionContext.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpConnectionContext.java
@@ -27,4 +27,9 @@ public interface HttpConnectionContext extends ConnectionContext {
 
     @Override
     HttpProtocolVersion protocol();
+
+    // Do not override the return type of parent() method to HttpConnectionContext as there is a possibility to have an
+    // HttpConnectionContext over a non-HTTP ConnectionContext. Also, with HttpConnectionContext as a return type it
+    // makes implementation a lot harder because we receive a non-HTTP NettyConnectionContext from the transport and
+    // wrap it later.
 }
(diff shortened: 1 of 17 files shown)


idelpivnitskiy:
Still need to add tests but the main change is ready for review.

bondolo:
I am curious if there are cases which _should_ override return type of `parent()` to `HttpConnectionContext`. I'll be looking for cases where the type of the result is `instanceof` checked or down casted. I can understand the comment why the return type is not changed in `HttpConnectionContext` itself, but am curious if there are any places where overriding is appropriate.

idelpivnitskiy:
I'm not aware of those cases.

idelpivnitskiy:
The only need for `HttpConnectionContext` if users want access to the `HttpExecutionContext` or `HttpProtocolVersion`. However, those are already available to users from the current ctx, they don't need to access `ctx.parent()` for that. If, for some reason, they really need it (let's say to pass somewhere else), they can either do an `instaceof` check or wrap the state of `parent` and the current `ctx` with `HttpConnectionContext` themselves.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpConnectionContext.java
```diff
@@ -27,4 +27,9 @@ public interface HttpConnectionContext extends ConnectionContext {
 
     @Override
     HttpProtocolVersion protocol();
+
+    // Do not override the return type of parent() method to HttpConnectionContext as there is a possibility to have an
+    // HttpConnectionContext over a non-HTTP ConnectionContext. Also, with HttpConnectionContext as a return type it
+    // makes implementation a lot harder because we receive a non-HTTP NettyConnectionContext from the transport and
+    // wrap it later.
```
idelpivnitskiy, line 34: Note that we can make this change only now or never, as even adding a `default` impl will be a breaking change. Users who already implement `parent()` returning `ConnectionContext` type will be broken if we change it to `HttpConnectionContext` later.

idelpivnitskiy: However, we can always add something like `HttpConnectionContext  httpParent()` later, if there is a use-case for that.

bondolo: The covariant return override should only be used if it can be known for certain that the return type will be the override type. If there are no certain cases, that's OK.

tkountis (approved):

Issue #2321, opened 2022-08-12
Flaky test: PublisherFlatMapConcatSingleTest


idelpivnitskiy:
https://github.com/apple/servicetalk/pull/2311/checks?check_run_id=7811182599

```
PublisherFlatMapConcatSingleTest.errorPropagatedInOrderLast(int, int, int, boolean) [2] begin=0 end=25 maxConcurrency=10 delayError=false

java.lang.AssertionError: 
Expected: iterable containing ["0x", "1x", "2x", "3x", "4x", "5x", "6x", "7x", "8x", "9x", "10x", "11x", "12x", "13x", "14x", "15x", "16x", "17x", "18x", "19x", "20x", "21x", "22x", "23x"]
     but: item 20: was null
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at io.servicetalk.concurrent.api.PublisherFlatMapConcatSingleTest.errorPropagatedInOrderLast(PublisherFlatMapConcatSingleTest.java:84)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
```

PR #2320, opened 2022-08-11
Move abstract TCK classes to testFixtures to use in other modules

idelpivnitskiy:
Motivation:

We have a few internal operators for `liftSync` or `liftSyncToSingle` in
other modules that currently not tested with RS TCK.

Modifications:

- Move `Abstract*TckTest` classes to `testFixtures`, make their `public`
methods `protected`, add javadoc;
- Add missing `@Test` annotation for implementations of those abstract
tests;
- Use new `servicetalk-concurrent-reactivestreams-testFixtures` for
`SpliceFlatStreamToMetaSingle` and `BeforeFinallyHttpOperator`;
- Fix violation of 1.09 rule found in `SpliceFlatStreamToMetaSingle`;

Result:

Our internal operators can be tested with RS TCK.

PR #2323, opened 2022-08-12
Publisher#flatMapConcatSingle null value concurrency fix

Scottmitch:
Motivation:
Publisher#flatMapConcatSingle maintains a queue of elements
on top of Publisher#flatMapSingle to maintain ordering. Elements
within this queue have non-volatile state due to outer concurrency
protections, however there are two independent variables indicating
"terminal type" and "terminal value". This may result in reading
the "terminal type" the "terminal value" may not be visible yet
and return `null`.

Modifications:
- Collapse the 2 variables into a single variable, so when reading
  if it is non-null we have the "type" and "value" visible.

Result:
Fixes https://github.com/apple/servicetalk/issues/2321

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatUtils.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatUtils.java
@@ -134,6 +134,13 @@ final class PublisherFlatMapConcatUtils {
     private static final class Item<R> {
         @Nullable
         SingleSource.Subscriber<? super R> subscriber;
+        /**
+         * Visibility is provided by {@link OrderedMapper#consumerLockUpdater}. There are multiple producer threads
+         * modifying independent {@link Item}s, but only a single thread consumes and calls {@link #tryTerminate()}.
+         * Since the state is a single reference either the consumer thread sees the state, or it doesn't. If it does
+         * then the item can be consumed and we are done. If it doesn't then the
+         * {@link OrderedMapper#consumerLockUpdater} requires another lock attempt and the state becomes visible.
+         */
         @Nullable
         private Object result;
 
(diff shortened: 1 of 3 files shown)


bondolo:
If the incomplete update can be observed then it must not be happening under the outer concurrency protection. Can multiple threads be racing to update the publisher or only one other thread observing the update? I am wondering if separate concurrency is needed on the publisher. Reducing to one variable will help, but I am uncertain that there still can't be update races--mostly null check followed by update rather than some flavor of compareAndSet`(null, value)`.

Scottmitch:
The producer/consumer may happen on different threads. The consumer is single threaded (via lock) but the producer could be happening on another thread. If the producer is on another thread then the consumer may see partial state (if it is spread out over multiple variables). We can run through this offline as well to confirm assumptions.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatUtils.java
```diff
@@ -134,31 +136,27 @@ private void tryPollQueue() {
         SingleSource.Subscriber<? super R> subscriber;
         @Nullable
         private Object result;
```
idelpivnitskiy, line 138: Consider adding a comment to clarify why it doesn't have to be `volatile`

idelpivnitskiy (approved):
Thanks for the quick fix!
While you are in RS-mode, please have a look at https://github.com/apple/servicetalk/pull/2320#discussion_r943757428 too.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapConcatUtils.java
tkountis: Yes I also think that this needs to be `volatile`. If not, what guarantees visibility from the producer to the consumer?

Scottmitch:
discussed offline with @tkountis @bondolo @idelpivnitskiy . everyone is satisfied with the current approach and I added comments to clarify.

PR #2324, opened 2022-08-13
Protobuf2 compatibility

bondolo:


servicetalk-data-protobuf/src/main/java/io/servicetalk/data/protobuf/ProtobufSerializerFactory.java
```diff
@@ -109,8 +112,20 @@ private <T extends MessageLite> Parser<T> newParser(Class<T> clazz) {
         final MethodHandle mh;
         try {
             mh = MethodHandles.publicLookup().findStatic(clazz, PARSER_METHOD_NAME, PARSER_METHOD_TYPE);
-        } catch (IllegalAccessException | NoSuchMethodException e) {
+        } catch (IllegalAccessException e) {
             throw new IllegalArgumentException("Unable to find " + clazz + "." + PARSER_METHOD_NAME, e);
+        } catch (NoSuchMethodException nsme) {
+            // perhaps the class is a protobuf 2.x generated class
+            try {
```
Scottmitch, line 119: can we add this code to an independent method so it in the expected case (protbuf 3) the byte code for this method is smaller?

servicetalk-data-protobuf/src/test/java/io/servicetalk/data/protobuf/ProtobufSerializerFactoryTest.java
```diff
@@ -143,4 +149,114 @@ private static DummyMessage newMsg(int length) {
...
+
+        static final Protobuf2Message MESSAGE = new Protobuf2Message() {
+            @Override
+            public boolean isInitialized() {
+                return true;
+            }
+        };
+
+        /**
+         * This declaration is intended to be compatible with how protoc 2.5 generated message classes.
```
Scottmitch, line 174: to test compatibility we should create a new project that brings in protobuf 2.x as a test dependency, generates some code, and verify the serializer works. there maybe other things that we change/depend upon over time. I also don't think we want to commit/guarantee protnbuf 2 compatibility externally ([protobuf 3 was released in 2016](https://github.com/protocolbuffers/protobuf/releases/tag/v3.0.0), last [protobuf 2 release was in 2014](https://github.com/protocolbuffers/protobuf/releases/tag/v2.6.1)), and if we have to advertise anything it would be "best effort" as this may restrict our ability to move forward with new features, etc.

Scottmitch:
lets close this for now as we are encouraging folks to update to protobuf 3

PR #2327, opened 2022-08-18
BeforeFinallyHttpOperator dereference subscriber after termination

Scottmitch:
Motivation:
After terminating the Single subscriber we need to dereference the Single subscriber as otherwise
we may retain a reference to a downstream Subscriber and violate RS 3.13.
https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.4/README.md#3.13


Scottmitch:
test for this included in https://github.com/apple/servicetalk/pull/2320

tkountis (approved):

idelpivnitskiy (commented):
Thanks!

PR #2328, opened 2022-08-18
Add io_uring dependencies with classifiers in `servicetalk-dependencies`

idelpivnitskiy:
Motivation:

Gradle intentionally skips dependencies constraints with a classifier
when it generates Maven BOM file.
See https://github.com/gradle/gradle/issues/8561.
It negatively affects maven users bcz maven marks
`servicetalk-transport-netty-internal` POM file as invalid:

```
[WARNING] The POM for io.servicetalk:servicetalk-transport-netty-internal:jar:0.42.14 is invalid, transitive dependencies (if any) will not be available: 2 problems were encountered while building the effective model for io.servicetalk:servicetalk-transport-netty-internal:0.42.14
[ERROR] 'dependencies.dependency.version' for io.netty.incubator:netty-incubator-transport-native-io_uring:jar:linux-x86_64 is missing. @
[ERROR] 'dependencies.dependency.version' for io.netty.incubator:netty-incubator-transport-native-io_uring:jar:linux-aarch_64 is missing. @
```

Modifications:
- Use `withXml` to add entries with a classifier explicitly;

Result:

Maven users don't have build issues anymore.

### DIFF
--- a/servicetalk-dependencies/build.gradle
+++ b/servicetalk-dependencies/build.gradle
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-core"
 apply plugin: "java-platform"
 
@@ -40,8 +39,6 @@ dependencies {
 
     // Ideally these come from the Netty BOM
     api "io.netty.incubator:netty-incubator-transport-native-io_uring:$nettyIoUringVersion"
-    api "io.netty.incubator:netty-incubator-transport-native-io_uring:$nettyIoUringVersion:linux-x86_64"
-    api "io.netty.incubator:netty-incubator-transport-native-io_uring:$nettyIoUringVersion:linux-aarch_64"
 
     api "io.opentracing:opentracing-api:$openTracingVersion"
     api "io.zipkin.reporter2:zipkin-reporter:$zipkinReporterVersion"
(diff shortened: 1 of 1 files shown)


tkountis (approved):

idelpivnitskiy:
I've verified locally that the new published snapshot contains 2 new entries:
```
      <dependency>
        <groupId>io.netty.incubator</groupId>
        <artifactId>netty-incubator-transport-native-io_uring</artifactId>
        <version>0.0.14.Final</version>
        <classifier>linux-x86_64</classifier>
      </dependency>
      <dependency>
        <groupId>io.netty.incubator</groupId>
        <artifactId>netty-incubator-transport-native-io_uring</artifactId>
        <version>0.0.14.Final</version>
        <classifier>linux-aarch_64</classifier>
      </dependency>
```
and the sample project I used to reproduce a maven warning doesn't complain anymore with the new snapshot.

Scottmitch (approved):

servicetalk-dependencies/build.gradle
```diff
@@ -96,6 +93,16 @@ publishing {
           system = 'ServiceTalk CI'
           url = "${ciManagementUrl}"
         }
+        // Gradle ignores dependencies constraints with classifiers when it generates pom.xml. To workaround it, we are
+        // forced to use withXml. See https://github.com/gradle/gradle/issues/8561
+        withXml {
+          Node pomNode = asNode()
+          Node dependencies = pomNode.dependencyManagement.dependencies.get(0)
```
bondolo, line 100: Is there any possibility that this node won't already exist and needs to be created?

servicetalk-dependencies/build.gradle
```diff
@@ -125,3 +132,11 @@ if (!!findProperty("signingKey") && !!findProperty("signingPassword")) {
     sign publishing.publications.mavenJava
   }
 }
+
+def addDependency(Node dependencies, String groupId, String artifactId, String version, String classifier) {
```
bondolo, line 136: Perhaps make `version` and `classier` optional and move to plugin utils?

servicetalk-dependencies/build.gradle
idelpivnitskiy: No, `withXml` part is invoked after the pom.xml is generated. It always has some `dependencies`.

servicetalk-dependencies/build.gradle
idelpivnitskiy: This method is required for version with classifier, they are not optional.
We don't need to share it with any other `gradle` file bcz it's needed only for `servicetalk-dependencies`. If we have a use-case later, we can move it then.

PR #2329, opened 2022-08-22
Improve parsing of Set-Cookie headers

chrisvest:
Motivation:
The parsing of Set-Cookie headers is deviating from the standard in a couple of ways.
- Cookie values are allowed to contain the equals sign character.
- Percentage-prefixed hexadecimal literals are not a thing.
- Only cookie-values need to adhere to the cookie-octet rule.
- Defined but empty values should decode to the empty string rather than null, so we can distinguish between their empty and absent cases.
- When parsing Set-Cookie values, every ; must be followed by a space.
  See the `set-cookie-string` rule in https://www.rfc-editor.org/rfc/rfc6265#section-4.1.1
  The parser should check for this, since it can lead to mis-parsing later data and produce confusing error messages, or the produce a cookie object with wrong data.

Modification:
When the parser decodes the '=' character, it now no longer updates the 'begin' index if the parse state is ParsingValue.
This means the parsed '=' character will be part of the cookie value.

The parse rule for the '%' character has been removed.
The associated validation code has also been removed, as it is no longer used.
This means the '%' character no longer has a special meaning to the parser.

When the parser validates a non-special character (a character that falls into the default case in the parser), the cookie-octet validation function is now only used if the parser is in the ParsingValue state.
Otherwise, we use a new cookie-attribute validation.
Most cookie-attributes need only follow the simple rule of "any CHAR except CTLs or ';'".

The parser already has a final-processing step for non-empty tails.
We add a similar step for when the tail is the empty string.
This way, cookie values, paths, and domains, that are defined but have no value, will be decoded to have the empty string as a value.
This way, we can tell if those attributes were defined but empty, or absent. 

The parser for Set-Cookie HTTP headers now checks that ';' are followed by a space byte.
When cookie validation is enabled, an exception will be thrown when this is not the case.
When cookie validation is disabled, the space byte will no longer be blindly assumed to be there, but instead checked for, and the parser position will be adjusted accordingly.
This allows, e.g. "Set-Cookie: a=b;Expires=Mon, 22 Aug 2022 20:12:35 GMT" to be parsed correctly in non-validating mode.
Previously, the above would have thrown an exception for using a ',' character in an attribute value that do not allow them ("xpires" instead of "Expires").

The parse-state-switching algorithm has been changed.
It no longer relies on a map from attribute name to a parse-state-as-a-char-sequence.
Instead, we do a hard-coded binary search based on the attribute length (which we can read without paying for a bounds-check), and then we compare the field name with what would be expected for that length.
This gives us a significant boost to parsing speed, because we do no hashing, fewer string-compares, fewer memory indirections, and we let the CPU branch predictor do the heavy lifting.
It also takes up less memory as we no longer need the map of AV field names.
We can also remove the awkward ParseStateCharSequence class.

Improve the error message for when a quoted cookie value contains the ';' character, which is not allowed.
This also means we won't run into later errors about unbalanced quotation or the like.

Result:
The parsing of Set-Cookie headers is now more correct, and faster, and has better error reporting.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
@@ -15,15 +15,11 @@
  */
 package io.servicetalk.http.api;
 
-import io.servicetalk.buffer.api.CharSequences;
-
 import javax.annotation.Nullable;
 
 import static io.servicetalk.buffer.api.CharSequences.caseInsensitiveHashCode;
 import static io.servicetalk.buffer.api.CharSequences.contentEquals;
 import static io.servicetalk.buffer.api.CharSequences.contentEqualsIgnoreCase;
-import static io.servicetalk.buffer.api.CharSequences.equalsIgnoreCaseLower;
-import static io.servicetalk.buffer.api.CharSequences.newAsciiString;
 import static io.servicetalk.buffer.api.CharSequences.parseLong;
 import static io.servicetalk.http.api.HeaderUtils.validateCookieNameAndValue;
 import static io.servicetalk.http.api.HeaderUtils.validateToken;
@@ -43,23 +39,36 @@ public final class DefaultHttpSetCookie implements HttpSetCookie {
     private static final String ENCODED_LABEL_SECURE = "; secure";
     private static final String ENCODED_LABEL_SAMESITE = "; samesite=";
 
-    /**
-     * An underlying size of 16 has been shown with the current {@link CharSequences#newAsciiString(CharSequence)} hash
-     * algorithm to have no collisions with the current set of supported cookie names.
-     * If more cookie names are supported, or the hash algorithm changes this initial value should be re-evaluated.
-     * <p>
-     * We use {@link HttpHeaders} here because we need a case-insensitive way to compare multiple keys. Instead of a
-     * if/else block we lean on {@link HttpHeaders} which provides an associative array that compares keys in a case
-     * case-insensitive manner.
-     */
-    private static final HttpHeaders AV_FIELD_NAMES = new DefaultHttpHeaders(16, false, false, false);
-
-    static {
-        AV_FIELD_NAMES.add(newAsciiString("path"), new ParseStateCharSequence(ParseState.ParsingPath));
-        AV_FIELD_NAMES.add(newAsciiString("domain"), new ParseStateCharSequence(ParseState.ParsingDomain));
-        AV_FIELD_NAMES.add(newAsciiString("expires"), new ParseStateCharSequence(ParseState.ParsingExpires));
-        AV_FIELD_NAMES.add(newAsciiString("max-age"), new ParseStateCharSequence(ParseState.ParsingMaxAge));
-        AV_FIELD_NAMES.add(newAsciiString("samesite"), new ParseStateCharSequence(ParseState.ParsingSameSite));
+    private static ParseState parseStateOf(CharSequence fieldName) {
+        // Try a binary search based on length. We can read length without bounds checks.
+        int len = fieldName.length();
+        if (len >= 4 && len <= 8) {
+            if (len < 7) {
+                if (len == 4) {
+                    if (contentEqualsIgnoreCase("path", fieldName)) {
+                        return ParseState.ParsingPath;
+                    }
+                } else if (len == 6) {
+                    if (contentEqualsIgnoreCase("domain", fieldName)) {
+                        return ParseState.ParsingDomain;
+                    }
+                }
+            } else {
+                if (len == 7) {
+                    if (contentEqualsIgnoreCase("expires", fieldName)) {
+                        return ParseState.ParsingExpires;
+                    }
+                    if (contentEqualsIgnoreCase("max-age", fieldName)) {
+                        return ParseState.ParsingMaxAge;
+                    }
+                } else {
+                    if (contentEqualsIgnoreCase("samesite", fieldName)) {
+                        return ParseState.ParsingSameSite;
+                    }
+                }
+            }
+        }
+        return ParseState.Unknown;
     }
 
     private final CharSequence name;
@@ -98,7 +107,7 @@ public final class DefaultHttpSetCookie implements HttpSetCookie {
      * <a href="https://tools.ietf.org/html/rfc6265#section-4.1.1">cookie-value</a>.
      * @param secure the <a href="https://tools.ietf.org/html/rfc6265#section-4.1.1">secure-av</a>.
      * @param httpOnly the <a href="https://tools.ietf.org/html/rfc6265#section-4.1.1">httponly-av</a> (see
-     * <a href="http://www.owasp.org/index.php/HTTPOnly">HTTP-only</a>).
+     * <a href="https://www.owasp.org/index.php/HTTPOnly">HTTP-only</a>).
      */
     public DefaultHttpSetCookie(final CharSequence name, final CharSequence value,
                                 final boolean wrapped, final boolean secure, final boolean httpOnly) {
@@ -120,7 +129,7 @@ public final class DefaultHttpSetCookie implements HttpSetCookie {
      * <a href="https://tools.ietf.org/html/rfc6265#section-4.1.1">cookie-value</a>.
      * @param secure the <a href="https://tools.ietf.org/html/rfc6265#section-4.1.1">secure-av</a>.
      * @param httpOnly the <a href="https://tools.ietf.org/html/rfc6265#section-4.1.1">httponly-av</a> (see
-     * <a href="http://www.owasp.org/index.php/HTTPOnly">HTTP-only</a>).
+     * <a href="https://www.owasp.org/index.php/HTTPOnly">HTTP-only</a>).
      * @param sameSite the
      * <a href="https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05#section-5.3.7">SameSite attribute</a>.
      */
@@ -162,7 +171,8 @@ public final class DefaultHttpSetCookie implements HttpSetCookie {
             begin = 0;
         }
 
-        while (i < setCookieString.length()) {
+        int length = setCookieString.length();
+        while (i < length) {
             final char c = setCookieString.charAt(i);
             switch (c) {
                 case '=':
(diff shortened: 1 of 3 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
```diff
@@ -248,7 +248,18 @@ static HttpSetCookie parseSetCookie(final CharSequence setCookieString, boolean
                             break;
                     }
                     parseState = ParseState.Unknown;
-                    i += 2;
+                    if (validateContent) {
+                        if (i >= setCookieString.length() || ' ' != setCookieString.charAt(i)) {
+                            throw new IllegalArgumentException(
+                                    "a space is required after ; in cookie attribute-value lists");
```
idelpivnitskiy, line 254: Multiple other tests fail with this exception. Example:
```
DefaultHttpSetCookiesTest > decodeDuplicateNamesRO() FAILED
    java.lang.IllegalArgumentException: a space is required after ; in cookie attribute-value lists
        at io.servicetalk.http.api.DefaultHttpSetCookie.parseSetCookie(DefaultHttpSetCookie.java:253)
        at io.servicetalk.http.api.HeaderUtils.parseSetCookie(HeaderUtils.java:364)
        at io.servicetalk.http.api.ReadOnlyHttpHeaders.getSetCookiesIterator(ReadOnlyHttpHeaders.java:295)
        at io.servicetalk.http.api.DefaultHttpSetCookiesTest.decodeDuplicateNames(DefaultHttpSetCookiesTest.java:60)
        at io.servicetalk.http.api.DefaultHttpSetCookiesTest.decodeDuplicateNamesRO(DefaultHttpSetCookiesTest.java:56)
```
See build logs for full info.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultHttpSetCookiesTest.java
```diff
@@ -769,6 +769,32 @@ private static void valueIteratorThrowsIfNoNextCall(HttpHeaders headers) {
         itr.remove();
     }
 
+    @Test
+    void throwIfNoSpaceBeforeCookieAttributeValue() {
+        final HttpHeaders headers = DefaultHttpHeadersFactory.INSTANCE.newHeaders();
+        headers.add("set-cookie", "first=12345;Extension");
+        headers.add("set-cookie", "second=12345;Expires=Mon, 22 Aug 2022 20:12:35 GMT");
+        throwIfNoSpaceBeforeCookieAttributeValue(headers);
```
idelpivnitskiy, line 777: Can you please add tests to verify "lenient" mode when validation is disabled?
`DefaultHttpHeadersFactory` has ctor args to control validation.

idelpivnitskiy (commented):
@chrisvest Thank you for finding and contributing!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
chrisvest: This was probably caused by my attempt to move the improvements over bit-by-bit. I have updated the PR to contain all of the improvements in one go.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultHttpSetCookiesTest.java
chrisvest: I added a test.

Scottmitch:
```
> Task :servicetalk-http-api:pmdMain FAILED
```

you can run `./gradlew pmd` locally to find issues.

chrisvest:
@idelpivnitskiy Build passed now.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
```diff
@@ -535,7 +535,7 @@ protected final void initNext(CharSequence cookieHeaderValue) {
          * @return the next {@link HttpCookiePair} value for {@link #next()}, or {@code null} if all have been parsed.
          */
         private HttpCookiePair findNext(CharSequence cookieHeaderValue) {
-            int semiIndex = indexOf(cookieHeaderValue, ';', nextNextStart);
+            int semiIndex = nextDelimiter(cookieHeaderValue);
```
idelpivnitskiy, line 538: Should we use the same approach instead of `indexOf` in all other cases in this class (lines 396, 442, 645)?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
```diff
@@ -519,46 +544,43 @@ private static SameSite fromSequence(CharSequence cs, int begin, int end) {
...
-        if (c2 != 'X' && c2 != 'x') {
-            throw new IllegalArgumentException("unexpected hex indicator " + c2);
-        }
-        c2 = cookieHeaderValue.charAt(i + 2);
-        final char c3 = cookieHeaderValue.charAt(i + 3);
-        // The MSB can only be 0,1,2 so we do a cheaper conversion of hex -> decimal.
-        final int hexValue = (c2 - '0') * 16 + hexToDecimal(c3);
-        validateCookieOctetHexValue(hexValue);
+    private static boolean equalsIgnoreCaseLower(char c, char k) {
+        return c == k || c >= 'A' && c <= 'Z' && c == k - 32;
```
idelpivnitskiy, line 548: Can we use pre-existing `CharSequences.equalsIgnoreCaseLower`? Looks like this method does the same

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
```diff
@@ -519,46 +544,43 @@ private static SameSite fromSequence(CharSequence cs, int begin, int end) {
...
+            throw unexpectedHexValue(hexValue, index);
         }
     }
 
-    private static int hexToDecimal(final char c) {
-        return c >= '0' && c <= '9' ? c - '0' : c >= 'a' && c <= 'f' ? (c - 'a') + 10 : c >= 'A' && c < 'F' ?
-                (c - 'A') + 10 : -1;
+    private static IllegalArgumentException unexpectedHexValue(int hexValue, int index) {
+        return new IllegalArgumentException(
+                "Unexpected hex value at index " + index + ": " + Integer.toHexString(hexValue));
```
idelpivnitskiy, line 584: Consider adding `0x` before the hexValue string

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
```diff
@@ -462,27 +480,34 @@ private enum ParseState {
...
+    private static ParseState parseStateOf(CharSequence fieldName) {
+        // Try a binary search based on length. We can read length without bounds checks.
+        int len = fieldName.length();
+        if (len >= 4 && len <= 8) {
+            if (len < 7) {
+                if (len == 4) {
+                    if (contentEqualsIgnoreCase("path", fieldName)) {
+                        return ParseState.ParsingPath;
+                    }
+                } else if (len == 6 && contentEqualsIgnoreCase("domain", fieldName)) {
```
idelpivnitskiy, line 492: Is checking `len` give visible advantage? Impl of `contentEqualsIgnoreCase` checks `a.length() != b.length()` before starting comparison of each character

servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultHttpSetCookiesTest.java
```diff
@@ -91,18 +91,18 @@ private static void decodeSecureCookieNames(final HttpHeaders headers) {
 
     @Test
     void decodeDifferentCookieNames() {
-        final HttpHeaders headers = new ReadOnlyHttpHeaders("set-cookie",
```
idelpivnitskiy, line 94: IIUC `ReadOnlyHttpHeaders` is a pkg-private class that is not used for anywhere except tests. Can you please remove it with corresponding tests?

idelpivnitskiy (commented):
Looks great, thanks for the nice improvements!

servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultHttpSetCookiesTest.java
Scottmitch: Lets consider removal of ReadOnlyHeaders as a followup PR if necessary. The intention was to potentially use it for some gRPC cases but headers now being exposed in filters makes that difficult todo.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
chrisvest: An earlier version used just `contentEqualsIgnoreCase`, and using the length for the branch tree like this was definitely faster. This `parseStateOf` method is some of the hottest code in the parser.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultHttpSetCookiesTest.java
chrisvest: Oh, didn't see @Scottmitch's comment in time. I already added a commit that removes it. Which way do you prefer?

chrisvest:
@idelpivnitskiy comments addressed

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
```diff
@@ -545,7 +547,7 @@ private static SameSite fromSequence(CharSequence cs, int begin, int end) {
     }
 
     private static boolean equalsIgnoreCaseLower(char c, char k) {
-        return c == k || c >= 'A' && c <= 'Z' && c == k - 32;
+        return CharSequences.equalsIgnoreCaseLower(c, k);
```
idelpivnitskiy, line 550: Consider removing the whole private method and just add a static import for `CharSequences.equalsIgnoreCaseLower`. It has the same name, same params.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/DefaultHttpSetCookiesTest.java
idelpivnitskiy: Let's move the last commit into a separate PR for visibility and easier revert if necessary.

idelpivnitskiy (approved):

idelpivnitskiy (approved):

chrisvest:
@idelpivnitskiy Done. Moved it to #2354

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/2298

PR #2331, opened 2022-08-29
Avoid race when HTTP/1.x request is cancelled at connection level filter

idelpivnitskiy:
Motivation:

#2266 (released in 0.42.13) introduces a race between cancel marking a
connection as "closing" (requires a hop to the event-loop), followed by
propagation of `onError` (on the cancel thread) that marks the request
as "finished", and the next request selecting the same connection. This
only affects rare users who cancel at the connection level, users who
use `TimeoutHttpRequesterFilter` as `appendClientFilter` are not
affected.

Modifications:
- Move `onClosing` to `NettyChannelListenableAsyncCloseable`;
- Notify `onClosing` asap for `channelInactive` event;
- Log a warning if `AbstractLBHttpConnectionFactory` can not access
`onClosing` (users incorrectly wrapped a connection);
- Log a warning if `LoadBalancedStreamingHttpClient` takes ownership of
the `OnStreamClosedRunnable` (users prevented propagation of request
context);
- Don't do anything with HTTP/2 connection in
`AbstractStreamingHttpConnection`;
- Improve comments for concurrency control in
`LoadBalancedStreamingHttpClient`, `AbstractStreamingHttpConnection`,
and `H2ClientParentConnectionContext`;
- Add a test for described issue;

Result:

No race for HTTP/1.x users who cancel at the connection level.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ParentConnectionContext.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ParentConnectionContext.java
@@ -163,6 +163,10 @@ class H2ParentConnectionContext extends NettyChannelListenableAsyncCloseable imp
         });
     }
 
+    private void notifyOnClosingImpl() {    // For access from AbstractH2ParentConnection
+        notifyOnClosing();
+    }
+
     final void trackActiveStream(Channel streamChannel) {
         keepAliveManager.trackActiveStream(streamChannel);
     }
@@ -228,7 +232,7 @@ class H2ParentConnectionContext extends NettyChannelListenableAsyncCloseable imp
         }
 
         private void doChannelClosed(final String method) {
-            parentContext.notifyOnClosing();
+            parentContext.notifyOnClosingImpl();
 
             if (hasSubscriber()) {
                 tryFailSubscriber(StacklessClosedChannelException.newInstance(H2ParentConnectionContext.class, method));
@@ -276,7 +280,7 @@ class H2ParentConnectionContext extends NettyChannelListenableAsyncCloseable imp
                 // We trigger the graceful close process here (with no timeout) to make sure the socket is closed once
                 // the existing streams are closed. The MultiplexCodec may simulate a GOAWAY when the stream IDs are
                 // exhausted so we shouldn't rely upon our peer to close the transport.
-                parentContext.keepAliveManager.initiateGracefulClose(parentContext::notifyOnClosing);
+                parentContext.keepAliveManager.initiateGracefulClose(parentContext::notifyOnClosingImpl);
             } else if (msg instanceof Http2PingFrame) {
                 parentContext.keepAliveManager.pingReceived((Http2PingFrame) msg);
             } else if (!(msg instanceof Http2SettingsAckFrame)) { // we ignore SETTINGS(ACK)
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ParentConnectionContext.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ParentConnectionContext.java
@@ -16,9 +16,7 @@
 package io.servicetalk.http.netty;
 
 import io.servicetalk.concurrent.Cancellable;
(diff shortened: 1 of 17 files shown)


idelpivnitskiy:
😢 we actually have many tests (related to graceful closure) that rely on `onClosing` happening from the event-loop. Trying to see if there are any other event that can be used for those tests...

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancedStreamingHttpClient.java
```diff
@@ -104,8 +111,23 @@ public void onError(final Throwable throwable) {
...
+                                // For HTTP/1.x cancellation is handled in AbstractStreamingHttpConnection.
+                                // For HTTP/2 cancellation is handled by OnStreamClosedRunnable owned by the actual
+                                // Stream. To avoid leaking the resource in case users wiped/modified the
+                                // request.context() before OnStreamClosedRunnable was propagated to the
+                                // H2ClientParentConnectionContext, we double-check ownership here and mark the request
+                                // as "finished". In normal circumstances, ownership will be taken by
+                                // H2ClientParentConnectionContext prior propagation of the Cancellable down.
                                 if (onStreamClosed != null && onStreamClosed.own()) {
+                                    if (!onStreamClosedWarningLogged) {
+                                        //noinspection AssignmentToStaticFieldFromInstanceMethod
```
Scottmitch, line 123: is there a non-IDE specific way to ignore this (e.g. `SuppressWarning(..)`)?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -941,6 +919,7 @@ public void channelActive(ChannelHandlerContext ctx) {
 
         @Override
         public void channelInactive(ChannelHandlerContext ctx) {
+            connection.notifyOnClosing();
```
Scottmitch, line 922: if a channel was never active this method may not be called (e.g. just handler removed) do we need to also invoke this method in `handlerRemove`?

idelpivnitskiy: I think it's enough to notify only from `channelInactive`. If channel was never active, and we close locally, the `channel.closeFuture()` or `closeAsync/closeAsyncGracefully` will execute first. This is here specifically for the case when channel was active but unexpectedly closed by the server and for some reason we haven't seen `ChannelInputShutdownReadComplete` before `channelInactive`. IIRC, it can happen with Java NIO networking.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ParentConnectionContext.java
```diff
@@ -88,8 +84,6 @@ class H2ParentConnectionContext extends NettyChannelListenableAsyncCloseable imp
         this.sslConfig = sslConfig;
         this.idleTimeoutMs = idleTimeoutMs;
         this.keepAliveManager = keepAliveManager;
-        // Just in case the channel abruptly closes, we should complete the onClosing Completable.
-        onClose().subscribe(onClosing::onComplete);
```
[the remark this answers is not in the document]
idelpivnitskiy: Being extra cautious will cover any potential gaps. It should not cause much overhead and close events are suppose to be infrequent. The logic still exists, just moved to `NettyChannelListenableAsyncCloseable`.

tkountis (approved):
Thanks for the code comments - helped a lot. lgtm

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ParentConnectionContext.java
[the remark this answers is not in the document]
Scottmitch: discussed offline, I thought the future listener was being removed but it was just moved to another location

Issue #2265, opened 2022-07-11
IoUringTest.ioUringIsAvailableOnLinux()


Scottmitch:
https://github.com/apple/servicetalk/runs/7258768345?check_suite_focus=true

```
java.util.concurrent.TimeoutException: ioUringIsAvailableOnLinux() timed out after 30 seconds

Suppressed: java.lang.InterruptedException
		at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1000)
		at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1308)
		at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:231)
		at io.servicetalk.concurrent.api.SourceToFuture.get(SourceToFuture.java:91)
		at io.servicetalk.http.netty.IoUringTest.ioUringIsAvailableOnLinux(IoUringTest.java:86)
		at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
		
```

Scottmitch:
https://github.com/apple/servicetalk/runs/8242401899

```
java.util.concurrent.TimeoutException: ioUringIsAvailableOnLinux() timed out after 30 seconds
	at org.junit.jupiter.engine.extension.TimeoutExceptionFactory.create(TimeoutExceptionFactory.java:29)
	at org.junit.jupiter.engine.extension.SameThreadTimeoutInvocation.proceed(SameThreadTimeoutInvocation.java:58)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:156)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:147)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:86)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(InterceptingExecutableInvoker.java:103)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.lambda$invoke$0(InterceptingExecutableInvoker.java:93)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
[...]

PR #2334, opened 2022-08-30
Add CacheConnectionHttpLoadBalanceFactory

Scottmitch:
Motivation:
When traffic spikes occur ServiceTalk will create as many
new connections as concurrent requests cannot be satisfied
with existing connections. This is common at startup and
results in more connections than necessary for protocols that
support multiple requests in parallel on the same connection
(http/1.x pipelining, h2 multiplexing).

Modifications:
- Add CacheConnectionHttpLoadBalanceFactory which wraps a
  HttpLoadBalancerFactory and will cache connection creation attempts
  according to max concurrent requests supported by the connection.
- LoadBalancerFactory#newLoadBalancer is deprecated in favor of a
  variant that doesn't have generic override on the method. This makes
  it possible for the LoadBalancer implementation to more easily wrap
  the connection and augment behavior if necessary.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
@@ -43,8 +43,7 @@ public interface LoadBalancerFactory<ResolvedAddress, C extends LoadBalancedConn
      * and hence will call {@link ConnectionFactory#closeAsync()} when {@link LoadBalancer#closeAsync()} is called.
      * @param <T> Type of connections created by the passed {@link ConnectionFactory}.
      * @return a new {@link LoadBalancer}.
-     * @deprecated In the future only {@link #newLoadBalancer(String, Publisher, ConnectionFactory)} will remain,
-     * please use that method instead.
+     * @deprecated Use {@link #newLoadBalancerTyped(String, Publisher, ConnectionFactory)}.
      */
     @Deprecated // FIXME: 0.43 - remove deprecated method
     default <T extends C> LoadBalancer<T> newLoadBalancer(
@@ -69,11 +68,35 @@ public interface LoadBalancerFactory<ResolvedAddress, C extends LoadBalancedConn
      * and hence will call {@link ConnectionFactory#closeAsync()} when {@link LoadBalancer#closeAsync()} is called.
      * @param <T> Type of connections created by the passed {@link ConnectionFactory}.
      * @return a new {@link LoadBalancer}.
+     * @deprecated Use {@link #newLoadBalancerTyped(String, Publisher, ConnectionFactory)}.
      */
+    @Deprecated // FIXME: 0.43 - remove deprecated method
     <T extends C> LoadBalancer<T> newLoadBalancer(
             String targetResource,
             Publisher<? extends Collection<? extends ServiceDiscovererEvent<ResolvedAddress>>> eventPublisher,
             ConnectionFactory<ResolvedAddress, T> connectionFactory);
+
+    /**
+     * Create a new {@link LoadBalancer}.
+     *
+     * @param targetResource A {@link String} representation of the target resource for which the created instance
+     * will perform load balancing. Bear in mind, load balancing is performed over the a collection of hosts provided
+     * via the {@code eventPublisher} which may not correspond directly to a single unresolved address, but potentially
+     * a merged collection.
+     * @param eventPublisher A stream of {@link Collection}&lt;{@link ServiceDiscovererEvent}&gt;
+     * which the {@link LoadBalancer} can use to connect to physical hosts. Typically generated
+     * from {@link ServiceDiscoverer#discover(Object) ServiceDiscoverer}.
+     * @param connectionFactory {@link ConnectionFactory} that the returned {@link LoadBalancer} will use to generate
+     * new connections. Returned {@link LoadBalancer} will own the responsibility for this {@link ConnectionFactory}
+     * and hence will call {@link ConnectionFactory#closeAsync()} when {@link LoadBalancer#closeAsync()} is called.
+     * @return a new {@link LoadBalancer}.
+     */
+    default LoadBalancer<C> newLoadBalancerTyped(
+            String targetResource,
+            Publisher<? extends Collection<? extends ServiceDiscovererEvent<ResolvedAddress>>> eventPublisher,
+            ConnectionFactory<ResolvedAddress, C> connectionFactory) {
+        return newLoadBalancer(targetResource, eventPublisher, connectionFactory);
+    }
 
     @Override
     default ExecutionStrategy requiredOffloads() {
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
@@ -187,7 +187,7 @@ final class ReservableRequestConcurrencyControllers {
 
         @Override
         public String toString() {
-            return "pendingRequests: " + pendingRequests + " maxRequests: " + maxConcurrencyHolder.lastSeenValue(-1);
+            return "pendingRequests=" + pendingRequests + " maxRequests=" + maxConcurrencyHolder.lastSeenValue(-1);
         }
     }
 
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
@@ -184,6 +184,11 @@ final class ReservableRequestConcurrencyControllers {
         final boolean casPendingRequests(int oldValue, int newValue) {
             return pendingRequestsUpdater.compareAndSet(this, oldValue, newValue);
         }
+
+        @Override
+        public String toString() {
+            return "pendingRequests: " + pendingRequests + " maxRequests: " + maxConcurrencyHolder.lastSeenValue(-1);
+        }
     }
 
     private static final class ReservableRequestConcurrencyControllerMulti
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/CacheConnectionFactory.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/CacheConnectionFactory.java
@@ -89,6 +89,9 @@ final class CacheConnectionFactory<ResolvedAddress, C extends ListenableAsyncClo
                                             result1.onClose().whenFinally(this::lockRemoveFromMap).subscribe();
                                         }
                                     } catch (Throwable cause) {
+                                        if (result1 != null) {
+                                            result1.closeAsync().subscribe();
+                                        }
                                         subscriber.onError(cause);
                                         return;
                                     }
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/CacheConnectionFactory.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/CacheConnectionFactory.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright © 2022 Apple Inc. and the ServiceTalk project authors
+ *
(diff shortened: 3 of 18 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2Setting.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+/**
+ * Represents an <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.1">HTTP/2 Setting</a>
+ * &lt;Identifier, Value&gt; pair.
+ */
+public interface H2Setting {
```
Scottmitch, line 22: this PR can be decoupled in to multiple PRs, but first wanted to get feedback on the general approach.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/LoadBalancerFactory.java
```diff
@@ -69,12 +68,36 @@ default <T extends C> LoadBalancer<T> newLoadBalancer(
      * and hence will call {@link ConnectionFactory#closeAsync()} when {@link LoadBalancer#closeAsync()} is called.
      * @param <T> Type of connections created by the passed {@link ConnectionFactory}.
      * @return a new {@link LoadBalancer}.
+     * @deprecated Use {@link #newLoadBalancerTyped(String, Publisher, ConnectionFactory)}.
```
Scottmitch, line 71: @tkountis - wdyt?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpKeepAlive.java
```diff
@@ -42,7 +42,9 @@ enum HttpKeepAlive {
     // In the interest of performance we are not accommodating for the spec allowing multiple header fields
     // or comma-separated values for the Connection header. See: https://tools.ietf.org/html/rfc7230#section-3.2.2
     static HttpKeepAlive responseKeepAlive(final HttpMetaData metaData) {
-        if (HTTP_1_1.equals(metaData.version())) {
+        // If multiple protocols are supported we don't know which protocol will be negotiated. Treat any major
```
Scottmitch, line 45: this change was moved to https://github.com/apple/servicetalk/pull/2342

Scottmitch:
build failure caused by https://github.com/apple/servicetalk/issues/2265

tkountis (approved):

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -416,12 +416,13 @@ private Single<C> selectConnection0(final Predicate<C> selector, @Nullable final
                     // Invoke the selector before adding the connection to the pool, otherwise, connection can be
                     // used concurrently and hence a new connection can be rejected by the selector.
                     if (!selector.test(newCnx)) {
-                        // Failure in selection could be temporary, hence add it to the queue and be consistent
-                        // with the fact that select failure does not close a connection.
-                        return newCnx.closeAsync().concat(failed(StacklessConnectionRejectedException.newInstance(
+                        // Failure in selection could be the result of connection factory returning queued connection,
+                        // and not having visibility into max-concurrent-requests. Propagate the exception and rely upon
+                        // retry strategy.
+                        return failed(StacklessConnectionRejectedException.newInstance(
```
idelpivnitskiy, line 422: This is concerning. I understand that right now the only known use-case for the selector to reject a new connection is a described caching use-case. However, we don't know what users do with concurrency controllers (they are public now) and it becomes harder for us to remember about this path if we make any changes for selector in the future. There is a risk to leak a real (not cached) connection. If we return an exception we should add connection to the host.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/CacheConnectionHttpLoadBalanceFactory.java
```diff
@@ -0,0 +1,198 @@
...
+import static io.servicetalk.http.api.HttpEventKey.MAX_CONCURRENCY;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A {@link HttpLoadBalancerFactory} that will cache successive connection creation attempts and return the same
+ * {@link Single} instead of creating a new connection each time. This is useful when a spike of connections occurs
+ * instead of creating a new connection for each request, if the underlying protocol version supports concurrency
+ * (pipelining, multiplexing) a single connection creation attempt can be used before the connection is actually
+ * established, which will reduce the overall number of connectiosn required.
+ * @param <ResolvableAddress> The resolved address type.
```
idelpivnitskiy, line 51: ResolvableAddress -> ResolvedAddress?

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/CacheConnectionFactory.java
```diff
@@ -0,0 +1,174 @@
...
+                                @Override
+                                public void onSuccess(@Nullable final C result1) {
+                                    try {
+                                        if (result1 == null) {
+                                            lockRemoveFromMap();
+                                        } else {
+                                            result1.onClose().whenFinally(this::lockRemoveFromMap).subscribe();
+                                        }
+                                    } catch (Throwable cause) {
+                                        subscriber.onError(cause);
```
idelpivnitskiy, line 92: Should we close the connection in case of an error here? If not, please add a comment in the code explaining why.

idelpivnitskiy (changes requested):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
```diff
@@ -184,6 +184,11 @@ final int pendingRequests() {
         final boolean casPendingRequests(int oldValue, int newValue) {
             return pendingRequestsUpdater.compareAndSet(this, oldValue, newValue);
         }
+
+        @Override
+        public String toString() {
+            return "pendingRequests: " + pendingRequests + " maxRequests: " + maxConcurrencyHolder.lastSeenValue(-1);
```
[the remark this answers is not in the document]
Scottmitch: I'll remove the class names for now as a result of the other comments, we can add them back later if we want.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: hmmm good call. Currently selector should only reject if it is closed or concurrency exceeded (already added to the host) so I don't think leaking would be and issue now however adding to host is safer.

Note that I will also filter duplicates from the collection to reduce biased toward connections and keep the false-positive selection rate down.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -620,6 +627,13 @@ boolean addConnection(C connection) {
                 }
 
                 final Object[] existing = previous.connections;
+                // Brute force iteration to avoid duplicates. If connections grow larger and faster lookup is required
+                // we can keep a Set for faster lookups (at the cost of more memory) as well as array.
+                for (final Object o : existing) {
+                    if (o.equals(connection)) {
+                        return true;
```
Scottmitch, line 634: @idelpivnitskiy - note this change. not sure connection counts are commonly high enough to justify an additional `Set` but I can add one if preferred.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/CacheConnectionFactory.java
```diff
@@ -0,0 +1,177 @@
...
+                                public void onSuccess(@Nullable final C result1) {
+                                    try {
+                                        if (result1 == null) {
+                                            lockRemoveFromMap();
+                                        } else {
+                                            result1.onClose().whenFinally(this::lockRemoveFromMap).subscribe();
+                                        }
+                                    } catch (Throwable cause) {
+                                        if (result1 != null) {
+                                            result1.closeAsync().subscribe();
```
idelpivnitskiy, line 93: Consider logging (at least at debug level) the fact we close a connection with the cause

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -416,12 +416,19 @@ private Single<C> selectConnection0(final Predicate<C> selector, @Nullable final
...
-                        // Failure in selection could be temporary, hence add it to the queue and be consistent
-                        // with the fact that select failure does not close a connection.
-                        return newCnx.closeAsync().concat(failed(StacklessConnectionRejectedException.newInstance(
+                        // Failure in selection could be the result of connection factory returning cached connection,
+                        // and not having visibility into max-concurrent-requests, or other threads already selected the
+                        // connection which uses all the max concurrent request count.
+
+                        // Just in case the connection is not closed add it to the host so we don't lose track,
+                        // duplicates will be filtered out.
+                        host.addConnection(newCnx);
```
idelpivnitskiy, line 425: It looks like we can still leak the connection if `addConnection` method returns `false`

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
idelpivnitskiy: I'm fine to start with brute force and defer any optimizations here. The only concern is comparison by `equals`. If we use `equals`, the `toLoadBalancedConnection` method have to guarantee that it can distinguish between a cached duplicate of the connection and a new connection when it applies concurrency controller and the scoring. This opens a risk to cause unexpected behavior for existing users if anyone does not provide this gurantee.

Scottmitch: `CacheConnectionFactory` doesn't wrap the connection and it returns w/e the underlying factory provides. Not sure I follow all the control flow for `toLoadBalancedConnection` but shouldn't it just be comparing objects against the same instances of them selves when caching is enabled?

idelpivnitskiy: Discussed offline and clarified ordering. After CF returns a real connection we first pass it to `toLoadBalancedConnection` and apply a concurrency controller, then apply caching layer. Therefore, comparing using default `equals` works as expected.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -416,12 +416,19 @@ private Single<C> selectConnection0(final Predicate<C> selector, @Nullable final
...
+                        // If there is caching Propagate the exception and rely upon retry strategy.
+                        Single<C> failedSingle = failed(StacklessConnectionRejectedException.newInstance(
                                 "Newly created connection " + newCnx + " for " + targetResource
                                         + " was rejected by the selection filter.",
-                                RoundRobinLoadBalancer.class, "selectConnection0(...)")));
+                                RoundRobinLoadBalancer.class, "selectConnection0(...)"));
+
+                        // Just in case the connection is not closed add it to the host so we don't lose track,
+                        // duplicates will be filtered out.
+                        return host.addConnection(newCnx) ? failedSingle : newCnx.closeAsync().concat(failedSingle);
```
idelpivnitskiy, line 431: While our `newCnx.closeAsync()` impl always completed, who knows what users applied in their filters. How do you think, in all similar cases inside RRLB, should be just subscribe to` closeAsync()` and then return a failed source instead of concatenating them?

idelpivnitskiy (commented):
Great feature, thanks!

PR #2335, opened 2022-08-30
Override `toString()` for delegating connections

idelpivnitskiy:
Motivation:

#2278 added a new connection wrapper
`DefaultFilterableStreamingHttpLoadBalancedConnection` that does not
implement `toString()`. It hides the channel id in RRLB logs.

Modifications:

- Implement `toString()` for
`DefaultFilterableStreamingHttpLoadBalancedConnection` in
`HttpLoadBalancerFactory` and `DefaultHttpLoadBalancerFactory`;
- Make `AbstractStreamingHttpConnection#toString()` final;

Result:

RRLB logs include channel id.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
@@ -230,5 +230,10 @@ public interface HttpLoadBalancerFactory<ResolvedAddress>
         public ReservedBlockingHttpConnection asBlockingConnection() {
             return toReservedBlockingConnection(this, executionContext().executionStrategy());
         }
+
+        @Override
+        public String toString() {
+            return delegate.toString();
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
```diff
@@ -230,5 +230,10 @@ public ReservedBlockingStreamingHttpConnection asBlockingStreamingConnection() {
         public ReservedBlockingHttpConnection asBlockingConnection() {
             return toReservedBlockingConnection(this, executionContext().executionStrategy());
         }
+
+        @Override
+        public String toString() {
+            return delegate.toString();
```
Scottmitch, line 236: we typically do something like `getClass().getSimpleName() + '(' delegate + ')'` does that make sense here (and other wrapper cases)?

Scottmitch (approved):

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLoadBalancerFactory.java
idelpivnitskiy: Then it's becomes a very long chain of delegation in RRLB logs. Let's keep it this way for now as these wrappers neither apply any user-visible behavior change for the connection nor add concurrency controller output in logs.

PR #2337, opened 2022-08-31
Fix flaky `ResponseCancelTest`

### LINKED ISSUE #2297: ResponseCancelTest
https://github.com/apple/servicetalk/runs/7573293427

```
io.servicetalk.transport.netty.internal.StacklessClosedChannelException
	at io.servicetalk.transport.netty.internal.DefaultNettyConnection.channelInactive(...)(Unknown Source)
```

possibly related: https://github.com/apple/servicetalk/issues/952

idelpivnitskiy:
Motivation:

#2297 indicates that `ResponseCancelTest` sometimes fails. It happens
because we incorrectly request signals from `delayedClientTermination`
queue. After cancel we may or may not see a terminal event. In rare
cases, `StacklessClosedChannelException` is propagated to the subscriber
after cancel. The next request does not assume that there is a pending
`ClientTerminationSignal` in the queue and considers this exception as
failure for a new request.

Modifications:
- Introduce a `requestId` to associate `ClientTerminationSignal` with
a proper request;
- Discard signals for prior requests inside `resume` logic;
- Wrap `signal.err` with `AssertionError` to preserve a caller stack
trace;

Result:

Fixes #2297.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseCancelTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseCancelTest.java
@@ -359,7 +359,7 @@ class ResponseCancelTest {
                 // request.
                 signal = signals.take();
                 if (signal.requestId != requestId) {
-                    LOGGER.info("Skipping {} looking for requestId={}", signal, requestId);
+                    LOGGER.info("Skipped {} because looking for requestId={}", signal, requestId);
                 }
             } while (signal.requestId != requestId);
             if (signal.err != null) {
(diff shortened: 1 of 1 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ResponseCancelTest.java
```diff
@@ -312,13 +337,19 @@ void resume() {
...
-            ClientTerminationSignal signal = signals.take();
+                           int requestId,
+                           CountDownLatch latch) throws Throwable {
+            ClientTerminationSignal signal;
+            do {
+                // In case of cancel, a terminal signal may or may not arrive to the subscriber. The requestId helps
+                // to make sure we discard optional signals of all previous requests and resuming only for the current
+                // request.
+                signal = signals.take();
+            } while (signal.requestId != requestId);
```
Scottmitch, line 349: nit: would logging `if (signal.requestId != requestId)` provide any value here in case the test stalls?

Scottmitch (approved):

PR #2340, opened 2022-09-03
Introduce `IdleTimeoutConnectionFilter` as alternative to `IDLE_TIMEOUT`

idelpivnitskiy:
Motivation:

When `IDLE_TIMEOUT` is used as a mechanism to close unused (excess)
connections, it introduces a risk to impact long-polling requests. Users
always have to coordinate request timeout value with `IDLE_TIMEOUT`
value.

Modifications:

- Add `IdleTimeoutConnectionFilter` and tests;

Result:

Users can still close excess connection without interference with
long-polling requests.

### DIFF
--- a/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
+++ b/servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
@@ -1,0 +1,242 @@
+/*
+ * Copyright © 2022 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.utils;
+
+import io.servicetalk.concurrent.Cancellable;
+import io.servicetalk.concurrent.Executor;
+import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.http.api.FilterableStreamingHttpConnection;
+import io.servicetalk.http.api.HttpExecutionStrategy;
+import io.servicetalk.http.api.StreamingHttpConnectionFilter;
+import io.servicetalk.http.api.StreamingHttpConnectionFilterFactory;
+import io.servicetalk.http.api.StreamingHttpRequest;
+import io.servicetalk.http.api.StreamingHttpResponse;
+import io.servicetalk.transport.api.ExecutionContext;
+import io.servicetalk.transport.api.RetryableException;
+import io.servicetalk.transport.api.ServiceTalkSocketOptions;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.nio.channels.ClosedChannelException;
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import javax.annotation.Nullable;
+
+import static io.servicetalk.concurrent.api.Single.defer;
+import static io.servicetalk.concurrent.api.Single.failed;
+import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
+import static java.lang.Integer.MIN_VALUE;
+import static java.time.Duration.ZERO;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+/**
+ * A connection-level filter that closes idle connections.
+ * <p>
+ * This filter is an alternative to {@link ServiceTalkSocketOptions#IDLE_TIMEOUT} at L7 layer. It helps to close idle
+ * connections that were not used to send any requests for the specified duration without affecting any in-flight
+ * requests.
+ * <ul>
+ *     <li>Connections that have in-flight requests are considered "in-use".</li>
+ *     <li>If response payload body was not consumed, the connection is still considered "in-use" and does not start
+ *     counting the timer.</li>
+ * </ul>
+ */
+public final class IdleTimeoutConnectionFilter implements StreamingHttpConnectionFilterFactory {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(IdleTimeoutConnectionFilter.class);
+
+    private static final Cancellable CANCELLED = () -> { };
+
+    private final long timeoutNs;
+    @Nullable
+    private final Executor timeoutExecutor;
+
+    /**
+     * Creates a new instance.
+     *
+     * @param timeout timeout duration after which an idle connection is closed
+     */
+    IdleTimeoutConnectionFilter(final Duration timeout) {
+        this.timeoutNs = ensurePositive(timeout).toNanos();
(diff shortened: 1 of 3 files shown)


idelpivnitskiy:
For visibility in release notes, I will send a separate PR to start using this filter instead of `IDLE_TIMEOUT` by default.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
```diff
@@ -0,0 +1,242 @@
...
+        this.timeoutExecutor = null;
+    }
+
+    /**
+     * Creates a new instance.
+     *
+     * @param timeout timeout duration after which an idle connection is closed
+     * @param timeoutExecutor the {@link Executor} to use for scheduling the timer notifications
+     */
+    IdleTimeoutConnectionFilter(final Duration timeout, final Executor timeoutExecutor) {
```
Scottmitch, line 86: how do folks create this filter if the constructors are package private?

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
```diff
@@ -0,0 +1,242 @@
...
+        this.timeoutNs = ensurePositive(timeout).toNanos();
+        this.timeoutExecutor = requireNonNull(timeoutExecutor);
+    }
+
+    private static Duration ensurePositive(final Duration timeout) {
+        if (ZERO.compareTo(timeout) >= 0) {
```
Scottmitch, line 92: use `timeout.isNegative()`?

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        @Override
+        public Single<StreamingHttpResponse> request(final StreamingHttpRequest request) {
+            return defer(() -> {
+                final int inFlightRequests = requestsUpdater.incrementAndGet(this);
+                if (inFlightRequests < 0) {
+                    // To mitigate the tiny risk of going into positive numbers with Integer.MAX_VALUE concurrent
+                    // requests (unrealistic), reset the counter back to MIN_VALUE.
+                    requests = MIN_VALUE;
```
Scottmitch, line 164: if we are protecting against overflow we should do it atomically. Otherwise is likely to cause harm if another thread decrements the value while it is negative we will never get back to 0 as expected.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
```diff
@@ -0,0 +1,242 @@
...
+                        .liftSync(new BeforeFinallyHttpOperator(() -> {
+                            // It's acceptable to use 2 volatile variables instead of a single object state here. Even
+                            // if 2 threads race between updating "lastResponseTime" and "requests", the delay for a new
+                            // timer task will be close to "timeoutNs".
+                            requestsUpdater.updateAndGet(this, prev -> {
+                                if (prev > 1) {
+                                    return prev - 1;
+                                }
+                                assert prev == 1 : "Unexpected requests value: " + prev;
+                                lastResponseTime = nanoTime();
```
Scottmitch, line 177: the updater function maybe called repeated times if the atomic operation fails, and doesn't ensure the count is `0` ... this should be moved outside the scope of the updater.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
```diff
@@ -0,0 +1,242 @@
...
+                                    connection, NANOSECONDS.toMillis(timeoutNs));
+                            connection.closeAsync().subscribe();
+                            return;
+                        }
+                    } else {
+                        updateIdleTimeout(nextDelayNs);
+                        return;
+                    }
+                } else {
+                    // Should never happen. Keep it just in case to prevent infinite loop.
```
Scottmitch, line 220: this may happen unless we atomically prevent overflow.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
```diff
@@ -0,0 +1,242 @@
...
+
+        private volatile int requests;
+        @Nullable
+        private volatile Cancellable timeoutTask;
+
+        private final long timeoutNs;
+        private final Executor timeoutExecutor;
+
+        // While it may look like "volatile" is not required for this variable for "happens-before" visibility because
+        // we always write to "requests" after updating "lastResponseTime" and read "requests" before reading
```
Scottmitch, line 130: `we always write to "requests" after updating "lastResponseTime"` -> this isn't quite true yet pending the comment below

Scottmitch: Consider simplifying/clarifying this comment to say the volatile is not for general visibility but to prevent [word tearing](https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.6)

Scottmitch (changes requested):

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
idelpivnitskiy: Since this is a connection level filter, it always works after concurrency controller, which will make sure we never have more than `Integer.MAX_VALUE` concurrent requests on a single connection. Technically, the overflow can happen for a reserve connections.

I don't think using `addWithOverflowProtection` is the right choice here as it will lead to a problem of incorrectly detecting when connection is idle. 
For example, if there are `MAX_VALUE + 1` concurrent requests on a connection (for h2 that extra request will fail asap, for h1 it's possible to have quite long pipeline), finishing the first `MAX_VALUE` requests will return the counter back to 0, but there might be still 1 request running. 

WDYT if this filter will start rejecting new requests instead of overflowing? `Integer.MAX_VALUE`  seems a reasonable values users should not exceed. Alternatively, we can switch to `Long` and start rejecting only for long overflow.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
idelpivnitskiy: Assuming there is no overflow: the increment and decrement are tight to a request. It can go to 0 only if all requests complete. It may race between N to N-1 transition but it will never race from 1 to 0.

Scottmitch: This code looks better 👍 .The issue was updating `lastResponseTime` state was inside the updater function (which actually writes before `requestsUpdater` is updated and maybe invoked multiple times if the atomic operation fails).

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
Scottmitch: discussed offline. rejecting to prevent overflow makes sense.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/IdleTimeoutConnectionFilter.java
Scottmitch: I didn't consider the `zero` isn't valid, thx for clarifying. fine as is 👍

Scottmitch (approved):

PR #2341, opened 2022-09-03
HTTP/2 set initial settings

Scottmitch:
Motivaiton:
There is no way to define initial values for
[http/2 settings](https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2).

Modifications:
- Enhance H2ProtocolConfigBuilder to accept Map<Character, Integer> to
  set the initial settings values.
- Set the default value of settings with bound max header list size
  and 1mb default initial flow control window.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
@@ -29,7 +29,7 @@ public interface Http2Settings {
      * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_HEADER_TABLE_SIZE</a>.
      */
     @Nullable
-    Integer headerTableSize();
+    Long headerTableSize();
 
     /**
      * Get the value for
@@ -38,7 +38,7 @@ public interface Http2Settings {
      * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_MAX_CONCURRENT_STREAMS</a>.
      */
     @Nullable
-    Integer maxConcurrentStreams();
+    Long maxConcurrentStreams();
 
     /**
      * Get the value for
@@ -65,7 +65,7 @@ public interface Http2Settings {
      * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_MAX_HEADER_LIST_SIZE</a>.
      */
     @Nullable
-    Integer maxHeaderListSize();
+    Long maxHeaderListSize();
 
     /**
      * Get the setting value associated with an
@@ -74,11 +74,11 @@ public interface Http2Settings {
      * @return {@code null} if no setting value corresponding {@code identifier} exists, otherwise the value.
      */
     @Nullable
-    Integer settingValue(char identifier);
+    Long settingValue(char identifier);
 
     /**
      * Iterate over all the &lt;identifier, value&gt; tuple in this settings object.
      * @param action Invoked on each &lt;identifier, value&gt; tuple.
      */
-    void forEach(BiConsumer<? super Character, ? super Integer> action);
+    void forEach(BiConsumer<? super Character, ? super Long> action);
 }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright © 2022 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.api;
+
+/**
+ * Utilities for <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.1">HTTP/2 Setting</a>.
+ */
+public final class Http2Settings {
+    /**
+     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_HEADER_TABLE_SIZE</a>.
+     */
+    public static final char HEADER_TABLE_SIZE = 0x1;
+
+    /**
+     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_ENABLE_PUSH</a>.
+     */
+    public static final char ENABLE_PUSH = 0x2;
+
+    /**
+     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">
+     *     SETTINGS_MAX_CONCURRENT_STREAMS</a>.
+     */
+    public static final char MAX_CONCURRENT_STREAMS = 0x3;
+
+    /**
+     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">
+     *     SETTINGS_INITIAL_WINDOW_SIZE</a>.
+     */
+    public static final char INITIAL_WINDOW_SIZE = 0x4;
+
+    /**
+     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_MAX_FRAME_SIZE</a>.
+     */
+    public static final char MAX_FRAME_SIZE = 0x5;
+
+    /**
+     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">
+     *     SETTINGS_MAX_HEADER_LIST_SIZE</a>.
+     */
+    public static final char MAX_HEADER_LIST_SIZE = 0x6;
+
+    private Http2Settings() {
+    }
+}
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
@@ -15,43 +15,70 @@
  */
 package io.servicetalk.http.api;
 
+import java.util.function.BiConsumer;
+import javax.annotation.Nullable;
+
 /**
- * Utilities for <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.1">HTTP/2 Setting</a>.
+ * Object representing a <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.1">HTTP/2 Setting</a> frame.
  */
-public final class Http2Settings {
+public interface Http2Settings {
     /**
-     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_HEADER_TABLE_SIZE</a>.
+     * Get the value for
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_HEADER_TABLE_SIZE</a>.
+     * @return the value for
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_HEADER_TABLE_SIZE</a>.
      */
-    public static final char HEADER_TABLE_SIZE = 0x1;
+    @Nullable
+    Integer headerTableSize();
 
     /**
-     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_ENABLE_PUSH</a>.
+     * Get the value for
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_MAX_CONCURRENT_STREAMS</a>.
+     * @return the value for
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_MAX_CONCURRENT_STREAMS</a>.
      */
-    public static final char ENABLE_PUSH = 0x2;
+    @Nullable
+    Integer maxConcurrentStreams();
 
     /**
-     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">
-     *     SETTINGS_MAX_CONCURRENT_STREAMS</a>.
+     * Get the value for
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_INITIAL_WINDOW_SIZE</a>.
+     * @return the value for
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_INITIAL_WINDOW_SIZE</a>.
      */
-    public static final char MAX_CONCURRENT_STREAMS = 0x3;
+    @Nullable
+    Integer initialWindowSize();
 
     /**
-     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">
-     *     SETTINGS_INITIAL_WINDOW_SIZE</a>.
+     * Get the value for
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_MAX_FRAME_SIZE</a>.
+     * @return the value for
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_MAX_FRAME_SIZE</a>.
      */
-    public static final char INITIAL_WINDOW_SIZE = 0x4;
+    @Nullable
+    Integer maxFrameSize();
 
     /**
-     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_MAX_FRAME_SIZE</a>.
+     * Get the value for
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_MAX_HEADER_LIST_SIZE</a>.
+     * @return the value for
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_MAX_HEADER_LIST_SIZE</a>.
      */
-    public static final char MAX_FRAME_SIZE = 0x5;
+    @Nullable
+    Integer maxHeaderListSize();
 
     /**
-     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">
-     *     SETTINGS_MAX_HEADER_LIST_SIZE</a>.
+     * Get the setting value associated with an
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.1">identifier</a>.
+     * @param identifier the <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.1">identifier</a>.
+     * @return {@code null} if no setting value corresponding {@code identifier} exists, otherwise the value.
      */
-    public static final char MAX_HEADER_LIST_SIZE = 0x6;
+    @Nullable
+    Integer settingValue(char identifier);
 
-    private Http2Settings() {
-    }
+    /**
+     * Iterate over all the &lt;identifier, value&gt; tuple in this settings object.
+     * @param action Invoked on each &lt;identifier, value&gt; tuple.
+     */
+    void forEach(BiConsumer<? super Character, ? super Integer> action);
 }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2SettingsBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2SettingsBuilder.java
@@ -25,6 +25,9 @@ import javax.annotation.Nullable;
  * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.1">HTTP/2 Setting</a>.
  */
 public final class Http2SettingsBuilder {
+    private static final long MAX_UNSIGNED_INT = 0xffffffffL;
+    private static final int MAX_FRAME_SIZE_LOWER_BOUND = 0x4000;
+    private static final int MAX_FRAME_SIZE_UPPER_BOUND = 0xffffff;
     /**
      * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_HEADER_TABLE_SIZE</a>.
      */
@@ -48,7 +51,7 @@ public final class Http2SettingsBuilder {
      *     SETTINGS_MAX_HEADER_LIST_SIZE</a>.
      */
     private static final char MAX_HEADER_LIST_SIZE = 0x6;
-    private final Map<Character, Integer> settings;
+    private final Map<Character, Long> settings;
 
     /**
      * Create a new instance.
@@ -72,7 +75,8 @@ public final class Http2SettingsBuilder {
      * @param value The value.
      * @return {@code this}.
      */
-    public Http2SettingsBuilder headerTableSize(int value) {
+    public Http2SettingsBuilder headerTableSize(long value) {
+        validate32Unsigned(value);
         settings.put(HEADER_TABLE_SIZE, value);
         return this;
     }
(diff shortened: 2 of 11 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpProtocolConfigs.java
```diff
@@ -52,6 +54,11 @@ public static H1ProtocolConfigBuilder h1() {
     /**
      * Returns {@link H2ProtocolConfig} with the default configuration for
      * <a href="https://tools.ietf.org/html/rfc7540">HTTP/2</a>.
+     * <p>
+     * Note this doesn't necessarily provide {@link H2ProtocolConfig#initialSettings()} that corresponds to
```
Scottmitch, line 58: worth calling out for visibility...  default values were set based upon experience and this carve out is meant to highlight that folks should be explicit if they depend upon specific values (defaults are subject to change).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
```diff
@@ -60,6 +61,21 @@ default String alpnId() {
...
+     * settings may be ignored if not supported (e.g. push promise).
+     */
+    Map<Character, Integer> initialSettings();
+
+    /**
+     * Provide a hint on the number of bytes that the flow controller will attempt to give to a stream for each
+     * allocation (assuming the stream has this much eligible data).
+     * @return number of bytes.
+     */
+    int flowControlQuantum();
```
Scottmitch, line 77: now that folks can control the initial flow control we should also allow for them to control flow control quantum. incase the flow control window is very big or small they may want to adjust this.

Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/2343

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public static final char HEADER_TABLE_SIZE = 0x1;
+
+    /**
+     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_ENABLE_PUSH</a>.
+     */
+    public static final char ENABLE_PUSH = 0x2;
+
+    /**
+     * Identifier <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">
```
idelpivnitskiy, line 33: Consider clarifying in javadoc that ServiceTalk supports this option only for the server bcz client does not support server push.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
```diff
@@ -84,4 +87,9 @@ public void channelRegistered(final ChannelHandlerContext ctx) {
             // Http2ConnectionHandler.processGoAwayWriteResult will close the connection after GO_AWAY is flushed
         }
     }
+
+    private static Http2Settings applyClientDefaultSettings(Http2Settings settings) {
+        // Notify server that this client does not support server push and request it to be disabled.
+        return settings.pushEnabled(false).maxConcurrentStreams(0L);
```
idelpivnitskiy, line 93: This is hard to debug if we silently override these 2 settings. With a custom `Http2Settings` interface we can prevent users from specifying `pushEnabled` at the API level. For `maxConcurrentStreams`, consider validating settings at the client builder and throw an exception if users try to provide `maxConcurrentStreams`.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
```diff
@@ -0,0 +1,57 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.api;
+
+/**
+ * Utilities for <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.1">HTTP/2 Setting</a>.
+ */
+public final class Http2Settings {
```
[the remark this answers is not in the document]
Scottmitch: I will try this approach. We need a way to iterate over keys and get any arbitrary keys as folks could add any settings but I can add this.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
Scottmitch: I will remove this as part of the HttpSettings API rework.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
```diff
@@ -60,6 +61,21 @@ default String alpnId() {
...
+     * Get a {@link Map} which provides a hint for the initial settings for any h2 connection. Note that some settings
+     * may be ignored if not supported (e.g. push promise).
+     * @return a {@link Map} which provides a hint for the initial settings for any h2 connection. Note that some
+     * settings may be ignored if not supported (e.g. push promise).
+     */
+    Map<Character, Integer> initialSettings();
+
+    /**
+     * Provide a hint on the number of bytes that the flow controller will attempt to give to a stream for each
+     * allocation (assuming the stream has this much eligible data).
```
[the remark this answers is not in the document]
Scottmitch: done. ptal.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
```diff
@@ -0,0 +1,84 @@
...
+    Integer maxHeaderListSize();
+
+    /**
+     * Get the setting value associated with an
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.1">identifier</a>.
+     * @param identifier the <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.1">identifier</a>.
+     * @return {@code null} if no setting value corresponding {@code identifier} exists, otherwise the value.
+     */
+    @Nullable
+    Integer settingValue(char identifier);
```
idelpivnitskiy, line 77: Do we currently have a use-case for a custom setting? If not, let's keep API minimal by exposing only settings from the spec that we know the behavior for. Arbitrary settings may lead to unexpected behavior or override known settings.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Iterate over all the &lt;identifier, value&gt; tuple in this settings object.
+     * @param action Invoked on each &lt;identifier, value&gt; tuple.
+     */
+    void forEach(BiConsumer<? super Character, ? super Integer> action);
```
idelpivnitskiy, line 83: If we can defer adding `settingValue`, let's defer adding this method too.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2SettingsBuilder.java
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        @Override
+        public int hashCode() {
+            return settings.hashCode();
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            return o instanceof DefaultHttp2Settings && settings.equals(((DefaultHttp2Settings) o).settings);
```
idelpivnitskiy, line 198: For the purpose of `equals` and `hashCode` it should not matter what is the impl type of the `Http2Settings` interface. We should compare only the actual values present.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
```diff
@@ -84,4 +82,15 @@ public void channelRegistered(final ChannelHandlerContext ctx) {
...
         }
     }
+
+    private static Http2Settings validateClientSettings(Http2Settings settings) {
+        final Boolean pushEnabled = settings.pushEnabled();
+        if (pushEnabled == null) {
+            // Notify server that this client does not support server push and request it to be disabled.
+            settings.pushEnabled(false);
+        } else if (pushEnabled) {
+            throw new IllegalArgumentException("push is enabled but not supported. settings=" + settings);
```
idelpivnitskiy, line 92: We should also keep maxConcurrentStreams set to 0L and don't let users override it.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentChannelInitializer.java
```diff
@@ -70,4 +73,34 @@ static void initFrameLogger(final Http2FrameCodecBuilder multiplexCodecBuilder,
                             frameLoggerConfig.logLevel()), frameLoggerConfig.logUserData()));
         }
     }
+
+    static io.netty.handler.codec.http2.Http2Settings toNettySettings(Http2Settings h2Settings) {
+        io.netty.handler.codec.http2.Http2Settings nettySettings = new io.netty.handler.codec.http2.Http2Settings();
```
idelpivnitskiy, line 78: Using the default `io.netty.handler.codec.http2.Http2Settings()` will override the default settings set by Netty: maxHeaderListSize=8192.
Consider taking `multiplexCodecBuilder.initialSettings()` and apply only non-null ST settings on top of it. By using `multiplexCodecBuilder.initialSettings()` instead of  `io.netty.handler.codec.http2.Http2Settings.defaultSettings()` you save allocation as the `multiplexCodecBuilder` always invokes `defaultSettings()`.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2SettingsBuilder.java
```diff
@@ -0,0 +1,201 @@
...
+     */
+    public Http2Settings build() {
+        return new DefaultHttp2Settings(settings);
+    }
+
+    private static final class DefaultHttp2Settings implements Http2Settings {
+        private final Map<Character, Integer> settings;
+
+        private DefaultHttp2Settings(final Map<Character, Integer> settings) {
+            this.settings = settings;
```
idelpivnitskiy, line 142: If ctor doesn't make a copy of the `Map`, builder still have reference to the same map and can change values after `build()`. If you remove support for custom settings, consider using pojo without the internal map.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
```diff
@@ -60,6 +61,28 @@ default String alpnId() {
     @Nullable
     KeepAlivePolicy keepAlivePolicy();
 
+    /**
+     * Get the {@link Http2Settings} that provides a hint for the initial settings. Note that some settings may be
+     * ignored if not supported (e.g. push promise).
+     * @return the {@link Http2Settings} that provides a hint for the initial settings. Note that some settings may be
+     * ignored if not supported (e.g. push promise).
```
idelpivnitskiy, line 68: Consider clarifying in javadoc how to build settings and/or add `@see` tag to link to the builder for settings.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
idelpivnitskiy: Do you plan to add connection level flow-control in this PR? Without it, increased `INITIAL_WINDOW_SIZE` for streams won't have much effect.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfigBuilder.java
```diff
@@ -103,33 +106,79 @@ public H2ProtocolConfigBuilder keepAlivePolicy(final KeepAlivePolicy policy) {
...
      */
     public H2ProtocolConfig build() {
-        return new DefaultH2ProtocolConfig(headersFactory, headersSensitivityDetector, frameLoggerConfig,
-                keepAlivePolicy);
+        return new DefaultH2ProtocolConfig(h2Settings, headersFactory, headersSensitivityDetector, frameLoggerConfig,
+                keepAlivePolicy, flowControlQuantum);
     }
 
-    private static final class DefaultH2ProtocolConfig implements H2ProtocolConfig {
+    static Http2SettingsBuilder newDefaultSettingsBuilder() {
```
idelpivnitskiy, line 146: Consider making this method private

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfigBuilder.java
```diff
@@ -103,33 +106,79 @@ public H2ProtocolConfigBuilder keepAlivePolicy(final KeepAlivePolicy policy) {
...
+                .initialWindowSize(1048576) // 1mb default window size
+                .maxHeaderListSize((int) DEFAULT_HEADER_LIST_SIZE);
+    }
+
+    /**
+     * Default allocation quantum to use for the remote flow controller.
+     * @return Default allocation quantum to use for the remote flow controller.
+     */
+    private static int defaultFlowControlQuantum() {
+        return 1024 * 16;
```
idelpivnitskiy, line 157: Since this method is private, consider assigning the value directly to the `flowControlQuantum`, then we can kill the method. It will be easier for developers to read default value without the need to jump to the end of the class.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2SettingsBuilder.java
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * Set the value for
+     * <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2">SETTINGS_MAX_CONCURRENT_STREAMS</a>.
+     *
+     * @param value The value.
+     * @return {@code this}.
+     */
+    public Http2SettingsBuilder maxConcurrentStreams(int value) {
+        settings.put(MAX_CONCURRENT_STREAMS, value);
```
idelpivnitskiy, line 88: Consider adding validation for each value to make sure those are within expected range

idelpivnitskiy (changes requested):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentChannelInitializer.java
idelpivnitskiy: > I removed the empty/default behavior here as defaults are already applied in the builder.

Ok, I saw later that you set all default at ST level. I don't mind either way.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
Scottmitch: Yes I can add this. increased flow control windows will still have effect without it, just that single streams can exhaust connection window more easily without it.

Scottmitch: Actually we ported increasing the connection window to Netty (doubles it by default). I will add an option to extend even further in ServiceTalk but set to 0 by default.

https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java#L229-L243

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
Scottmitch: I think we should leave this as there maybe arbitrary setting identifiers (not only for us to process) and we won't be able to easily add it later once the API stabilizes.

idelpivnitskiy: We can add `default` method later if necessary and support by our impl. It's hard to imagine a valid use-case for a custom setting, and it opens more doors for misuse. We are not netty that have to be flexible for any purpose, the controlled environment that doesn't let users break themselves is more user-friendly.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentChannelInitializer.java
Scottmitch: Defaults being applied at the builder makes more sense ... then folks can override them and we will use them and/or validate them later.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
Scottmitch: This is the accessor side, so folks don't need to know how to build a `Http2Settings` ... there is already a `@see Http2SettingsBuilder` on the `H2ProtocolConfigBuilder#initialSettings` method.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfigBuilder.java
Scottmitch: I will remove it and inline it as it is no longer necessary.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfigBuilder.java
Scottmitch: I killed the method.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2SettingsBuilder.java
Scottmitch: good catch, I will copy.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2SettingsBuilder.java
idelpivnitskiy: On the other hand, maybe we should. Otherwise, users can extend `Http2Settings` to add more characteristics. Let's keep the `instanceof`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
```diff
@@ -60,6 +61,35 @@ default String alpnId() {
     @Nullable
     KeepAlivePolicy keepAlivePolicy();
 
+    /**
+     * Get the {@link Http2Settings} that provides a hint for the initial settings. Note that some settings may be
+     * ignored if not supported (e.g. push promise).
```
idelpivnitskiy, line 66: Consider clarifying that we throw an exception instead of ignoring

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
```diff
@@ -60,6 +61,35 @@ default String alpnId() {
...
+     * consider <a href="https://datatracker.ietf.org/doc/html/rfc7540#section-5.3">stream priority</a>) that have data
+     * pending to write and allow them to write a subset of the available flow control window (aka a "quantum"). The
+     * larger this value may result in increased goodput/throughput on the connection, but increase latency on some
+     * streams (if flow control windows become constrained).
+     * @return number of bytes.
+     */
+    int flowControlQuantum();
+
+    /**
+     * Increment to apply to {@link Http2Settings#initialWindowSize()} for the connection stream. This expands the
```
idelpivnitskiy, line 87: From the spec:
> The WINDOW_UPDATE frame can be specific to a stream or to the entire connection.

"for the connection stream" is a bit confusing here. It sounds like the setting increments window for frames with streamId=0. Consider saying that it applies to the entire connection, not a "connection stream".

Also, consider linking to [WINDOW_UPDATE](https://datatracker.ietf.org/doc/html/rfc7540#section-6.9) instead of `Http2Settings#initialWindowSize()`.

Last, please clarify in the javadoc that this is an increment in bytes.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
Scottmitch: discussed offline ... I think it is useful to keep for now as we would ideally use the same object in the future to expose to users (e.g. for visibility in transport observer, `StreamingHttpConnection.transportEventStream`, ..). The risk it will confuse folks is minimal as for creation they will mostly go through the builder.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2SettingsBuilder.java
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        @Override
+        public void forEach(final BiConsumer<? super Character, ? super Long> action) {
+            settings.forEach(action);
+        }
+
+        @Override
+        public String toString() {
+            return settings.toString();
```
idelpivnitskiy, line 211: The output of this impl is: `{=16384, =10}`.
Consider implementing it in the way it prints hex value for the identifier and maybe a string name for known identifiers.

idelpivnitskiy (commented):
Overall looks great, last findings:

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
idelpivnitskiy: Agreed, would be nice to reuse the same API for observing settings. I like that you haven't added the builder method for a custom setting yet. Only advanced users will be able to configure that by implementing `Http2Settings` contract.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Http2Settings.java
```diff
@@ -0,0 +1,84 @@
...
+     * @return {@code null} if no setting value corresponding {@code identifier} exists, otherwise the value.
+     */
+    @Nullable
+    Long settingValue(char identifier);
+
+    /**
+     * Iterate over all the &lt;identifier, value&gt; tuple in this settings object.
+     * @param action Invoked on each &lt;identifier, value&gt; tuple.
+     */
+    void forEach(BiConsumer<? super Character, ? super Long> action);
```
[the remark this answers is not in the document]
Scottmitch: `super` allows for super cases to be as argument types instead of exact types and is what `Map` uses so the following (and other variations) are possible:

```java
BiConsumer<Object, Object> foo = (key, value) -> {
};
h2Settings.forEach(foo);
```

Let just leave it consistent with `Map` unless this causes an issue.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
```diff
@@ -51,21 +56,30 @@ public void init(final Channel channel) {
...
 
         channel.pipeline().addLast(multiplexCodecBuilder.build(),
                 new Http2MultiplexHandler(H2PushStreamHandler.INSTANCE));
+        if (config.flowControlWindowIncrement() > 0) {
+            // Must be after Http2ConnectionHandler does its initialization. The client must wait until after
+            // the connection preface and settings are sent.
+            channel.pipeline().addLast(new ChannelInboundHandlerAdapter() {
+                @Override
+                public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
+                    if (evt instanceof Http2ConnectionPrefaceAndSettingsFrameWrittenEvent) {
```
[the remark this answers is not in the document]
Scottmitch: singleton instance is package private so we can't check by reference.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfig.java
Scottmitch: the behavior may vary depending upon the use site and implementation, some settings maybe OK to ignore ... I don't think we need to over specify exception usage (illegal argument exception may always be thrown).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
```diff
@@ -22,25 +22,30 @@
...
 
 final class H2ClientParentChannelInitializer implements ChannelInitializer {
-
     private final H2ProtocolConfig config;
+    private final io.netty.handler.codec.http2.Http2Settings nettySettings;
 
     H2ClientParentChannelInitializer(final H2ProtocolConfig config) {
         this.config = config;
+        final io.servicetalk.http.api.Http2Settings h2Settings = config.initialSettings();
+        nettySettings = validateClientSettings(toNettySettings(h2Settings));
```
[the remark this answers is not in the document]
Scottmitch: I will validate push when setting the protocol, but I will remove validation of max concurrent streams bcz it shouldn't hurt if push is disabled (and it would now require wrapping the object to override, plus not sure we want to override the value anyways as this can be confusing).

Scottmitch:
build failure caused by https://github.com/apple/servicetalk/issues/2345

PR #2342, opened 2022-09-03
HTTP/2 avoid premature connection closure when negotiating via ALPN

Scottmitch:
Motivation:
ServiceTalk can negotiate http/1.x or http/2 through ALPN when TLS
is enabled. ServiceTalk doesn't know which request factory to use
until the protocol negotiation completes. It is possible that the
http/1.x request factory is used when speaking http/2.x which may
result in premature connection closure due to keep alive logic bug.

Modifications:
- http/2+ should be treated as always using keep alive and doesn't
  use connection specific keep alive headers [1].

[1] https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2.2


Scottmitch:
discovered while testing https://github.com/apple/servicetalk/pull/2334 and connections were being prematurely closed resulting in higher number than expected connections.

idelpivnitskiy (approved):

PR #2344, opened 2022-09-07
Make io_uring tests on linux more lenient

idelpivnitskiy:
Motivation:

Users may run tests on linux kernel that does not support io_uring. To avoid failure of `IoUringTest.ioUringIsAvailableOnLinux()` we need to run it only when io_uring is available.

Modifications:

- Use `assumeTrue` instead of `assertTrue`;
- Test with & without offloading;

Result:

Users can run tests on any linux kernel without failure due to io_uring unavailability.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/IoUringTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/IoUringTest.java
@@ -15,6 +15,7 @@
  */
 package io.servicetalk.http.netty;
 
+import io.servicetalk.concurrent.internal.TestTimeoutConstants;
 import io.servicetalk.http.api.BlockingHttpClient;
 import io.servicetalk.http.api.HttpRequest;
 import io.servicetalk.http.api.HttpResponse;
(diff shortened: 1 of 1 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/IoUringTest.java
```diff
@@ -58,22 +60,25 @@ void ioUringIsNotAvailableOnMacOs() {
...
+            assumeTrue(IoUringUtils.isAvailable(), "io_uring is unavailable on " +
+                    System.getProperty("os.name") + ' ' + System.getProperty("os.version"));
             IOUring.ensureAvailability();
 
             ioUringExecutor = NettyIoExecutors.createIoExecutor(2, "io-uring");
             assertThat(ioUringExecutor.eventLoopGroup(), is(instanceOf(IOUringEventLoopGroup.class)));
 
             try (ServerContext serverContext = HttpServers.forAddress(localAddress(0))
                     .ioExecutor(ioUringExecutor)
+                    .executionStrategy(offloadNone())
```
Scottmitch, line 77: should this test be parameterized? it should work with and without offloading IIUC?

idelpivnitskiy: it shouldn't matter for this test, but I added a param anyway

PR #2346, opened 2022-09-08
Revert default HTTP/2 settings sent by the client

idelpivnitskiy:
Motivation:

#2341 changed default HTTP/2 settings that client sends to notify the server it does not support Server Push.

Modification:

- Set `ENABLE_PUSH=0, MAX_CONCURRENT_STREAMS=0` in settings frame that client sends to a server;
- Validate custom initial settings that users pass to `H2ProtocolConfig` and throw if users try to override ENABLE_PUSH/MAX_CONCURRENT_STREAMS on the client or set ENABLE_PUSH other than 0 for the server;
- Include content message in GO_AWAY frame if client receives a Push Stream frame;

Result:

Default client behavior doesn't change, custom settings validated to prevent programming mistakes.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
@@ -27,6 +27,7 @@ import io.netty.handler.codec.http2.Http2ConnectionPrefaceAndSettingsFrameWritte
 import io.netty.handler.codec.http2.Http2FrameCodecBuilder;
 import io.netty.handler.codec.http2.Http2MultiplexHandler;
 
+import static io.netty.buffer.ByteBufUtil.writeAscii;
 import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
 import static io.servicetalk.http.netty.H2ServerParentChannelInitializer.initFrameLogger;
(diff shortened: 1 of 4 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
```diff
@@ -37,8 +38,7 @@ final class H2ClientParentChannelInitializer implements ChannelInitializer {
 
     H2ClientParentChannelInitializer(final H2ProtocolConfig config) {
         this.config = config;
-        final io.servicetalk.http.api.Http2Settings h2Settings = config.initialSettings();
-        nettySettings = toNettySettings(h2Settings);
+        nettySettings = toNettySettings(config.initialSettings()).pushEnabled(false).maxConcurrentStreams(0L);
```
Scottmitch, line 41: If we set `pushEnabled` to false we don't really need to set `maxConcurrentStreams`

Scottmitch (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
idelpivnitskiy: Either way it doesn't hurt. This is what grpc-java does. Let's be consistent.

PR #2348, opened 2022-09-09
Publisher#flatMapMerge allow terminal propagation after invalid demand

Scottmitch:
Motivation:
Publisher#flatMapMerge has internal state that tracks demand. This state is used to drive the internal state machine and prevents invalid demand to avoid internal state being corrupted. However this code didn't always allow for the subsequent terminal (e.g. likely onError) to be propagated downstream which may result in a hang.

Modifications:
- FlatMapPublisherSubscriber demand tracking shouldn't prevent terminal if there is too many items delivered.
- Remove conditionals that are unnecessary for needsDemand because the call sites know if demand is required.


tkountis (approved):

tkountis:
Build failures due to https://github.com/apple/servicetalk/issues/2245

PR #2349, opened 2022-09-09
Use `IdleTimeoutConnectionFilter` instead of SocketOption by default

idelpivnitskiy:
Motivation:

`ServiceTalkSocketOptions.IDLE_TIMEOUT` currently has the default value of 5 min. This default value makes it harder for users to coordinate between per-request timeout and per-client `IDLE_TIMEOUT`. Example: long-polling requests that use much higher timeout might be interrupted by lower `IDLE_TIMEOUT` value.

Modifications:

- Change the default value for `IDLE_TIMEOUT` to 0;
- If users didn't append `IdleTimeoutConnectionFilter`, configure one by default with 5 min timeout;
- Allow configuring `IdleTimeoutConnectionFilter` with `Duration.ZERO`;
- Add `toString()` impl for `IdleTimeoutConnectionFilter`;

Result:

Long-polling requests are not affected by default `IDLE_TIMEOUT` value, while ST users still have a filter that closes connections that were not used to serve any requests for 5 minutes. To change this value, users should append their own `IdleTimeoutConnectionFilter` with different value. To disable timeouts, users should append
`IdleTimeoutConnectionFilter` with `Duration#ZERO` as a timeout value.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
@@ -48,6 +48,7 @@ import io.servicetalk.http.api.StreamingHttpConnectionFilterFactory;
 import io.servicetalk.http.api.StreamingHttpRequest;
 import io.servicetalk.http.api.StreamingHttpRequestResponseFactory;
 import io.servicetalk.http.utils.HostHeaderHttpRequesterFilter;
+import io.servicetalk.http.utils.IdleTimeoutConnectionFilter;
 import io.servicetalk.logging.api.LogLevel;
 import io.servicetalk.transport.api.ClientSslConfig;
 import io.servicetalk.transport.api.ExecutionStrategy;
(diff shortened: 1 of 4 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -491,6 +503,12 @@ private void ifHostHeaderHttpRequesterFilter(final Object filter) {
         }
     }
 
+    private void ifIdleTimeoutConnectionFilter(final StreamingHttpConnectionFilterFactory factory) {
```
Scottmitch, line 506: `ifIdleTimeoutConnectionFilter` -> `checkIfIdleTimeoutConnectionFilter`

Scottmitch (approved):
🎉

idelpivnitskiy:
All checks passed but CodeQL result was not propagated to this PR, still shows as "In Progress". On [Actions](https://github.com/apple/servicetalk/actions/workflows/codeql-analysis.yml) page we can see that CodeQL also finished successfully. 
There is no UI to restart CodeQL task.

PR #2351, opened 2022-09-09
Update jctools 3.3.1-ea -> 4.0.1

Scottmitch:
Modifications:
- `io.servicetalk.internal.queues.useUnpadded` value defaults to "false" to use unpadded queues by default. This property will be removed in future releases, but exists in case there are any regressions detected.

### DIFF
--- a/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent.java
+++ b/servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent.java
@@ -40,11 +40,11 @@ import org.jctools.queues.atomic.MpscLinkedAtomicQueue;
 import org.jctools.queues.atomic.MpscUnboundedAtomicArrayQueue;
 import org.jctools.queues.atomic.SpscGrowableAtomicArrayQueue;
 import org.jctools.queues.atomic.SpscUnboundedAtomicArrayQueue;
-import org.jctools.queues.ea.unpadded.MpscChunkedUnpaddedArrayQueue;
-import org.jctools.queues.ea.unpadded.MpscLinkedUnpaddedQueue;
-import org.jctools.queues.ea.unpadded.MpscUnboundedUnpaddedArrayQueue;
-import org.jctools.queues.ea.unpadded.SpscChunkedUnpaddedArrayQueue;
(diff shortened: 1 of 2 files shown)


idelpivnitskiy:
API removal in jctools 4.x introduces a risk of breaking other jctools use-cases outside of ServiceTalk and Netty scope. Let's defer this for now and consider for later releases. Maybe 0.43.x.

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/PlatformDependent.java
```diff
@@ -277,12 +278,13 @@ private static final class Queues {
         static {
             // Internal property, subject to removal after validation completed.
             final String useUnpaddedQueuesName = "io.servicetalk.internal.queues.useUnpadded";
-            boolean useUnpaddedQueues = getBoolean(useUnpaddedQueuesName);
+            boolean useUnpaddedQueues = parseBoolean(getProperty(useUnpaddedQueuesName, "true"));
```
idelpivnitskiy, line 281: To minimize changes, wdyt if we keep this flag `false` for the next release and change it to `true` in release after?

Scottmitch: the flag has been false since the feature was introduced. The behavior should be the same in the `ea` and regular release, and the intention was to move default behavior to conserve memory (bcz our queues are typically small so constant queue overhead is relatively large). we can defer flipping the default if you prefer but I don't think going between `ea` and regular changes any behavior.

idelpivnitskiy: The behavior will be the same only for those who already tried to set the flag to `true` and had exactly 3.3.1-ea in the classpath. In my experience, some users don't like `ea` in their classpath and either downgraded the version to 3.3.0 or already upgraded to 4.0.1.
The safer approach will be to update the version in one release and start using the new feature by default in another release.

idelpivnitskiy: Follow up: #2490

PR #2352, opened 2022-09-09
`AsciiBuffer.indexOf` returns incorrect result for 2-bytes characters

idelpivnitskiy:
Motivation:

`AsciiBuffer.indexOf` casts `char` to `byte` by dropping the higher 8 bits. As the result, it may return an index of an ascii character that has the same lower 8 bits.

Modifications:

- Check character length and return `-1` if it's 2-bytes character;

Result:

Correct result when users try to find an index of a 2-bytes character in `AsciiBuffer`. Consistency with other `CharSequence` implementations, like `String` and `StringBuilder`.

### DIFF
--- a/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/AsciiBuffer.java
+++ b/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/AsciiBuffer.java
@@ -114,7 +114,14 @@ final class AsciiBuffer implements CharSequence {
      * @throws NullPointerException if {@code subString} is {@code null}.
      */
     int indexOf(char ch, int start) {
+        if (!singleByte(ch)) {
+            return -1;
+        }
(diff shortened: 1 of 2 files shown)


servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/AsciiBuffer.java
```diff
@@ -114,9 +114,16 @@ Buffer unwrap() {
      * @throws NullPointerException if {@code subString} is {@code null}.
      */
     int indexOf(char ch, int start) {
+        if (!singleByte(ch)) {
+            return -1;
+        }
         return buffer.indexOf(start, buffer.writerIndex(), (byte) ch);
```
Scottmitch, line 120: nit: simplify and remove negation: `return singleByte(ch) ? buffer.indexOf(start, buffer.writerIndex(), (byte) ch) : -1;`

Scottmitch (approved):
good catch!

PR #2364, opened 2022-09-14
Publisher#flatMapMerge subscribe after cancel

Scottmitch:
Motivation:
Publisher#flatMapMerge doesn't subscribe to mapped sources if it has been cancelled. This behavior is inconsistent with Publisher#flatMapMergeSingle and also may be surprising to have mapped Publishers be discarded without any subscribe and terminal event. For example there could be control flows that use `multicast` with a minimum amount of subscribers before subscribing upstream, and if flatMap doesn't subscribe the control flow will hang.

Modifications:
- Publisher#flatMapMerge should always subscribe and cancel the mapped Publishers if necessary.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
@@ -15,9 +15,7 @@
  */
 package io.servicetalk.concurrent.api;
 
-import io.servicetalk.concurrent.Cancellable;
 import io.servicetalk.concurrent.internal.ConcurrentSubscription;
-import io.servicetalk.concurrent.internal.DelayedSubscription;
 import io.servicetalk.concurrent.internal.FlowControlUtils;
 import io.servicetalk.concurrent.internal.QueueFullException;
 import io.servicetalk.concurrent.internal.SubscriberUtils;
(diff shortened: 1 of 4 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapMerge.java
```diff
@@ -198,11 +196,18 @@ public void onSubscribe(final Subscription s) {
         public void onNext(@Nullable final T t) {
             final Publisher<? extends R> publisher = requireNonNull(source.mapper.apply(t),
                     () -> "Mapper " + source.mapper + " returned null");
-            FlatMapPublisherSubscriber<T, R> subscriber = new FlatMapPublisherSubscriber<>(this);
-            if (cancellableSet.add(subscriber) && activeMappedSourcesUpdater.incrementAndGet(this) > 0) {
-                publisher.subscribeInternal(subscriber);
+            for (;;) {
+                final int currValue = this.activeMappedSources;
+                if (currValue < 0) {
+                    throw new IllegalStateException("onNext(" + t + ") after terminal signal delivered to " + this);
```
idelpivnitskiy, line 202: 1. `t` may contain sensitive user information, we should not use it as-is for string messages.
2. Consider adding a comment to clarify why we throw from `onNext` instead of delivering `onError`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherFlatMapSingle.java
```diff
@@ -186,10 +186,18 @@ public void onSubscribe(Subscription s) {
         public void onNext(T t) {
             final Single<? extends R> next = requireNonNull(source.mapper.apply(t),
                     () -> "Mapper " + source.mapper + " returned null");
-            if (activeMappedSourcesUpdater.incrementAndGet(this) > 0) {
-                next.subscribeInternal(new FlatMapSingleSubscriber());
+            for (;;) {
+                final int currValue = this.activeMappedSources;
+                if (currValue < 0) {
+                    throw new IllegalStateException("onNext(" + t + ") after terminal signal delivered to " + this);
```
idelpivnitskiy, line 192: Same comments here

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapMergeTest.java
```diff
@@ -355,6 +356,88 @@ void testDuplicateTerminal(boolean errorFirst, boolean errorSecond) {
         }
     }
 
+    @ParameterizedTest(name = "{displayName} [{index}] onError={0} delayError={1}")
+    @CsvSource(value = {/*"true,true", "true,false",*/ "false,true"/*, "false,false"*/})
```
idelpivnitskiy, line 360: Debugging leftover or intentionally skipped?

PR #2365, opened 2022-09-15
ContentEncodingHttpRequesterFilter to remove content-length

Scottmitch:
Motivation:
ContentEncodingHttpRequesterFilter will apply compression or decompression. This may change the overall payload length and if the content-length header is present it will likely be incorrect after.

Modifications:
- Remove the content-length header if encoding/decoding is applied in ContentEncodingHttpRequesterFilter. Headers are already manipulated to add/remove content-encoding.

### DIFF
--- a/servicetalk-http-api/src/test/java/io/servicetalk/http/api/ContentEncodingHttpServiceFilterTest.java
+++ b/servicetalk-http-api/src/test/java/io/servicetalk/http/api/ContentEncodingHttpServiceFilterTest.java
@@ -15,22 +15,36 @@
  */
 package io.servicetalk.http.api;
 
+import io.servicetalk.buffer.api.Buffer;
 import io.servicetalk.buffer.api.CharSequences;
+import io.servicetalk.concurrent.api.Publisher;
+import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.encoding.api.BufferDecoderGroupBuilder;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


servicetalk-http-api/src/test/java/io/servicetalk/http/api/ContentEncodingHttpServiceFilterTest.java
```diff
@@ -73,4 +87,46 @@ void testMatchAndRemoveEncodingMiddleDoesNotMatch() {
...
+        request.headers().add(CONTENT_LENGTH, String.valueOf(payloadBody.length()));
+        request.contentEncoding(gzipDefault());
+        StreamingHttpResponse response = filter.request(request).toFuture().get();
+        assertContentLength(response.headers(), response.payloadBody());
+    }
+
+    private static void assertContentLength(HttpHeaders headers, Publisher<Buffer> publisher) throws Exception {
+        CharSequence reqLen = headers.get(CONTENT_LENGTH);
+        if (reqLen != null) {
+            final int len = publisher.collect(() -> 0, (sum, buff) -> sum + buff.readableBytes()).toFuture().get();
```
idelpivnitskiy, line 128: The test never goes inside this `if` branch. Instead, consider asserting that there is not CL header bcz this is what we expect from a streaming client.

idelpivnitskiy (approved):
Makes sense!

servicetalk-http-api/src/test/java/io/servicetalk/http/api/ContentEncodingHttpServiceFilterTest.java
Scottmitch: I'll just leave as is for now ... currently we just remove the header but from the test perspective either it is not present, or it needs to be correct.

idelpivnitskiy:
Can you submit a similar PR for `ContentEncodingHttpServiceFilter` please?

Scottmitch:
https://github.com/apple/servicetalk/pull/2366

PR #2366, opened 2022-09-16
ContentEncodingHttpServiceFilter to remove content-length

Scottmitch:
Motivation:
ContentEncodingHttpServiceFilter will apply compression or decompression. This may change the overall payload length and if the content-length header is present it will likely be incorrect after.

Modifications:

- Remove the content-length header if encoding/decoding is applied in ContentEncodingHttpServiceFilter. Headers are already manipulated to add/remove content-encoding.

### DIFF
--- a/servicetalk-http-api/src/test/java/io/servicetalk/http/api/ContentEncodingHttpServiceFilterTest.java
+++ b/servicetalk-http-api/src/test/java/io/servicetalk/http/api/ContentEncodingHttpServiceFilterTest.java
@@ -17,13 +17,12 @@ package io.servicetalk.http.api;
 
 import io.servicetalk.buffer.api.Buffer;
 import io.servicetalk.buffer.api.CharSequences;
-import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.encoding.api.BufferDecoderGroupBuilder;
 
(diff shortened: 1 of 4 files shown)


servicetalk-http-api/src/test/java/io/servicetalk/http/api/ContentEncodingHttpServiceFilterTest.java
```diff
@@ -88,45 +87,35 @@ void testMatchAndRemoveEncodingMiddleDoesNotMatch() {
                 contentEqualsIgnoreCase(contentEncoding, " deflate , foo , gzip "), is(true));
     }
 
-    @ParameterizedTest(name = "{displayName} [{index}] {0}-{1}")
-    @MethodSource("requesterTypes")
-    void contentLengthRemovedOnEncode(final RequesterType type,
-                                      final SecurityType security) throws Exception {
+    @ParameterizedTest(name = "{displayName} [{index}] {0}")
+    @EnumSource(SecurityType.class)
+    void contentLengthRemovedOnEncode(final SecurityType security) throws Exception {
```
Scottmitch, line 92: diff looks ugly but I moved the `AbstractHttpRequesterFilterTest` test to a new class and kept this class focused on service filter.

idelpivnitskiy (approved):
Thank you!

PR #2367, opened 2022-09-21
Make HTTP/1.x server-side control flow consistent with the client-side

idelpivnitskiy:
Motivation:

The HTTP/1.x control flow for `NettyHttpServer` uses a single `write` operation to write an indefinite stream of request/response iterations as a repeated publisher. This difference requires all our transport utilities (e.g. `CloseHandler`, `WriteStreamSubscriber`, `FlushStrategy`) to have special behavior to accommodate the difference.

Modifications:
- Apply `repeat` operator to the returned `write` completable;
- Deprecate `SplittingFlushStrategy`, it's not required anymore;
- Temporarily introduce `ChangingFlushStrategy` to still support changes of the `FlushStrategy` per request with existing `FlushStrategy` API;
- `RequestResponseCloseHandler` emits `OutboundDataEndEvent` on every "payload end";
- Adjust existing tests to account for the new behavior;
- Enhance some tests to use different executors for client & server, it helps with debugging;

Result:

HTTP/1.x server-side control flow is consistent with the client-side.

Note that it changes the way server-side manages demand for new writes. Instead of having a single `WriteDemandEstimator` for the entire connection that remembers the size of previous requests, it now has a new `WriteDemandEstimator` per exchange, the same way as the client. If we will observe it as a problem, we can share the `WriteDemandEstimator` for the entire connection in a follow up, but we should do it consistently for client & serve sides.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -278,38 +269,7 @@ static final class NettyHttpServerConnection extends HttpServiceContext implemen
...
-                            if (itemWritten instanceof Buffer) {
-                                return contentLength > 0 &&
-                                        (contentLength -= ((Buffer) itemWritten).readableBytes()) <= 0 ?
-                                        End : InProgress;
-                            }
-                            return itemWritten instanceof HttpHeaders ? End : InProgress;
-                        }
-                    });
+            this.flushStrategy = new ChangingFlushStrategy(new FlushStrategyHolder(connection.defaultFlushStrategy()));
             connection.updateFlushStrategy((current, isCurrentOriginal) -> flushStrategy);
```
Scottmitch, line 273: >Temporarily introduce ChangingFlushStrategy to still support changes of the FlushStrategy per request with existing FlushStrategy API;

can you clarify what your plan is for removal of flush strategy on the connection level? is this planned for followup PR or to be included in this PR?

very happy to see `SplittingFlushStrategy` and related hacks go 🏅

Scottmitch (approved):
lgtm! looking forward to followup benefits from this (flush strategy api nd hooking up to blocking apis, common use of utils, ..)

PR #2370, opened 2022-09-24
DefaultBlockingHttp objects to use cast and shareContextOnSubscribe

Scottmitch:
Motivation:
DefaultBlockingStreamingHttp* objects have control flow where they use map and cast to an Object, this can use the `cast` operator. There are also defer operators used that don't need to copy AsyncContext so `shareContextOnSubscribe` can be used to avoid copying.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultBlockingStreamingHttpRequest.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultBlockingStreamingHttpRequest.java
@@ -201,8 +201,9 @@ final class DefaultBlockingStreamingHttpRequest extends AbstractDelegatingHttpRe
         original.payloadHolder().messageBody(defer(() -> {
             HttpMessageBodyIterator<Buffer> body = messageBody.iterator();
             return fromIterable(() -> body)
-                    .map(o -> (Object) o)
-                    .concat(defer(() -> from(body.trailers()).filter(Objects::nonNull)));
+                    .cast(Object.class)
+                    .concat(defer(() -> from(body.trailers()).filter(Objects::nonNull).shareContextOnSubscribe()))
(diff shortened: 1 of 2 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultBlockingStreamingHttpRequest.java
```diff
@@ -201,8 +201,9 @@ public BlockingStreamingHttpRequest messageBody(final HttpMessageBodyIterable<Bu
         original.payloadHolder().messageBody(defer(() -> {
             HttpMessageBodyIterator<Buffer> body = messageBody.iterator();
             return fromIterable(() -> body)
-                    .map(o -> (Object) o)
-                    .concat(defer(() -> from(body.trailers()).filter(Objects::nonNull)));
+                    .cast(Object.class)
+                    .concat(defer(() -> from(body.trailers()).filter(Objects::nonNull).shareContextOnSubscribe()))
```
idelpivnitskiy, line 205: For internal code, consider manual check for null and return `empty()` to reduce number of operators.
We can also consider adding `Publisher.fromNullable(...)` to simplify this use-case.

idelpivnitskiy (approved):

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultBlockingStreamingHttpRequest.java
Scottmitch: I added a package private static `fromFilterNull` method for local usage. not sure yet we want to expose composite operators in the public API but we can always promote it in a followup PR.

Scottmitch:
build failure attributed to https://github.com/apple/servicetalk/issues/2378

PR #2371, opened 2022-09-24
Add http byte[] aggregated example

Scottmitch:
Motivation:
Some times folks want to get the payload body as a `byte[]` and `ByteArraySerializer` can be used for this.

### DIFF
--- a/servicetalk-examples/http/serialization/bytes/src/main/java/io/servicetalk/examples/http/serialization/bytes/blocking/BlockingBytesClient.java
+++ b/servicetalk-examples/http/serialization/bytes/src/main/java/io/servicetalk/examples/http/serialization/bytes/blocking/BlockingBytesClient.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright © 2018 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


servicetalk-examples/http/serialization/bytes/src/main/java/io/servicetalk/examples/http/serialization/bytes/blocking/BlockingBytesClient.java
```diff
@@ -0,0 +1,34 @@
...
+
+import static io.servicetalk.examples.http.serialization.bytes.SerializerUtils.SERIALIZER;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public final class BlockingBytesClient {
+    public static void main(String[] args) throws Exception {
+        try (BlockingHttpClient client = HttpClients.forSingleAddress("localhost", 8080).buildBlocking()) {
+            HttpResponse resp = client.request(client.get("/bytes"));
+            System.out.println(resp.toString((name, value) -> value));
+            final byte[] bytes = resp.payloadBody(SERIALIZER); // Deserialize the whole payload body as a byte[].
```
Scottmitch, line 30: wanted to give some usage of `ByteArraySerializer` ... as converting to `byte[]` may not be obvious for folks.

idelpivnitskiy (approved):

PR #2372, opened 2022-09-24
Add Publisher#concatPropagateCancel(Completable) to force cancel propagation

Scottmitch:
Motivation:
In some scenarios concat is used where the right hand side has assocated state that is desirable to always trigger (concat payload body to meta data for http response processing). Regular concat will not subscribe if the Publisher terminates with onError or is cancelled which will prevent terminal event visibility.


Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/2069

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherConcatWithCompletable.java
```diff
@@ -31,20 +31,154 @@
...
+                            next.subscribeInternal(this);
+                        }
+                        break;
+                    }
+                } else if (c == CANCELLED) {
+                    // CANCELLED means downstream doesn't care about being notified and internal state tracking also
+                    // uses this state if terminal has been propagated, so avoid duplicate terminal. We may subscribe
+                    // to both sources in parallel if cancellation occurs, so allowing terminal to propagate would mean
+                    // ordering and concurrency needs to be accounted for between Publisher and completed, because
+                    // cancel allows for no more future delivery we avoid future invocation of the target subscriber.
```
idelpivnitskiy, line 93: Thanks for clarifying this. While we can ignore terminal events after cancel, we try to still deliver all originally emitted events in other operators. It's good if "deliver terminal after cancel" decision can be honored from the original source instead of modified by an operator.
IIUC, we need one more state to handle this. `cancel` needs to use CAS instead of getAndSet to flip subscription into that new state and all other states to `CANCELLED`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherConcatWithCompletable.java
```diff
@@ -31,20 +31,154 @@
...
+        private volatile Cancellable cancellable = IGNORE_CANCEL;
+
+        ConcatSubscriberCancel(Subscriber<? super T> target, Completable next) {
+            this.target = target;
+            this.next = next;
+        }
+
+        @Override
+        public void onSubscribe(Subscription s) {
+            cancellable = new FirstSubscription(s);
```
idelpivnitskiy, line 65: Can we use `instanceof Subscription` instead of wrapping, like `ConcatSubscriber` does? If not, please leave a comment in code to clarify why we need it.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -1942,9 +1942,37 @@ public final Publisher<T> concat(Single<? extends T> next) {
...
+     * successfully.
+     * @param propagateCancel if {@code true} then {@code next} will be subscribed to and cancelled if this
+     * {@link Publisher} is cancelled or terminates with {@link Subscriber#onError(Throwable)}. if {@code false} this
+     * method is equivalent to {@link #concat(Completable)}.
+     * @return A {@link Publisher} that emits all items from this {@link Publisher} and then awaits successful
+     * completion of {@code next} {@link Completable}.
+     * @see <a href="https://reactivex.io/documentation/operators/concat.html">ReactiveX concat operator.</a>
+     * @see #concat(Completable)
+     */
+    public final Publisher<T> concat(Completable next, boolean propagateCancel) {
```
idelpivnitskiy, line 1974: I'm still unsure about adding an overload vs using a new name `concatPropagateCancel`. I'm not concerned of people moving between sources. Chance someone will move from `Publisher.concat(Completable, true)` to `Single.concat(Publisher, true)` is low. More worried about people memorizing meaning of the 2nd boolean in one case and being surprised the meaning have changed for another use-case. I would be less worried if the order of booleans will be the same, but we can not change it for `Single.concat(Publisher)`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherConcatWithCompletable.java
Scottmitch: We need to differentiate between the first `onSubscribe(Subscription)` and the second `onSubscribe(Cancellable)`. someone could pass a `Cancellable` that is also a `Subscription` to the second method so we need some additional state.

Scottmitch: added comments

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
Scottmitch: discussed offline. will go with `concatPropagateCancel` for now

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletablePropagateCancelTest.java
```diff
@@ -0,0 +1,65 @@
...
+    PublisherConcatWithCompletablePropagateCancelTest() {
+        setup(true);
+    }
+
+    @Test
+    @Override
+    void sourceCancel() {
+        subscriber.awaitSubscription().cancel();
+        assertThat("Source subscription not cancelled.", subscription.isCancelled(), is(true));
+        assertThat("Next source subscribed on cancellation.", completable.isSubscribed(), is(true));
```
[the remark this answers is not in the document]
Scottmitch: refactored everything to be in the same class and parametrized.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherConcatWithCompletable.java
Scottmitch: I will add a new state to propagate here.

Scottmitch: actually this fails the tck bcz we don't necessarily know if the publish was done emitting or not. lets not do this for now.

Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/2265

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletableTest.java
```diff
@@ -78,27 +105,70 @@ void nextError() {
...
+        if (propagateCancel) {
+            completable.awaitSubscribed();
+            assertThat(cancellable.isCancelled(), is(true));
+            if (onError) {
+                completable.onError(DELIBERATE_EXCEPTION);
+            } else {
+                completable.onComplete();
+            }
+            // Cancel before the publisher completes means we don't propagate terminals.
+            assertThat(subscriber.pollTerminal(10, MILLISECONDS), is(nullValue()));
```
idelpivnitskiy, line 129: Ok, now I see the problem with propagating terminals after `cancel`. Thanks for this test!

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletableTest.java
```diff
@@ -78,27 +105,70 @@ void nextError() {
...
+            source.onError(DELIBERATE_EXCEPTION);
+        } else {
+            source.onComplete();
+        }
+        assertThat("Next source not subscribed.", completable.isSubscribed(), is(propagateCancel || !onError));
         subscriber.awaitSubscription().cancel();
-        completable.onSubscribe(cancellable);
-        assertThat("Next cancellable not cancelled.", cancellable.isCancelled(), is(true));
+        if (propagateCancel) {
+            completable.awaitSubscribed();
```
idelpivnitskiy, line 154: We already checked at line 151 that completable is subscribed

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletableTest.java
```diff
@@ -78,27 +105,70 @@ void nextError() {
...
+    @CsvSource(value = {"false,false", "false,true", "true,false", "true,true"})
+    void nextCancel(boolean propagateCancel, boolean onError) {
+        setup(propagateCancel);
+        if (onError) {
+            source.onError(DELIBERATE_EXCEPTION);
+        } else {
+            source.onComplete();
+        }
+        assertThat("Next source not subscribed.", completable.isSubscribed(), is(propagateCancel || !onError));
         subscriber.awaitSubscription().cancel();
```
idelpivnitskiy, line 152: After this line, consider adding:
```
assertThat("Source subscription unexpectedly cancelled.", subscription.isCancelled(), is(false));
```

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletableTest.java
```diff
@@ -78,27 +105,70 @@ void nextError() {
...
-        completable.onSubscribe(cancellable);
-        assertThat("Next cancellable not cancelled.", cancellable.isCancelled(), is(true));
+        if (propagateCancel) {
+            completable.awaitSubscribed();
+            assertThat("Next cancellable not cancelled.", cancellable.isCancelled(), is(true));
+        } else {
+            if (!onError) {
+                completable.awaitSubscribed();
```
idelpivnitskiy, line 158: This is already checked at line 151

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherConcatWithCompletable.java
```diff
@@ -31,20 +31,163 @@
...
+                }
+            }
+        }
+
+        /**
+         * We need to differentiate between the first {@link Subscriber#onSubscribe(Subscription)} and the second
+         * {@link CompletableSource.Subscriber#onSubscribe(Cancellable)}. This unique type allows distinguishing
+         * from a {@link Cancellable} that is also a {@link Subscription} passed to the second method.
+         */
+        private static final class FirstSubscription implements Subscription {
```
idelpivnitskiy, line 168: Thanks for clarifying. Do we need the same trick for `ConcatSubscriber`? It may have the same issue

servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/PublisherConcatWithCompletableCancelTckTest.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package io.servicetalk.concurrent.reactivestreams.tck;
+
+import io.servicetalk.concurrent.api.Completable;
+import io.servicetalk.concurrent.api.Publisher;
+
+import org.testng.annotations.Test;
+
+@Test
+public class PublisherConcatWithCompletableCancelTckTest extends AbstractPublisherOperatorTckTest<Integer> {
```
idelpivnitskiy, line 24: PublisherConcatWithCompletableCancelTckTest -> PublisherConcatWithCompletablePropagateCancelTckTest ?

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletableTest.java
```diff
@@ -78,27 +105,70 @@ void nextError() {
...
+    @ParameterizedTest(name = "{displayName} [{index}] propagateCancel={0} onError={1}")
+    @CsvSource(value = {"false,false", "false,true", "true,false", "true,true"})
+    void nextCancel(boolean propagateCancel, boolean onError) {
+        setup(propagateCancel);
+        if (onError) {
+            source.onError(DELIBERATE_EXCEPTION);
+        } else {
+            source.onComplete();
+        }
+        assertThat("Next source not subscribed.", completable.isSubscribed(), is(propagateCancel || !onError));
```
idelpivnitskiy, line 151: The message will be incorrect when `is` results in `false` and condition fails.

idelpivnitskiy (approved):
My comments related to tests are in this patch:
```patch
Index: servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletableTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletableTest.java b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletableTest.java
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletableTest.java	(revision bcad9a44fec7c69744d6b54edd87bb5350c3a1b3)
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletableTest.java	(date 1664812054953)
@@ -148,16 +148,16 @@
         } else {
             source.onComplete();
         }
-        assertThat("Next source not subscribed.", completable.isSubscribed(), is(propagateCancel || !onError));
-        subscriber.awaitSubscription().cancel();
-        if (propagateCancel) {
-            completable.awaitSubscribed();
+        if (propagateCancel || !onError) {
+            assertThat("Next source not subscribed.", completable.isSubscribed(), is(true));
+            subscriber.awaitSubscription().cancel();
+            assertThat("Source subscription unexpectedly cancelled.", subscription.isCancelled(), is(false));
             assertThat("Next cancellable not cancelled.", cancellable.isCancelled(), is(true));
         } else {
-            if (!onError) {
-                completable.awaitSubscribed();
-            }
-            assertThat("Next cancellable not cancelled.", cancellable.isCancelled(), is(!onError));
+            assertThat("Next source unexpectedly subscribed.", completable.isSubscribed(), is(false));
+            subscriber.awaitSubscription().cancel();
+            assertThat("Source subscription not cancelled.", subscription.isCancelled(), is(true));
+            assertThat("Next cancellable unexpectedly cancelled.", cancellable.isCancelled(), is(false));
         }
 
         if (onError) {
Index: servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/PublisherConcatWithCompletableCancelTckTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/PublisherConcatWithCompletableCancelTckTest.java b/servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/PublisherConcatWithCompletablePropagateCancelTckTest.java
rename from servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/PublisherConcatWithCompletableCancelTckTest.java
rename to servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/PublisherConcatWithCompletablePropagateCancelTckTest.java
--- a/servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/PublisherConcatWithCompletableCancelTckTest.java	(revision bcad9a44fec7c69744d6b54edd87bb5350c3a1b3)
+++ b/servicetalk-concurrent-reactivestreams/src/test/java/io/servicetalk/concurrent/reactivestreams/tck/PublisherConcatWithCompletablePropagateCancelTckTest.java	(date 1664810883896)
@@ -21,7 +21,7 @@
 import org.testng.annotations.Test;
 
 @Test
-public class PublisherConcatWithCompletableCancelTckTest extends AbstractPublisherOperatorTckTest<Integer> {
+public class PublisherConcatWithCompletablePropagateCancelTckTest extends AbstractPublisherOperatorTckTest<Integer> {
     @Override
     protected Publisher<Integer> composePublisher(Publisher<Integer> publisher, int elements) {
         return publisher.concatPropagateCancel(Completable.completed());
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherConcatWithCompletable.java
Scottmitch: We could use the same trick but it isn't strictly necessary bcz we don't need to distinguish in terminal events if we have transitioned to the second subscription (bcz we don't transition until the first one completes).

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/PublisherConcatWithCompletableTest.java
Scottmitch: I applied your diff above, thx!

PR #2377, opened 2022-09-29
Enhance testing of the server-side control flow

idelpivnitskiy:
Motivation:

Cover a few more use-cases to increase confidence in server-side control flow behavior.

Modifications:

- Add `ServerPipelineControlFlowTest`;
- Add utilities for easier setup of debugging options for client/server builder;

Result:

More tests for the server-side control flow.


idelpivnitskiy:
This is just a test, need it to enhance confidence in #2367 changes. Merging.

PR #2379, opened 2022-09-29
Expand allowed types to accommodate custom service discoverer events

mgodave:
Motivation:

The types in the client builders are too restrictive and prevent us from using
them with custom ServiceDiscoverers with custom ServiceDiscovererEvent
implementations.

Changes:

Open up the types so we can use these APIs with a larger selection of custom
ServiceDiscoverers. This change is not breaking.


idelpivnitskiy (approved):
Optionally, we can add a test to verify a custom SD event can be specified. No need to focus on a test right now, we can make a follow-up later.

PR #2380, opened 2022-09-30
NettyCompression gzip streaming decompression over multiple buffers fix

Scottmitch:
Motivation:
NettyCompression's gzip streaming decompression fails to decompress
payloads when split over multiple buffers and may result in:

```
DecompressionException: Input is not in the GZIP format
```

Modifications:
- NettyCompressionStreamingSerializer skips bytes unconditionally after
  they are feed to the Netty decoders. However in some scenarios not all
  bytes maybe consumed, and if we skip the bytes the aggregation maintained
  by Netty's ByteToMessageDecoder will be discarded. Instead of skipping
  unconditionally we should only advance the reader index if the ByteBuf
  and Buffer don't already match.

### DIFF
--- a/servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionSerializer.java
+++ b/servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionSerializer.java
@@ -56,9 +56,7 @@ final class NettyCompressionSerializer implements SerializerDeserializer<Buffer>
         final MessageToByteEncoder<ByteBuf> encoder = encoderSupplier.get();
         final EmbeddedChannel channel = newEmbeddedChannel(encoder, allocator);
         try {
-            channel.writeOutbound(extractByteBufOrCreate(toSerialize));
-            toSerialize.skipBytes(toSerialize.readableBytes());
-
+            writeAndUpdateIndex(channel, toSerialize, false);
             // May produce footer
             preparePendingData(channel);
             drainChannelQueueToSingleBuffer(channel.outboundMessages(), nettyDst);
@@ -84,9 +82,7 @@ final class NettyCompressionSerializer implements SerializerDeserializer<Buffer>
         final ByteToMessageDecoder decoder = decoderSupplier.get();
         final EmbeddedChannel channel = newEmbeddedChannel(decoder, allocator);
         try {
-            channel.writeInbound(toByteBuf(serializedData));
-            serializedData.skipBytes(serializedData.readableBytes());
-
+            writeAndUpdateIndex(channel, serializedData, true);
             drainChannelQueueToSingleBuffer(channel.inboundMessages(), nettyDst);
             // no need to advance writerIndex -> NettyBuffer's writerIndex reflects the underlying ByteBuf value.
             cleanup(channel);
@@ -94,6 +90,21 @@ final class NettyCompressionSerializer implements SerializerDeserializer<Buffer>
         } catch (Throwable e) {
             safeCleanup(channel);
             throw new BufferEncodingException("Unexpected exception during decoding", e);
+        }
+    }
+
+    static void writeAndUpdateIndex(EmbeddedChannel channel, Buffer toSerialize, boolean inbound) {
+        ByteBuf byteBuf = extractByteBufOrCreate(toSerialize);
+        final int beforeReadableBytes = byteBuf.readableBytes();
+        if (inbound) {
+            channel.writeInbound(byteBuf);
+        } else {
+            channel.writeOutbound(byteBuf);
+        }
+        // extractByteBufOrCreate may have to copy if it isn't able to unwrap NettyBuffer and in this case we have to
+        // manually advance the Buffer indexes to reflect what was consumed.
+        if (byteBuf.readableBytes() != toSerialize.readableBytes()) {
+            toSerialize.skipBytes(byteBuf.readableBytes() - beforeReadableBytes);
         }
     }
 
--- a/servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
+++ b/servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
@@ -36,13 +36,13 @@ import java.util.function.Supplier;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.buffer.api.ReadOnlyBufferAllocators.DEFAULT_RO_ALLOCATOR;
-import static io.servicetalk.buffer.netty.BufferUtils.extractByteBufOrCreate;
 import static io.servicetalk.buffer.netty.BufferUtils.getByteBufAllocator;
 import static io.servicetalk.buffer.netty.BufferUtils.newBufferFrom;
 import static io.servicetalk.concurrent.api.Single.succeeded;
 import static io.servicetalk.encoding.netty.NettyCompressionSerializer.cleanup;
 import static io.servicetalk.encoding.netty.NettyCompressionSerializer.preparePendingData;
 import static io.servicetalk.encoding.netty.NettyCompressionSerializer.safeCleanup;
+import static io.servicetalk.encoding.netty.NettyCompressionSerializer.writeAndUpdateIndex;
 import static java.util.Objects.requireNonNull;
 
 final class NettyCompressionStreamingSerializer implements StreamingSerializerDeserializer<Buffer> {
@@ -80,8 +80,7 @@ final class NettyCompressionStreamingSerializer implements StreamingSerializerDe
                 }
 
                 try { // onNext will produce AT-MOST N items (as received)
-                    channel.writeInbound(extractByteBufOrCreate(next));
-                    next.skipBytes(next.readableBytes());
+                    writeAndUpdateIndex(channel, next, true);
                     Buffer buffer = drainChannelQueueToSingleBuffer(channel.inboundMessages(), allocator);
                     if (buffer != null && buffer.readableBytes() > 0) {
                         subscriber.onNext(buffer);
@@ -148,8 +147,7 @@ final class NettyCompressionStreamingSerializer implements StreamingSerializerDe
                                     subscriber.onNext(buffer);
                                 }
                             } else {
-                                channel.writeOutbound(extractByteBufOrCreate(next));
-                                next.skipBytes(next.readableBytes());
+                                writeAndUpdateIndex(channel, next, false);
                                 Buffer buffer = drainChannelQueueToSingleBuffer(channel.outboundMessages(), allocator);
                                 if (buffer != null && buffer.readableBytes() > 0) {
                                     subscriber.onNext(buffer);
--- a/servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializerTest.java
+++ b/servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializerTest.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright © 2022 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.encoding.netty;
+
+import io.servicetalk.buffer.api.Buffer;
+import io.servicetalk.buffer.api.BufferAllocator;
+import io.servicetalk.buffer.api.CompositeBuffer;
+import io.servicetalk.concurrent.api.Publisher;
+import io.servicetalk.serializer.api.SerializerDeserializer;
+import io.servicetalk.serializer.api.StreamingSerializerDeserializer;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+
+import java.util.concurrent.ThreadLocalRandom;
+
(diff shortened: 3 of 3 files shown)


Scottmitch:
thx @colestanfield for providing a reproducer!

servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializerTest.java
```diff
@@ -0,0 +1,76 @@
...
+        StreamingSerializerDeserializer<Buffer> serializer = type.serializer;
+        String rawString = "hello";
+        byte[] rawBytes = new byte[1024];
+        ThreadLocalRandom.current().nextBytes(rawBytes);
+        Buffer[] clearText = new Buffer[] {allocator.fromAscii(rawString), allocator.wrap(rawBytes)};
+        Publisher<Buffer> serializedPub = serializer.serialize(from(clearText), allocator);
+
+        CompositeBuffer serializedBuf = serializedPub.collect(allocator::newCompositeBuffer, (comp, buf) -> {
+            comp.addBuffer(buf);
+            return comp;
```
mgodave, line 62: nit: I think `addBuffer` returns `this` (looking at the ST buffer API) so the block and return and unnecessary.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
```diff
@@ -80,8 +83,21 @@ public void onNext(@Nullable final Buffer next) {
                 }
 
                 try { // onNext will produce AT-MOST N items (as received)
-                    channel.writeInbound(extractByteBufOrCreate(next));
-                    next.skipBytes(next.readableBytes());
+                    final Buffer bufferToDecode;
+                    if (aggregate != null && aggregate.readableBytes() != 0) {
+                        aggregateNext(next);
+                        bufferToDecode = aggregate;
+                    } else {
```
mgodave, line 90: It was unclear from the test if this branch is being tested.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
```diff
@@ -80,8 +83,21 @@ public void onNext(@Nullable final Buffer next) {
...
+                        bufferToDecode = next;
+                    }
+                    ByteBuf byteBuf = toByteBuf(bufferToDecode);
+                    channel.writeInbound(byteBuf);
+                    assert byteBuf.readableBytes() == bufferToDecode.readableBytes();
+                    if (bufferToDecode.readableBytes() > 0) {
+                        aggregateRemainder(bufferToDecode);
+                    } else {
```
mgodave, line 98: `aggregate == null` is an interesting case to make sure gets tested.

mgodave (approved):

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
```diff
@@ -110,6 +126,34 @@ public void onComplete() {
 
                 subscriber.onComplete();
             }
+
+            private void aggregateNext(Buffer next) {
+                assert aggregate != null;
+                if (!aggregate.isReadOnly() && aggregate.maxWritableBytes() >= next.readableBytes()) {
+                    aggregate.writeBytes(next);
+                } else {
+                    Buffer newAgg = allocator.newBuffer(aggregate.readableBytes() + next.readableBytes(),
```
idelpivnitskiy, line 135: This can overflow

servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializerTest.java
```diff
@@ -0,0 +1,76 @@
...
+        GZIP(gzipDefaultStreaming()), DEFLATE(deflateDefaultStreaming());
+
+        final StreamingSerializerDeserializer<Buffer> serializer;
+
+        StreamingType(StreamingSerializerDeserializer<Buffer> serializer) {
+            this.serializer = serializer;
+        }
+    }
+
+    @ParameterizedTest(name = "{index}: flushStrategy = {0}")
```
idelpivnitskiy, line 49: flushStrategy -> algorithm?

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
```diff
@@ -80,8 +83,21 @@ public void onNext(@Nullable final Buffer next) {
                 }
 
                 try { // onNext will produce AT-MOST N items (as received)
-                    channel.writeInbound(extractByteBufOrCreate(next));
-                    next.skipBytes(next.readableBytes());
+                    final Buffer bufferToDecode;
+                    if (aggregate != null && aggregate.readableBytes() != 0) {
+                        aggregateNext(next);
+                        bufferToDecode = aggregate;
```
idelpivnitskiy, line 89: Consider chaning `aggregateNext` signature to return `Buffer`. Then we can do `bufferToDecode = aggregateNext(next);`

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
```diff
@@ -80,8 +83,21 @@ public void onNext(@Nullable final Buffer next) {
...
+                    } else {
+                        bufferToDecode = next;
+                    }
+                    ByteBuf byteBuf = toByteBuf(bufferToDecode);
```
idelpivnitskiy, line 93: Consider keeping pre-existing `extractByteBufOrCreate` to make sure we can work with `ReadOnlyByteBuffer`(s) too. Would be great to have a test.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
idelpivnitskiy: +1 on increasing test coverage. There are missing branches in `aggregateNext`, `aggregateRemainder`, and `drainChannelQueueToSingleBuffer`.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
```diff
@@ -110,6 +126,34 @@ public void onComplete() {
...
                 subscriber.onComplete();
             }
+
+            private void aggregateNext(Buffer next) {
+                assert aggregate != null;
+                if (!aggregate.isReadOnly() && aggregate.maxWritableBytes() >= next.readableBytes()) {
+                    aggregate.writeBytes(next);
+                } else {
+                    Buffer newAgg = allocator.newBuffer(aggregate.readableBytes() + next.readableBytes(),
+                            Integer.MAX_VALUE);
```
idelpivnitskiy, line 136: This is equal to the default `maxCapacity`, consider using an overload without 2nd argument.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
```diff
@@ -110,6 +126,34 @@ public void onComplete() {
...
+
+            private void aggregateRemainder(Buffer bufferToDecode) {
+                if (aggregate == null || aggregate.readableBytes() == 0) {
+                    aggregate = bufferToDecode;
+                } else if (!aggregate.isReadOnly() &&
+                        aggregate.maxWritableBytes() >= bufferToDecode.readableBytes()) {
+                    aggregate.writeBytes(bufferToDecode);
+                } else {
+                    Buffer newAgg = allocator.newBuffer(
+                            aggregate.readableBytes() + bufferToDecode.readableBytes(), Integer.MAX_VALUE);
```
idelpivnitskiy, line 151: Same comments here

servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializerTest.java
Scottmitch: good catch. I'll change to map to parameter name `type`.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
Scottmitch: yes. we need to feed Netty a `ByteBuf` and since Netty doesn't queue data during the decode (which is why this PR is required) the most pending data we can feed is a single `ByteBuf`. So either we use `CompositeByteBuf` for aggregation or a single buffer with max of 2gb single chunks. Since this is streaming decoder the assumption is chunks can be smaller sizes to avoid overflow.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
Scottmitch: unless its guaranteed by the interface I will keep it (I don't think we need to guarantee this in the interface) bcz this use case is specifically using the buffer for aggregation

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializer.java
idelpivnitskiy: agreed, it's ok to use a single buffer with max 2gb capacity. consider handling this use-case anyway to throw a more informative exception (`BufferEncodingException`?) instead of `IllegalArgumentException` that will come from allocator.

servicetalk-encoding-netty/src/main/java/io/servicetalk/encoding/netty/NettyCompressionSerializer.java
```diff
@@ -97,6 +93,21 @@ public Buffer deserialize(final Buffer serializedData, final BufferAllocator all
...
+    static void writeAndUpdateIndex(EmbeddedChannel channel, Buffer toSerialize, boolean inbound) {
+        ByteBuf byteBuf = extractByteBufOrCreate(toSerialize);
+        final int beforeReadableBytes = byteBuf.readableBytes();
+        if (inbound) {
+            channel.writeInbound(byteBuf);
+        } else {
+            channel.writeOutbound(byteBuf);
+        }
+        // extractByteBufOrCreate may have to copy if it isn't able to unwrap NettyBuffer and in this case we have to
+        // manually advance the Buffer indexes to reflect what was consumed.
```
idelpivnitskiy, line 105: 🎉

servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializerTest.java
```diff
@@ -0,0 +1,121 @@
...
+        final SerializerDeserializer<Buffer> serializer;
+        final BufferAllocator allocator;
+
+        AggType(SerializerDeserializer<Buffer> serializer, BufferAllocator allocator) {
+            this.serializer = serializer;
+            this.allocator = allocator;
+        }
+    }
+
+    @ParameterizedTest(name = "{index}: type = {0}")
```
idelpivnitskiy, line 73:
```suggestion
    @ParameterizedTest(name = "{displayName} [{index}] type = {0}")
```
For Intellij IDEA `displayName` is not important, but in jUnit HTML report that will be published on CI the name won't be included without it.

servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializerTest.java
```diff
@@ -0,0 +1,121 @@
...
+                        allocator.wrap(readAndCopy(serializedBuf, serializedBuf.readableBytes()))
+                ), DEFAULT_ALLOCATOR)
+                .collect(DEFAULT_ALLOCATOR::newCompositeBuffer, CompositeBuffer::addBuffer).toFuture().get();
+
+        assertThat(result.readableBytes(), equalTo(rawString.length() + rawBytes.length));
+        assertThat(result.readBytes(rawString.length()).toString(UTF_8), equalTo(rawString));
+        assertThat(result, equalTo(allocator.wrap(rawBytes)));
+    }
+
+    @ParameterizedTest(name = "{index}: type = {0}")
```
idelpivnitskiy, line 101:
```suggestion
    @ParameterizedTest(name = "{displayName} [{index}] type = {0}")
```

idelpivnitskiy (approved):
🚢

servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializerTest.java
```diff
@@ -0,0 +1,121 @@
...
+        String rawString = "hello";
+        byte[] rawBytes = new byte[1024];
+        ThreadLocalRandom.current().nextBytes(rawBytes);
+        Buffer[] clearText = new Buffer[] {allocator.fromAscii(rawString), allocator.wrap(rawBytes)};
+        Publisher<Buffer> serializedPub = serializer.serialize(from(clearText), DEFAULT_ALLOCATOR);
+
+        CompositeBuffer serializedBuf = serializedPub
+                .collect(DEFAULT_ALLOCATOR::newCompositeBuffer, CompositeBuffer::addBuffer).toFuture().get();
+
+
```
idelpivnitskiy, line 87:
```suggestion

```

```
Error: eckstyle] [ERROR] /home/runner/work/servicetalk/servicetalk/servicetalk-encoding-netty/src/test/java/io/servicetalk/encoding/netty/NettyCompressionStreamingSerializerTest.java:85:109: There is more than 1 empty line after this line. [EmptyLineSeparator]
```

tkountis (approved):
lgtm nice finding

PR #2381, opened 2022-09-30
Single#concatPropagateCancel(Publisher) to force cancel propagation

Scottmitch:
Motivation:
In some scenarios concat is used where the right hand side has assocated state that is desirable to always trigger (concat payload body to meta data for http response processing). Regular concat will not subscribe if the Single terminates with onError or is cancelled which will prevent terminal event visibility.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
@@ -766,7 +766,7 @@ public abstract class Single<T> {
      * @see #concat(Publisher, boolean)
      */
     public final Publisher<T> concat(Publisher<? extends T> next) {
-        return new SingleConcatWithPublisher<>(this, next, false);
+        return concat(next, false);
     }
 
     /**
@@ -793,7 +793,29 @@ public abstract class Single<T> {
      * all elements from {@code next} {@link Publisher}.
      */
     public final Publisher<T> concat(Publisher<? extends T> next, boolean deferSubscribe) {
-        return new SingleConcatWithPublisher<>(this, next, deferSubscribe);
+        return new SingleConcatWithPublisher<>(this, next, deferSubscribe, false);
+    }
+
+    /**
+     * Returns a {@link Publisher} that first emits the result of this {@link Single} and then subscribes and emits all
+     * elements from {@code next} {@link Publisher}. Any error emitted by this {@link Single} or {@code next}
+     * {@link Publisher} is forwarded to the returned {@link Publisher}.
+     * <p>
+     * This method provides a means to sequence the execution of two asynchronous sources and in sequential programming
+     * is similar to:
+     * <pre>{@code
+     *     List<T> results = new ...;
+     *     results.add(resultOfThisSingle());
+     *     results.addAll(nextStream());
+     *     return results;
+     * }</pre>
+     * @param next {@link Publisher} to concat. will be subscribed to and cancelled if this {@link Publisher} is
+     * cancelled or terminates with {@link Subscriber#onError(Throwable)}.
+     * @return New {@link Publisher} that first emits the result of this {@link Single} and then subscribes and emits
+     * all elements from {@code next} {@link Publisher}.
+     */
+    public final Publisher<T> concatPropagateCancel(Publisher<? extends T> next) {
+        return new SingleConcatWithPublisher<>(this, next, false, true);
     }
 
     /**
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
@@ -764,6 +764,7 @@ public abstract class Single<T> {
      * @return New {@link Publisher} that first emits the result of this {@link Single} and then subscribes and emits
      * all elements from {@code next} {@link Publisher}.
      * @see #concat(Publisher, boolean)
+     * @see #concatPropagateCancel(Publisher)
      */
     public final Publisher<T> concat(Publisher<? extends T> next) {
         return concat(next, false);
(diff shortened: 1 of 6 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -793,7 +793,37 @@ public final Publisher<T> concat(Publisher<? extends T> next) {
...
+     * {@code next} {@link Publisher} does not or might not support multiple subscribers (non-replayable). Choosing the
+     * immediate subscribe ({@code false}) behavior may have better performance and may be a preferable choice for
+     * replayable {@link Publisher}(s) or when eager subscribe is beneficial.
+     * @param propagateCancel if {@code true} then {@code next} will be subscribed to and cancelled if this
+     * {@link Publisher} is cancelled or terminates with {@link Subscriber#onError(Throwable)}. if {@code false} this
+     * method is equivalent to {@link #concat(Publisher, boolean)}.
+     * @return New {@link Publisher} that first emits the result of this {@link Single} and then subscribes and emits
+     * all elements from {@code next} {@link Publisher}.
+     */
+    public final Publisher<T> concat(Publisher<? extends T> next, boolean deferSubscribe, boolean propagateCancel) {
```
idelpivnitskiy, line 825: The `deferSubscribe` and `propagateCancel` look conflicting. If users use `deferSubscribe == true` they do not expect subscribe to the `next` without demand and success for the `Single`. Javadoc for `deferSubscribe` describes a use-case with non-replayable publisher. Do you have a legit use-case in mind for `deferSubscribe == true && propagateCancel == true`?

My expectation was that we will modify only `deferSubscribe == false` variant to make it consistent with all other `concat` variants: it should subscribe to the `next` after `onSuccess` even if `cancel` was already observed. Because for our use-case we use `Single.succeeded`, it will be enough. No need for extra complexity.

Scottmitch: These two options are independent and augment the operator behavior in a non-mutually exclusive way. `deferSubscribe` applies to the happy path, `propagateCancel` applies to the not-happy path. In the specific call site we have today in https://github.com/apple/servicetalk/pull/2369 we are using `Single.succeeded` but the semantics we desire from `concat` are provided by `propagateCancel`, so I think it is safer to provide these semantics in the operator rather than rely upon the `Single` implementation (e.g. what if we change it to respect cancel in the future). This also allows us to provide consistent semantics with https://github.com/apple/servicetalk/pull/2372 which is required at our current call-site.

Scottmitch: discussed offline, we can do both and I will rename operator to `concatPropagateCancel`.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/single/SingleConcatWithPublisherTest.java
```diff
@@ -60,7 +62,12 @@ class SingleConcatWithPublisherTest {
     private TestPublisher<Integer> next = new TestPublisher.Builder<Integer>().disableAutoOnSubscribe().build();
 
     void setUp(boolean deferSubscribe) {
-        toSource(source.concat(next, deferSubscribe)).subscribe(subscriber);
+        setUp(deferSubscribe, false);
+    }
+
+    void setUp(boolean deferSubscribe, boolean propagateCancel) {
+        toSource(propagateCancel ? source.concatPropagateCancel(next) : source.concat(next, deferSubscribe))
```
idelpivnitskiy, line 69: Tests will be a bit misleading. They assume 4 combinations, but we have only 3 combinations. Let's try to avoid testing with `true, true` as it's the same as `false, true`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
```diff
@@ -120,36 +192,65 @@ final void superCancel() {
...
+            delayedSubscription(EMPTY_SUBSCRIPTION);
+        }
     }
 
     private static final class ConcatSubscriber<T> extends AbstractConcatSubscriber<T> {
         /**
          * If {@link #request(long)} (with a valid n) invoked before {@link #onSuccess(Object)}.
          */
         private static final Object REQUESTED = new Object();
+        private static final Object REQUESTED_SUBSCRIBED = new Object();
```
idelpivnitskiy, line 230: Please add javadoc to describe this state. All other states have description

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
```diff
@@ -90,28 +96,94 @@ public final void onSubscribe(final Subscription subscription) {
...
+        private void onErrorPropagateCancel(Throwable t) {
+            for (;;) {
+                final Object oldValue = mayBeResult;
+                if (oldValue == CANCELLED) {
+                    // CANCELLED means downstream doesn't care about being notified and internal state tracking also
+                    // uses this state if terminal has been propagated, so avoid duplicate terminal. We may
+                    // subscribe to both sources in parallel if cancellation occurs, so allowing terminal to
+                    // propagate would mean ordering and concurrency needs to be accounted for between Single and
+                    // Publisher, because cancel allows for no more future delivery we avoid future invocation of
+                    // the target subscriber.
```
idelpivnitskiy, line 122: This is similar to https://github.com/apple/servicetalk/pull/2372#discussion_r984911440, but this operator has a lot more states. Will it be possible to make behavior consistent?

Scottmitch: we can propagate error as long as `propagateCancel == false` to avoid concurrency and ordering issues.

idelpivnitskiy:
Thank you!

idelpivnitskiy (approved):

PR #2382, opened 2022-09-30
Add regression tests for netty client factories

mgodave:
Motivation:

We recently expanded the allowed types for the Netty client factories and in
order to prevent a regression we should make sure they are used in a test and
fail the build if the types are narrowed.

Changes:

Create two simple tests that will fail to compile of a regression occurs in the
scope of the allowed types.


idelpivnitskiy (approved):
🚢

PR #2383, opened 2022-09-30
Make `Single.concat(Publisher)` behavior consistent on cancel

idelpivnitskiy:
Motivation:

Behavior of `Single.concat(Publisher)` is different compare to all other `concat` variants: `Single.concat(Completable)`,
`Single.concat(Single)`, `Completable.concat(Completable), `Completable.concat(Single)`, `Completable.concat(Publisher)`, `Publisher.concat(Completable)`, `Publisher.concat(Single)`, `Publisher.concat(Publisher)`. It does not subscribe to the next source to propagate cancellation if `onSuccess` is delivered after `cancel`.

Modifications:

- Make tests for all `concat` valiants consistent in regards to cancellation;
- Modify `Single.concat(Publisher)` behavior to subscribe in case `onSuccess` is delivered after `cancel`;

Result:

Behavior of `Single.concat(Publisher)` is consistent with all other `concat` variants.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
@@ -150,7 +150,11 @@ final class SingleConcatWithPublisher<T> extends AbstractNoHandleSubscribePublis
                         next.subscribeInternal(this);
                     }
                     break;
-                } else if (oldValue == CANCELLED || mayBeResultUpdater.compareAndSet(this, INITIAL, result)) {
+                } else if (oldValue == CANCELLED) {
+                    // Subscribe to the next source to propagate cancellation
+                    next.subscribeInternal(this);
+                    break;
+                } else if (mayBeResultUpdater.compareAndSet(this, INITIAL, result)) {
                     break;
                 }
             }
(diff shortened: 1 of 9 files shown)


idelpivnitskiy:
Decided to open a PR now as it shouldn't cause merge conflicts with #2381.

idelpivnitskiy:
@Scottmitch rebased after your changes, ready for review

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SingleConcatWithPublisher.java
```diff
@@ -142,7 +144,7 @@ private void onErrorPropagateCancel(Throwable t) {
 
         @Override
         public final void onComplete() {
-            if (propagateCancel) {
+            if (propagateCancel || !deferSubscribe()) {
```
Scottmitch, line 147: I wasn't expecting the changes described in this PR to be made here. for example I was expecting the `SingleSubscriber.onSuccess(..)` methods such that we subscribe eve when completed if `CANCELLED` (perhaps conditionally?)

thing to watch out for is if `propagateCancel` we may subscribe to both sources concurrently, and `mayBeResultUpdater` is used to ensure we only terminate a single time. 

can we make the termination probation (on error / on complete) is consistent.

Scottmitch:
@idelpivnitskiy - should we close this or do you still think we need this?

PR #2384, opened 2022-10-01
[wip] PublisherAsBlockingIterable LinkedBlockingQueue -> SpscBlockingQueue

Scottmitch:


Scottmitch:
> make client-flame TEST=grpc-st-blocking METHOD=StrAgg PLSIZE=medium NO_OFFLOADS=false LOADCONN=1024 LOADTHREAD=1024 JDK=11 REFLECTION=false GC=Parallel

no real benefits observed on the above profile ... more analysis needed...

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherAsBlockingIterable.java
```diff
@@ -54,7 +61,7 @@
     private final int queueCapacityHint;
 
     PublisherAsBlockingIterable(final Publisher<T> original) {
-        this(original, 16);
+        this(original, 8);
```
Scottmitch, line 64: current min forced by JCTools is 16 anyways, but will consider followup contribution to allow for smaller initial queue size.

[1] https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/SpscUnboundedArrayQueue.java#L32

Scottmitch:
LinkedTransferQueue also does `Thread.yield` and provides similar results for yield_count == 1 ... lets start with this queue first: https://github.com/apple/servicetalk/pull/2386

PR #2385, opened 2022-10-04
Implement `toString()` for `TestPublisherSubscriber#pollTerminal(...)`

idelpivnitskiy:
Motivation:

We have assertions that expect no terminal signals. When they fail, hamcrest prints a lambda name that is hard to understand.

Modifications:

- Implement `toString()` for the returned `Supplier` from `TestPublisherSubscriber#pollTerminal(...)`;

Result:

Easier to understand what was the terminal signal when assertion fails.


idelpivnitskiy:
Impacts only tests, improves debuggability. Merging

PR #2388, opened 2022-10-05
`Single.concatDeferSubscribe(Publisher)` instead of `Single.concat(Publisher, boolean)`

idelpivnitskiy:
Motivation:

Boolean arguments for operators are harder to read, especially when number of possible boolean args is growing.

Modifications:

- Introduce `Single.concatDeferSubscribe(Publisher)` that behaves like `Single.concat(Publisher, true)`;
- Deprecate `Single.concat(Publisher, boolean)`;

Result:

Clear behavior based on operator name, consistency with `concatPropagateCancel`.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
@@ -748,8 +748,7 @@ public abstract class Single<T> {
      * elements from {@code next} {@link Publisher}. Any error emitted by this {@link Single} or {@code next}
      * {@link Publisher} is forwarded to the returned {@link Publisher}.
      * <p>
-     * Note: this method is an overload for {@link #concat(Publisher, boolean)} with {@code deferSubscribe} equal to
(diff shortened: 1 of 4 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -792,13 +791,41 @@ public final Publisher<T> concat(Publisher<? extends T> next) {
      * replayable {@link Publisher}(s) or when eager subscribe is beneficial.
      * @return New {@link Publisher} that first emits the result of this {@link Single} and then subscribes and emits
      * all elements from {@code next} {@link Publisher}.
+     * @deprecated Use {@link #concat(Publisher)} or {@link #concatDeferSubscribe(Publisher)} instead.
      */
+    @Deprecated
```
Scottmitch, line 796: I'm mixed on this... imho using method overload language feature avoids callers needing to have conditional code if they need to switch between different options. however we have both variants now (qualified name and method overload) so I'm not opposed to this.

PR #2389, opened 2022-10-07
`Buffer.[set|write]Bytes` doesn't respect length when reading from `InputStream`

idelpivnitskiy:
Motivation:

Javadoc for `Buffer.writeBytes` and `Buffer.setBytes` says it reads "length bytes have been read, the end of stream is reached, or an exception is thrown". However, its internal implementation in Netty does only a single `InputStream.read(byte[], int, int)` invocation. `InputStream` contract does not guarantee to read the whole requested `length`, is says "reads up to len bytes of data from the input stream". There are implementations that may return less than `len` bytes on a single read attempt.

Modifications:

- `NettyBuffer` keeps reading from the `InputStream` until length bytes have been read;
- Add tests to demonstrate a problem;
- Also add tests for `writeBytesUntilEndStream` bcz it's a custom method that does not exist in Netty;

Result:

`Buffer.writeBytes` and `Buffer.setBytes` behave according to their javadoc when reading from `InputStream`.

### DIFF
--- a/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/Buffer.java
+++ b/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/Buffer.java
@@ -924,7 +924,7 @@ public interface Buffer {
     Buffer setBytes(int index, ByteBuffer src);
 
     /**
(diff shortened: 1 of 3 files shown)


mgodave (approved):
lgtm

PR #2392, opened 2022-10-11
Prepare for upgrade to Netty 4.1.84

idelpivnitskiy:
Motivation:

Netty 4.1.84 includes https://github.com/netty/netty/pull/12760 which moves validation of connection-related headers for HTTP/2 from `HpackDecoder` to `DefaultHttp2Headers`. This breaks our existing control flow because in case there is no `content-length` header, we always unconditionally add `transfer-encoding: chunked` header that will be removed later in HTTP/2 pipeline if the HTTP/2 is the actual negotiated protocol.

Modification:

- Improve our control flow to take protocol into account when it adds `transfer-encoding: chunked` header;
- Adjust existing tests that target `transfer-encoding: chunked` handling to always use `DefaultHttpHeadersFactory`, even  when testing HTTP/2, it helps to avoid validation in netty's `DefaultHttp2Headers`;

Result:

All tests pass with both versions of netty: 4.1.82 and 4.1.84.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
@@ -57,8 +57,8 @@ final class H2ClientParentChannelInitializer implements ChannelInitializer {
                 // the user to apply their own timeout at the call site.
                 .gracefulShutdownTimeoutMillis(-1)
                 .initialSettings(nettySettings)
-                // We always validate incoming headers, define this explicitly instead of inheriting a default value.
-                .validateHeaders(true)
+                // Inherit headers validation setting from the HttpHeadersFactory.
+                .validateHeaders(config.headersFactory().validateNames())
                 .headerSensitivityDetector(config.headersSensitivityDetector()::test);
 
         initFrameLogger(multiplexCodecBuilder, config.frameLoggerConfig());
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
@@ -57,6 +57,8 @@ final class H2ClientParentChannelInitializer implements ChannelInitializer {
                 // the user to apply their own timeout at the call site.
                 .gracefulShutdownTimeoutMillis(-1)
                 .initialSettings(nettySettings)
+                // We always validate incoming headers, define this explicitly instead of inheriting a default value.
+                .validateHeaders(true)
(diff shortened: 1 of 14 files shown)


idelpivnitskiy:
We will need to refactor some of our HTTP control flow logic because of a change in Netty: https://github.com/netty/netty/pull/12760
```
Caused by: io.netty.handler.codec.http2.Http2Exception: Illegal connection-specific header 'transfer-encoding' encountered.
	at app//io.netty.handler.codec.http2.Http2Exception.connectionError(Http2Exception.java:109)
	at app//io.netty.handler.codec.http2.DefaultHttp2Headers$2.validateName(DefaultHttp2Headers.java:83)
	at app//io.netty.handler.codec.http2.DefaultHttp2Headers$2.validateName(DefaultHttp2Headers.java:42)
	at app//io.netty.handler.codec.DefaultHeaders.validateName(DefaultHeaders.java:1008)
	at app//io.netty.handler.codec.http2.DefaultHttp2Headers.validateName(DefaultHttp2Headers.java:168)
	at app//io.netty.handler.codec.http2.DefaultHttp2Headers.validateName(DefaultHttp2Headers.java:33)
	at app//io.netty.handler.codec.DefaultHeaders.add(DefaultHeaders.java:329)
	at app//io.servicetalk.http.netty.NettyH2HeadersToHttpHeaders.add(NettyH2HeadersToHttpHeaders.java:115)
	at app//io.servicetalk.http.netty.HeaderUtils.addRequestTransferEncodingIfNecessary(HeaderUtils.java:329)
	at app//io.servicetalk.http.netty.AbstractStreamingHttpConnection.lambda$request$2(AbstractStreamingHttpConnection.java:185)
	at app//io.servicetalk.concurrent.api.SingleDefer.handleSubscribe(SingleDefer.java:41)
```

idelpivnitskiy:
I've made changes that allow the default control flow to work with the latest Netty version. It doesn't solve the problem of users manually adding connection-specific headers, but it at least unblocks those users who don't: who rely on our control flow to populate payload-related headers or gRPC users.

The remaining use-case will be addressed either in Netty or in a follow-up PR.

This PR verified that it works with both netty versions, and intentionally doesn't update the version. It helps to unblock users who have Netty version updated transitively through some other libraries.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ClientParentChannelInitializer.java
```diff
@@ -57,6 +57,8 @@ public void init(final Channel channel) {
                 // the user to apply their own timeout at the call site.
                 .gracefulShutdownTimeoutMillis(-1)
                 .initialSettings(nettySettings)
+                // We always validate incoming headers, define this explicitly instead of inheriting a default value.
+                .validateHeaders(true)
```
Scottmitch, line 61: should this be derived from the headers factory used for h2?

Scottmitch (approved):

PR #2393, opened 2022-10-12
Run scheduled tests with Netty 4.1.X-SNAPSHOT

idelpivnitskiy:
Motivation:

To catch any issues with newer Netty releases, run a scheduled job to test ServiceTalk with the latest netty 4.1.X-SNAPSHOT.

Modifications:

- Adjust `build.gradle` files to exclude macos-specific artifacts when we build with a netty snapshot;
- Add CI job to run with Netty 4.1.+;

Result:

Regular tests with the latest Netty snapshot.

### DIFF
--- a/servicetalk-transport-netty-internal/build.gradle
+++ b/servicetalk-transport-netty-internal/build.gradle
@@ -17,7 +17,6 @@
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-library"
 
 ext.isNettySnapshot = "$nettyVersion".endsWithAny("SNAPSHOT", "+")
-println("!!! $nettyVersion")
 
 dependencies {
   api platform(project(":servicetalk-dependencies"))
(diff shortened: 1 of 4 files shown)


normanmaurer:
@idelpivnitskiy thanks a lot!

servicetalk-transport-netty-internal/build.gradle
```diff
@@ -16,6 +16,9 @@
 
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-library"
 
+ext.isNettySnapshot = "$nettyVersion".endsWithAny("SNAPSHOT", "+")
+println("!!! $nettyVersion")
```
Scottmitch, line 20: is this needed? if so can this be more descriptive (e.g. nettyVersion: ...)?

Scottmitch (approved):

servicetalk-transport-netty-internal/build.gradle
idelpivnitskiy: 👀 leftover

PR #2398, opened 2022-10-16
H2 TE header parsing bug fix

Scottmitch:
Motivation:
HTTP/2 has special rules that TE header is only allowed to have a value of `trailers`. The sanitizing of the TE headers may have infinite looped in some scenarios.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1Utils.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1Utils.java
@@ -15,7 +15,6 @@
  */
 package io.servicetalk.http.netty;
 
-import io.servicetalk.buffer.api.CharSequences;
 import io.servicetalk.http.api.HttpHeaders;
 
 import io.netty.handler.codec.http.HttpHeaderNames;
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1Utils.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1Utils.java
@@ -151,19 +151,35 @@ final class H2ToStH1Utils {
         Iterator<? extends CharSequence> teItr = h1Headers.valuesIterator(TE);
         boolean addTrailers = false;
         while (teItr.hasNext()) {
-            String teValue = teItr.next().toString();
-            int i = teValue.indexOf(',');
-            if (i != -1) {
-                int start = 0;
-                do {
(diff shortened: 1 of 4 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2PriorKnowledgeFeatureParityTest.java
```diff
@@ -322,6 +325,67 @@ void cookiesRoundTrip(HttpTestExecutionStrategy strategy, boolean h2PriorKnowled
...
+        setUp(strategy, h2PriorKnowledge);
+        InetSocketAddress serverAddress = bindHttpEchoServer();
+        try (BlockingHttpClient client = forSingleAddress(HostAndPort.of(serverAddress))
+                .protocols(h2PriorKnowledge ? h2Default() : h1Default())
+                .executionStrategy(clientExecutionStrategy).buildBlocking()) {
+            // Test individual headers
+            HttpRequest request = client.get("/");
+            request.addHeader(TE, "foo");
+            request.addHeader(TE, TRAILERS);
+            request.addHeader(TE, "bar");
```
idelpivnitskiy, line 340: In the next Netty version 4.1.84 the `DefaultHttp2Headers` impl will throw when users add `TE` value different from `trailers`. To keep the test coverage of `H2ToStH1Utils`, consider using `DefaultHttpHeadersFactory` when `h2PriorKnowledge == true` instead of `H2HeadersFactory`. 

Use-cases for `H2HeadersFactory` (backed by `DefaultHttp2Headers` from Netty) are already covered by tests in Netty.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1Utils.java
```diff
@@ -151,19 +151,35 @@ static Http2Headers h1HeadersToH2Headers(HttpHeaders h1Headers) {
...
-            int i = teValue.indexOf(',');
-            if (i != -1) {
-                int start = 0;
-                do {
-                    if (teValue.substring(start, i).compareToIgnoreCase(TRAILERS.toString()) == 0) {
+            final CharSequence teSequence = teItr.next();
+            if (addTrailers) {
+                teItr.remove();
+            } else {
+                final String teValue = teSequence.toString();
```
idelpivnitskiy, line 158: Looks like we don't need conversion to `String`, consider using `CharSequences.indexOf` and then `CharSequence.subSequence` with `contentEqualsIgnoreCase`

PR #2400, opened 2022-10-17
H2 conversion support OWS for connection header csv

Scottmitch:
Motivation:
H2 doesn't support the connection header, and must remove all headers that are referenced in the value of the conneciton header. In the event the connection header contains a CSV we should support OWS.


idelpivnitskiy (commented):
Consider testing with `DefaultHttpHeadersFactory` when `h2PriorKnowledge == true`.
Reminder that in Netty 4.1.84 `DefaultHttp2Headers` will throw if someone tries to add a `Connection` header.

PR #2409, opened 2022-10-26
`BeforeFinallyHttpOperator`: support re-subscribe to the message body

idelpivnitskiy:
Motivation:

`BeforeFinallyHttpOperator` has a single `state` inside `ResponseCompletionSubscriber` that is shared between `Single` and `Publisher` of the message body. If users re-subscribe to the message body, the state is already in `TERMINATED` position and does not propagate signals.

Modifications:

- Move message body processing logic into a new `MessageBodySubscriber` class;
- Each subscribe operation creates a new `MessageBodySubscriber` and their states are managed independently;
- Improve `IdleTimeoutConnectionFilterTest` to always create a new response object (otherwise, the response payload body can be transformed multiple times);

Result:

Users of `BeforeFinallyOperator` always see all terminal events, even for re-subscribe. Terminal signal will always be delivered to the second subscriber, even when `discardEventsAfterCancel` is set.


Scottmitch (approved):

Scottmitch (commented):
`IdleTimeoutConnectionFilterTest.twoConcurrentRequests()` test failure

idelpivnitskiy:
The problem was in the test, it reused a response object which resulted in applying transform on the payload body twice.

PR #2412, opened 2022-11-01
HostAndPort to support parsing ip+port string

Scottmitch:
Motivation:
Parsing HostAndPort from 'ip+port' string is a common case, but we don't have tooling to support this.


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/DefaultHostAndPort.java
```diff
@@ -30,8 +40,71 @@ final class DefaultHostAndPort implements HostAndPort {
      * @param port the port.
      */
     DefaultHostAndPort(String hostName, int port) {
+        this(hostName, port, isIPv6(hostName) ? '[' + hostName + "]:" + port : hostName + ':' + port);
+    }
+
+    private DefaultHostAndPort(String hostName, int port, String toString) {
         this.hostName = requireNonNull(hostName);
         this.port = port;
```
idelpivnitskiy, line 48: Should we check the port range?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/DefaultHostAndPort.java
```diff
@@ -62,4 +135,126 @@ public boolean equals(Object o) {
...
+    }
+
+    private static void validateIPv4(String inetAddress, int startIndex, int endIndex) {
+        int nibbles = 0;
+        int digits = 0;
+        for (int i = startIndex; i < endIndex; ++i) {
+            final char currChar = inetAddress.charAt(i);
+            if (currChar >= '0' && currChar <= '9') {
+                if (++digits > 3) {
+                    throw new IllegalArgumentException("No more than 3 digits per section expected");
```
idelpivnitskiy, line 237: Consider including a substring of the validating area to let users see what they gave as a parameter

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/DefaultHostAndPort.java
```diff
@@ -62,4 +135,126 @@ public boolean equals(Object o) {
...
+                        }
+                        if (shouldInsertRemainder) {
+                            compressedIPv6Builder.append(c);
+                            ++i;
+                            for (; i < rawIp.length(); ++i) {
+                                compressedIPv6Builder.append(rawIp.charAt(i));
+                            }
+                            break;
+                        } else {
+                            throw new IllegalArgumentException("Invalid IPv6 address[" + i + "]=" + c);
```
idelpivnitskiy, line 201: Consider including `rawIp`

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/DefaultHostAndPort.java
```diff
@@ -30,8 +40,71 @@ final class DefaultHostAndPort implements HostAndPort {
...
+     * @param startIndex The index at which the address parsing starts.
+     * @return A {@link HostAndPort} where the hostname is the IP address and the port is parsed from the string.
+     */
+    static HostAndPort parseFromIpPort(String ipPort, int startIndex) {
+        String inetAddress;
+        final boolean isv6;
+        int i;
+        if (ipPort.charAt(startIndex) == '[') { // check if ipv6
+            if (ipPort.length() - startIndex > MAX_IPV6_LEN) {
+                throw new IllegalArgumentException("Invalid IPv6 address: " + ipPort);
```
idelpivnitskiy, line 66: Should we use only substring starting from `startIndex` instead of the whole `ipPort` in the exception message?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/DefaultHostAndPort.java
```diff
@@ -62,4 +135,126 @@ public boolean equals(Object o) {
...
+            } else {
+                throw new IllegalArgumentException("Unexpected character in IPv4 address[" + i + "]=" + currChar);
+            }
+        }
+        if (nibbles != 3) {
+            throw new IllegalArgumentException("3 IP section separators expected, found " + nibbles);
+        }
+    }
+
+    private static boolean isIPv6(String address) {
```
idelpivnitskiy, line 253: Can we reuse anything from `io.servicetalk.utils.internal.NetworkUtils`?

Scottmitch: good point, forgot about this class!

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/DefaultHostAndPort.java
Scottmitch: I removed the validation embedded in this method and just used NetworkUtils ... we can combine/optimize later if needed.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/DefaultHostAndPort.java
Scottmitch: I used NetworkUtils for this now and print substring as suggested

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/DefaultHostAndPort.java
Scottmitch: I tried this but we have some code that depends upon passing in an invalid port to clarify it isn't specified (see `AbstractHttpRequestMetaDataTest` test failures). if we want to change this lets do it in a followup PR.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/DefaultHostAndPort.java
```diff
@@ -30,8 +46,86 @@ final class DefaultHostAndPort implements HostAndPort {
      * @param port the port.
      */
     DefaultHostAndPort(String hostName, int port) {
+        if (isValidIpV6Address(requireNonNull(hostName))) { // Normalize ipv6 so equals/hashCode works as expected
```
Scottmitch, line 49: @idelpivnitskiy - note that I made IPv6 do compression to be consistent in all cases.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/DefaultHostAndPort.java
```diff
@@ -30,8 +46,86 @@ final class DefaultHostAndPort implements HostAndPort {
...
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("invalid port " + ipPort.substring(startIndex), e);
+        }
+        if (!isValidPort(port) || ipPort.charAt(i + 1) == '+') { // parseInt allows '+' but we don't want this
+            throw new IllegalArgumentException("invalid port " + ipPort.substring(startIndex));
+        }
+
+        if (isv6) {
+            inetAddress = compressIPv6(inetAddress, 0, inetAddress.length());
+            if (!isValidIpV6Address(inetAddress)) {
```
idelpivnitskiy, line 120: Should we run validation before compressing it?

idelpivnitskiy (approved):
Thank you!

PR #2413, opened 2022-11-07
Publisher#takeUntil cancel before terminate

Scottmitch:
Motivation:
Publisher#takeUntil will terminate downstream before cancelling upstream. There are some sources (e.g. NettyChannelPublisher) that allow resubscribe in a sequential fashion, but in order for this to work cancellation must be first.

Modifications:
- Switch ordering in TakeUtil to cancel before terminate
- Add TestResubscribePublisher utility that wraps TestPublisher which allows for sequential resbuscribes.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -2224,11 +2224,37 @@ public abstract class Publisher<T> {
      *
      * @param until {@link Completable}, termination of which, terminates the returned {@link Publisher}.
      * @return A {@link Publisher} that only emits the items till {@code until} {@link Completable} is completed.
-     *
+     * @deprecated Use {@link #takeUntil(Supplier)}.
      * @see <a href="https://reactivex.io/documentation/operators/takeuntil.html">ReactiveX takeUntil operator.</a>
      */
+    @Deprecated
     public final Publisher<T> takeUntil(Completable until) {
-        return new TakeUntilPublisher<>(this, until);
+        return takeUntil(() -> until);
+    }
+
+    /**
+     * Takes elements until {@link Completable} is terminated successfully or with failure.
+     * <p>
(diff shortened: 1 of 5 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2224,11 +2224,37 @@ public final Publisher<T> takeWhile(Predicate<? super T> predicate) {
...
+     *     }
+     *     return results;
+     * }</pre>
+     *
+     * @param untilSupplier {@link Supplier} that is invoked on each subscribe that provides a {@link Completable},
+     * termination of which, terminates the returned {@link Publisher}.
+     * @return A {@link Publisher} that only emits the items till {@code until} {@link Completable} is completed.
+     * @see <a href="https://reactivex.io/documentation/operators/takeuntil.html">ReactiveX takeUntil operator.</a>
+     */
+    public final Publisher<T> takeUntil(Supplier<? extends Completable> untilSupplier) {
```
Scottmitch, line 2256: @idelpivnitskiy - I know we have had discussion about providing `Supplier` vs using `defer(() -> ..)` in the past. We could continue to assume folks know they should use `defer(() -> ..)` but in this case it was missed in internal usage and when `Completable` terminates the Publisher is also terminated, which may trigger re-subscribe...

Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/2414

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
Scottmitch: discussed offline ... `Publisher.defer(() -> ..)` can be used as a building block in this case. we can add followup docs/examples to clarify making operators that may retain state re-subscribe safe.

Scottmitch:
merging as no API changes and discussed offline

idelpivnitskiy (commented):
Late LGTM, thanks!

PR #2419, opened 2022-11-12
Server control flow should close connection on `DecoderException`

idelpivnitskiy:
Motivation:

#2367 changed server-side control flow. After that, `connection.write` doesn't start before we received a request meta-data. As the result, all errors that happen during parsing of the meta-data do not reach `WriteStreamSubscriber` and can leave a connection in undefined state, when we logged an error but didn't clean up the channel state.

Modifications:

- Close the channel if an error occurs on the `Single` of request meta-data;
- Add tests to verify the expected behavior;

Result:

Channel is closed when an early request error happens, client doesn't hang forever.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
@@ -49,6 +49,7 @@ import io.servicetalk.transport.api.ConnectionContext;
 import io.servicetalk.transport.api.ConnectionObserver;
 import io.servicetalk.transport.api.ServerContext;
 import io.servicetalk.transport.api.SslConfig;
+import io.servicetalk.transport.netty.internal.ChannelCloseUtils;
 import io.servicetalk.transport.netty.internal.ChannelInitializer;
 import io.servicetalk.transport.netty.internal.CloseHandler;
 import io.servicetalk.transport.netty.internal.CloseHandler.CloseEventObservedException;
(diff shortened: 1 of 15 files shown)


idelpivnitskiy:
The plan for the next release is to revert #2367, cut the release, reapply #2367, then merge this fix.

idelpivnitskiy:
This PR will go in after #2367 is reverted back, see #2423.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -298,7 +298,10 @@ public FlushStrategy defaultFlushStrategy() {
 
         private Completable handleRequestAndWriteResponse(final Single<StreamingHttpRequest> requestSingle,
                                                           final boolean handleMultipleRequests) {
-            final Completable exchange = requestSingle.flatMapCompletable(rawRequest -> {
+            final Completable exchange = requestSingle.whenOnError(t -> {
+                // If an error happened before we could start writing, we need to close the channel.
+                nettyChannel().close();
```
Scottmitch, line 303: when we close under the hood it can be difficult for users to understand who initiated closure. Consider either:
- moving this close operation into `ErrorLoggingHttpSubscriber` after it logs the reason
- adding a comment that `ErrorLoggingHttpSubscriber` will log the exception if required.

Scottmitch:
lgtm after the above is addressed.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
idelpivnitskiy: Good point, I've moved closure to `ErrorLoggingHttpSubscriber`

idelpivnitskiy:
I've reverted back server-side control flow as the first commit and applied a change of this PR as the second commit. Will use "Rebase and merge"

PR #2420, opened 2022-11-12
Preserve the caller's stacktrace for blocking API

idelpivnitskiy:
Motivation:

Users of blocking API do not expect `ExecutionException`, like users of `Future.get()`. However, `ExecutionException` plays an important role to capture the caller's stacktrace, which helps to understand which path made a blocking call. Currently, we unwrap `ExecutionException` and lose this information.

Modifications:

- Instead of dropping the `ExecutionException` wrapper, move it as a suppressed exception for the original one;
- Test new behavior;

Result:

Users can understand what path requested blocking invocation when an async operation fails with an exception.


Scottmitch (approved):

tkountis (approved):
🚢

PR #2421, opened 2022-11-12
Move gRPC server deadline filter, allow custom insertion order

Scottmitch:
Motivation:
The default insertion order for gRPC deadline filter is after user filters. If a user's filter deadlocks the timeout will never be applied.

Some use cases may necessitate applying the deadline timeout filter in a custom order (e.g. relative to retries). There is currently no way to customize the append ordering.

Modifications:
- Move the gRPC server deadline filter to the beginning of the filter chain so timeouts are applied regardless if user filter code hangs.
- Add appendTimeoutFilter method to gRPC client and server builder to give users control over filter append ordering.
- Introduce GrpcFilters utility class to create the filters that apply deadline propagation.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
@@ -16,6 +16,7 @@
 package io.servicetalk.grpc.api;
 
 import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
+import io.servicetalk.http.api.StreamingHttpClientFilterFactory;
 
 import java.time.Duration;
 
@@ -71,6 +72,23 @@ public interface GrpcClientBuilder<U, R> {
      * @return {@code this}.
      */
     GrpcClientBuilder<U, R> defaultTimeout(Duration defaultTimeout);
+
+    /**
+     * Determine if a filter will be inserted by this builder that enforces the
+     * <a href="https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests">Timeout deadline
+     * propagation</a>.
+     * <p>
+     * To insert {@link GrpcFilters#newGrpcDeadlineClientFilterFactory()} in your preferred order use
+     * {@link #initializeHttp} and
+     * {@link SingleAddressHttpClientBuilder#appendClientFilter(StreamingHttpClientFilterFactory)}.
+     * <p>
+     * {@link #defaultTimeout(Duration)} is independent of this method, and may still inject state (even if no timeout
+     * is applied locally because this {@code append} is {@code false} and no timeout filter is appended).
+     * @param append {@code true} if this builder should append the timeout filter, {@code false} if it should not.
+     * @return {@code this}.
+     * @see GrpcFilters#newGrpcDeadlineClientFilterFactory()
+     */
+    GrpcClientBuilder<U, R> appendTimeoutFilter(boolean append);
 
     /**
      * Builds a <a href="https://www.grpc.io">gRPC</a> client.
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
@@ -19,6 +19,7 @@ import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
 import io.servicetalk.http.api.StreamingHttpClientFilterFactory;
 
 import java.time.Duration;
+import javax.annotation.Nullable;
 
 /**
  * A builder for building a <a href="https://www.grpc.io">gRPC</a> client.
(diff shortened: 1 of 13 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExceptionMapperServiceFilter.java
```diff
@@ -84,9 +84,11 @@ private static StreamingHttpResponse convertToGrpcErrorResponse(final String wha
                 ctx.executionContext().bufferAllocator(), null);
         final CharSequence codeValue = response.headers().get(GRPC_STATUS);
         assert codeValue != null;
-        LOGGER.error("Unexpected exception during a {} processing for connection={}, request='{} {} {}' was mapped " +
-                        "to grpc-status: {} ({})", what, ctx, request.method(), request.requestTarget(),
-                request.version(), codeValue, fromCodeValue(codeValue), cause);
+        if (LOGGER.isInfoEnabled()) {
```
Scottmitch, line 87: @idelpivnitskiy - note this happens for all timeouts, logging at error for timeouts seems too high, wdyt?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcFilters.java
```diff
@@ -0,0 +1,102 @@
...
+
+import static io.servicetalk.grpc.internal.DeadlineUtils.GRPC_DEADLINE_KEY;
+import static io.servicetalk.grpc.internal.DeadlineUtils.readTimeoutHeader;
+import static io.servicetalk.utils.internal.DurationUtils.ensurePositive;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+/**
+ * Utility filters for gRPC.
+ */
+public final class GrpcFilters {
```
Scottmitch, line 41: this doesn't strictly need to be in `grpc.api` (could go in `grpc.netty`) but I put it here to reference from javadocs on the builder method for easier discovery.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
```diff
@@ -72,6 +73,23 @@ default HttpInitializer<U, R> append(HttpInitializer<U, R> toAppend) {
      */
     GrpcClientBuilder<U, R> defaultTimeout(Duration defaultTimeout);
```
Scottmitch, line 74: on the server side we "could" use this method to set a `null` value as disabling the filter insertion (because the value is ignored if you don't append the filter ... which is different than the client where a default can be inserted without enforcing locally). however I went with `appendTimeoutFilter` for both client and server to be consistent (but slight awkwardness where this value is ignored on the server if  `appendTimeoutFilter(false)`).

servicetalk-grpc-netty/build.gradle
```diff
@@ -32,7 +32,7 @@ dependencies {
   api project(":servicetalk-grpc-utils")
   api project(":servicetalk-http-netty")
 
-  implementation project(":servicetalk-http-utils")
+  compileOnly project(":servicetalk-http-utils")
```
Scottmitch, line 35: only timeout filter from this project is referenced and just from javadocs...

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatus.java
```diff
@@ -162,9 +188,12 @@ public GrpcStatusCode code() {
      * Returns the cause, can be null.
      *
      * @return the cause, can be null.
+     * @deprecated Use {@link GrpcStatusException#fromThrowable(Throwable)} to create a {@link GrpcStatusException}
+     * then {@link GrpcStatusException#getCause()}.
      */
     @Nullable
-    public Throwable cause() {
+    @Deprecated
+    public Throwable cause() { // FIXME: 0.43 - remove deprecated method
```
Scottmitch, line 196: the relationship between `GrpcStatus` and `GrpcStatusException` has always been awkward. This change will move us toward `GrpcStatus` just being a status (not containing an exception) and using  `GrpcStatusException` when an exception is in play

Scottmitch: moved to https://github.com/apple/servicetalk/pull/2422

Scottmitch:
PR depends upon https://github.com/apple/servicetalk/pull/2422 (for consistent error codes return when client times out)

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
```diff
@@ -116,7 +112,9 @@ public <BlockingClient extends BlockingGrpcClient<?>> BlockingClient buildBlocki
     private GrpcClientCallFactory newGrpcClientCallFactory() {
         SingleAddressHttpClientBuilder<U, R> builder = httpClientBuilderSupplier.get().protocols(h2Default());
         builder.appendClientFilter(CatchAllHttpClientFilter.INSTANCE);
-        builder.appendClientFilter(new TimeoutHttpRequesterFilter(GRPC_TIMEOUT_REQHDR, true));
+        if (appendTimeoutFilter) {
+            builder.appendClientFilter(newGrpcDeadlineClientFilterFactory());
+        }
         httpInitializer.initialize(builder);
```
idelpivnitskiy, line 118: There are 2 more approaches to consider:
1. An approach that does not require new API would be to always add a conditional deadline filter, then wrap `builder` before running it through `httpInitializer` and after initializer returns check if user added another `TimeoutHttp*Filter` or not. If yes, we change a predicate for our conditional filter to be always `false` to keep only their filter.
2. Another option is to add another `defaultTimeout` overload that takes `Duration` and `boolean`. Users will have only one method that controls both: default value and should we add a filter for them or not.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExceptionMapperServiceFilter.java
idelpivnitskiy: Can we log at different levels for timeout exceptions and keep `error` for everything else?

servicetalk-grpc-netty/build.gradle
idelpivnitskiy: Consider either adding a comment here so we don't forget or changing javadoc from using `@link` to `@code` with full class name, then you can kill this line.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcClientBuilder.java
Scottmitch: >we can mitigate it by still adding the defaultTimeout into AsyncContext

I would prefer not to complicate things and add dependencies on async context for the default value. users will just have to provide the default value again when they create/insert the filter.

servicetalk-grpc-netty/build.gradle
Scottmitch: I'll add a comment here. `{@link}` ensures code doesn't become stale and makes javadocs easier to navigate.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcFilters.java
```diff
@@ -0,0 +1,102 @@
...
+            Duration timeout = null != requestTimeout ? requestTimeout : defaultTimeout;
+
+            if (null != timeout) {
+                // Store the timeout in the context as a deadline to be used for any client requests created
+                // during the context of handling this request.
+                try {
+                    Long deadline = timeSource.currentTime(NANOSECONDS) + timeout.toNanos();
+                    AsyncContext.put(GRPC_DEADLINE_KEY, deadline);
+                } catch (UnsupportedOperationException ignored) {
+                    LOGGER.debug("Async context disabled, timeouts will not be propagated to client requests");
```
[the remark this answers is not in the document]
Scottmitch: I'll kill it for now

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcClientBuilder.java
Scottmitch: I would prefer not todo (1). If there is code that doesn't need to be instantiated lets have the user be explicit about it to avoid any surprises.

I will do (2). we would need to allow `@Nullable` for default timeout as users could not want to insert a default if one isn't present.

idelpivnitskiy (approved):
🚀 
Thank you for taking care of this!

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExceptionMapperServiceFilter.java
Scottmitch: yes let me exclude this log for timeout exception ... folks can get visibility via Observer if required.

PR #2422, opened 2022-11-12
Decouple GrpcStatusException creation from GrpcStatus

Scottmitch:
Motivation:
Creation of GrpcStatusException must first go through GrpcStatus. This includes propagation of root cause and unwrapping of existing Throwables. This coupling is unnecessary and complicates the creation and relationship between these classes.

Modifications:
- Deprecate GrpcStatus methods related to creation of GrpcStatusException
- Add public methods to GrpcStatusException that allow for creation and take GrpcStatus as state


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatus.java
```diff
@@ -70,25 +78,38 @@ public GrpcStatus(GrpcStatusCode code, @Nullable Throwable cause) {
...
 
+    /**
+     * Constructs a status with cause and additional description.
+     * @param code status code.
+     * @param description additional description.
+     */
+    public GrpcStatus(GrpcStatusCode code, @Nullable CharSequence description) {
+        this.code = requireNonNull(code);
+        this.cause = null;
+        this.description = description == null ? null : description.toString();
```
idelpivnitskiy, line 99: If it's always a `String` internally, consider either changing the input argument to `String` or adding an alternative `description()` method that returns `CharSequence`.

servicetalk-grpc-health/src/main/java/io/servicetalk/grpc/health/DefaultHealthService.java
```diff
@@ -81,8 +82,8 @@ public DefaultHealthService(Predicate<String> watchAllowed) {
     public Single<HealthCheckResponse> check(final GrpcServiceContext ctx, final HealthCheckRequest request) {
         HealthValue health = serviceToStatusMap.get(request.getService());
         if (health == null) {
-            return Single.failed(new GrpcStatus(NOT_FOUND, null, "unknown service: " + request.getService())
-                    .asException());
+            return Single.failed(new GrpcStatusException(
+                    new GrpcStatus(NOT_FOUND, "unknown service: " + request.getService())));
```
idelpivnitskiy, line 86: Could be a follow-up, but something to consider: for use-cases like this one it looks useful to be able to create a stackless `GrpcStatusException` because it's not used for anything else except population of the `grpc-status` & `grpc-message`.
WDYT about adding `GrpcStatusException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace)` ctor? That way we and users can create static exceptions with no stacktrace and no suppressions.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatus.java
```diff
@@ -144,8 +171,11 @@ public GrpcStatusException asException() {
      *
      * @param applicationStatusSupplier the {@link Supplier} for the {@link Status}.
      * @return the current status wrapped in a {@link GrpcStatusException}.
+     * @deprecated Use {@link GrpcStatusException#GrpcStatusException(GrpcStatus)}.
```
idelpivnitskiy, line 174: `GrpcStatusException#of(Status)` seems like a better alternative for `GrpcStatus#asException(Supplier<Status>)`

idelpivnitskiy (commented):
While I don't have a strong preference to any approach, note that this change complicates discoverability of the `GrpcStatusException` type. Before this one, users had to know only about `GrpcStatus`. Now, they have to know what exception type they should throw.

servicetalk-grpc-health/src/main/java/io/servicetalk/grpc/health/DefaultHealthService.java
Scottmitch: lets consider this as a followup if necessary.

Scottmitch:
Users still had to be aware of `GrpcStatusException` (e.g. if they customize there Status, or ensuring throwing this type of exception) ... they just couldn't create it without going through `GrpcStatus`.  This has confused users (and myself) in the past.

idelpivnitskiy (approved):

Issue #2378, opened 2022-09-29
Flaky H2PriorKnowledgeFeatureParityTest.serverGracefulClose


idelpivnitskiy:
Unfortunately, #2375 didn't fix the issue. Saw a new test failure:

https://github.com/apple/servicetalk/runs/8607775207

```
H2PriorKnowledgeFeatureParityTest.serverGracefulClose(HttpTestExecutionStrategy, boolean) [1] client=NO_OFFLOAD, h2PriorKnowledge=true
java.util.concurrent.TimeoutException: serverGracefulClose(io.servicetalk.http.netty.HttpTestExecutionStrategy, boolean) timed out after 30 seconds
...
	Suppressed: java.lang.InterruptedException
		at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1048)
		at java.base/java.util.concurrent.CountDownLatch.await(CountDownLatch.java:230)
		at io.servicetalk.http.netty.H2PriorKnowledgeFeatureParityTest.serverGracefulClose(H2PriorKnowledgeFeatureParityTest.java:1200)
		at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
		...
```

Scottmitch:
```
H2PriorKnowledgeFeatureParityTest.serverGracefulClose(HttpTestExecutionStrategy, boolean) [1] client=NO_OFFLOAD, h2PriorKnowledge=true
java.util.concurrent.TimeoutException: serverGracefulClose(io.servicetalk.http.netty.HttpTestExecutionStrategy, boolean) timed out after 30 seconds

	Suppressed: java.lang.InterruptedException
		at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1040)
		at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1345)
		at java.base/java.util.concurrent.CountDownLatch.await(CountDownLatch.java:232)
		at io.servicetalk.http.netty.H2PriorKnowledgeFeatureParityTest.serverGracefulClose(H2PriorKnowledgeFeatureParityTest.java:1200)
```

idelpivnitskiy:
https://github.com/apple/servicetalk/runs/8648726776

```
[...]

PR #2424, opened 2022-11-15
DefaultDnsClient ConcurrentModificationException

Scottmitch:
Motivation:
DefaultDnsClient takes the List from the Netty's DNS client and sorts it. This may result in a ConcurrentModificationException if the underlying cache modifies this list.

Modifications:
- Make a copy of the List we get back from Netty before we sort the collection and retain a reference.


Scottmitch:
test failure due to https://github.com/apple/servicetalk/issues/2378

PR #2426, opened 2022-11-18
ScheduledExecutorService to remove timeout objects on cancel

Scottmitch:
Motivation:
By default the JDK's ScheduledExecutorService does not remove timer objects from its queues upon cancel. Cancellation is expected to be common when requests have timeouts applied, and therefore the object retention lifetime will be longer than required and may negatively impact GC performance.


idelpivnitskiy (approved):
Good idea!

PR #2429, opened 2022-11-19
Add gRPC Request Response Context Example

idelpivnitskiy:
Motivation:

Demonstrate how users can use gRPC request/response context to access HTTP headers.

Modifications:

- Add `servicetalk-examples-grpc-request-response-context` module;

Result:

Users can see how to use gRPC request/response context.


Scottmitch (approved):
❤️

PR #2430, opened 2022-11-20
Add ListenableAsyncCloseable#onClosing()

Scottmitch:
Motivation:
It is useful to get a notified on the leading edge when closing beings. This may drive decisions like stopping to use a connection or resources that is being closed and not suitable for reuse.

Modifications:
- Use existing transport mechansims to drive onClosing for connections.
- Remove NettyConnectionContext override of onClosing and defer to the base class.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/CloseUtils.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/CloseUtils.java
@@ -44,6 +44,9 @@ final class CloseUtils {
      * @param closingStarted a {@link CountDownLatch} to notify
      */
     static void onGracefulClosureStarted(ConnectionContext cc, CountDownLatch closingStarted) {
+        // cc.onClosing() will trigger on the leading edge of closure, which maybe when the user calls closeAsync().
+        // The tests that depend upon this method need to wait until protocol events occur to ensure no more data will
+        // be processed, which isn't the same as cc.onClosing().
         NettyConnectionContext nettyCtx = (NettyConnectionContext) cc;
         if (cc.protocol() == HTTP_1_1) {
             nettyCtx.transportError().subscribe(t -> {
--- a/servicetalk-client-api-internal/src/main/java/io/servicetalk/client/api/internal/DefaultPartitionedClientGroup.java
+++ b/servicetalk-client-api-internal/src/main/java/io/servicetalk/client/api/internal/DefaultPartitionedClientGroup.java
@@ -132,6 +132,11 @@ public final class DefaultPartitionedClientGroup<U, R, Client extends Listenable
     }
 
     @Override
+    public Completable onClosing() {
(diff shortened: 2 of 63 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/CloseUtils.java
```diff
@@ -44,6 +44,9 @@ private CloseUtils() {
      * @param closingStarted a {@link CountDownLatch} to notify
      */
     static void onGracefulClosureStarted(ConnectionContext cc, CountDownLatch closingStarted) {
+        // cc.onClosing() will trigger on the leading edge of closure, which maybe when the user calls closeAsync().
```
Scottmitch, line 47: this was an interesting observation... if we needed to differentiate between "any trigger for closing" (which is what `onClosing` is) and "transport/protocol is closing" we could add another method in the future, but currently this seems to be mostly a concern in tests.

idelpivnitskiy (changes requested):
Thanks for catching other cases when we miss overriding a `default` method!
Everything looks good. However, there are a couple more cases that need an override for `onClosing()`:
- AbstractNettyIoExecutor
- ClientFactory in generated gRPC code

Also, please apply this patch:
```patch
Index: servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelListenableAsyncCloseable.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelListenableAsyncCloseable.java b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelListenableAsyncCloseable.java
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelListenableAsyncCloseable.java	(revision cfd0bf0cbf7d299d87076e714c32ec2c55ad1da9)
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyChannelListenableAsyncCloseable.java	(date 1669842880284)
@@ -101,6 +101,7 @@
      *
      * @return a {@link Completable} that notifies when the connection has begun its closing sequence.
      */
+    @Override
     public final Completable onClosing() {
         return fromSource(onClosing);
     }
@@ -162,7 +163,7 @@
         return onClose;
     }
 
-    final Completable onCloseNoOffload() {
+    private Completable onCloseNoOffload() {
         return onCloseNoOffload;
     }
 ```

Scottmitch:
> AbstractNettyIoExecutor

This was intentional as there isn't a reliable leading edge trigger for `EventLoopGroup`. I can add some state in this class to provide best effort leading edge notification though.

idelpivnitskiy:
Thanks!

PR #2431, opened 2022-11-20
CloseHandler force RST on close if still reading to avoid half close hang

Scottmitch:
Motivation:
HttpLifecycleObserverTest#testClientCancelsRequestAfterResponse tests half close
behavior but is subject to the OS's TCP implementation sending a RST before
the test writes, and therefore the write returning an error. However some
scenarios the client would just send a FIN and the server write would succeed.

Modifications:
- Server should wait to write until it has recieved the RST
- CloseHandler should force a RST before full close is done. This can
  be achieved by setting SO_LINGER to 0.

Result:
Fixes `HttpLifecycleObserverTest#testClientCancelsRequestAfterResponse(HttpProtocol) [1] protocol=HTTP_1`
prevents hangs when read is complete and write is still open.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
@@ -139,7 +139,7 @@ abstract class AbstractLifecycleObserverHttpFilter implements HttpExecutionStrat
             try {
                 responseSingle = responseFunction.apply(transformed);
             } catch (Throwable t) {
-                onExchange.onResponseError(t);
+                exchangeContext.onError(t); // Go through exchange context so finally methods are invoked.
                 return Single.<StreamingHttpResponse>failed(t).shareContextOnSubscribe();
             }
             return responseSingle
@@ -249,6 +249,11 @@ abstract class AbstractLifecycleObserverHttpFilter implements HttpExecutionStrat
                 safeReport(onExchange::onExchangeFinally, onExchange, "onExchangeFinally");
                 clearContext.run();
             }
+        }
+
+        @Override
+        public String toString() {
+            return "remaining=" + remaining;
         }
     }
 
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CloseHandler.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CloseHandler.java
@@ -164,7 +164,7 @@ public abstract class CloseHandler {
     abstract void channelCloseNotify(ChannelHandlerContext ctx);
 
     /**
-     * Signal that {@link Channel#close()} is above to be called.
+     * Signal that {@link Channel#close()} is about to be called.
      * @param channel The channel.
      */
     void channelClose(Channel channel) {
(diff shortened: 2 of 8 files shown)


Scottmitch:
this PR depends upon https://github.com/apple/servicetalk/pull/2430 (test uses the new `onClosing()` method)

Scottmitch:
without `SO_LINGER=0` the server writes after FIN were accepted by the OS, and the server only receives RST from client, after the client reads more data after it already sent a FIN.

Screen Shot 2022-11-20 at 9 04 26 AM

with `SO_LINGER=0` the client sends a RST (instead of a FIN) and then the server writes fail.

Screen Shot 2022-11-20 at 9 05 23 AM

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
```diff
@@ -139,7 +139,7 @@ public void cancel() {
             try {
                 responseSingle = responseFunction.apply(transformed);
             } catch (Throwable t) {
-                onExchange.onResponseError(t);
+                exchangeContext.onError(t); // Go through exchange context so finally methods are invoked.
```
idelpivnitskiy, line 142: 👍

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CloseHandler.java
```diff
@@ -163,6 +163,14 @@ public static CloseHandler forNonPipelined(boolean isClient, ChannelConfig confi
      */
     abstract void channelCloseNotify(ChannelHandlerContext ctx);
 
+    /**
+     * Signal that {@link Channel#close()} is above to be called.
```
idelpivnitskiy, line 167: above -> about?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NonPipelinedCloseHandler.java
```diff
@@ -162,6 +163,18 @@ void gracefulUserClosing(final Channel channel) {
...
 
+    @Override
+    void channelClose(final Channel channel) {
+        // We need to set SO_LINGER before we invoke close if we are not writing and still reading. This forces sending
+        // RST (instead of FIN) which the peer will interpret as "full close" and any future writes by the peer will
+        // fail (if the peer is half closed, not reading, but still writing). Otherwise, the peer write attempts may
+        // be accepted by the OS, only to generate RST when the local peer receives the data, but the peer has finished
+        // reading which may prevent processing the RST and be hung.
+        if (isAllSet(state, READ) && !isAllSet(state, WRITE) && channel instanceof SocketChannel) {
+            setSocketResetOnClose((SocketChannel) channel);
```
idelpivnitskiy, line 174: NonPipelinedCloseHandler is used by child channels that represent a virtual stream. Does setting a socket option here make any difference?

idelpivnitskiy (approved):
Thanks!

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NonPipelinedCloseHandler.java
Scottmitch: No `Http2StreamChannel` currently extends `Channel` currently (not `SocketChannel`). However if/when it does I don't think this will meaningfully change behavior.

PR #2435, opened 2022-11-23
Add utilities to create ForkJoinPool

Scottmitch:
Motivation:
ForkJoinPool has a custom ThreadFactory interface ForkJoinWorkerThread that we can implement internally to make its use easier.

Modifications:
- Add new ForkJoinThread type and method which accepts ForkJoinWorkerThreadFactory to Executors


Scottmitch:
I was experimenting with FJP and added this to capture later, not sure how/if it can be used to improve default perf...

Scottmitch:
closing as this isn't a current priority.

PR #2436, opened 2022-11-23
Reduce allocation in PublishAndSubscribeOn*

Scottmitch:
Motivation:
Intermediate allocations from PublishAndSubscribeOn* are in the top 10 on jfr reports from perf bencmarks.

Modifications:
- Remove intermediate allocations for safe invocation of handleSubscribe. The parent method is internal and which calls a final method which handles exceptions.


Scottmitch:
Before this PR:

Screen Shot 2022-11-22 at 5 57 13 PM

After this PR:
Screen Shot 2022-11-22 at 5 57 31 PM

idelpivnitskiy (approved):
🚀

PR #2437, opened 2022-11-23
Add FlowControlUtils.addWithUnderOverflowProtection for int

Scottmitch:
Motivation:
FlowControlUtils.addWithUnderOverflowProtection exists for long types but not int types.


servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/FlowControlUtils.java
```diff
@@ -124,4 +124,22 @@ public static long addWithUnderOverflowProtection(final long x, final long y) {
...
+     * @param x first value.
+     * @param y second value.
+     * @return
+     * <ul>
+     *     <li>{@code x + y} if no overflow</li>
+     *     <li>{@link Integer#MAX_VALUE} if overflow in the positive direction</li>
+     *     <li>{@link Integer#MIN_VALUE} if otherwise in the negative direction</li>
+     * </ul>
+     */
+    public static int addWithUnderOverflowProtection(final int x, final int y) {
```
Scottmitch, line 140: @tkountis - plan to use this to replace `DefaultLoadBalancedAddress.addWithUnderflowOrOverflowProtection`.

PR #2446, opened 2022-11-28
Pin specific JDK11 version to avoid javadoc bug

idelpivnitskiy:
Motivation:

CodeQL job started to fail with javadoc bug:
https://github.com/apple/servicetalk/actions/runs/3567715145/jobs/5995737033
Looks like they updated JDK11 version.

Modifications:

- Force CodeQL task to use a specific JDK11 version;

Result:

CodeQL task continues to work.


idelpivnitskiy:
Will merge to unblock other PRs.

PR #2450, opened 2022-12-01
RoundRobinLoadBalancer fix connection leak during graceful closure

Scottmitch:
Motivation:
RoundRobinLoadBalancer clears state of all hosts and each host clears all connections when closing. However a common pattern is to initiate gracefulClosure with a timeout, and then force a full close after a timeout interval. In this scenario RRLB has discarded the reference to connections and will not full close the connections.

Modifications:
- asyncCloseable needs to keep state about what hosts are being closed and each host can't dereference connections until they are fully closed.
- switch back to using onClose to cleanup the connection list as otherwise we may leak connections that have started graceful closure and never finished.

### DIFF
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -45,13 +45,19 @@ import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Comparator;
+import java.util.Iterator;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Map.Entry;
+import java.util.Spliterator;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.function.UnaryOperator;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
 
@@ -91,7 +97,6 @@ final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalancedConnec
         implements LoadBalancer<C> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(RoundRobinLoadBalancer.class);
-    private static final List<?> CLOSED_LIST = new ArrayList<>(0);
     private static final Object[] EMPTY_ARRAY = new Object[0];
 
     @SuppressWarnings("rawtypes")
@@ -128,8 +133,6 @@ final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalancedConnec
     private final ConnectionFactory<ResolvedAddress, ? extends C> connectionFactory;
     private final int linearSearchSpace;
     private final ListenableAsyncCloseable asyncCloseable;
-    @Nullable
-    private CompositeCloseable compositeCloseable;
 
     /**
      * Creates a new instance.
@@ -178,7 +181,7 @@ final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalancedConnec
                     @SuppressWarnings("unchecked")
                     final List<Host<ResolvedAddress, C>> usedAddresses =
                             usedHostsUpdater.updateAndGet(RoundRobinLoadBalancer.this, oldHosts -> {
-                                if (oldHosts == CLOSED_LIST) {
+                                if (isClosedList(oldHosts)) {
                                     return oldHosts;
                                 }
                                 final ResolvedAddress addr = requireNonNull(event.address());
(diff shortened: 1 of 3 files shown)


servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -640,12 +663,12 @@ previous, new ConnState(newList, newState))) {
             LOGGER.trace("Load balancer for {}: added a new connection {} to {} after {} attempt(s).",
                     targetResource, connection, this, addAttempt);
             // Instrument the new connection so we prune it on close
-            connection.onClosing().beforeFinally(() -> {
+            connection.onClose().beforeFinally(() -> {
```
Scottmitch, line 666: while this is preferred to remove connections from the selection process we still need to manage the lifecycle of the connections. we can use `connection.onClosing` in the future to disable selection eligibility but the connection by adding extra state or another collection but this PR prioritizes correctness and we can followup to optimize.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -891,4 +918,166 @@ public static StacklessConnectionRejectedException newInstance(String message, C
...
+        }
+
+        @Override
+        public boolean removeAll(final Collection<?> c) {
+            return delegate.removeAll(c);
+        }
+
+        @Override
+        public boolean removeIf(final Predicate<? super T> filter) {
+            return List.super.removeIf(filter);
```
idelpivnitskiy, line 1000: Use `delegate`

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -891,4 +918,166 @@ public static StacklessConnectionRejectedException newInstance(String message, C
...
+        }
+
+        @Override
+        public boolean retainAll(final Collection<?> c) {
+            return delegate.retainAll(c);
+        }
+
+        @Override
+        public void replaceAll(final UnaryOperator<T> operator) {
+            List.super.replaceAll(operator);
```
idelpivnitskiy, line 1010: Use `delegate`

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -891,4 +918,166 @@ public static StacklessConnectionRejectedException newInstance(String message, C
...
+        }
+
+        @Override
+        public void sort(final Comparator<? super T> c) {
+            List.super.sort(c);
```
idelpivnitskiy, line 1015: Use `delegate`

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -891,4 +918,166 @@ public static StacklessConnectionRejectedException newInstance(String message, C
...
+            delegate.forEach(action);
+        }
+
+        @Override
+        public Object[] toArray() {
+            return delegate.toArray();
+        }
+
+        @Override
+        public <T1> T1[] toArray(final T1[] a) {
```
idelpivnitskiy, line 964: There is also `toArray(final IntFunction<T1[]> generator)` overload that is not delegated. Also, consider delegating methods from `Object`

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -640,12 +667,12 @@ previous, new ConnState(newList, newState))) {
             LOGGER.trace("Load balancer for {}: added a new connection {} to {} after {} attempt(s).",
                     targetResource, connection, this, addAttempt);
             // Instrument the new connection so we prune it on close
-            connection.onClosing().beforeFinally(() -> {
+            connection.onClose().beforeFinally(() -> {
                 int removeAttempt = 0;
                 for (;;) {
                     ++removeAttempt;
```
idelpivnitskiy, line 673: Consider moving `++removeAttempt;` after checking the state

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -620,7 +647,7 @@ boolean addConnection(C connection) {
             for (;;) {
                 ++addAttempt;
```
idelpivnitskiy, line 648: Consider moving `++addAttempt;` after checking the state

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -533,6 +550,16 @@ void markClosed() {
             }
         }
 
+        private ConnState closeConnState() {
+            for (;;) {
+                final ConnState oldState = connState;
+                if (oldState.state == State.CLOSED || connStateUpdater.compareAndSet(this, oldState,
+                        new ConnState(oldState.connections, State.CLOSED))) {
```
idelpivnitskiy, line 557: Consider adding a comment to describe why we need to keep `oldState.connections` around even if `markClosed()` tries to close them. Would be helpful for someone new.

idelpivnitskiy (approved):
Good findings, thanks!

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: oof great catch

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
Scottmitch: `toArray(final IntFunction<T1[]> generator)` -> this is JDK11 only so we can't add it.

PR #2451, opened 2022-12-01
Remove unused BufferHolder interface

daschl:
Motivation:

The `BufferHolder` interface is currently not used and should be removed.

Modifications:

This changeset removes the interface itself and the related code checking for its type.


daschl:
Slight change of plans, hold off on review for now.

PR #2452, opened 2022-12-02
Deprecate BufferHandler and BufferHolder

daschl:
Motivation:

Both the BufferHandler and the BufferHolder did see very limited use and are scheduled for removal in the next minor release.

Modifications:

To preserve backwards compatibility and to notify potential users, both APIs are marked as deprecated now in preparation for removal.

### DIFF
--- a/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/BufferHolder.java
+++ b/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/BufferHolder.java
@@ -18,7 +18,7 @@ package io.servicetalk.buffer.api;
 /**
  * An object which contains a {@link Buffer}.
  *
- * @deprecated This API is going to be removed in the next minor version with no planned replacement. If it cannot be
+ * @deprecated This API is going to be removed in future releases with no planned replacement. If it cannot be
  *             removed from your application, consider copying it into your codebase.
  */
 @Deprecated // FIXME: 0.43 - Remove deprecation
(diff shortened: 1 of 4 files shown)


servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/BufferHolder.java
```diff
@@ -17,7 +17,11 @@
 
 /**
  * An object which contains a {@link Buffer}.
+ *
+ * @deprecated This API is going to be removed in the next minor version with no planned replacement. If it cannot be
```
idelpivnitskiy, line 21: Bcz we are not quite correct in semantic versioning until 1.0.0, consider saying "in the future releases" instead of giving more specifics on minor/major version

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/BufferHandler.java
```diff
@@ -40,8 +40,12 @@
  * </ul>
  *
  * This also releases any {@link ByteBuf} once converted to {@link Buffer}.
+ *
+ * @deprecated This API is going to be removed in the next minor version with no planned replacement. If it cannot be
```
idelpivnitskiy, line 44: Same

idelpivnitskiy (approved):
LGTM, thanks!

daschl:
@idelpivnitskiy resolved in a follow-up commit.

idelpivnitskiy:
Thanks!
If it's already approved, feel free to merge after CI finishes. Looks like you have power now :) 
We always use "squash and merge" option via GitHub UI.

PR #2454, opened 2022-12-05
Provide better exception when connection limit reached in LimitingCon…

daschl:
…nectionFactoryFilter

Motivation:

Before this change, when the LimitingConnectionFactoryFilter does not allow more connections for a given host, a generic ConnectException is returned. Developer productivity can be improved since at the moment the user needs to inspect the text message for further information and to distinguish different failure cases.

Modifications:

This changeset subclasses the ConnectException for backwards compatibility and returns a ConnectionLimitReachedException instead. The actual message is also preserved for compatibility reasons.

Result:

Better user experience when matching on the exception instance while still being backwards compatible.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionLimitReachedException.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionLimitReachedException.java
@@ -20,6 +20,8 @@ import java.net.ConnectException;
 /**
  * Thrown when the number of connections reached their limit for a given resource (i.e. a host)
  * depending on the context.
+ *
+ * @see LimitingConnectionFactoryFilter
  */
 public class ConnectionLimitReachedException extends ConnectException {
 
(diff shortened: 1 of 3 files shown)


servicetalk-client-api/src/main/java/io/servicetalk/client/api/ConnectionLimitReachedException.java
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+package io.servicetalk.client.api;
+
+import java.net.ConnectException;
+
+/**
+ * Thrown when the number of connections reached their limit for a given resource (i.e. a host)
+ * depending on the context.
+ */
```
idelpivnitskiy, line 23: Consider adding a `@see` tag with a link to `LimitingConnectionFactoryFilter` and enhancing `LimitingConnectionFactoryFilter`'s javadoc with explanation what exception type will be thrown when the limit is reached.

idelpivnitskiy (approved):

PR #2455, opened 2022-12-07
Add null checks to generated GRPC API

daschl:
Motivation:

At the moment the generated GRPC code does not perform input validation in all necessary methods and might pass invalid/null inputs down into the GRPC layer where it is harder to track down what went wrong in the first place.

Modifications:

This changeset adds explicit requireNonNull to all generated methods which accept user input and adds a test case to verify that the requireNonNull is performed at the earliest point in time and not deeper in the call stack.

Result:

Early argument null checks to improve debugability in case of invalid argument values.


idelpivnitskiy (approved):
🚀

PR #2456, opened 2022-12-07
Provide common HTTP Service interface for all API variants.

daschl:
**Motivation**

  Right now four different Service API variants are provided, all
  with their async and sync listen* overloads. If the service
  type is known at compile time those overloads suffice, but in
  situations where the type is only determined at runtime (for
  example when DI is used) the user needs to write verbose code
  that can be provided by ServiceTalk and improve the developer
  experience.

  **Modifications**

  This changeset introduces a new "Service" interface in the
  http package and it now acts as a parent for the four different
  service implementations. The HttpServerBuilder gets two more
  overloads (blocking and non-blocking) to accept the Service
  type and then internally performs the runtime type checking
  and delegates to the appropriate methods.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
@@ -32,6 +32,7 @@ import io.servicetalk.transport.api.TransportObserver;
 import java.net.SocketOption;
 import java.net.StandardSocketOptions;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.BooleanSupplier;
 import java.util.function.Predicate;
 
@@ -318,6 +319,27 @@ public interface HttpServerBuilder {
      * Starts this server and returns the {@link HttpServerContext} after the server has been successfully started.
      * <p>
      * If the underlying protocol (e.g. TCP) supports it this will result in a socket bind/listen on {@code address}.
+     * <p>
+     * Note that this method is generic in the sense that it accepts all HTTP {@link Service} implementations to be
+     * passed in, namely {@link StreamingHttpService}, {@link HttpService}, {@link BlockingStreamingHttpService} and
+     * {@link BlockingHttpService}. It is especially useful when Dependency Injection is used and the type of service
+     * is not known at compile time.
+     *
+     * @param service Service invoked for every request received by this server. The returned {@link HttpServerContext}
+     * manages the lifecycle of the {@code service}, ensuring it is closed when the {@link HttpServerContext} is closed.
+     * @return A {@link HttpServerContext} by blocking the calling thread until the server is successfully started or
+     * throws an {@link Exception} if the server could not be started.
+     * @throws IllegalArgumentException if an unknown/unsupported {@link Service} is being provided.
+     * @throws Exception if the server could not be started.
+     */
+    default HttpServerContext listenServiceAndAwait(Service service) throws Exception {
+        return blockingInvocation(listenService(service));
+    }
+
+    /**
+     * Starts this server and returns the {@link HttpServerContext} after the server has been successfully started.
+     * <p>
+     * If the underlying protocol (e.g. TCP) supports it this will result in a socket bind/listen on {@code address}.
      *
      * @param service Service invoked for every request received by this server. The returned {@link HttpServerContext}
      * manages the lifecycle of the {@code service}, ensuring it is closed when the {@link HttpServerContext} is closed.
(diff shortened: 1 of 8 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -374,6 +396,38 @@ default HttpServerContext listenBlockingStreamingAndAwait(BlockingStreamingHttpS
...
+     * is not known at compile time.
+     *
+     * @param service Service invoked for every request received by this server. The returned {@link HttpServerContext}
+     * manages the lifecycle of the {@code service}, ensuring it is closed when the {@link HttpServerContext} is closed.
+     * @return A {@link Single} that completes when the server is successfully started or terminates with an error if
+     * the server could not be started.
+     * @throws IllegalArgumentException if an unknown/unsupported {@link Service} is being provided.
+     */
+    default Single<HttpServerContext> listenService(final Service service) {
+        Objects.requireNonNull(service);
```
daschl, line 416: @idelpivnitskiy should we defer() the whole block or keep it eager as is?

idelpivnitskiy: No, in this case we don't need to defer bcz the incoming argument is checked once and never changes later, and we don't do any meaningful work here (we only checking types). `defer` is useful when we make some operations and want to re-run those operations on every re-subscribe. Example: logic that changes headers of a request.

Also, `requireNonNull` isn't necessary here, it will just fall under `IllegalArgumentException`.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -374,6 +396,38 @@ default HttpServerContext listenBlockingStreamingAndAwait(BlockingStreamingHttpS
...
+        if (service instanceof HttpService) {
+            return listen((HttpService) service);
+        } else if (service instanceof StreamingHttpService) {
+            return listenStreaming((StreamingHttpService) service);
+        } else if (service instanceof BlockingHttpService) {
+            return listenBlocking((BlockingHttpService) service);
+        } else if (service instanceof BlockingStreamingHttpService) {
+            return listenBlockingStreaming((BlockingStreamingHttpService) service);
+        } else {
+            throw new IllegalArgumentException("Unknown/Unsupported service type: " + service);
```
idelpivnitskiy, line 427: 1. Methods that return an async source should not throw (unless NPE). Instead, those methods should returns a failed source. See `Single.failed(...)`.
2. Consider keeping only one variant, let's say "Unsupported".

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -374,6 +396,38 @@ default HttpServerContext listenBlockingStreamingAndAwait(BlockingStreamingHttpS
...
+     * Note that this method is generic in the sense that it accepts all HTTP {@link Service} implementations to be
+     * passed in, namely {@link StreamingHttpService}, {@link HttpService}, {@link BlockingStreamingHttpService} and
+     * {@link BlockingHttpService}. It is especially useful when Dependency Injection is used and the type of service
+     * is not known at compile time.
+     *
+     * @param service Service invoked for every request received by this server. The returned {@link HttpServerContext}
+     * manages the lifecycle of the {@code service}, ensuring it is closed when the {@link HttpServerContext} is closed.
+     * @return A {@link Single} that completes when the server is successfully started or terminates with an error if
+     * the server could not be started.
+     * @throws IllegalArgumentException if an unknown/unsupported {@link Service} is being provided.
```
idelpivnitskiy, line 413: Consider: `if an unsupported {@link Service} type is being provided.`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -314,6 +315,27 @@ HttpServerBuilder appendServiceFilter(Predicate<StreamingHttpRequest> predicate,
...
+     * Note that this method is generic in the sense that it accepts all HTTP {@link Service} implementations to be
+     * passed in, namely {@link StreamingHttpService}, {@link HttpService}, {@link BlockingStreamingHttpService} and
+     * {@link BlockingHttpService}. It is especially useful when Dependency Injection is used and the type of service
+     * is not known at compile time.
+     *
+     * @param service Service invoked for every request received by this server. The returned {@link HttpServerContext}
+     * manages the lifecycle of the {@code service}, ensuring it is closed when the {@link HttpServerContext} is closed.
+     * @return A {@link HttpServerContext} by blocking the calling thread until the server is successfully started or
+     * throws an {@link Exception} if the server could not be started.
+     * @throws IllegalArgumentException if an unknown/unsupported {@link Service} is being provided.
```
idelpivnitskiy, line 332: Consider: `if an unsupported {@link Service} type is being provided.`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Service.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Parent interface for all available HTTP services.
+ *
+ * @see HttpService
+ * @see StreamingHttpService
+ * @see BlockingHttpService
+ * @see BlockingStreamingHttpService
+ */
+public interface Service extends HttpExecutionStrategyInfluencer {
```
idelpivnitskiy, line 26: As we discussed offline, let's consider an http-specific name, like `HttpServiceBase` or `BaseHttpService`.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+            try (HttpServerContext ctx = HttpServers.forAddress(localAddress(0)).listenServiceAndAwait(service)) {
+                ctx.closeGracefully();
+            }
+        });
+    }
+
+    static Stream<Arguments> serviceProvider() {
+        return Stream.of(
+                arguments(named("BlockingHttpService",
```
idelpivnitskiy, line 66: TIL about `named` 😮

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import static io.servicetalk.transport.netty.internal.AddressUtils.localAddress;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Named.named;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+
+public class HttpServiceTest {
+
+    @ParameterizedTest
```
idelpivnitskiy, line 42: Consider adding a name for the test in the following format:
```
@ParameterizedTest(name = "{displayName} [{index}] {0}")
```
Without this, Intellij IDEA shows the name correctly but generated HTML report that is published by CI has no name for parametrized test. This helps to fix it.

idelpivnitskiy (approved):
🚀

PR #2459, opened 2022-12-09
`RetryingHttpRequesterFilter`: rename `ofImmediate` to `ofImmediateBounded`

idelpivnitskiy:
Motivation:

It's not obvious for users that `BackOffPolicy.ofImmediate()` retries only 3 times.

Modifications:

- Deprecated `BackOffPolicy.ofImmediate()`;
- Introduce `BackOffPolicy.ofImmediateBounded()` that behaves the same way;
- Cache the returning value for `BackOffPolicy.ofImmediateBounded()`;

Result:

Easier for users to understand that immediate `BackOffPolicy` has a limit for number of retries.


daschl (approved):

idelpivnitskiy:
Discussed with @tkountis offline and he recommended another name: `ofImmediateBounded()` instead of `ofImmediateWithDefaultLimit()`

PR #2460, opened 2022-12-09
Remove ServiceAdapterHolder from DefaultHttpServerBuilder

daschl:
Motivation:

The ServiceAdapterHolder is not needed anymore conceptually since the Service now carries its offloading properties (HttpExecutionStrategy) as part of the class itself.

Modifications:

This changeset is the first step to remove the ServiceAdapterHolder by removing it from the DefaultHttpServerBuilder. The HttpExecutionStrategy is still computed but instead of going through the ServiceAdapterHolder indirection the codepath now consistently uses the StreamingHttpService codepath.

Result:

Simplified DefaultHttpServerBuilder as well as removing most of the current ServiceAdapterHolder usage (there is still some left which can be cleaned up in follow-up changesets).

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpApiConversions.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpApiConversions.java
@@ -288,7 +288,7 @@ public final class HttpApiConversions {
      * @param influencer {@link HttpExecutionStrategyInfluencer} to influence the strategy for invoking the resulting
      * {@link StreamingHttpService}.
      * @return {@link ServiceAdapterHolder} containing the service adapted to the streaming programming model.
-     * @deprecated Use overload with {@link HttpExecutionStrategy} rather than {@link HttpExecutionStrategyInfluencer}
+     * @deprecated Use {@link #toStreamingHttpService(HttpExecutionStrategy, HttpService)} instead.
      */
     @Deprecated // FIXME: 0.43 - remove deprecated method
     public static ServiceAdapterHolder toStreamingHttpService(HttpService service,
@@ -297,7 +297,7 @@ public final class HttpApiConversions {
     }
 
     /**
-     * Convert from a {@link HttpService} to a {@link ServiceAdapterHolder}.
+     * Convert from a {@link HttpService} to a {@link StreamingHttpService}.
      *
      * @param service The {@link HttpService} to convert.
      * @param strategy required strategy for the service when invoking the resulting {@link ServiceAdapterHolder}.
@@ -314,7 +314,7 @@ public final class HttpApiConversions {
      *
      * @param service The {@link HttpService} to convert.
      * @param strategy required strategy for the service when invoking the resulting {@link StreamingHttpService}.
-     * @return {@link StreamingHttpService} containing the service adapted to the streaming programming model.
+     * @return the converted {@link StreamingHttpService} to be used for the streaming programming model.
      */
     public static StreamingHttpService toStreamingHttpService(HttpExecutionStrategy strategy, HttpService service) {
         return new ServiceToStreamingService(service, strategy);
@@ -327,7 +327,7 @@ public final class HttpApiConversions {
      * @param influencer {@link HttpExecutionStrategyInfluencer} to influence the strategy for invoking the resulting
      * {@link StreamingHttpService}.
      * @return {@link ServiceAdapterHolder} containing the service adapted to the streaming programming model.
-     * @deprecated Use overload with {@link HttpExecutionStrategy} rather than {@link HttpExecutionStrategyInfluencer}
+     * @deprecated Use {@link #toStreamingHttpService(HttpExecutionStrategy, BlockingStreamingHttpService)} instead.
      */
     @Deprecated // FIXME: 0.43 - remove deprecated method
     public static ServiceAdapterHolder toStreamingHttpService(BlockingStreamingHttpService service,
@@ -336,7 +336,7 @@ public final class HttpApiConversions {
     }
 
     /**
-     * Convert from a {@link BlockingStreamingHttpService} to a {@link ServiceAdapterHolder}.
+     * Convert from a {@link BlockingStreamingHttpService} to a {@link StreamingHttpService}.
      *
      * @param service The {@link BlockingStreamingHttpService} to convert.
      * @param strategy required strategy for the service when invoking the resulting {@link StreamingHttpService}.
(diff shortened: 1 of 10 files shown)


daschl:
@idelpivnitskiy this change has one thing to consider/discuss: The change as applied right now is backwards incompatible in the sense that .adaptor() is not available from outside of the package anymore. If we decide that the class is internal only and not worth bothering I'm good with that. The other two options (I think) are:

- As proposed offline, add new methods with different arg order and use those and leave the others alone
- Make the concrete classes (i.e. `BlockingToStreamingService` public) which would solve the accessibility issue as well.

If one of the approaches above is chosen, the test case changes can be reverted I think.

daschl:
Updated per offline comments and ready to review.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpApiConversions.java
```diff
@@ -290,24 +290,36 @@ public static BlockingStreamingHttpClient toBlockingStreamingClient(StreamingHtt
      * @return {@link ServiceAdapterHolder} containing the service adapted to the streaming programming model.
      * @deprecated Use overload with {@link HttpExecutionStrategy} rather than {@link HttpExecutionStrategyInfluencer}
```
idelpivnitskiy, line 291: While you are here, can you please update this description with a link to the correct method?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpApiConversions.java
```diff
@@ -353,17 +378,31 @@ public static ServiceAdapterHolder toStreamingHttpService(BlockingHttpService se
     }
 
     /**
-     * Convert from a {@link BlockingStreamingHttpService} to a {@link StreamingHttpService}.
+     * Convert from a {@link BlockingStreamingHttpService} to a {@link ServiceAdapterHolder}.
```
idelpivnitskiy, line 381: BlockingStreamingHttpService -> BlockingHttpService, same in lines 365, 367, 383, 395, 398

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -257,7 +256,8 @@ public HttpServerBuilder bufferAllocator(final BufferAllocator allocator) {
 
     @Override
     public Single<HttpServerContext> listen(final HttpService service) {
-        return listenForAdapter(toStreamingHttpService(service, computeServiceStrategy(HttpService.class, service)));
+        HttpExecutionStrategy executionStrategy = computeServiceStrategy(HttpService.class, service);
+        return listenForService(toStreamingHttpService(executionStrategy, service), executionStrategy);
```
idelpivnitskiy, line 260: I'm puzzled why none of the tests caught this, but this is incorrect. Previous code of `listenForAdapter` used `adapterHolder.serviceInvocationStrategy()` which could be modified further by the conversion layer:
```java
    ServiceToStreamingService(final HttpService original, HttpExecutionStrategy strategy) {
        super(defaultStrategy() == strategy ? DEFAULT_STRATEGY : strategy);
    }
```
Going forward, we need to use `requireOffloads()` method of the converted `StreamingHttpService` instead of `serviceInvocationStrategy()`.

idelpivnitskiy (changes requested):
TBH, it will be easier to review all associated changes in one PR instead of splitting in into multiple. With many PRs it's harder to understand what is forgotten, what will be fixed in a follow-up, and how many follow-ups to expect. 
I would just go ahead, deprecate `ServiceAdapterHolder`, and make necessary changes in `HttpPredicateRouterBuilder`, `GrpcRouter`, and `BlockingStreamingToStreamingServiceTest`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
idelpivnitskiy: The name is confusing, but `defaultStrategy() != DEFAULT_STRATEGY ` 🤯

daschl: I think I found the root cause of the issue: https://github.com/apple/servicetalk/pull/2460/files#diff-e1735bd3a1c16ec930f09a406521ed467238de450ea9a921adb469741ffd651bR39

The `ServiceAdapterHolder` did not override the `requiredOffloads()` so it gave conflicting strategies when asked for it via `requiredOffloads ` or `serviceInvocationStrategy `. If not overridden, it would always return from the parent:

```
        // safe default--implementations are expected to override
        return HttpExecutionStrategies.offloadAll();
    }
```

idelpivnitskiy: Correct, we have to override `AbstractServiceAdapterHolder.requiredOffloads()` and use it instead of `serviceInvocationStrategy()`. That's one part.

The second part is `listenForAdapter` replacement. The current changes are not identical to how it behaved before. `listenForAdapter` used `adapterHolder.serviceInvocationStrategy()` for the 2nd argument of `listenForService`, but now you use `computeServiceStrategy`, which is not the same.
Consider removing `HttpExecutionStrategy` as an argument for `listenForService`. Instead, do this as the first line inside `listenForService`:
```java
final HttpExecutionStrategy computedStrategy = rawService.requiredOffloads();
```

idelpivnitskiy:
Everything else looks good, except the one unresolved conversation above.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
daschl: Applied with latest commit.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -268,13 +269,15 @@ public Single<HttpServerContext> listenStreaming(final StreamingHttpService serv
...
     public Single<HttpServerContext> listenBlocking(final BlockingHttpService service) {
         HttpExecutionStrategy executionStrategy = computeServiceStrategy(BlockingHttpService.class, service);
-        return listenForService(toStreamingHttpService(executionStrategy, service), executionStrategy);
+        StreamingHttpService streamingService = toStreamingHttpService(executionStrategy, service);
+        return listenForService(streamingService, streamingService.requiredOffloads());
     }
 
     @Override
     public Single<HttpServerContext> listenBlockingStreaming(final BlockingStreamingHttpService service) {
         HttpExecutionStrategy executionStrategy = computeServiceStrategy(BlockingStreamingHttpService.class, service);
```
idelpivnitskiy, line 278: Since the `executionStrategy` is used only once, consider inlining it instead of using a local variable.

idelpivnitskiy (approved):
Thank you!

PR #2467, opened 2022-12-13
Perform GRPC route building lazily on bind

daschl:
Motivation:

At the moment building a GRPC route is performed right when the application registers a new one at the high level API.

Since every route needs an execution strategy when being built, the execution strategy needs to be provided at construction time of the ServiceFactory/Builder. In order to improve the user experience and to make the execution strategy an optional builder method the building of the routes needs to be moved to a later point in the build stage.

Modifications:

This changeset, in preparation to making the execution strategy optional in a later commit, makes the route building lazy by moving it inside a closure which is executed at "bind" time instead of being evaluated immediately.

To preserve the current semantics (especially that an error is thrown if two routes on the same path are provided by the user), the error reporting is still performed eagerly. The merge code also had to be changed slightly to take this invariant into account.

Result:

Lazy GRPC route building in preparation to make the execution strategy optional.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
@@ -32,14 +32,18 @@ import io.servicetalk.router.api.RouteExecutionStrategyFactory;
 import io.servicetalk.transport.api.ExecutionContext;
 
 import java.lang.reflect.Method;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.function.Consumer;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.api.Completable.completed;
 import static io.servicetalk.grpc.api.GrpcExecutionStrategies.offloadNever;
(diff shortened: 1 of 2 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
```diff
@@ -340,7 +371,11 @@ protected final <Req, Resp> void addStreamingRoute(// FIXME: 0.43 - remove depre
     protected final <Req, Resp> void addStreamingRoute(
             final GrpcExecutionStrategy executionStrategy, MethodDescriptor<Req, Resp> methodDescriptor,
             BufferDecoderGroup decompressors, List<BufferEncoder> compressors, StreamingRoute<Req, Resp> route) {
-        routeBuilder.addStreamingRoute(methodDescriptor, decompressors, compressors, executionStrategy, route);
+        final String path = methodDescriptor.httpPath();
+        verifyNoOverrides(null, methodDescriptor.httpPath(), deferredRoutes);
```
idelpivnitskiy, line 375: `methodDescriptor.httpPath() -> path`

idelpivnitskiy (approved):

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcRoutes.java
daschl: 👍 spotted it in another place as well.

PR #2468, opened 2022-12-15
Make RouteExecutionStrategyFactory optional on Builder

daschl:
Motivation:

Now that route building has been made lazy in a previous commit, the RouteExecutionStrategyFactory can be moved to an optional builder argument and therefore clean up the API.

Modifications:

In addition to making RouteExecutionStrategyFactory available on the Builder, the corresponding constructors have been marked as Deprecated since they are not needed anymore and can be removed in a future version.

NOTE: the test only asserts that the custom factory is called, due to a different potential bug identified we cannot assert that custom offloading parameters are actually applied (yet). Once resolved, this will be implemented in a follow-up commit.

Result:

Simpler user-facing API when building GRPC services.


servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -728,6 +739,10 @@ private TypeSpec.Builder addServiceFactory(final State state, final TypeSpec.Bui
                         .addJavadoc(JAVADOC_PARAM + strategyFactory +
                                 " a factory that creates an execution strategy for different {@link $L#id() id}s" +
                                 lineSeparator(), RouteExecutionStrategy)
+                        .addJavadoc(JAVADOC_DEPRECATED + "Use {@link $L#$L()} and set the custom strategy " +
+                                "on {@link $L#$L($T)} instead." + lineSeparator(), Builder, Builder, Builder,
```
idelpivnitskiy, line 743: Consider also changing internals of this constructor as suggested in the javadoc. It will be easier for us to remove deprecated code with less changes later.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -792,6 +807,10 @@ supportedMessageCodings, serviceFactoryBuilderInitChain(state.serviceProto, fals
                         .addJavadoc(JAVADOC_PARAM + strategyFactory +
                                 " a factory that creates an execution strategy for different {@link $L#id() id}s" +
                                 lineSeparator(), RouteExecutionStrategy)
+                        .addJavadoc(JAVADOC_DEPRECATED + "Use {@link $L#$L()} and set the custom strategy " +
+                                        "on {@link $L#$L($T)} instead." + lineSeparator(), Builder, Builder, Builder,
```
idelpivnitskiy, line 811: Same here

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ServiceFactoryCustomizationTest.java
```diff
@@ -0,0 +1,56 @@
...
+class ServiceFactoryCustomizationTest {
+
+    @Test
+    void usesCustomRouteExecutionStrategyFactory() throws Exception {
+        final AtomicBoolean customFactoryCalled = new AtomicBoolean(false);
+
+        GrpcServers
+                .forAddress(localAddress(0))
+                .listenAndAwait(new Greeter.ServiceFactory.Builder()
+                        .routeExecutionStrategyFactory(id -> {
```
idelpivnitskiy, line 41: Instead of creating a new test class, consider modifying the existing `ExecutionStrategyTest`. After your changes, its `RouteExecutionStrategy` uses deprecated ctors. It will be great to migrate those to new API.
However, it's also great to make sure that deprecated API still works as expected. So, instead of just migrating to the new API, consider also keeping the deprecated variant for now.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
idelpivnitskiy: Actually, this ctor is also deprecated. Nvm. Let's keep as-is.

PR #2470, opened 2022-12-17
Do not apply `InternalRetryingHttpClientFilter` for reserved connections

idelpivnitskiy:
Motivation:

Reserved connections don't have concurrency controller. Users are responsible for managing traffic through the connection themselves. If they violate Netty settings, they may get `MaxConcurrentStreamsViolatedStacklessHttp2Exception`. We should not silently retry in this case.

Modifications:

- Implement `InternalRetryingHttpClientFilter#reserveConnection(...)` method to simply delegate invocation without applying any filtering;

Result:

The retrying logic of the `request(...)` method won't be applied by `StreamingHttpClientFilter#reserveConnection(...)`.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
@@ -26,8 +26,10 @@ import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.concurrent.api.internal.SubscribableCompletable;
+import io.servicetalk.http.api.FilterableReservedStreamingHttpConnection;
 import io.servicetalk.http.api.FilterableStreamingHttpClient;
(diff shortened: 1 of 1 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
```diff
@@ -317,6 +319,14 @@ protected Single<StreamingHttpResponse> request(final StreamingHttpRequester del
                             .retry((count, t) -> count <= 32 &&
                                     t instanceof MaxConcurrentStreamsViolatedStacklessHttp2Exception);
                 }
+
+                @Override
+                public Single<? extends FilterableReservedStreamingHttpConnection> reserveConnection(
+                        final HttpRequestMetaData metaData) {
+                    // Pass through. Concurrency controller isn't applied for reserved connections. Users have to manage
+                    // all traffic themselves and should see all exceptions from Netty.
+                    return delegate().reserveConnection(metaData);
```
idelpivnitskiy, line 328: Caught it when we went through #2461 with @mgodave 🎉

tkountis (approved):

PR #2473, opened 2022-12-23
Fix increased latency for gRPC after adding `ListenableAsyncCloseable.onClosing()`

idelpivnitskiy:
Motivation:

#2430 adds `ListenableAsyncCloseable.onClosing()` method. Its implementations require allocation of a `CompletableProcessor`.
Because we wrap channel's `EventLoop` with `IoExecutor` on every new connection and every new HTTP/2 stream, it caused significant increase of allocations and pressure for GC. As the result, latency increased for all HTTP/2 and gRPC use-cases.

Modifications:

- Add `ExecutionContextUtils` that adapt `EventLoop` to `IoExecutor` API for `ExecutionContext` and cache its value in `FastThreadLocal` to reduce allocations;
- Make sure that HTTP/1.x and parent HTTP/2 connections use `ExecutionContextUtils`;
- For each new HTTP/2 stream (child channel) take execution context from the parent channel because child channel uses the same `EventLoop` as its parent channel;
- Add a new `DefaultNettyConnection.initChannel` overload that takes already constructed `ExecutionContext`;
- Adjust existing tests for new API;

Result:

Significantly reduced allocation rate and less GC pauses for HTTP/2 and gRPC.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ExecutionContextUtils.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ExecutionContextUtils.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright © 2022 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+import io.servicetalk.http.api.DelegatingHttpExecutionContext;
+import io.servicetalk.http.api.HttpExecutionContext;
+import io.servicetalk.transport.api.IoExecutor;
+
+import io.netty.channel.Channel;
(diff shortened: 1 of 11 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ExecutionContextUtils.java
```diff
@@ -0,0 +1,62 @@
...
+                builderExecutionContext.ioExecutor().isIoThreadSupported());
+        return new DelegatingHttpExecutionContext(builderExecutionContext) {
+            @Override
+            public IoExecutor ioExecutor() {
+                return channelIoExecutor;
+            }
+        };
+    }
+
+    private static IoExecutor fromChannel(final Channel channel, boolean isIoThreadSupported) {
```
daschl, line 47: Could it happen that isIoThreadSupported changes from true/false on the same thread through the HttpExecutionContext? If so the cached IoExecutor might be the wrong one that is returned?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ExecutionContextUtils.java
```diff
@@ -0,0 +1,62 @@
...
+
+final class ExecutionContextUtils {
+
+    private static final FastThreadLocal<IoExecutor> CHANNEL_IO_EXECUTOR = new FastThreadLocal<>();
+
+    private ExecutionContextUtils() {
+        // No instances
+    }
+
+    static HttpExecutionContext channelExecutionContext(final Channel channel,
```
daschl, line 35: Since this class is part of an optimization, for future-self and to avoid perf regressions consider some javadocs to explain the rationale and use case

daschl (approved):
Looks good to me, just one clarification and one javadoc nit

PR #2476, opened 2022-12-27
Allow to open a new connection when reserving (HTTP).

daschl:
Motivation:

When a connection is being reserved on one of the HttpClient variants, at the moment the user has no control over the decision if an existing connection should be reused or a new one will be opened.

Use cases have been reported where being able to explicitly reserve a new connection instead of reusing one would be benefitial.

Modifications:

This changeset adds support for this feature by introducing two new APIs at different levels:

 - At the HttpClient level a HTTP_FORCE_NEW_CONNECTION context key has been added which signals the intention at the reserveConnection level by the user.
 - The LoadBalancer now features a newConnection method which by default throws an UnsupportedOperationException for backwards compatibility.

The RoundRobinLoadBalancer implements the newConnection API by reusing all of the existing internal functionality (only skipping the connection reuse part and having to explicitly reserve since no predicate is present).

Result:

The user can now explicitly ask for a new connection when reserving a connection instead of potentially receiving an already opened one.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
@@ -48,7 +48,7 @@ public final class HttpContextKeys {
             newKey("HTTP_TARGET_ADDRESS_BEHIND_PROXY", Object.class);
 
     /**
-     * If set to true, forces creating a new connection versus potentially reusing an already established one.
+     * If set to true, forces creating a new connection versus potentially selecting an already established one.
      * <p>
      * This key is only available when reserving a connection and will be ignored when performing a regular
      * request on the client (for example through {@link HttpClient#request(HttpRequest)}).
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
@@ -47,6 +47,17 @@ public final class HttpContextKeys {
     public static final Key<Object> HTTP_TARGET_ADDRESS_BEHIND_PROXY =
             newKey("HTTP_TARGET_ADDRESS_BEHIND_PROXY", Object.class);
 
+    /**
+     * If set to true, forces creating a new connection versus potentially reusing an already established one.
+     * <p>
+     * This key is only available when reserving a connection and will be ignored when performing a regular
+     * request on the client (for example through {@link HttpClient#request(HttpRequest)}).
+     *
+     * @see HttpClient#reserveConnection(HttpRequestMetaData)
+     */
+    public static final Key<Boolean> HTTP_FORCE_NEW_CONNECTION =
+            newKey("HTTP_FORCE_NEW_CONNECTION", Boolean.class);
+
     private HttpContextKeys() {
         // No instances
     }
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancedStreamingHttpClient.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancedStreamingHttpClient.java
@@ -52,7 +52,6 @@ import static io.servicetalk.http.netty.AbstractStreamingHttpConnection.requestE
 import static io.servicetalk.http.netty.LoadBalancedStreamingHttpClient.OnStreamClosedRunnable.areStreamsSupported;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.atomic.AtomicIntegerFieldUpdater.newUpdater;
-import static java.util.function.Function.identity;
 
 final class LoadBalancedStreamingHttpClient implements FilterableStreamingHttpClient {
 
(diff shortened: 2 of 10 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
```diff
@@ -47,6 +47,17 @@ public final class HttpContextKeys {
     public static final Key<Object> HTTP_TARGET_ADDRESS_BEHIND_PROXY =
             newKey("HTTP_TARGET_ADDRESS_BEHIND_PROXY", Object.class);
 
+    /**
+     * If set to true, forces creating a new connection versus potentially reusing an already established one.
```
idelpivnitskiy, line 51:
```suggestion
     * If set to true, forces creating a new connection versus potentially selecting an already established one.
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancedStreamingHttpClient.java
```diff
@@ -153,8 +155,16 @@ private static void notifyConnectionSelected(final HttpRequestMetaData requestMe
     @Override
     public Single<ReservedStreamingHttpConnection> reserveConnection(final HttpRequestMetaData metaData) {
         return Single.defer(() -> {
-            Single<ReservedStreamingHttpConnection> connection =
-                    loadBalancer.selectConnection(SELECTOR_FOR_RESERVE, metaData.context()).map(identity());
+            final ContextMap context = metaData.context();
+            final Boolean forceNew = context.get(HttpContextKeys.HTTP_FORCE_NEW_CONNECTION);
```
idelpivnitskiy, line 159: Consider using `Boolean.TRUE.equals(context.get(HttpContextKeys.HTTP_FORCE_NEW_CONNECTION))`. This way we can skip autoboxing and use `?:` instead of `if` below

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
```diff
@@ -753,6 +754,33 @@ void handleAllDiscoveryEvents() throws Exception {
...
 
+    /**
+     * This test verifies that the {@link io.servicetalk.client.api.LoadBalancer#newConnection(ContextMap)} API is
+     * supported.
+     */
+    @Test
+    void registersNewConnections() throws Exception {
+        serviceDiscoveryPublisher.onComplete();
+
+        lb = (RoundRobinLoadBalancer<String, TestLoadBalancedConnection>)
```
idelpivnitskiy, line 765: consider `lb = defaultLb();`

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
```diff
@@ -753,6 +754,33 @@ void handleAllDiscoveryEvents() throws Exception {
...
+        assertAddresses(lb.usedAddresses(), EMPTY_ARRAY);
+
+        sendServiceDiscoveryEvents(upEvent("address-1"));
+        assertAddresses(lb.usedAddresses(), "address-1");
+
+        assertTrue(connectionsCreated.isEmpty());
+        lb.newConnection(null).toFuture().get();
+        assertEquals(1, connectionsCreated.size());
+        lb.newConnection(null).toFuture().get();
+        assertEquals(2, connectionsCreated.size());
```
idelpivnitskiy, line 781: Consider verifying that the connection was reserved

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -431,6 +439,13 @@ private Single<C> selectConnection0(final Predicate<C> selector, @Nullable final
         }
         return establishConnection
                 .flatMap(newCnx -> {
+                    if (forceNewConnectionAndReserve && !newCnx.tryReserve()) {
+                        return failed(StacklessConnectionRejectedException.newInstance(
```
idelpivnitskiy, line 443: This will leak the connection. We have to follow the same pattern as below for other exceptions:
```
return newCnx.closeAsync().concat(failed(...));
```

idelpivnitskiy (commented):
LGTM after comments addressed:

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerTest.java
```diff
@@ -761,24 +763,21 @@ void handleAllDiscoveryEvents() throws Exception {
...
         assertTrue(connectionsCreated.isEmpty());
-        lb.newConnection(null).toFuture().get();
+        TestLoadBalancedConnection conn1 = lb.newConnection(null).toFuture().get();
         assertEquals(1, connectionsCreated.size());
-        lb.newConnection(null).toFuture().get();
+        TestLoadBalancedConnection conn2 = lb.newConnection(null).toFuture().get();
         assertEquals(2, connectionsCreated.size());
+
+        verify(conn1, atLeastOnce()).tryReserve();
+        verify(conn2, atLeastOnce()).tryReserve();
```
idelpivnitskiy, line 780: Should we expect only one invocation?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancedStreamingHttpClient.java
```diff
@@ -153,8 +155,13 @@ private static void notifyConnectionSelected(final HttpRequestMetaData requestMe
...
     public Single<ReservedStreamingHttpConnection> reserveConnection(final HttpRequestMetaData metaData) {
         return Single.defer(() -> {
-            Single<ReservedStreamingHttpConnection> connection =
-                    loadBalancer.selectConnection(SELECTOR_FOR_RESERVE, metaData.context()).map(identity());
+            final ContextMap context = metaData.context();
+            final boolean forceNew = Boolean.TRUE.equals(context.get(HttpContextKeys.HTTP_FORCE_NEW_CONNECTION));
+
+            Single<ReservedStreamingHttpConnection> connection = forceNew ?
+                    loadBalancer.newConnection(context).map(identity()) :
+                    loadBalancer.selectConnection(SELECTOR_FOR_RESERVE, context).map(identity());
```
idelpivnitskiy, line 163:
```suggestion
            Single<FilterableStreamingHttpLoadBalancedConnection> connection = forceNew ?
                    loadBalancer.newConnection(context) :
                    loadBalancer.selectConnection(SELECTOR_FOR_RESERVE, context);
```

idelpivnitskiy (approved):

PR #2478, opened 2022-12-30
Introduce GrpcExecutionStrategies#offloadNone

daschl:
Motivation:

To ensure partiy with HttpExecutionStrategies it makes sense to introduce "offloadNone" as a static utility method instead of only making it accessible through a builder.

Modifications:

This changeset adds the static "offloadNone" utility method in GrpcExecutionStrategies and clarifies the javadoc of "offloadNever" (which is deprecated) on alternative usage.

Result:

Feature parity between GRPC and HTTP on the "offloadNone" execution strategy creation.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
@@ -68,7 +68,7 @@ public final class GrpcExecutionStrategies {
      *
      * @return {@link GrpcExecutionStrategy} that disables all request-response path offloads.
      * @see #offloadNone()
-     * @deprecated Use a custom strategy with no offloads instead; (for example {@link #offloadNone()})
+     * @deprecated Use {@link #offloadNone()} instead.
      */
     // FIXME: 0.43 - remove deprecated method
     @Deprecated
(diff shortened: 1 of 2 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcExecutionStrategies.java
```diff
@@ -64,15 +67,28 @@ public static GrpcExecutionStrategy defaultStrategy() {
      * When merged with another execution strategy the result is always this strategy.
      *
      * @return {@link GrpcExecutionStrategy} that disables all request-response path offloads.
-     * @deprecated Use a custom strategy with no offloads instead;
-     * {@code GrpcExecutionStrategies.customStrategyBuilder().offloadNone().build()}
+     * @see #offloadNone()
+     * @deprecated Use a custom strategy with no offloads instead; (for example {@link #offloadNone()})
```
idelpivnitskiy, line 71:
```suggestion
     * @deprecated Use {@link #offloadNone()} instead.
```

idelpivnitskiy (changes requested):
We should also improve `GrpcExecutionStrategy.from(HttpExecutionStrategy)` to account for the new `offloadNone()` strategy.

idelpivnitskiy (approved):
🚢

PR #2479, opened 2023-01-02
Deprecate ConnectionAcceptor(+ Factory) "append"

daschl:
Motivation:

At the moment the user experience on how to apply multiple ConnectionAcceptorFactories is confusing because there are many different ways on how to perform this task.

In order to improve the ambiguity there should only be one common way to append multiple acceptor factories.

Modifications:

This changeset deprecates the append methods on both the ConnectionAcceptor and the ConnectionAcceptorFactory. A new ConnectionAcceptorFactoryAppender is added which is used internally and in tests, but can be used by a user who needs to port functionality forward when relying on the old behavior.

All internal use of the deprecated APIs has been replaced with the helper class in preparation for later removal.

Result:

The primary API to chain the acceptors should be through the regular HttpServerBuilder which is now expressed through deprecation and later removal of the other APIs.


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionAcceptorFactoryAppender.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A special {@link ConnectionAcceptorFactory} which allows to append two factories together.
+ * <p>
+ * Note that both factories' execution strategies are merged together and that result is used when
+ * {@link #requiredOffloads()} is called.
+ */
+public final class ConnectionAcceptorFactoryAppender implements ConnectionAcceptorFactory {
```
idelpivnitskiy, line 26: By exposing a new public class we don't reduce number of ways how to append ConnectionAcceptorFactories. The expectation was that `HttpServerBuilder#appendConnectionAcceptorFilter(ConnectionAcceptorFactory)` will be the only way to append acceptors, working in a way similar to `appendServiceFilter`

idelpivnitskiy (changes requested):

idelpivnitskiy:
Let's finish with an early connection acceptor before proceeding with this PR. Maybe we won't need to make any changes here if we come up with new interfaces and deprecate `ConnectionAcceptor`

daschl:
Will abandon this one for a new one which just deprecates on top of the new early and late acceptors.

PR #2480, opened 2023-01-03
Deprecate ConnectionFactoryFilter "append"

daschl:
Motivation:

At the moment the user experience on how to apply multiple ConnectionFactoryFilters is confusing because there are multiple ways on how to perform this task.

In order to improve the ambiguity there should only be one common way to append multiple acceptor factories.

Modifications:

This changeset deprecates the append methods on the ConnectionFactoryFilter. A new ConnectionFactoryFilterAppender is added which is used internally and in tests, but can be used by a user who needs to port functionality forward when relying on the old behavior.

All internal use of the deprecated APIs has been replaced with the helper class in preparation for later removal.

Result:

The primary API to chain the factories should be through the regular HttpClientBuilder which is now expressed through deprecation and later removal of the other APIs.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -521,7 +522,8 @@ private static StreamingHttpConnectionFilterFactory appendConnectionFilter(
     @Override
     public DefaultSingleAddressHttpClientBuilder<U, R> appendConnectionFactoryFilter(
             final ConnectionFactoryFilter<R, FilterableStreamingHttpConnection> factory) {
-        connectionFactoryFilter = connectionFactoryFilter.append(requireNonNull(factory));
+        connectionFactoryFilter = new ConnectionFactoryFilterAppender<>(connectionFactoryFilter,
+                requireNonNull(factory));
         strategyComputation.add(factory);
```
idelpivnitskiy, line 527: Same here, the expectation is that `appendConnectionFactoryFilter` will be the only way to append them. Compare to the connection accepted case, this one already has `strategyComputation.add(factory)` that correctly computes the resulting strategy. We only need to sequence factories correctly.

idelpivnitskiy (changes requested):

idelpivnitskiy (approved):
Thanks!

PR #2482, opened 2023-01-05
Consider request strategy when we schedule timeouts or retries

idelpivnitskiy:
Motivation:

Currently, timeout and retrying filter uses an `HttpExecutionStrategy` from the `ExecutionContext`. However, on the requester side users can either use a different API that optimizes the pre-computed strategy or they can put a custom strategy per request.

Modifications:

- For `RetryingHttpRequesterFilter`, compute `alternativeTimerExecutor` based on `HTTP_EXECUTION_STRATEGY_KEY`;
- For `TimeoutHttpRequesterFilter`, compute `useForTimeout` based on `HTTP_EXECUTION_STRATEGY_KEY`;

Result:

1. On the client-side, the actual `HttpExecutionStrategy` associated with the request is considered.
2. Less use of offloading executor when there is no need to offload.

### DIFF
--- a/servicetalk-http-utils/src/test/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilterTest.java
+++ b/servicetalk-http-utils/src/test/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilterTest.java
@@ -17,6 +17,7 @@ package io.servicetalk.http.utils;
 
 import io.servicetalk.concurrent.TimeSource;
 import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.context.api.ContextMap;
 import io.servicetalk.http.api.DefaultHttpExecutionContext;
 import io.servicetalk.http.api.FilterableStreamingHttpConnection;
 import io.servicetalk.http.api.HttpConnectionContext;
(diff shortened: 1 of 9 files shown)


servicetalk-http-utils/src/test/java/io/servicetalk/http/utils/TimeoutHttpRequesterFilterTest.java
```diff
@@ -72,6 +74,10 @@ private static Single<StreamingHttpResponse> applyFilter(TimeoutHttpRequesterFil
         when(connection.request(any())).thenReturn(responseSingle);
 
         StreamingHttpRequester requester = filterFactory.create(connection);
-        return requester.request(mock(StreamingHttpRequest.class));
+        StreamingHttpRequest request = mock(StreamingHttpRequest.class);
+        ContextMap requestContext = mock(ContextMap.class);
+        when(request.context()).thenReturn(requestContext);
+        when(requestContext.getOrDefault(HTTP_EXECUTION_STRATEGY_KEY, strategy)).thenReturn(strategy);
```
daschl, line 80: Would it make sense to add a test case which asserts the newly added functionality (so one where HTTP_EXECUTION_STRATEGY_KEY is present and applied correctly?)

idelpivnitskiy: I made an attempt but then realized that with mocks it won't really make any difference because it's anyway have to go through `requestContext.getOrDefault(...)`. So, having the value for `HTTP_EXECUTION_STRATEGY_KEY` or fallback to default results in the same line for mock

daschl: 👍

daschl (approved):

PR #2486, opened 2023-01-11
Introduce Early and Late ConnectionAcceptors

daschl:
Motivation:

The currently available ConnectionAcceptor has the following shortcomings which are attempted to be
addressed with this changeset:

 - Not being able to reject connections before the TLS handshake is performed.
 - Complex semantics how to append / chain multiple Acceptors together for the same service.

Modifications:

The early and late stages in accepting a connection are now broken out into two separate interfaces.

Each acceptor now only knows itself and the appending of multiple acceptors is performed by the HttpServerBuilder, simplifying appending multiple acceptors from a user perspective.

Since they are different interfaces, different offloading strategies can be put in place for each - still defaulting to full offloading.

Note: this changeset does not deprecate the "old"
acceptor, this will be done in a follow-up commit.

Result:

Improved functionality and user experience when needing to reject connections early as well as chaining multiple acceptors together.


servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
```diff
@@ -169,6 +175,84 @@ protected void handleSubscribe(Subscriber<? super ServerContext> subscriber) {
...
+                }
+
+                @Override
+                public ExecutionContext<?> executionContext() {
+                    return executionContext;
+                }
+            };
+
+            final boolean ioThreadSupported = executionContext.ioExecutor().isIoThreadSupported();
+            final NettyIoExecutor ioExecutor = fromNettyEventLoop(channel.eventLoop(), ioThreadSupported);
```
idelpivnitskiy, line 218: Looks like we can move `ExecutionContextUtils.fromChannel` to transport-netty-internal in order to reuse here, but won't be able to move `channelExecutionContext` due to dependency on `HttpExecutionContext`. It's ok, we can have a similar utility that uses `DelegatingExecutionContext`.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
```diff
@@ -169,6 +175,84 @@ protected void handleSubscribe(Subscriber<? super ServerContext> subscriber) {
...
+                }
+
+                @Override
+                public SocketAddress remoteAddress() {
+                    return channel.remoteAddress();
+                }
+
+                @Override
+                public ExecutionContext<?> executionContext() {
+                    return executionContext;
```
idelpivnitskiy, line 213: Here we should use the result of the utility that references an IoExecutor backed by the channel's event loop. The goal is to give access only to the resources related to the connection that users are going to accept or reject.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
```diff
@@ -169,6 +175,84 @@ protected void handleSubscribe(Subscriber<? super ServerContext> subscriber) {
...
+            Single<CC> connection,
+            final Channel channel,
+            final ExecutionContext<?> executionContext,
+            @Nullable final EarlyConnectionAcceptor earlyConnectionAcceptor,
+            @Nullable final LateConnectionAcceptor lateConnectionAcceptor,
+            @Nullable final InfluencerConnectionAcceptor connectionAcceptor) {
+        final Executor offloadExecutor = executionContext.executor();
+
+        if (earlyConnectionAcceptor != null) {
+            ReducedConnectionInfo info = new ReducedConnectionInfo() {
```
idelpivnitskiy, line 200: It feels like only `sslSession()` will always be null for `EarlyConnectionAcceptor`. Other fields can be useful to make the decision:
- `sslConfig` signals if the connection is going to have a handshake or not. We can say that it's always known when you build the server, but in practice users responsible for `EarlyConnectionAcceptor` and the server builder can be different people. Another use-case is that `SslConfig` can be dynamic and return different SSL options over time. It's nice to have access to it from the acceptor.
- `socketOptions` - these can also be dynamic in practice (not something that we offer right now, but we can in the future) and influence decisions;
- `protocol` - server can listen for both UDP/QUIC and TCP when we add HTTP/3 support. Users can decide to accept only a certain amount of connections for each protocol;

WDYT if we reuse the fill `ConnectionInfo` interface here?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/EarlyConnectionAcceptor.java
```diff
@@ -0,0 +1,51 @@
...
+@FunctionalInterface
+public interface EarlyConnectionAcceptor extends ExecutionStrategyInfluencer<ConnectExecutionStrategy>, AsyncCloseable {
+
+    /**
+     * Accept or reject an incoming connection.
+     *
+     * @param info additional information about the connection to make an acceptance decision.
+     * @return a completed (to accept) or a failed (to reject) {@link Completable}
+     */
+    Completable accept(ReducedConnectionInfo info);
```
idelpivnitskiy, line 35: Let's assume that we change this type to `ConnectionInfo`. Then `EarlyConnectionAcceptor` will be identical to `LateConnectionAcceptor`. We will be able to reuse the same interface keeping different configuration methods on the server builder. But then I'm not sure what will be a good name for a new interface. Doesn't look like we can reuse the existing `ConnectionAcceptor`. WDYT?

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
```diff
@@ -87,7 +97,9 @@ public static <CC extends ConnectionContext> Single<ServerContext> bind(SocketAd
             final ReadOnlyTcpServerConfig config, final boolean autoRead, final ExecutionContext<?> executionContext,
             @Nullable final InfluencerConnectionAcceptor connectionAcceptor,
             final BiFunction<Channel, ConnectionObserver, Single<CC>> connectionFunction,
-            final Consumer<CC> connectionConsumer) {
+            final Consumer<CC> connectionConsumer,
+            @Nullable final EarlyConnectionAcceptor earlyConnectionAcceptor,
+            @Nullable final LateConnectionAcceptor lateConnectionAcceptor) {
```
idelpivnitskiy, line 102: Even though this is an internal module, for 0.42.x we try to be extra cautious. Because it's a public method, let's deprecate the existing overload and create a new one.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -183,6 +186,52 @@ public interface HttpServerBuilder {
      */
     HttpServerBuilder appendConnectionAcceptorFilter(ConnectionAcceptorFactory factory);
 
+    /**
+     * Appends the {@link EarlyConnectionAcceptor} to be called when a new connection has been created.
```
idelpivnitskiy, line 190: Would be useful to clarify what early/later means in practice.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -183,6 +186,52 @@ public interface HttpServerBuilder {
...
+     *     acceptor1 ⇒ acceptor2 ⇒ acceptor3
+     * </pre>
+     * <p>
+     * The acceptor is offloaded by default. If an acceptor in the chain fails the {@link Completable}, the later ones
+     * will not be called.
+     *
+     * @param acceptor the acceptor to append to the chain of acceptors.
+     * @return this {@link HttpServerBuilder} for chaining purposes.
+     */
+    HttpServerBuilder appendEarlyConnectionAcceptor(EarlyConnectionAcceptor acceptor);
```
idelpivnitskiy, line 210: These methods should have `default` impl for backward compatibility

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -384,6 +410,54 @@ private <T extends HttpExecutionStrategyInfluencer> HttpExecutionStrategy comput
...
+     * @return the combined acceptor with merged execution strategies.
+     */
+    @Nullable
+    private EarlyConnectionAcceptor buildEarlyConnectionAcceptor(final List<EarlyConnectionAcceptor> acceptors) {
+        return acceptors
+                .stream()
+                .reduce((prev, acceptor) -> new EarlyConnectionAcceptor() {
+                    @Override
+                    public Completable accept(final ReducedConnectionInfo info) {
+                        return prev.accept(info).concat(acceptor.accept(info));
```
idelpivnitskiy, line 426: I see that `ConnectionAcceptorAppender` wraps the 2nd one with `defer` for the same purpose as we discussed in `TcpServerBinder` - isolating the `AsyncContext` for each acceptor. Let's preserve this behavior for now. And please, leave a comment to explain why the `defer` is necessary.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -384,6 +410,54 @@ private <T extends HttpExecutionStrategyInfluencer> HttpExecutionStrategy comput
...
     }
 
+    /**
+     * Combines all early acceptors into one by concatenating the callbacks and merging their execution strategies.
+     *
+     * @param acceptors the acceptors to combine into one.
+     * @return the combined acceptor with merged execution strategies.
+     */
+    @Nullable
+    private EarlyConnectionAcceptor buildEarlyConnectionAcceptor(final List<EarlyConnectionAcceptor> acceptors) {
```
idelpivnitskiy, line 420: IIUC, we can make these methods `static`

idelpivnitskiy (changes requested):

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/EarlyConnectionAcceptor.java
daschl: Agreed that a common interface would be ok, but can't figure out a good name for it since we cannot reuse the current `ConnectionAcceptor` yeah. On the other hand, having 2 interfaces might not be too bad since it would allow to evolve them differently in the future if needed and since most people will use it with a lambda anyways it will not be as visible on the user-facing APIs.

So In my opinion if we do not come up with a good name for a common interface, I would be okay with keeping them separate.

daschl:
Ok, so new round of feedback applied. @idelpivnitskiy aside from another round of reviews and polish potentially, I think we have two open questions:

- Do we need factories or can we keep  using just the acceptors without them.
- Do we want to merge the Early and Late interfaces into one, and if yes, how do we name it?

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/EarlyConnectionAcceptor.java
idelpivnitskiy: Good points!
Also, with a single interface it will be possible to reuse the same impl in early and later stage, which is likely more error prone. For example, users may end up not having `SSLSession` and rejecting requests unintentionally if the place "later impl" in early stage, or they may consume more resources by placing "early impl" at later stage. So, too easy to misuse. Better to force people to think about the position and behavior when they implement interface instead of when they find a good spot for existing impl.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptorOffloadingTest.java
```diff
@@ -15,60 +15,187 @@
...
+
+        assertThat("LateConnectionAcceptor was not invoked", offloaded.get(), is(notNullValue()));
+        assertThat("Incorrect offloading for LateConnectionAcceptor", offloaded.get(), is(offload));
+    }
+
+    /**
+     * Tests the offload merging and makes sure that if at least one is offloaded, both are.
+     */
+    @Test
+    void testMultipleEarlyAcceptorOffloading() throws Exception {
```
idelpivnitskiy, line 119: Consider enhancing "multiple" tests by:
1. Adding the 3rd acceptor and verifying the order of acceptors execution.
2. Placing "offloaded" at 1, 2, 3rd positions to make sure that any offloaded acceptor enabled offloading for all of them.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptorOffloadingTest.java
```diff
@@ -15,60 +15,187 @@
...
 
 import static io.servicetalk.concurrent.api.Single.succeeded;
 import static io.servicetalk.http.api.HttpSerializers.textSerializerUtf8;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.core.Is.is;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 class ConnectionAcceptorOffloadingTest {
```
idelpivnitskiy, line 48: Consider additional tests to make sure all acceptors work with HTTP/1.1 and HTTP/2 and with/without SslConfig. It will help to verify that `EarlyConnectionAcceptorHandler` works as expected.
If you'd like, we can aggregate all tests under `ConnectionAcceptorTest` class in http-netty instead of splitting into `ConnectionAcceptorOffloadingTest` and `ConnectionAcceptorRejectionTest`.

idelpivnitskiy (changes requested):

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -207,11 +214,20 @@ public interface HttpServerBuilder {
...
+    default HttpServerBuilder appendEarlyConnectionAcceptor(EarlyConnectionAcceptor acceptor) {
+        throw new UnsupportedOperationException("appendEarlyConnectionAcceptor is not supported " +
+                "by this HttpServerBuilder");
+    }
 
     /**
      * Appends the {@link LateConnectionAcceptor} to be called when a new connection has been created.
      * <p>
+     * The {@link LateConnectionAcceptor} (compared to the {@link EarlyConnectionAcceptor}) gets called later in the
+     * connection establishment process. Instead of being invoked right after the connection has been created, this
```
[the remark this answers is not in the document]
daschl: We can use `accepted` which I already used in the early one above?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -183,6 +187,70 @@ public interface HttpServerBuilder {
...
+    }
+
+    /**
+     * Appends the {@link LateConnectionAcceptor} to be called when a new connection has been created.
+     * <p>
+     * The {@link LateConnectionAcceptor} (compared to the {@link EarlyConnectionAcceptor}) gets called later in the
+     * connection establishment process. Instead of being invoked right after the connection has been created, this
+     * acceptor gets called after the TLS handshake completes and as a result has more contextual information available
+     * in the {@link ConnectionInfo}. When used on an unsecured connection, the late acceptor will be called immediately
+     * after all early acceptors completed.
```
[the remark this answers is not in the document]
daschl: I'll remove this sentence like above, causes more confusion than it does good.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
```diff
@@ -169,6 +205,116 @@ protected void handleSubscribe(Subscriber<? super ServerContext> subscriber) {
...
+            Completable earlyCompletable = Completable.defer(() -> earlyConnectionAcceptor.accept(info));
+
+            if (earlyConnectionAcceptor.requiredOffloads().isConnectOffloaded()) {
+                earlyCompletable = earlyCompletable.subscribeOn(offloadExecutor);
+            }
+
+            connection = earlyCompletable
+                    .publishOn(channelExecutionContext.ioExecutor(), () -> !channel.eventLoop().inEventLoop())
+                    .concat(connection)
+                    .whenFinally(acceptorHandler::releaseEvents);
```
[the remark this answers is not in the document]
daschl: So in the failure case the handler would be released since `handlerRemoved` is triggered?

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/EarlyConnectionAcceptorHandler.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.NoSuchElementException;
+import java.util.Queue;
+import javax.annotation.Nullable;
+
+/**
+ * Netty handler which queues pipeline events until {@link #releaseEvents()} is called or the handler is removed.
+ */
+class EarlyConnectionAcceptorHandler extends ChannelInboundHandlerAdapter {
+
+    private final Queue<Runnable> queuedEvents = new ArrayDeque<>();
```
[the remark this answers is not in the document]
daschl: Shrinking it down to 4 seems reasonable?

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
```diff
@@ -169,6 +205,116 @@ protected void handleSubscribe(Subscriber<? super ServerContext> subscriber) {
...
+        }
+
+        if (lateConnectionAcceptor != null) {
+            connection = connection.flatMap(conn -> {
+                // Defer is required to isolate the context between the user accept callback and
+                // the rest of the connection
+                Single<CC> deferred = defer(() -> lateConnectionAcceptor.accept(conn).concat(succeeded(conn)));
```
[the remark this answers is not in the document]
daschl: I've looked into it and I'm not sure if this will work as you described since we need the `ConnectionContext` as an argument on the `accept` - so we need to stay inside the `flatMap` - maybe I am missing something?

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/EarlyConnectionAcceptorHandler.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    public void handlerRemoved(final ChannelHandlerContext ctx) throws Exception {
+        releaseEvents();
+        super.handlerRemoved(ctx);
+    }
+
+    @Override
+    public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
+        enqueueOrRunEvent(() -> ctx.fireChannelRead(msg));
+        releaseEvents();
```
idelpivnitskiy, line 93: Thanks a lot for pointing me to this method offline. I finally realized how this works, why SSL case works as expected and found out that HTTP/2 plaintext fails when run repeatedly because h2 uses `AUTO_READ`. As the result, we may receive a `channelRead` event automatically, while the early acceptor is still running on an offloaded thread.

After thinking about it more, I realized that there is no need for HTTP/2 to set AUTO_READ proactively on the Bootstrap. 
Its [H2ParentConnectionContext](https://github.com/apple/servicetalk/blob/e5f0853adec1c8e8a0f3acba46453904fd349065/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ParentConnectionContext.java#L210-L213) already has a logic to override `AUTO_READ` when h2 pipeline initialized. We have to do it this way for ALPN use-case, and we can stick to this approach for plaintext h2-prior-knowledge use-case as well without any impact. I will elaborate more when we meet.

After adding the suggested tests you will see HTTP/2 plaintext failing. This patch will solve the issue:

```patch
Subject: [PATCH] Patch
---
Index: servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DeferredServerChannelBinder.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DeferredServerChannelBinder.java b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DeferredServerChannelBinder.java
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DeferredServerChannelBinder.java	(revision 710e6c7b6736b179eb607f2604b356fc3984c648)
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DeferredServerChannelBinder.java	(date 1674109872536)
@@ -68,9 +68,7 @@
                 (channel, connectionObserver) -> alpnInitChannel(listenAddress, channel, config, executionContext,
                         service, drainRequestPayloadBody, connectionObserver);
 
-        // We disable auto read by default so we can handle stuff in the ConnectionFilter before we accept any content.
-        // In case ALPN negotiates h2, h2 connection MUST enable auto read for its Channel.
-        return TcpServerBinder.bind(listenAddress, tcpConfig, false, executionContext, connectionAcceptor, channelInit,
+        return TcpServerBinder.bind(listenAddress, tcpConfig, executionContext, connectionAcceptor, channelInit,
                 serverConnection -> {
                     // Start processing requests on http/1.1 connection:
                     if (serverConnection instanceof NettyHttpServerConnection) {
Index: servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpRequestEncoderTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpRequestEncoderTest.java b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpRequestEncoderTest.java
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpRequestEncoderTest.java	(revision 710e6c7b6736b179eb607f2604b356fc3984c648)
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpRequestEncoderTest.java	(date 1674109872572)
@@ -436,7 +436,7 @@
 
             ReadOnlyTcpServerConfig sConfig = new TcpServerConfig().asReadOnly();
             ServerContext serverContext = resources.prepend(
-                    TcpServerBinder.bind(localAddress(0), sConfig, false,
+                    TcpServerBinder.bind(localAddress(0), sConfig,
                             SEC, null,
                             (channel, observer) -> DefaultNettyConnection.initChannel(channel, SEC.bufferAllocator(),
                                     SEC.executor(), SEC.ioExecutor(),
Index: servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java	(revision 710e6c7b6736b179eb607f2604b356fc3984c648)
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java	(date 1674109872563)
@@ -30,7 +30,6 @@
 import io.servicetalk.tcp.netty.internal.TcpServerBinder;
 import io.servicetalk.tcp.netty.internal.TcpServerChannelInitializer;
 import io.servicetalk.tcp.netty.internal.TcpServerConfig;
-import io.servicetalk.transport.api.ConnectionObserver;
 import io.servicetalk.transport.api.ServerContext;
 import io.servicetalk.transport.netty.internal.ExecutionContextExtension;
 
@@ -117,15 +116,14 @@
         final ReadOnlyTcpServerConfig tcpReadOnly = new TcpServerConfig().asReadOnly();
 
         try {
-            serverContext = TcpServerBinder.bind(localAddress(0), tcpReadOnly, true,
+            serverContext = TcpServerBinder.bind(localAddress(0), tcpReadOnly,
                     httpExecutionContext, null,
                     (channel, observer) -> {
-                        final ConnectionObserver connectionObserver = config.tcpConfig().transportObserver()
-                                .onNewConnection(channel.localAddress(), channel.remoteAddress());
+                        channel.config().setAutoRead(true);
                         return initChannel(channel, httpExecutionContext, config,
-                                new TcpServerChannelInitializer(tcpReadOnly, connectionObserver)
+                                new TcpServerChannelInitializer(tcpReadOnly, observer)
                                         .andThen(channel1 -> channel1.pipeline().addLast(interceptor)), service,
-                                true, connectionObserver);
+                                true, observer);
                     },
                     connection -> connection.process(true), null, null)
                     .map(delegate -> new NettyHttpServerContext(delegate, service, httpExecutionContext))
Index: servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java	(revision 710e6c7b6736b179eb607f2604b356fc3984c648)
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java	(date 1674109872543)
@@ -96,8 +96,7 @@
             return failed(newH2ConfigException());
         }
         final ReadOnlyTcpServerConfig tcpServerConfig = config.tcpConfig();
-        // Auto read is required for h2
-        return TcpServerBinder.bind(listenAddress, tcpServerConfig, true, executionContext, connectionAcceptor,
+        return TcpServerBinder.bind(listenAddress, tcpServerConfig, executionContext, connectionAcceptor,
                 (channel, connectionObserver) -> initChannel(listenAddress, channel, executionContext, config,
                         new TcpServerChannelInitializer(tcpServerConfig, connectionObserver), service,
                         drainRequestPayloadBody, connectionObserver),
Index: servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java	(revision 710e6c7b6736b179eb607f2604b356fc3984c648)
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java	(date 1674109872553)
@@ -127,8 +127,7 @@
         }
         // This state is read only, so safe to keep a copy across Subscribers
         final ReadOnlyTcpServerConfig tcpServerConfig = config.tcpConfig();
-        // We disable auto read so that we can handle stuff in the ConnectionFilter before we accept any content.
-        return TcpServerBinder.bind(address, tcpServerConfig, false, executionContext, connectionAcceptor,
+        return TcpServerBinder.bind(address, tcpServerConfig, executionContext, connectionAcceptor,
                 (channel, connectionObserver) -> initChannel(channel, executionContext, config,
                         new TcpServerChannelInitializer(tcpServerConfig, connectionObserver), service,
                         drainRequestPayloadBody, connectionObserver),
Index: servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerChannelInitializer.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerChannelInitializer.java b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerChannelInitializer.java
--- a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerChannelInitializer.java	(revision 710e6c7b6736b179eb607f2604b356fc3984c648)
+++ b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerChannelInitializer.java	(date 1674109872597)
@@ -32,7 +32,7 @@
 /**
  * {@link ChannelInitializer} for TCP.
  */
-public class TcpServerChannelInitializer implements ChannelInitializer {
+public class TcpServerChannelInitializer implements ChannelInitializer {    // FIXME: 0.43 - make this class final
 
     private final ChannelInitializer delegate;
 
Index: servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
--- a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java	(revision 710e6c7b6736b179eb607f2604b356fc3984c648)
+++ b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java	(date 1674109893157)
@@ -101,7 +101,7 @@
             @Nullable final InfluencerConnectionAcceptor connectionAcceptor,
             final BiFunction<Channel, ConnectionObserver, Single<CC>> connectionFunction,
             final Consumer<CC> connectionConsumer) {
-        return bind(listenAddress, config, autoRead, executionContext, connectionAcceptor, connectionFunction,
+        return bind(listenAddress, config, executionContext, connectionAcceptor, connectionFunction,
                 connectionConsumer, null, null);
     }
 
@@ -111,7 +111,6 @@
      *
      * @param listenAddress The address to bind to.
      * @param config The {@link ReadOnlyTcpServerConfig} to use for the bind socket and newly accepted sockets.
-     * @param autoRead if {@code true} auto read will be enabled for new {@link Channel}s.
      * @param executionContext The {@link ExecutionContext} to use for the bind socket.
      * @param connectionAcceptor The {@link ConnectionAcceptor} used to filter newly accepted sockets.
      * @param connectionFunction Used to create a new {@link NettyConnection} from a {@link Channel}.
@@ -124,7 +123,7 @@
      * listening on the {@code listenAddress}.
      */
     public static <CC extends ConnectionContext> Single<ServerContext> bind(SocketAddress listenAddress,
-            final ReadOnlyTcpServerConfig config, final boolean autoRead, final ExecutionContext<?> executionContext,
+            final ReadOnlyTcpServerConfig config, final ExecutionContext<?> executionContext,
             @Nullable final InfluencerConnectionAcceptor connectionAcceptor,
             final BiFunction<Channel, ConnectionObserver, Single<CC>> connectionFunction,
             final Consumer<CC> connectionConsumer,
@@ -135,7 +134,7 @@
         listenAddress = toNettyAddress(listenAddress);
         EventLoopAwareNettyIoExecutor nettyIoExecutor = toEventLoopAwareNettyIoExecutor(executionContext.ioExecutor());
         ServerBootstrap bs = new ServerBootstrap();
-        configure(config, autoRead, bs, nettyIoExecutor.eventLoopGroup(), listenAddress.getClass());
+        configure(config, bs, nettyIoExecutor.eventLoopGroup(), listenAddress.getClass());
 
         ChannelSet channelSet = new ChannelSet(
                 executionContext.executionStrategy().isCloseOffloaded() ? executionContext.executor() : immediate());
@@ -315,7 +314,7 @@
         return connection;
     }
 
-    private static void configure(ReadOnlyTcpServerConfig config, boolean autoRead, ServerBootstrap bs,
+    private static void configure(ReadOnlyTcpServerConfig config, ServerBootstrap bs,
                                   @Nullable EventLoopGroup eventLoopGroup,
                                   Class<? extends SocketAddress> bindAddressClass) {
         if (eventLoopGroup == null) {
@@ -335,7 +334,10 @@
             bs.option(option, opt.getValue());
         }
 
-        bs.childOption(ChannelOption.AUTO_READ, autoRead);
+        // We disable AUTO_READ by default to prevent reading any data before the decision to accept the connection is
+        // made and to allow managing back-pressure after that. Protocols that require AUTO_READ have to enable it at
+        // their handler.
+        bs.childOption(ChannelOption.AUTO_READ, false);
 
         // Set the correct ByteBufAllocator based on our BufferAllocator to minimize memory copies.
         ByteBufAllocator byteBufAllocator = POOLED_ALLOCATOR;
Index: servicetalk-tcp-netty-internal/src/testFixtures/java/io/servicetalk/tcp/netty/internal/TcpServer.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-tcp-netty-internal/src/testFixtures/java/io/servicetalk/tcp/netty/internal/TcpServer.java b/servicetalk-tcp-netty-internal/src/testFixtures/java/io/servicetalk/tcp/netty/internal/TcpServer.java
--- a/servicetalk-tcp-netty-internal/src/testFixtures/java/io/servicetalk/tcp/netty/internal/TcpServer.java	(revision 710e6c7b6736b179eb607f2604b356fc3984c648)
+++ b/servicetalk-tcp-netty-internal/src/testFixtures/java/io/servicetalk/tcp/netty/internal/TcpServer.java	(date 1674109872603)
@@ -82,7 +82,7 @@
                               Function<NettyConnection<Buffer, Buffer>, Completable> service,
                               ExecutionStrategy executionStrategy)
-        return TcpServerBinder.bind(localAddress(port), config, false, executionContext, connectionAcceptor,
+        return TcpServerBinder.bind(localAddress(port), config, executionContext, connectionAcceptor,
                 (channel, connectionObserver) -> DefaultNettyConnection.<Buffer, Buffer>initChannel(channel,
                         executionContext.bufferAllocator(), executionContext.executor(), executionContext.ioExecutor(),
                         UNSUPPORTED_PROTOCOL_CLOSE_HANDLER, config.flushStrategy(), config.idleTimeoutMs(),
Index: servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java
--- a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java	(revision 710e6c7b6736b179eb607f2604b356fc3984c648)
+++ b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java	(date 1674109872578)
@@ -33,7 +33,7 @@
 /**
  * {@link ChannelInitializer} for TCP client.
  */
-public class TcpClientChannelInitializer implements ChannelInitializer {
+public class TcpClientChannelInitializer implements ChannelInitializer {    // FIXME: 0.43 - make this class final
 
     private final ChannelInitializer delegate;
 

```

idelpivnitskiy (commented):
This is the last round of review. Everything else looks great!
Thanks a lot for taking care of all the issues around the old acceptor. The new API is much simpler without the factory and works well.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/EarlyConnectionAcceptorHandler.java
idelpivnitskiy: ^ I added 2 comments for 0.43. It's not expected that these classes are not final. Unlikely we will break anyone (they are internal), just trying to play extra safe game for 0.42.x

idelpivnitskiy (approved):
```patch
Index: servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/EarlyAndLateConnectionAcceptorTest.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/EarlyAndLateConnectionAcceptorTest.java b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/EarlyAndLateConnectionAcceptorTest.java
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/EarlyAndLateConnectionAcceptorTest.java	(revision 53ba6e2f7cf3deab7205890235707c0e4dffdbf9)
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/EarlyAndLateConnectionAcceptorTest.java	(date 1674208226674)
@@ -29,42 +29,46 @@
 import io.servicetalk.transport.api.ConnectionAcceptorFactory;
 import io.servicetalk.transport.api.ConnectionInfo;
 import io.servicetalk.transport.api.EarlyConnectionAcceptor;
+import io.servicetalk.transport.api.HostAndPort;
 import io.servicetalk.transport.api.IoThreadFactory;
 import io.servicetalk.transport.api.LateConnectionAcceptor;
 import io.servicetalk.transport.api.ServerContext;
 import io.servicetalk.transport.api.ServerSslConfig;
 import io.servicetalk.transport.api.ServerSslConfigBuilder;
 
-import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.Arguments;
-import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.jupiter.params.provider.CsvSource;
+import org.junit.jupiter.params.provider.EnumSource;
 import org.junit.jupiter.params.provider.ValueSource;
 
 import java.io.IOException;
-import java.net.SocketAddress;
+import java.net.InetSocketAddress;
 import java.util.Queue;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.stream.Stream;
 
 import static io.servicetalk.concurrent.api.Single.succeeded;
 import static io.servicetalk.http.api.HttpSerializers.textSerializerUtf8;
 import static io.servicetalk.http.netty.HttpProtocolConfigs.h1Default;
 import static io.servicetalk.http.netty.HttpProtocolConfigs.h2Default;
+import static io.servicetalk.test.resources.DefaultTestCerts.serverPemHostname;
+import static java.time.Duration.ofMillis;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.core.Is.is;
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assumptions.assumeFalse;
 
 class EarlyAndLateConnectionAcceptorTest {
 
-    @ParameterizedTest
+    @ParameterizedTest(name = "{displayName} [{index}] offload={0}")
     @ValueSource(booleans = {false, true})
     void testSingleAcceptorOffloading(boolean offload) throws Exception {
         AtomicReference<Boolean> offloaded = new AtomicReference<>();
@@ -76,13 +80,13 @@
                             return original.accept(context);
                         },
                         offload ? ConnectExecutionStrategy.offloadAll() : ConnectExecutionStrategy.offloadNone()));
-        doSuccessRequestResponse(builder, ServerType.H1);
+        doSuccessRequestResponse(builder, Protocol.H1);
 
         assertThat("ConnectionAcceptor was not invoked", offloaded.get(), is(notNullValue()));
         assertThat("Incorrect offloading for ConnectionAcceptor", offloaded.get(), is(offload));
     }
 
-    @ParameterizedTest
+    @ParameterizedTest(name = "{displayName} [{index}] offload={0}")
     @ValueSource(booleans = {false, true})
     void testSingleEarlyAcceptorOffloading(boolean offload) throws Exception {
         AtomicReference<Boolean> offloaded = new AtomicReference<>();
@@ -100,13 +104,13 @@
                 return offload ? ConnectExecutionStrategy.offloadAll() : ConnectExecutionStrategy.offloadNone();
             }
         });
-        doSuccessRequestResponse(builder, ServerType.H1);
+        doSuccessRequestResponse(builder, Protocol.H1);
 
         assertThat("EarlyConnectionAcceptor was not invoked", offloaded.get(), is(notNullValue()));
         assertThat("Incorrect offloading for EarlyConnectionAcceptor", offloaded.get(), is(offload));
     }
 
-    @ParameterizedTest
+    @ParameterizedTest(name = "{displayName} [{index}] offload={0}")
     @ValueSource(booleans = {false, true})
     void testSingleLateAcceptorOffloading(boolean offload) throws Exception {
         AtomicReference<Boolean> offloaded = new AtomicReference<>();
@@ -124,7 +128,7 @@
                 return offload ? ConnectExecutionStrategy.offloadAll() : ConnectExecutionStrategy.offloadNone();
             }
         });
-        doSuccessRequestResponse(builder, ServerType.H1);
+        doSuccessRequestResponse(builder, Protocol.H1);
 
         assertThat("LateConnectionAcceptor was not invoked", offloaded.get(), is(notNullValue()));
         assertThat("Incorrect offloading for LateConnectionAcceptor", offloaded.get(), is(offload));
@@ -133,32 +137,9 @@
     /**
      * Tests the offload merging and makes sure that if at least one is offloaded, both are.
      */
-    @ParameterizedTest
-    @MethodSource("multipleAcceptorsOffloadingArgs")
-    void testMultipleAcceptorsOffloadingH1(boolean firstOffloaded, boolean secondOffloaded, boolean thirdOffloaded)
-            throws Exception {
-        final AtomicInteger earlyOffloaded = new AtomicInteger();
-        final AtomicInteger lateOffloaded = new AtomicInteger();
-        final Queue<Integer> executionOrder = new ArrayBlockingQueue<>(6);
-
-        HttpServerBuilder builder = serverBuilder()
-                .appendEarlyConnectionAcceptor(earlyAcceptor(firstOffloaded, earlyOffloaded, executionOrder, 1))
-                .appendEarlyConnectionAcceptor(earlyAcceptor(secondOffloaded, earlyOffloaded, executionOrder, 2))
-                .appendEarlyConnectionAcceptor(earlyAcceptor(thirdOffloaded, earlyOffloaded, executionOrder, 3))
-                .appendLateConnectionAcceptor(lateAcceptor(firstOffloaded, lateOffloaded, executionOrder, 4))
-                .appendLateConnectionAcceptor(lateAcceptor(secondOffloaded, lateOffloaded, executionOrder, 5))
-                .appendLateConnectionAcceptor(lateAcceptor(thirdOffloaded, lateOffloaded, executionOrder, 6));
-
-        doSuccessRequestResponse(builder, ServerType.H1);
-
-        assertEquals(3, earlyOffloaded.get());
-        assertEquals(3, lateOffloaded.get());
-        assertArrayEquals(new Integer[] {1, 2, 3, 4, 5, 6}, executionOrder.toArray(new Integer[0]));
-    }
-
-    @ParameterizedTest
-    @MethodSource("multipleAcceptorsOffloadingArgs")
-    void testMultipleAcceptorsOffloadingH1TLS(boolean firstOffloaded, boolean secondOffloaded, boolean thirdOffloaded)
+    @ParameterizedTest(name = "{displayName} [{index}] firstOffloaded={0} secondOffloaded={1} thirdOffloaded={2}")
+    @CsvSource({"true,false,false", "false,true,false", "false,false,true"})
+    void testMultipleAcceptorsOffloading(boolean firstOffloaded, boolean secondOffloaded, boolean thirdOffloaded)
             throws Exception {
         final AtomicInteger earlyOffloaded = new AtomicInteger();
         final AtomicInteger lateOffloaded = new AtomicInteger();
@@ -172,89 +153,30 @@
                 .appendLateConnectionAcceptor(lateAcceptor(secondOffloaded, lateOffloaded, executionOrder, 5))
                 .appendLateConnectionAcceptor(lateAcceptor(thirdOffloaded, lateOffloaded, executionOrder, 6));
 
-        doSuccessRequestResponse(builder, ServerType.H1_TLS);
+        doSuccessRequestResponse(builder, Protocol.H1);
 
         assertEquals(3, earlyOffloaded.get());
         assertEquals(3, lateOffloaded.get());
         assertArrayEquals(new Integer[] {1, 2, 3, 4, 5, 6}, executionOrder.toArray(new Integer[0]));
     }
 
-    @ParameterizedTest
-    @MethodSource("multipleAcceptorsOffloadingArgs")
-    @Disabled("Disabled due to a H2 plain AUTO_READ issue which will be addressed in a later commit")
-    void testMultipleAcceptorsOffloadingH2(boolean firstOffloaded, boolean secondOffloaded, boolean thirdOffloaded)
-            throws Exception {
+    @ParameterizedTest(name = "{displayName} [{index}] protocol={0}")
+    @EnumSource(Protocol.class)
+    void testEarlyAcceptorWithOffloadingAndDifferentProtocols(Protocol protocol) throws Exception {
+        assumeFalse(protocol == Protocol.H2,
+                "Disabled due to a H2 plain AUTO_READ issue which will be addressed in a later commit");
         final AtomicInteger earlyOffloaded = new AtomicInteger();
-        final AtomicInteger lateOffloaded = new AtomicInteger();
-        final Queue<Integer> executionOrder = new ArrayBlockingQueue<>(6);
 
         HttpServerBuilder builder = serverBuilder()
-                .appendEarlyConnectionAcceptor(earlyAcceptor(firstOffloaded, earlyOffloaded, executionOrder, 1))
-                .appendEarlyConnectionAcceptor(earlyAcceptor(secondOffloaded, earlyOffloaded, executionOrder, 2))
-                .appendEarlyConnectionAcceptor(earlyAcceptor(thirdOffloaded, earlyOffloaded, executionOrder, 3))
-                .appendLateConnectionAcceptor(lateAcceptor(firstOffloaded, lateOffloaded, executionOrder, 4))
-                .appendLateConnectionAcceptor(lateAcceptor(secondOffloaded, lateOffloaded, executionOrder, 5))
-                .appendLateConnectionAcceptor(lateAcceptor(thirdOffloaded, lateOffloaded, executionOrder, 6));
-
-        doSuccessRequestResponse(builder, ServerType.H2);
-
-        assertEquals(3, earlyOffloaded.get());
-        assertEquals(3, lateOffloaded.get());
-        assertArrayEquals(new Integer[] {1, 2, 3, 4, 5, 6}, executionOrder.toArray(new Integer[0]));
-    }
-
-    @ParameterizedTest
-    @MethodSource("multipleAcceptorsOffloadingArgs")
-    void testMultipleAcceptorsOffloadingH2TLS(boolean firstOffloaded, boolean secondOffloaded, boolean thirdOffloaded)
-            throws Exception {
-        final AtomicInteger earlyOffloaded = new AtomicInteger();
-        final AtomicInteger lateOffloaded = new AtomicInteger();
-        final Queue<Integer> executionOrder = new ArrayBlockingQueue<>(6);
-
-        HttpServerBuilder builder = serverBuilder()
-                .appendEarlyConnectionAcceptor(earlyAcceptor(firstOffloaded, earlyOffloaded, executionOrder, 1))
-                .appendEarlyConnectionAcceptor(earlyAcceptor(secondOffloaded, earlyOffloaded, executionOrder, 2))
-                .appendEarlyConnectionAcceptor(earlyAcceptor(thirdOffloaded, earlyOffloaded, executionOrder, 3))
-                .appendLateConnectionAcceptor(lateAcceptor(firstOffloaded, lateOffloaded, executionOrder, 4))
-                .appendLateConnectionAcceptor(lateAcceptor(secondOffloaded, lateOffloaded, executionOrder, 5))
-                .appendLateConnectionAcceptor(lateAcceptor(thirdOffloaded, lateOffloaded, executionOrder, 6));
-
-        doSuccessRequestResponse(builder, ServerType.H2_TLS);
-
-        assertEquals(3, earlyOffloaded.get());
-        assertEquals(3, lateOffloaded.get());
-        assertArrayEquals(new Integer[] {1, 2, 3, 4, 5, 6}, executionOrder.toArray(new Integer[0]));
-    }
-
-    @ParameterizedTest
-    @MethodSource("multipleAcceptorsOffloadingArgs")
-    void testMultipleAcceptorsOffloadingH2ALPN(boolean firstOffloaded, boolean secondOffloaded, boolean thirdOffloaded)
-            throws Exception {
-        final AtomicInteger earlyOffloaded = new AtomicInteger();
-        final AtomicInteger lateOffloaded = new AtomicInteger();
-        final Queue<Integer> executionOrder = new ArrayBlockingQueue<>(6);
-
-        HttpServerBuilder builder = serverBuilder()
-                .appendEarlyConnectionAcceptor(earlyAcceptor(firstOffloaded, earlyOffloaded, executionOrder, 1))
-                .appendEarlyConnectionAcceptor(earlyAcceptor(secondOffloaded, earlyOffloaded, executionOrder, 2))
-                .appendEarlyConnectionAcceptor(earlyAcceptor(thirdOffloaded, earlyOffloaded, executionOrder, 3))
-                .appendLateConnectionAcceptor(lateAcceptor(firstOffloaded, lateOffloaded, executionOrder, 4))
-                .appendLateConnectionAcceptor(lateAcceptor(secondOffloaded, lateOffloaded, executionOrder, 5))
-                .appendLateConnectionAcceptor(lateAcceptor(thirdOffloaded, lateOffloaded, executionOrder, 6));
-
-        doSuccessRequestResponse(builder, ServerType.H2_ALPN);
-
-        assertEquals(3, earlyOffloaded.get());
-        assertEquals(3, lateOffloaded.get());
-        assertArrayEquals(new Integer[] {1, 2, 3, 4, 5, 6}, executionOrder.toArray(new Integer[0]));
-    }
+                .appendEarlyConnectionAcceptor(info -> {
+                    if (!IoThreadFactory.IoThread.currentThreadIsIoThread()) {
+                        earlyOffloaded.incrementAndGet();
+                    }
+                    return info.executionContext().executor().timer(ofMillis(50));
+                });
+        doSuccessRequestResponse(builder, protocol);
 
-    private static Stream<Arguments> multipleAcceptorsOffloadingArgs() {
-        return Stream.of(
-                Arguments.of(true, false, false),
-                Arguments.of(false, true, false),
-                Arguments.of(false, false, true)
-        );
+        assertEquals(1, earlyOffloaded.get());
     }
 
     private static EarlyConnectionAcceptor earlyAcceptor(boolean shouldOffload, final AtomicInteger numOffloaded,
@@ -315,7 +237,7 @@
         }
     }
 
-    enum ServerType {
+    private enum Protocol {
         H1,
         H1_TLS,
         H2,
@@ -323,41 +245,40 @@
         H2_ALPN
     }
 
-    private static void doSuccessRequestResponse(final HttpServerBuilder serverBuilder, final ServerType serverType)
+    private static void doSuccessRequestResponse(final HttpServerBuilder serverBuilder, final Protocol protocol)
             throws Exception {
 
         ServerSslConfig serverSslConfig = new ServerSslConfigBuilder(
                 DefaultTestCerts::loadServerPem, DefaultTestCerts::loadServerKey).build();
 
-        if (serverType == ServerType.H1) {
+        if (protocol == Protocol.H1) {
             serverBuilder.protocols(h1Default());
-        } else if (serverType == ServerType.H1_TLS) {
+        } else if (protocol == Protocol.H1_TLS) {
             serverBuilder.protocols(h1Default()).sslConfig(serverSslConfig);
-        } else if (serverType == ServerType.H2) {
+        } else if (protocol == Protocol.H2) {
             serverBuilder.protocols(h2Default());
-        } else if (serverType == ServerType.H2_TLS) {
+        } else if (protocol == Protocol.H2_TLS) {
             serverBuilder.protocols(h2Default()).sslConfig(serverSslConfig);
-        } else if (serverType == ServerType.H2_ALPN) {
+        } else if (protocol == Protocol.H2_ALPN) {
             serverBuilder.protocols(h2Default(), h1Default()).sslConfig(serverSslConfig);
         }
 
         final HttpService service = (ctx, request, responseFactory) ->
                 succeeded(responseFactory.ok().payloadBody("Hello World!", textSerializerUtf8()));
         try (ServerContext server = serverBuilder.listenAndAwait(service)) {
-            SocketAddress serverAddress = server.listenAddress();
-
-            ClientSslConfig clientSslConfig = new ClientSslConfigBuilder(DefaultTestCerts::loadServerCAPem).build();
-            final SingleAddressHttpClientBuilder<SocketAddress, SocketAddress> clientBuilder =
-                    HttpClients.forResolvedAddress(serverAddress);
-            if (serverType == ServerType.H1) {
+            ClientSslConfig clientSslConfig = new ClientSslConfigBuilder(DefaultTestCerts::loadServerCAPem)
+                    .peerHost(serverPemHostname()).build();
+            final SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> clientBuilder =
+                    HttpClients.forSingleAddress(serverHostAndPort(server));
+            if (protocol == Protocol.H1) {
                 clientBuilder.protocols(h1Default());
-            } else if (serverType == ServerType.H1_TLS) {
+            } else if (protocol == Protocol.H1_TLS) {
                 clientBuilder.protocols(h1Default()).sslConfig(clientSslConfig);
-            } else if (serverType == ServerType.H2) {
+            } else if (protocol == Protocol.H2) {
                 clientBuilder.protocols(h2Default());
-            } else if (serverType == ServerType.H2_TLS) {
+            } else if (protocol == Protocol.H2_TLS) {
                 clientBuilder.protocols(h2Default()).sslConfig(clientSslConfig);
-            } else if (serverType == ServerType.H2_ALPN) {
+            } else if (protocol == Protocol.H2_ALPN) {
                 clientBuilder.protocols(h2Default(), h1Default()).sslConfig(clientSslConfig);
             }
 
@@ -378,13 +299,9 @@
 
         final HttpService service = (ctx, request, responseFactory) ->
                 succeeded(responseFactory.ok().payloadBody("Hello World!", textSerializerUtf8()));
-        try (ServerContext server = builder.listenAndAwait(service)) {
-            SocketAddress serverAddress = server.listenAddress();
-
-            try (BlockingHttpClient client = HttpClients.forResolvedAddress(serverAddress).buildBlocking()) {
-
-                Assertions.assertThrows(IOException.class, () -> client.request(client.get("/sayHello")));
-            }
+        try (ServerContext server = builder.listenAndAwait(service);
+             BlockingHttpClient client = HttpClients.forSingleAddress(serverHostAndPort(server)).buildBlocking()) {
+            assertThrows(IOException.class, () -> client.request(client.get("/sayHello")));
         }
     }
 
@@ -398,17 +315,13 @@
 
         final HttpService service = (ctx, request, responseFactory) ->
                 succeeded(responseFactory.ok().payloadBody("Hello World!", textSerializerUtf8()));
-        try (ServerContext server = builder.listenAndAwait(service)) {
-            SocketAddress serverAddress = server.listenAddress();
-
-            try (BlockingHttpClient client = HttpClients.forResolvedAddress(serverAddress).buildBlocking()) {
-
-                Assertions.assertThrows(IOException.class, () -> client.request(client.get("/sayHello")));
-            }
+        try (ServerContext server = builder.listenAndAwait(service);
+             BlockingHttpClient client = HttpClients.forSingleAddress(serverHostAndPort(server)).buildBlocking()) {
+            assertThrows(IOException.class, () -> client.request(client.get("/sayHello")));
         }
     }
 
     private static HttpServerBuilder serverBuilder() {
-        return HttpServers.forPort(0);
+        return HttpServers.forAddress(localAddress(0));
     }
 }
```

The main changes are:
1. Improve test names for CI. If only `@ParameterizedTest` is used, it works fine in IntellijIDEA, but in CI the names is empty.
2. `HttpServers.forPort(0)` -> `HttpServers.forAddress(localAddress(0))`
3. `peerHost(serverPemHostname())` - required for hostname verification algorithm
4. Removed repeated test code.
5. Used `timer` to delay early acceptor. It helps to make the H2 tests reproducible for every run.

PR #2490, opened 2023-01-13
Use unpadded jctools queues by default

idelpivnitskiy:
Modifications:
- `io.servicetalk.internal.queues.useUnpadded` value defaults to "true" to use unpadded queues by default. This property will be removed in future releases, but exists in case there are any regressions detected.


idelpivnitskiy:
Will keep in draft mode for now until. This change will go into 0.42.26

Scottmitch (approved):

PR #2491, opened 2023-01-16
Deprecate old ConnectionAcceptor

daschl:
I think this can be merged once the new early and late acceptors are merged.


daschl:
@idelpivnitskiy (I'll rebase once the parent changeset is merged)

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -184,7 +184,9 @@ public interface HttpServerBuilder {
      * @param factory {@link ConnectionAcceptorFactory} to append. Lifetime of this
      * {@link ConnectionAcceptorFactory} is managed by this builder and the server started thereof.
      * @return {@code this}
+     * @deprecated use {@link #appendLateConnectionAcceptor(LateConnectionAcceptor)} instead.
      */
+    @Deprecated // FIXME 0.43 - remove deprecated method
     HttpServerBuilder appendConnectionAcceptorFilter(ConnectionAcceptorFactory factory);
```
idelpivnitskiy, line 190: We should add `default` implementation to let users stop implementing this method in their implementations. Users wrap builders using `HttpProviders`. Throwing `UnsupportedOperationException` should be fine. I'm open to other ideas too.

idelpivnitskiy (approved):

PR #2492, opened 2023-01-19
Always disable AUTO_READ for TcpServerBinder

daschl:
Motivation:

AUTO_READ should be disabled by default to prevent reading any data before the decision to accept the connection is made and to allow managing back-pressure after that. Protocols that require AUTO_READ have to enable it at their handler (i.e. H2).

Modifications:

This changeset removes the autoRead option from the bind method of the TcpServerBinder and always sets it to false.

The code also re-enables the previously failing test in EarlyAndLateConnectionAcceptorTest.


daschl:
Will be rebased after the the connection acceptor change is merged.

idelpivnitskiy (approved):
Last commit LGTM

PR #2494, opened 2023-01-24
Update release tag regex to allow optional 4th version section

idelpivnitskiy:
Motivation:

To allow us releasing versions like a.b.c and a.b.c.d, we need to improve the regex to account for the optional 4th section.

Modifications:

- Update ci-release.yml

Result:

Release job will be triggered for both a.b.c and a.b.c.d versions.

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -2,7 +2,7 @@ name: Release Publisher
 on:
   push:
     tags:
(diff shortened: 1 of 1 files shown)


idelpivnitskiy:
Tested with regex tool. This PR doesn't have to wait for all CI pipelines to complete. Merging to unblock the release.

PR #2495, opened 2023-01-24
Split release tag regex into 2 separate patterns

idelpivnitskiy:
Motivation:

GitHub Actions pattern matching does not support full regex features: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

Modifications:

- Split one regex into 2 different patterns;
- Do the same for ci-snapshot.yml to ignore these releases;

Result:

Release pipeline should be triggered for a.b.c and a.b.c.d releases.

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -2,7 +2,8 @@ name: Release Publisher
 on:
   push:
     tags:
-      - "^[0-9]+.[0-9]+.[0-9]+(.[0-9]+)?$"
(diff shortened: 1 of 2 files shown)


idelpivnitskiy:
This PR doesn't need to wait for all CI pipelines to complete. Merging to unblock the release.

idelpivnitskiy:
Reminder for the future me: changes to GitHub Actions have to be made (and cherry-picked) before a tag is created.

PR #2496, opened 2023-01-24
Guard against trying to complete a cancelled Future

mgodave:
Problem/Solution/Result

The DNS client can attempt to complete its result promise even after the Promise
itself has been cancelled upstream, ie the subscriber is no longer interested in
the DNS query results; this will result in an unhandled exception in what should
be a normal use case. We should instead replace calls to setSuccess with
trySuccess on the DNS Promise to guard against this situation.


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -409,7 +409,7 @@ protected Future<DnsAnswer<InetAddress>> doDnsQuery() {
                                 dnsAnswerPromise.setFailure(cause2);
```
idelpivnitskiy: We should likely do the same for all `setFailure` too

idelpivnitskiy (approved):
🚀

PR #2497, opened 2023-01-27
Implement TLS Certificate Compression Support

daschl:
Motivation:

RFC 8879 defines a TLS extension called "TLS Certificate Compression" which allows to compress certificates and possibly reduce the amount of data exchanged during a TLS handshake.

Since netty already added support for it when used together with BoringSSL (netty-tcnative), ServiceTalk can benefit from a higher-level integration.

Modifications:

This changeset implements ServiceTalk API for Certificate Compression and hooks it up with the Netty APIs internally.

The main entry point for the user is through the SslConfigBuilder where they can set the supported algorithms (for now only ZLIB) for both client and server use cases. We also support both the compression and decompression depending on the use case.

At the moment it is not possible to define custom algorithms, although the API has been design with this use-case in mind to open it up later if needed.

ZlibOpenSslCertificateCompressionAlgorithm performs the actual compression and decompression and is called from the netty layer when doing so.

NOTE: since certificate compression is a "best effort" optimization, there is no failure if the feature is not supported or not available at runtime. If needed, inspection through a TransportObserver can determine if the number of bytes exchanged during a handshake are decreased or not.

Result:

TLS certificate compression with ZLIB can now be enabled and lead to less network overhead when performing the TLS handshake.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateCompressionTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateCompressionTest.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright © 2023 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+import io.servicetalk.client.api.TransportObserverConnectionFactoryFilter;
+import io.servicetalk.http.api.BlockingHttpClient;
+import io.servicetalk.http.api.HttpServerBuilder;
+import io.servicetalk.http.api.HttpService;
+import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
+import io.servicetalk.test.resources.DefaultTestCerts;
+import io.servicetalk.transport.api.CertificateCompressionAlgorithms;
+import io.servicetalk.transport.api.ClientSslConfigBuilder;
+import io.servicetalk.transport.api.ConnectionInfo;
+import io.servicetalk.transport.api.ConnectionObserver;
+import io.servicetalk.transport.api.HostAndPort;
+import io.servicetalk.transport.api.ServerContext;
+import io.servicetalk.transport.api.ServerSslConfigBuilder;
+import io.servicetalk.transport.api.SslProvider;
+import io.servicetalk.transport.api.TransportObserver;
+import io.servicetalk.transport.netty.internal.NoopTransportObserver;
+
+import org.junit.jupiter.api.Test;
+
+import java.net.InetSocketAddress;
+import java.util.Collections;
+import javax.annotation.Nullable;
+import javax.net.ssl.SSLSession;
+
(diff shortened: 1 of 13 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ZlibOpenSslCertificateCompressionAlgorithm.java
```diff
@@ -0,0 +1,100 @@
...
+            int bufferSizeEstimate = (int) Math.ceil(uncompressedLength * 1.001) + 12;
+
+            byte[] compressionBuffer = new byte[bufferSizeEstimate];
+            while (!deflater.finished()) {
+                int bytesCompressed = deflater.deflate(compressionBuffer);
+                outputStream.write(compressionBuffer, 0, bytesCompressed);
+            }
+            return outputStream.toByteArray();
+        } catch (Exception cause) {
+            throw new CertificateCompressionException("Failed to compress certificate with " + getClass(), cause);
```
idelpivnitskiy, line 59: Consider: "Failed to compress certificate with ZLIB". 
The class name will be visible in stack-trace

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ZlibOpenSslCertificateCompressionAlgorithm.java
```diff
@@ -0,0 +1,100 @@
...
+            // size on decompress is provided as an argument and anything larger would be a violation of the RFC so
+            // it will be rejected with an Exception.
+            byte[] output = new byte[uncompressedLen];
+            int bytesWritten = 0;
+            while (!inflater.finished()) {
+                int decompressedBytes = inflater.inflate(output, bytesWritten, uncompressedLen - bytesWritten);
+                bytesWritten += decompressedBytes;
+                if (bytesWritten > uncompressedLen) {
+                    throw new CertificateCompressionException("Number of bytes written exceeds the " +
+                            "uncompressed certificate length");
```
idelpivnitskiy, line 85: Consider including values of `bytesWritten` and `uncompressedLen` in the exception message

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ZlibOpenSslCertificateCompressionAlgorithm.java
```diff
@@ -0,0 +1,100 @@
...
+                }
+            }
+            return output;
+        } catch (Exception cause) {
+            throw new CertificateCompressionException("Failed to decompress certificate with " + getClass(), cause);
```
idelpivnitskiy, line 90: Consider: "Failed to decompress certificate with ZLIB".
The class name will be visible in stack-trace

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateCompressionTest.java
```diff
@@ -0,0 +1,179 @@
...
+            succeeded(responseFactory.ok().payloadBody("Hello World!", textSerializerUtf8()));
+
+        long readWithoutCompression;
+        long readWithCompression;
+        long writtenWithoutCompression;
+        long writtenWithCompression;
+
+        try (ServerContext server = serverBuilder(false).listenAndAwait(service)) {
+            SslBytesReadTransportObserver observer = new SslBytesReadTransportObserver();
+            try (BlockingHttpClient client = clientBuilder(server, false, observer).buildBlocking()) {
```
idelpivnitskiy, line 66: Even if the client-side compression currently doesn't work in netty, consider making this test parametrized and run it with/without mTLS to cover both cases.

idelpivnitskiy (approved):
Great! LGTM after comments addressed:

normanmaurer (approved):
LGTM

normanmaurer:
BoringSSL doesn't currently support support for mTLS.

idelpivnitskiy (approved):

PR #2500, opened 2023-02-06
Keep only preferred addresses for DNS resolutions

idelpivnitskiy:
Motivation:

Default value for `DnsResolverAddressTypes` is calculated based on available network interfaces and java.net system properties. When resolver uses `IPV4_PREFERRED` or `IPV6_PREFERRED`, we currently propagate all addresses to the LoadBalancer. Other client implementations, like JDK client/resolver, always use preferred address from the whole result set.

Modifications:

- Make `DefaultDnsServiceDiscovererBuilder.dnsResolverAddressTypes` non-null, initialize it with Netty's default value;
- When `IPV4_PREFERRED` or `IPV6_PREFERRED` is used and the final result contains both types of addresses, keep only preferred type and propagate it down to LB;
- Introduce `IPV4_PREFERRED_RETURN_ALL` and `IPV6_PREFERRED_RETURN_ALL` types to let users enforce pre-existing behavior;
- Enhance tests to verify new changes;

Result:

Default behavior will be aligned with other implementations. If the load balancing is desired for all addresses in result set, users can use new types to enforce that. The change is going to lower overall number of connections and reduce the round-robin circle size when DNS returns IPv4 and IPv6 address for each available host.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -333,6 +333,8 @@ final class DefaultDnsClient implements DnsClient {
                                 if (cause != null) {
                                     promise.tryFailure(cause);
                                 } else {
+                                    LOGGER.trace("DnsClient {}, original result for {}: {}",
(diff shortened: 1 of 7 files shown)


Scottmitch (approved):

idelpivnitskiy:
Moved all general test changes to #2503. Will rebase this PR after #2503 is merged.

PR #2501, opened 2023-02-07
GrpcDeadlineServerFilter: use inner class to capture external state

idelpivnitskiy:
Motivation:

`GrpcDeadlineServerFilter` defines a function that fallbacks to `defaultTimeout`. Because we use lambda, the `defaultTimeout` is not visible during heap dump investigation.

Modifications:

- Use a static inner class `GrpcDetermineTimeoutForRequestFunction` to capture the state;

Result:

`defaultTimeout` value can be found in heap dump.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcFilters.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcFilters.java
@@ -59,6 +59,8 @@ public final class GrpcFilters {
      */
     public static StreamingHttpServiceFilterFactory newGrpcDeadlineServerFilterFactory(
             @Nullable Duration defaultTimeout) {
+        // Use an inner class instead of a lambda to capture the external state (defaultTimeout) for visibility during
+        // heap dump analysis.
         return new TimeoutHttpServiceFilter(new GrpcDetermineTimeoutForRequestFunction(defaultTimeout), true);
     }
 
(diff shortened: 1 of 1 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcFilters.java
```diff
@@ -59,19 +59,29 @@ public static StreamingHttpClientFilterFactory newGrpcDeadlineClientFilterFactor
...
-    private static BiFunction<HttpRequestMetaData, TimeSource, Duration> grpcDetermineTimeout(
-            @Nullable Duration defaultTimeout) {
-        return (request, timeSource) -> {
-            /*
-             * Return the timeout duration extracted from the GRPC timeout HTTP header if present or default timeout.
-             *
-             * @param request The HTTP request to be used as source of the GRPC timeout header
-             * @return The non-negative timeout duration which may be null
-             */
+    private static final class GrpcDetermineTimeoutForRequestFunction
```
daschl, line 65: Consider adding some documentation why this standalone class is there so noone by accident optimizes it back into a lambda.

daschl (approved):

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcFilters.java
```diff
@@ -85,6 +95,6 @@ private static BiFunction<HttpRequestMetaData, TimeSource, Duration> grpcDetermi
             }
 
             return timeout;
-        };
+        }
```
Scottmitch, line 98: do you want to implement `toString()` as well?

Scottmitch (approved):

PR #2503, opened 2023-02-08
Simplify `DefaultDnsClientTest`, avoid depending on real time

idelpivnitskiy:
Motivation:

#2500 showed that `DefaultDnsClientTest` can be flaky. Because these tests rely on real time, they intentionally use larger TTL to make sure all updates are set before the next resolution. However, CI can be too slow and even 1 second delay may not be enough. Some tests deadlock. Also, there are too many unnecessary parametrized tests and all tests take more than 3 minutes.

Modifications:

- Create `NettyIoExecutorWithTestTimer` that delegates to `EventLoopAwareNettyIoExecutor` but uses `TestExecutor` as a time source and to schedule follow-up resolutions;
- Manually advance time and clear the cache before each follow-up query;
- Use cursor to select `nextIp6` instead of random selection. If random selects the same IPv6 address, tests fail their expectations;
- Remove unnecessary parametrized tests, no need to test `missingRecordStatus` for every test;
- Make `setup` extensible to avoid the need to re-create client for some tests;

Result:

Tests are predictable: don't rely on real time or how OS will schedule threads. Total time to run all `DefaultDnsClientTest` tests reduced from more than 3 minutes to less than 3 seconds.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
+++ b/servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
@@ -887,7 +887,8 @@ class DefaultDnsClientTest {
         private final TestExecutor timer;
 
         NettyIoExecutorWithTestTimer(EventLoopAwareNettyIoExecutor delegate, TestExecutor timer) {
-            this.delegate = delegate.next();    // Take a single EventLoopIoExecutor
+            // We need to take a single EventLoopIoExecutor to make sure that isCurrentThreadEventLoop() works.
+            this.delegate = delegate.next();
             this.timer = timer;
         }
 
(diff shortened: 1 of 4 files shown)


servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
```diff
@@ -914,11 +874,127 @@ private static void assertHasEvent(Collection<ServiceDiscovererEvent<InetAddress
...
+
+        @Override
+        public long currentTime(final TimeUnit unit) {
+            return timer.currentTime(unit);
+        }
+
+        @Override
+        public Cancellable schedule(final Runnable task, final long delay, final TimeUnit unit)
+                throws RejectedExecutionException {
+            return timer.schedule(() -> delegate.execute(task), delay, unit);
```
daschl, line 982: is that task run on the timer threads or the delegate threads?

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
```diff
@@ -914,11 +874,127 @@ private static void assertHasEvent(Collection<ServiceDiscovererEvent<InetAddress
...
+        }
+
+        @Override
+        public Cancellable schedule(final Runnable task, final Duration delay) throws RejectedExecutionException {
+            return timer.schedule(() -> delegate.execute(task), delay);
```
daschl, line 987: Same question as above (since looking at the publishOn below for the timer)

daschl (approved):
One question at the end of the tests, but otherwise looks good to me - if you want another pair of eyes though who is more familiar with the DNS code to make sure we did not loose any coverage in the refactoring..

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
idelpivnitskiy: If we pass original `IoExecutor`, it will schedule and execute tasks on EventLoop. With this modification, we move scheduling to `TestExecutor`, then have to manually offload execution back to EventLoop, because internal implementation of `DefaultDnsClient` expects it with multiple assertions. I will leave a comment in code.

PR #2507, opened 2023-02-14
Use SslProvider.isOptionSupported to determine certificate compression support

daschl:
Motivation:

Netty recently added a getter to SslProvider which allows us to check if certain options are supported or not.

Modifications:

This changeset changes the SslContextFactory to use this new getter and explicitly check for
OpenSslContextOption.CERTIFICATE_COMPRESSION_ALGORITHMS before enabling certificate compression.

Result:

More reliable check if certificate compression is supported or not.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateCompressionTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateCompressionTest.java
@@ -33,6 +33,8 @@ import io.servicetalk.transport.api.TransportObserver;
 import io.servicetalk.transport.netty.internal.NoopTransportObserver;
 
 import io.netty.handler.ssl.OpenSsl;
+import io.netty.handler.ssl.OpenSslContextOption;
+import org.junit.jupiter.api.condition.EnabledIf;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 
@@ -48,18 +50,23 @@ import static io.servicetalk.transport.netty.internal.AddressUtils.localAddress;
 import static io.servicetalk.transport.netty.internal.AddressUtils.serverHostAndPort;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
-import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 class SslCertificateCompressionTest {
+
+    private static boolean certCompressionAvailable() {
+        return OpenSsl.isAvailable() && io.netty.handler.ssl.SslProvider.isOptionSupported(
+                io.netty.handler.ssl.SslProvider.OPENSSL, OpenSslContextOption.CERTIFICATE_COMPRESSION_ALGORITHMS);
+    }
 
     /**
      * Compares the bytes written and read when certificate compression is enabled vs. when it is not.
      */
     @ParameterizedTest(name = "{displayName} [{index}] {arguments}")
     @ValueSource(booleans = {true, false})
+    @EnabledIf(
+            value = "certCompressionAvailable",
+            disabledReason = "OpenSSL not available or certificate compression is not supported")
     void negotiatesCertificateCompression(boolean clientEnabled) throws Exception {
-        assumeTrue(OpenSsl.isAvailable());
-
         final HttpService service = (ctx, request, responseFactory) ->
             succeeded(responseFactory.ok().payloadBody("Hello World!", textSerializerUtf8()));
 
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateCompressionTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateCompressionTest.java
@@ -32,6 +32,7 @@ import io.servicetalk.transport.api.SslProvider;
 import io.servicetalk.transport.api.TransportObserver;
 import io.servicetalk.transport.netty.internal.NoopTransportObserver;
 
+import io.netty.handler.ssl.OpenSsl;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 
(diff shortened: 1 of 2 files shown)


Scottmitch (approved):

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateCompressionTest.java
```diff
@@ -56,6 +58,8 @@ class SslCertificateCompressionTest {
     @ParameterizedTest(name = "{displayName} [{index}] {arguments}")
     @ValueSource(booleans = {true, false})
     void negotiatesCertificateCompression(boolean clientEnabled) throws Exception {
+        assumeTrue(OpenSsl.isAvailable());
```
normanmaurer, line 61: FYI. You can also use `@EnabledIf`

daschl: Good idea! At the moment (if my grep foo works) it looks like `assumeTrue` is exclusively used in ServiceTalk, I'll check if there is a benefit migrating to the annotation.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SslContextFactory.java
```diff
@@ -180,8 +194,7 @@ private static void configureCertificateCompression(SslConfig config, SslContext
             nettySslProvider = forServer ? SslContext.defaultServerProvider() : SslContext.defaultClientProvider();
         }
 
-        // TODO: change once https://github.com/netty/netty/pull/13145 is merged and released
-        if (nettySslProvider != SslProvider.OPENSSL) {
+        if (!SslProvider.isOptionSupported(nettySslProvider, OpenSslContextOption.CERTIFICATE_COMPRESSION_ALGORITHMS)) {
```
idelpivnitskiy, line 197: It's a frequent case that users have an older version of netty in the classpath. Sometimes due to their build issues, sometimes intentionally. 
When netty releases a new feature, we try to use method handle for that and use it only when a correct netty version is in the classpath. 
Example: https://github.pie.apple.com/servicetalk/servicetalk/blob/40d2f02d6a0723f2d3198bdd59862a4f02f044b8/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptimizedHttp2FrameCodecBuilder.java#L43

idelpivnitskiy (changes requested):

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SslContextFactory.java
idelpivnitskiy: Otherwise, users may be forced to downgrade ST version too.

daschl: Wouldn't this code only be triggered once someone opts-in to the cert compression feature though?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateCompressionTest.java
normanmaurer: I suspect this is just because they did migrate from junit4 and did not update these yet.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SslContextFactory.java
idelpivnitskiy: It will at ST level, which doesn't mean users can not downgrade netty version or have a mess in the classpath.
Another vector is that the feature can be enabled by a library behind a scene, not by an app owner.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateCompressionTest.java
idelpivnitskiy: 1. Should we use `SslProvider.isOptionSupported(OPENSSL, OpenSslContextOption.CERTIFICATE_COMPRESSION_ALGORITHMS)` instead of `OpenSsl.isAvailable()`?
2. Consider adding a message that will clarify why the test is disabled.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SslContextFactory.java
```diff
@@ -168,21 +195,40 @@ private static void configureTrustManager(SslConfig config, SslContextBuilder bu
...
+     * @param builder netty's builder for the SSL context.
+     * @param nettySslProvider the (potentially null) SSL provider used with netty.
+     * @param forServer if this is for a server or client context.
+     */
     private static void configureCertificateCompression(SslConfig config, SslContextBuilder builder,
                                                         @Nullable io.netty.handler.ssl.SslProvider nettySslProvider,
                                                         boolean forServer) {
         final List<CertificateCompressionAlgorithm> algorithms = config.certificateCompressionAlgorithms();
-        if (algorithms == null || algorithms.isEmpty()) {
+        if (algorithms == null || algorithms.isEmpty() || SSL_PROVIDER_OPTION_SUPPORTED == null) {
```
idelpivnitskiy, line 216: The certificate compression feature is available in prior Netty versions too. We should not skip it if the new method is not available.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SslContextFactory.java
```diff
@@ -168,21 +195,40 @@ private static void configureTrustManager(SslConfig config, SslContextBuilder bu
...
 
         if (nettySslProvider == null) {
             nettySslProvider = forServer ? SslContext.defaultServerProvider() : SslContext.defaultClientProvider();
         }
 
-        // TODO: change once https://github.com/netty/netty/pull/13145 is merged and released
-        if (nettySslProvider != SslProvider.OPENSSL) {
-            return;
+        try {
+            boolean compressionSupported = (boolean) SSL_PROVIDER_OPTION_SUPPORTED.invokeExact(
```
idelpivnitskiy, line 225: Consider using this method only when it's available, otherwise fallback to the pre-existing check.

idelpivnitskiy (approved):
Thanks!

PR #2508, opened 2023-02-16
Implement explicit, customizable DNS maxTTL

daschl:
Motivation:

At the moment it is possible to configure a minTTL for DNS resolution, but maxTTL is not configured (the cache uses Integer.MAX_INT) and the actual maxTTL for resolution is applied to whatever the DNS entry returns from the server.

To provide more customizations as well as being able to explicitly "cap" a TTL at a maximum value, it makes sense to allow setting an explicit maximum TTL.

Modifications:

This changeset introduces a configurable maxTTL option on the DefaultDnsServiceDiscovererBuilder with a sensible default setting. This value then:

 - Propagates to netty's DefaultDnsCache as the maximum value

as well as

 - Caps the returned TTL from the server to this value and logs a INFO level message if it has been capped (since reaching the MAX is really not expected in a well configured environment and can point to issues the actual deployed system).

Result:

Configurable DNS maxTTL as well as TTL capping for the DNS records to provide better safety at runtime.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -730,9 +730,9 @@ final class DefaultDnsClient implements DnsClient {
 
                     ttlNanos = dnsAnswer.ttlNanos();
                     if (ttlNanos > maxTTLNanos) {
-                        LOGGER.info("{} DNS Record {} has a high TTL of {}ms which is larger than maxTTL of {}ms, " +
-                                        "capping to maxTTL.", DefaultDnsClient.this, addresses,
-                                NANOSECONDS.toMillis(ttlNanos), NANOSECONDS.toMillis(maxTTLNanos));
+                        LOGGER.info("{} result for {} has a high TTL={}s which is larger than configured maxTTL={}s.",
+                                DefaultDnsClient.this, AbstractDnsPublisher.this,
+                                NANOSECONDS.toSeconds(ttlNanos), NANOSECONDS.toSeconds(maxTTLNanos));
                         ttlNanos = maxTTLNanos;
                     }
 
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -114,7 +114,7 @@ final class DefaultDnsClient implements DnsClient {
     private final EventLoopAwareNettyIoExecutor nettyIoExecutor;
     private final DnsNameResolver resolver;
     private final MinTtlCache ttlCache;
-    private final int maxTTL;
+    private final long maxTTLNanos;
     private final ListenableAsyncCloseable asyncCloseable;
     @Nullable
     private final DnsServiceDiscovererObserver observer;
@@ -141,7 +141,7 @@ final class DefaultDnsClient implements DnsClient {
         if (srvConcurrency <= 0) {
             throw new IllegalArgumentException("srvConcurrency: " + srvConcurrency + " (expected >0)");
         }
-        this.maxTTL = maxTTL;
+        this.maxTTLNanos = SECONDS.toNanos(maxTTL);
         this.srvConcurrency = srvConcurrency;
         this.srvFilterDuplicateEvents = srvFilterDuplicateEvents;
         this.inactiveEventsOnError = inactiveEventsOnError;
@@ -344,7 +344,7 @@ final class DefaultDnsClient implements DnsClient {
                                     promise.tryFailure(cause);
                                 } else {
                                     final DnsAnswer<HostAndPort> dnsAnswer;
-                                    long answerTTL = Long.MAX_VALUE;
+                                    long minTTLSeconds = Long.MAX_VALUE;
                                     List<DnsRecord> toRelease = null;
                                     try {
                                         @SuppressWarnings("unchecked")
@@ -356,14 +356,8 @@ final class DefaultDnsClient implements DnsClient {
                                                 throw new IllegalArgumentException(
                                                         "Unsupported DNS record type for SRV query: " + dnsRecord);
                                             }
-                                            if (dnsRecord.timeToLive() < answerTTL) {
-                                                answerTTL = dnsRecord.timeToLive();
-                                            }
-                                            if (answerTTL > maxTTL) {
-                                                answerTTL = maxTTL;
-                                                LOGGER.info("DNS Record {} has a high TTL of {}s which is " +
-                                                        "larger than maxTTL of {}, capping to maxTTL.",
-                                                        dnsRecord, dnsRecord.timeToLive(), maxTTL);
+                                            if (dnsRecord.timeToLive() < minTTLSeconds) {
+                                                minTTLSeconds = dnsRecord.timeToLive();
                                             }
                                             ByteBuf content = ((DnsRawRecord) dnsRecord).content();
                                             // https://tools.ietf.org/html/rfc2782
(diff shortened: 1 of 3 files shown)


daschl:
Marking this as a draft since I need to figure out how to test the capping properly in `DefaultDnsClientTest`, but wanted to vet the overall approach first before polish.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -348,8 +351,14 @@ protected Future<DnsAnswer<HostAndPort>> doDnsQuery() {
...
                                                         "Unsupported DNS record type for SRV query: " + dnsRecord);
                                             }
-                                            if (dnsRecord.timeToLive() < minTTLSeconds) {
-                                                minTTLSeconds = dnsRecord.timeToLive();
+                                            if (dnsRecord.timeToLive() < answerTTL) {
+                                                answerTTL = dnsRecord.timeToLive();
+                                            }
+                                            if (answerTTL > maxTTL) {
+                                                answerTTL = maxTTL;
+                                                LOGGER.info("DNS Record {} has a high TTL of {}s which is " +
```
idelpivnitskiy, line 359: 1. There is a chance that DNS can return different TTL values for different records, but in practice they are usually the same. This for loop tries to identity the lowest ttl value for all records. Consider moving this out of the loop to log only once after we identified the effective answerTTL/minTTLSeconds, no need to bother with every record.
2. Having it here only covers SRV resolutions, for A/AAAA resolutions there is a different subscription impl. To cover all of them, consider moving this logging to `handleResolveDone0` where we schedule the next query.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -53,6 +62,7 @@ public final class DefaultDnsServiceDiscovererBuilder {
     @Nullable
     private Duration queryTimeout;
     private int minTTLSeconds = 10;
```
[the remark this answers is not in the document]
idelpivnitskiy: Discussed with Scott offline: 5 seconds will be too aggressive. Let's keep 10 by default for now.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -39,6 +40,14 @@
  * resolve {@code A}, {@code AAAA}, {@code CNAME}, and  {@code SRV} type queries.
  */
 public final class DefaultDnsServiceDiscovererBuilder {
+
+    // Specifies the maximum TTL time in seconds that a DNS entry is allowed to have. If the server
+    // returns a larger one, it will be capped at this value.
+    //
+    // Note: this value is set to 1 hour by default because CoreDNS also uses 3600 seconds as its
+    // maximum value.
+    private static final int DEFAULT_MAX_TTL_SECS = (int) TimeUnit.HOURS.toSeconds(1);
```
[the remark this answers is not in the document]
daschl: I guess one hour is conservative and users can lower it themselves if needed -- I'm happy to adjust lower if you think that would work better as a default.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -80,6 +90,20 @@ public DefaultDnsServiceDiscovererBuilder minTTL(final int minTTLSeconds) {
         return this;
     }
 
+    /**
+     * The maximum allowed TTL. This will be the maximum poll interval as well as the maximum dns cache value.
+     *
+     * @param maxTTLSeconds the maximum amount of time a cache entry will be considered valid (in seconds).
+     * @return {@code this}.
+     */
+    public DefaultDnsServiceDiscovererBuilder maxTTL(final int maxTTLSeconds) {
```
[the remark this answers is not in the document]
daschl: I do not have a strong preference either - the advantage of keeping it separate is that a user who only wants to tweak one is not forced to think about the default value of the other (or we provide public access to the defaults we set). My assumption would be that since you usually construct the builder and then build immediately there won't be really a big difference when the exception is thrown.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -724,7 +727,15 @@ private void handleResolveDone0(final Future<DnsAnswer<T>> addressFuture,
                             comparator(), resolutionObserver == null ? null : (nAvailable, nMissing) ->
                                     reportResolutionResult(resolutionObserver, dnsAnswer, nAvailable, nMissing),
                             missingRecordStatus);
+
                     ttlNanos = dnsAnswer.ttlNanos();
+                    if (ttlNanos > maxTTLNanos) {
+                        LOGGER.info("{} DNS Record {} has a high TTL of {}ms which is larger than maxTTL of {}ms, " +
+                                        "capping to maxTTL.", DefaultDnsClient.this, addresses,
+                                NANOSECONDS.toMillis(ttlNanos), NANOSECONDS.toMillis(maxTTLNanos));
```
idelpivnitskiy, line 735: Consider:
```suggestion
                        LOGGER.info("{} result for {} has a high TTL={}s which is larger than configured maxTTL={}s.",
                                DefaultDnsClient.this, AbstractDnsPublisher.this,
                                NANOSECONDS.toSeconds(ttlNanos), NANOSECONDS.toSeconds(maxTTLNanos));
```

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
```diff
@@ -984,6 +984,25 @@ void srvExceptionInSubscriberOnNext() throws Exception {
...
+        TestPublisherSubscriber<ServiceDiscovererEvent<InetAddress>> subscriber = dnsQuery(domain);
+        Subscription subscription = subscriber.awaitSubscription();
+        subscription.request(Long.MAX_VALUE);
+
+        System.err.println(subscriber.takeOnNext());
+        recordStore.removeIPv4Address(domain, 1, ip);
+        advanceTime(3);
+        System.err.println(subscriber.takeOnNext());
+
+        //assertEvent(subscriber.takeOnNext(), ipv6, AVAILABLE);
```
idelpivnitskiy, line 1003: Discussed offline: we have to add a new IP in order for events to propagate.
Please also add a test for SRV records.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -53,6 +54,7 @@ public final class DefaultDnsServiceDiscovererBuilder {
     @Nullable
     private Duration queryTimeout;
     private int minTTLSeconds = 10;
+    private int maxTTLSeconds = (int) TimeUnit.MINUTES.toSeconds(5);
```
idelpivnitskiy, line 57: minutes instead of seconds?

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
```diff
@@ -984,6 +984,56 @@ void srvExceptionInSubscriberOnNext() throws Exception {
...
+        recordStore.addIPv4Address(domain, 5, ip1);
+
+        TestPublisherSubscriber<ServiceDiscovererEvent<InetAddress>> subscriber = dnsQuery(domain);
+        Subscription subscription = subscriber.awaitSubscription();
+        subscription.request(Long.MAX_VALUE);
+
+        assertEvent(subscriber.takeOnNext(), ip1, AVAILABLE);
+        recordStore.removeIPv4Address(domain, 5, ip1);
+        recordStore.addIPv4Address(domain, 5, ip2);
+        advanceTime(3);
```
idelpivnitskiy, line 1002: Consider using a local variable to store `maxTTL` value to make it clear you advance only `maxTTL`.

idelpivnitskiy (commented):
Final comments then LGTM. Thanks!

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
idelpivnitskiy: nvm, I misread this line as `TimeUnit.ofSeconds(5)`

idelpivnitskiy (approved):

PR #2509, opened 2023-02-17
`DefaultDnsClient`: enhance logging

idelpivnitskiy:
Motivation:

Help debugging DNS issues when TTL is unexpectedly high.

Modifications:
- Include TTL value for records every time we can access it for visibility of the internal state;
- Include hashCode value for A*/SRV publishers to distinguish different publishers for the same hostname in logs;
- Adjust wording;
- Use simple name for `DefaultDnsClient` to reduce noise in logs;

Result:

Easier to understand internal state of `DefaultDnsClient` for debugging.


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -192,6 +193,12 @@ final class DefaultDnsClient implements DnsClient {
         resolver = builder.build();
     }
 
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + // FIXME: change to the name/id when builder requires it
```
idelpivnitskiy, line 198: This will be "id" we discussed with you when you will add ServiceLoader provider

daschl (approved):

PR #2514, opened 2023-02-21
`RoundRobinLoadBalancer`: re-subscribe when all hosts become unhealthy

idelpivnitskiy:
Motivation:

It’s possible for the RRLB to turn into a state when all available
hosts become unhealthy. Possible scenarios:
- DNS returned incorrect results;
- All hosts were restarted and got different addresses.
RRL will be in a dead state until TTL expires, which may take some time.

Modifications:
- When RRLB detects that all hosts are unhealthy, it re-subscribes to the
SD events publisher to trigger a new resolution;
- Add `RoundRobinLoadBalancerFactory.Builder#healthCheckResubscribeInterval`
option to configure the new feature;
- Wrap `backgroundExecutor` with `NormalizedTimeSourceExecutor` to make
sure `currentTime` is always positive;
- Test behavior when the new feature is enabled/disabled;
- Add `DurationUtils.ensureNonNegative(...)` utility for validation;
- Make `TestExecutor(long)` constructor public to test
`NormalizedTimeSourceExecutor`;
- Add `SequentialPublisherSubscriberFunction.numberOfSubscribers()` to
verify how many subscribers the `TestPublisher` already saw;

Result:

RRLB forces SD to re-resolve addresses by cancelling the current subscription
and re-subscribing to the publisher when it detects that all hosts become
unhealthy.

---
~This PR depends on #2513 and #2516, review only the last commit.~

TODO:
- [x] Add tests for RRLB;
- [x] #2518;

### DIFF
--- a/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/SequentialPublisherSubscriberFunction.java
+++ b/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/SequentialPublisherSubscriberFunction.java
@@ -19,6 +19,7 @@ import io.servicetalk.concurrent.PublisherSource.Subscriber;
 import io.servicetalk.concurrent.PublisherSource.Subscription;
 
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import javax.annotation.Nullable;
 
@@ -32,6 +33,7 @@ public final class SequentialPublisherSubscriberFunction<T>
         implements Function<Subscriber<? super T>, Subscriber<? super T>> {
 
     private final AtomicBoolean subscribed = new AtomicBoolean();
+    private final AtomicInteger numberOfSubscribers = new AtomicInteger();
     @Nullable
     private volatile Subscriber<? super T> subscriber;
 
@@ -41,6 +43,7 @@ public final class SequentialPublisherSubscriberFunction<T>
             throw new IllegalStateException("Duplicate subscriber: " + subscriber);
         }
         this.subscriber = subscriber;
+        numberOfSubscribers.incrementAndGet();
         return new DelegatingPublisherSubscriber<T>(subscriber) {
             @Override
             public void onSubscribe(final Subscription s) {
@@ -98,5 +101,14 @@ public final class SequentialPublisherSubscriberFunction<T>
      */
     public boolean isSubscribed() {
         return subscribed.get();
+    }
+
+    /**
+     * Returns total number of observed {@link Subscriber}s.
+     *
+     * @return total number of observed {@link Subscriber}s.
+     */
+    public int numberOfSubscribers() {
+        return numberOfSubscribers.get();
     }
 }
--- a/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/SequentialPublisherSubscriberFunction.java
+++ b/servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/SequentialPublisherSubscriberFunction.java
@@ -33,7 +33,7 @@ public final class SequentialPublisherSubscriberFunction<T>
         implements Function<Subscriber<? super T>, Subscriber<? super T>> {
 
     private final AtomicBoolean subscribed = new AtomicBoolean();
-    private final AtomicInteger numberOfSubscribers = new AtomicInteger();
+    private final AtomicInteger numberOfSubscribersSeen = new AtomicInteger();
     @Nullable
     private volatile Subscriber<? super T> subscriber;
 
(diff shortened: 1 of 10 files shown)


servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/SequentialPublisherSubscriberFunction.java
```diff
@@ -41,6 +43,7 @@ public Subscriber<? super T> apply(final Subscriber<? super T> subscriber) {
             throw new IllegalStateException("Duplicate subscriber: " + subscriber);
         }
         this.subscriber = subscriber;
+        numberOfSubscribers.incrementAndGet();
```
daschl, line 46: we only increment, do we ever need to decrement this counter too (i.e. on unsubscribe?)

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -154,188 +165,288 @@
...
+                        }
+                    })
+                    .beforeOnComplete(() -> usedHosts = new ClosedList<>(emptyList()));
+        });
+        subscribeToEvents(false);
+    }
 
-        toSource(eventPublisher).subscribe(
-                new Subscriber<Collection<? extends ServiceDiscovererEvent<ResolvedAddress>>>() {
+    private void subscribeToEvents(boolean resubscribe) {
```
daschl, line 198: is there a possibility that this method is called multiple times concurrently (and if so, do you think it has racy conditions around cancelation?)

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -202,27 +211,59 @@ public RoundRobinLoadBalancerFactory.Builder<ResolvedAddress, C> healthCheckInte
...
-                throw new IllegalArgumentException("Jitter plus/minus interval underflow/overflow");
-            }
+            validate(interval, jitter);
             this.healthCheckInterval = interval;
             this.healthCheckJitter = jitter;
             return this;
         }
 
+        /**
+         * Configure an interval for re-subscribing to the original events stream in case all existing hosts become
```
daschl, line 229: consider mentioning the default interval used if not overridden

servicetalk-concurrent-api/src/testFixtures/java/io/servicetalk/concurrent/api/SequentialPublisherSubscriberFunction.java
idelpivnitskiy: In a sequential mode there is only one subscriber allowed at a time and this can be verified by `isSubscribed()` method. The goal of this feature is to count how many subscribers it saw during the lifetime of the publisher. Maybe it needs a better name. I will try `numberOfSubscribersSeen()`. LMK if you have other ideas.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
idelpivnitskiy: `nextResubscribeTime` is used to make sure there is only one thread at a time that performs re-subscribe (unless I messed up something around lines 540-543). I will add `assert nextResubscribeTime == RESUBSCRIBING` at the beginning of this method.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
idelpivnitskiy: We made a decision not to leak any default values in javadoc everywhere in the project to give us more flexibility to change those in the future. We can reconsider the approach when we are 1.0.

daschl: ah gotcha 👍

servicetalk-utils-internal/src/main/java/io/servicetalk/utils/internal/DurationUtils.java
```diff
@@ -67,6 +67,22 @@ public static Duration ensurePositive(final Duration duration, final String name
...
+    /**
+     * Ensures the duration is non-negative.
+     *
+     * @param duration the {@link Duration} to validate
+     * @param name name of the {@link Duration} variable
+     * @return the passed duration if all checks pass
+     * @throws NullPointerException if the passed duration is {@code null}
+     * @throws IllegalArgumentException if the passed duration is not greater or equal to {@link Duration#ZERO}
+     */
+    public static Duration ensureNonNegative(final Duration duration, final String name) {
```
daschl, line 79: the check above has a requireNonNull - consider one here too.

servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/NormalizedTimeSourceExecutorTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+@SuppressWarnings("NumericOverflow")
+class NormalizedTimeSourceExecutorTest {
+
+    private TestExecutor testExecutor;
+    private Executor executor;
+
+    @AfterEach
+    void tearDown() throws Exception {
+        executor.closeAsync().toFuture().get();
```
daschl, line 39: TestExecutor does not need to be shut down?

PR #2515, opened 2023-02-21
Sequential `Subscriber` is not reset after cancel for all test sources

idelpivnitskiy:
Motivation:

When `Sequential*SubscriberFunction` (this is a default mode for all test sources) is used, `cancel()` doesn't reset the current `Subscriber`. Therefore, following re-subscribes fail with "Duplicate subscriber" exception. This happens because `AutoOnSubscribe*SubscriberFunction` is applied before `subscriberCardinalityFunction`. As the result, `onSubscribe` is invoked before `Sequential*SubscriberFunction` wraps the `Subscriber`.

Modifications:

- Always apply `autoOnSubscribeSubscriberFunction` last, after `subscriberCardinalityFunction`;
- Add tests to verify the described scenario;
- Always reset the current `Subscriber` when `Sequential*SubscriberFunction` resets;
- Clarify that `Concurrent*SubscriberFunction` always requires manyual `onSubscribe` management;
- Disable auto on-subscribe when `concurrentSubscribers` mode is used;

Result:

It's possible to re-subscribe to the test source after the previous subscription is cancelled when `sequentialSubscribers` mode is used.

### DIFF
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/TestCompletableTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/TestCompletableTest.java
@@ -15,6 +15,7 @@
  */
 package io.servicetalk.concurrent.api;
 
+import io.servicetalk.concurrent.Cancellable;
 import io.servicetalk.concurrent.test.internal.TestCompletableSubscriber;
 
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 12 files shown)


servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/TestCompletableTest.java
```diff
@@ -58,30 +61,68 @@ void testNonResubscribeableCompletable() {
...
+        source.onComplete();
+        assertThat(sequentialCompletableSubscriberFunction.subscriber(), is(nullValue()));
+        assertThat(sequentialCompletableSubscriberFunction.isSubscribed(), is(false));
+        subscriber2.awaitOnComplete();
+    }
+
     @Test
     void testConcurrentSubscribeCompletable() {
         TestCompletable source = new TestCompletable.Builder()
                 .concurrentSubscribers()
```
idelpivnitskiy, line 115: Note that `concurrentSubscribers()` mode isn't used anywhere outside these tests for `TestPublisher|Completable|Single`. Disabling auto on-subscribe won't have any impact.
Furthermore, we do not advertise `testFixtures` as stable public API.

idelpivnitskiy:
Discussed this PR with @Scottmitch offline, all good.

PR #2516, opened 2023-02-21
`SequentialCancellable`: add `cancelCurrent()` method

idelpivnitskiy:
Motivation:

In scenarios when `SequentialCancellable` is used to manage global cancellation, there might be cases when we have to re-subscribe. Cancelling the current `Cancellable` will allow us to do this.

Modifications:

- Add `SequentialCancellable#cancelCurrent()` method;
- Test the new feature;

Result:

It's possible to cancel only the current `Cancellable` inside `SequentialCancellable` and then re-subscribe.

### DIFF
--- a/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/SequentialCancellable.java
+++ b/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/SequentialCancellable.java
@@ -64,8 +64,17 @@ public class SequentialCancellable implements Cancellable {
     @Override
     public final void cancel() {
         Cancellable oldVal = currentUpdater.getAndSet(this, CANCELLED);
-            oldVal.cancel();
-        }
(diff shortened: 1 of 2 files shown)


servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/SequentialCancellable.java
```diff
@@ -64,8 +64,17 @@ public final void nextCancellable(Cancellable next) {
     @Override
     public final void cancel() {
         Cancellable oldVal = currentUpdater.getAndSet(this, CANCELLED);
-            oldVal.cancel();
-        }
+        oldVal.cancel();
+    }
+
+    /**
+     * Cancels only the {@link Cancellable} that is currently hold without side effect for any
```
daschl, line 71: consider "..that is currently _held_ without side effect"..

daschl (approved):

PR #2518, opened 2023-02-22
Disable cache by default for DNS ServiceDiscoverer

idelpivnitskiy:
Motivation:

When we poll DNS in the background the cache is not important because we always schedule the next query after cache expires. It only helps with concurrent resolutions of the same address, which in practice should be minimal if users properly reuse clients. The disadvantage of having a cache is that if it gets poisoned with invalid or stale entries there is no way to clear the cache. Cancelling the events publisher and re-subscribing doesn't help because re-subscribe always hits the cache. See #2514.

Modifications:

- Disable cache by default;
- Provide API to opt-in for caching (can be useful if users resolve per new connection instead of polling in the background);
- Deprecate `minTTL`, remove `maxTTL` builder methods, introduce `ttl(min, max)` and `ttl(min, max, minCache, maxCache)` instead;
- Invoke `ttlCache.prepareForResolution(name)` only for scheduled queries, keep it unchanged when cancel/re-subscribe to correctly offset ttl;
- Ignore empty list inside `MinTtlCache.get(...)`;
- Make `DefaultDnsClient` logging more consistent;
- Include `id` in the returned `ServiceDiscoverer.toString()` from the builder;
- Enhance testing;

Result:

Caching is disabled by default. Polling is driven by TTL. In case of re-subscribe, we always send a new query. Users have API to configure min/max polling intervals and caching.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -266,7 +266,7 @@ final class DefaultDnsClient implements DnsClient {
                         assertInEventloop();
                         // If this error is because the SRV entry was detected as inactive, then propagate the error and
                         // don't retry. Otherwise this is a resolution exception (e.g. UnknownHostException), and retry.
-                        return cause == SrvAddressRemovedException.DNS_SRV_ADDR_REMOVED ||
+                        return cause.getClass().equals(SrvAddressRemovedException.class) ||
                                 aRecordMap.remove(srvEvent.address().hostName()) == null ?
                                 Completable.failed(cause) : srvHostNameRepeater.apply(i);
                     }).onErrorComplete(); // retryWhen will propagate onError, but we don't want this.
@@ -276,7 +276,8 @@ final class DefaultDnsClient implements DnsClient {
                 } else {
                     final ARecordPublisher aPublisher = aRecordMap.remove(srvEvent.address().hostName());
                     if (aPublisher != null) {
-                        aPublisher.cancelAndFail0(SrvAddressRemovedException.DNS_SRV_ADDR_REMOVED);
+                        aPublisher.cancelAndFail0(
+                                SrvAddressRemovedException.newInstance(DefaultDnsClient.class, "dnsSrvQuery"));
                     }
                     return empty();
                 }
@@ -1001,10 +1002,12 @@ final class DefaultDnsClient implements DnsClient {
 
     private static final class SrvAddressRemovedException extends RuntimeException {
         private static final long serialVersionUID = -4083873869084533456L;
-        private static final SrvAddressRemovedException DNS_SRV_ADDR_REMOVED =
-                unknownStackTrace(new SrvAddressRemovedException(), DefaultDnsClient.class, "dnsSrvQuery");
 
         private SrvAddressRemovedException() {
+        }
+
+        static SrvAddressRemovedException newInstance(Class<?> clazz, String method) {
+            return unknownStackTrace(new SrvAddressRemovedException(), clazz, method);
         }
     }
 }
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -44,6 +44,7 @@ import io.netty.handler.codec.dns.DnsRecord;
 import io.netty.resolver.dns.DefaultDnsCache;
 import io.netty.resolver.dns.DnsNameResolver;
 import io.netty.resolver.dns.DnsNameResolverBuilder;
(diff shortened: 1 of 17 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererBuilder.java
```diff
@@ -30,26 +30,29 @@
...
+     * than {@code 0}.
+     * @param maxSeconds The maximum about of time the result will be considered valid (in seconds), must be greater
+     * than {@code minSeconds}.
+     * @param cache If {@code true}, DNS responses will be cached locally for the specified time. Any concurrent
+     * resolutions for the same address will hit the cache if it's not expired. Otherwise, all resolutions will generate
+     * a new query for DNS server.
      * @return {@code this}.
      */
-    DnsServiceDiscovererBuilder maxTTL(int maxTTLSeconds);
+    DnsServiceDiscovererBuilder ttl(int minSeconds, int maxSeconds, boolean cache);
```
daschl, line 50: If we start using a new method, consider using `Duration` instead of int seconds

daschl: Also, should we add a different builder for the caching enable/disable?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -98,11 +109,14 @@ public DefaultDnsServiceDiscovererBuilder minTTL(final int minTTLSeconds) {
     }
 
     @Override
-    public DefaultDnsServiceDiscovererBuilder maxTTL(final int maxTTLSeconds) {
-        if (minTTLSeconds <= 0) {
-            throw new IllegalArgumentException("maxTTLSeconds: " + maxTTLSeconds + " (expected > 0)");
+    public DefaultDnsServiceDiscovererBuilder ttl(final int minSeconds, final int maxSeconds, final boolean cache) {
+        if (minSeconds < 0 || maxSeconds < minSeconds) {
```
daschl, line 113: If we use a duration, we should check that the supplied duration is > seconds

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -696,17 +701,21 @@ private void cancelAndTerminate0(Throwable cause) {
             }
 
             private void scheduleQuery0(final long nanos) {
+                scheduleQuery0(nanos, nanos);
+            }
+
+            private void scheduleQuery0(final long nanos, final long originalTtlNanos) {
```
daschl, line 707: consider adding a javadoc or renaming "nanos" to something more descriptive

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererBuilder.java
idelpivnitskiy: As we discussed offline, I went with:
```
.ttl(min, max)
.ttl(min, max, minCache, maxCache)
```
Thanks for brainstorming!

idelpivnitskiy:
Build failures will be resolved after #2523 is merged.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererBuilder.java
```diff
@@ -30,26 +30,54 @@
...
+     * Controls min/max TTL values that will influence polling intervals.
+     * <p>
+     * The created {@link ServiceDiscoverer} polls DNS server based on TTL value of the resolved records. Min/max values
+     * help to make sure polling stays within reasonable boundaries. Too frequent DNS queries may generate too much load
+     * for the DNS server, too rare DNS queries may lead to incorrect state if the remote servers changed IPs before
+     * original TTL expired.
+     * <p>
+     * With this overload, there will be no local caching for resolved records.
      *
-     * @param minTTLSeconds The minimum amount of time a cache entry will be considered valid (in seconds).
```
daschl, line 35: Consider describing the default values on each param so that if a user only wants to tweak one off from the default it's easier to find what to put in place.

daschl (approved):

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererBuilder.java
idelpivnitskiy: Discussed offline, leaving a comment here for others: we decided not to leak default values in javadoc at least until we are 1.0 to have some flexibility to adjust them.

PR #2520, opened 2023-02-23
Option to create a client that runs DNS resolutions for every new connection

idelpivnitskiy:
Motivation:

In scenarios when an application creates many clients, and there is a low rate of creating new connections, continuous background DNS resolutions can add too much load on DNS server.

Modifications:

- Introduce `HttpClients.DiscoveryStrategy` that allows users to change the resolutions strategy, with an option to  run ServiceDiscoverer on every new connection;
- Introduce `HttpClients` factories that take `DiscoveryStrategy` into account;
- Add similar factories for `GrpcClients`;
- Refactor `GlobalDnsServiceDiscoverer` to accommodate new behavior;
- Add `ResolvingConnectionFactoryFilter` that implement a new discovery strategy;
- Disable `retryServiceDiscoveryErrors` when it's known that SD will never fail;
- Don't allow users to override `ServiceDiscoverer` through the builder API when `DiscoveryStrategy.ON_NEW_CONNECTION` is used;
- Skip `RetryingServiceDiscoverer` and `StatusAwareServiceDiscoverer` when `HttpClients.NoRetriesStrategy` is set;
- Test new behavior;

Result:

Users can create a client that runs DNS resolutions every time the client opens a new connection.

### DIFF
--- a/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcClients.java
+++ b/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcClients.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019, 2022 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019, 2022-2023 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,6 +22,7 @@ import io.servicetalk.grpc.api.GrpcClientBuilder;
 import io.servicetalk.grpc.api.GrpcProviders.GrpcClientBuilderProvider;
 import io.servicetalk.http.api.HttpHeaderNames;
 import io.servicetalk.http.netty.HttpClients;
+import io.servicetalk.http.netty.HttpClients.DiscoveryStrategy;
 import io.servicetalk.transport.api.HostAndPort;
 
 import org.slf4j.Logger;
@@ -62,7 +63,7 @@ public final class GrpcClients {
 
     /**
      * Creates a {@link GrpcClientBuilder} for an address with default {@link LoadBalancer} and DNS
-     * {@link ServiceDiscoverer}.
+     * {@link ServiceDiscoverer} using {@link DiscoveryStrategy#BACKGROUND background} discovery strategy.
      * <p>
      * The returned builder can be customized using {@link GrpcClientBuilderProvider}.
      *
@@ -77,7 +78,7 @@ public final class GrpcClients {
 
     /**
      * Creates a {@link GrpcClientBuilder} for an address with default {@link LoadBalancer} and DNS
-     * {@link ServiceDiscoverer}.
+     * {@link ServiceDiscoverer} using {@link DiscoveryStrategy#BACKGROUND background} discovery strategy.
      * <p>
      * The returned builder can be customized using {@link GrpcClientBuilderProvider}.
      *
@@ -92,54 +93,37 @@ public final class GrpcClients {
 
     /**
      * Creates a {@link GrpcClientBuilder} for an address with default {@link LoadBalancer} and DNS
-     * {@link ServiceDiscoverer} that will run resolutions before creating a new connection.
-     * <p>
-     * Important side effects to take into account:
-     * <ol>
-     *     <li>The total latency for opening a new connection will be increased by a latency of DNS resolution.</li>
-     *     <li>If the target host has more than one resolved address, created clients loose ability to load balance for
-     *     each request. Instead, the balancing will happen only for every new created connection.</li>
-     *     <li>Created clients won't be able to move/shift traffic based on changes in DNS records until the remote
-     *     server closes existing connections.</li>
-     * </ol>
+     * {@link ServiceDiscoverer} using the specified {@link DiscoveryStrategy}.
      * <p>
      * The returned builder can be customized using {@link GrpcClientBuilderProvider}.
      *
-     * @param host host to connect to, resolved by default using a DNS {@link ServiceDiscoverer} for every new
-     * connection.
+     * @param host host to connect to, resolved by default using a DNS {@link ServiceDiscoverer}.
      * @param port port to connect to
+     * @param discoveryStrategy {@link DiscoveryStrategy} to use
      * @return new builder for the address
      * @see GrpcClientBuilderProvider
      */
-    public static GrpcClientBuilder<HostAndPort, InetSocketAddress> forAddressResolveOnDemand(
-            final String host, final int port) {
-        return forAddressResolveOnDemand(HostAndPort.of(host, port));
+    public static GrpcClientBuilder<HostAndPort, InetSocketAddress> forAddress(
+            final String host, final int port, final DiscoveryStrategy discoveryStrategy) {
+        return forAddress(HostAndPort.of(host, port), discoveryStrategy);
     }
 
     /**
      * Creates a {@link GrpcClientBuilder} for an address with default {@link LoadBalancer} and DNS
-     * {@link ServiceDiscoverer} that will run resolutions before creating a new connection.
-     * <p>
-     * Important side effects to take into account:
-     * <ol>
-     *     <li>The total latency for opening a new connection will be increased by a latency of DNS resolution.</li>
-     *     <li>If the target host has more than one resolved address, created clients loose ability to load balance for
-     *     each request. Instead, the balancing will happen only for every new created connection.</li>
-     *     <li>Created clients won't be able to move/shift traffic based on changes in DNS records until the remote
-     *     server closes existing connections.</li>
-     * </ol>
+     * {@link ServiceDiscoverer} using the specified {@link DiscoveryStrategy}.
      * <p>
      * The returned builder can be customized using {@link GrpcClientBuilderProvider}.
      *
-     * @param address the {@code UnresolvedAddress} to connect to, resolved using a DNS {@link ServiceDiscoverer} for
-     * every new connection.
+     * @param address the {@code UnresolvedAddress} to connect to, resolved by default using a DNS
+     * {@link ServiceDiscoverer}.
+     * @param discoveryStrategy {@link DiscoveryStrategy} to use
      * @return new builder for the address
      * @see GrpcClientBuilderProvider
      */
-    public static GrpcClientBuilder<HostAndPort, InetSocketAddress> forAddressResolveOnDemand(
-            final HostAndPort address) {
+    public static GrpcClientBuilder<HostAndPort, InetSocketAddress> forAddress(
+            final HostAndPort address, final DiscoveryStrategy discoveryStrategy) {
         return applyProviders(address,
-                new DefaultGrpcClientBuilder<>(() -> HttpClients.forSingleAddressResolveOnDemand(address)));
+                new DefaultGrpcClientBuilder<>(() -> HttpClients.forSingleAddress(address, discoveryStrategy)));
     }
 
     /**
--- a/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcClients.java
+++ b/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcClients.java
@@ -205,7 +205,7 @@ public final class GrpcClients {
 
     /**
      * Creates a {@link GrpcClientBuilder} for a custom address type with default {@link LoadBalancer} and user
-     * provided {@link ServiceDiscoverer}.
+     * provided {@link ServiceDiscoverer} using {@link DiscoveryStrategy#BACKGROUND background} discovery strategy.
      * <p>
      * The returned builder can be customized using {@link GrpcClientBuilderProvider}.
      *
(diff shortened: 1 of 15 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
```diff
@@ -175,6 +205,64 @@ public static SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> for
...
+     * The returned builder can be customized using {@link SingleAddressHttpClientBuilderProvider}.
+     *
+     * @param address the {@code UnresolvedAddress} to connect to, resolved using a DNS {@link ServiceDiscoverer} for
+     * every new connection. This address will also be used for the {@link HttpHeaderNames#HOST} header value.
+     * Use {@link SingleAddressHttpClientBuilder#unresolvedAddressToHost(Function)} if you want to override that
+     * value or {@link SingleAddressHttpClientBuilder#hostHeaderFallback(boolean)} if you want to disable this behavior.
+     * @return new builder for the address
+     * @see SingleAddressHttpClientBuilderProvider
+     */
+    public static SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> forSingleAddressResolveOnDemand(
```
idelpivnitskiy, line 258: Please help with a better name 🙈

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
```diff
@@ -175,6 +206,65 @@ public static SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> for
...
+            final HostAndPort address) {
+        return applyProviders(address,
+                new DefaultSingleAddressHttpClientBuilder<>(address, unresolvedServiceDiscoverer()))
+                // Apply after providers to let them see these customizations.
+                .appendConnectionFactoryFilter(ResolvingConnectionFactoryFilter.INSTANCE)
+                .retryServiceDiscoveryErrors(NoRetriesStrategy.INSTANCE);
```
idelpivnitskiy, line 265: @daschl for context: this approach requires 3 (or at least 2) settings on the builder:
1. swapping default SD with `unresolvedServiceDiscoverer()`;
2. adding `ResolvingConnectionFactoryFilter`
3. (optional) disabling `retryServiceDiscoveryErrors` bcz retries are not required for `unresolvedServiceDiscoverer()`

I was trying to find a way how to simplify users' experience for configuring all required components with a single API call. As the result I introduced these new factories on `HttpClients`. 
I also considered something like `HttpClientFunctions` that will effectively be `UnaryOperator<SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress>>` and users will apply it to the builder when the want this mode, but found it a bit difficult for users to discover. 

I'm open to any other ideas how to make users experience better.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
```diff
@@ -175,6 +206,65 @@ public static SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> for
...
+     * value or {@link SingleAddressHttpClientBuilder#hostHeaderFallback(boolean)} if you want to disable this behavior.
+     * @return new builder for the address
+     * @see SingleAddressHttpClientBuilderProvider
+     */
+    public static SingleAddressHttpClientBuilder<HostAndPort, InetSocketAddress> forSingleAddressResolveOnDemand(
+            final HostAndPort address) {
+        return applyProviders(address,
+                new DefaultSingleAddressHttpClientBuilder<>(address, unresolvedServiceDiscoverer()))
+                // Apply after providers to let them see these customizations.
+                .appendConnectionFactoryFilter(ResolvingConnectionFactoryFilter.INSTANCE)
```
idelpivnitskiy, line 264: Heads up: the position of this `ResolvingConnectionFactoryFilter` has some differences. If it's the first one in the chain, all subsequent filters will see already resolved address (like they do today with background resolutions). On the other hand, users of `TransportObserverConnectionFactoryFilter` won't be able to take DNS resolution time into account for opening a new connection. The only way for those users to properly measure time is to apply `TransportObserverConnectionFactoryFilter` via providers.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
daschl: Maybe if it is difficult to pick a name it can be worth exploring other options: we could reuse our current overloads and add one on the same name with an enum at the end i.e. (`ResolutionStrategy`) which could default to `POLL` but could be customized by the user to `ON_DEMAND`. That also leaves the door open to more customized strategies in the future if we ever need to.

daschl: Another option, but nor sure if feasible, would be to move it into the Builder itself as a configuration option?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
daschl: Is this something we should document in the javadoc of the Builder / Factory somewhere?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
daschl: I agree that wrapping this in a user-friendly way is good - see my additional comments on the potential other overload.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpClientResolvesOnDemandTest.java
```diff
@@ -0,0 +1,46 @@
...
+import static io.servicetalk.transport.netty.internal.AddressUtils.serverHostAndPort;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+class HttpClientResolvesOnDemandTest {
+
+    @ParameterizedTest(name = "{displayName} [{index}]: protocol={0}")
+    @EnumSource(HttpProtocol.class)
+    void test(HttpProtocol protocol) throws Exception {
+        try (HttpServerContext serverContext = HttpServers.forAddress(localAddress(0))
```
daschl, line 36: similar question as on GRPC if this exercises the new code path.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ResolvingConnectionFactoryFilter.java
```diff
@@ -0,0 +1,101 @@
...
+                return sd.discover(HostAndPort.of(address)).takeAtMost(1).firstOrError()
+                        .flatMap(resolvedAddresses -> {
+                            if (resolvedAddresses.size() > 1) {
+                                // In case DNS server returns multiple IPs, it's recommended to shuffle the result to
+                                // make sure the client balances load between all available IPs.
+                                List<ServiceDiscovererEvent<InetSocketAddress>> list =
+                                        resolvedAddresses instanceof List ?
+                                                (List<ServiceDiscovererEvent<InetSocketAddress>>) resolvedAddresses :
+                                                new ArrayList<>(resolvedAddresses);
+                                Collections.shuffle(list);
```
daschl, line 78: From reading this I assume shuffling is the best thing we can do (compared to keeping state for a "true" round robin)?

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcClientResolvesOnDemandTest.java
```diff
@@ -0,0 +1,48 @@
...
+    void test() throws Exception {
+        String greetingPrefix = "Hello ";
+        String name = "foo";
+        String expectedResponse = greetingPrefix + name;
+        try (GrpcServerContext serverContext = GrpcServers.forAddress(localAddress(0))
+                .listenAndAwait((GreeterService) (ctx, request) ->
+                        succeeded(HelloReply.newBuilder().setMessage(greetingPrefix + request.getName()).build()));
+             BlockingGreeterClient client = GrpcClients.forAddressResolveOnDemand(serverHostAndPort(serverContext))
+                     .buildBlocking(new ClientFactory())) {
+            assertEquals(expectedResponse,
```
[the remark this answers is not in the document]
idelpivnitskiy: ideally, always hamcrest because it generates better description messages in case of failures. but we are not always consistent. hope it will improve over-time.
I copied this test from `GrpcUdsTest` and modified without noticing it uses junit. I will adjust both tests

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcClientResolvesOnDemandTest.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class GrpcClientResolvesOnDemandTest {
+
+    @Test
+    void test() throws Exception {
+        String greetingPrefix = "Hello ";
+        String name = "foo";
+        String expectedResponse = greetingPrefix + name;
+        try (GrpcServerContext serverContext = GrpcServers.forAddress(localAddress(0))
```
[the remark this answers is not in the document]
idelpivnitskiy: For a new server, we always want to use `localAddress(0)` because it helps to avoid running on `0.0.0.0` address. That address might be restricted in some environments and cause test failures.

But I will modify the client factory to use "localhost" instead of taking hostname from `serverContext` for better clarity. Good call!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ResolvingConnectionFactoryFilter.java
idelpivnitskiy: Yes, our best effort to somehow distribute where the client connects to.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
idelpivnitskiy: Actually, users will not be able to see it regardless of the order. I documented it as the last item on `DiscoveryStrategy.ON_NEW_CONNECTION` and took some notes to create follow-up tickets.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
idelpivnitskiy: I really like the `enum` idea with overloads. It makes the API clear and gives a lot of freedom to document each strategy. Thanks a lot for the suggestion!

idelpivnitskiy: Actually, your idea helped to solve another problem: how users can customize SD for this discovery mode. I made the necessary adjustments in https://github.com/apple/servicetalk/pull/2520/commits/c1779b28f78b592b250b5848208f5bae29511e47 and it's possible now 🎉 
Also helped to test all edge cases.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
```diff
@@ -335,4 +478,112 @@ public Completable closeAsyncGracefully() {
...
+        @Override
+        public Completable apply(final int i, final Throwable t) {
+            return Completable.failed(t);
+        }
+    }
+
+    /**
+     * Defines {@link ServiceDiscoverer} will be used.
+     */
+    public enum DiscoveryStrategy {
```
daschl, line 522: 👍 great explanation of each strategy.

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcClients.java
```diff
@@ -90,6 +91,41 @@ public static GrpcClientBuilder<HostAndPort, InetSocketAddress> forAddress(final
...
+     * The returned builder can be customized using {@link GrpcClientBuilderProvider}.
+     *
+     * @param host host to connect to, resolved by default using a DNS {@link ServiceDiscoverer}.
+     * @param port port to connect to
+     * @param discoveryStrategy {@link DiscoveryStrategy} to use
+     * @return new builder for the address
+     * @see GrpcClientBuilderProvider
+     */
+    public static GrpcClientBuilder<HostAndPort, InetSocketAddress> forAddress(
+            final String host, final int port, final DiscoveryStrategy discoveryStrategy) {
```
daschl, line 107: Consider moving the `DiscoveryStrategy` to a namespace where it is not tied to http or grpc, since it could be technically also used for other protocols?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ResolvingConnectionFactoryFilter.java
```diff
@@ -0,0 +1,139 @@
...
+                        .flatMap(resolvedAddresses -> {
+                            @Nullable
+                            ServiceDiscovererEvent<R> resolved;
+                            if (resolvedAddresses.isEmpty()) {
+                                resolved = null;
+                            } else if (resolvedAddresses.size() == 1) {
+                                resolved = resolvedAddresses instanceof List ?
+                                        ((List<ServiceDiscovererEvent<R>>) resolvedAddresses).get(0) :
+                                        resolvedAddresses.stream().findFirst().orElse(null);
+                                if (!AVAILABLE.equals(resolved.status())) {
```
daschl, line 86: if resolved is not a list and nothing found (so orElse(null)) triggers (I know it's very unlikely) would this resolved.status() lead to a NPE?

daschl (approved):
looks good to me, just some minor comments to consider.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ResolvingConnectionFactoryFilter.java
idelpivnitskiy: This path executes only when `resolvedAddresses.size() == 1` but I will add a check anyway to avoid the worst case scenario if there is a tricky collection impl.
I used `resolvedAddresses.stream().findFirst().orElse(null)` as a short way instead of getting an Iterator and checking `hasNext()`.

PR #2521, opened 2023-02-23
Extend DnsServiceDiscovererObserver functionality

daschl:
Motivation:

To enhance the utility of the DnsServiceDiscovererObserver a couple additional APIs can be provided:

 - The newly introduced ServiceDiscoverer ID is brought in scope
 - Further lifecycle callbacks are added to the DnsDiscoveryObserver

Modifications:

This changeset adds the two enhancements mentioned in the motivation in a backwards-compatible way. In the next minor released the interface can be cleaned up.

Tests have been enhanced to test the newly added functionality.

Result:

More functionality around the DnsServiceDiscovererObserver.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -24,6 +24,7 @@ import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
 import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.concurrent.api.PublisherOperator;
+import io.servicetalk.concurrent.api.TerminalSignalConsumer;
 import io.servicetalk.concurrent.api.internal.SubscribablePublisher;
 import io.servicetalk.concurrent.internal.CancelImmediatelySubscriber;
 import io.servicetalk.concurrent.internal.DuplicateSubscribeException;
@@ -230,22 +231,30 @@ final class DefaultDnsClient implements DnsClient {
             Publisher<? extends Collection<ServiceDiscovererEvent<InetAddress>>> events = inactiveEventsOnError ?
                     recoverWithInactiveEvents(pub, false) :
                     pub;
-            return events
-                    .afterCancel(() -> {
-                        if (discoveryObserver != null) {
-                            discoveryObserver.discoveryCanceled();
-                        }
-                    })
-                    .afterOnError(cause -> {
-                        if (discoveryObserver != null) {
+            return discoveryObserver == null ? events : events.beforeFinally(new TerminalSignalConsumer() {
+                    @Override
+                    public void onComplete() {
+                        // this event will never be triggered
+                    }
+
+                    @Override
+                    public void onError(final Throwable cause) {
+                        try {
                             discoveryObserver.discoveryFailed(cause);
+                        } catch (Throwable err) {
+                            LOGGER.warn("Unexpected exception from observer while reporting error event", err);
                         }
-                    })
-                    .afterOnComplete(() -> {
-                        if (discoveryObserver != null) {
-                            discoveryObserver.discoveryCompleted();
+                    }
+
+                    @Override
+                    public void cancel() {
+                        try {
+                            discoveryObserver.discoveryCanceled();
+                        } catch (Throwable err) {
+                            LOGGER.warn("Unexpected exception from observer while reporting cancel event", err);
                         }
-                    });
+                    }
+                });
         });
     }
 
@@ -261,7 +270,8 @@ final class DefaultDnsClient implements DnsClient {
         // We "recover" unconditionally to force inactive events to propagate to all mapped A* publishers to cancel
         // any pending scheduled tasks. SrvInactiveCombinerOperator is used to filter the aggregated collection of
         // inactive events if necessary.
-        return recoverWithInactiveEvents(new SrvRecordPublisher(serviceName, discoveryObserver), true)
+        Publisher<Collection<ServiceDiscovererEvent<InetSocketAddress>>> events =
+                recoverWithInactiveEvents(new SrvRecordPublisher(serviceName, discoveryObserver), true)
                 .flatMapConcatIterable(identity())
                 .flatMapMerge(srvEvent -> {
                 assertInEventloop();
@@ -279,7 +289,7 @@ final class DefaultDnsClient implements DnsClient {
                                 recoverWithInactiveEvents(aPublisher, false);
                         return srvFilterDuplicateEvents ?
                                 srvFilterDups(returnPub, availableAddresses, srvEvent.address().port()) :
-                                returnPub.map(events -> mapEventList(events, inetAddress ->
+                                returnPub.map(ev -> mapEventList(ev, inetAddress ->
                                         new InetSocketAddress(inetAddress, srvEvent.address().port())));
                     }).retryWhen((i, cause) -> {
                         assertInEventloop();
@@ -301,20 +311,30 @@ final class DefaultDnsClient implements DnsClient {
                     return empty();
                 }
             }, srvConcurrency)
-            .liftSync(inactiveEventsOnError ? SrvInactiveCombinerOperator.EMIT : SrvInactiveCombinerOperator.NO_EMIT)
-            .afterCancel(() -> {
-                if (discoveryObserver != null) {
-                    discoveryObserver.discoveryCanceled();
+            .liftSync(inactiveEventsOnError ? SrvInactiveCombinerOperator.EMIT : SrvInactiveCombinerOperator.NO_EMIT);
+
+            return discoveryObserver == null ? events : events.beforeFinally(new TerminalSignalConsumer() {
+                @Override
+                public void onComplete() {
+                    // this event will never be triggered
                 }
-            })
-            .afterOnError(cause -> {
-                if (discoveryObserver != null) {
-                    discoveryObserver.discoveryFailed(cause);
+
+                @Override
+                public void onError(final Throwable cause) {
+                    try {
+                        discoveryObserver.discoveryFailed(cause);
+                    } catch (Throwable err) {
+                        LOGGER.warn("Unexpected exception from observer while reporting error event", err);
+                    }
                 }
-            })
-            .afterOnComplete(() -> {
-                if (discoveryObserver != null) {
-                    discoveryObserver.discoveryCompleted();
+
+                @Override
+                public void cancel() {
+                    try {
+                        discoveryObserver.discoveryCanceled();
+                    } catch (Throwable err) {
+                        LOGGER.warn("Unexpected exception from observer while reporting cancel event", err);
+                    }
                 }
             });
         });
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -96,6 +96,7 @@ import static io.servicetalk.transport.netty.internal.BuilderUtils.socketChannel
 import static io.servicetalk.transport.netty.internal.EventLoopAwareNettyIoExecutors.toEventLoopAwareNettyIoExecutor;
 import static io.servicetalk.utils.internal.ThrowableUtils.addSuppressed;
 import static java.lang.Integer.toHexString;
+import static java.lang.System.identityHashCode;
 import static java.nio.ByteBuffer.wrap;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
@@ -156,7 +157,7 @@ final class DefaultDnsClient implements DnsClient {
         this.ttlJitterNanos = ttlJitterNanos;
         this.observer = observer;
         this.missingRecordStatus = missingRecordStatus;
-        this.id = id;
+        this.id = id + " (instance @" + toHexString(identityHashCode(this)) + ')';
         asyncCloseable = toAsyncCloseable(graceful -> {
             if (nettyIoExecutor.isCurrentThreadEventLoop()) {
                 closeAsync0();
@@ -200,7 +201,7 @@ final class DefaultDnsClient implements DnsClient {
 
     @Override
     public String toString() {
-        return id + " (instance @" + toHexString(hashCode()) + ')';
+        return id;
     }
 
     // visible for testing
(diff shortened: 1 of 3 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -226,8 +226,27 @@ private DnsDiscoveryObserver newDiscoveryObserver(final String address) {
...
+                    recoverWithInactiveEvents(pub, false) :
+                    pub;
+            return events
+                    .afterCancel(() -> {
+                        if (discoveryObserver != null) {
+                            discoveryObserver.discoveryCanceled();
+                        }
+                    })
+                    .afterOnError(cause -> {
+                        if (discoveryObserver != null) {
```
idelpivnitskiy, line 241: Consider checking this before applying an operator to minimize number of operators on the hot path when there is no observer.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -226,8 +226,27 @@ private DnsDiscoveryObserver newDiscoveryObserver(final String address) {
...
-            return inactiveEventsOnError ? recoverWithInactiveEvents(pub, false) : pub;
+            final DnsDiscoveryObserver discoveryObserver = newDiscoveryObserver(address);
+            ARecordPublisher pub = new ARecordPublisher(address, discoveryObserver);
+            Publisher<? extends Collection<ServiceDiscovererEvent<InetAddress>>> events = inactiveEventsOnError ?
+                    recoverWithInactiveEvents(pub, false) :
+                    pub;
+            return events
+                    .afterCancel(() -> {
+                        if (discoveryObserver != null) {
+                            discoveryObserver.discoveryCanceled();
```
idelpivnitskiy, line 237: This is what I was talking about previously :) without "catch-all" observer wrapper it's so easy to miss a try-catch for a new callback. 
`DnsServiceDiscovererObserver` was the first observer we implemented (before `TransportObserver`). Here, the need for "catch-all" wrapper was less obvious because the entire scope is within one file. Will be nice to make it consistent for `DnsServiceDiscovererObserver` too, but not necessary for this PR as it will expand the scope. Let's add try-catches for now and we can introduce "catch-all" wrapper in a follow-up.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserverTest.java
```diff
@@ -62,6 +67,7 @@ class DnsServiceDiscovererObserverTest {
     private static final String SERVICE_NAME = "servicetalk";
     private static final String INVALID = "invalid.";
     private static final int DEFAULT_TTL = 1;
+    private static final String DISCOVERER_ID = "test";
```
idelpivnitskiy, line 70: Consider using `getClass().getSimpleName()` to guarantee that each test class specifies its own id.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserverTest.java
```diff
@@ -107,9 +113,37 @@ void srvQueryTriggersNewDiscoveryObserver() throws Exception {
...
+            @Override
+            public DnsDiscoveryObserver onNewDiscovery(final String name) {
+                fail("This method must not be called anymore when overridden");
+                return null;
+            }
+
+            @Override
+            public DnsDiscoveryObserver onNewDiscovery(final String serviceDiscovererId, final String name) {
+                newDiscoveryCalls.add(name);
+                assertEquals(DISCOVERER_ID, serviceDiscovererId);
```
idelpivnitskiy, line 127: We are not always consistent, but we try to use hamcrest for assertions because it generates better messages compare to jUnit.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserverTest.java
```diff
@@ -107,9 +113,37 @@ void srvQueryTriggersNewDiscoveryObserver() throws Exception {
...
                                           String expectedName) throws Exception {
         BlockingQueue<String> newDiscoveryCalls = new LinkedBlockingDeque<>();
-        DnsClient client = dnsClient(name -> {
-            newDiscoveryCalls.add(name);
-            return NoopDnsDiscoveryObserver.INSTANCE;
+        final AtomicBoolean discoveryCanceledCalled = new AtomicBoolean();
+        DnsClient client = dnsClient(new DnsServiceDiscovererObserver() {
+            @Override
+            public DnsDiscoveryObserver onNewDiscovery(final String name) {
+                fail("This method must not be called anymore when overridden");
```
idelpivnitskiy, line 120: When ST plugs in any observer, we always try-catch its callbacks to make sure misbehaving observability does not impact the main execution flow. This assertion will not fail the test. Instead, consider using a `BlockingQueue` where you can add `AssertionException` and they verify that the queue is empty at the end of the test. We have a utility for that, see `TestUtils.assertNoAsyncErrors(...)` and how it's used by other tests.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserverTest.java
```diff
@@ -107,9 +113,37 @@ void srvQueryTriggersNewDiscoveryObserver() throws Exception {
...
+                        // the takeAtMost operator below will trigger a cancellation, not a completion event
+                        discoveryCanceledCalled.set(true);
+                    }
+
+                    @Override
+                    public void discoveryCompleted() { }
+
+                    @Override
+                    public void discoveryFailed(final Throwable cause) { }
+                };
```
idelpivnitskiy, line 145: Consider verifying that this method was not invoked.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserverTest.java
```diff
@@ -155,6 +190,47 @@ void srvQueryTriggersNewResolutionObserver() throws Exception {
...
+
+        Publisher<?> publisher = publisherFactory.apply(client, INVALID);
+        assertThat("Unexpected calls to discoveryFailed(t)", discoveryFailures, hasSize(0));
+        // Wait until SD returns at least one address:
+        ExecutionException ee = assertThrows(ExecutionException.class,
+                () -> publisher.takeAtMost(1).ignoreElements().toFuture().get());
+        Throwable cause = ee.getCause();
+        assertThat(cause, instanceOf(UnknownHostException.class));
+        assertThat("Unexpected number of calls to discoveryFailed(t)", discoveryFailures, hasSize(1));
+        assertThat("Unexpected name for discoveryFailed(t)", discoveryFailures, hasItem(sameInstance(cause)));
```
idelpivnitskiy, line 231: name -> cause?

idelpivnitskiy (changes requested):

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -226,8 +226,27 @@ private DnsDiscoveryObserver newDiscoveryObserver(final String address) {
...
         return defer(() -> {
-            ARecordPublisher pub = new ARecordPublisher(address, newDiscoveryObserver(address));
-            return inactiveEventsOnError ? recoverWithInactiveEvents(pub, false) : pub;
+            final DnsDiscoveryObserver discoveryObserver = newDiscoveryObserver(address);
+            ARecordPublisher pub = new ARecordPublisher(address, discoveryObserver);
+            Publisher<? extends Collection<ServiceDiscovererEvent<InetAddress>>> events = inactiveEventsOnError ?
+                    recoverWithInactiveEvents(pub, false) :
+                    pub;
+            return events
+                    .afterCancel(() -> {
```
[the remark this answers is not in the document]
daschl: That's great context, thanks for sharing!

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserver.java
```diff
@@ -30,9 +30,23 @@ public interface DnsServiceDiscovererObserver {
      * @param name the name of DNS record to be discovered
      * @return {@link DnsDiscoveryObserver} that provides visibility into individual DNS resolutions behind the
      * associated discovery
+     * @deprecated use {@link #onNewDiscovery(String, String)} instead.
      */
+    @Deprecated // FIXME 0.43: remove deprecated method
```
idelpivnitskiy, line 35: Because `DnsServiceDiscovererObserver` is a `@FunctionalInterface` right now, we won't be able to simply remove this method in the next release. Consider describing the recommended approach for users, similar to this example: https://github.com/apple/servicetalk/pull/2525/files#diff-5d32e457fd45a95d48cdea9552a11778b7063fe42b5ff9ef6ea5c792a28c42e9R74-R77

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserver.java
```diff
@@ -46,6 +60,18 @@ interface DnsDiscoveryObserver {
          * @return {@link DnsResolutionObserver} that provides visibility into results of the current DNS resolution
          */
         DnsResolutionObserver onNewResolution(String name);
+
+        /**
+         * Notifies that the current DNS discovery got canceled (did not complete successfully).
```
idelpivnitskiy, line 65: Consider describing to users the expected behavior of terminal callbacks. See `HttpLifecycleObserver` as an example:
https://github.com/apple/servicetalk/blob/b4543f4a42111758903d9fdbcd795b1be4b71a61/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLifecycleObserver.java#L104-L110

https://github.com/apple/servicetalk/blob/b4543f4a42111758903d9fdbcd795b1be4b71a61/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLifecycleObserver.java#L146-L153

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -226,8 +227,35 @@ private DnsDiscoveryObserver newDiscoveryObserver(final String address) {
...
+                    public void onComplete() {
+                        // this event will never be triggered
+                    }
+
+                    @Override
+                    public void onError(final Throwable cause) {
+                        try {
+                            discoveryObserver.discoveryFailed(cause);
+                        } catch (Throwable err) {
+                            LOGGER.warn("Unexpected exception from observer while reporting error event", err);
```
idelpivnitskiy, line 246: 1. Consider adding `cause` as suppressed exception for `err`, similar to `resolutionFailed`:
https://github.com/apple/servicetalk/blob/38d5166d0633da240371a29391bcc29f97b63246/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java#L784-L790

2. Consider "while reporting error event" -> "while reporting discovery failure" to give users a hint what callback we tried to invoke.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -226,8 +227,35 @@ private DnsDiscoveryObserver newDiscoveryObserver(final String address) {
...
+                        }
+                    }
+
+                    @Override
+                    public void cancel() {
+                        try {
+                            discoveryObserver.discoveryCanceled();
+                        } catch (Throwable err) {
+                            LOGGER.warn("Unexpected exception from observer while reporting cancel event", err);
```
idelpivnitskiy, line 255: Consider "while reporting cancel event" -> "while reporting discovery cancelled"

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserver.java
```diff
@@ -46,6 +60,18 @@ interface DnsDiscoveryObserver {
...
+         */
+        default void discoveryCanceled() { } // FIXME: 0.43 remove default
```
idelpivnitskiy, line 67: I know that both versions are grammatically correct. Only for consistency with other public API we have, consider using double LL: `discoveryCancelled()`.

Example:
https://github.com/apple/servicetalk/blob/d45724ddbda177b91d0426b7737757e506218bfb/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/SequentialCancellable.java#L84-L85

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObserverTest.java
```diff
@@ -107,9 +113,38 @@ void srvQueryTriggersNewDiscoveryObserver() throws Exception {
...
+            @Override
+            public DnsDiscoveryObserver onNewDiscovery(final String name) {
+                errors.add(new AssertionError("This method must not be called anymore when overridden"));
+                return ignored -> NoopDnsResolutionObserver.INSTANCE;
+            }
+
+            @Override
+            public DnsDiscoveryObserver onNewDiscovery(final String serviceDiscovererId, final String name) {
+                newDiscoveryCalls.add(name);
+                assertThat("Unexpected serviceDiscovererId", serviceDiscovererId, equalTo(DISCOVERER_ID));
```
idelpivnitskiy, line 129: This assertion will not fail the test because we always try-catch observer callbacks internally. Consider using `errors` queue. You can create your own `AssertionError` or try-catch `assertThat`.

idelpivnitskiy (approved):
Everything looks good! Only minor recommendations for working and 🚢 it

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -214,7 +215,7 @@ private DnsDiscoveryObserver newDiscoveryObserver(final String address) {
             return null;
         }
         try {
-            return observer.onNewDiscovery(address);
+            return observer.onNewDiscovery(id, address);
```
idelpivnitskiy, line 218: Btw, while you are here, how do you think should we do in the constructor something like this?
```
this.id = id + '@' + toHexString(identityHashCode(this));
```
Then we can simplify `toString()` to be:
```java
    @Override
    public String toString() {
        return id;
    }
```
There is a risk of having multiple instances built be the same builder. All of them will have the same id. `identityHashCode` will always help the observer to see unique IDs.

PR #2523, opened 2023-02-24
Apply `TestDnsServiceDiscovererBuilderProvider` only for a specific id

idelpivnitskiy:
Motivation:

Gradle can run test classes in different order. In case it runs `DnsServiceDiscovererBuilderProviderTest` after any other test, the numbers are off because existing `TestDnsServiceDiscovererBuilderProvider` is applied to any builder.

Modifications:

- Apply `TestDnsServiceDiscovererBuilderProvider` logic only for `DnsServiceDiscovererBuilderProviderTest`;

Result:

Tests pass regardless of the execution order.


idelpivnitskiy:
Example of the build failure: https://github.com/apple/servicetalk/pull/2518/checks?check_run_id=11541294089

daschl (approved):

PR #2524, opened 2023-02-26
Don't allow to override ServiceDiscoverer for resolved address client

idelpivnitskiy:
Motivation:

`HttpClients.forResolvedAddress(...)` uses a special `ServiceDiscoverer`
that users should not override. Otherwise, behavior will be
unpredictable.

Modifications:

- Use `withUnmodifiableServiceDiscoverer` to protect from overriding the
`ServiceDiscoverer`;
- Add `ForResolvedAddressTest` to validate this behavior;

Result:

Users can not change `ServiceDiscoverer` used by `HttpClients
.forResolvedAddress(...)`. In case they need a custom SD for this use
case, they can use `HttpClients.forSingleAddress(ServiceDiscoverer, U)`.

---

Depends on #2520, review only the latest commit: https://github.com/apple/servicetalk/pull/2524/commits/3713d8dc64cecdda0fb56e6f07d85db71eb41ebc


daschl (approved):
Last commit looks good.

PR #2511, opened 2023-02-17
Implement DnsServiceDiscovererBuilderProvider

daschl:
Motivation:

Similar to the related functionality in GRPC and HTTP servers, it can be useful to configure the DNS service discoverer also through a ServiceProvider.

Modifications:

This changeset introduces ServiceProvider functionality on top of the DefaultDnsServiceDiscovererBuilder. To do this, a new interface is introduced (to also provide a Delegate implementation) as well as a DnsServiceDiscoverers factory class to construct a builder with an ID.

Note that uniqueness on this ID is not enforced by ServiceTalk, but is something to recommend to the application developer when using this mechanism. The default construct is marked as deprecated and will use a UUID as the ID.

Result:

It is now possible to configure the DNS service discoverer through the ServiceProvider mechanism.

PR #2525, opened 2023-02-26
Always required an id for creating a `MultiAddressHttpClientBuilder`

idelpivnitskiy:
`MultiAddressHttpClientBuilderProvider` doesn't have a way to
distinguish different `MultiAddressHttpClientBuilder`(s). In https://github.com/apple/servicetalk/pull/2511 we
introduced "id" for `DnsServiceDiscovererBuilderProvider` to solve this
problem. Let's use a similar approach for `MultiAddressHttpClientBuilder`.

Modifications:

- Introduce `HttpClients.forMultiAddressUrl(String)` factory;
- Change recently added overload `forMultiAddressUrl(DiscoveryStrategy)`
to `forMultiAddressUrl(String, DiscoveryStrategy)`;
- Deprecate `MultiAddressHttpClientBuilderProvider#newBuilder(builder)`;
- Introduce `MultiAddressHttpClientBuilderProvider#newBuilder(id,
builder)` instead;
- Update all existing tests to use new API;

Result:

Users can provide id/name/label for the `MultiAddressHttpClientBuilder`.
This id can be used by `MultiAddressHttpClientBuilderProvider` to distinguish
different builders before deciding what modifications to apply.

---

This PR depends on #2520, review only the last commit: https://github.com/apple/servicetalk/pull/2525/commits/ccd94f8d3d7fe6431dfafd45da7983af30007b4e

### DIFF
--- a/.github/workflows/ci-prb-reports.yml
+++ b/.github/workflows/ci-prb-reports.yml
@@ -23,7 +23,7 @@ jobs:
           # File location set in ci-prb.yml and must be coordinated.
           name: test-results-${{ matrix.os }}-${{ matrix.java }}
       - name: Publish Test Report
-        uses: scacap/action-surefire-report@4cbb611f23a9b5a030c73c2b32190a80983956ed
+        uses: scacap/action-surefire-report@a2d14baac06028516812d93c36cb9c99aa50198e
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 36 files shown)


daschl (approved):
Reviewed the last commit and looks good to me, the only thing we can consider if we should deprecate/remove the non-id method or if we should keep it for users convenience (which would then just have the uuid in?)

PR #2526, opened 2023-02-26
`DnsServiceDiscovererObserver` should always report adjusted TTL

idelpivnitskiy:
Motivation:

#2508 introduced `maxTTL` setting. When we report TTL value in `ResolutionResult`, it's always adjusted by `minTTL` but not by `maxTTL`.

Modifications:

- Move `ttlNanos` adjustment before `reportResolutionResult`;
- Add tests to verify that TTL is adjusted by `maxTTL` value;

Result:

`DnsServiceDiscovererObserver` always sees the correct TTL value that will be used to schedule the next DNS query.


daschl (approved):
Ah, rearranging that was easier than I thought in the DefaultDnsClient- nice!

PR #2527, opened 2023-02-27
Implement RoundRobinLoadBalancerBuilderProvider

daschl:
Motivation:

It can be useful to configure the RoundRobinLoadBalancer through a ServiceProvider, similar to other components already present in ServiceTalk.

Modifications:

This changeset introduces ServiceProvider functionality on top of the RoundRobinLoadBalancerBuilder. To do this, the builder has been refactored into an interface and a delegate implementation is also introduced.

Result:

It is now possible to configure the RoundRobinLoadBalancer through the ServiceProvider mechanism.

### DIFF
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
@@ -19,7 +19,6 @@ import io.servicetalk.client.api.ConnectionFactory;
 import io.servicetalk.client.api.LoadBalancedConnection;
 import io.servicetalk.client.api.LoadBalancer;
 import io.servicetalk.client.api.LoadBalancerFactory;
-import io.servicetalk.client.api.ServiceDiscoverer;
 import io.servicetalk.client.api.ServiceDiscovererEvent;
 import io.servicetalk.concurrent.api.DefaultThreadFactory;
 import io.servicetalk.concurrent.api.Executor;
@@ -31,6 +30,7 @@ import io.servicetalk.transport.api.ExecutionStrategy;
 
 import java.time.Duration;
 import java.util.Collection;
+import java.util.UUID;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.function.Predicate;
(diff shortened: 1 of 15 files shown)


servicetalk-loadbalancer/src/test/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerBuilderProviderTest.java
```diff
@@ -0,0 +1,58 @@
...
+        assertThat("TestRoundRobinLoadBalancerBuilderProvider not called", buildCounter.get(), is(1));
+        assertThat("Builder method not intercepted", linearSearchSpaceIntercept.get(), is(1234L));
+    }
+
+    public static final class TestRoundRobinLoadBalancerBuilderProvider
+            implements RoundRobinLoadBalancerBuilderProvider {
+        @Override
+        public <ResolvedAddress, C extends LoadBalancedConnection> RoundRobinLoadBalancerBuilder<ResolvedAddress, C>
+        newBuilder(final RoundRobinLoadBalancerBuilder<ResolvedAddress, C> builder) {
+            buildCounter.incrementAndGet();
```
idelpivnitskiy, line 46: If gradle decides to run any other test class in this package before the current one, numbers will be off. 
After we have an id, we can use the same trick as in https://github.com/apple/servicetalk/pull/2523

For cases when there is no id, I used a flag to control when the provider should be applied: https://github.com/apple/servicetalk/blob/a5f73b39641f7344434820895a0d0556e6486377/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpProvidersTest.java#L299
And always reset numbers before running each test: https://github.com/apple/servicetalk/blob/a5f73b39641f7344434820895a0d0556e6486377/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpProvidersTest.java#L292-L295

idelpivnitskiy (changes requested):

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -124,9 +124,13 @@ public ExecutionStrategy requiredOffloads() {
      *
      * @param <ResolvedAddress> The resolved address type.
      * @param <C> The type of connection.
+     * @deprecated this class will be made package-private in the future, rely on the
+     * {@link RoundRobinLoadBalancerBuilder} instead.
      */
+    @Deprecated // FIXME: 0.43 - make package private
```
idelpivnitskiy, line 130: Consider deprecating the top level `RoundRobinLoadBalancerFactory` class too

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancerFactory.java
```diff
@@ -140,24 +144,18 @@ public ExecutionStrategy requiredOffloads() {
...
+         *
+         * @deprecated use {@link RoundRobinLoadBalancers#builder(String)} instead.
          */
+        @Deprecated // FIXME: 0.43 - remove deprecated constructor
         public Builder() {
+            this(UUID.randomUUID().toString());
+        }
+
+        Builder(final String id) {
+            this.id = id;
```
idelpivnitskiy, line 156: Either here or in `RoundRobingLoadBalancers`, consider adding a check that the `id` is not empty

idelpivnitskiy (approved):
Thank you!

PR #2531, opened 2023-03-03
Enhance `SniTest` to verify that SNI is present in `SSLSession`

idelpivnitskiy:
Motivation:

It may be necessary for users to read SNI used for the SSL handshake to apply different policies. Let's make sure the SNI is always present in `SSLSession`.

Modifications:

- Enhance `SniTest` to validate that `getRequestedServerNames()` contains value regardless of the used `SslProvider` and that `SNIHostName` matches the expected value;

Result:

Better test coverage for SNI use-case.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SniTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SniTest.java
@@ -17,7 +17,6 @@ package io.servicetalk.http.netty;
 
 import io.servicetalk.http.api.BlockingHttpClient;
 import io.servicetalk.http.api.BlockingHttpService;
-import io.servicetalk.http.api.HttpProtocolConfig;
 import io.servicetalk.http.api.HttpRequest;
 import io.servicetalk.http.api.HttpResponseStatus;
 import io.servicetalk.http.api.ReservedBlockingHttpConnection;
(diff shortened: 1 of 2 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SniTest.java
```diff
@@ -26,109 +25,128 @@
...
+                                                 HttpProtocol protocol, boolean useALPN) throws Exception {
+        try (ServerContext serverContext = newServerBuilder(SERVER_CTX, protocol)
+                .sslConfig(trustedServerConfig(serverSslProvider, alpnIds(protocol, useALPN)),
+                        singletonMap("no_match" + SNI_HOSTNAME, untrustedServerConfig(serverSslProvider)))
                 .listenBlockingAndAwait(newSslVerifyService());
              BlockingHttpClient client = HttpClients.forSingleAddress(
                      getLoopbackAddress().getHostName(),
                      serverHostAndPort(serverContext).port())
-                     .protocols(protocolConfigs(protocols))
+                     .protocols(protocol.config)
```
daschl, line 147: I see that this PR changes from a list of protocols to a single protocol (everywhere?). For my own curiosity, are we loosing any test coverage now by providing only one vs. multiple?

idelpivnitskiy: No, the previous code always used `singletonList`

daschl (approved):

PR #2532, opened 2023-03-06
Allow to wrap Executor with custom TimeSource

daschl:
Motivation:

When an Executor needs to wrapped with a custom TimeSource (i.e for testing) it is possible, but not very obvious or user friendly at the moment.

Modifications:

This changeset introduces a factory method to help delegate calles from a provided executor and also use a custom TimeSource.

Result:

It is now easier to use an executor with a custom TimeSource if needed.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Executors.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Executors.java
@@ -15,6 +15,8 @@
  */
 package io.servicetalk.concurrent.api;
 
+import io.servicetalk.concurrent.TimeSource;
+
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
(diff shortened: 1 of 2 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Executors.java
```diff
@@ -254,4 +256,23 @@ public static Executor from(java.util.concurrent.Executor jdkExecutor,
...
+     *
+     * @param delegate the {@link Executor} which should use the provided time source.
+     * @param timeSource the {@link TimeSource} that should be used when calling {@link Executor#currentTime(TimeUnit)}.
+     * @return The wrapped {@link Executor} with a custom {@link TimeSource}.
+     */
+    public static Executor withTimeSource(final Executor delegate, final TimeSource timeSource) {
+        return new DelegatingExecutor(delegate) {
+            @Override
+            public long currentTime(final TimeUnit unit) {
+                return timeSource.currentTime(unit);
```
idelpivnitskiy, line 274: It's not enough just to swap the `TimeSource` because all scheduled tasks will still depend on the real time.
Since the primary use-case is testing, consider not adding this on the main public API. 
There is already a test-fixture `TestExecutor` that allows control of the time source and execution of the tasks.

idelpivnitskiy (changes requested):

daschl:
Good points, closing this for now and will reevaluate if needed.

PR #2536, opened 2023-03-09
Update tags for OpenTelemetry

aleqi200:
- http.url will not have localhost as it's not available all the time
- added http.target, http.route, http.flavor, http.user_agent, http.scheme, net.host.name, net.host.port
- allow http client filter to use host as default value, making component name optional

### DIFF
--- a/servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
+++ b/servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
@@ -36,7 +36,6 @@ import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 
-import java.util.Optional;
 import java.util.function.UnaryOperator;
 import javax.annotation.Nullable;
 
@@ -134,6 +133,10 @@ public final class OpenTelemetryHttpRequestFilter extends AbstractOpenTelemetryF
         if (componentName != null && !componentName.trim().isEmpty()) {
             return componentName;
         }
-        return Optional.ofNullable(request.host()).orElse(request.requestTarget());
+        String host = request.host();
+        if (host != null) {
+            return host;
+        }
+        return request.requestTarget();
     }
 }
--- a/servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
+++ b/servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
@@ -27,6 +27,7 @@ import io.servicetalk.http.api.StreamingHttpConnectionFilterFactory;
 import io.servicetalk.http.api.StreamingHttpRequest;
 import io.servicetalk.http.api.StreamingHttpRequester;
 import io.servicetalk.http.api.StreamingHttpResponse;
+import io.servicetalk.transport.api.HostAndPort;
 
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.OpenTelemetry;
@@ -128,12 +129,12 @@ public final class OpenTelemetryHttpRequestFilter extends AbstractOpenTelemetryF
     }
 
     private String getSpanName(StreamingHttpRequest request) {
-        if (componentName != null && !componentName.isEmpty()) {
+        if (!componentName.isEmpty()) {
             return componentName;
         }
-        String host = request.host();
-        if (host != null) {
-            return host;
+        HostAndPort hostAndPort = request.effectiveHostAndPort();
+        if (hostAndPort != null) {
+            return hostAndPort.hostName();
         }
         return request.requestTarget();
     }
--- a/servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
+++ b/servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
@@ -37,7 +37,6 @@ import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 
 import java.util.function.UnaryOperator;
-import javax.annotation.Nullable;
 
 /**
  * An HTTP filter that supports <a href="https://opentelemetry.io/docs/instrumentation/java/">open telemetry</a>.
@@ -54,7 +53,6 @@ import javax.annotation.Nullable;
 public final class OpenTelemetryHttpRequestFilter extends AbstractOpenTelemetryFilter
     implements StreamingHttpClientFilterFactory, StreamingHttpConnectionFilterFactory {
 
-    @Nullable
     private final String componentName;
 
     /**
@@ -63,9 +61,9 @@ public final class OpenTelemetryHttpRequestFilter extends AbstractOpenTelemetryF
      * @param openTelemetry the {@link OpenTelemetry}.
      * @param componentName The component name used during building new spans.
      */
-    public OpenTelemetryHttpRequestFilter(final OpenTelemetry openTelemetry, @Nullable String componentName) {
+    public OpenTelemetryHttpRequestFilter(final OpenTelemetry openTelemetry, String componentName) {
         super(openTelemetry);
-        this.componentName = componentName;
+        this.componentName = componentName.trim();
     }
 
     /**
@@ -73,7 +71,7 @@ public final class OpenTelemetryHttpRequestFilter extends AbstractOpenTelemetryF
      *
      * @param componentName The component name used during building new spans.
      */
-    public OpenTelemetryHttpRequestFilter(@Nullable String componentName) {
+    public OpenTelemetryHttpRequestFilter(String componentName) {
         this(GlobalOpenTelemetry.get(), componentName);
     }
 
@@ -82,7 +80,7 @@ public final class OpenTelemetryHttpRequestFilter extends AbstractOpenTelemetryF
      * using the hostname as the component name.
      */
     public OpenTelemetryHttpRequestFilter() {
-        this(GlobalOpenTelemetry.get(), null);
+        this(GlobalOpenTelemetry.get(), "");
     }
 
     @Override
@@ -130,7 +128,7 @@ public final class OpenTelemetryHttpRequestFilter extends AbstractOpenTelemetryF
     }
 
     private String getSpanName(StreamingHttpRequest request) {
-        if (componentName != null && !componentName.trim().isEmpty()) {
+        if (componentName != null && !componentName.isEmpty()) {
             return componentName;
         }
         String host = request.host();
--- a/servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
+++ b/servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
@@ -36,8 +36,9 @@ import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 
-import java.util.Objects;
+import java.util.Optional;
 import java.util.function.UnaryOperator;
+import javax.annotation.Nullable;
 
 /**
  * An HTTP filter that supports <a href="https://opentelemetry.io/docs/instrumentation/java/">open telemetry</a>.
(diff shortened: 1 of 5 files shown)


servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
```diff
@@ -111,12 +121,19 @@ private Single<StreamingHttpResponse> trackRequest(final StreamingHttpRequester
...
             return Single.failed(t);
         }
         return tracker.track(response);
     }
+
+    private String getSpanName(StreamingHttpRequest request) {
+        if (componentName != null && !componentName.trim().isEmpty()) {
+            return componentName;
+        }
+        return Optional.ofNullable(request.host()).orElse(request.requestTarget());
```
daschl, line 137: Consider just using regular if / else with null, I don't think we need to construct an Optional instance here since it gets discarded anyways.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestTagExtractor.java
```diff
@@ -32,15 +34,22 @@ private static String getRequestMethod(HttpRequestMetaData req) {
     }
 
     private static String getHttpUrl(HttpRequestMetaData req) {
-        return (req.scheme() == null ? "http" : req.scheme()) + "://" +
-            (req.effectiveHostAndPort() == null ? "localhost:8080" : req.effectiveHostAndPort())
-            + req.rawPath()
+        return req.path()
```
daschl, line 37: does this equal the code you removed, or is this a behavioral change for the tag value? If so, is this acceptable?

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpServerFilter.java
```diff
@@ -122,6 +122,6 @@ protected void tagStatusCode() {
      * @return the operation name to build the span with.
      */
     private static String getOperationName(HttpRequestMetaData metaData) {
-        return metaData.method().name() + ' ' + metaData.requestTarget();
+        return metaData.method().name() + ' ' + metaData.path();
```
daschl, line 125: What changes in this case? Is this a breaking change?

servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilterTest.java
```diff
@@ -99,14 +99,14 @@ void testInjectWithNoParent() throws Exception {
                 otelTesting.assertTraces()
                     .hasTracesSatisfyingExactly(ta ->
                         assertThat(ta.getSpan(0).getAttributes().get(SemanticAttributes.HTTP_URL))
-                            .startsWith("http://localhost:8080"));
+                            .isEqualTo("/"));
```
daschl, line 102: I see this commit changes the values as well, in the commit message can you please explicitly list what has changed and to what? So people can track down changes later when they upgrade.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestTagExtractor.java
```diff
@@ -32,15 +34,22 @@ private static String getRequestMethod(HttpRequestMetaData req) {
...
     }
 
     static Span reportTagsAndStart(SpanBuilder span, HttpRequestMetaData httpRequestMetaData) {
         span.setAttribute("http.url", getHttpUrl(httpRequestMetaData));
         span.setAttribute("http.method", getRequestMethod(httpRequestMetaData));
+        span.setAttribute("http.target", getHttpUrl(httpRequestMetaData));
+        span.setAttribute("http.route", httpRequestMetaData.rawPath());
+        span.setAttribute("http.flavor", httpRequestMetaData.version().major() + "."
+            + httpRequestMetaData.version().minor());
+        Optional.ofNullable(httpRequestMetaData.userInfo())
```
daschl, line 48: Do we need optionals here or can we just use a regular if not null?

daschl: (consider avoiding Optional allocations on the hot code path if we don't have to do it)

daschl (changes requested):

aleqi200:
Tags in Spring MVC:

Name: /hello

http.flavor | 1.1 |  
-- | -- | --
http.method | GET |  
http.response_content_length | 77 |  
http.route | /hello |  
http.scheme | http |  
http.status_code | 200 |  
http.target | /hello?test=123&some=another%20thing |  
http.user_agent | Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/110.0.0.0 Safari/537.36 |  
internal.span.format | proto |  
net.host.name | localhost |  
net.host.port | 8080 |  
net.sock.host.addr | 0:0:0:0:0:0:0:1 |  
net.sock.peer.addr | 0:0:0:0:0:0:0:1 |  
net.sock.peer.port | 53475 |  
net.transport | ip_tcp |  
otel.library.name | io.opentelemetry.tomcat-7.0 |  
otel.library.version | 1.19.2-alpha |  
span.kind | server |  
thread.id | 23 |  
thread.name | http-nio-8080-exec-3

Tags in the current ServiceTalk:

Name: GET /job?name=Alex&some=thing

http.method | GET |  
-- | -- | --
http.status_code | 200 |  
http.url | http://localhost:8082/job?name=Alex&some=thing |  
internal.span.format | proto |  
otel.library.name | io.servicetalk |  
span.kind | server |  
thread.id | 49 |  
thread.name | servicetalk-global-io-executor-1-18

This will add some extra information to bring the library in parity with other libraries, in this case comparing to Spring MVC.
The most important part is that the http.url is used as the main name which makes the span name really verbose and with some unecessary information, making it shorter will make it easier to apply filters and see it all together.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestTagExtractor.java
aleqi200: That makes sense. I removed them.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpServerFilter.java
aleqi200: I highlight the reasons below but basically, this is too verbose and hard to filter:
Screenshot 2023-03-10 at 9 54 21 AM
Compared to this (Spring MVC):
Screenshot 2023-03-10 at 9 54 08 AM
Yes that last one had some query params but it still shows it in a more concise way. There's no specification on what it should look like but I think we should avoid having it too big to make it easy to filter or apply other things to it.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestTagExtractor.java
aleqi200: behavioral change.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
aleqi200: switched to an if

daschl (approved):

idelpivnitskiy (changes requested):

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
```diff
@@ -111,12 +120,23 @@ private Single<StreamingHttpResponse> trackRequest(final StreamingHttpRequester
...
             return Single.failed(t);
         }
         return tracker.track(response);
     }
+
+    private String getSpanName(StreamingHttpRequest request) {
+        if (componentName != null && !componentName.trim().isEmpty()) {
+            return componentName;
+        }
+        String host = request.host();
```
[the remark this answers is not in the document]
aleqi200: ok, that makes sense. I changed it to that.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
```diff
@@ -62,20 +63,28 @@ public final class OpenTelemetryHttpRequestFilter extends AbstractOpenTelemetryF
...
+        this.componentName = componentName;
     }
 
     /**
      * Create a new instance, searching for any instance of an opentelemetry available.
      *
      * @param componentName The component name used during building new spans.
      */
-    public OpenTelemetryHttpRequestFilter(String componentName) {
+    public OpenTelemetryHttpRequestFilter(@Nullable String componentName) {
```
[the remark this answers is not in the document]
aleqi200: ok, I kept it as non null and no value will be empty string.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestTagExtractor.java
```diff
@@ -32,15 +32,30 @@ private static String getRequestMethod(HttpRequestMetaData req) {
...
     static Span reportTagsAndStart(SpanBuilder span, HttpRequestMetaData httpRequestMetaData) {
         span.setAttribute("http.url", getHttpUrl(httpRequestMetaData));
         span.setAttribute("http.method", getRequestMethod(httpRequestMetaData));
+        span.setAttribute("http.target", getHttpUrl(httpRequestMetaData));
+        span.setAttribute("http.route", httpRequestMetaData.rawPath());
+        span.setAttribute("http.flavor", httpRequestMetaData.version().major() + "."
+            + httpRequestMetaData.version().minor());
+        String userInfo = httpRequestMetaData.userInfo();
+        if (userInfo != null) {
+            span.setAttribute("http.user_agent", userInfo);
```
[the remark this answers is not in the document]
aleqi200: oh thanks for that. I updated it.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestTagExtractor.java
```diff
@@ -32,15 +32,30 @@ private static String getRequestMethod(HttpRequestMetaData req) {
...
+        }
+        String scheme = httpRequestMetaData.scheme();
+        if (scheme != null) {
+            span.setAttribute("http.scheme", scheme);
+        }
+        String hostName = httpRequestMetaData.host();
+        if (hostName != null) {
+            span.setAttribute("net.host.name", hostName);
+        }
+        span.setAttribute("net.host.port", httpRequestMetaData.port());
```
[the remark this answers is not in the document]
aleqi200: done

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestTagExtractor.java
```diff
@@ -32,15 +32,30 @@ private static String getRequestMethod(HttpRequestMetaData req) {
...
             + (req.rawQuery() == null ? "" : "?" + req.rawQuery());
     }
 
     static Span reportTagsAndStart(SpanBuilder span, HttpRequestMetaData httpRequestMetaData) {
         span.setAttribute("http.url", getHttpUrl(httpRequestMetaData));
         span.setAttribute("http.method", getRequestMethod(httpRequestMetaData));
+        span.setAttribute("http.target", getHttpUrl(httpRequestMetaData));
+        span.setAttribute("http.route", httpRequestMetaData.rawPath());
+        span.setAttribute("http.flavor", httpRequestMetaData.version().major() + "."
+            + httpRequestMetaData.version().minor());
```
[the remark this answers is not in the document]
idelpivnitskiy: sorry, didn't know if you finished or not and pushed https://github.com/apple/servicetalk/pull/2536/commits/27a25e2bde99c1f639c72a63c8c36093f19234ed
since this format is specific to OTEL, let's keep it private for now. If in the future we will have another case, we can add a new method on `HttpProtocolVersion`.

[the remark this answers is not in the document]
aleqi200: no problem, let's keep it with the private method to cache the version for now.

idelpivnitskiy (approved):
Thank you!

PR #2538, opened 2023-03-10
Only depend on OpenTelemetry api and not the overall BOM

daschl:
Motivation:

Depending on the BOM instead of just the minimal -api package causes downstream issues in terms of how these dependency versions are managed.

Modifications:

In order to more tighly control what we are depending on, this changeset only depends on the-api package.


daschl:
@aleqi200 this might also be interesting to you?

idelpivnitskiy (approved):

PR #2541, opened 2023-03-20
RetryingHttpRequesterFilter to duplicate payload body

Scottmitch:
Motivation:
In order for payload body to be reused the Buffers need to be duplicated. When SSL is used it will reliably move buffer indexes and if the payload is replayable the retried request may hang because no payload body is actually sent.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -221,7 +223,8 @@ public Single<? extends FilterableReservedStreamingHttpConnection> reserveConnec
         @Override
         protected Single<StreamingHttpResponse> request(final StreamingHttpRequester delegate,
                                                         final StreamingHttpRequest request) {
-            Single<StreamingHttpResponse> single = delegate.request(request);
+            final StreamingHttpRequest duplicatedRequest = request.transformMessageBody(messageBodyDuplicator());
```
idelpivnitskiy, line 226: Because most users don't customize default retry strategy, we don't need to transform the body by default. Consider applying duplicator only when users customized any of `retry*` functions on the builder.

PR #2544, opened 2023-03-28
`ProxyConnectConnectionFactoryFilter`: report when the channel is closed

idelpivnitskiy:
Motivation:

When proxy returns 2xx status code with `Connection: closed` header, users see an awkward error saying:
Failed to find a handler of type DeferSslHandler in channel pipeline. It confuses users. We can handle this case with a better exception message.

Modifications:
- Check if the channel is still active before generating `IllegalStateException`;

Result:

More clear exception message when proxy closes the connection.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
@@ -125,7 +125,8 @@ final class ProxyConnectConnectionFactoryFilter<ResolvedAddress, C extends Filte
             final DeferSslHandler deferSslHandler = channel.pipeline().get(DeferSslHandler.class);
             if (deferSslHandler == null) {
                 if (!channel.isActive()) {
-                    LOGGER.info("{} is unexpectedly closed after receiving response: {}",
+                    LOGGER.info("{} is unexpectedly closed after receiving response: {}. " +
+                                    "Investigate logs on a proxy side to identify the cause.",
                             connection, response.toString((name, value) -> value));
                     return failed(StacklessClosedChannelException.newInstance(
                             ProxyConnectConnectionFactoryFilter.class, "handleConnectResponse: " +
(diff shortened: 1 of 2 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -86,7 +87,7 @@ public Single<C> newConnection(final ResolvedAddress resolvedAddress,
                 return delegate().newConnection(resolvedAddress, contextMap, observer).flatMap(c -> {
                     try {
                         return c.request(c.connect(connectAddress).addHeader(CONTENT_LENGTH, ZERO))
-                                .flatMap(response -> handleConnectResponse(c, response))
+                                .flatMap(response -> handleConnectResponse(c, response).shareContextOnSubscribe())
```
daschl, line 90: for my own education: is this related to the changes below and if so why?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -123,7 +124,14 @@ public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt
 
             final DeferSslHandler deferSslHandler = channel.pipeline().get(DeferSslHandler.class);
             if (deferSslHandler == null) {
-                return failed(new IllegalStateException("Failed to find a handler of type " +
+                if (!channel.isActive()) {
+                    LOGGER.info("{} is unexpectedly closed after receiving response: {}",
```
daschl, line 128: Is there a hint we can give to the user that this is proxy-related?

daschl (approved):
Looks good to me, just two comments.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
idelpivnitskiy: no, just observed as part of the main fix. The `flatMap` is like `defer`, will make a copy of a context when it subscribes to a returned single. Because `handleConnectResponse` is 100% internal method (users can not do anything in the middle of that method) it doesn't matter if we copy the context or share. Added `shareContextOnSubscribe()` to avoid unnecessary copying, just to reduce allocations.

PR #2545, opened 2023-03-30
MultiAddressUrlClient: incorrect transformation of target without path

idelpivnitskiy:
Motivation:
`java.net.URI` considers `http://localhost?param=value` a correct absolute address. However, `MultiAddressUrlClient` sends `/` instead of `/?param=value` to the server after transforming absolute to relative request target.

Modifications:

- Enhance `DefaultMultiAddressUrlHttpClientBuilder.absoluteToRelativeFormRequestTarget` to account for this use-case;
- Sync `RedirectSingle.absoluteToRelativeFormRequestTarget`;

Result:

`MultiAddressUrlClient` and redirect logic correctly transform `http://localhost?param=value` to a relative `/?param=value` request target.


daschl (approved):
LGTM

PR #2551, opened 2023-04-06
New filter: `JavaNetSoTimeoutHttpConnectionFilter`

idelpivnitskiy:
Motivation:

For backward compatibility with blocking Java libraries, it's helpful to have a filter that mimics behavior of `SO_TIMEOUT` socket option.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/JavaNetSoTimeoutHttpConnectionFilterTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/JavaNetSoTimeoutHttpConnectionFilterTest.java
@@ -1,0 +1,287 @@
+/*
+ * Copyright © 2023 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+import io.servicetalk.buffer.api.Buffer;
+import io.servicetalk.buffer.api.CharSequences;
+import io.servicetalk.buffer.api.CompositeBuffer;
+import io.servicetalk.concurrent.api.Executor;
+import io.servicetalk.concurrent.api.Publisher;
+import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.context.api.ContextMap.Key;
+import io.servicetalk.http.api.BlockingHttpClient;
+import io.servicetalk.http.api.BlockingStreamingHttpClient;
+import io.servicetalk.http.api.BlockingStreamingHttpRequest;
+import io.servicetalk.http.api.BlockingStreamingHttpResponse;
+import io.servicetalk.http.api.HttpRequest;
+import io.servicetalk.http.api.HttpResponse;
+import io.servicetalk.http.api.StreamingHttpResponse;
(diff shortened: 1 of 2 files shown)


servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/JavaNetSoTimeoutHttpConnectionFilter.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.concurrent.TimeoutException;
+import java.util.function.BiFunction;
+import javax.annotation.Nullable;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A filter to mimics {@link SocketOptions#SO_TIMEOUT} behavior on the client-side.
+ * <p>
+ * While {@link TimeoutHttpRequesterFilter} apples a timeout for the overall duration to receive either the response
```
daschl, line 53: apples / applies

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/JavaNetSoTimeoutHttpConnectionFilter.java
```diff
@@ -0,0 +1,213 @@
...
+ * To protect from the described use-cases, consider also using {@link TimeoutHttpRequesterFilter} before applying this
+ * filter.
+ * <p>
+ * This filter implements only {@link StreamingHttpConnectionFilterFactory} and therefore can be applied only at the
+ * connection level. This restriction ensures that the timeout is applied only for the response read operations
+ * (similar to {@link SocketOptions#SO_TIMEOUT} used by Java blocking API), without waiting for selecting or
+ * establishing a connection.
+ * <p>
+ * {@link SocketTimeoutException} (or its subtype) will be propagated when the timeout is reached.
+ *
```
daschl, line 68: It would be great if this javadoc could also elaborate more on the http 100 continue behavior, it took me a while to figure this bit out in the code and users could benefit from a (terse) explanation as well.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/JavaNetSoTimeoutHttpConnectionFilterTest.java
```diff
@@ -0,0 +1,287 @@
...
+import static java.nio.charset.StandardCharsets.US_ASCII;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.endsWith;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class JavaNetSoTimeoutHttpConnectionFilterTest {
```
daschl, line 68: final?

daschl (approved):
Looks good to me, just some comments about javadoc during polish?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/JavaNetSoTimeoutHttpConnectionFilterTest.java
idelpivnitskiy: IIRC, jUnit4 didn't allow final test classes. Tested that with jUnit5 we can define them as final, but since none of the existing test classes are final, let's keep it consistent

idelpivnitskiy: In production classes, we try to keep everything `final` (even private inner classes and package-private classes) for 2 reasons:
1. if someone ever makes the class public, at least it will be `final`;
2. some JVM impls can apply more optimizations to the generated bytecode when you explicitly tell that the class is final. Maybe modern compilers are smarter, but we follow this from netty for many years with the assumption that netty can be run in whatever env

daschl: 👍

PR #2554, opened 2023-04-18
Allow null values with FormUrlEncodedHttp[De]Serializer

daschl:
Motivation:

At the moment the behavior around null (and to an extent empty) values is a little bit unflexible. ServiceTalk makes the decision to ignore null values on encoding and converting them to the same value as an empty string on decoding.

Since there is no spec which says that null values are disallowed, it is better if ServiceTalk is flexible and lets the user decide if they want to ignore values (which they can do by checking and then just not encoding them in the first place).

Modifications:

This changeset modifies behavior as follows.

On the encoding side:
 - null values are now allowed and encoded as "just" the key, instead of not being encoded at all. (Distinguished without an "=" compared to an empty string)

On the decoding side:
 - Null values are now not turned into an empty string, but if just the key is present turned into "null" - and if the "=" is present, an empty string is decoded.

Note: this also implicitly changes the resulting HttpQuery behavior when calling lazyParseQueryString on HttpRequestMetaData, since it uses the same functionality underneath. Since the "get" method is marked as @Nullable already, there is no illegal behavior observed.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/FormUrlEncodedHttpSerializer.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/FormUrlEncodedHttpSerializer.java
@@ -27,6 +27,7 @@ import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
@@ -46,12 +47,17 @@ final class FormUrlEncodedHttpSerializer implements HttpSerializer<Map<String, L
     static final FormUrlEncodedHttpSerializer UTF8 = new FormUrlEncodedHttpSerializer(UTF_8,
             headers -> headers.set(CONTENT_TYPE, APPLICATION_X_WWW_FORM_URLENCODED_UTF_8));
 
+    private static final byte EQUALS_BYTE = '=';
+    private static final byte AMPERSAND_BYTE = '&';
+
     private final Charset charset;
+    private final boolean isOptimizedCharset;
     private final Consumer<HttpHeaders> addContentType;
 
     FormUrlEncodedHttpSerializer(final Charset charset, final Consumer<HttpHeaders> addContentType) {
         this.charset = charset;
         this.addContentType = addContentType;
+        this.isOptimizedCharset = charset == UTF_8 || charset == StandardCharsets.US_ASCII;
     }
 
     @SuppressWarnings("ConstantConditions")
@@ -185,8 +191,14 @@ final class FormUrlEncodedHttpSerializer implements HttpSerializer<Map<String, L
             values.forEach(value -> {
                 writeKey(buffer, isContinuation, key);
                 if (value != null) {
-                    buffer.writeBytes("=".getBytes(charset));
-                    buffer.writeBytes(urlEncode(value).getBytes(charset));
+                    if (isOptimizedCharset) {
+                        buffer.writeByte(EQUALS_BYTE);
+                    } else {
+                        buffer.writeBytes("=".getBytes(charset));
+                    }
+                    if (!value.isEmpty()) {
+                        buffer.writeBytes(urlEncode(value).getBytes(charset));
+                    }
                 }
             });
         });
@@ -195,7 +207,11 @@ final class FormUrlEncodedHttpSerializer implements HttpSerializer<Map<String, L
 
     private void writeKey(final Buffer buffer, boolean isContinuation, String key) {
         if (buffer.writerIndex() != 0 || isContinuation) {
-            buffer.writeBytes("&".getBytes(charset));
+            if (isOptimizedCharset) {
+                buffer.writeByte(AMPERSAND_BYTE);
+            } else {
+                buffer.writeBytes("&".getBytes(charset));
+            }
         }
         buffer.writeBytes(urlEncode(key).getBytes(charset));
     }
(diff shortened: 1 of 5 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/FormUrlEncodedHttpSerializer.java
```diff
@@ -168,34 +168,38 @@ private Buffer serialize(@Nullable final Map<String, List<String>> parameters, f
...
                 return;
             }
 
             values.forEach(value -> {
-                if (value == null) {
-                    return;
+                writeKey(buffer, isContinuation, key);
+                if (value != null) {
+                    buffer.writeBytes("=".getBytes(charset));
+                    buffer.writeBytes(urlEncode(value).getBytes(charset));
```
idelpivnitskiy, line 189: Should we protect this line by `!value.isEmpty()` check? Looks like we can write only `=` and go to the next value in this case.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/FormUrlEncodedHttpSerializer.java
```diff
@@ -168,34 +168,38 @@ private Buffer serialize(@Nullable final Map<String, List<String>> parameters, f
...
+                writeKey(buffer, isContinuation, key);
                 return;
             }
 
             values.forEach(value -> {
-                if (value == null) {
-                    return;
+                writeKey(buffer, isContinuation, key);
+                if (value != null) {
+                    buffer.writeBytes("=".getBytes(charset));
```
idelpivnitskiy, line 188: Since you are here, consider caching the value for `UFT-8` and `US_ASCII` encodings. For those we can use `buffer.writeByte(intValueOfTheByte)` and skip allocation of the array.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/FormUrlEncodedHttpSerializer.java
```diff
@@ -168,34 +168,38 @@ private Buffer serialize(@Nullable final Map<String, List<String>> parameters, f
...
-                buffer.writeBytes("=".getBytes(charset));
-                buffer.writeBytes(urlEncode(value).getBytes(charset));
             });
         });
         return buffer;
     }
 
+    private void writeKey(final Buffer buffer, boolean isContinuation, String key) {
+        if (buffer.writerIndex() != 0 || isContinuation) {
+            buffer.writeBytes("&".getBytes(charset));
```
idelpivnitskiy, line 198: Consider also caching `&` for well-known charsets

servicetalk-http-api/src/test/java/io/servicetalk/http/api/FormUrlEncodedHttpDeserializerTest.java
```diff
@@ -148,12 +137,29 @@ public boolean hasNext() {
...
+    @ValueSource(booleans = { true, false })
+    void deserializesNullOrEmptyValues(boolean paramIsNull) {
+        final String separator = paramIsNull ? "" : "=";
+        final String formParameters = String.format("key1%s&key2%s&key2%s", separator, separator, separator);
+
+        final Map<String, List<String>> deserialized = deserializer.deserialize(FE_HEADERS, toBuffer(formParameters));
+        assertThat(deserialized.size(), is(2));
+
+        assertThat(deserialized.get("key1").size(), is(1));
+        assertThat(deserialized.get("key2").size(), is(2));
```
idelpivnitskiy, line 156: Consider:
```
        assertThat(deserialized.get("key1"), hasSize(1));
        assertThat(deserialized.get("key2"), hasSize(2));
```
It generates better exception message in case the assertion fails.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/FormUrlEncodedHttpDeserializerTest.java
```diff
@@ -148,12 +137,29 @@ public boolean hasNext() {
...
+
+        assertThat(deserialized.get("key1").get(0), paramIsNull ? nullValue() : emptyString());
+        assertThat(deserialized.get("key2").get(0), paramIsNull ? nullValue() : emptyString());
+        assertThat(deserialized.get("key2").get(1), paramIsNull ? nullValue() : emptyString());
+    }
+
     private Buffer toBuffer(final String value) {
```
idelpivnitskiy, line 163: While you are here, please make it `static`

servicetalk-http-api/src/test/java/io/servicetalk/http/api/FormUrlEncodedHttpSerializerTest.java
```diff
@@ -237,6 +245,27 @@ public boolean hasNext() {
...
+
+        final Buffer serialized = SERIALIZER.serialize(headers, formParameters, DEFAULT_ALLOCATOR);
+        assertEquals("key1&key2&key3&key3", serialized.toString(UTF_8), "Unexpected serialized content.");
+    }
+
+    @Test
+    void serializesKeysWithEmptyValue() {
+        final Map<String, List<String>> formParameters = new HashMap<>();
+        formParameters.put("key1", Collections.singletonList(""));
+        formParameters.put("key2", Arrays.asList("", ""));
```
idelpivnitskiy, line 263: Consider testing a mix of `null`, empty, and non-empty string for the same key

idelpivnitskiy (approved):
Only minor comments, then 🚢 
Thanks!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/FormUrlEncodedHttpSerializer.java
```diff
@@ -46,12 +47,17 @@ final class FormUrlEncodedHttpSerializer implements HttpSerializer<Map<String, L
...
+    private static final byte AMPERSAND_BYTE = '&';
+
     private final Charset charset;
+    private final boolean isOptimizedCharset;
     private final Consumer<HttpHeaders> addContentType;
 
     FormUrlEncodedHttpSerializer(final Charset charset, final Consumer<HttpHeaders> addContentType) {
         this.charset = charset;
         this.addContentType = addContentType;
+        this.isOptimizedCharset = charset == UTF_8 || charset == StandardCharsets.US_ASCII;
```
idelpivnitskiy, line 60: Consider using `equals` because it's a class, not enum

idelpivnitskiy (approved):

PR #2555, opened 2023-04-21
Publisher#timeout always propagate TimeoutException

Scottmitch:
Motivation:
Publisher#timeout operator is inconsistent with Single and Completable timeout operators in that when a timeout triggers it will not ensure the timeout exception is propagated. It is possible for upstream to propagate a different exception even though the source of the error is a timeout exception.

Modifications:
- Enhance ConcurrentTerminalSubscriber to allow for external termination and take advantage of that in Publisher#timeout.
- Add tests for Completable and Single timeout operators to ensure existing expected behavior doesn't regress.


Scottmitch:
test failure attributed to https://github.com/apple/servicetalk/issues/2245

idelpivnitskiy (approved):
Thanks a lot!

PR #2556, opened 2023-04-21
ConcurrentTerminalSubscriber fix deferred terminal concurrency

Scottmitch:
Motivation:
ConcurrentTerminalSubscriber#unwrapMarkTerminated() isn't safe and may result in concurrent invocation on the downstream subscriber between onNext and the terminal method.

Modifications:
- Remove unwrapMarkTerminated()
- Add deferredOnError(Throwable) and deferredOnComplete() to "queue" a terminal signal that can later be delivered by onNext (if concurrently being invoked) or deliverDeferredTerminal().


idelpivnitskiy (commented):
Cool, thanks a lot for the follow-up!

Questions:
1. Should we use the new methods in `TakeUntilSubscriber`? It seems like we have a similar issue there as well.
2. `ConcurrentTerminalSubscriber` is used only in `concurrent-api` module. We should be able to deprecate it in `concurrent-internal` now and move to `concurrent-api` for the next major release. Should we?

PR #2557, opened 2023-04-21
DefaultDnsClient reduce log level for TTL > max

Scottmitch:
Motivation:
DefaultDnsClient logs at info level if the TTL from DNS is larger than the maximum configured TTL value. This level of logging can be noisy in production and not necessarily be within the client's control.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -814,7 +814,7 @@ final class DefaultDnsClient implements DnsClient {
                     ttlNanos = dnsAnswer.ttlNanos();
                     if (ttlNanos > maxTTLNanos) {
                         if (LOGGER.isDebugEnabled()) {
-                            LOGGER.debug("{} result for {} has TTL={}s > maxTTL={}s",
+                            LOGGER.debug("{} result for {} has TTL={} > maxTTL={}",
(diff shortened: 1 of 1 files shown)


colestanfield (approved):

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -813,9 +813,11 @@ private void handleResolveDone0(final Future<DnsAnswer<T>> addressFuture,
                     final List<T> addresses = dnsAnswer.answer();
                     ttlNanos = dnsAnswer.ttlNanos();
                     if (ttlNanos > maxTTLNanos) {
-                        LOGGER.info("{} result for {} has a high TTL={}s which is larger than configured maxTTL={}s.",
-                                DefaultDnsClient.this, AbstractDnsPublisher.this,
-                                NANOSECONDS.toSeconds(ttlNanos), NANOSECONDS.toSeconds(maxTTLNanos));
+                        if (LOGGER.isDebugEnabled()) {
+                            LOGGER.debug("{} result for {} has a high TTL={}s which is larger than configured maxTTL={}s.",
```
colestanfield: Ya know, putting `s` on the numeric value makes it harder to parse. I might recommend the unit be part of the field name.

Scottmitch: I'll just remove the units. TTL is always in seconds for DNS.

idelpivnitskiy (approved):

Scottmitch:
build failure caused by https://github.com/apple/servicetalk/issues/2117

PR #2563, opened 2023-04-26
Improve how we copy payload body in case of redirect

idelpivnitskiy:
Motivation:

- No need to always use `concat` when we carry message body from original to redirect request;
- We can access `DefaultPayloadInfo` to copy original request flags;

Modifications:

- Discard an empty publisher instead of using concatenation;
- Give pkg-private access to `DefaultPayloadInfo`;
- Copy original flags when available instead of using pessimistic approach that always sets "mayHaveTrailers";

Result:

More optimal copying of the message body, preserving original PayloadInfo flags.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/RedirectConfigBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/RedirectConfigBuilder.java
@@ -432,10 +432,18 @@ public final class RedirectConfigBuilder {
         private static void fullCopy(final StreamingHttpRequest originalRequest,
                                      final StreamingHttpRequest redirectRequest) {
             redirectRequest.setHeaders(originalRequest.headers());
-            redirectRequest.transformMessageBody(p -> p.ignoreElements().concat(originalRequest.messageBody()));
-            // Use `transform` to update PayloadInfo flags, assuming trailers may be included in the message body
-            redirectRequest.transform(NOOP_TRAILERS_TRANSFORMER);
-            // FIXME: instead of `transform`, preserve original PayloadInfo/FlushStrategy when it's API is available
+            preserveMessageBody(originalRequest, redirectRequest);
+            // Sync PayloadInfo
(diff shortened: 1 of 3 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/RedirectConfigBuilder.java
```diff
@@ -432,10 +432,18 @@ public StreamingHttpRequest apply(final boolean relative,
         private static void fullCopy(final StreamingHttpRequest originalRequest,
                                      final StreamingHttpRequest redirectRequest) {
             redirectRequest.setHeaders(originalRequest.headers());
-            redirectRequest.transformMessageBody(p -> p.ignoreElements().concat(originalRequest.messageBody()));
-            // Use `transform` to update PayloadInfo flags, assuming trailers may be included in the message body
-            redirectRequest.transform(NOOP_TRAILERS_TRANSFORMER);
-            // FIXME: instead of `transform`, preserve original PayloadInfo/FlushStrategy when it's API is available
+            preserveMessageBody(originalRequest, redirectRequest);
+            // Sync PayloadInfo
+            if (DefaultStreamingHttpRequest.class.equals(redirectRequest.getClass()) &&
```
daschl, line 437: Since DefaultStreamingHttpRequest is final, would instance of also work here? (I always wonder if instanceof is quicker or this class check)

daschl (approved):
Looks good to me, just one style question / nit.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/RedirectConfigBuilder.java
idelpivnitskiy: It will, but I'm not sure if it has any optimizations for `final` classes. I only remember that for non-final classes and interfaces it's slower. Even if it has, I believe it will optimize it to `class.equals`.

PR #2565, opened 2023-04-27
`Publisher.takeUntil(Completable)`: preserve original terminal signal

idelpivnitskiy:
Motivation:

For `takeUntil` operator, when either `Publisher` or `Completable` terminates, downstream `Subscriber` should see the original terminal signal, even if the other source terminates concurrently.

Modifications:

- Use recently added "deferred" methods of `ConcurrentTerminalSubscriber`;
- Add test to reproduce the original issue;
- Refactor existing tests to use `ParameterizedTest`;

Result:

`takeUntil` operators propagates the initial terminal signal.


Scottmitch (approved):

Scottmitch:
🚀 good find!

bryce-anderson (approved):
Good catch.

PR #2566, opened 2023-04-27
Deprecated some internal classes in `concurrent-internal`

idelpivnitskiy:
Motivation:

`concurrent-internal` has classes that either unused at all or can be moved as a pkg-private class to `concurrent-api` module.

Modifications:

- Deprecate `ArrayUtils`;
- Deprecate `ConcurrentTerminalSubscriber`;
- Deprecate `QueueFullAndRejectedSubscribeException`;

Result:

Less public classes in the next major release.


Scottmitch:
can you update TimeoutDemandPublishser then lgtm

PR #2567, opened 2023-04-27
`ConcurrentTerminalSubscriber`: consolidate code for `process*` methods

idelpivnitskiy:
Motivation:

`processOnComplete` and `processOnError` methods are very similar. We can consolidate their code in one private method.


daschl (approved):
Looks like no behavioral change, just refactoring (did not see any issues while doing that).

PR #2568, opened 2023-04-27
`RetryingHttpRequesterFilter`: share context when drain response payload

idelpivnitskiy:
Motivation:

`RetryingHttpRequesterFilter` uses an async `flatMap` operator to drain response payload body, but it doesn't share the context. It will result in `AsyncContext` propagation if some other user filters manipulate with context during payload body read.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
@@ -232,10 +232,12 @@ public final class RetryingHttpRequesterFilter
             if (responseMapper != null) {
                 single = single.flatMap(resp -> {
                     final HttpResponseException exception = responseMapper.apply(resp);
-                    return exception != null ?
+                    return (exception != null ?
                             // Drain response payload body before discarding it:
-                            resp.payloadBody().ignoreElements().onErrorComplete().concat(Single.failed(exception)) :
-                            Single.succeeded(resp);
+                            resp.payloadBody().ignoreElements().onErrorComplete()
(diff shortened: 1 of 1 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -232,10 +232,12 @@ protected Single<StreamingHttpResponse> request(final StreamingHttpRequester del
...
                     final HttpResponseException exception = responseMapper.apply(resp);
-                    return exception != null ?
+                    return (exception != null ?
                             // Drain response payload body before discarding it:
-                            resp.payloadBody().ignoreElements().onErrorComplete().concat(Single.failed(exception)) :
-                            Single.succeeded(resp);
+                            resp.payloadBody().ignoreElements().onErrorComplete()
+                                    .concat(Single.<StreamingHttpResponse>failed(exception)) :
+                            Single.succeeded(resp))
+                            .shareContextOnSubscribe();
```
daschl, line 240: would it make a difference if the shareContextOnSubscribe is inside our side of the flatMap?

daschl (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
idelpivnitskiy: I'm not sure I understood the question, feel free to continue if I misunderstood.
Because the function returns an async source, `flatMap` will subscribe to it. `shareContextOnSubscribe()` should be the last operator in the chain so that subscribe happens directly to `shareContextOnSubscribe()` to have any effect.

PR #2570, opened 2023-04-28
Add SslConfig#maxCertificateListBytes()

Scottmitch:
Motivation:
Some use cases have certificate chains larger than the BoringSSL default limit. This may result in handshake failures and there is no way to configure the maximum value.

Modifications:
- Add SslConfig#maxCertificateListBytes() and corresponding builder methods to allow configuring the limit (if supported by SSLEngine).

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateMaxSizeTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateMaxSizeTest.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright © 2023 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+import io.servicetalk.http.api.BlockingHttpClient;
+import io.servicetalk.http.api.HttpServerBuilder;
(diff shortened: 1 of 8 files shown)


Scottmitch:
PR build expected to fail currently. lower limit isn't failing handshake as expected, more investigation required.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SslContextFactory.java
```diff
@@ -256,6 +226,38 @@ private static void configureCertificateCompression(SslConfig config, SslContext
...
     }
 
+    private static void configureNettyOptions(SslConfig config, SslContextBuilder builder) {
+        final int maxCertificateListBytes = config.maxCertificateListBytes();
+        if (maxCertificateListBytes > 0) {
+            builder.option(MAX_CERTIFICATE_LIST_BYTES, maxCertificateListBytes);
+        }
+    }
+
+    private static SslContext configureBuilder(SslConfig config, SslContextBuilder builder, boolean forServer) {
```
bryce-anderson, line 236: Nice cleanup. 👍

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SslCertificateMaxSizeTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+    // JDK9+ can use InputStream.readAllBytes()
+    private static byte[] readAllBytes(InputStream is) throws IOException {
+        try {
+            final int readSize = 2048;
+            ByteArrayOutputStream buf = new ByteArrayOutputStream(readSize);
+            byte[] bytes = new byte[readSize];
+            while (is.read(bytes) > 0) {
+                buf.write(bytes);
+            }
```
bryce-anderson, line 123: You may not read the full buffer on each iteration.
```suggestion
            byte[] bytes = new byte[readSize];
            int length;
            while ((length = is.read(bytes)) != -1) {
                buf.write(bytes, 0, length);
            }
```
Since it seems that you're only using this to get the length of the data stream you could skip the `ByteArrayOutputStream` stuff and just sum up the non `-1` values you get from the read calls and return that.

bryce-anderson (approved):

idelpivnitskiy (commented):
Late LGTM, thanks!

Issue #2502, opened 2023-02-07
Streamline creation of GRPC Exceptions


daschl:
At the moment, a GRPC exception can be created as 

```java
// Without Description
throw new GrpcStatusException(GrpcStatusCode.PERMISSION_DENIED.status());

// With Description
throw new GrpcStatusException(new GrpcStatus(GrpcStatusCode.PERMISSION_DENIED, "my-description"));
```

Other GRPC implementations provide convenience methods like:

```java
throw PERMISSION_DENIED.withDescription("denied!").asException();
```

We could potentially add something similar to streamline exception generation (also including `withCause`, etc).

cmwigard:
I had a go at this issue as my first contribution to ServiceTalk, I hope you don't mind.
Let me know if there is anything you would like me to change or discuss.

PR #2571, opened 2023-05-01
Streamlined the creation of GRPC Exceptions by adding convenience method

cmwigard:
Motivation:

In reference to issue #2502, there should be a more convenient way to initialize GRPC Exceptions analogous to other GRPC implementations.

Modifications:

- Removed the deprecation from GrpcStatus.asException as it is required for this implementation of desired functionality.
- Added additional constructor for GrpcStatus to support initialization from codeValue and description.
- Added convenience method GrpcStatusCode.withDescription to support desired functionality.
- Added GrpcStatusTest.testGrpcStatusExceptionFromGrpcStatusCode for test-driven development and better test coverage.

Result:

GRPC Exceptions can now be initialized and thrown more conveniently.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatus.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatus.java
@@ -166,7 +166,6 @@ public final class GrpcStatus {
         return exception == null ? null : exception.status();
     }
 
-
     /**
      * Returns the current status wrapped in a {@link GrpcStatusException}.
      *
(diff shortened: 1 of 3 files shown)


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcStatus.java
```diff
@@ -156,10 +170,8 @@ public static GrpcStatus fromThrowableNullable(Throwable t) { // FIXME: 0.43 - r
      * Returns the current status wrapped in a {@link GrpcStatusException}.
      *
      * @return the current status wrapped in a {@link GrpcStatusException}.
-     * @deprecated Use {@link GrpcStatusException#GrpcStatusException(GrpcStatus)}.
      */
-    @Deprecated
-    public GrpcStatusException asException() { // FIXME: 0.43 - remove deprecated method
+    public GrpcStatusException asException() {
```
82baras, line 174: Should we updated deprecated code?

PR #2575, opened 2023-05-02
Convert IPv6 colons to dots for the new peer host

idelpivnitskiy:
Motivation:

#1958 adds IP-address to the peer host to increase probability for SSL cache hit. JDK can fallback to peer host for hostname verification if it can not find SNI hostname in the certs. When this happens with IPv6, users see
`IllegalArgumentException: Contains non-LDH ASCII characters`.

Modifications:

- Replace colons with dots when convert Inet6Address to String;

Result:

Users see expected `CertificateException`: No subject alternative DNS name matching <...> found.


idelpivnitskiy:
Didn't know how to test it with IPv6. If we have an idea, I will submit a follow-up. For now, verified with a custom URL and example project.

Scottmitch (approved):
Lgtm. Test would be nice but can be followup

PR #2576, opened 2023-05-03
Add Publisher#timeoutDemand operator

Scottmitch:
Motivation:
Some use cases involve long-poll style streaming where it is OK if the upstream doesn't deliver data as long as there is outstanding demand. The timeout operator will unconditionally timeout if there are no signals, independent of demand.

Modifications:
- Add Publisher#timeoutDemand operator that applies a timeout only when there is 0 outstanding demand to upstream.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutDemandPublisher.java
```diff
@@ -0,0 +1,160 @@
...
+                    // one actually fires.
+                    if (demandUpdater.compareAndSet(this, currDemand, currDemand)) {
+                        break;
+                    }
+                } else if (demandUpdater.compareAndSet(this, currDemand, DEMAND_TIMER_FIRED)) {
+                    try {
+                        stopTimer(true); // clear the reference and prevent future timers.
+                    } finally {
+                        // Concurrent/multiple termination is protected by ConcurrentTerminalSubscriber.
+                        offloadTimeout(new TimeoutException("timeout after " + NANOSECONDS.toMillis(parent.durationNs) +
```
idelpivnitskiy, line 126: Consider clarifying what exactly timed out, describing that there was no demand

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/TimeoutPublisherDemandTest.java
```diff
@@ -0,0 +1,296 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.sameInstance;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+final class TimeoutPublisherDemandTest {
```
idelpivnitskiy, line 63: Consider adding a test similar to `timeoutExceptionDeliveredBeforeUpstreamException()`, we need to make sure another `onError` after `cancel` is not propagated

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutPublisher.java
```diff
@@ -228,7 +291,7 @@ private void timerFires() {
                 final long nextTimeoutNs = parent.durationNs - (currentTimeNs - lastStartNS);
                 if (nextTimeoutNs <= 0) { // Timeout!
                     offloadTimeout(new TimeoutException("timeout after " + NANOSECONDS.toMillis(parent.durationNs) +
```
idelpivnitskiy, line 293: Users will appreciate if we use `restartAtOnNext` to enhance this exception message by clarifying if a timeout happened between `onNext` or for the terminal event

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutDemandPublisher.java
```diff
@@ -0,0 +1,160 @@
...
+        @Override
+        public void request(final long n) {
+            final Subscription subscription = this.subscription;
+            assert subscription != null;
+            if (n > 0 && demandUpdater.getAndAccumulate(this, n,
+                    FlowControlUtils::addWithOverflowProtectionIfNotNegative) == 0) {
+                stopTimer(false);
+            }
+            subscription.request(n);
+        }
```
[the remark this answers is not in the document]
Scottmitch: right, we need to set the state of `demandUpdater` before `request(1)` upstream and we should be good.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/publisher/TimeoutPublisherDemandTest.java
Scottmitch: sgtm I will share tests too

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TimeoutDemandPublisher.java
```diff
@@ -0,0 +1,160 @@
...
+                    try {
+                        stopTimer(true); // clear the reference and prevent future timers.
+                    } finally {
+                        // Concurrent/multiple termination is protected by ConcurrentTerminalSubscriber.
+                        offloadTimeout(new TimeoutException("timeout after " + NANOSECONDS.toMillis(parent.durationNs) +
+                                "ms"), parent.timeoutExecutor);
+                    }
+                    break;
+                }
+            }
```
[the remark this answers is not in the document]
Scottmitch: I will simplify. this logic was meant to decrease reliance on executor cancel preventing execution of the callback (e.g. incase its best effort, ..) but isn't necessary.

PR #2577, opened 2023-05-04
RRLB: reduce number of state transitions when health-check passed

idelpivnitskiy:
Motivation:

When health-checking opens a new connection,
`host.addConnection(newCnx)` only adds a connection to the list, but keeps the state as `UNHEALTHY`. The following `markHealthy` will flip the state. This makes it possible for a racy state transitions in between. We can enhance `addConnection` to also change the state to keep everything atomic.

Modifications:

- `addConnection` transitions `ACTIVE` and `UNHEALTHY` states to `STATE_ACTIVE_NO_FAILURES`;
- Update comment to clarify why `markHealthy` can `cancelIfHealthCheck`;

Result:

Unhealthy host transitions to `ACTIVE` state at the same time when it adds a new connection to the list.

### DIFF
--- a/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
+++ b/servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
@@ -592,9 +592,10 @@ final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalancedConnec
 
                         // Just in case the connection is not closed add it to the host so we don't lose track,
                         // duplicates will be filtered out.
-                        return host.addConnection(newCnx) ? failedSingle : newCnx.closeAsync().concat(failedSingle);
+                        return host.addConnection(newCnx, null) ?
+                                failedSingle : newCnx.closeAsync().concat(failedSingle);
                     }
-                    if (host.addConnection(newCnx)) {
+                    if (host.addConnection(newCnx, null)) {
                         return succeeded(newCnx);
                     }
                     return newCnx.closeAsync().concat(isClosedList(this.usedHosts) ? failedLBClosed(targetResource) :
@@ -809,7 +810,7 @@ final class RoundRobinLoadBalancer<ResolvedAddress, C extends LoadBalancedConnec
             return HealthCheck.class.equals(connState.state.getClass());
         }
 
-        boolean addConnection(C connection) {
+        boolean addConnection(final C connection, final @Nullable HealthCheck<Addr, C> currentHealthCheck) {
             int addAttempt = 0;
             for (;;) {
                 final ConnState previous = connStateUpdater.get(this);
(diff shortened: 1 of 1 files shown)


servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -888,7 +894,13 @@ currentConnState, new ConnState(newList, currentConnState.state))) {
...
                 LOGGER.trace("{}: removed connection {} from {} after {} attempt(s).",
                         lbDescription, connection, this, removeAttempt);
-            }).subscribe();
+            // Use onErrorComplete instead of whenOnError to avoid double logging of an error inside subscribe():
+            // SimpleCompletableSubscriber.
+            }).onErrorComplete(t -> {
+                        LOGGER.error("{}: unexpected error while processing connection.onClose() for {}.",
+                                lbDescription, connection, t);
+                        return true;
+                    }).subscribe();
```
Scottmitch, line 903: nit: fix indentation

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -829,7 +829,9 @@ boolean addConnection(C connection) {
                 Object[] newList = Arrays.copyOf(existing, existing.length + 1);
                 newList[existing.length] = connection;
 
-                Object newState = ActiveState.class.equals(previous.state.getClass()) ?
+                // If we were able to add a new connection to the list, we should mark the host as ACTIVE again and
+                // reset its failures counter.
+                Object newState = ActiveState.class.equals(previous.state.getClass()) || Host.isUnhealthy(previous) ?
```
Scottmitch, line 834: (general improvement suggestion) `ActiveState.class.equals..` -> consider adding a utility method for this too (it is done in a few spots). Also `isActiveAndHealthy` could call the new static method to share code too.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -743,7 +743,7 @@ void markExpired() {
         }
 
         void markHealthy(final HealthCheck<Addr, C> originalHealthCheckState) {
-            // Marking healthy is generally called from a successful health check, after a connection was added.
+            // Marking healthy is called when we need to recover from an unexpected error.
             // However, it is possible that in the meantime, the host entered an EXPIRED state, then ACTIVE, then failed
             // to open connections and entered the UNHEALTHY state before the original thread continues execution here.
             // In such case, the flipped state is not the same as the one that just succeeded to open a connection.
```
Scottmitch, line 749: (line 753) what if `previous.state == CLOSED`? Should we not set back to `ACTIVE`?

idelpivnitskiy: It flips only if `Host.isUnhealthy(previous)`

PR #2579, opened 2023-05-06
HttpObjectDecoder chunk size long -> int

Scottmitch:
Motivation:
HttpObjectDecoder parses chunk size as a long but in practice we truncate the value to an integer internally. We should parse the value as a int and raise an error early rather than discarding data.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
@@ -81,6 +81,7 @@ import static io.servicetalk.http.netty.HeaderUtils.removeTransferEncodingChunke
 import static io.servicetalk.http.netty.HttpKeepAlive.shouldClose;
 import static java.lang.Character.isISOControl;
 import static java.lang.Character.isWhitespace;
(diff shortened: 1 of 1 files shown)


bryce-anderson (approved):

idelpivnitskiy:
> HttpObjectDecoder parses chunk size as a long but in practice we truncate the value to an integer internally.

Can you point where we truncate? I see that we have an assertion in `READ_CHUNKED_CONTENT` to check that it's less than `Integer.MAX_VALUE`, and we may have int overflow when we cast it to int, but otherwise we don't truncate. If we remove an assert and fix overflow, we can read `long` chunks. I don't see why we will put a synthetic limit in ST.

The latest revision of [RFC9112](https://httpwg.org/specs/rfc9112.html#rfc.section.7.1) says:
> HTTP/1.1 does not define any means to limit the size of a chunked response such that an intermediary can be assured of buffering the entire response. Additionally, very large chunk sizes may cause overflows or loss of precision if their values are not represented accurately in a receiving implementation. Therefore, recipients MUST anticipate potentially large hexadecimal numerals and prevent parsing errors due to integer conversion overflows or precision loss due to integer representation.

Scottmitch:
https://github.com/apple/servicetalk/pull/2594

PR #2586, opened 2023-05-08
`Publisher.fromInputStream`: clarify `InputStream` ownership in javadoc

idelpivnitskiy:
Motivation:

Users are not sure if they have to close the `InputStream` or not after converting it into a `Publisher` or after passing into a request object.

Modifications:

- Clarify `InputStream` ownership for `Publisher.fromInputStream`;
- Enhance `FromInputStreamPublisherTest`;

Result:

Users have better understanding if they should close `InputStream` or not.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -4062,6 +4062,13 @@ public abstract class Publisher<T> {
      * Create a new {@link Publisher} that when subscribed will emit all data from the {@link InputStream} to the
      * {@link Subscriber} and then {@link Subscriber#onComplete()}.
      * <p>
+     * The resulting publisher is not replayable and supports only a single {@link Subscriber}.
+     * <p>
+     * After a returned {@link Publisher} is subscribed, it owns the passed {@link InputStream}, meaning that the
+     * {@link InputStream} will be automatically closed when the {@link Publisher} is cancelled or terminated. Not
+     * necessary to close the {@link InputStream} after subscribe, but it should be closed when control flow never
+     * subscribes to the returned {@link Publisher}.
+     * <p>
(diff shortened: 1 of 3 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -4062,6 +4062,13 @@ public static <T> Publisher<T> fromBlockingIterable(BlockingIterable<? extends T
      * Create a new {@link Publisher} that when subscribed will emit all data from the {@link InputStream} to the
      * {@link Subscriber} and then {@link Subscriber#onComplete()}.
      * <p>
+     * The resulting publisher is not replayable and supports only a single {@link Subscriber}.
```
daschl, line 4065: Great addition. Consider also clarifying what happens if the user happens to close the input stream as well.

idelpivnitskiy: `java.io.Closeable#close()` method is idempotent per its javadoc, nothing will happen. If you mean something different, please provide an example of how you would phrase it.

daschl: I was thinking of something like:

> If the `InputStream` is prematurely closed by the caller, no error will be raised.

I've been imagining a situation where the user might close the stream and no error is raised but the "payload" on the other side comes out shorter than expected.

idelpivnitskiy: This is an implementation detail of a particular `InputStream`. There are some that ignore close, like `ByteArrayInputStream`, some return `-1` after closing, but there are some that throw `IOException` if they see a `read` attempt after `close`, like `PipedInputStream`.

daschl (approved):

PR #2589, opened 2023-05-09
update jackson 2.14.2 -> 2.15.0

Scottmitch:


Scottmitch:
concerns about incompatibility require more investigation, closing for now

PR #2591, opened 2023-05-10
Publisher#buffer retry with sequential subscription

Scottmitch:
Motivation:
If the BufferStrategy terminates and the operator is applied to a Publisher that only supports sequential subscription model any retry operations will fail with duplicate Subscriber exception.

Modifications:
- Cancel before propagating error, so any retry operator will have state reset on the publisher.


idelpivnitskiy (approved):
🎖️

PR #2592, opened 2023-05-10
Logging LifeCycleObservers quote connection

Scottmitch:
Motivation:
LoggingGrpcLifecycleObserver and LoggingGrpcLifecycleObserver log the connection context, but the connection context string has spaces and special characters that may impact log indexers.

Modifications:
- quote places that log connection strings


idelpivnitskiy (commented):
Thanks!

PR #2594, opened 2023-05-11
HttpObjectDecoder chunk size int -> long

Scottmitch:
Motivation:
https://httpwg.org/specs/rfc9112.html#rfc.section.7.1
> HTTP/1.1 does not define any means to limit the size of a chunked response such that an intermediary can be assured of buffering the entire response. Additionally, very large chunk sizes may cause overflows or loss of precision if their values are not represented accurately in a receiving implementation. Therefore, recipients MUST anticipate potentially large hexadecimal numerals and prevent parsing errors due to integer conversion overflows or precision loss due to integer representation.


idelpivnitskiy (approved):
Thanks!

PR #2595, opened 2023-05-11
Publisher#validateOutstandingDemand operator

Scottmitch:
Motivation:
Some use cases use custom operators and sources that may invalidate demand
which can invalidate state in other operators and leave the system in
an undefined state. There is no easy way to validate this and get visibility.

Modifications:
- Add Publisher#validateOutstandingDemand operator

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -47,6 +47,7 @@ import java.util.function.IntFunction;
 import java.util.function.IntPredicate;
 import java.util.function.LongConsumer;
 import java.util.function.LongSupplier;
+import java.util.function.ObjLongConsumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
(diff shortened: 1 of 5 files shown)


tkountis (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/BiLongConsumer.java
```diff
@@ -0,0 +1,34 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * A special consumer that takes an {@code long} and a custom arguments.
+ *
+ * @param <T> The other argument to this function.
+ */
+@FunctionalInterface
+public interface BiLongConsumer<T> {
```
chrisvest, line 26: This is very similar to `java.util.function.ObjLongConsumer`. The argument order is just flipped.

Scottmitch: good catch! I will use this instead.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -1801,6 +1802,33 @@ public final Publisher<T> whenCancel(Runnable onCancel) {
         return beforeCancel(onCancel);
     }
 
+    /**
+     * Validate the outstanding demand ({@link Subscription#request(long)} minus
+     * {@link Subscriber#onNext(Object)}) does not go negative.
+     * @return The new {@link Publisher}.
```
idelpivnitskiy, line 1808: Consider clarifying what will happen when invalid demand is detected. Would be nice to also have some pseudo code that will demonstrate the flow, like we do for other operators.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -1801,6 +1802,33 @@ public final Publisher<T> whenCancel(Runnable onCancel) {
...
+    /**
+     * Validate the outstanding demand ({@link Subscription#request(long)} minus
+     * {@link Subscriber#onNext(Object)}) does not go negative.
+     * <p>
+     * ReactiveStreams specification allows {@link Subscriber} and {@link Subscription} to be concurrent, therefore the
+     * arguments of this function maybe invoked concurrently.
+     * @param onNextConsumer Called when {@link Subscriber#onNext(Object)} is invoked with first argument is the signal
+     * and second argument is demand (after decrementing).
+     * @param requestConsumer Called when {@link Subscription#request(long)} is invoked with first argument of {@code n}
+     * and second argument of current demand (after incrementing). This method may only be called if {@code n} is valid.
```
idelpivnitskiy, line 1824: Consider clarifying how these functions can be used. Are those only to observe or can they also throw an exception?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ValidateDemandPublisher.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        @Override
+        public void onComplete() {
+            subscriber.onComplete();
+        }
+    }
+
+    static final class InvalidDemandException extends IllegalStateException {
+        InvalidDemandException(String message) {
```
idelpivnitskiy, line 103: Our users found it hard to do anything with custom exceptions that are not public. In this case, they will see `InvalidDemandException` in logs, but won't have any clue what to do with it until they look at the source code and see that it's actually an `IllegalStateException`. Consider throwing `IllegalStateException` exception instead or make this one public.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
idelpivnitskiy: Will be also great to clarify a use-case for this operator.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ValidateDemandPublisher.java
```diff
@@ -0,0 +1,107 @@
...
+        public void onSubscribe(final Subscription subscription) {
+            subscriber.onSubscribe(new Subscription() {
+                @Override
+                public void request(final long n) {
+                    try {
+                        if (isRequestNValid(n)) {
+                            final long currDemand = demandUpdater.accumulateAndGet(ValidateDemandSubscriber.this, n,
+                                    FlowControlUtils::addWithOverflowProtection);
+                            parent.requestConsumer.accept(n, currDemand);
+                        }
```
idelpivnitskiy, line 68: The operator suppose to validate the demand. As a user, I would expect that any non-positive demand is also invalid and it have to be detected by this operator. Why we don't do anything for `else` case?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ValidateDemandPublisher.java
```diff
@@ -0,0 +1,107 @@
...
+                    } finally {
+                        subscription.request(n);
```
idelpivnitskiy, line 70: Why do we propagate demand even if the `requestConsumer` throws? I would expect we propagate `cancel`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ValidateDemandPublisher.java
Scottmitch: naming is hard ... this operator is meant to validate outstanding demand as delivered by async sources or operators, and not prevent invalid request(n) from propagating. Catching/failing preemptively would defeat the goal of this operator (meant to catch bugs in other operators and observe outstanding demand w/out influencing it). Any suggestions on names, here are some options:
- `trackDemand`
- `validateOutstandingDemand`

idelpivnitskiy: Thanks for clarifying. Agreed, naming is hard. With `trackDemand` users won't expect any exceptions to be propagated at all, only callbacks for observing it. Let's go with `validateOutstandingDemand` and clarifying in javadoc that invalid `request(n)` will be propagated as-is.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ValidateDemandPublisher.java
Scottmitch: `InvalidDemandException` helps validate the test condition but I will go with `IllegalStateException` for now if its an issue.

PR #2599, opened 2023-05-12
ThreadInterruptingCancellable consolidate state

Scottmitch:
Motivation:
ThreadInterruptingCancellable has integer internal state that can be removed by just using the Thread and atomically update it.

### DIFF
--- a/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/ThreadInterruptingCancellable.java
+++ b/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/ThreadInterruptingCancellable.java
@@ -17,7 +17,8 @@ package io.servicetalk.concurrent.internal;
 
 import io.servicetalk.concurrent.Cancellable;
 
-import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import javax.annotation.Nullable;
 
 import static java.lang.Thread.interrupted;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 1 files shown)


bryce-anderson (approved):
This seems correct but is, imo, more complicated. This isn't far from the same constructor pattern as an `AtomicInteger` which just sets the field on construction and expects the users to safely publish the `AtomicInteger` if it's going to be shared, most commonly by assigning the `AtomicInteger` to a `final` field. This is what is done with the usages of `ThreadInterruptingCancellable` that I've found.

If that's not enough and we expect unsafe storing of a `ThreadInterruptingCancellable` then maybe keeping a second field isn't such a big deal if it means we don't have to point people to the java memory model. That is, however, subjective.

idelpivnitskiy:
Agreed with @bryce-anderson. Pre-existing code is easier to read and understand, and does not require any clarification comments.

Scottmitch:
agreed this is more complex but the complexity is encapsulated and trade-off is memory vs complexity.

> This isn't far from the same constructor pattern as an AtomicInteger
AtomicInteger doesn't have any other options iiuc, however in this case bcz we have a terminal state we can be more defensive.

> This is what is done with the usages of ThreadInterruptingCancellable that I've found.
Currently `ThreadInterruptingCancellable` is used a local variables which afaik don't have the same guarantees of a final member variable (relative to access to the instance).

bryce-anderson:
> Currently ThreadInterruptingCancellable is used a local variables which afaik don't have the same guarantees of a final member variable (relative to access to the instance).

You're right: I misread the search result. It would still take some pretty adversarial code to be a problem but it's definitely possible.

> agreed this is more complex but the complexity is encapsulated and trade-off is memory vs complexity.

I'd personally not worry about an integer field in this context without some data to say it's a real contributor to allocations. Of course that is only my opinion, and not a very strong one, so feel free to ship it especially if these jump out in profiling data.

Scottmitch:
I will merge as complexity is encapsulated and we won't retain reference to the thread after done.

PR #2601, opened 2023-05-16
Deprecate `H2KeepAlivePolicies.disabled()`

idelpivnitskiy:
Motivation:

A special static constant to disable keep-alive is hard to discover. The internal state of `H2ProtocolConfigBuilder` converts `disabled()` to `null`. The `H2ProtocolConfig.keepAlivePolicy()` is nullable. Using `null` is expected.

Modifications:

- Deprecate `H2KeepAlivePolicies.disabled()`;
- Make `H2ProtocolConfigBuilder.keepAlivePolicy(...)` argument nullable;

Result:

Users can use `null` instead of `H2KeepAlivePolicies.disabled()`.


daschl (approved):
I find `H2KeepAlivePolicies.disabled()` better describes the intent than `null` (since other than looking at the javadoc, null could also mean use the default or something similar), but the PR itself looks fine to me.

idelpivnitskiy:
I didn't think about it that way. You have a good point. Agreed that we should not deprecate `disabled()` policy. Let me close this PR and if we need any adjustments for `H2ProtocolConfigBuilder.keepAlivePolicy(...)`, I will include those in #2602.

PR #2602, opened 2023-05-16
`KeepAlivePolicy`: validation, improved defaults & javadoc, ns precision

idelpivnitskiy:
Motivation:

1. Currently, if users configure a custom `KeepAlivePolicy` with
`idleDuration` less than 1 second (might look reasonable for testing),
they end up with behavior similar to disabled `KeepAlivePolicy`. This is
unexpected and hard to debug.
2. It's possible to configure negative durations, which may lead to
unexpected behavior.
3. `DEFAULT_ACK_TIMEOUT` is too high for a practical use-case. PING 
frames are acknowledged immediately by the protocol. 5 seconds is more
than enough for a round-trip TCP packet around the world.
4. Javadoc doesn't describe relationships between durations and how their
values affect the system.
5. Users need a way to configure ack timeout for the graceful closure process
without enabling keep-alive PING frames.

Modifications:
- Enhance `KeepAlivePolicy` and `H2KeepAlivePolicies` javadoc to clarify
its methods and arguments;
- `H2KeepAlivePolicies`: enhance validation for configured durations;
- Also validate in `H2ProtocolConfigBuilder.keepAlivePolicy(...)` in
case users pass a custom implementation of `KeepAlivePolicy`;
- `H2KeepAlivePolicies.DEFAULT_ACK_TIMEOUT`: default value changed
from 30 to 5 seconds;
- `H2KeepAlivePolicies.disabled()`: use `idleDuration == 0`, keep 
`DEFAULT_ACK_TIMEOUT`;
- `KeepAliveManager`: use `NANOSECONDS` for idle timeout;
- Make `H2ProtocolConfig.keepAlivePolicy()` non-nullable;
- `KeepAliveManager`: configure `pingAckTimeoutNanos` even when
`idleDuration` is zero (disabled);

Results:

1. Users see an exception when they misconfigure `KeepAlivePolicy`, smaller
`idleDuration` works as expected and it never rounded to a full second.
2. Improved default values, descriptive javadoc.
3. Users can configure `ackTimeout` even when idleness detection is disabled.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2KeepAlivePolicies.java
```diff
@@ -30,7 +34,7 @@ public final class H2KeepAlivePolicies {
     static final KeepAlivePolicy DISABLE_KEEP_ALIVE =
             new DefaultKeepAlivePolicy(ofDays(365), ofDays(365), false);
     static final Duration DEFAULT_IDLE_DURATION = ofSeconds(30);
-    static final Duration DEFAULT_ACK_TIMEOUT = ofSeconds(30);
+    static final Duration DEFAULT_ACK_TIMEOUT = ofSeconds(5);
```
daschl, line 37: Consider adding the default value as information to a javadoc block (config builder and/or config)

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2KeepAlivePolicies.java
```diff
@@ -135,7 +144,37 @@ public KeepAlivePolicyBuilder withoutActiveStreams(final boolean withoutActiveSt
...
+            Duration ackTimeout = this.ackTimeout;
+            if (ackTimeout == null) {
+                ackTimeout = min(idleDuration, DEFAULT_ACK_TIMEOUT);
+            }
+            validateRelativeDurations(idleDuration, ackTimeout);
             return new DefaultKeepAlivePolicy(idleDuration, ackTimeout, withoutActiveStreams);
         }
+
+        static Duration validateIdleDuration(final Duration idleDuration) {
+            if (idleDuration.getSeconds() < 1) {
```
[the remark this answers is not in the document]
idelpivnitskiy: If `idleDuration` is null, it will automatically generate NPE. Is there any benefit to explicitly add `requireNonNull`?

idelpivnitskiy:
TBH, I didn't like a magical rounding to a full second. Then I discovered that `IdleStateHandler` actually supports other time units. Let me do some more refactoring.

idelpivnitskiy:
@daschl did some re-work, updated PR description, wdyt?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -117,10 +116,15 @@ protected void channelIdle(final ChannelHandlerContext ctx, final IdleStateEvent
...
+        // Before 0.42.30, H2ProtocolConfig.keepAlivePolicy() was @Nullable. For backward compatibility, we keep
+        // tolerance for null values.
+        if (keepAlivePolicy != null) {  // FIXME: 0.43.x - consider removing null check
+            // KeepAlivePolicy with idlenessThresholdNanos <= 0 disables PINGs, but allows configuring
+            // pingAckTimeoutNanos for graceful closure (GO_AWAY).
             disallowKeepAliveWithoutActiveStreams = !keepAlivePolicy.withoutActiveStreams();
             pingAckTimeoutNanos = keepAlivePolicy.ackTimeout().toNanos();
-            pingWriteCompletionListener = future -> {
+            final long idlenessThresholdNanos = keepAlivePolicy.idleDuration().toNanos();
+            pingWriteCompletionListener = idlenessThresholdNanos > 0 ? future -> {
```
idelpivnitskiy, line 127: Didn't move `pingWriteCompletionListener` under a single `if` block to avoid indentation change. Wanted to keep git changes minimal to preserve the history.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2KeepAlivePolicies.java
daschl: 👍

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2KeepAlivePolicies.java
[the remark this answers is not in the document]
daschl: If no message is provided it's the same, yeah.

daschl (approved):

PR #2603, opened 2023-05-17
Add fragment methods to HttpRequestMetaData

mikehepple:
Motivation:

Without fragment support users are not able to build a full request target using only builder methods

Modifications:

- Adds setter and getter for fragment in HttpRequestMetaData implementations
- Adds tests to AbstractHttpRequestMetaDataTest

Result:

Users are now able to set a fragment for request targets

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultBlockingStreamingHttpRequest.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultBlockingStreamingHttpRequest.java
@@ -154,6 +154,18 @@ final class DefaultBlockingStreamingHttpRequest extends AbstractDelegatingHttpRe
         return this;
     }
 
+    @Nullable
+    @Override
+    public String fragment() {
+        return original.fragment();
+    }
+
+    @Override
+    public HttpRequestMetaData fragment(@Nullable String fragment) {
+        original.fragment(fragment);
+        return this;
+    }
+
     @Override
     public BlockingIterable<Buffer> payloadBody() {
         return original.payloadBody().toIterable();
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultBlockingStreamingHttpRequest.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultBlockingStreamingHttpRequest.java
@@ -154,14 +154,8 @@ final class DefaultBlockingStreamingHttpRequest extends AbstractDelegatingHttpRe
         return this;
     }
 
-    @Nullable
     @Override
-    public String fragment() {
-        return original.fragment();
-    }
-
-    @Override
-    public HttpRequestMetaData fragment(@Nullable String fragment) {
+    public BlockingStreamingHttpRequest fragment(@Nullable String fragment) {
         original.fragment(fragment);
         return this;
     }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequest.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequest.java
@@ -145,6 +145,18 @@ final class DefaultHttpRequest extends AbstractDelegatingHttpRequest
         return this;
     }
 
+    @Nullable
+    @Override
+    public String fragment() {
+        return original.fragment();
+    }
+
+    @Override
+    public HttpRequest fragment(@Nullable String fragment) {
+        original.fragment(fragment);
+        return this;
+    }
+
     @Override
     public HttpRequest addHeader(final CharSequence name, final CharSequence value) {
         original.addHeader(name, value);
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequest.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequest.java
@@ -145,12 +145,6 @@ final class DefaultHttpRequest extends AbstractDelegatingHttpRequest
         return this;
     }
 
-    @Nullable
-    @Override
-    public String fragment() {
-        return original.fragment();
-    }
-
     @Override
     public HttpRequest fragment(@Nullable String fragment) {
         original.fragment(fragment);
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequestMetaData.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequestMetaData.java
@@ -25,6 +25,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import javax.annotation.Nullable;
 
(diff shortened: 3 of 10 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultBlockingStreamingHttpRequest.java
```diff
@@ -154,6 +154,18 @@ public BlockingStreamingHttpRequest setQueryParameters(String key, String... val
         return this;
     }
 
+    @Nullable
+    @Override
+    public String fragment() {
```
idelpivnitskiy, line 159: There is no need to add override for "getter" in all classes that implement `HttpRequestMetaData`. It's enough to have it only in:
1. AbstractDelegatingHttpRequest
2. DefaultHttpRequestMetaData

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultBlockingStreamingHttpRequest.java
```diff
@@ -154,6 +154,18 @@ public BlockingStreamingHttpRequest setQueryParameters(String key, String... val
...
+        return original.fragment();
+    }
+
+    @Override
+    public HttpRequestMetaData fragment(@Nullable String fragment) {
```
idelpivnitskiy, line 164: But "setter" we need to override in all following classes because we have to change the return type:
1. HttpRequest
2. StreamingHttpRequest
3. BlockingStreamingHttpRequest
4. DefaultHttpRequestMetaData
5. DefaultHttpRequest
6. DefaultStreamingHttpRequest
7. DefaultBlockingStreamingHttpRequest

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequestMetaData.java
```diff
@@ -433,6 +433,33 @@ private void query(final Map<String, List<String>> params) {
...
+    public HttpRequestMetaData fragment(@Nullable String fragment) {
+
+      if (fragment == null && fragment() == null) {
+        return this;
+      }
+
+      String originalRequestTarget = requestTarget();
+
+      if (lazyParseRequestTarget().fragment() != null) {
+        int fragmentLength = lazyParseRequestTarget().fragment().length() + 1;
```
idelpivnitskiy, line 452: The short equivalent of `lazyParseRequestTarget().fragment()` will be `fragment()`. Looks like we need it for all code paths, consider caching in a local variable at the top of the method.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequestMetaData.java
```diff
@@ -433,6 +433,33 @@ private void query(final Map<String, List<String>> params) {
...
+    @Nullable
+    @Override
+    public String fragment() {
+      return lazyParseRequestTarget().fragment();
+    }
+
+    @Override
+    public HttpRequestMetaData fragment(@Nullable String fragment) {
+
+      if (fragment == null && fragment() == null) {
```
idelpivnitskiy, line 445: Consider also returning immediately if `fragment.equals(currFragment)`

servicetalk-http-api/src/test/java/io/servicetalk/http/api/AbstractHttpRequestMetaDataTest.java
```diff
@@ -801,6 +801,52 @@ void testEncodeToRequestTargetWithAbsoluteFormWithUserInfo() {
...
+        createFixture("/some/path?query=value#fragment");
+        fixture.fragment("overwrite");
+        assertEquals("overwrite", fixture.fragment());
+        assertEquals("/some/path?query=value#overwrite", fixture.requestTarget());
+    }
+
+    @Test
+    void testClearFragment() {
+        createFixture("/some/path?query=value#fragment");
+        fixture.fragment(null);
```
idelpivnitskiy, line 845: Thanks for the good test coverage! Consider 3 more tests:
1. When requestTarget has an empty fragment: `/some/path#`, `assertThat(fixture.fragment(), is(emptyString()))`
2. When users set an empty string as a fragment
3. When users set the same fragment

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequestMetaData.java
```diff
@@ -433,6 +433,33 @@ private void query(final Map<String, List<String>> params) {
...
+      String originalRequestTarget = requestTarget();
+
+      if (lazyParseRequestTarget().fragment() != null) {
+        int fragmentLength = lazyParseRequestTarget().fragment().length() + 1;
+        // remove the existing fragment
+        originalRequestTarget = originalRequestTarget.substring(0, originalRequestTarget.length() - fragmentLength);
+      }
+
+      return requestTarget(fragment == null || fragment.isEmpty()
+              ? originalRequestTarget
```
idelpivnitskiy, line 458: I tested how the `lazyParseRequestTarget()` works, it actually results in the following:
1. `/path` -> fragment() returns `null`
2. `/path#` -> fragment() returns an empty string
3. `/path#foo` -> fragment returns `foo`

Will be great if the setter works in the opposite way: `fragment("")` results in `/path#`, etc.

idelpivnitskiy (changes requested):
Thank you for the PR @mikehepple!

mikehepple:
@idelpivnitskiy suggested changes made. Submitted as a 2nd commit but can squash if you prefer.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequest.java
```diff
@@ -145,6 +145,18 @@ public HttpRequest setQueryParameters(final String key, final String... values)
         return this;
     }
 
+    @Nullable
+    @Override
+    public String fragment() {
```
idelpivnitskiy, line 150: Reminder to remove this override from all classes, except `AbstractDelegatingHttpRequest` and `DefaultHttpRequestMetaData`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpRequestMetaData.java
```diff
@@ -406,6 +406,27 @@ default boolean hasQueryParameter(final String key) {
      */
     boolean removeQueryParameters(String key, String value);
 
+    /**
+     * Returns the <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-3.5">fragment</a> part of the request target.
+     *
+     * @return the fragment part of the request target.
+     */
+    @Nullable
+    default String fragment() {
```
idelpivnitskiy, line 415: Forgot to ask, please add the following comment here:
```
// FIXME: 0.43 - consider removing default implementations
```
It's a reminder for us when it's expected to have some API changes.

idelpivnitskiy (commented):
One override to remove, then LGTM! Thanks a lot. The 2nd commit is the best to see a diff. We squash via UI when it's time to merge, no worries about it.

idelpivnitskiy:
Running CI now...

PR #2604, opened 2023-05-17
Test HTTP/2 `KeepAliveManager` with `DuplexChannel`

idelpivnitskiy:
Motivation:

`KeepAliveManagerTest` covers only non-duplex `Channel` use-case. However, `KeepAliveManager` state machine is a lot more complex for `DuplexChannel`. We need to enhance test coverage for the most used case.

Modifications:
- Parametrize `KeepAliveManagerTest` to test with duplex and non-duplex `Channel`, add more assertions;
- Modify `H2ParentConnectionContext` and `KeepAliveManager` to accept `Http2StreamChannel` instead of `Channel`;
- `KeepAliveTest`: fix `clientBuilder.initializeHttp` override issue (`CLIENT_CTX` is not set), do some other minor refactoring;

Result:

Enhanced test coverage for `KeepAliveManager`.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/KeepAliveManagerTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/KeepAliveManagerTest.java
@@ -16,24 +16,36 @@
 package io.servicetalk.http.netty;
 
 import io.servicetalk.http.netty.H2ProtocolConfig.KeepAlivePolicy;
+import io.servicetalk.transport.netty.internal.EmbeddedDuplexChannel;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.channel.socket.ChannelInputShutdownReadComplete;
+import io.netty.channel.socket.ChannelOutputShutdownEvent;
 import io.netty.handler.codec.http2.DefaultHttp2PingFrame;
 import io.netty.handler.codec.http2.Http2Error;
 import io.netty.handler.codec.http2.Http2GoAwayFrame;
 import io.netty.handler.codec.http2.Http2PingFrame;
+import io.netty.handler.codec.http2.Http2StreamChannel;
+import io.netty.handler.ssl.SslCloseCompletionEvent;
 import io.netty.util.concurrent.Promise;
 import org.hamcrest.Matcher;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadLocalRandom;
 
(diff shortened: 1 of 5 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/KeepAliveManagerTest.java
```diff
@@ -47,41 +59,54 @@
 
 class KeepAliveManagerTest {
 
+    private final BlockingQueue<ScheduledTask> scheduledTasks = new LinkedBlockingQueue<>();
```
daschl, line 62: I know it wasn't done in the original code / this refactoring, but would it make sense in the BeforeEach method (or setUp) to clear this queue to have every test start with a deterministic fresh state?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/KeepAliveManagerTest.java
```diff
@@ -338,4 +396,33 @@ private static final class ScheduledTask {
...
+    private static final class KeepAliveManagerHandler extends ChannelInboundHandlerAdapter {
+        private KeepAliveManager keepAliveManager;
+
+        void keepAliveManager(KeepAliveManager keepAliveManager) {
+            this.keepAliveManager = requireNonNull(keepAliveManager);
+        }
+
+        @Override
+        public void channelInactive(final ChannelHandlerContext ctx) {
+            keepAliveManager.channelClosed();
```
daschl, line 409: IIRC handlerRemoved and channelInactive can be both called one after another in the pipeline, is it necessary to add an "already closed" check in `KeepAliveManager#channelClosed`? The implementation of the method looks fine to me side-effect wise, just double checking with you if it's ok.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/KeepAliveManagerTest.java
```diff
@@ -338,4 +396,33 @@ private static final class ScheduledTask {
...
+        }
+
+        @Override
+        public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) throws Exception {
+            if (evt == ChannelInputShutdownReadComplete.INSTANCE || evt == SslCloseCompletionEvent.SUCCESS) {
+                keepAliveManager.channelInputShutdown();
+            } else if (evt == ChannelOutputShutdownEvent.INSTANCE) {
+                keepAliveManager.channelOutputShutdown();
+            } else {
+                release(evt);
```
daschl, line 424: Should any other user-triggered events cause failure? (i.e. they are unexpected?)

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/KeepAliveManagerTest.java
idelpivnitskiy: Thanks for calling it out. It looks idempotent to me. It only overrides the states and cancels if there are any futures.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/KeepAliveManagerTest.java
idelpivnitskiy: I don't think so. My understanding is that user events are fired as best effort to notify a pipeline about something. Other handlers are free to ignore those events if they do not depend on them. 

This handler is a simplification of what we have in `H2ParentConnectionContext.AbstractH2ParentConnection`.

daschl (approved):

PR #2605, opened 2023-05-17
`KeepAliveManager`: improve debuggability

idelpivnitskiy:
Motivation:

`KeepAliveManager` has very limited logging that doesn't describe what is going on during graceful closure process. Also, it's internal state is not user-friendly when analyzing heap dumps.

Modifications:
- Add more logging, include the parent `channel` in every log statement with extra information from internal fields when necessary;
- Use enum for internal state instead of `Object` constants;
- Rename `activeChildChannels` -> `activeStreams` to align name with public API;

Result:

More debug-level visibility, easy to read internal state for heap dump analysis.


daschl (approved):
Good refactoring!

PR #2606, opened 2023-05-20
WriteStreamSubscriberTest verify writability change

Scottmitch:
Motivation:
WriteStreamSubscriberTest should verify assumptions about how Channel#byteBeforeUnwritable works relative to writability change notifications as otherwise we may stop requesting data and hang.


Scottmitch:
Expected to fail until https://github.com/netty/netty/pull/13389 is merged/released.

tkountis (approved):

PR #2608, opened 2023-05-22
Handle graceful closure for indefinite streams

idelpivnitskiy:
Motivation:

Indefinitely running streams do not allow graceful closure to finish. Users have to apply a timeout for graceful closure and force close the channel. We need to verify that this scenario correctly closes the connection on both sides.

Modifications:
- Enhance `GracefulConnectionClosureHandlingTest` to validate the described use-case.
- Adjust `RequestResponseCloseHandler` to close the channel when input is shutdown before it can read the full incoming message. This reverts #53. Affects only HTTP/1.x.
- Adjust `RequestResponseCloseHandlerTest` for new behavior.
- `KeepAliveManager`: close the channel if write fails.

Results:

1. Connection is correctly closed on both ends when a closure is forced after incomplete graceful closure.
2. `KeepAliveManager` doesn't leave a connection in a hanging state if one of its write operations fails.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
@@ -130,7 +130,8 @@ final class KeepAliveManager {
             pingWriteCompletionListener = idlenessThresholdNanos > 0 ? future -> {
                 assert channel.eventLoop().inEventLoop();
                 if (!future.isSuccess()) {
-                    LOGGER.debug("{} Failed writing a PING frame after idleness is detected", channel, future.cause());
+                    LOGGER.debug("{} Failed to write a PING frame after idleness is detected, closing the channel",
+                            channel, future.cause());
                     close0(future.cause());
                 } else if (keepAliveState == State.KEEP_ALIVE_ACK_PENDING) {
                     // Schedule a task to verify ping ack within the pingAckTimeoutMillis
@@ -144,8 +145,8 @@ final class KeepAliveManager {
                             channel.writeAndFlush(new DefaultHttp2GoAwayFrame(NO_ERROR))
                                     .addListener(f -> {
                                         if (!f.isSuccess()) {
-                                            LOGGER.debug("{} Failed writing the last GO_AWAY after PING(ACK) timeout",
-                                                    channel, f.cause());
+                                            LOGGER.debug("{} Failed to write the last GO_AWAY after PING(ACK) " +
+                                                            "timeout, closing the channel", channel, f.cause());
                                         }
                                         close0(f.cause());
                                     });
@@ -324,7 +325,8 @@ final class KeepAliveManager {
         channel.writeAndFlush(new DefaultHttp2PingFrame(GRACEFUL_CLOSE_PING_CONTENT)).addListener(future -> {
             assert channel.eventLoop().inEventLoop();
             if (!future.isSuccess()) {
-                LOGGER.debug("{} Failed writing the first GO_AWAY and PING frames", channel, future.cause());
+                LOGGER.debug("{} Failed to write the first GO_AWAY and PING frames, closing the channel",
+                        channel, future.cause());
                 close0(future.cause());
             } else if (gracefulCloseState == State.GRACEFUL_CLOSE_START) {
                 // If gracefulCloseState is not GRACEFUL_CLOSE_START that means we have already received the PING(ACK)
@@ -353,7 +355,7 @@ final class KeepAliveManager {
 
         channel.writeAndFlush(new DefaultHttp2GoAwayFrame(NO_ERROR)).addListener(future -> {
             if (!future.isSuccess()) {
-                LOGGER.debug("{} Failed writing the second GO_AWAY", channel, future.cause());
+                LOGGER.debug("{} Failed to write the second GO_AWAY, closing the channel", channel, future.cause());
                 close0(future.cause());
             } else if (activeStreams == 0) {
                 close0(null);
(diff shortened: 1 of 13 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -129,7 +129,10 @@ protected void channelIdle(final ChannelHandlerContext ctx, final IdleStateEvent
             final long idlenessThresholdNanos = keepAlivePolicy.idleDuration().toNanos();
             pingWriteCompletionListener = idlenessThresholdNanos > 0 ? future -> {
                 assert channel.eventLoop().inEventLoop();
-                if (future.isSuccess() && keepAliveState == State.KEEP_ALIVE_ACK_PENDING) {
+                if (!future.isSuccess()) {
+                    LOGGER.debug("{} Failed writing a PING frame after idleness is detected", channel, future.cause());
```
daschl, line 133: Consider mentioning in the log that the cannel is now going to be proactively closed.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -317,9 +323,12 @@ private void doCloseAsyncGracefully0(final Runnable whenInitiated) {
         channel.write(goAwayFrame);
         channel.writeAndFlush(new DefaultHttp2PingFrame(GRACEFUL_CLOSE_PING_CONTENT)).addListener(future -> {
             assert channel.eventLoop().inEventLoop();
-            // If gracefulCloseState is not GRACEFUL_CLOSE_START that means we have already received the PING(ACK) and
-            // there is no need to apply the timeout.
-            if (future.isSuccess() && gracefulCloseState == State.GRACEFUL_CLOSE_START) {
+            if (!future.isSuccess()) {
+                LOGGER.debug("{} Failed writing the first GO_AWAY and PING frames", channel, future.cause());
+                close0(future.cause());
```
daschl, line 328: Consider mentioning in the log that the cannel is now going to be proactively closed.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -343,13 +352,16 @@ private void gracefulCloseWriteSecondGoAway() {
         gracefulCloseState = State.GRACEFUL_CLOSE_SECOND_GO_AWAY_SENT;
 
         channel.writeAndFlush(new DefaultHttp2GoAwayFrame(NO_ERROR)).addListener(future -> {
-            if (activeStreams == 0) {
-                close0();
+            if (!future.isSuccess()) {
+                LOGGER.debug("{} Failed writing the second GO_AWAY", channel, future.cause());
+                close0(future.cause());
```
daschl, line 357: Consider mentioning in the log that the cannel is now going to be proactively closed.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/GracefulConnectionClosureHandlingTest.java
```diff
@@ -146,26 +158,28 @@ class GracefulConnectionClosureHandlingTest {
...
-                    SERVER_CTX.ioExecutor().isUnixDomainSocketSupported(),
+            assumeTrue(SERVER_CTX.ioExecutor().isUnixDomainSocketSupported(),
                     "Server's IoExecutor does not support UnixDomainSocket");
-            Assumptions.assumeTrue(
-                    CLIENT_CTX.ioExecutor().isUnixDomainSocketSupported(),
+            assumeTrue(CLIENT_CTX.ioExecutor().isUnixDomainSocketSupported(),
                     "Client's IoExecutor does not support UnixDomainSocket");
             assumeFalse(viaProxy, "UDS cannot be used via proxy");
         }
         assumeFalse(protocol == HTTP_2 && viaProxy, "Proxy is not supported with HTTP/2");
```
daschl, line 178: nit: could simplify to

```suggestion
if (viaProxy) {
  assumeFalse(!secure, "Proxy tunnel works only with secure connections");
  assumeFalse(protocol == HTTP_2, "Proxy is not supported with HTTP/2");
}

```

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/GracefulConnectionClosureHandlingTest.java
```diff
@@ -241,26 +268,57 @@ public Completable accept(final ConnectionContext context) {
...
                         }
-                        data.add(Arguments.of(protocol, initiateClosureFromClient, useUds, viaProxy));
                     }
                 }
             }
         }
         return data;
     }
 
+    private static Collection<Arguments> dataWithKeepAlive() {
```
daschl, line 307: consider adding a docblock here, the method is quite dense with array manipulation - just briefly explaining what it does to the arguments?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -876,7 +876,9 @@ public void handlerRemoved(ChannelHandlerContext ctx) {
 
         @Override
         public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
-            connection.nettyChannelPublisher.channelOnError(unwrapThrowable(cause));
+            final Throwable throwable = unwrapThrowable(cause);
+            tryFailSubscriber(throwable);
```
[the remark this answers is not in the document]
idelpivnitskiy: Actually, let me revert this change. It's not required for new tests in this PR and I think it's not required at all because the subscriber will be notified inside either channelActive or SSL handshake completion. If an exception is propagated before any of these 2 events, the channel will be closed anyway.
I would say we need a reproducer before we add this.

daschl (approved):
Looks good to me, but as talked about offline @Scottmitch should also look over the state machine parts.

Scottmitch (approved):

mgodave (approved):
This looks to be an issue we are currently seeing as well.

tkountis (approved):
most approvals ever

PR #2610, opened 2023-05-25
Enable more `GracefulConnectionClosureHandlingTest`s after netty upgrade

idelpivnitskiy:
Motivation:

Netty 4.1.93 was released with the required fix, we can enable for tests for HTTP/2 graceful closure now.


idelpivnitskiy:
@daschl I will merge bcz it's just a test, you already reviewed it earlier. Just FYI that it works as expected with newer Netty version

PR #2611, opened 2023-05-27
`KeepAliveManager`: force close the channel in case of any issues

idelpivnitskiy:
Motivation:

1. If we failed to write a `PING` or `GO_AWAY` frame, the channel is already corrupted and we should not expect a gentle FIN exchange, simply close the channel.
2. If we shutdown output, wait for input shutdown only for a reasonable time frame to prevent a resource leak because network can lose a FIN back.

Modifications:
- Close the channel immediately if `close0` is invoked with an exception;
- Do `SslHandler.closeOutbound()` only for a duplex channel;
- Schedule a timer for `pingAckTimeoutNanos` to wait for input shutdown and force close if it's not received;

Result:

HTTP/2 connections don't hang indefinitely if the remote peer does not reply with a FIN back or if FIN was lost.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
@@ -98,6 +98,8 @@ final class KeepAliveManager {
     @Nullable
     private Object keepAliveState;
     @Nullable
+    private Future<?> inputShutdownFuture;
+    @Nullable
     private final GenericFutureListener<Future<? super Void>> pingWriteCompletionListener;
 
     KeepAliveManager(final Channel channel, @Nullable final KeepAlivePolicy keepAlivePolicy) {
@@ -206,8 +208,10 @@ final class KeepAliveManager {
 
         cancelIfStateIsAFuture(gracefulCloseState);
         cancelIfStateIsAFuture(keepAliveState);
+        cancelIfStateIsAFuture(inputShutdownFuture);
         gracefulCloseState = State.CLOSED;
         keepAliveState = State.CLOSED;
+        inputShutdownFuture = null;
     }
 
     void initiateGracefulClose(final Runnable whenInitiated) {
@@ -240,6 +244,8 @@ final class KeepAliveManager {
 
     void channelInputShutdown() {
         assert channel.eventLoop().inEventLoop();
+        cancelIfStateIsAFuture(inputShutdownFuture);
+        inputShutdownFuture = null;
         channelHalfShutdown("input", DuplexChannel::isOutputShutdown);
     }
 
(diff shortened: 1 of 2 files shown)


idelpivnitskiy:
Need to think about a good test for that, but ready for review.

idelpivnitskiy:
Hm... also need to adjust `KeepAliveManagerTest`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -377,38 +383,51 @@ private void close0(@Nullable Throwable cause) {
...
-                            channel, activeStreams);
+        final DuplexChannel duplexChannel = (DuplexChannel) channel;
+        duplexChannel.shutdownOutput().addListener(f -> {
+            if (duplexChannel.isInputShutdown()) {
+                LOGGER.debug("{} Input and output shutdown, closing the channel", channel);
+                channel.close();
+            } else {
+                // If we are in this state, we already finished GO_AWAY exchange and there are no more active streams.
+                // Give the remote peer some time to propagate InputShutdown, then force close the channel if it didn't
+                // happen withing reasonable time frame.
```
Scottmitch, line 419: nit: `withing` -> within`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -240,6 +244,8 @@ void channelOutputShutdown() {
 
     void channelInputShutdown() {
         assert channel.eventLoop().inEventLoop();
+        cancelIfStateIsAFuture(inputShutdownFuture);
```
Scottmitch, line 247: do we need to keep `inputShutdownFuture state?
- if no: can we just always initiate shutdown when this method is called?
- if yes: can we we return early from this method if `inputShutdownFuture != null` (e.g. we are already shutting everything down)?

idelpivnitskiy: the future is a timeout we are ready to wait for the input shutdown event. When we receive `ChannelInputShutdownReadComplete`, this method is invoked and we can cancel the future and finish the shutdown. 
if the timeout wins, we close the channel and I don't expect `ChannelInputShutdownReadComplete` to propagate. But even if it propagates, going though `channelHalfShutdown` once again should have no actual effect. wdyt?

I renamed `inputShutdownFuture` -> `inputShutdownTimeoutFuture`.

Scottmitch: makes sense (this method is notification that input has already changed, we don't want to leave the schedule task hanging). the event may propagate but as you say should be a noop.

Scottmitch (approved):

PR #2615, opened 2023-06-02
`KeepAliveManager`: include human-readable debug data in h2 frames

idelpivnitskiy:
Motivation:

1. `GO_AWAY` allows including a content for debugging purposes. It helps to debug closure based on frame-logging data.
2. `PING` frames always include data, using even/odd numbers help to understand originator of the frame.

Modifications:

- Include content when sending `GO_AWAY` frames;
- Use even numbers for keep-alive `PING`s and odd numbers for graceful close `PING`s;
- Test new behavior;

Result:

Possible to understand the cause for GO_AWAY and PING frame when frame logging is enabled.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
@@ -72,8 +72,8 @@ final class KeepAliveManager {
             AtomicIntegerFieldUpdater.newUpdater(KeepAliveManager.class, "activeStreams");
 
     // Use the last digit (even or odd) to distinguish PING frames when frame logging is enabled.
-    private static final long GRACEFUL_CLOSE_PING_CONTENT = ThreadLocalRandom.current().nextLong() | 0x01L;
-    private static final long KEEP_ALIVE_PING_CONTENT = ThreadLocalRandom.current().nextLong() & ~0x01L;
+    private static final long GRACEFUL_CLOSE_PING_CONTENT = ThreadLocalRandom.current().nextLong() | 0x01L; // odd
+    private static final long KEEP_ALIVE_PING_CONTENT = ThreadLocalRandom.current().nextLong() & ~0x01L;    // even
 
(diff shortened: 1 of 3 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
```diff
@@ -65,8 +70,17 @@ private enum State {
     private static final Logger LOGGER = LoggerFactory.getLogger(KeepAliveManager.class);
     private static final AtomicIntegerFieldUpdater<KeepAliveManager> activeStreamsUpdater =
             AtomicIntegerFieldUpdater.newUpdater(KeepAliveManager.class, "activeStreams");
-    private static final long GRACEFUL_CLOSE_PING_CONTENT = ThreadLocalRandom.current().nextLong();
-    private static final long KEEP_ALIVE_PING_CONTENT = ThreadLocalRandom.current().nextLong();
+
+    // Use the last digit (even or odd) to distinguish PING frames when frame logging is enabled.
+    private static final long GRACEFUL_CLOSE_PING_CONTENT = ThreadLocalRandom.current().nextLong() | 0x01L;
+    private static final long KEEP_ALIVE_PING_CONTENT = ThreadLocalRandom.current().nextLong() & 0xfffffffffffffffeL;
```
tkountis, line 76: nit: would `^ 0x01` result in the same and be easier to read?

tkountis (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/KeepAliveManager.java
tkountis: it wouldn't ignore that.

idelpivnitskiy: Nope, XOR won't work here because if the random bit was set to 0, it will flip it to 1.

idelpivnitskiy: I will do `& ~0x01L`

PR #2616, opened 2023-06-13
Change default SD retry strategy to exponential backoff with full jitter

idelpivnitskiy:
Motivation:

Currently, clients fallback to `retryWithConstantBackoffDeltaJitter` retry strategy for `ServiceDiscovery` if users didn't provide one. Retrying with a constant backoff could add more pressure on DNS servers if they start failing consistently. Exponential backoff will help us to slow down for permanent errors.

Modifications:
- `retryWithConstantBackoffDeltaJitter(10, 5)` -> `retryWithExponentialBackoffFullJitter(8, 256)` - changes retries from every 5-15 seconds to exponentially increasing delay from 0-8 to 0-256 seconds;

Result:

Less pressure on DNS servers for permanent failures.


bryce-anderson (approved):
🕶️

PR #2617, opened 2023-06-15
Bind DNS channels to a local wildcard address

idelpivnitskiy:
Motivation:

Currently, Netty's DNS resolver is configured to bind automatically after we write a packet. As the result, logs show only channel id but not local address:port pair. As the result, it's hard to correlate the packets we send with server-side DNS logs.

Modifications:
- Introduce `DnsServiceDiscovererBuilder` API for users to specify a local `SocketAddress`;
- Use `new InetSocketAddress(0)` by default for binding;
- Temporarily add a system property that allows users to quickly revert pre-existing default behavior:
`-Dio.servicetalk.dns.discovery.netty.skipBinding=boolean`;
- Log pre-initialized default values of `DefaultDnsServiceDiscovererBuilder` at debug level;

Result:

Users can see `[id: 0x1d761bfc, L:/0.0.0.0:63085]` instead of just `[id: 0x1d761bfc]` in Netty logs when DEBUG level is enabled for `io.netty.resolver.dns`.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
@@ -20,7 +20,11 @@ import io.servicetalk.client.api.ServiceDiscovererEvent;
 import io.servicetalk.transport.api.HostAndPort;
 import io.servicetalk.transport.api.IoExecutor;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
@@ -32,6 +36,7 @@ import static io.servicetalk.dns.discovery.netty.DnsClients.asSrvDiscoverer;
 import static io.servicetalk.dns.discovery.netty.DnsResolverAddressTypes.systemDefault;
 import static io.servicetalk.transport.netty.internal.GlobalExecutionContext.globalExecutionContext;
 import static io.servicetalk.utils.internal.DurationUtils.ensurePositive;
+import static java.lang.Boolean.getBoolean;
 import static java.time.Duration.ofSeconds;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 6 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -44,10 +49,43 @@
...
 public final class DefaultDnsServiceDiscovererBuilder implements DnsServiceDiscovererBuilder {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultDnsServiceDiscovererBuilder.class);
+
+    /**
+     * @deprecated This system property is introduced temporarily as a way for users to skip binding and revert the
+     * preexisting behavior.
+     */
+    @Deprecated // FIXME: 0.43 - consider removing this system property
+    private static final String SKIP_BINDING_PROPERTY = "io.servicetalk.dns.discovery.netty.skipBinding";
```
idelpivnitskiy, line 60: This is a quick escape way in case binding to the wildcard address causes any issues for users.

bryce-anderson: I'm not seeing where `SKIP_BINDING` is used. Did you intend to test it when constructing the `DefaultDnsClient` or when defining `DEFAULT_LOCAL_ADDRESS`? Presuming so, which do you prefer?

idelpivnitskiy: Good catch! I forget to use it 🙈

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -44,10 +49,44 @@
...
+    private static final boolean SKIP_BINDING = getBoolean(SKIP_BINDING_PROPERTY);
```
bryce-anderson, line 61: feel free to ignore: there seems to be little value in storing the value since it's only used to evaluate a static final field right below it.

bryce-anderson (approved):
I really like it. The small things that make it easier to debug are the undervalued heros of libraries.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -44,10 +49,44 @@
...
+
+    /**
+     * @deprecated This system property is introduced temporarily as a way for users to skip binding and revert the
+     * preexisting behavior.
+     */
+    @Deprecated // FIXME: 0.43 - consider removing this system property
+    private static final String SKIP_BINDING_PROPERTY = "io.servicetalk.dns.discovery.netty.skipBinding";
+    @Nullable
+    private static final SocketAddress DEFAULT_LOCAL_ADDRESS =
+            getBoolean(SKIP_BINDING_PROPERTY) ? null : new InetSocketAddress(0);
```
idelpivnitskiy, line 63: For reference, `java.net.DatagramSocket#send(DatagramPacket)` uses `new InetSocketAddress(0)` if it's not bound yet.

normanmaurer (approved):
LGTM

PR #2618, opened 2023-06-15
Client should log observed errors from `ServiceDiscoverer`

idelpivnitskiy:
Motivation:

If `ServiceDiscoverer` fails for the first discovery call of a new client, the exception is propagated to the request caller. If it fails after the first collection of resolved addresses is returned, users never see errors from `ServiceDiscoverer` unless they opt-in for a `ServiceDiscoverer`-specific observer. This hides errors for users by default.

Modification:

- `DefaultSingleAddressHttpClientBuilder$StatusAwareServiceDiscoverer` logs all observed errors at `debug` level;

Result:

Users can see errors from `ServiceDiscoverer` and react when error rate increases.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
@@ -80,6 +80,7 @@ import static io.servicetalk.http.api.HttpProtocolVersion.HTTP_2_0;
 import static io.servicetalk.http.netty.AlpnIds.HTTP_2;
 import static io.servicetalk.http.netty.StrategyInfluencerAwareConversions.toConditionalClientFilterFactory;
 import static io.servicetalk.http.netty.StrategyInfluencerAwareConversions.toConditionalConnectionFilterFactory;
+import static io.servicetalk.utils.internal.ThrowableUtils.rootCause;
 import static java.lang.Integer.parseInt;
 import static java.time.Duration.ofMinutes;
 import static java.time.Duration.ofSeconds;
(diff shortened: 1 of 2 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -730,7 +732,15 @@ void resetError() {
         @Override
         public Publisher<Collection<E>> discover(final U u) {
             return delegate().discover(u)
-                    .beforeOnError(status::nextError)
+                    .beforeOnError(t -> {
+                        if (LOGGER.isDebugEnabled()) {
+                            LOGGER.warn("Observed an error from {} while discovering '{}':", delegate(), u, t);
+                        } else {
+                            LOGGER.warn("Observed an error from {} while discovering '{}': {}",
+                                    delegate(), u, rootCause(t));
```
idelpivnitskiy, line 740: I considered not logging it if publisher didn't emit any `onNext` events, but decided to not over complicate the logic. If it's a permanent error, like `UnknownHostException`, users have to stop the app, reconfigure it, and run again. If it's a short blip, it won't hurt if we log it here and then fail the request. Without debug level we log only type + msg of the root cause.

bryce-anderson:
Do we have a sense of how common these will be? I worry that it could be something that floods the logs. In a perfect world this would be a metric.

bryce-anderson:
> Agreed, metric will be nice. Since it's internal to ST, log is the only option for now.
> I discussed with some folks. Let's start from `debug` level in ST. In addition, we will add metrics on users side. Based on that we can decide later if we need to promote to higher log level or not.

Sounds like a good plan to me.

bryce-anderson (approved):
Looks good. I don't know if you wanted to adjust your commit message since the log level of debug is probably worth mentioning, but I'll leave that up to you.

PR #2624, opened 2023-06-21
Add `SslConfig.handshakeTimeout()`

idelpivnitskiy:
Motivation:

Users need control the timeout for SSL handshake.

Modifications:

- Add `SslConfig.handshakeTimeout()` and `AbstractSslConfigBuilder.handshakeTimeout(Duration)`;
- Refactor `SslContextFactory` to store timeout value in `SslContext` attributes and reuse code between `forClient` and `forServer` factories;
- Refactor `SslUtils` to apply handshake timeout when a new `SslHandler` is created;
- Add `SslHandshakeTimeoutTest`;
- Add javadoc references from `SslConfig` builders to `SslConfig`;

Result:

Users can control SSL handshake timeouts via `SslConfig` API.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2021, 2023 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 package io.servicetalk.transport.api;
 
 import java.io.InputStream;
+import java.time.Duration;
 import java.util.List;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
@@ -23,6 +24,8 @@ import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLSessionContext;
 import javax.net.ssl.TrustManagerFactory;
+
+import static io.servicetalk.transport.api.AbstractSslConfigBuilder.DEFAULT_HANDSHAKE_TIMEOUT;
 
 /**
  * Specifies the configuration for TLS/SSL.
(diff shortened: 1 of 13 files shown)


bryce-anderson (approved):

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
```diff
@@ -155,9 +158,22 @@ public interface SslConfig {
...
         return null;
     }
+
+    /**
+     * Get the timeout for the handshake process.
+     * <p>
+     * Implementations can round the returned {@link Duration} to full time units, depending on their time granularity.
+     * {@link Duration#ZERO Zero duration} disables the timeout.
+     *
+     * @return the timeout for the handshake process or {@link Duration#ZERO} to disable it.
```
daschl, line 174: consider @see to the DEFAULT_HANDSHAKE_TIMEOUT

servicetalk-transport-netty-internal/src/testFixtures/java/io/servicetalk/transport/netty/internal/AddressUtils.java
```diff
@@ -56,7 +57,17 @@ public static InetSocketAddress localAddress(final int port) {
...
+        return serverHostAndPort(ctx.listenAddress());
+    }
+
+    /**
+     * Returns a {@link HostAndPort} representation of server's listening address.
+     *
+     * @param address The {@link SocketAddress} of the server
+     * @return a {@link HostAndPort} representation of server's listening address.
+     */
+    public static HostAndPort serverHostAndPort(final SocketAddress address) {
```
daschl, line 69: Since it would otherwise fail at runtime, consider using InetSocketAddress directly as the type for the argument

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
idelpivnitskiy: I can not, it's pkg-private constant.
Until 1.0 we made a decision to avoid clarifying default values in javadoc to give us some flexibility to adjust them. We can reconsider after 1.0

servicetalk-transport-netty-internal/src/testFixtures/java/io/servicetalk/transport/netty/internal/AddressUtils.java
idelpivnitskiy: The goal was to side cast to `InetSocketAddress` too.
Since it's a test utility only, expectation is that if there is a class cast issue, it will blow up for users immediately. I will keep it as-is for now. Thanks for noticing!

PR #2625, opened 2023-06-21
Add `ClientHello` configuration options for SNI

idelpivnitskiy:
Motivation:

Netty 4.1.94 exposed a new option to configure maximum length of a
`ClientHello` message. `SniHandler` also allows configuring a timeout
for parsing SNI info before starting the handshake.
See https://github.com/netty/netty/commit/535da17e45201ae4278c0479e6162bb4127d4c32

Modifications:

- Add `HttpServerBuilder.sslConfig(ServerSslConfig, Map, int, long)`
overload to give users a way for `SniHandler` configuration;
- Propagate new options via `TcpServerConfig`;
- Enhance `SniServerChannelInitializer` to apply new options when the
min required Netty version is detected;
- Set reasonable default values;
- Add `SniTest.sniTimeout(...)`;

Result:

Users can control the maximum `ClientHello` length and timeout for parsing
SNI information.

### DIFF
--- a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
+++ b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
@@ -39,8 +39,9 @@ import static java.util.Objects.requireNonNull;
  */
 public final class TcpServerConfig extends AbstractTcpConfig<ServerSslConfig> {
     /**
-     * The maximum length of client hello message as defined by
-     * <a href="https://www.rfc-editor.org/rfc/rfc5246#section-6.2.1">RFC5246</a>.
+     * The maximum length of ClientHello message as defined by
+     * <a href="https://www.rfc-editor.org/rfc/rfc5246#section-7.4.1.2">RFC5246</a> and
+     * <a href="https://datatracker.ietf.org/doc/html/rfc6347#section-3.2.3">RFC6347</a> as {@code 2^24 - 1}.
      */
     private static final int MAX_CLIENT_HELLO_LENGTH = 0xFFFFFF;
     private static final Duration DEFAULT_CLIENT_HELLO_TIMEOUT = ofSeconds(10); // same as default in Netty SslHandler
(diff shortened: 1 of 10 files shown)


servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
```diff
@@ -24,24 +24,41 @@
...
 
 /**
  * Configuration for TCP based servers.
  */
 public final class TcpServerConfig extends AbstractTcpConfig<ServerSslConfig> {
+    /**
+     * The maximum length of client hello message as defined by
+     * <a href="https://www.rfc-editor.org/rfc/rfc5246#section-6.2.1">RFC5246</a>.
+     */
+    private static final int MAX_CLIENT_HELLO_LENGTH = 0xFFFFFF;
```
idelpivnitskiy, line 45: @normanmaurer Can you point me where `0xFFFFFF` value comes from?
I followed the RFC link and it says `2^14` instead of `2^24 - 1`.

Source: https://github.com/netty/netty/commit/535da17e45201ae4278c0479e6162bb4127d4c32#diff-1c7c1e64a0afabaa56b28c9fd8a3c9173a8904b2cee64be327563b7d97b69086R41-R45

normanmaurer: It's been a while so I dont remember exactly but its also mentioned in other RFCs , for example: https://datatracker.ietf.org/doc/html/rfc6347#section-3.2.3.

normanmaurer: also that said I am not sure if this is a good default ... we were just very "conservative" as we didn't want to break people (there was no limit before). I think in practice using something much smaller is better.

daschl (approved):
lgtm

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
idelpivnitskiy: Thanks!
After reading https://www.rfc-editor.org/rfc/rfc5246#section-7.4.1.2, it doesn't look like ClientHello specifically has a limit but I'm agreed that TLS in general defines a limit for its messages as 2^24 - 1.

Agreed that it's not a perfect default. At least, it's better than unlimited. We will start with this for backward compatibility and we will try to reduce it after we experiment in apps.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/SniServerChannelInitializer.java
```diff
@@ -22,29 +22,86 @@
...
+        this.maxClientHelloLength = maxClientHelloLength;
+        this.clientHelloTimeoutMillis = clientHelloTimeoutMillis;
     }
 
     @Override
     public void init(final Channel channel) {
-        channel.pipeline().addLast(new SniHandlerWithPooledAllocator(sniMapping));
+        channel.pipeline().addLast(CAN_SET_ALL_SETTINGS ?
+                new SniHandlerWithAllSettings(sniMapping, maxClientHelloLength, clientHelloTimeoutMillis) :
+                new SniHandlerWithPooledAllocator(sniMapping));
```
idelpivnitskiy, line 104: Tested locally that this works as expected if older Netty version is at runtime

PR #2628, opened 2023-06-27
Allow configuring "negative" DNS TTL for failed resolutions

idelpivnitskiy:
Motivation:

Both JDK and Netty resolvers allow configuring "negative TTL" for
caching unsuccessful DNS resolutions. By default, it's configured to
0, but there is a Java system property to change it for
`java.net.InetAddress`.

Modifications:

- Add `DnsServiceDiscovererBuilder#ttl(int, int, int, int, int)`
overload that allows configuring negative TTL;
- Recognize `networkaddress.cache.negative.ttl` system property to
adjust the default value for Netty resolver;
- Log the default value at debug level;

Result:

Users can configure DNS caching for unsuccessful results.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
@@ -38,6 +38,7 @@ import static io.servicetalk.transport.netty.internal.GlobalExecutionContext.glo
 import static io.servicetalk.utils.internal.DurationUtils.ensurePositive;
 import static java.lang.Boolean.getBoolean;
 import static java.lang.Math.min;
+import static java.lang.System.getProperty;
 import static java.time.Duration.ofSeconds;
 import static java.util.Objects.requireNonNull;
 
@@ -68,9 +69,19 @@ public final class DefaultDnsServiceDiscovererBuilder implements DnsServiceDisco
     private static final int DEFAULT_MIN_TTL_CACHE_SECONDS = 0;
     private static final int DEFAULT_MAX_TTL_CACHE_SECONDS = 30;
     private static final int DEFAULT_TTL_POLL_JITTER_SECONDS = 4;
+    // System property recognized by JVM: https://docs.oracle.com/javase/8/docs/technotes/guides/net/properties.html
+    // We do not support "networkaddress.cache.ttl" because it's behavior in JDK is different. It overrides the original
+    // TTL from the server and caches exactly for the specified amount of time via the property.
+    private static final String NEGATIVE_TTL_CACHE_SECONDS_PROPERTY = "networkaddress.cache.negative.ttl";
+    private static final int DEFAULT_NEGATIVE_TTL_CACHE_SECONDS;
     private static final ServiceDiscovererEvent.Status DEFAULT_MISSING_RECOREDS_STATUS = EXPIRED;
 
     static {
+        final Integer negativeCacheTtlValue = parseProperty(NEGATIVE_TTL_CACHE_SECONDS_PROPERTY);
+        DEFAULT_NEGATIVE_TTL_CACHE_SECONDS = negativeCacheTtlValue == null ? 0 :
+                // A value of -1 indicates "cache forever".
+                (negativeCacheTtlValue < 0 ? Integer.MAX_VALUE : negativeCacheTtlValue);
+
         if (LOGGER.isDebugEnabled()) {
             LOGGER.debug("-D{}: {}", SKIP_BINDING_PROPERTY, getBoolean(SKIP_BINDING_PROPERTY));
             LOGGER.debug("Default local address to bind to: {}", DEFAULT_LOCAL_ADDRESS);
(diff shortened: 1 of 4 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -65,20 +67,38 @@ public final class DefaultDnsServiceDiscovererBuilder implements DnsServiceDisco
     private static final int DEFAULT_MIN_TTL_POLL_SECONDS = 10;
     private static final int DEFAULT_MAX_TTL_POLL_SECONDS = (int) TimeUnit.MINUTES.toSeconds(5);
     private static final int DEFAULT_MIN_TTL_CACHE_SECONDS = 0;
-    private static final int DEFAULT_MAX_TTL_CACHE_SECONDS = 0;
+    private static final int DEFAULT_MAX_TTL_CACHE_SECONDS = 30;
+    private static final int DEFAULT_TTL_POLL_JITTER_SECONDS = 4;
+    // System property recognized by JVM: https://docs.oracle.com/javase/8/docs/technotes/guides/net/properties.html
```
Scottmitch, line 72: nit: consider putting in java doc comment `/** <a href=".."></a> */`

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -170,11 +212,10 @@ public DefaultDnsServiceDiscovererBuilder ttl(final int minSeconds, final int ma
                     " (expected: 0 <= minCacheSeconds <= minSeconds(" + minSeconds +
                     ") <= maxCacheSeconds <= maxSeconds(" + maxSeconds + "))");
         }
-        this.minTTLSeconds = minSeconds;
-        this.maxTTLSeconds = maxSeconds;
-        this.minTTLCacheSeconds = minCacheSeconds;
-        this.maxTTLCacheSeconds = maxCacheSeconds;
-        return this;
+        if (negativeTTLCacheSeconds < 0) {
```
Scottmitch, line 215: this is called in a static context and will prevent the app from starting which maybe difficult to understand. did you consider logging an error and choosing a default value if invalid values are specified?

idelpivnitskiy: Actually, it's a leftover from some intermediate work when I had more properties. With only one property, there is no need to validate inside the static block as we always choose the default, even if we can not parse the value.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -68,9 +69,22 @@ public final class DefaultDnsServiceDiscovererBuilder implements DnsServiceDisco
...
+     * original TTL from the server and caches exactly for the specified amount of time instead of just being a max cap.
+     */
+    private static final String NEGATIVE_TTL_CACHE_SECONDS_PROPERTY = "networkaddress.cache.negative.ttl";
+    private static final int DEFAULT_NEGATIVE_TTL_CACHE_SECONDS;
     private static final ServiceDiscovererEvent.Status DEFAULT_MISSING_RECOREDS_STATUS = EXPIRED;
 
     static {
+        final Integer negativeCacheTtlValue = parseProperty(NEGATIVE_TTL_CACHE_SECONDS_PROPERTY);
+        DEFAULT_NEGATIVE_TTL_CACHE_SECONDS = negativeCacheTtlValue == null ? 0 :
+                // A value of -1 indicates "cache forever".
```
bryce-anderson, line 85: The way it's coded it's a value less than 0 indicates "cache forever".

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -339,11 +369,26 @@ private static DnsClientFilterFactory appendFilter(@Nullable final DnsClientFilt
...
+        final String propertyValue = getProperty(propertyName);
+        if (propertyValue == null) {
+            return null;
+        }
+        try {
+            return Integer.valueOf(propertyValue);
+        } catch (NumberFormatException e) {
+            // Ignore
+            return null;
+        }
```
bryce-anderson, line 392: I agree with @Scottmitch that if this happens we should log it as an error.

bryce-anderson (approved):

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
idelpivnitskiy: Updated the comment to clarify

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererBuilder.java
```diff
@@ -72,9 +73,45 @@ public interface DnsServiceDiscovererBuilder {
...
+     * than or equal to {@code minCacheSeconds}, and less than or equal to {@code maxSeconds}.
+     * @param negativeCacheSeconds The amount of time an unsuccessful (failed) result will be cached locally (in
+     * seconds), must be greater than or equal to {@code 0}. If other overloads are used, the default value will
+     * recognize the standard Java system property {@code networkaddress.cache.negative.ttl},
+     * like {@link java.net.InetAddress} does.
+     * @return {@code this}.
+     * @see #ttl(int, int)
+     * @see #ttl(int, int, int, int)
+     */
+    default DnsServiceDiscovererBuilder ttl(int minSeconds, int maxSeconds, int minCacheSeconds, int maxCacheSeconds,
```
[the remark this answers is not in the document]
idelpivnitskiy: I'm agreed that 5 look like at the limit. Taking into account that:
- it's a very rare use-case that someone needs this overload
- we already have 2 other overloads with the same pattern
- proposed approach requires designing more public API classes and deprecating existing overloads
I would propose sticking to the pattern and shipping as-is.
In general, it's rare that users change any settings on this builder.

daschl (approved):

PR #2629, opened 2023-06-27
Add `DnsServiceDiscovererBuilder.consolidateCacheSize(int)`

idelpivnitskiy:
Motivation:

Starting from Netty 4.1.88, it added a new features to consolidate all concurrent lookups for the same hostname instead of sending independent requests.

Modifications:

- Add new `DnsServiceDiscovererBuilder.consolidateCacheSize(int)` API;
- Consolidate 1024 concurrent DNS hostnames by default, log default value;

Result:

By default, up to 1024 different hostnames will consolidate concurrent lookups. Users have API to change this value, if necessary.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
@@ -62,7 +62,7 @@ public final class DefaultDnsServiceDiscovererBuilder implements DnsServiceDisco
     private static final SocketAddress DEFAULT_LOCAL_ADDRESS =
             getBoolean(SKIP_BINDING_PROPERTY) ? null : new InetSocketAddress(0);
     private static final DnsResolverAddressTypes DEFAULT_DNS_RESOLVER_ADDRESS_TYPES = systemDefault();
-    private static final int DEFAULT_CONSOLIDATE_CACHE_SIZE = 1024;
+    static final int DEFAULT_CONSOLIDATE_CACHE_SIZE = 1024;
     private static final int DEFAULT_MIN_TTL_POLL_SECONDS = 10;
     private static final int DEFAULT_MAX_TTL_POLL_SECONDS = (int) TimeUnit.MINUTES.toSeconds(5);
     private static final int DEFAULT_MIN_TTL_CACHE_SECONDS = 0;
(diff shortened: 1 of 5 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererBuilder.java
```diff
@@ -30,6 +30,21 @@
  * resolve {@code A}, {@code AAAA}, {@code CNAME}, and {@code SRV} type queries.
  */
 public interface DnsServiceDiscovererBuilder {
+    /**
+     * Set the maximum size of the cache that is used to consolidate concurrent lookups for different hostnames.
```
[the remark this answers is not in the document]
idelpivnitskiy: No, it's the max size of a hash map that holds `<Hostname, Future<Response>>` mappings. 
When a new request starts, it adds a new entry in the hash map. Then any other concurrent requests for the same hostname will take the future from the map.
When the Futute completes, it will remove the entry from the map.
Different hostnames will create new entries in the map.

bryce-anderson (approved):

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -62,6 +62,7 @@ public final class DefaultDnsServiceDiscovererBuilder implements DnsServiceDisco
     private static final SocketAddress DEFAULT_LOCAL_ADDRESS =
             getBoolean(SKIP_BINDING_PROPERTY) ? null : new InetSocketAddress(0);
     private static final DnsResolverAddressTypes DEFAULT_DNS_RESOLVER_ADDRESS_TYPES = systemDefault();
+    private static final int DEFAULT_CONSOLIDATE_CACHE_SIZE = 1024;
```
daschl, line 65: Is there any reasoning how we ended up with 1024 as the cache size number?

idelpivnitskiy: Not really, just a number that feels not too big and not too low, handling most of the use-cases that feel reasonable (having up to this number of unique hostnames). Happy to adjust if there are other proposals

daschl: No counter proposal, sounds reasonable to me!

daschl (approved):

PR #2631, opened 2023-06-28
Avoid double offloading a service when `offloadAll()` is used

idelpivnitskiy:
Motivation:

When `offloadAll()` strategy is used, offloading is applied for all paths. Invocation of the `delegate.handle(...)` becomes a single of the response. Because the response single also uses `subscribeOn(...)`, we are already offloaded. But the `shouldOffload` predicate for `handle(...)` is validated upfront, when we were on `IoThread`. This results in double offloading.

Modifications:

- If both `isMetadataReceiveOffloaded()` and `isSendOffloaded()` are true, wrap the `offloadHandle` with `defer` to validate the `shouldOffload` predicate at the time we already subscribed to the response single;

Result:

Less thread hops.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpServiceToOffloadedStreamingHttpService.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpServiceToOffloadedStreamingHttpService.java
@@ -70,12 +70,18 @@ public class StreamingHttpServiceToOffloadedStreamingHttpService implements Stre
             }
             final Single<StreamingHttpResponse> resp;
             if (additionalOffloads.isMetadataReceiveOffloaded() && shouldOffload.getAsBoolean()) {
-                final StreamingHttpRequest r = request;
-                resp = useExecutor.submit(
-                                () -> delegate.handle(wrappedCtx, r, responseFactory).shareContextOnSubscribe())
-                        // exec.submit() returns a Single<Single<StreamingHttpResponse>>, so flatten nested Single.
-                        // No need to apply shareContextOnSubscribe() again because unwrapped Single already shares ctx.
-                        .flatMap(identity());
+                final StreamingHttpRequest finalRequest = request;
+                // If isSendOffloaded() then we can already be offloaded by the time it subscribes to SubmitSingle.
+                // Wrap with defer to verify the shouldOffload predicate after subscribe.
+                resp = additionalOffloads.isSendOffloaded() ? Single.defer(() -> {
+                    if (shouldOffload.getAsBoolean()) {
+                        return offloadHandle(useExecutor, wrappedCtx, finalRequest, responseFactory)
+                                .shareContextOnSubscribe();
+                    } else {
+                        return delegate.handle(wrappedCtx, finalRequest, responseFactory)
+                                .shareContextOnSubscribe();
+                    }
+                }) : offloadHandle(useExecutor, wrappedCtx, request, responseFactory);
             } else {
                 resp = delegate.handle(wrappedCtx, request, responseFactory);
             }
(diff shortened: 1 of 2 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpServiceToOffloadedStreamingHttpService.java
```diff
@@ -70,12 +70,18 @@ public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
...
-                final StreamingHttpRequest r = request;
-                resp = useExecutor.submit(
-                                () -> delegate.handle(wrappedCtx, r, responseFactory).shareContextOnSubscribe())
-                        // exec.submit() returns a Single<Single<StreamingHttpResponse>>, so flatten nested Single.
-                        // No need to apply shareContextOnSubscribe() again because unwrapped Single already shares ctx.
-                        .flatMap(identity());
+                final StreamingHttpRequest finalRequest = request;
+                // If isSendOffloaded() then we can already be offloaded by the time it subscribes to SubmitSingle.
+                // Wrap with defer to verify the shouldOffload predicate after subscribe.
+                resp = additionalOffloads.isSendOffloaded() ? Single.defer(() -> {
```
Scottmitch, line 76: control flow can be simplified to reduce to 1 `shareContextOnSubscribe` to make it more clear its always applied: 

```diff
diff --git a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpServiceToOffloadedStreamingHttpService.java b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpServiceToOffloadedStreamingHttpService.java
index 9d46504f8..01a8481d3 100644
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpServiceToOffloadedStreamingHttpService.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpServiceToOffloadedStreamingHttpService.java
@@ -73,15 +73,11 @@ public class StreamingHttpServiceToOffloadedStreamingHttpService implements Stre
                 final StreamingHttpRequest finalRequest = request;
                 // If isSendOffloaded() then we can already be offloaded by the time it subscribes to SubmitSingle.
                 // Wrap with defer to verify the shouldOffload predicate after subscribe.
-                resp = additionalOffloads.isSendOffloaded() ? Single.defer(() -> {
-                    if (shouldOffload.getAsBoolean()) {
-                        return offloadHandle(useExecutor, wrappedCtx, finalRequest, responseFactory)
-                                .shareContextOnSubscribe();
-                    } else {
-                        return delegate.handle(wrappedCtx, finalRequest, responseFactory)
-                                .shareContextOnSubscribe();
-                    }
-                }) : offloadHandle(useExecutor, wrappedCtx, request, responseFactory);
+                resp = additionalOffloads.isSendOffloaded() ? Single.defer(() -> (shouldOffload.getAsBoolean() ?
+                        offloadHandle(useExecutor, wrappedCtx, finalRequest, responseFactory) :
+                        delegate.handle(wrappedCtx, finalRequest, responseFactory))
+                            .shareContextOnSubscribe()) :
+                        offloadHandle(useExecutor, wrappedCtx, request, responseFactory);
             } else {
                 resp = delegate.handle(wrappedCtx, request, responseFactory);
             }
```

Scottmitch (approved):

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpServiceToOffloadedStreamingHttpService.java
idelpivnitskiy: Considered that but found it less readable and easy to miss the correct bracket position to `shareContextOnSubscribe()`. 
I like ternary operator for a single case, but found it hard to read when multiple ternary operators are nested across multiple lines. Will keep it as-is.

Scottmitch: ack. you can also adjust parenthesis and formatting to make it more readable

PR #2632, opened 2023-06-29
[Single|Completable].toFuture() cancellable handle exception

Scottmitch:
Motivation:
[Single|Completable].toFuture() may hang if there is a thread blocked on get(), a different thread invokes cancel, and the Cancellable for the Subscriber chain throws.

Modifications:
- Unblock the CountDownLatch in a finally block
- Preserve stack trace from the thread that invokes cancel

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceToFuture.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceToFuture.java
@@ -29,12 +29,11 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import javax.annotation.Nullable;
 
+import static io.servicetalk.concurrent.api.ThrowableWrapper.isThrowableWrapper;
 import static java.util.Objects.requireNonNull;
 
 abstract class SourceToFuture<T> implements Future<T> {
-
     static final Object NULL = new Object();
-    private static final Object CANCELLED = new Object();
 
     @SuppressWarnings("rawtypes")
     private static final AtomicReferenceFieldUpdater<SourceToFuture, Object> valueUpdater =
@@ -65,9 +64,13 @@ abstract class SourceToFuture<T> implements Future<T> {
 
     @Override
     public final boolean cancel(final boolean mayInterruptIfRunning) {
-        if (valueUpdater.compareAndSet(this, null, CANCELLED)) {
-            cancellable.cancel();
-            latch.countDown();
+        if (value == null && valueUpdater.compareAndSet(this, null,
+                new CancellationWrapper(new CancellationException()))) {
+            try {
+                cancellable.cancel();
+            } finally {
+                latch.countDown();
+            }
             return true;
         }
         return false;
@@ -75,7 +78,7 @@ abstract class SourceToFuture<T> implements Future<T> {
 
     @Override
     public final boolean isCancelled() {
-        return value == CANCELLED;
+        return CancellationWrapper.isCancellationWrapper(value);
     }
 
     @Override
@@ -122,10 +125,10 @@ abstract class SourceToFuture<T> implements Future<T> {
         if (value instanceof Throwable) {
             throw new ExecutionException((Throwable) value);
         }
-        if (value == CANCELLED) {
-            throw new CancellationException();
+        if (CancellationWrapper.isCancellationWrapper(value)) {
+            throw ((CancellationWrapper) value).exception;
         }
-        if (value instanceof ThrowableWrapper) {
+        if (isThrowableWrapper(value)) {
             return (T) ((ThrowableWrapper) value).unwrap();
         }
         return (T) value;
@@ -168,6 +171,18 @@ abstract class SourceToFuture<T> implements Future<T> {
         @Override
         public void onComplete() {
             setValue(NULL);
+        }
+    }
+
+    private static final class CancellationWrapper {
+        private final CancellationException exception;
+
+        private CancellationWrapper(final CancellationException exception) {
+            this.exception = exception;
+        }
+
+        static boolean isCancellationWrapper(@Nullable Object o) {
+            return o != null && o.getClass().equals(CancellationWrapper.class);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/AbstractToFutureTest.java
```diff
@@ -67,6 +71,28 @@ void testSubscribed() {
         assertThat(isSubscribed(), is(true));
     }
 
+    @Test
+    void testCancellableThrows() throws InterruptedException, ExecutionException {
+        doThrow(DELIBERATE_EXCEPTION).when(mockCancellable).cancel();
+        Future<T> future = toFuture();
+        ExecutorService executorService = Executors.newCachedThreadPool();
```
idelpivnitskiy, line 78: Consider re-using already available `EXEC`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceToFuture.java
```diff
@@ -122,10 +125,10 @@ private T reportGet(@Nullable final Object value) throws ExecutionException {
         if (value instanceof Throwable) {
             throw new ExecutionException((Throwable) value);
         }
-        if (value == CANCELLED) {
-            throw new CancellationException();
+        if (CancellationWrapper.isCancellationWrapper(value)) {
+            throw ((CancellationWrapper) value).exception;
```
idelpivnitskiy, line 129: As a user, I will much more appreciate if the stack-trace shows me what code path of my business logic throws `CancellationException` (where I blocked) instead of seeing what code path cancelled, because it helps me to understand what request failed. 
The pre-existing approach aligns with Java behavior for `java.util.concurrent.FutureTask`.

If there is a real need to preserve a stack trace that cancelled, consider adding it as a cause or a suppressed exception for the one we create at this line today.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceToFuture.java
```diff
@@ -65,17 +64,21 @@ public final void onError(final Throwable t) {
...
-        if (valueUpdater.compareAndSet(this, null, CANCELLED)) {
-            cancellable.cancel();
-            latch.countDown();
+        if (value == null && valueUpdater.compareAndSet(this, null,
+                new CancellationWrapper(new CancellationException()))) {
+            try {
+                cancellable.cancel();
+            } finally {
+                latch.countDown();
+            }
```
idelpivnitskiy, line 73: Great catch!

idelpivnitskiy (changes requested):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceToFuture.java
Scottmitch: [CompletableFuture](https://github.com/openjdk/jdk17/blob/master/src/java.base/share/classes/java/util/concurrent/CompletableFuture.java#L2479) takes the stack trace from "where it was cancelled". I think this provides more value bcz the place that does a `get()` knows its stack, but doesn't know where/why it was cancelled.

As you say, `java.util.concurrent.FutureTask` does it the other way (throws from `get()`).

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/AbstractToFutureTest.java
Scottmitch: Since we are testing `Future` here I didn't want to use our executors which we would also use `Future` to block.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceToFuture.java
Scottmitch: CompletableFuture javadocs talk to this briefly:

https://github.com/openjdk/jdk17/blob/master/src/java.base/share/classes/java/util/concurrent/CompletableFuture.java#L97-L103
```java
 * <p>CompletableFuture also implements {@link Future} with the following
 * policies: <ul>
 *
 * <li>Since (unlike {@link FutureTask}) this class has no direct
 * control over the computation that causes it to be completed,
 * cancellation is treated as just another form of exceptional
 * completion.  Method {@link #cancel cancel} has the same effect as
 * {@code completeExceptionally(new CancellationException())}. Method
 * {@link #isCompletedExceptionally} can be used to determine if a
 * CompletableFuture completed in any exceptional fashion.
```

idelpivnitskiy: Yeah, no objections on preserving both. It's useful addition. Just based on my experience, these "async exception stack-traces" bring more confusion vs help. So, preserving the caller's stack-trace at the first place is more important. Thanks for improving it!

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceToFuture.java
```diff
@@ -65,17 +64,21 @@ public final void onError(final Throwable t) {
 
     @Override
     public final boolean cancel(final boolean mayInterruptIfRunning) {
-        if (valueUpdater.compareAndSet(this, null, CANCELLED)) {
-            cancellable.cancel();
-            latch.countDown();
+        if (value == null && valueUpdater.compareAndSet(this, null,
+                new CancellationWrapper(new CancellationException()))) {
```
idelpivnitskiy, line 68: Consider adding a message saying this is where it was cancelled

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceToFuture.java
```diff
@@ -122,10 +125,12 @@ private T reportGet(@Nullable final Object value) throws ExecutionException {
         if (value instanceof Throwable) {
             throw new ExecutionException((Throwable) value);
         }
-        if (value == CANCELLED) {
-            throw new CancellationException();
+        if (CancellationWrapper.isCancellationWrapper(value)) {
+            CancellationException exception = new CancellationException();
```
idelpivnitskiy, line 129: Consider adding a message describing this is what path was cancelled

idelpivnitskiy (approved):

PR #2633, opened 2023-06-29
Do not fail IoUring test on older CI systems

daschl:
Motivation:

Currently, the ioUringIsAvailableOnLinux test assumes that if when run under CI, the kernel version is always up to par with what netty needs to get IoUring enabled. This assumption might not always hold true, so allow to skip the test gracefully when CI has an older kernel version instead of failing the test suite.


idelpivnitskiy (approved):

idelpivnitskiy:
Thanks!

PR #2634, opened 2023-06-30
Apply optimized execution strategy for converted connections

idelpivnitskiy:
Motivation:

When client is converted to a different API, the API optimizes default strategy. We should do the same when we convert a reserved connection to a different API.

Modifications:

- Apply an execution strategy inside request method of every `ReservedStreamingHttpConnectionToReserved*HttpConnection`;

Result:

Converted connections apply the same strategy as converted clients.


servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpClientToBlockingHttpClient.java
```diff
@@ -144,6 +145,7 @@ public <T> BlockingIterable<? extends T> transportEventIterable(final HttpEventK
 
         @Override
         public HttpResponse request(final HttpRequest request) throws Exception {
+            request.context().putIfAbsent(HTTP_EXECUTION_STRATEGY_KEY, strategy);
```
idelpivnitskiy, line 148: Sorry, I don't have cycles right now to deal with `ClientEffectiveStrategyTest`.
I've verified locally that it works as expected after my fix and filed #2635.
The idea is that the request method for every `ReservedStreamingHttpConnectionToReserved*HttpConnection` should behave the same way as its top-level `StreamingHttpClientTo*HttpClient`.

Scottmitch: can you add some comments maybe an internal static utility function that makes it easier to track?

Scottmitch (approved):

PR #2639, opened 2023-07-05
Publisher[retry|repeat] operators demand management if onNext throws

Scottmitch:
Motivation:
Publisher `retry*` and `repeat*` operators need to persist outstanding demand between "re-do" operations. Otherwise the downstream Subscriber will have requested signals, but they won't be requested from the new async source after switching. If a downstream operator (or Subscriber) throws from onNext the retry/repeat operator has to decrement demand to avoid violating the Reactive Streams specification (no more than request(n) amount of onNext(..)), but downstream Subscribers that didn't see the onNext (because an earlier operator threw) will be hung waiting for request(n) that will never be delivered.

Modifications:
- Publisher[retry|repeat] introduce a flag to determine if downstream onNext throwing should be terminal. This mode should be used unless it is known the onNext signal always makes it to the last Subscriber and its request(n) accounting can tolerate the exception it throws.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -2253,15 +2253,10 @@ public abstract class Publisher<T> {
      * <ul>
      *     <li>{@code true} means that exceptions thrown from downstream {@link Subscriber#onNext(Object)} will be
      *     caught, cancel the {@link Subscription}, propagate a {@link Subscriber#onError(Throwable)} downstream, and
-     *     no retry will be attempted. <b>Unless you are sure all downstream operators consume the
-     *     {@link Subscriber#onNext(Object)} this option SHOULD be used</b>. Otherwise, the outstanding demand counting
-     *     in this operator will be incorrect and may lead to a "hang" (e.g. this operator thinks demand has been
-     *     consumed downstream so won't request it upstream after the retry, but if not all downstream operators see the
-     *     signal because one before threw, they may wait for a signal they requested but will never be delivered).</li>
+     *     no retry will be attempted.</li>
      *     <li>{@code false} means that exceptions thrown from downstream {@link Subscriber#onNext(Object)} will NOT
-     *     be caught and will propagate upstream. Use with caution as this may lead to incorrect demand accounting and
-     *     "hang" (see above). Example use case where this option makes sense is if the last {@link Subscriber} is
-     *     the only one expected to throw and demand is correctly accounted.</li>
+     *     be caught and will propagate upstream. May lead to incorrect demand accounting and "hangs" if this operator
+     *     isn't the last in the chain.</li>
      * </ul>
      * @param shouldRetry {@link BiIntPredicate} that given the retry count and the most recent {@link Throwable}
      * emitted from this {@link Publisher} determines if the operation should be retried.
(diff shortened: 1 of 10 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2205,20 +2205,74 @@ public final Publisher<T> concatPropagateCancel(Completable next) {
...
+     * emitted from this {@link Publisher} determines if the operation should be retried.
      * @return A {@link Publisher} that emits all items from this {@link Publisher} and re-subscribes if an error is
      * emitted if the passed {@link BiIntPredicate} returned {@code true}.
-     *
      * @see <a href="https://reactivex.io/documentation/operators/retry.html">ReactiveX retry operator.</a>
+     * @see #retry(boolean, BiIntPredicate)
      */
     public final Publisher<T> retry(BiIntPredicate<Throwable> shouldRetry) {
-        return new RedoPublisher<>(this,
+        return retry(false, shouldRetry);
```
Scottmitch, line 2216: I'm tempted to have the default be `true` because it is generally safer, error propagation is more explicit, and you will never "hang". However that would be a breaking change, and may not be the right choice if the `Subscriber` is the one that throws (see `PublisherBufferTest.java` below).

wdyt?

tkountis: The spec says
>Calling onSubscribe, onNext, onError or onComplete MUST [return normally](https://github.com/reactive-streams/reactive-streams-jvm#term_return_normally) except when any provided parameter is null in which case it MUST throw a java.lang.NullPointerException to the caller, for all other situations the only legal way for a Subscriber to signal failure is by cancelling its Subscription. In the case that this rule is violated, any associated Subscription to the Subscriber MUST be considered as cancelled, and the caller MUST raise this error condition in a fashion that is adequate for the runtime environment.

which covers the solutions here, doesn't it?
I think the test may be biased toward our implementation of the operators in the chain. I believe defaulting to `true` makes more sense.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2256,23 +2310,83 @@ public final Publisher<T> retry(BiIntPredicate<Throwable> shouldRetry) {
...
+     *             }
+     *         }
+     *     }
+     * }</pre>
+     * @param terminateOnNextException
+     * <ul>
+     *     <li>{@code true} means that exceptions thrown from downstream {@link Subscriber#onNext(Object)} will be
+     *     caught, cancel the {@link Subscription}, propagate a {@link Subscriber#onError(Throwable)} downstream, and
+     *     no retry will be attempted. <b>Unless you are sure all downstream operators consume the
+     *     {@link Subscriber#onNext(Object)} this option SHOULD be used</b>. Otherwise, the outstanding demand counting
```
tkountis, line 2369: my two cents, this is too much information for a ST user. This should be javadoc for the try...catch inside the operator. Consider keeping the first sentence only.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
Scottmitch: sounds good. unless there are objections in this PR I'll keep default at `false` and submit a followup PR to make the change more explicit/visible.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
Scottmitch: I reduce the javadocs on these operators.

tkountis (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
idelpivnitskiy: Oh, that's a very good catch!

Interestingly, I've checked other RS implementations (reactor and rxjava3) and they let an exception from `onNext` propagate upstream, as we do when ``terminateOnNextException == false`.

> The spec says...

The most frequent source of exceptions will be operators like `Publisher.map` or `filter`. I've seen users throwing from the mapper intentionally. 

I would say that we need to handle retry and repeat separately. For repeat, it makes sense to try-catch, cancel, terminate (like `terminateOnNextException == true`) and that should be the only behavior. We do not need an overload that takes a boolean. This will be consistent with `Single.repeat*`.

The retry is a little bit more interesting because it handles exceptions. However, I would say that  `terminateOnNextException == true` should be an expected behavior.
Reasoning: `retry*` should retry exceptions originated by the upstream. If we already delivered `onNext` and the exception was thrown by `onNext` downstream, it should not be considered for retry predicate.
Hesitations:
1. Other RS implementations behave like ``terminateOnNextException == false` 😞 
2. It's a behavior change for existing users.

idelpivnitskiy:
Code change LGTM

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
tkountis: >retry* should retry exceptions originated by the upstream. If we already delivered onNext and the exception was thrown by onNext downstream, it should not be considered for retry predicate.

Exactly that 👍 - It's very unexpected to see this behavior, and hard to debug, because you are paying attention to the wrong part of the stream.

>Other RS implementations behave like terminateOnNextException == false

sadly true, which makes me think that they likely have the same bug.

>It's a behavior change for existing users.

The risk of having a silent bug like this is real, I would say let's use our communication channels, to highlight this change, and have a release that introduces just this, so people can move back to the previous version if needed, we can iterate if we notice bad feedback. Maybe even add a WARN in the output when the retry catches an uncaught exception?

PR #2642, opened 2023-07-05
Publisher[retry|repeat] change default exception handling mode

Scottmitch:
Motivation:
The current default mode of Publisher retry/repeat operators allows exceptions to propagate from onNext. However this may result in incorrect demand management and lead to "hangs" due to off-by-one demand (downstream subscriber requested X, but may only see X-1 if an earlier downstream subscriber threw). ReactiveStreams spec specifies that Subscriber methods are generally not expected to throw [1]. For these reasons the current default maybe unsafe and isn't commonly expected. The alternative mode makes exception handling explicit and won't result in hangs.

[1] https://github.com/reactive-streams/reactive-streams-jvm#2.13

Modifications:
- Change default mode of operation to not allow downstream exceptions to propagate upstream.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
@@ -2213,7 +2213,7 @@ public abstract class Publisher<T> {
      * @see #retry(boolean, BiIntPredicate)
      */
     public final Publisher<T> retry(BiIntPredicate<Throwable> shouldRetry) {
-        return retry(false, shouldRetry);
+        return retry(true, shouldRetry);
     }
 
     /**
@@ -2315,7 +2315,7 @@ public abstract class Publisher<T> {
      * @see #retryWhen(boolean, BiIntFunction)
      */
     public final Publisher<T> retryWhen(BiIntFunction<Throwable, ? extends Completable> retryWhen) {
-        return retryWhen(false, retryWhen);
+        return retryWhen(true, retryWhen);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


tkountis (approved):
thx

bryce-anderson (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2404,7 +2404,7 @@ public final Publisher<T> retryWhen(boolean terminateOnNextException,
      * @see #repeat(boolean, IntPredicate)
      */
     public final Publisher<T> repeat(IntPredicate shouldRepeat) {
-        return repeat(false, shouldRepeat);
+        return repeat(true, shouldRepeat);
```
idelpivnitskiy, line 2407: I've provided a detailed reply on the original PR: https://github.com/apple/servicetalk/pull/2639#discussion_r1253763692

Consider removing `repeat*` overloads that take the boolean and always use `terminateOnNextException == true`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2213,7 +2213,7 @@ public final Publisher<T> concatPropagateCancel(Completable next) {
      * @see #retry(boolean, BiIntPredicate)
      */
     public final Publisher<T> retry(BiIntPredicate<Throwable> shouldRetry) {
-        return retry(false, shouldRetry);
+        return retry(true, shouldRetry);
```
idelpivnitskiy, line 2216: While I'm +1 for this default behavior change, it's disappointing that reasons to have an extra overload that takes a boolean are:
1. Other RS implementations behave like terminateOnNextException == false
2. It's a behavior change for existing users.

Scottmitch: agreed with 1 & 2 but also there are legit cases (you are subscribing, and your subscriber throws due to existing application logic) that would be difficult to orchestrate control flow w/out this option.

PR #2645, opened 2023-07-05
Enhance debug logging for `ReservableRequestConcurrencyControllers`

idelpivnitskiy:
Motivation:

Provide more visibility for `ReservableRequestConcurrencyControllers` and how it changes internal state when new events received.

Modifications:

- Pass a connection to `ReservableRequestConcurrencyControllers` to always include connectionId when logging;
- Log when retrying MaxConcurrentStreamsViolatedStacklessHttp2Exception;
- Remove logging transport events from `AbstractLBHttpConnectionFactory` because we log it in `ReservableRequestConcurrencyControllers` now;

Results:

All `ReservableRequestConcurrencyControllers` logs contain connectionId.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLBHttpConnectionFactory.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLBHttpConnectionFactory.java
@@ -17,11 +17,9 @@ package io.servicetalk.http.netty;
 
 import io.servicetalk.client.api.ConnectionFactory;
 import io.servicetalk.client.api.ConnectionFactoryFilter;
-import io.servicetalk.client.api.ConsumableEvent;
 import io.servicetalk.client.api.ReservableRequestConcurrencyController;
 import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
-import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.context.api.ContextMap;
 import io.servicetalk.http.api.FilterableStreamingHttpConnection;
@@ -43,7 +41,6 @@ import java.util.function.Function;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.api.AsyncCloseables.emptyAsyncCloseable;
-import static io.servicetalk.http.netty.AbstractStreamingHttpConnection.MAX_CONCURRENCY_NO_OFFLOADING;
 import static io.servicetalk.transport.api.TransportObservers.asSafeObserver;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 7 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
```diff
@@ -113,6 +113,7 @@ public String toString() {
 
     private abstract static class AbstractReservableRequestConcurrencyController
             implements ReservableRequestConcurrencyController {
+        private static final HttpEventKey<ConsumableEvent<Integer>> MAX_CONCURRENCY_KEY = MAX_CONCURRENCY_NO_OFFLOADING;
```
bryce-anderson, line 116: Do we anticipate this changing in the future or something? I would have personally just inlined it.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
```diff
@@ -207,12 +211,13 @@ public void onNext(@Nullable final ConsumableEvent<Integer> event) {
 
                 @Override
                 public void onError(final Throwable t) {
-                    LOGGER.info("Unexpected error from transportEventStream(MAX_CONCURRENCY).", t);
+                    LOGGER.info("{} Unexpected error from transportEventStream({})",
+                            connection, MAX_CONCURRENCY_KEY, t);
```
bryce-anderson, line 215: I see two placeholders and three args here. Sadly I don't think we can use the `info(String, Throwable)` with other arguments.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLBHttpConnectionFactory.java
```diff
@@ -121,11 +118,8 @@ public final Single<FilterableStreamingHttpLoadBalancedConnection> newConnection
                     // Apply connection filters:
                     FilterableStreamingHttpConnection filteredConnection =
                             connectionFilterFunction != null ? connectionFilterFunction.create(conn) : conn;
-                    return protocolBinding.bind(filteredConnection, newConcurrencyController(
-                            filteredConnection.transportEventStream(MAX_CONCURRENCY_NO_OFFLOADING)
-                                    .beforeOnNext(event -> LOGGER.debug("{} Received {} event: {}",
-                                            conn, MAX_CONCURRENCY_NO_OFFLOADING, event)),
-                            filteredConnection.onClosing()), context);
+                    return protocolBinding.bind(filteredConnection, newConcurrencyController(filteredConnection),
+                            context);
```
bryce-anderson, line 122: This change sure feels cleaner. 👍

bryce-anderson (approved):
lgtm after fixing the log format string.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
idelpivnitskiy: This is expected, `connection` and `MAX_CONCURRENCY_KEY` will go into placeholders, while `t` will be translated into stack-trace. If you look at the code before, it had zero placeholders and a single `t`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
idelpivnitskiy: No, just wanted to make sure we use the same key everywhere: as an argument for `transportEvents` and for logging. 
It helps to make it clear, because we also have a public `HttpEventKey.MAX_CONCURRENCY` key that offloads events.

bryce-anderson: I did see the `MAX_CONCURRENCY` which is part of why this seemed somewhat less clear, but it's obviously subjective. Feel free to ignore, especially if you don't have anything else to commit with it.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReservableRequestConcurrencyControllers.java
bryce-anderson: TIL. The code before used the `info(Object, Throwable)` method so it was obvious.

bryce-anderson (approved):

PR #2647, opened 2023-07-11
`RetryingHttpRequesterFilter`: don't wait for LB if it's already ready

idelpivnitskiy:
Motivation:

`NoAvailableHostException` can be thrown when either LB is not ready yet or when it's already ready, but all hosts turned into unhealthy state. In this case, there is no need to wait for "ready" event, because the `onHostsAvailable` completable is already completed.

Modifications:
- `LoadBalancerReadySubscriber` returns `null` if it's already "ready";
- `RetryingHttpRequesterFilter.OuterRetryStrategy` fails fast when LB is ready but we received `NoAvailableHostException`;
- Test this behavior;

Result:

We do not keep retrying when LB is already ready.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancerReadySubscriber.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancerReadySubscriber.java
@@ -15,6 +15,7 @@
  */
 package io.servicetalk.http.netty;
 
+import io.servicetalk.client.api.LoadBalancer;
 import io.servicetalk.client.api.LoadBalancerReadyEvent;
 import io.servicetalk.concurrent.CompletableSource.Processor;
 import io.servicetalk.concurrent.PublisherSource.Subscriber;
@@ -24,7 +25,6 @@ import io.servicetalk.concurrent.internal.DelayedCancellable;
 
 import javax.annotation.Nullable;
 
-import static io.servicetalk.concurrent.api.Completable.completed;
 import static io.servicetalk.concurrent.api.Processors.newCompletableProcessor;
 import static io.servicetalk.concurrent.api.SourceAdapters.fromSource;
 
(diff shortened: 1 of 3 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/LoadBalancerReadySubscriber.java
```diff
@@ -38,13 +38,18 @@ final class LoadBalancerReadySubscriber extends DelayedCancellable implements Su
...
      * @return A {@link Completable} that will complete when a {@link LoadBalancerReadyEvent} returns {@code true}
-     * from {@link LoadBalancerReadyEvent#isReady()}.
+     * from {@link LoadBalancerReadyEvent#isReady()} or {@code null} if the {@link LoadBalancer} is already ready.
      */
-    public Completable onHostsAvailable() {
+    @Nullable
+    Completable onHostsAvailable() {
         Processor onHostsAvailable = this.onHostsAvailable;
-        return onHostsAvailable == null ? completed() : fromSource(onHostsAvailable);
+        return onHostsAvailable == null ? null : fromSource(onHostsAvailable);
```
daschl, line 52: I'm not sure if `null` is a great signal here that something is ready - the old behavior (completing immediately) seemed like a clear signal that something is available.

For my understanding, it seems that there are two different signals "merged" here:

-  Load balance is ready (LoadBalancerReadyEvent#isReady())
-  Hosts are available

So if my understanding is correct the load balancer can become ready only once (so completable semantics would suffice) vs. hosts can become available, then unavailable, then available again. If that is true, would it make sense to turn hostsAvailable from a Completable into a Publisher which emits state? (Available / Unavailable) and caches the last one? So it doesn't matter if one subscribes to the hot publisher, they would just add a .takeUntil(e == available) or similar.

idelpivnitskiy: `onHostsAvailable` is the same as LB publisher emits "ready" event. 
LB can turn into "not-ready" state at any time late if SD propagated all "unavailable" events.
We can not use `Publisher` here because we have to translate publisher events from LB to `Completable`, bcz this is what is used as a retry signal.

idelpivnitskiy: Alternative proposal is in #2648

idelpivnitskiy: @daschl Thanks for the question! It helped to find another way that does not require changes for existing `LoadBalancerReadySubscriber`

PR #2651, opened 2023-07-13
Javadoc: fix broken RFC link.

vchimishuk:
`https://tools.ietf.org/search/` prefix is broken. All other Javadoc links has `https://tools.ietf.org/html/` prefix and works just fine.


idelpivnitskiy (approved):
Thank you!

PR #2652, opened 2023-07-14
Publisher.timeoutDemand(Duration) not timing out demand

Scottmitch:
Motivation:
The Publisher.timeoutDemand(Duration) method calls timeout operator overload instead of the timeoutDemand overload and is therefor applying the wrong timeout.


Scottmitch:
good find @tkountis !

tkountis (approved):

PR #2658, opened 2023-07-21
Remove `gradle-pitest-plugin`

idelpivnitskiy:
Motivation:

`gradle-pitest-plugin` is not an official plugin. Its latest version was released in Nov 2022. Gradle 8 complains that it uses deprecated API. Since we never actually used pitest, we should consider removing it.

Modifications:

- Remove `gradle-pitest-plugin` and its configuration;
- Wrap `sourceCompatibility/targetCompatibility` with `java` block, according to the guideline here:
https://docs.gradle.org/8.2.1/userguide/upgrading_version_8.html#java_convention_deprecation

Result:

No warnings from Gradle 8 about using deprecated API.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -73,7 +73,6 @@ assertJCore=3.23.1
 hamcrestVersion=2.2
 mockitoCoreVersion=4.11.0
 spotbugsPluginVersion=5.0.13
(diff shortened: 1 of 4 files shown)


Scottmitch (commented):
👍

PR #2662, opened 2023-07-25
use semantic convention APIs for opentelemetry

aleqi200:
Motivation:

Today the tags are manually handcrafted with strings, if the opentelemetry api changes, the tags will be outdated, using the instrumenter available from the ```io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-semconv``` we can make it all dynamic based on the specification without understanding the specification and its tags.

Result:

API will become more stable and evolve easier together with opentelemetry, no more handcrafted strings and the conventions will be driven by the instrumentation API.


aleqi200:
The tags are now showing the following:
Screenshot 2023-07-27 at 2 20 12 PM
The tags are driven by the semantic conventions available in the API.

| Tag | Value | 
| -- |  -- |
| http.method | POST |  
| http.request.header.transaction_id | [     "1234" ] |  
| http.request.header.x_shield_req_id | [     "1234-asdf" ] |  
| http.request_content_length | 87 |  
| http.response_content_length | 36 |  
| http.route | /job |  
| http.status_code | 200 |  
| http.target | /job |  
| internal.span.format | proto |  
| net.host.name | localhost |  
| net.host.port | 8082 |  
| net.protocol.name | http |  
| net.protocol.version | 1.1 |  
| otel.library.name | io.servicetalk |  
| otel.status_code | OK |  
| span.kind | server |  
| thread.id | 24 |  
| thread.name | servicetalk-global-io-executor-1-3 |  
| user_agent.original | curl/7.88.1

This will allow capture of request/response headers like above ```http.request.header.transaction_id```.

aleqi200:
Metrics will be available for spans as well.
The following metrics were captured using the updated API with opentelemetry and prometheus:
Screenshot 2023-07-26 at 9 50 46 AM

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpClientCommonAttributesGetter.java
```diff
@@ -0,0 +1,82 @@
...
+import io.servicetalk.http.api.HttpResponseMetaData;
+import io.servicetalk.transport.api.HostAndPort;
+
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpClientAttributesGetter;
+
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.Nullable;
+
+final class ServicetalkHttpClientCommonAttributesGetter
```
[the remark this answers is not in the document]
daschl: (same on the server getter)

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpClientCommonAttributesGetter.java
```diff
@@ -0,0 +1,82 @@
...
+        if (value != null) {
+            return Collections.singletonList(value.toString());
+        }
+        return Collections.emptyList();
+    }
+
+    @Nullable
+    @Override
+    public String getUrlFull(HttpRequestMetaData request) {
+        HostAndPort effectiveHostAndPort = request.effectiveHostAndPort();
```
daschl, line 73: @idelpivnitskiy is there a more effective way to get the full URL? Also related, should path or requestTarget be used here?

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpServerFilter.java
```diff
@@ -52,21 +59,62 @@
  */
 public final class OpenTelemetryHttpServerFilter extends AbstractOpenTelemetryFilter
```
daschl, line 60: I wonder if we should rename this to `OpenTelemetryHttpServiceFilter` to have proper alignment with the client?

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
```diff
@@ -54,17 +63,42 @@
 public final class OpenTelemetryHttpRequestFilter extends AbstractOpenTelemetryFilter
```
daschl, line 63: I wonder if we should rename this to `OpenTelemetryHttpClientFilter` to have proper alignment with the service?

daschl (changes requested):

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkNetClientAttributesGetter.java
```diff
@@ -0,0 +1,70 @@
...
+      return null;
+    }
+    return "http";
+  }
+
+  @Nullable
+  @Override
+  public String getNetworkProtocolVersion(HttpRequestMetaData request,
+                                          @Nullable HttpResponseMetaData response) {
+    if (response == null) {
```
daschl, line 51: Can we take it from the request which is not nullable?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpProtocolVersion.java
```diff
@@ -149,6 +149,24 @@ public String name() {
         return httpVersion;
     }
 
+    /**
+     * Resolves and return the http version number as a String. e.g. : 1.1 or 2.0.
+     * @return the http version number string.
+     */
+    public String fullVersion() {
+        if (major == 1) {
```
daschl, line 157: Instead of special casing, consider caching `major + '.' + minor` in a local variable in the constructor and then just returning that `fullVersion` string directly.

aleqi200: Good idea. Storing that instead.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpServerFilter.java
```diff
@@ -52,21 +59,62 @@
...
+        instrumenter =
+            serverInstrumenterBuilder.buildServerInstrumenter(RequestHeadersPropagatorGetter.INSTANCE);
     }
 
     /**
      * Create a new Instance, searching for any instance of an opentelemetry available.
      */
     public OpenTelemetryHttpServerFilter() {
-        this(GlobalOpenTelemetry.get());
+        this(GlobalOpenTelemetry.get(), OpentelemetryOptions.newBuilder().build());
```
idelpivnitskiy, line 99: Consider `this(DEFAULT_OPTIONS)`

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestHeadersPropagatorGetter.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright © 2022 Apple Inc. and the ServiceTalk project authors
```
idelpivnitskiy, line 2: 2023

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ScopeTracker.java
```diff
@@ -18,30 +18,36 @@
...
 class ScopeTracker implements TerminalSignalConsumer {
 
     private final Scope currentScope;
-    private final Span span;
+    private final Context context;
+    private final StreamingHttpRequest request;
+    private final Instrumenter<HttpRequestMetaData, HttpResponseMetaData> instrumenter;
 
     @Nullable
     protected HttpResponseMetaData metaData;
```
idelpivnitskiy, line 43: metaData -> responseMetaData, protected -> private

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ScopeTracker.java
```diff
@@ -74,23 +76,12 @@ public void onError(final Throwable throwable) {
     @Override
     public void cancel() {
         try {
-            tagStatusCode();
-            span.setStatus(StatusCode.ERROR);
+            instrumenter.end(context, request, metaData, null);
```
idelpivnitskiy, line 79: I wonder if this case should report with an exception because passing `null` indicates "success", but cancel doesn't mean success in practice.
Consider passing a static instance of `new Exception("cancelled", null, false, false)` here and then check by reference inside `ServicetalkSpanStatusExtractor`.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpClientAttributesGetter.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    public String getHttpRequestMethod(HttpRequestMetaData httpRequestMetaData) {
+        return httpRequestMetaData.method().name();
+    }
+
+    @Override
+    public List<String> getHttpRequestHeader(HttpRequestMetaData httpRequestMetaData, String name) {
+        CharSequence value = httpRequestMetaData.headers().get(name);
```
idelpivnitskiy, line 45: This is incorrect, HTTP allows multiple values for the same header name. You should use `headers().values(name)` instead. 
Please add a test that verifies a case with multiple values.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpClientAttributesGetter.java
```diff
@@ -0,0 +1,82 @@
...
+    @Nullable
+    @Override
+    public String getUrlFull(HttpRequestMetaData request) {
+        HostAndPort effectiveHostAndPort = request.effectiveHostAndPort();
+        String requestScheme = request.scheme() != null ? request.scheme()
+            : "http";
+        CharSequence hostAndPort = request.headers().contains("host") ? request.headers().get("host")
+            : effectiveHostAndPort != null ?
+            String.format("%s:%d", effectiveHostAndPort.hostName(), effectiveHostAndPort.port()) : null;
+        return hostAndPort != null ? String.format("%s://%s%s", requestScheme, hostAndPort, request.path()) :
```
idelpivnitskiy, line 79: The whole method has multiple issues:

1. `request.scheme()` is always `null` today for single-address client. Falling back to `http` is not correct. Please open a GH issue, and we can try to prefill scheme based on `SslConfig` configured at the client builder.
2. There is no need to check `host` header if you already parsed `effectiveHostAndPort`. Look at its javadoc, it already takes the `host` header into account. In cases when you need a header, use constants instead of hard coded header names.
3. Usage of `String.format` is strongly discouraged, it has the worst performance. Consider `+` or `StringBuilder` as alternatives.
4. `request.path()` is not enough, javadoc for this method described that you need everything: path, query parameter, fragment. 

For now, consider only returning `request.requestTarget()`. It will contain everything you need and will give a full URL if users passed it in absolute-form. Then we can improve after (1) is resolved.
In addition, consider implementing `UrlAttributesExtractor`, it will give users path, query, etc.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpServerAttributesGetter.java
```diff
@@ -0,0 +1,89 @@
...
+        CharSequence value = httpResponseMetaData.headers().get(name);
+        if (value != null) {
+            return Collections.singletonList(value.toString());
+        }
+        return Collections.emptyList();
+    }
+
+    @Override
+    public String getHttpRoute(HttpRequestMetaData httpRequestMetaData) {
+        return httpRequestMetaData.path();
```
idelpivnitskiy, line 87: Javadoc for `SemanticAttributes.HTTP_ROUTE` says:

> MUST NOT be populated when this is not supported by the HTTP server framework as the route attribute should have low-cardinality and the URI path can NOT substitute it. SHOULD include the application root if there is one.
 
Returning `path()` here is not correct, don't implement this method.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpServerAttributesGetter.java
```diff
@@ -0,0 +1,89 @@
...
+                                             HttpResponseMetaData httpResponseMetaData,
+                                             @Nullable Throwable error) {
+        return httpResponseMetaData.status().code();
+    }
+
+    @Override
+    public List<String> getHttpResponseHeader(HttpRequestMetaData httpRequestMetaData,
+                                              HttpResponseMetaData httpResponseMetaData,
+                                              String name) {
+        CharSequence value = httpResponseMetaData.headers().get(name);
```
idelpivnitskiy, line 78: Same comments as for request headers

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkNetServerAttributesGetter.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private ServicetalkNetServerAttributesGetter() {
+  }
+
+  @Nullable
+  @Override
+  public String getNetworkProtocolName(HttpRequestMetaData request, @Nullable HttpResponseMetaData response) {
+
+    if (response == null) {
+      return null;
```
idelpivnitskiy, line 45: No need to check for `null` here as we know ST can communicate only over `http`

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkNetServerAttributesGetter.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+    return "http";
+  }
+
+  @Override
+  public String getNetworkProtocolVersion(HttpRequestMetaData request,
+                                          @Nullable HttpResponseMetaData response) {
+    HttpProtocolVersion version = request.version();
+    return version.fullVersion();
```
idelpivnitskiy, line 54: Prefer a version from `response` if it's not null

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkSpanStatusExtractor.java
```diff
@@ -0,0 +1,57 @@
...
+            spanStatusBuilder.setStatus(StatusCode.ERROR);
+        } else if (status != null) {
+            switch (status.status().statusClass()) {
+                case INFORMATIONAL_1XX:
+                case SUCCESSFUL_2XX:
+                case REDIRECTION_3XX:
+                    spanStatusBuilder.setStatus(StatusCode.OK);
+                    break;
+                default:
+                    spanStatusBuilder.setStatus(StatusCode.ERROR);
```
idelpivnitskiy, line 51: Consider using the opposite strategy: mark as `ERROR` for 4xx and 5xx, keep as `OK` for everything else as we don't know what `UNKNOWN` means

idelpivnitskiy (changes requested):

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkNetServerAttributesGetter.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  @Override
+  @Nullable
+  public String getServerAddress(HttpRequestMetaData request) {
+    HostAndPort effectiveHostAndPort = request.effectiveHostAndPort();
```
idelpivnitskiy, line 60: 2 spaces is not enough, checkstyle will complain.
make sure that `./gradlew build` passes locally before pushing more commits to this branch.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
```diff
@@ -73,15 +107,35 @@ public OpenTelemetryHttpRequestFilter(final OpenTelemetry openTelemetry, String
...
+        this(GlobalOpenTelemetry.get(), componentName, OpentelemetryOptions.newBuilder().build());
+    }
+
+    /**
+     * Create a new instance, searching for any instance of an opentelemetry available.
+     *
+     * @param openTelemetry the {@link OpenTelemetry}.
+     * @param componentName The component name used during building new spans.
+     */
+    public OpenTelemetryHttpRequestFilter(final OpenTelemetry openTelemetry, String componentName) {
```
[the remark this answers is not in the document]
aleqi200: ok, deprecated the constructors with the Opentelemetry.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpServerFilter.java
```diff
@@ -52,21 +59,62 @@
...
+    }
+
+    /**
+     * Create a new instance.
+     *
+     * @param opentelemetryOptions extra options to create the opentelemetry filter
+     */
+    public OpenTelemetryHttpServerFilter(OpentelemetryOptions opentelemetryOptions) {
```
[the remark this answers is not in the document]
aleqi200: Deprecated them.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpServerFilter.java
aleqi200: ok, will do in a follow up PR

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkNetClientAttributesGetter.java
aleqi200: ok, using the response if not null and request otherwise as a fallback.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
```diff
@@ -73,15 +107,35 @@ public OpenTelemetryHttpRequestFilter(final OpenTelemetry openTelemetry, String
...
+        this(openTelemetry, componentName, OpentelemetryOptions.newBuilder().build());
```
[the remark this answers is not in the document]
aleqi200: done.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpServerFilter.java
aleqi200: done

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpServerAttributesGetter.java
aleqi200: true, looking at the path it should contain the path params and that's not possible to find.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkNetServerAttributesGetter.java
aleqi200: ok, using request as a fallback only.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkNetServerAttributesGetter.java
aleqi200: using a constant instead.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpServerAttributesGetter.java
aleqi200: ok, refactored this.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpClientAttributesGetter.java
aleqi200: ok, I refactored this a little.
I created the issue https://github.com/apple/servicetalk/issues/2674

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpClientAttributesGetter.java
```diff
@@ -0,0 +1,82 @@
...
+                                             HttpResponseMetaData httpResponseMetaData,
+                                             @Nullable Throwable error) {
+        return httpResponseMetaData.status().code();
+    }
+
+    @Override
+    public List<String> getHttpResponseHeader(HttpRequestMetaData httpRequestMetaData,
+                                              HttpResponseMetaData httpResponseMetaData,
+                                              String name) {
+        CharSequence value = httpResponseMetaData.headers().get(name);
```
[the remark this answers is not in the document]
aleqi200: ok, moved this to a common class as utility method.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpClientAttributesGetter.java
aleqi200: ok, refactored this.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ScopeTracker.java
aleqi200: ok, I created an exception but the exception is enough as the ```ServicetalkSpanStatusExtractor ``` will mark the span as error.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ScopeTracker.java
aleqi200: updated.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ScopeTracker.java
```diff
@@ -18,30 +18,36 @@
...
 
-import static io.servicetalk.http.api.HttpResponseStatus.StatusClass.SERVER_ERROR_5XX;
 import static java.util.Objects.requireNonNull;
 
 class ScopeTracker implements TerminalSignalConsumer {
 
     private final Scope currentScope;
-    private final Span span;
+    private final Context context;
+    private final StreamingHttpRequest request;
```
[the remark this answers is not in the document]
aleqi200: renamed.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestHeadersPropagatorGetter.java
aleqi200: fixed.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkHttpServerAttributesGetter.java
```diff
@@ -0,0 +1,75 @@
...
+
+import io.servicetalk.http.api.HttpRequestMetaData;
+import io.servicetalk.http.api.HttpResponseMetaData;
+
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpServerAttributesGetter;
+
+import java.util.List;
+import javax.annotation.Nullable;
+
+final class ServicetalkHttpServerAttributesGetter
```
idelpivnitskiy: Merged into a single `ServicetalkHttpAttributesGetter` class to reduce duplication for common methods

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServicetalkNetServerAttributesGetter.java
```diff
@@ -0,0 +1,66 @@
...
+
+import io.opentelemetry.instrumentation.api.instrumenter.net.NetServerAttributesGetter;
+
+import javax.annotation.Nullable;
+
+/**
+ * This class is internal and is hence not for public use. Its APIs are unstable and can change at
+ * any time.
+ */
+final class ServicetalkNetServerAttributesGetter
```
idelpivnitskiy: Merged into a single `ServicetalkNetAttributesGetter` class to reduce duplication for common methods

idelpivnitskiy (commented):
Pushed 2 commits to address the last comments I have below:

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequestFilter.java
```diff
@@ -54,17 +63,43 @@
...
      * Create a new instance.
      *
-     * @param openTelemetry the {@link OpenTelemetry}.
-     * @param componentName The component name used during building new spans.
+     * @param openTelemetry        the {@link OpenTelemetry}.
+     * @param componentName        The component name used during building new spans.
+     * @param opentelemetryOptions extra options to create the opentelemetry filter.
      */
-    public OpenTelemetryHttpRequestFilter(final OpenTelemetry openTelemetry, String componentName) {
+    OpenTelemetryHttpRequestFilter(final OpenTelemetry openTelemetry, String componentName,
```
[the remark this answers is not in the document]
aleqi200: Preserving history makes sense.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/HeadersPropagatorGetter.java
```diff
@@ -67,4 +70,11 @@ public String get(@Nullable HttpHeaders carrier, final String key) {
         final CharSequence value = carrier.get(key);
         return value == null ? null : value.toString();
     }
+
+    static List<String> getHeadersValue(String name, HttpHeaders headers) {
+        Iterable<? extends CharSequence> value = headers.values(name);
+        return StreamSupport.stream(value.spliterator(), false)
```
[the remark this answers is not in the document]
aleqi200: I was not sure on the best approach there, I wanted to avoid streams but I was skeptical on the amount of code necessary now I can see the benefits outweighs the short code.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/OpentelemetryOptions.java
```diff
@@ -0,0 +1,101 @@
...
+    private final boolean enableMetrics;
+
+    OpentelemetryOptions(List<String> captureRequestHeaders, List<String> captureResponseHeaders,
+                         boolean enableMetrics) {
+        this.captureRequestHeaders = captureRequestHeaders;
+        this.captureResponseHeaders = captureResponseHeaders;
+        this.enableMetrics = enableMetrics;
+    }
+
+    public static Builder newBuilder() {
```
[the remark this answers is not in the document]
aleqi200: I was not sure about this. It makes sense.

aleqi200:
the changes look good. I made a test and it looks good.

idelpivnitskiy (approved):
Thank you!

PR #2664, opened 2023-07-28
DefaultNettyConnection.initChannel() does not work on channels that already completed a SSL handshake.

thomdev:
###  Motivation
The value of `NettyToStChannelHandler.waitForSslHandshake` is coming from `NettyPipelineSslUtils.isSslEnabled()`. When it is true, `NettyToStChannelHandler` relies on the `SslHandshakeCompletionEvent` user event, which has been sent already, to notify the subscriber. This makes it impossible to use `DefaultNettyConnection.initChannel()` to “upgrade” channels that have a `SslHandler` but already completed the handshake.

### Modifications
- Add new `NettyPipelineSslUtils.hasPendingSslHandshake()` helper function.
- `DefaultNettyConnection.initChannel()` uses `NettyPipelineSslUtils.hasPendingSslHandshake()` instead of `NettyPipelineSslUtils.isSslEnabled()`.

### Result
`DefaultNettyConnection` works with `Channel`s that already completed a SSL handshake.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
@@ -18,10 +18,12 @@ package io.servicetalk.transport.netty.internal;
 import io.servicetalk.transport.api.ConnectionObserver.SecurityHandshakeObserver;
 import io.servicetalk.transport.netty.internal.ConnectionObserverInitializer.ConnectionObserverHandler;
 
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelPipeline;
 import io.netty.handler.ssl.SniHandler;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.ssl.SslHandshakeCompletionEvent;
+import io.netty.util.concurrent.Future;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
```diff
@@ -45,11 +48,40 @@ private NettyPipelineSslUtils() {
...
+            final Future<Channel> handshakeFuture = sslHandler.handshakeFuture();
+            if (handshakeFuture.isDone()) {
+                if (handshakeFuture.cause() != null) {
+                    if (handshakeFuture.cause() instanceof SSLException) {
+                        throw (SSLException) handshakeFuture.cause();
+                    } else {
+                        throw new SSLException(handshakeFuture.cause());
+                    }
+                }
+                return false;
```
idelpivnitskiy, line 78:
```
AlpnClientAndServerTest.serverAlpnProtocols=[h2, http/1.1], clientAlpnProtocols=[h2, http/1.1], expectedProtocol=HTTP/2.0, expectedExceptionType=null
java.lang.AssertionError: 
Expected: is not null
     but: was null
	at io.servicetalk.http.netty.AlpnClientAndServerTest.testAlpnConnection(AlpnClientAndServerTest.java:186)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```
ALPN is exactly the case when handshake is already complete by the time we add the last handler into pipeline. As the result, we do not extract the SSLSession. I don't remember how to extract the session after handshake. Maybe through `SslHandler#engine()`.

idelpivnitskiy: Yes, exactly, we extract it via `sslHandler.engine().getSession()`

idelpivnitskiy:
Fixed in https://github.com/apple/servicetalk/pull/2672

PR #2666, opened 2023-08-02
RetryingHttpRequesterFilter use original Publisher avoid SOOE

Scottmitch:
Motivation:
RetryingHttpRequesterFilter uses the request across retry operations. The request's Publisher state is shared across retries and operators applied via transformMessageBody will therefore be reapplied on each retry. This typically isn't desirable from a semantics perspective but may also lead to StackOverflowException if enough retries are attempted.

Modifications:
- RetryingHttpRequesterFilter should reuse the original Publisher for request payload body on each retry attempt.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
@@ -1717,7 +1717,7 @@ public abstract class Single<T> {
      * @param context The {@link ContextMap} to use for {@link AsyncContext} when subscribed.
      * @return A {@link Single} that will use the {@link ContextMap} for {@link AsyncContext} when subscribed.
      */
-    public final Single<T> shareContextOnSubscribe(ContextMap context) {
+    public final Single<T> setContextOnSubscribe(ContextMap context) {
         return new SingleSetContextOnSubscribe<>(this, context);
     }
 
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
@@ -1710,6 +1710,18 @@ public abstract class Single<T> {
     }
 
     /**
+     * Specify the {@link ContextMap} to use for {@link AsyncContext} when the returned {@link Single} is subscribed to.
+     * <p>
+     * This operator only impacts behavior if the returned {@link Single} is subscribed directly after this operator,
+     * that means this must be the "last operator" in the chain for this to have an impact.
+     * @param context The {@link ContextMap} to use for {@link AsyncContext} when subscribed.
+     * @return A {@link Single} that will use the {@link ContextMap} for {@link AsyncContext} when subscribed.
+     */
+    public final Single<T> shareContextOnSubscribe(ContextMap context) {
+        return new SingleSetContextOnSubscribe<>(this, context);
+    }
+
+    /**
      * <strong>This method requires advanced knowledge of building operators. Before using this method please attempt
      * to compose existing operator(s) to satisfy your use case.</strong>
      * <p>
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingToStreamingService.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingToStreamingService.java
@@ -37,7 +37,7 @@ final class BlockingToStreamingService extends AbstractServiceAdapterHolder {
                                                 final StreamingHttpRequest request,
                                                 final StreamingHttpResponseFactory responseFactory) {
         return request.toRequest().flatMap(req -> fromCallable(() -> original.handle(
-                ctx, req, ctx.responseFactory())).map(HttpResponse::toStreamingResponse));
+                ctx, req, ctx.responseFactory())).map(HttpResponse::toStreamingResponse).shareContextOnSubscribe());
     }
 
     @Override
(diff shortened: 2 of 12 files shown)


idelpivnitskiy (commented):
Good idea, it should work!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -226,9 +226,14 @@ public Single<? extends FilterableReservedStreamingHttpConnection> reserveConnec
...
                                                         final StreamingHttpRequest request) {
-            final StreamingHttpRequest duplicatedRequest = mayReplayRequestPayload ?
-                    request.transformMessageBody(messageBodyDuplicator()) : request;
-            Single<StreamingHttpResponse> single = delegate.request(duplicatedRequest);
+            // State intentionally outside the defer because the request state is shared across subscribes. If
+            // re-applying operators duplicates logic that isn't desirable and lead to StackOverflowException.
+            final Publisher<Object> originalMessageBody = request.messageBody();
+            Single<StreamingHttpResponse> single = Single.defer(() -> (mayReplayRequestPayload ?
+                    delegate.request(request.transformMessageBody(messageBodyDuplicator(originalMessageBody))) :
+                    delegate.request(request.transformMessageBody(p -> originalMessageBody))
```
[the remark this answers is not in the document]
Scottmitch: > When mayReplayRequestPayload == false, we don't need to do anything with originalMessageBody. 

The unit test demonstrates the problem exists regardless of `mayReplayRequestPayload`. In either case if we don't apply the original publisher we will keep appending operators and eventually SOOE.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RedoPublisher.java
```diff
@@ -49,9 +49,9 @@ final class RedoPublisher<T> extends AbstractNoHandleSubscribePublisher<T> {
     void handleSubscribe(Subscriber<? super T> subscriber, ContextMap contextMap,
                          AsyncContextProvider contextProvider) {
         // For the current subscribe operation we want to use contextMap directly, but in the event a re-subscribe
-        // operation occurs we want to restore the original state of the AsyncContext map, so we save a copy upfront.
```
Scottmitch, line 52: the copy isn't done consistently after reloading changes were done a while back. I'm also not sure we need the copy any more so I made them consistent.

Scottmitch: moved to https://github.com/apple/servicetalk/pull/2670

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -226,9 +229,20 @@ public Single<? extends FilterableReservedStreamingHttpConnection> reserveConnec
...
+            final Publisher<Object> originalMessageBody = request.messageBody();
+            final AtomicReference<ContextMap> contextRef = new AtomicReference<>();
+            Single<StreamingHttpResponse> single = Single.defer(() -> {
+                final Single<StreamingHttpResponse> reqSingle = delegate.request(
+                        request.transformMessageBody(mayReplayRequestPayload ?
+                                messageBodyDuplicator(originalMessageBody) : p -> originalMessageBody));
+                final ContextMap map = contextRef.get();
+                return map == null && contextRef.compareAndSet(null, AsyncContext.context()) ?
+                        reqSingle.shareContextOnSubscribe() :
+                        reqSingle.shareContextOnSubscribe(contextRef.get());
```
bryce-anderson, line 243: ffti: I think the point is that on follow-up requests we want to reuse the same context entry but why is non-obvious. If there is a short comment about this part it might be helpful, or as likely, a reference to an existing comment somewhere else.

bryce-anderson (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1709,6 +1709,18 @@ public final Single<T> shareContextOnSubscribe() {
...
+     * Specify the {@link ContextMap} to use for {@link AsyncContext} when the returned {@link Single} is subscribed to.
+     * <p>
+     * This operator only impacts behavior if the returned {@link Single} is subscribed directly after this operator,
+     * that means this must be the "last operator" in the chain for this to have an impact.
+     * @param context The {@link ContextMap} to use for {@link AsyncContext} when subscribed.
+     * @return A {@link Single} that will use the {@link ContextMap} for {@link AsyncContext} when subscribed.
+     */
+    public final Single<T> shareContextOnSubscribe(ContextMap context) {
+        return new SingleSetContextOnSubscribe<>(this, context);
+    }
```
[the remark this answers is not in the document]
Scottmitch: yes sounds good, I was debating the same

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingToStreamingService.java
```diff
@@ -37,7 +37,7 @@ public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
                                                 final StreamingHttpRequest request,
                                                 final StreamingHttpResponseFactory responseFactory) {
         return request.toRequest().flatMap(req -> fromCallable(() -> original.handle(
-                ctx, req, ctx.responseFactory())).map(HttpResponse::toStreamingResponse));
+                ctx, req, ctx.responseFactory())).map(HttpResponse::toStreamingResponse).shareContextOnSubscribe());
```
idelpivnitskiy, line 40: Good catch!

idelpivnitskiy (commented):
LGTM

PR #2670, opened 2023-08-09
Retry/repeat operators consistently not copy AsyncContext

Scottmitch:
Motivation:
Some of the retry/repeat operators copy AsyncContext on subscribe and others doen't. We should consistently not copy in these operators and if copy is desired it can be implemented externally via another operator.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RedoPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RedoPublisher.java
@@ -48,8 +48,10 @@ final class RedoPublisher<T> extends AbstractNoHandleSubscribePublisher<T> {
     @Override
     void handleSubscribe(Subscriber<? super T> subscriber, ContextMap contextMap,
                          AsyncContextProvider contextProvider) {
-        // For the current subscribe operation we want to use contextMap directly, but in the event a re-subscribe
-        // operation occurs we want to restore the original state of the AsyncContext map.
+        // Current expected behavior is to capture the context on the first subscribe, save it, and re-use it on each
+        // resubscribe. This allows for async context to be shared across each request retry, and follows the same
+        // shared state model as the request object on the client. If copy-on-each-resubscribe is desired this could
+        // be provided by an independent operator, or manually cleared/overwritten.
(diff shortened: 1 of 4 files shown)


daschl (commented):
Change itself looks good to me, but for my understanding: is there any potential impact for current users, and also does this affect the behavior of a `shareContextOnSubscribe` in some way?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/RedoPublisher.java
```diff
@@ -49,9 +49,9 @@ final class RedoPublisher<T> extends AbstractNoHandleSubscribePublisher<T> {
     void handleSubscribe(Subscriber<? super T> subscriber, ContextMap contextMap,
                          AsyncContextProvider contextProvider) {
         // For the current subscribe operation we want to use contextMap directly, but in the event a re-subscribe
-        // operation occurs we want to restore the original state of the AsyncContext map, so we save a copy upfront.
+        // operation occurs we want to restore the original state of the AsyncContext map.
```
daschl, line 52: Is this comment now still accurate? (original state will not be preserved anymore if a re-subscribe occurs, or?)

Scottmitch: I clarified these comments.

Scottmitch:
> is there any potential impact for current users

The first retry will now see the original context map, in most/all cases the retry logic will repopulate the map which shouldn't be an issue. However if folks are checking "do I have existing state in the AsyncContext map" they may see state on the first retry that wasn't there previously, however note that the state would have already been shared on the 2nd (and later) retries so the current behavior isn't consistent and likely can't be relied upon.

I don't think the current behavior makes much sense. The retry operator making a copy means that on retry the expected context may not be restored (bcz the original is passed upstream), and only making a copy the first time means behavior is inconsistent after subsequent retries.

> also does this affect the behavior of a shareContextOnSubscribe in some way?

No I don't think so. That operator is still independent. I debated making another operator that is something like "copyContextOnSubscribe" that can make a copy anywhere in the operator chain (by default copy is made at the the "bottom" right when subscribe happens and passed up the chain). However I think we can wait until its necessary.

idelpivnitskiy (commented):
Thank you!
LGTM

PR #2671, opened 2023-08-11
Clean up discarded response message content on the service side

daschl:


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardCleanerServiceFilter.java
```diff
@@ -0,0 +1,131 @@
...
+            @Override
+            public HttpResponseObserver onResponse(final HttpResponseMetaData responseMetaData) {
+                this.responseMetaData = responseMetaData;
+                return NoopHttpLifecycleObserver.NoopHttpResponseObserver.INSTANCE;
+            }
+
+            @Override
+            public void onExchangeFinally() {
+                if (requestMetaData != null && responseMetaData != null) {
+                    Boolean subscribed = requestMetaData.context().get(payloadSubscribedKey);
```
Scottmitch, line 107: if a retry occurs the request context map state will be shared, do we reset this state anywhere?

Scottmitch: ah server side retries are less likely but in principal could be done by inserting a filter that has a fallback to in-memory publisher if failure occurs.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardCleanerServiceFilter.java
```diff
@@ -0,0 +1,136 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+
+import static io.servicetalk.http.netty.HttpPayloadDiscardWatchdogServiceFilter.payloadPublisherKey;
+import static io.servicetalk.http.netty.HttpPayloadDiscardWatchdogServiceFilter.payloadSubscribedKey;
+
+final class HttpPayloadDiscardCleanerServiceFilter implements StreamingHttpServiceFilterFactory, HttpLifecycleObserver {
```
idelpivnitskiy, line 43: We should not implement 2 unrelated interfaces at the same time, it opens a risk for misuse. 
Instead of creating a new `StreamingHttpServiceFilterFactory`, consider implementing only `HttpLifecycleObserver` and use a cached constant of `HttpLifecycleObserverServiceFilter`. It will save some boilerplate code around `create` and `requireOffloads` overrides.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardCleanerServiceFilter.java
```diff
@@ -0,0 +1,136 @@
...
+
+            @Override
+            public void onResponseError(final Throwable cause) {
+            }
+
+            @Override
+            public void onResponseCancel() {
+            }
+
+            private void tryCleanPayload() {
```
idelpivnitskiy, line 131: Unused method

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardCleanerServiceFilter.java
```diff
@@ -0,0 +1,136 @@
...
+    public HttpExchangeObserver onNewExchange() {
+
+        return new HttpExchangeObserver() {
+
+            @Nullable
+            private HttpRequestMetaData requestMetaData;
+
+            @Override
+            public HttpRequestObserver onRequest(final HttpRequestMetaData requestMetaData) {
+                this.requestMetaData = requestMetaData;
```
idelpivnitskiy, line 92: Consider saving only `request.context()`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardWatchdogServiceFilter.java
```diff
@@ -0,0 +1,84 @@
...
+                        .map(response -> {
+                            if (!HttpApiConversions.isPayloadEmpty(response)) {
+                                // If the payload is not empty, always write the buffer publisher into the
+                                // request context. When a downstream subscriber arrives, mark the payload
+                                // as subscribed explicitly (having a payload present and no subscription is
+                                // an indicator that it must be freed later on).
+                                request.context().put(payloadPublisherKey, response.payloadBody());
+                                return response.transformPayloadBody(bufferPublisher ->
+                                        bufferPublisher.beforeOnSubscribe(subscription ->
+                                                request.context().put(payloadSubscribedKey, true)));
```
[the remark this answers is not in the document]
idelpivnitskiy: Alternatively, consider using `AtomitReference` for the publisher and set it back to `null` on subscribe. That way we need only a single key for context map.
This will also help with retry case mentioned by Scott. If we see `response.map` execution again, while reference is not null, it's an indicator that we have to drain the previous publisher too.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardCleanerServiceFilter.java
```diff
@@ -0,0 +1,136 @@
...
+            public void onExchangeFinally() {
+                if (requestMetaData != null) {
+                    final ContextMap requestContext = requestMetaData.context();
+                    if (requestContext.get(payloadSubscribedKey) == null) {
+                        // No-one subscribed to the payload (or there is none), so if there is a payload
+                        // proactively clean it up.
+                        Publisher<?> payload = requestContext.get(payloadPublisherKey);
+                        if (payload != null) {
+                            LOGGER.debug("Proactively cleaning up HTTP response payload which has been dropped - " +
+                                            "this is a strong indication of a bug in a filter. Request: {}",
```
idelpivnitskiy, line 111: I would probably use `warn` here or maybe even `error`. With `debug`, nobody will see it. 
However, there is a risk to flood logs if we do it for every request. Consider logging only once, by remembering the "logged" flag in a static variable. No need to use volatile, it's ok it if logs a few more times before it becomes visible to other threads. 

Also, consider enhancing the message to clarify that it's in a custom user-defined filter, to make sure they don't think about ST filters.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpServicePayloadDiscardWatchdogTest.java
```diff
@@ -0,0 +1,104 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+final class HttpServicePayloadDiscardWatchdogTest {
+
+    @Test
+    void cleansPayloadBodyIfDiscardedInFilter() throws Exception {
+        final AtomicBoolean payloadSubscribed = new AtomicBoolean(false);
+        try (HttpServerContext serverContext = HttpServers
+                .forPort(0)
```
idelpivnitskiy, line 45: Prefer using test utilities, either `AddressUtils.localAddress(0)` or even better to use `BuilderUtils.newServerBuilder`. 
We saw issues when tests use `forPort(0)` when they run on a system (CI) where wildcard address is blocked.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpServicePayloadDiscardWatchdogTest.java
```diff
@@ -0,0 +1,104 @@
...
+                .appendServiceFilter(service -> new StreamingHttpServiceFilter(service) {
+                    @Override
+                    public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
+                                                                final StreamingHttpRequest request,
+                                                                final StreamingHttpResponseFactory responseFactory) {
+                        return delegate()
+                                .handle(ctx, request, responseFactory)
+                                .beforeOnSuccess(res -> res.transformPayloadBody(payload -> Publisher.empty()));
```
idelpivnitskiy, line 53: Consider parametrizing the test to make sure our detection logic captures all overloads of `transform*` function

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardWatchdogServiceFilter.java
```diff
@@ -0,0 +1,84 @@
...
+                        .handle(ctx, request, responseFactory)
+                        .map(response -> {
+                            if (!HttpApiConversions.isPayloadEmpty(response)) {
+                                // If the payload is not empty, always write the buffer publisher into the
+                                // request context. When a downstream subscriber arrives, mark the payload
+                                // as subscribed explicitly (having a payload present and no subscription is
+                                // an indicator that it must be freed later on).
+                                request.context().put(payloadPublisherKey, response.payloadBody());
+                                return response.transformPayloadBody(bufferPublisher ->
+                                        bufferPublisher.beforeOnSubscribe(subscription ->
```
idelpivnitskiy, line 71: `beforeOnSubscribe` is a little late. `onSubscribed` is invoked by the publisher in response to `Publisher.subscribe` invocation from the `Subscriber`. Instead, consider using `beforeSubscriber`. Do the logic insider `Supplier`, and returns a noop Subscriber.

Ideally, we should add `beforeSubscribe(Runnable)` operator. We already discussed this with Scott, I need it for some other use-case. If you'd like to work on a new operator, I can guide you. Otherwise, just leave a todo comment and we can improve this when the new operator is available.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpServicePayloadDiscardWatchdogTest.java
```diff
@@ -0,0 +1,104 @@
...
+                    }
+                })
+                .listenStreamingAndAwait((ctx, request, responseFactory) -> {
+                    final Publisher<Buffer> buffer = Publisher
+                            .from(ctx.executionContext().bufferAllocator().fromUtf8("Hello, World!"))
+                            .beforeOnSubscribe(subscription -> payloadSubscribed.set(true));
+                    return Single.succeeded(responseFactory.ok().payloadBody(buffer));
+                })) {
+            try (BlockingHttpClient client = HttpClients.forSingleAddress(
+                    HostAndPort.of((InetSocketAddress) serverContext.listenAddress())).buildBlocking()) {
```
idelpivnitskiy, line 63: See `AddressUtils.serverHostAndPort` and/or `BuilderUtils.newClientBuilder`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpServicePayloadDiscardWatchdogTest.java
```diff
@@ -0,0 +1,104 @@
...
+                .forPort(0)
+                .appendServiceFilter(service -> new StreamingHttpServiceFilter(service) {
+                    @Override
+                    public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
+                                                                final StreamingHttpRequest request,
+                                                                final StreamingHttpResponseFactory responseFactory) {
+                        return delegate()
+                                .handle(ctx, request, responseFactory)
+                                .map(res -> {
+                                    throw new IllegalStateException("I don't like this response!");
```
idelpivnitskiy, line 85: Consider using `DeliberateException` to indicate it's expected

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpServicePayloadDiscardWatchdogTest.java
```diff
@@ -0,0 +1,104 @@
...
+                HttpResponse response = client.request(client.get("/"));
+                assertEquals(0, response.payloadBody().readableBytes());
+            }
+
+            assertTrue(payloadSubscribed.get());
+        }
+    }
+
+    @Test
+    void cleansPayloadBodyIfErrorInFilter() throws Exception {
```
idelpivnitskiy, line 73: Consider adding a test that uses `map` to drop the original response and create a new one

idelpivnitskiy (changes requested):
Approach lgtm, my comments only for details of the impl:

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardCleanerServiceFilter.java
daschl: Caching the whole metadata helps with debug logging the surrounding context - do you think it's worth the optimization still?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpServicePayloadDiscardWatchdogTest.java
daschl: Added transformMessageBody and transformPayloadBody drops, do we also need to consider the vanilla transform for the trailers?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardWatchdogServiceFilter.java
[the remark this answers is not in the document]
daschl: Added a test

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardCleanerServiceFilter.java
daschl: Added a test in the latest changeset.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardCleanerServiceFilter.java
idelpivnitskiy: I saw that you used it for logging and forgot to update my comment. Even though we filter headers by default,  potential risk there is that you can log sensitive request-target if path contains auth token.
It shouldn't matter what request caused it bcz if users have a problem, it's likely applicable for every request. And if they want more details, they can add correlation key in MDC.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpServicePayloadDiscardWatchdogTest.java
idelpivnitskiy: No, we are good with only these 2. Transform for trailers does not let users drop old payload publisher. Thanks!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogServiceFilter.java
```diff
@@ -0,0 +1,127 @@
...
+                            // arrives, mark the message as subscribed explicitly (having a message present and no
+                            // subscription is an indicator that it must be freed later on).
+                            final AtomicReference<?> previous = request.context()
+                                    .put(MESSAGE_PUBLISHER_KEY, new AtomicReference<>(response.messageBody()));
+                            if (previous != null) {
+                                // If a previous message exists, the response publisher got resubscribed to (i.e.
+                                // during a retry) and so also needs to be cleaned up.
+                                Publisher<?> message = (Publisher<?>) previous.get();
+                                if (message != null) {
+                                    LOGGER.debug("Cleaning up HTTP response message which has been resubscribed to - " +
```
idelpivnitskiy, line 81: Recommended rewording above and promoted to `warn` level.

In this case, we probably don't need to worry about "logging once" because (1) it's not expected that users will ever have retries configured and (2) retries are not expected to happen all the time to flood logs. wdyt?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardCleanerServiceLifecycleObserver.java
```diff
@@ -0,0 +1,118 @@
...
+                    final AtomicReference<?> maybePublisher = requestContext.get(MESSAGE_PUBLISHER_KEY);
+                    if (maybePublisher != null) {
+                        Publisher<?> message = (Publisher<?>) maybePublisher.get();
+                        if (message != null) {
+                            // No-one subscribed to the message (or there is none), so if there is a message
+                            // proactively clean it up.
+                            if (!loggedError) {
+                                LOGGER.error("Proactively cleaning up HTTP response message which has been " +
+                                                "dropped - this is a strong indication of a bug in a user-defined " +
+                                                "filter. Request: {}",
```
idelpivnitskiy: Consider:
```
Automatically draining HTTP response message body which has been dropped by user code -  this is a strong indication of a bug in a user-defined filter. Responses (or their message body) must be fully consumed before discarding.
```
No need to log context too, you never know what's there.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogServiceTest.java
```diff
@@ -0,0 +1,170 @@
...
+                    return responseFactory.ok().payloadBody(buffer);
+                }))) {
+
+            try (BlockingHttpClient client = newClientBuilder(serverContext, CLIENT_CTX)
+                    .buildBlocking()) {
+                HttpResponse response = client.request(client.get("/"));
+                assertEquals(0, response.payloadBody().readableBytes());
+            }
+
+            assertTrue(payloadSubscriptionCounter.await(5, TimeUnit.SECONDS));
```
idelpivnitskiy, line 87: just fyi: not necessary to explicitly set timeouts because every test has a default timeout of 10 seconds (30 when running on CI)

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpPayloadDiscardCleanerServiceFilter.java
daschl: Ok thanks for the clarification - updated

idelpivnitskiy (commented):
LGTM after the last comments:

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogServiceFilter.java
daschl: Agreed - if we see it spamming we can always bring in the check here too.

idelpivnitskiy (approved):
Thank you!

PR #2672, opened 2023-08-14
Don't wait for handshake if it's already done when adding last handlers

idelpivnitskiy:
Motivation:

Currently, `waitForSslHandshake` flag is defined by presence of `SslHandler` in the pipeline.
However, there could be a use-case when `DefaultNettyConnection` is added after handshake
is already complete and there is no need to wait for it anymore.

Modifications:
- Deprecate `NettyPipelineSslUtils.isSslEnabled(...)`, instead introduce
`NettyPipelineSslUtils.extractSslSessionAndReport(...)` overload that can extract `SSLSession`
if the handshake is already complete and avoid `waitForSslHandshake`;
- Use the new utility in `DefaultNettyConnection`, `H2ClientParentConnectionContext`,
and `H2ServerParentConnectionContext`;
- Enhance tests to make sure behavior for all use-cases does not change;

Result:

The last handler in the pipeline does not wait for handshake if it's already done.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
@@ -18,16 +18,20 @@ package io.servicetalk.transport.netty.internal;
 import io.servicetalk.transport.api.ConnectionObserver.SecurityHandshakeObserver;
 import io.servicetalk.transport.netty.internal.ConnectionObserverInitializer.ConnectionObserverHandler;
 
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelPipeline;
 import io.netty.handler.ssl.SniHandler;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.ssl.SslHandshakeCompletionEvent;
+import io.netty.util.concurrent.Future;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.function.Consumer;
 import javax.annotation.Nullable;
 import javax.net.ssl.SSLSession;
+
+import static io.servicetalk.utils.internal.ThrowableUtils.throwException;
 
 /**
  * Utilities for {@link ChannelPipeline} and SSL/TLS.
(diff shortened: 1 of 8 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
```diff
@@ -45,11 +52,55 @@ private NettyPipelineSslUtils() {
...
+     * @param pipeline {@link ChannelPipeline} which contains a handler containing the {@link SSLSession}
+     * @param connectionObserver {@link ConnectionObserver} in case the handshake status should be reported
+     * @return The {@link SSLSession} or {@code null} if none can be found
+     */
+    @Nullable
+    public static SSLSession extractSslSessionAndReport(@Nullable final SslConfig sslConfig,
+                                                        final ChannelPipeline pipeline,
+                                                        final ConnectionObserver connectionObserver) {
+        if (sslConfig == null) {
+            assert noSslHandlers(pipeline);
```
daschl, line 78: Consider throwing a more descriptive exception if this happens for better debugging.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
```diff
@@ -45,11 +52,55 @@ private NettyPipelineSslUtils() {
...
     }
 
+    /**
+     * Extracts the {@link SSLSession} from the {@link ChannelPipeline} if the handshake is already done
+     * and reports the result to {@link SecurityHandshakeObserver} if available.
+     *
+     * @param sslConfig {@link SslConfig} if SSL/TLS is expected
+     * @param pipeline {@link ChannelPipeline} which contains a handler containing the {@link SSLSession}
+     * @param connectionObserver {@link ConnectionObserver} in case the handshake status should be reported
+     * @return The {@link SSLSession} or {@code null} if none can be found
```
daschl, line 71: consider adding @throws for IllegalStateException if the handler could not be found

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
```diff
@@ -65,25 +116,34 @@ public static SSLSession extractSslSessionAndReport(ChannelPipeline pipeline,
...
+                deliverFailureCause(failureConsumer, unableToFindSslHandler(), observer);
             }
         } else {
             deliverFailureCause(failureConsumer, sslEvent.cause(), observer);
         }
         return null;
     }
 
+    private static boolean noSslHandlers(final ChannelPipeline pipeline) {
+        return pipeline.get(SslHandler.class) == null && pipeline.get(DeferSslHandler.class) == null &&
```
daschl, line 134: side note: would it make sense to give all of these handlers a common name so that they are easier to locate by name instead of their class type?

idelpivnitskiy: That's a good idea that we can keep for future enhancements. The only one potential problem is that sometimes some name is hardcoded outside of ST control. See `io.netty.handler.ssl.SniHandler.replaceHandler` as an example.

daschl (approved):

PR #2673, opened 2023-08-16
Shuffle DNS records before propagating to `LoadBalancer`

idelpivnitskiy:
Motivation:

If multiple client instances receive identical DNS records set, they will start connecting to the same upstream instances because the collection is always sorter.

Modifications:

- Shuffle the list before propagating to the load balancer;

Result:

More uniform distribution of the initial connections.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -68,6 +68,7 @@ import java.nio.channels.ClosedChannelException;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -846,6 +847,10 @@ private void handleResolveDone0(final Future<DnsAnswer<T>> addressFuture,
                             cancellableForQuery = null;
                         }
                         try {
+                            // Shuffle events to avoid multiple clients connecting to the same upstream IP address if
+                            // they receive identical result set from DNS.
+                            Collections.shuffle(events);
```
idelpivnitskiy, line 852: I was considering adding a builder option to let users opt-out of shuffling, but then I couldn't find a good reason why they may need a sorted list.
Worst case, they can wrap SD interface and apply `map` on the publisher that will sort again.

bryce-anderson (approved):
This looks good. It seems an implementation detail that we have a mutable List which can be sorted in place but if we care to safeguard against that a follow-up would be fine.

idelpivnitskiy:
We make a copy before, so it has no impact on what users do after we emit a list

PR #2676, opened 2023-08-18
Enable Keep-Alive socket option by default

tkountis:
Motivation
There are cases where a TCP socket can be disconnected unexpectedly (terminated without adhering to the protocol), and when there is no interaction with the socket, the staleness will not be detected. Its commonly the case where TCP keep-alive is used as a counter measure for such cases, to force the socket to get into the right term state to reflect the actual condition.

Modifications
Changed ST default to always enable Keep-Alive socket option.

Results
Expectation is to always be able to detect stale sockets in the application space, and force streams attached to these socket to react on that condition.


Scottmitch (approved):

idelpivnitskiy (approved):
Before merging, please add PR description with motivation and behavior impact that we will be able to use for release notes

PR #2678, opened 2023-08-20
Add Publisher.switchMap

Scottmitch:
Motivation:
Publisher.switchMap can be used to flatten an async stream of
publishers while always taking results from the latest publisher
and cancelling the previous Publisher.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherSwitchMap.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherSwitchMap.java
@@ -196,7 +196,7 @@ final class PublisherSwitchMap<T, R> extends AbstractAsynchronousPublisherOperat
         private final class RSubscriber implements Subscriber<R> {
             volatile int state;
             @Nullable
-            private final RSubscriber prevPublisher;
+            private RSubscriber prevPublisher;
             @Nullable
             private Subscription localSubscription;
             @Nullable
@@ -210,7 +210,9 @@ final class PublisherSwitchMap<T, R> extends AbstractAsynchronousPublisherOperat
             public void onSubscribe(Subscription subscription) {
                 localSubscription = requireNonNull(subscription);
                 if (prevPublisher != null) {
-                    prevPublisher.dispose(subscription);
+                    final RSubscriber localPrev = prevPublisher;
+                    prevPublisher = null; // Set the reference to null to avoid memory leak.
+                    localPrev.dispose(subscription);
                 } else {
                     switchTo(subscription);
                 }
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherSwitchMap.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherSwitchMap.java
@@ -137,7 +137,7 @@ final class PublisherSwitchMap<T, R> extends AbstractAsynchronousPublisherOperat
             if (currPublisher != null) {
                 try {
                     if (parent.maxDelayedErrors <= 0) {
-                        currPublisher.dispose(EMPTY_SUBSCRIPTION);
+                        currPublisher.dispose(EMPTY_SUBSCRIPTION, false);
                     }
                 } finally {
                     final Throwable cause = outerErrorUpdateState(t);
@@ -167,7 +167,8 @@ final class PublisherSwitchMap<T, R> extends AbstractAsynchronousPublisherOperat
                 final int cState = currPublisher.state;
                 if (stateUpdater.compareAndSet(currPublisher, cState, setOuterState(cState, OUTER_STATE_ERROR))) {
                     final int innerState = getInnerState(cState);
-                    return parent.maxDelayedErrors <= 0 && innerState != INNER_STATE_ERROR ||
+                    return (parent.maxDelayedErrors <= 0 &&
+                            innerState != INNER_STATE_ERROR && innerState != INNER_STATE_EMITTING) ||
                             (parent.maxDelayedErrors > 0 &&
                             (innerState == INNER_STATE_ERROR || innerState == INNER_STATE_COMPLETE)) ? t : null;
                 }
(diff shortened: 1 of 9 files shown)


servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherFlatMapLatestTest.java
```diff
@@ -0,0 +1,355 @@
...
+            publisher.onComplete();
+            subscriber.awaitOnComplete();
+        }
+    }
+
+    /**
+     * A signal containing the data from a series of {@link Publisher}s connected in a serial fashion.
+     * @param <T> Type of the data.
+     */
+    interface FlatMapLatestSignal<T> {
```
Scottmitch, line 279: This has shown to be useful but will keep internal to test for now.

When a switch happens application logic may need to take special action (e.g. replay initial/aggregated state as first event)

tkountis:
Will spend some time on it, but should we name it [switch](https://reactivex.io/documentation/operators/switch.html) for familiarity?

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherSwitchMapTest.java
```diff
@@ -0,0 +1,380 @@
...
+            publisher.onComplete();
+            subscriber.awaitOnComplete();
+        }
+    }
+
+    /**
+     * A signal containing the data from a series of {@link Publisher}s connected in a serial fashion.
+     * @param <T> Type of the data.
+     */
+    interface SwitchFlatMapSignal<T> {
```
Scottmitch, line 304: This has shown to be useful but will keep internal to test for now.

When a switch happens application logic may need to take special action (e.g. replay initial/aggregated state as first event)

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/PublisherSwitchMapTest.java
```diff
@@ -0,0 +1,429 @@
...
+            publisher.onComplete();
+            subscriber.awaitOnComplete();
+        }
+    }
+
+    /**
+     * A signal containing the data from a series of {@link Publisher}s connected in a serial fashion.
+     * @param <T> Type of the data.
+     */
+    interface SwitchMapSignal<T> {
```
Scottmitch, line 353: This has shown to be useful but will keep internal to test for now.

When a switch happens application logic may need to take special action (e.g. replay initial/aggregated state as first event)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SwitchMapSignal.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.function.Function;
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
+
+/**
+ * A signal containing the data from a series of {@link Publisher}s switched in a serial fashion.
+ * @param <T> Type of the data.
+ * @see Publisher#switchMap(Function)
+ */
+public interface SwitchMapSignal<T> {
```
Scottmitch, line 27: this isn't strictly required for the operator, but some control flows require knowing if this is the first emission from a new publisher (e.g. may update state that is preserved across retry, ..)

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherSwitchMap.java
```diff
@@ -0,0 +1,391 @@
...
+                }
+            }
+        }
+
+        @Nullable
+        private TerminalNotification outerCompleteUpdateState() {
+            assert currPublisher != null;
+            for (;;) {
+                final int cState = currPublisher.state;
+                if (stateUpdater.compareAndSet(currPublisher, cState, setOuterState(cState, OUTER_STATE_COMPLETE))) {
```
bryce-anderson, line 167: As far as I can tell, events from the outer `onError` and `onComplete` can race with events from the inner `Publisher` which may lead to RS faults such as an `onNext` after an `onComplete`. This can obviously be solved via `ConcurrentTerminalSubscription` but I think you're trying to avoid that. I think we need to also make sure that the inner state is not in the emitting state and also in in inner-emit code, check if our outer state happened to transition to a terminal state.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherSwitchMap.java
```diff
@@ -0,0 +1,391 @@
...
+            this.target = target;
+            this.parent = parent;
+        }
+
+        @Override
+        public void onSubscribe(Subscription subscription) {
+            // No need to wrap in ConcurrentSubscription because only the latest RSubscriber interacts with the
+            // tSubscription which is atomically protected by the state variable.
+            tSubscription = requireNonNull(subscription);
+            target.onSubscribe(rSubscription);
```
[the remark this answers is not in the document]
Scottmitch: good call, let me add this.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherSwitchMap.java
bryce-anderson: Here an example I'm thinking of:
initial state: [outer state idle, inner state idle, maxDelayedErrors = 0]
1. Thread1: enter `RSubscriber.onNext`. Set state to `[outer state idle, inner state emitting]`, break loop, pause before `target.onNext(result);`
2. Thread2: call outer `onError(t)`, 
    - call `currPublisher.dispose(EMPTY_SUBSCRIPTION)` sets state to `[outer state idle, inner state disposed]`
    - call `outerErrorUpdateState(t)` sets state to `[outer state error, inner state disposed]`, `parent.maxDelayedErrors <= 0 && innerState != INNER_STATE_ERROR == true -> return t`
    - call `target.onError(t);`
3. Thread1: wake up concurrently with the still running Thread2 which is about to call `target.onError(t)` and call `target.onNext(result)` concurrently.

It looks like we guard against the inner state being in the critical portion of the `onError` pathway but not the `onNext` pathway.

Presuming this all checks out to be true, it might be easier to reshape the outer `onError` method to call specialized version(s) of the inner `dispose` that store the state for the inner loop to handle once it exits the emitting state.

Scottmitch: good call, this makes sense and I will fix.

bryce-anderson (approved):

PR #2681, opened 2023-08-26
Remove usage of deprecated PlatformDependent.throwException

Scottmitch:
Motivation:
PlatformDependent.throwException was deprecated many releases ago but its usage hasn't been removed.


idelpivnitskiy (commented):
Thank you!

PR #2683, opened 2023-08-30
Publisher.multicast late subscriber cancel demand bug

Scottmitch:
Motivation:
Publisher.multicast tracks the active subscriber with the least demand in order to control upstream demand. When cancel happens if the subscriber with minimum demand is removed the delta with the new minimum should be propagated upstream. However there is a bug where the initial demand offset isn't considered for late subscribers that get cancelled with may lead to not propagating demand upstream and "hanging".


tkountis (approved):
good find

PR #2684, opened 2023-08-30
Add Publisher.replay

Scottmitch:
Motivation:
Publisher.replay provides the ability to keep state that is preserved for multiple subscribers and across resubscribes.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MulticastPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MulticastPublisher.java
@@ -19,6 +19,7 @@ import io.servicetalk.concurrent.api.DefaultPriorityQueue.Node;
 import io.servicetalk.concurrent.internal.ArrayUtils;
 import io.servicetalk.concurrent.internal.DelayedSubscription;
 import io.servicetalk.concurrent.internal.RejectedSubscribeException;
+import io.servicetalk.concurrent.internal.TerminalNotification;
 import io.servicetalk.context.api.ContextMap;
 
 import org.slf4j.Logger;
@@ -32,6 +33,7 @@ import java.util.function.BiConsumer;
 import java.util.function.Function;
 import javax.annotation.Nullable;
 
+import static io.servicetalk.concurrent.api.Completable.completed;
 import static io.servicetalk.concurrent.api.PublishAndSubscribeOnPublishers.deliverOnSubscribeAndOnError;
 import static io.servicetalk.concurrent.internal.ConcurrentUtils.releaseLock;
 import static io.servicetalk.concurrent.internal.ConcurrentUtils.tryAcquireLock;
@@ -47,8 +49,10 @@ import static java.util.Comparator.comparingLong;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
 
-final class MulticastPublisher<T> extends AbstractNoHandleSubscribePublisher<T> {
+class MulticastPublisher<T> extends AbstractNoHandleSubscribePublisher<T> {
     private static final Logger LOGGER = LoggerFactory.getLogger(MulticastPublisher.class);
+    static final int DEFAULT_MULTICAST_QUEUE_LIMIT = 64;
+    static final Function<Throwable, Completable> DEFAULT_MULTICAST_TERM_RESUB = t -> completed();
     private static final Subscriber<?>[] EMPTY_SUBSCRIBERS = new Subscriber[0];
     @SuppressWarnings("rawtypes")
     private static final AtomicReferenceFieldUpdater<MulticastPublisher.State, Subscriber[]>
@@ -61,7 +65,8 @@ final class MulticastPublisher<T> extends AbstractNoHandleSubscribePublisher<T> 
     private final int minSubscribers;
     private final boolean exactlyMinSubscribers;
     private final boolean cancelUpstream;
-    private volatile State state;
+    @Nullable
+    volatile State state;
 
     MulticastPublisher(Publisher<T> original, int minSubscribers, boolean exactlyMinSubscribers, boolean cancelUpstream,
                        int maxQueueSize, Function<Throwable, Completable> terminalResubscribe) {
@@ -76,16 +81,30 @@ final class MulticastPublisher<T> extends AbstractNoHandleSubscribePublisher<T> 
         this.exactlyMinSubscribers = exactlyMinSubscribers;
         this.cancelUpstream = cancelUpstream;
         this.terminalResubscribe = requireNonNull(terminalResubscribe);
-        state = new State(maxQueueSize, minSubscribers);
+    }
+
+    static <T> MulticastPublisher<T> newMulticastPublisher(
+            Publisher<T> original, int minSubscribers, boolean exactlyMinSubscribers, boolean cancelUpstream,
+            int maxQueueSize, Function<Throwable, Completable> terminalResubscribe) {
+        MulticastPublisher<T> publisher = new MulticastPublisher<>(original, minSubscribers, exactlyMinSubscribers,
+                cancelUpstream, minSubscribers, terminalResubscribe);
+        publisher.resetState(maxQueueSize, minSubscribers);
+        return publisher;
     }
 
     @Override
-    void handleSubscribe(Subscriber<? super T> subscriber, ContextMap contextMap,
+    final void handleSubscribe(Subscriber<? super T> subscriber, ContextMap contextMap,
                          AsyncContextProvider contextProvider) {
-        state.addSubscriber(subscriber, contextMap, contextProvider);
+        final State cState = state;
+        assert cState != null;
+        cState.addNewSubscriber(subscriber, contextMap, contextProvider);
     }
 
-    private final class State extends MulticastRootSubscriber<MulticastFixedSubscriber<T>> implements Subscriber<T> {
+    void resetState(int maxQueueSize, int minSubscribers) {
+        state = new State(maxQueueSize, minSubscribers);
+    }
+
+    class State extends MulticastRootSubscriber<MulticastFixedSubscriber<T>> implements Subscriber<T> {
         private final DefaultPriorityQueue<MulticastFixedSubscriber<T>> demandQueue;
         volatile int subscribeCount;
         @SuppressWarnings("unchecked")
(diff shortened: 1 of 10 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/MulticastPublisher.java
```diff
@@ -179,32 +239,26 @@ void addSubscriber(Subscriber<? super T> subscriber, ContextMap contextMap,
...
-                    ((TerminalSubscriber<?>) currSubs[0]).terminate(subscriber);
-                    break;
-                } else {
-                    @SuppressWarnings("unchecked")
-                    Subscriber<? super T>[] newSubs = (Subscriber<? super T>[])
-                            Array.newInstance(Subscriber.class, currSubs.length + 1);
-                    System.arraycopy(currSubs, 0, newSubs, 0, currSubs.length);
-                    newSubs[currSubs.length] = multiSubscriber;
-                    if (newSubscribersUpdater.compareAndSet(this, currSubs, newSubs)) {
-                        addSubscriber(multiSubscriber);
```
Scottmitch, line 194: note this code was moved to `addSubscriber` bcz replay requires that we acquire the lock before making the subscriber visible in the array (via `newSubscribersUpdater`) and this is preferred approach for multicast too.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
```diff
@@ -0,0 +1,244 @@
...
+                    break;
+                } else if (nanoTime - item.timeStamp >= ttlNanos) {
+                    final int currentSize = getSize(currentStateSize);
+                    if (stateSizeUpdater.compareAndSet(this, currentStateSize,
+                            buildStateSize(getState(currentStateSize) + 1, currentSize - 1))) {
+                        // When we add: we add to the queue we add first, then CAS sizeState.
+                        // When we remove: we CAS the atomic state first, then poll.
+                        // This avoids removing a non-expired item because if the "add" thread is running faster and
+                        // already polled "item" the CAS will fail, and we will try again on the next loop iteration.
+                        items.poll();
```
bryce-anderson, line 219: I think we may still have a form of race between `accumulate(T)` and `expireSignals()` that can make otherwise good elements disappear.
```
items: [expired=true], maxItems = 1.
Thread1: accumulate. items.add(..); currentSize == maxItems, stateSizeUpdater.cas(..) == true; pause before items.poll();
items: [expired=true, expired=false]
Thread2: timer fires, expireSignals(); item != null && item is expired, stateSizeUpdater.cas(..) == true; items.poll();
items: [expired=false]
Thread1: items.poll(); break;
items: []
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    boolean cancelUpstream();
+
+    /**
+     * Get a hint to limit the number of elements which will be queued for each {@link Subscriber} in order to
+     * compensate for unequal demand and late subscribers.
+     * @return a hint to limit the number of elements which will be queued for each {@link Subscriber} in order to
+     * compensate for unequal demand and late subscribers.
+     */
+    int queueLimitHint();
```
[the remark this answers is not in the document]
Scottmitch: The idea was to give some leeway in the API as it may not be strictly enforced as a limit but more of a guideline (e.g. approximately this much is the limit).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
```diff
@@ -0,0 +1,244 @@
...
+
+        private static int getState(long stateSize) {
+            return (int) (stateSize >>> 32);
+        }
+
+        private static long buildStateSize(int state, int size) {
+            return (((long) state) << 32) | size;
+        }
+
+        private void schedulerTimer(long nanos) {
```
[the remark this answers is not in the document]
Scottmitch: In other similar cases (e.g. `buffer(..)`) we also require a timer. If you just limit based upon time you may fall into the trap of too many items during your time interval, so to remove potential sharp edges by default we require both. You can simulate a "time only" option by setting the limit very high.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
Scottmitch: you are correct, great catch! will fix.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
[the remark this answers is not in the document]
bryce-anderson: Sorry, I wasn't terribly clear, I think we're on the same page for `maxItems` but not the value of the timer _task_. I know we have two different limiters and I think the `maxItems` is the overwhelmingly most valuable of the two limits and a must have.
The timer task, otoh, doesn't appear to buy a lot other than (theoretically) better GC behavior due to the task eagerly evicting expired items: we still have to potentially buffer `maxItems` that come in rapidly so the max memory usage is theoretically the same. However, even without the timer task we can still honor the time constraint by checking if items are expired when it's time to emit them and pruning then. By pruning during emission we save us the overhead of the timer tasks themselves and some extra concurrency primitive overhead since the emission happens inside the protection of our existing concurrency primitives.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
```diff
@@ -0,0 +1,287 @@
...
+        public void deliverAccumulation(final Consumer<T> consumer) {
+            int i = 0;
+            for (TimeStampSignal<T> timeStampSignal : items) {
+                consumer.accept(timeStampSignal.signal);
+                // The queue size maybe larger than maxItems if we weren't able to acquire the queueLock while adding.
+                // This is only a temporary condition while there is concurrent access between timer and accumulator.
+                // Guard against it here to preserve the invariant that we shouldn't deliver more than maxItems.
+                if (++i >= maxItems) {
+                    break;
+                }
```
bryce-anderson, line 180: This feels like it can turn to an item leak.
```
maxItems = 1, items=[T1]
Thread1: timer fires and we take the queueLock, and suspend.
items=[T1]
Thread2: accumulate(T2), fail to take queueLock, adds T2 to items, and exits.
items=[T1, T2]
Thread2: new subscription, enters deliverAccumulation(..), emit T1, ++i == maxItems -> break; <-!! sees T1 but missed item T2.
```
I think this is intrinsically racy behavior: a program shouldn't be able to tell whether it received [T1, T2] from accumulation, a post-subscribe onNext calls, or a mix, so we should just deliver everything in the queue.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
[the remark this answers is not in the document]
Scottmitch: Ah I see what you are saying. I'm not sure there is a "right answer" and depends upon the scenario.

Timer Task Desirable:
Desired lifetime of element is `<small number>` (e.g. 1 second), an upper bound of `<high number>`, data is expected to arrive on average in 100ms intervals (but can be bursty). I don't want to retain `high number` in memory at all times and the timer is expected to dictate queue size in most cases.

Timer Task is Extra Overhead:
Desired lifetime of element is `<high number>` (e.g. 15 seconds), an upper bound of `<small number>`, data is expected to arrive on average in 100ms intervals (but may have gaps in time). The timer isn't really useful here (but depending upon the granularity may not also be a primary resource consumer).

We could always provide an alternative strategy/option if we want to but rational is the timer task based approach provides a more conservatively bounded approach to start (although more complex from implementation perspective).

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
Scottmitch: great point. this is tricky condition to test test too.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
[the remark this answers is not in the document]
Scottmitch: I added a `historyTtlBuilder` overload that has a `lazy` argument to demonstrate the two approaches. The method is currently package private and I'm not sure we want to expose both yet

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
```diff
@@ -0,0 +1,371 @@
...
+     *     <li>{@code true} will evict expired items in a lazy fashion when new subscribers arrive. This approach
+     *     is more likely to retain {@code historyHint} elements in memory in steady state, but avoids cost of
+     *     scheduling timer tasks.</li>
+     *     <li>{@code false} will evict expired items eagerly when they expire. If {@code ttl} is lower that
+     *     {@code historyHint} relative to signal arrival rate this can use less memory but schedules time tasks.</li>
+     * </ul>
+     * @param <T> The type of {@link ReplayStrategyBuilder}.
+     * @return a {@link ReplayStrategyBuilder} using the historyHint and TTL strategy.
+     */
+    public static <T> ReplayStrategyBuilder<T> historyTtlBuilder(int historyHint, Duration ttl, Executor executor,
```
Scottmitch, line 93: @bryce-anderson - I added this method as a result of your suggestions.

`lazy = false` is likely to be useful if time duration is the dominant limiter, `lazy = true` if history is the dominant limiter.

wdyt?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
```diff
@@ -0,0 +1,382 @@
...
+
+    private static final class LazyTimeLimitedReplayAccumulator<T> implements ReplayAccumulator<T> {
+        @SuppressWarnings("rawtypes")
+        private static final AtomicIntegerFieldUpdater<LazyTimeLimitedReplayAccumulator> queueSizeUpdater =
+                AtomicIntegerFieldUpdater.newUpdater(LazyTimeLimitedReplayAccumulator.class, "queueSize");
+        private final Executor executor;
+        private final Queue<TimeStampSignal<T>> items;
+        private final long ttlNanos;
+        private final int maxItems;
+        private volatile int queueSize;
```
bryce-anderson, line 146: This is mostly the same as `MostRecentReplayAccumulator` now and I don't think we need to worry about concurrent calls to either `accumulate(..)` or `deliverAccumulation(..)` so we can use a standard queue and it's simple `.size()` method.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
```diff
@@ -0,0 +1,382 @@
...
+                        items.add(signal);
+                        break;
+                    }
+                } else if (queueSizeUpdater.compareAndSet(this, qSize, qSize)) {
+                    items.poll();
+                    items.add(signal);
+                    break;
+                }
+            }
+        }
```
bryce-anderson, line 177: This is also an opportunity to evict stale elements if we'd like. We can have a simple `private evictStale()` method and call it at the start of these public methods. Then the rest looks almost exactly like `MostRecentReplayAccumulator`.

bryce-anderson (approved):
Looks good other than some minor optimizations.
Thank you for your patience.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
bryce-anderson: I like it. Thank you.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategy.java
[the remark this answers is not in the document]
bryce-anderson: Sounds reasonable to me. Thank you for the explanation.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ReplayStrategies.java
Scottmitch: lets defer this bcz `LazyTimeLimitedReplayAccumulator` can do the iteration and removal at the same time which is a bit harder to generalize.

PR #2690, opened 2023-09-08
Add Publisher.zip

Scottmitch:
Motivation:
The zip operator helps combine results from multiple Publishers into a single Publisher.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherZipper.java
```diff
@@ -0,0 +1,302 @@
...
+                    sb.append("Publisher indexes: [");
+                    Iterator<Integer> itr = nonEmptyIndexes.iterator();
+                    sb.append(itr.next()); // safe to call next(), already checked is not empty.
+                    while (itr.hasNext()) {
+                        sb.append(", ").append(itr.next());
+                    }
+                    sb.append("] had onNext signals queued when onComplete terminal signal received");
+                    subscriber.onError(new IllegalStateException(sb.toString()));
+                }
+            }
```
bryce-anderson, line 218: I don't think this is the same behavior described in https://reactivex.io/documentation/operators/zip.html. It suggests that its okay to have extra signals emitted and they are simply dropped which is the behavior of most collection zip operators that I've seen. At a minimum, the behavior should be documented so users know they should have publishers of all the same length, which I think could be non-trivial in some cases.

Along these same lines, I don't think this implementation will ever terminate if one of the suppliers is an infinite stream. It seems pretty reasonable to have one source that is an infinite stream an the join it with another finite stream and expect a finite result.

AkshDesai04 (approved):

PR #2691, opened 2023-09-13
`CONNECT` request to a proxy must include `Host` header

idelpivnitskiy:
Motivation:

According to RFC7230 section 5.4
(https://datatracker.ietf.org/doc/html/rfc7230#section-5.4), all HTTP/1.1 requests must contain the `Host` header. If the target URI includes an authority component, then a client MUST send a field-value for Host that is identical to that authority component. On the other hand, `content-length: 0` header is not required because request payload body is not expected:
https://datatracker.ietf.org/doc/html/rfc9110#section-9.3.6

Modifications:
- Enhance `ProxyTunnel` to parse and validate initial line and the host header before passing data to `handler`, return 400 if it finds an error in request;
- Update `ProxyConnectConnectionFactoryFilter` to remove `content-lenght: 0` and add `host` header;
- Use `BuilderUtils` in `HttpsProxyTest` to allow enabling wire logging via system properties;

Result:

`CONNECT` request follows RFC guidelines and examples.

### DIFF
--- a/servicetalk-http-netty/src/testFixtures/java/io/servicetalk/http/netty/ProxyTunnel.java
+++ b/servicetalk-http-netty/src/testFixtures/java/io/servicetalk/http/netty/ProxyTunnel.java
@@ -28,10 +28,17 @@ import java.io.OutputStream;
 import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
+import java.util.Locale;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicInteger;
 import javax.annotation.Nullable;
 
+import static io.servicetalk.http.api.HttpHeaderNames.CONTENT_LENGTH;
+import static io.servicetalk.http.api.HttpHeaderNames.HOST;
+import static io.servicetalk.http.api.HttpProtocolVersion.HTTP_1_1;
+import static io.servicetalk.http.api.HttpRequestMethod.CONNECT;
+import static io.servicetalk.http.api.HttpResponseStatus.BAD_REQUEST;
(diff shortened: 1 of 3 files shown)


servicetalk-http-netty/src/testFixtures/java/io/servicetalk/http/netty/ProxyTunnel.java
```diff
@@ -79,12 +86,41 @@ public HostAndPort startProxy() throws IOException {
...
+                                        initialLine);
+                            }
+                            final int end = initialLine.indexOf(' ', CONNECT_PREFIX.length());
+                            final String authority = initialLine.substring(CONNECT_PREFIX.length(), end);
+                            final int colon = authority.indexOf(':');
+                            host = authority.substring(0, colon);
+                            port = Integer.parseInt(authority.substring(colon + 1));
+                            protocol = initialLine.substring(end + 1);
+
+                            final String hostHeader = readLine(in);
```
daschl, line 105: Where this parsing could potentially be improved is that if I read this correctly it expects the host header to always come as the first header. Others are possible i.e. `Proxy-Authorization` and they might have a different order, but since we use this for internal testing it's fine. Just wanted to mention it in case we want to beef up the testing in the future they could be parsed into a map and then you do your assertions against the map.

daschl (approved):

servicetalk-http-netty/src/testFixtures/java/io/servicetalk/http/netty/ProxyTunnel.java
idelpivnitskiy: That's a good observation. You are right, it can have other headers. Will improve later when we test those use-cases.

PR #2692, opened 2023-09-13
Drain proxy `CONNECT` response before starting TLS handshake

idelpivnitskiy:
Motivation:

1. Successful response to `CONNECT` never has a message body, and we are not interested in payload body for any non-200 status code. Drain it asap to free connection and RS resources before starting TLS handshake.
2. Result of the proxy tunneling is incorrectly offloaded, response offloading can be different from users-configured `ConnectExecutionStrategy`.

Modifications:
- Drain response message body before invoking `handleConnectResponse`;
- Use computed `ExecutionStrategy` to decide when `newConnection` should be offloaded, similar to what `AbstractLBHttpConnectionFactory` does;
- Update `ProxyConnectConnectionFactoryFilterTest` to assert new behavior;

Result:

Response to `CONNECT` request is drained and all publishers are released before starting any new activity on this connection.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
@@ -26,6 +26,9 @@ import io.servicetalk.http.api.FilterableStreamingHttpConnection;
 import io.servicetalk.http.api.HttpExecutionStrategies;
 import io.servicetalk.http.api.HttpExecutionStrategy;
 import io.servicetalk.http.api.StreamingHttpResponse;
+import io.servicetalk.transport.api.ConnectExecutionStrategy;
+import io.servicetalk.transport.api.ExecutionStrategy;
+import io.servicetalk.transport.api.IoThreadFactory;
 import io.servicetalk.transport.api.TransportObserver;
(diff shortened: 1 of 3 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -87,7 +93,13 @@ public Single<C> newConnection(final ResolvedAddress resolvedAddress,
                 return delegate().newConnection(resolvedAddress, contextMap, observer).flatMap(c -> {
                     try {
                         return c.request(c.connect(connectAddress).addHeader(CONTENT_LENGTH, ZERO))
-                                .flatMap(response -> handleConnectResponse(c, response).shareContextOnSubscribe())
+                                // Successful response to CONNECT never has a message body, and we are not interested in
+                                // payload body for any non-200 status code. Drain it asap to free connection and RS
+                                // resources before starting TLS handshake.
+                                .flatMap(response -> response.messageBody().ignoreElements()
+                                        .concat(Single.defer(() -> handleConnectResponse(c, response)
```
idelpivnitskiy, line 100: Tried to minimize git diff, added draining here instead of changing indentation inside `handleConnectResponse`

daschl (approved):

PR #2693, opened 2023-09-14
Calculate initial concurrency based on resulting connection protocol

idelpivnitskiy:
Motivation:

`RequestConcurrencyController` is always applied as the top level wrapper of the connection (after all connection-level filters) before returning it back to the `LoadBalancer`. As the result, there is no need to make an earlier decision for `initialConcurrency` value at implementations of `AbstractLBHttpConnectionFactory` (when the final protocol may be unknown) because it can be inferred based on the resulting connection protocol at the time a new
`ReservableRequestConcurrencyController` is created. Early initialization as it's now makes it harder to switch protocols after proxy connect.

Modifications:

- Convert abstract method `AbstractLBHttpConnectionFactory.newConcurrencyController` into a private method;
- Compute `initialConcurrency` based on connection's protocol and `ReadOnlyHttpClientConfig`;

Result:

No need to pre-initialize `initialConcurrency` value in `AbstractLBHttpConnectionFactory` implementations as it will be computed based on the connection's protocol at the time it's required.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLBHttpConnectionFactory.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLBHttpConnectionFactory.java
@@ -41,6 +41,7 @@ import java.util.function.Function;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.api.AsyncCloseables.emptyAsyncCloseable;
+import static io.servicetalk.http.netty.ReservableRequestConcurrencyControllers.newConcurrencyController;
 import static io.servicetalk.transport.api.TransportObservers.asSafeObserver;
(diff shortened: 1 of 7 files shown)


idelpivnitskiy:
I'm trying to extract meaningful code changes into independent PRs before finalizing PR for a proxy connect with ALPN (can switch from HTTP/1.1 to HTTP/2 protocol)

daschl (approved):

idelpivnitskiy:
@daschl as we discussed offline, reverted changes for `ReservableRequestConcurrencyControllers` and move computation logic to `AbstractLBHttpConnectionFactory`. Now the delta looks smaller and `ReservableRequestConcurrencyControllers` doesn't need to know anything about `ReadOnlyHttpClientConfig`.

PR #2697, opened 2023-09-19
Move HTTP proxy `CONNECT` logic before `ConnectionFactoryFilter`s

idelpivnitskiy:
Motivation:

Currently, HTTP proxy `CONNECT` logic is prepended before user-defined `ConnectionFactoryFilter`s. Because user filters can wrap `FilterableStreamingHttpConnection`:
1. It makes it hard to upgrade the connection if ALPN agrees on a different protocol.
2. `ProxyConnectConnectionFactoryFilter` can not access Netty's `Channel` to trigger TLS handshake.

Modifications:
- Create `ProxyConnectLBHttpConnectionFactory` as an alternative to `PipelinedLBHttpConnectionFactory` that makes HTTP proxy `CONNECT` request after `PipelinedStreamingHttpConnection` is created;
- Keep `ProxyConnectConnectionFactoryFilter` only for the purpose of propagating `HTTP_TARGET_ADDRESS_BEHIND_PROXY` key upstream;
- Add internal `NettyFilterableStreamingHttpConnection` to get access to Netty `Channel` without type cast;
- Enhance `StacklessClosedChannelException` to carry a message;
- Rename `ProxyConnectConnectionFactoryFilterTest` to `ProxyConnectLBHttpConnectionFactoryTest`, adjust for testing `processConnect` method;

Result:

1. No need to worry that HTTP CONNECT won't work if users wrapped `FilterableStreamingHttpConnection` in their `ConnectionFactoryFilter`s.
2. Makes it possible to change protocol based on ALPN (will be done in follow-up PRs).

Behavior change:
- Because user-defined `ConnectionFactoryFilter`s can not intercept HTTP `CONNECT` requests anymore, they lose ability to alter the request (for example, to set auth or debug headers). Follow-up PRs will add new API to let users intercept such requests.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
@@ -46,8 +46,13 @@ import static io.servicetalk.http.api.HttpProtocolVersion.HTTP_1_1;
 import static io.servicetalk.http.api.HttpResponseStatus.StatusClass.SUCCESSFUL_2XX;
 import static io.servicetalk.http.netty.StreamingConnectionFactory.buildStreaming;
 import static io.servicetalk.utils.internal.ThrowableUtils.addSuppressed;
-import static java.util.Objects.requireNonNull;
 
+/**
+ * {@link AbstractLBHttpConnectionFactory} implementation that handles HTTP/1.1 CONNECT when a client is configured to
+ * talk over HTTPS Proxy Tunnel.
+ *
+ * @param <ResolvedAddress> The type of resolved address.
+ */
 final class ProxyConnectLBHttpConnectionFactory<ResolvedAddress>
         extends AbstractLBHttpConnectionFactory<ResolvedAddress> {
 
@@ -62,8 +67,9 @@ final class ProxyConnectLBHttpConnectionFactory<ResolvedAddress>
             final ProtocolBinding protocolBinding) {
         super(config, executionContext, version -> reqRespFactory, connectStrategy, connectionFactoryFilter,
                 connectionFilterFunction, protocolBinding);
-        requireNonNull(config.h1Config(), "H1ProtocolConfig is required");
-        assert config.connectAddress() != null;
+        assert config.h1Config() != null : "H1ProtocolConfig is required";
+        assert config.tcpConfig().sslContext() != null : "Proxy CONNECT works only for TLS connections";
+        assert config.connectAddress() != null : "Address (authority) for CONNECT request is required";
         connectAddress = config.connectAddress().toString();
     }
 
@@ -93,7 +99,8 @@ final class ProxyConnectLBHttpConnectionFactory<ResolvedAddress>
                     .flatMap(response -> {
                         // Successful response to CONNECT never has a message body, and we are not interested in payload
                         // body for any non-200 status code. Drain it asap to free connection and RS resources before
-                        // starting TLS handshake or propagating an error.
+                        // starting TLS handshake or propagating an error. We do this after verifying the status to
+                        // preserve ProxyResponseException even if draining fails with an exception.
                         if (response.status().statusClass() != SUCCESSFUL_2XX) {
                             return drainPropagateError(response, new ProxyResponseException(c +
                                     " Non-successful response from proxy CONNECT " + connectAddress, response.status()))
@@ -168,7 +175,7 @@ final class ProxyConnectLBHttpConnectionFactory<ResolvedAddress>
     private static Single<FilterableStreamingHttpConnection> safeCompletePropagateError(
             final Completable completable, final Throwable error) {
         return completable
-                .onErrorResume(closeError -> Completable.failed(addSuppressed(error, closeError)))
+                .onErrorResume(completableError -> Completable.failed(addSuppressed(error, completableError)))
                 .concat(Single.failed(error));
     }
 }
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright © 2019-2023 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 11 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
```diff
@@ -0,0 +1,174 @@
...
+            final ReadOnlyHttpClientConfig config, final HttpExecutionContext executionContext,
+            @Nullable final StreamingHttpConnectionFilterFactory connectionFilterFunction,
+            final StreamingHttpRequestResponseFactory reqRespFactory,
+            final ExecutionStrategy connectStrategy,
+            final ConnectionFactoryFilter<ResolvedAddress, FilterableStreamingHttpConnection> connectionFactoryFilter,
+            final ProtocolBinding protocolBinding) {
+        super(config, executionContext, version -> reqRespFactory, connectStrategy, connectionFactoryFilter,
+                connectionFilterFunction, protocolBinding);
+        requireNonNull(config.h1Config(), "H1ProtocolConfig is required");
+        assert config.connectAddress() != null;
```
daschl, line 66: What's the reasoning behind mixing requireNonNull and assert != null?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
```diff
@@ -0,0 +1,174 @@
...
+import static io.servicetalk.http.api.HttpContextKeys.HTTP_EXECUTION_STRATEGY_KEY;
+import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
+import static io.servicetalk.http.api.HttpHeaderNames.HOST;
+import static io.servicetalk.http.api.HttpProtocolVersion.HTTP_1_1;
+import static io.servicetalk.http.api.HttpResponseStatus.StatusClass.SUCCESSFUL_2XX;
+import static io.servicetalk.http.netty.StreamingConnectionFactory.buildStreaming;
+import static io.servicetalk.utils.internal.ThrowableUtils.addSuppressed;
+import static java.util.Objects.requireNonNull;
+
+final class ProxyConnectLBHttpConnectionFactory<ResolvedAddress>
```
daschl, line 51: maybe a short javadoc describing what the connection factory does is helpful for someone looking through the code.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
```diff
@@ -0,0 +1,174 @@
...
+
+    private static Single<FilterableStreamingHttpConnection> closePropagateError(
+            final FilterableStreamingHttpConnection connection, final Throwable error) {
+        return safeCompletePropagateError(connection.closeAsync(), error);
+    }
+
+    private static Single<FilterableStreamingHttpConnection> safeCompletePropagateError(
+            final Completable completable, final Throwable error) {
+        return completable
+                .onErrorResume(closeError -> Completable.failed(addSuppressed(error, closeError)))
```
daschl, line 171: nit: `closeError` could also be a `drainError`? maybe refer to it as `suppressedError` or similar since that's where it ends up?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
```diff
@@ -0,0 +1,174 @@
...
+                    .flatMap(response -> {
+                        // Successful response to CONNECT never has a message body, and we are not interested in payload
+                        // body for any non-200 status code. Drain it asap to free connection and RS resources before
+                        // starting TLS handshake or propagating an error.
+                        if (response.status().statusClass() != SUCCESSFUL_2XX) {
+                            return drainPropagateError(response, new ProxyResponseException(c +
+                                    " Non-successful response from proxy CONNECT " + connectAddress, response.status()))
+                                    .shareContextOnSubscribe();
+                        }
+                        return response.messageBody().ignoreElements()
```
daschl, line 102: looks like draining is done in both codepaths (error or success) - consider draining it beforehand and then check the status to potentially bail out?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
```diff
@@ -0,0 +1,174 @@
...
+                                .concat(handshake(c))
```
daschl, line 103: the handshake is always added, looking for a `SslHandshakeCompletionEvent` - what happens in a non-HTTPS/TLS scenario?

idelpivnitskiy: The CONNECT logic works only for TLS connections, `DefaultSingleAddressHttpClientBuilder` won't add this factory if TLS is not configured. I will add an assertion in the constructor to verify that inside `ProxyConnectLBHttpConnectionFactory` too.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
idelpivnitskiy: I considered that, but then realized that if draining logic throws, we will never verify the response status and won't generate `ProxyResponseException`. It feels that knowing the response status will give users more information for troubleshooting vs seeing some weird error as a result of payload body draining. Because if that, decided to always drain after. Will add a comment to clarify.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
idelpivnitskiy: Worst case, if something is messed up, it won't find `DeferSslHandler` and will throw an exception. `ProxyConnectLBHttpConnectionFactoryTest` verifies these use-cases.

idelpivnitskiy: For non-TLS (http) use-cases, requests are sent as-is to the proxy server with an absolute-form request target. Proxy will forward it to the final destination. See `HttpProxyTest` for details.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
daschl: ah that makes sense now - thanks!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
daschl: 👍

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
daschl: 👍 thanks!

daschl (approved):

daschl (approved):

PR #2698, opened 2023-09-19
Add initializer API for HTTP proxy `CONNECT` request

idelpivnitskiy:
Motivation:

After #2697 moved HTTP proxy `CONNECT` logic before user-defined
`ConnectionFactoryFilter`s, users lost the ability to intercept `CONNECT`
requests for the purpose of adding custom headers, like auth.

Modifications:

- Add `SingleAddressHttpClientBuilder.proxyAddress(...)` overload that
takes `Consumer<StreamingHttpRequest>` as a 2nd argument;
- Recompute `HttpExecutionStrategy` after `CONNECT` request initializer
is invoked in `ProxyConnectLBHttpConnectionFactory`;
- Enhance `ProxyConnectLBHttpConnectionFactoryTest` to verify that the
initializer is invoked and users can alter the execution strategy;
- Enhance `ProxyTunnel` and `HttpsProxyTest` to verify that new API
can be used to send `Proxy-Authorization` header;

Result:

Users have explicit API to alter HTTP `CONNECT` requests if necessary.

---
Depends on #2697, review only the last commit.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
@@ -78,7 +78,7 @@ public interface SingleAddressHttpClientBuilder<U, R> extends HttpClientBuilder<
      * request. It can be used to add headers, like {@link HttpHeaderNames#PROXY_AUTHORIZATION}, debugging info, etc.
      * @return {@code this}.
      */
-    default SingleAddressHttpClientBuilder<U, R> proxyAddress(  // FIXME: 0.43 - remove default impl
+    default SingleAddressHttpClientBuilder<U, R> proxyAddress(// FIXME: 0.43 - remove default impl
             U proxyAddress, Consumer<StreamingHttpRequest> connectRequestInitializer) {
         throw new UnsupportedOperationException(
                 "Setting proxy address with request initializer is not yet supported by " + getClass().getName());
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018-2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2018-2023 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
```diff
@@ -63,6 +64,26 @@ default SingleAddressHttpClientBuilder<U, R> proxyAddress(U proxyAddress) { // F
...
+     *
+     * @param proxyAddress Unresolved address of the proxy. When used with a builder created for a resolved address,
+     * {@code proxyAddress} should also be already resolved – otherwise runtime exceptions may occur.
+     * @param requestInitializer {@link Consumer} of {@link StreamingHttpRequest} that can be used to add additional
+     * info to <a href="https://datatracker.ietf.org/doc/html/rfc9110#section-9.3.6">HTTP/1.1 CONNECT</a> request.
+     * It can be used to add headers, like {@link HttpHeaderNames#PROXY_AUTHORIZATION}, debugging information, etc.
+     * @return {@code this}.
+     */
+    default SingleAddressHttpClientBuilder<U, R> proxyAddress(U proxyAddress,   // FIXME: 0.43 - remove default impl
+                                                              Consumer<StreamingHttpRequest> requestInitializer) {
```
daschl, line 82: nit: I know it's in the javadoc, but consider renaming it to "connectRequestInitializer" to make it clear that it is really only for the initial connect and not any subsequent http requests going through the proxy.

servicetalk-http-netty/src/testFixtures/java/io/servicetalk/http/netty/ProxyTunnel.java
```diff
@@ -181,6 +194,22 @@ private static String readLine(final InputStream in) throws IOException {
         }
     }
 
+    private static Headers readHeaders(final InputStream in) throws IOException {
```
daschl, line 197: it's generic now, yay! 🥳

daschl (approved):

idelpivnitskiy:
We made a decision to defer adding this API for now. Taking into account that while before #2697 it was possible to add `proxy-authorization` header via `ConnectionFactoryFilter`, in practice that approach was too complicated for users to figure it out without our help and we are not aware of anyone requiring this feature for now. 
If the use-case appears, we will provide the required API. At this point, we would like to hear real user request before we prematurely add new API. This will help us to implement better API design and answer questions like:
1. Should we let them initialize the whole `StreamingHttpRequest` or limit it to only `HttpRequestMetaData` or `HttpHeaders`?
2. Do they need a function/consumer or a static way to provide auth info?
3. Do we need a more complex configuration API for proxies or not, like a way to let users configure a timeout for `CONNECT` request?

I cherry-picked enhancements for tests from this PR to #2697.

idelpivnitskiy:
Reworked in https://github.com/apple/servicetalk/pull/2744

PR #2703, opened 2023-09-21
Introduce defaultHttp(s)Port for MultiAddressHttpClientBuilder

daschl:
Motivation
----------
At the moment, if no port number is explicitly provided, the default HTTP and HTTPS ports are used (80 and 443). In some environments it can be useful (i.e. if a client is always talking to a high port number) to not have to specify it on every request.

Modifications
-------------
This changeset introduces two new builder options to allow customizing the defaultHttpPort as well as the defaultHttpsPort and applying it if no explicit port is provided on the request itself.

Result
------
It is now possible to not specify the port number on every request if they diverge from the default 80 or 443 port numbers.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
@@ -147,4 +147,28 @@ public interface MultiAddressHttpClientBuilder<U, R> extends HttpClientBuilder<U
      * @see RedirectConfigBuilder
      */
     MultiAddressHttpClientBuilder<U, R> followRedirects(@Nullable RedirectConfig config);
+
+    /**
+     * Configures the default port for the HTTP scheme if not explicitly provided as part of the
+     * {@link HttpRequestMetaData#requestTarget()}.
+     *
+     * @param port the port that should be used if not explicitly provided for HTTP requests.
+     * @return {@code this}.
+     */
+    default MultiAddressHttpClientBuilder<U, R> defaultHttpPort(int port) { // FIXME: 0.43 - remove default impl
+        throw new UnsupportedOperationException("Setting defaultHttpPort is not yet supported by "
+                + getClass().getName());
+    }
+
+    /**
+     * Configures the default port for the HTTPS scheme if not explicitly provided as part of the
+     * {@link HttpRequestMetaData#requestTarget()}.
+     *
+     * @param port the port that should be used if not explicitly provided for HTTPS requests.
+     * @return {@code this}.
+     */
+    default MultiAddressHttpClientBuilder<U, R> defaultHttpsPort(int port) { // FIXME: 0.43 - remove default impl
+        throw new UnsupportedOperationException("Setting defaultHttpsPort is not yet supported by "
+                + getClass().getName());
+    }
 }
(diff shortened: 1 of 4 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
```diff
@@ -147,4 +147,22 @@ default MultiAddressHttpClientBuilder<U, R> headersFactory(HttpHeadersFactory he
...
+    MultiAddressHttpClientBuilder<U, R> defaultHttpPort(int port);
+
+    /**
+     * Configures the default port for the HTTPS scheme if not explicitly provided as part of the
+     * {@link HttpRequestMetaData#requestTarget()}.
+     *
+     * @param port the port that should be used if not explicitly provided for HTTPS requests.
+     * @return {@code this}.
+     */
+    MultiAddressHttpClientBuilder<U, R> defaultHttpsPort(int port);
```
idelpivnitskiy, line 167: New methods should have default implementations for backward compatibility

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -453,4 +463,29 @@ public MultiAddressHttpClientBuilder<HostAndPort, InetSocketAddress> followRedir
...
+    }
+
+    /**
+     * Verifies that the given port number is within the allowed range.
+     *
+     * @param port the port number to verify.
+     * @return the port number if in range.
+     */
+    private static int verifyPortRange(final int port) {
+        if (port < 0 || port > 0xFFFF) {
```
idelpivnitskiy, line 486: port 0 also doesn't make sense in the URL, consider excluding 0.
AHC also does not allow 0.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -453,4 +463,29 @@ public MultiAddressHttpClientBuilder<HostAndPort, InetSocketAddress> followRedir
...
+            throw new IllegalArgumentException("Provided port number out of range: " + port);
```
idelpivnitskiy, line 487: 1. Missed "is" before "out"?
2. For consistency with any other validation we have, consider specifying expected range in the message.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MultiAddressUrlHttpClientTest.java
```diff
@@ -255,14 +255,33 @@ void requestWithAsteriskFormRequestTargetWithHostHeader() {
...
+        try (StreamingHttpClient client = HttpClients
+                .forMultiAddressUrl(ID).defaultHttpPort(serverPort).buildStreaming()) {
+            StreamingHttpRequest request = client.get(format("http://%s/200", serverHost));
+            requestAndValidate(client, request, OK, "/200");
+        }
+    }
+
+    @Test
+    void requestWithExplicitPortTakesPrecedenceOverCustomDefaultHttpPort() throws Exception {
+        int illegalPort = serverPort + 1;
```
idelpivnitskiy, line 279: There is a risk that env where someone runs tests has an HTTP server on `serverPort + 1`. Consider using this trick to find an available port:

```java
    static int availablePort() throws IOException {
        ServerSocket s = new ServerSocket(0);
        int port = s.getLocalPort();
        s.close();
        return port;
    }
```

idelpivnitskiy (changes requested):

PR #2707, opened 2023-09-21
CI: Enable jdk21 builds

Scottmitch:
This reverts commit a2704b9e24e7ae4eada9267427ea881abe30b073.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,6 +41,10 @@ task validateLocalDocSite(type: Exec) {
 quality.dependsOn validateLocalDocSite
 
 subprojects {
+  // mockito 5 only supports jdk11+
+  if (JavaVersion.current() < JavaVersion.VERSION_11) {
+    project.setProperty("mockitoCoreVersion", mockitoCorePreJdk11Version)
+  }
   // Used by ci-release.yaml to determine which modules need to be built/released with JDK11.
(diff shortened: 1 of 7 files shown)


Scottmitch:
wait to merge until jdk21 G1GC issue resolved https://github.com/apple/servicetalk/pull/2705

build.gradle
```diff
@@ -41,6 +41,10 @@ task validateLocalDocSite(type: Exec) {
 quality.dependsOn validateLocalDocSite
 
 subprojects {
+  // mockito 5 only supports jdk11+
+  if (JavaVersion.current() < JavaVersion.VERSION_11) {
+    project.setProperty("mockitoCoreVersion", "4.11.0")
```
idelpivnitskiy, line 46: If 4.x works well on JDK21, will be preferable to stick to that until we can bump the minimal required version for ST to JDK11. Using 4.x conditionally opens a risk of compilation errors if 5.x breaks API.
If there are any issues with 4.x on JDK21, consider adding an entry in gradle.properties for 4.x. If it's only here, it's easy to forget.

Scottmitch: 4.x doesn't work due to bytebuddy and incompatible class file structure

Scottmitch: I can create another property for this

PR #2709, opened 2023-09-22
Demonstrate multi-address client in HTTP examples

sonicloong:
Motivation:
Some HTTP examples only use single-address client. It would be helpful to also provide multi-address client examples.

Modification:
- Create `RetryUrlClient`
- Create `TimeoutUrlClient`
- Create `LifecycleObserverUrlClient `
- Create `CompressionFilterExampleUrlClient `

Result:
Both single-address and multi-address clients are demonstrated in HTTP examples.


idelpivnitskiy (approved):
Thank you!

PR #2710, opened 2023-09-25
Clean up client response payloads on error to free up connections

daschl:
Motivation
----------
When a client makes a request and during response processing an exception is thrown, the exception bubbles up to the client but the message payload body is then not available to be consumed.

In this case, the underlying pooled connection will not be freed up and a new connection is created.

Modifications
-------------
By tracking the message payloads when responses are bubbled up through the filter chain, we can detect if an error happens and if detected the message payload body is being proactively drained.

This will free up the connection again so it can be reused by a subsequent request.

Result
------
More correct handling of message payload bodies if an exception happens during response filter processing.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
@@ -322,6 +322,9 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
                         ctx.builder.retryingHttpRequesterFilter);
             }
 
+            currClientFilterFactory = appendFilter(currClientFilterFactory,
+                    HttpMessageDiscardWatchdogClientFilter.INSTANCE);
+
             // Internal retries must be one of the last filters in the chain.
             currClientFilterFactory = appendFilter(currClientFilterFactory, InternalRetryingHttpClientFilter.INSTANCE);
             FilterableStreamingHttpClient wrappedClient =
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
@@ -140,6 +140,8 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
         strategyComputation = new ClientStrategyInfluencerChainBuilder();
         this.loadBalancerFactory = DefaultHttpLoadBalancerFactory.Builder.<R>fromDefaults().build();
         this.serviceDiscoverer = requireNonNull(serviceDiscoverer);
+
+        clientFilterFactory = appendFilter(clientFilterFactory, HttpMessageDiscardWatchdogClientFilter.CLEANER);
     }
 
     private DefaultSingleAddressHttpClientBuilder(@Nullable final U address,
@@ -296,6 +298,7 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
                             targetAddress(ctx)));
 
             ContextAwareStreamingHttpClientFilterFactory currClientFilterFactory = ctx.builder.clientFilterFactory;
+
             if (roConfig.hasProxy() && sslContext == null) {
                 // If we're talking to a proxy over http (not https), rewrite the request-target to absolute-form, as
                 // specified by the RFC: https://tools.ietf.org/html/rfc7230#section-5.3.2
@@ -314,7 +317,12 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
                 currClientFilterFactory = appendFilter(currClientFilterFactory,
                         ctx.builder.retryingHttpRequesterFilter);
             }
-            // Internal retries must be the last filter in the chain, right before LoadBalancedStreamingHttpClient.
+
+            // This filter cleans up tracked and discarded message payloads.
+            currClientFilterFactory = appendFilter(currClientFilterFactory,
+                    HttpMessageDiscardWatchdogClientFilter.INSTANCE);
+
+            // Internal retries must be one of the last filters in the chain.
             currClientFilterFactory = appendFilter(currClientFilterFactory, InternalRetryingHttpClientFilter.INSTANCE);
             FilterableStreamingHttpClient wrappedClient =
                     currClientFilterFactory.create(lbClient, lb.eventStream(), ctx.sdStatus);
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
@@ -141,7 +141,8 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
         this.loadBalancerFactory = DefaultHttpLoadBalancerFactory.Builder.<R>fromDefaults().build();
         this.serviceDiscoverer = requireNonNull(serviceDiscoverer);
 
-        clientFilterFactory = appendFilter(clientFilterFactory, HttpMessageDiscardWatchdogClientFilter.CLEANER);
+        connectionFilterFactory = HttpMessageDiscardWatchdogClientFilter.CONNECTION_CLEANER;
+        clientFilterFactory = appendFilter(clientFilterFactory, HttpMessageDiscardWatchdogClientFilter.CLIENT_CLEANER);
     }
 
     private DefaultSingleAddressHttpClientBuilder(@Nullable final U address,
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
@@ -141,7 +141,6 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
         this.loadBalancerFactory = DefaultHttpLoadBalancerFactory.Builder.<R>fromDefaults().build();
         this.serviceDiscoverer = requireNonNull(serviceDiscoverer);
 
-        connectionFilterFactory = HttpMessageDiscardWatchdogClientFilter.CONNECTION_CLEANER;
         clientFilterFactory = appendFilter(clientFilterFactory, HttpMessageDiscardWatchdogClientFilter.CLIENT_CLEANER);
     }
 
@@ -322,9 +321,6 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
                 currClientFilterFactory = appendFilter(currClientFilterFactory,
                         ctx.builder.retryingHttpRequesterFilter);
             }
-
-            currClientFilterFactory = appendFilter(currClientFilterFactory,
-                    HttpMessageDiscardWatchdogClientFilter.INSTANCE);
 
             // Internal retries must be one of the last filters in the chain.
             currClientFilterFactory = appendFilter(currClientFilterFactory, InternalRetryingHttpClientFilter.INSTANCE);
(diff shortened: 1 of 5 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
```diff
@@ -0,0 +1,132 @@
...
+                        // (i.e. during a retry) and so previous message body needs to be cleaned up.
+                        LOGGER.warn("Automatically draining previous HTTP response message body that was " +
+                                "not consumed. Users-defined retry logic must drain response payload before " +
+                                "retrying.");
+
+                        previous.ignoreElements().subscribe();
+                    }
+
+                    return response.transformMessageBody(msgPublisher -> msgPublisher.beforeSubscriber(() -> {
+                        final AtomicReference<?> maybePublisher = request.context().get(MESSAGE_PUBLISHER_KEY);
```
idelpivnitskiy, line 88: Here and in `HttpMessageDiscardWatchdogServiceFilter`, can we reuse the above `reference` instead of querying `request.context()`? Doesn't look like we ever remove `MESSAGE_PUBLISHER_KEY`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private static final class CleanerStreamingHttpClientFilterFactory implements StreamingHttpClientFilterFactory {
+        @Override
+        public StreamingHttpClientFilter create(final FilterableStreamingHttpClient client) {
+            return new StreamingHttpClientFilter(client) {
+                @Override
+                protected Single<StreamingHttpResponse> request(final StreamingHttpRequester delegate,
+                                                                final StreamingHttpRequest request) {
+                    return delegate.request(request).onErrorResume(originalThrowable -> Single.defer(() -> {
```
idelpivnitskiy, line 111: I would assume that `defer` here is not required. The function of `onErrorResume` already executes the role of `defer`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
```diff
@@ -0,0 +1,132 @@
...
+                        final AtomicReference<?> maybePublisher = request.context().get(MESSAGE_PUBLISHER_KEY);
+                        if (maybePublisher != null) {
+                            Publisher<?> message = (Publisher<?>) maybePublisher.get();
+                            if (message != null) {
+                                // No-one subscribed to the message (or there is none), so if there is a message
+                                // proactively clean it up.
+                                return message.ignoreElements().concat(Single.failed(originalThrowable));
```
idelpivnitskiy, line 118: The problem with `concat` is that if `ignoreElements` fails, we loose `originalThrowable`. Consider using a similar trick I used in proxy LB factory to preserve the `originalThrowable`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
```diff
@@ -0,0 +1,132 @@
...
+                            }
+                        }
+                        return Single.failed(originalThrowable);
```
idelpivnitskiy, line 121: For both `return` cases we also need to apply `.shareContextOnSubscribe()` at the end

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilterTest.java
```diff
@@ -0,0 +1,153 @@
...
+                        return "Just delegation, no failure";
+                    }
+                }, null),
+                Arguments.of(new ResponseTransformer() {
+                    @Override
+                    public Single<StreamingHttpResponse> apply(final StreamingHttpRequester requester,
+                                                               final StreamingHttpRequest request) {
+                        return requester
+                                .request(request)
+                                .map(dropped -> {
```
idelpivnitskiy, line 138: Consider also testing a case with `flatMap` that returns `Single.failed(...)`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
daschl: Good catch / simplification! 👍

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
daschl: Can you help me understand what you mean here exactly? As in: dropping the error from `ignoreElements` or adding one as the cause from the other?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -317,7 +320,12 @@ connectionFilterFactory, new AlpnReqRespFactoryFunc(
                 currClientFilterFactory = appendFilter(currClientFilterFactory,
                         ctx.builder.retryingHttpRequesterFilter);
             }
-            // Internal retries must be the last filter in the chain, right before LoadBalancedStreamingHttpClient.
+
+            // This filter cleans up tracked and discarded message payloads.
+            currClientFilterFactory = appendFilter(currClientFilterFactory,
+                    HttpMessageDiscardWatchdogClientFilter.INSTANCE);
```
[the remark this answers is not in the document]
daschl: See my updated code - I tried that but the tests are failing .. maybe I put it in the wrong spot? 🤔

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
idelpivnitskiy: Recovering from any error, propagating the `originalThrowable`, and suppressing any new errors. Something similar to what I had here: https://github.com/apple/servicetalk/pull/2697/files#diff-740b3f38328018b08090bca154c032069df66494a187fc201ad50b5c9693a464R178

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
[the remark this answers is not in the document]
idelpivnitskiy: No, the initial spot is perfect. But it affected some other state. There are 2 stages that rely on full response being consumed: 
1. connection state
2. LoadBalancedStreamingHttpClient state

After we moved "catch" logic to the connection filter, it intercepts original payload body publisher, but it doesn't see the result of `liftSync` operator applied by `LoadBalancedStreamingHttpClient`. Something similar to what we observed on the server-side when users have `BeforeFinallyHttpOperator` in the middle of the chain. 

It seems like we need 2 "catch" points:
1. as the last connection filter for the initial catch
2. as the last client filter that will override the message body publisher inside the atomic reference if response reached client filter successfully.

That way we can at least guarantee that we release at both levels.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
```diff
@@ -0,0 +1,134 @@
...
+                protected Single<StreamingHttpResponse> request(final StreamingHttpRequester delegate,
+                                                                final StreamingHttpRequest request) {
+                    return delegate.request(request).onErrorResume(originalThrowable -> {
+                        final AtomicReference<?> maybePublisher = request.context().get(MESSAGE_PUBLISHER_KEY);
+                        if (maybePublisher != null) {
+                            Publisher<?> message = (Publisher<?>) maybePublisher.get();
+                            if (message != null) {
+                                // No-one subscribed to the message (or there is none), so if there is a message
+                                // proactively clean it up.
+                                return message
```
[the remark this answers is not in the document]
idelpivnitskiy: In such cases, an exception is happening in a user filter. Their responsibility is to clean up the state before propagating an exception. For example, instead of 
```java
response.flatMap(r -> Single.failed(...));
```
they suppose to return:
```java
response.flatMap(r -> r.messageBody().ignoreElements().concat(Single.failed(...)));
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
```diff
@@ -0,0 +1,132 @@
...
+                    return delegate
+                            .request(request)
+                            .onErrorResume(cause -> {
+                                final AtomicReference<?> maybePublisher = request.context().get(MESSAGE_PUBLISHER_KEY);
+                                if (maybePublisher != null && maybePublisher.getAndSet(null) != null) {
+                                    // No-one subscribed to the message (or there is none), so if there is a message
+                                    // tell the user to clean it up.
+                                    LOGGER.warn("Discovered un-drained HTTP response message body which has " +
+                                            "been dropped by user code - this is a strong indication of a bug " +
+                                            "in a user-defined filter. Responses (or their message body) must " +
```
idelpivnitskiy, line 118: Consider:
```patch
Index: servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java	(revision 93a8015b44ea0f38e6be4a292b2a6730b56c42c8)
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java	(date 1697817005489)
@@ -81,7 +81,7 @@
                         // user.
                         LOGGER.warn("Discovered un-drained HTTP response message body which has " +
                                 "been dropped by user code - this is a strong indication of a bug " +
-                                "in a user-defined filter. Responses (or their message body) must " +
+                                "in a user-defined filter. Response payload (message) body must " +
                                 "be fully consumed before retrying.");
                     }
 
@@ -115,7 +115,7 @@
                                     // tell the user to clean it up.
                                     LOGGER.warn("Discovered un-drained HTTP response message body which has " +
                                             "been dropped by user code - this is a strong indication of a bug " +
-                                            "in a user-defined filter. Responses (or their message body) must " +
+                                            "in a user-defined filter. Response payload (message) body must " +
                                             "be fully consumed before discarding.");
                                 }
                                 return Single.<StreamingHttpResponse>failed(cause).shareContextOnSubscribe();
```

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilterTest.java
```diff
@@ -0,0 +1,197 @@
...
+                int numRequests = 5;
+                for (int i = 0; i < numRequests; i++) {
+                    if (expectedException == null) {
+                        StreamingHttpResponse response = client.request(client.get("/")).toFuture().get();
+                        assertEquals(HttpResponseStatus.OK, response.status());
+                        // Consume the body to release the connection back to the pool
+                        response.messageBody().ignoreElements().toFuture().get();
+                    } else {
+                        ExecutionException ex = assertThrows(ExecutionException.class,
+                                () -> client.request(client.get("/")).timeout(Duration.ofMillis(100)).toFuture().get());
```
idelpivnitskiy, line 123: This is weird, but definitely not related to changes in this PR. Consider either skipping tests that target exceptions originated by connection level filter or keeping this timeout. We can do a follow-up PR to investigate the root cause.

idelpivnitskiy (approved):
Thank you!

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilterTest.java
daschl: I'll skip the failing ones since it makes it a little awkward to test the assertions. I think in the follow-up PR we should try to re-add them back if we figure out the root cause.

PR #2711, opened 2023-09-26
Add `ConnectionObserver.ProxyConnectObserver`

idelpivnitskiy:
Motivation:

After #2699 implemented HTTP proxy CONNECT with ALPN, it made observability
confusing. It broke the contract that only one `connectionEstablished` method can
be invoked, which happens when ALPN switches from HTTP/1.1 to HTTP/2 pipeline.
Also, for a proxy tunnel with HTTP/1.1 `ConnectionObserver` events were always confusing:
users had to wait for `handshakeComplete` after `connectionEstablished` to report that
the connection is "ready".

Modifications:

- Add `ConnectionObserver.ProxyConnectObserver` that reports events related to
`CONNECT` phase that starts right after `onTransportHandshakeComplete`;
- Refactor `PipelinedLBHttpConnectionFactory` to defer invocation of
`connectionEstablished`/`multiplexedConnectionEstablished` until after the handshake
is complete to make it consistent with non-proxied connections and make sure only
one of the "established" methods is invoked;
- Removed `ProxyConnectLBHttpConnectionFactoryTest` because it targeted code paths
that do not exist anymore. Instead, enhanced `HttpsProxyTest` to cover similar use-cases.
- Enhanced `ProxyTunnel` to allow behavior customization;

Result:

Enhanced observability for proxy tunnels, `ConnectionObserver` behavior is consistent
between proxied and non-proxied use-cases.

Expected sequence of observer callbacks:
1. onNewConnection
2. onTransportHandshakeComplete
3. onProxyConnect
4. write/flush/read
5. proxyConnectComplete
6. onSecurityHandshake
7. write/flush/read
8. handshakeComplete
9. connectionEstablished or multiplexedConnectionEstablished, depending on ALPN results
10. ...
11. connectionClosed

---
Depends on #2699, review only starting from "Add ConnectionObserver.ProxyConnectObserver" commit.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectChannelSingle.java
```diff
@@ -0,0 +1,191 @@
...
+                    failSubscriber(ctx, new ProxyResponseException("Non-successful response '" + response.status() +
+                            "' from proxy on CONNECT " + connectAddress, response.status()));
+                }
+            }
+        }
+
+        @Override
+        public void channelReadComplete(final ChannelHandlerContext ctx) throws Exception {
+            if (subscriber != null) {
+                ctx.read(); // Keep requesting until finished
```
daschl, line 137: Question for my understanding: why wouldn't we just run all the logic that is currently in `userEventTriggered` in here? (so, what's the benefit in waiting for the InboundDataEndEvent?)

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AlpnChannelSingle.java
```diff
@@ -94,20 +97,34 @@ protected void configurePipeline(final ChannelHandlerContext ctx, final String p
 
         @Override
         protected void handshakeFailure(final ChannelHandlerContext ctx, final Throwable cause) {
-            LOGGER.warn("{} TLS handshake failed:", ctx.channel(), cause);
+            LOGGER.debug("{} TLS handshake failed:", ctx.channel(), cause);
```
daschl, line 111: what's the motivation behind dropping it down to debug from warn?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
```diff
@@ -161,4 +175,23 @@ private static void removeH1Handlers(final Channel channel) {
             pipeline.remove(handlerClass);
         }
     }
+
+    private static Throwable handleException(final Throwable cause, final Channel channel) {
+        // If the Channel is still active, close it in case of any error to free resources.
+        if (channel.isActive()) {
+            channel.close();
```
daschl, line 183: should the `ChannelFuture` be consumed and logged in case of something failed during close?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectChannelSingle.java
idelpivnitskiy: It's ideal to wait for the internal state machine of `HttpResponseDecoder` to complete before we notify the subscriber. Completion is signaled by `InboundDataEndEvent`. It helps us to make sure the response from the proxy is correct and does not include any payload body. If a malformed proxy responds with `content-length` header and the payload body, we will detect it as an error.
Will add error handling, a test, and a comment.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AlpnChannelSingle.java
idelpivnitskiy: I thought it's useless because on the client-side users always get an exception from the client. Thinking about it again, it's useful on the server-side. I will revert this change and see if I can extract the ALPN changes into a separate PR. It was a general problem that handshake status was not reported when ALPN is used 😞

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
idelpivnitskiy: Looking at how we close the channel in any other place we never do that. The primary `cause` is the most important exception that we should propagate. Closing the channel is the best effort. I never saw a case when it can fail. Did you?
But while we are here, we should use `ChannelCloseUtils.close(channel, cause)` instead of a regular `channel.close()`.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectChannelSingle.java
daschl: Thanks!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AlpnChannelSingle.java
daschl: 👍

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectLBHttpConnectionFactory.java
daschl: IIRC I have only seen this in the past if the shutdown procedure races the channel closes (i.e. someone closes the event loops beforehand and then the channel complains at some point that it is shut down). While this would help identify an illegal usage / race condition - thinking about it again in this use case we likely won't benefit from it at all. So let's go with `ChannelCloseUtils` 👍

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AlpnChannelSingle.java
idelpivnitskiy: Extracted ALPN-related changes to #2712. For re-review, please ignore failure propagation changes in `AlpnChannelSingle`.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NoopTransportObserver.java
```diff
@@ -71,6 +72,11 @@ public void onFlush() {
         public void onTransportHandshakeComplete() {
         }
 
+        @Override
```
daschl, line 75: Maybe for a follow-up pr: but why do we have NoopTransportObservers in transport internal and transport api? Maybe we can make the internal one use the one in API?

daschl (approved):
let's 🚢

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NoopTransportObserver.java
idelpivnitskiy: The original motivation was to avoid null-checks in our internal codebase, but overtime we found it not user-friendly to always require to return an observer. Users struggle to understand what they suppose to return when they are not interested in callbacks of sub-observer. Even if we make it public in transport-api, it's still hard to discover and easy to return your own, which will not give expected reduction of the overhead. The long-term plan is to make all callbacks that return another observer `@Nullable`.

PR #2712, opened 2023-09-27
TLS handshake failures are not reported to observers when ALPN is used

idelpivnitskiy:
Motivation:

Observed that `AlpnChannelHandler` does not report handshake failure via `SecurityHandshakeObserver`. Moreover, `SslHandler` closes the `Channel` before propagating `SslHandshakeCompletionEvent` in case the handshake fails, which makes our reporting sequence incorrect (`ConnectionObserver.connectionClosed(Throwable)` is invoked before `SecurityHandshakeObserver.handshakeFailed(Throwable)`, but shouldn't).

Modifications:

- Instead of trying to intercept `SslHandshakeCompletionEvent` for all pipeline variations, add a listener for `SslHandler.handshakeFuture()` to reliably get notifications about handshake status before the `Channel` can be closed;
- Enhance `SecurityHandshakeObserverTest` to test failure case and verify ordering of events;
- Enhance `AlpnChannelHandler` to unwrap `SSLException` the same way its parent `ApplicationProtocolNegotiationHandler` does, fail subscriber on `channelInactive` event, close channel if can not propagate `handshakeFailure` to subscriber;
- Deprecate `NettyPipelineSslUtils` methods that are not required anymore;

Result:

`SecurityHandshakeObserver.handshakeFailed(Throwable)` is always reported for all protocol configurations and always happens before `ConnectionObserver.connectionClosed(Throwable)`.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AlpnChannelSingle.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AlpnChannelSingle.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019-2020 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019-2021, 2023 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,18 +17,22 @@ package io.servicetalk.http.netty;
 
 import io.servicetalk.concurrent.SingleSource;
 import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.transport.netty.internal.ChannelCloseUtils;
 import io.servicetalk.transport.netty.internal.ChannelInitializer;
+import io.servicetalk.transport.netty.internal.StacklessClosedChannelException;
 
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPipeline;
+import io.netty.handler.codec.DecoderException;
 import io.netty.handler.ssl.ApplicationProtocolNegotiationHandler;
 import io.netty.handler.ssl.SslHandler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
+import javax.net.ssl.SSLException;
 
 import static io.servicetalk.http.netty.AlpnIds.HTTP_1_1;
 import static io.servicetalk.transport.netty.internal.ChannelCloseUtils.assignConnectionError;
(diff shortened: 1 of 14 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
```diff
@@ -70,11 +68,30 @@ public static boolean isSslEnabled(ChannelPipeline pipeline) {  // FIXME: 0.43 -
...
+    public static SSLSession extractSslSessionAndReport(// FIXME: 0.43 - remove deprecated method
+            @Nullable final SslConfig sslConfig,
+            final ChannelPipeline pipeline,
+            @SuppressWarnings("unused") final ConnectionObserver connectionObserver) {
+        return extractSslSession(sslConfig, pipeline);
+    }
+
+    /**
+     * Extracts the {@link SSLSession} from the {@link ChannelPipeline} if the handshake is already done
+     * and reports the result to {@link SecurityHandshakeObserver} if available.
```
daschl, line 85: this comment is now outdated, right?

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NettyPipelineSslUtils.java
```diff
@@ -111,22 +123,41 @@ public static SSLSession extractSslSessionAndReport(@Nullable final SslConfig ss
...
-                                                        boolean shouldReport) {
-        final SecurityHandshakeObserver observer = lookForHandshakeObserver(pipeline, shouldReport);
-        if (sslEvent.isSuccess()) {
+                                                        @SuppressWarnings("unused") boolean shouldReport) {
+        return extractSslSession(pipeline, sslEvent, failureConsumer);
+    }
+
+    /**
+     * Extracts the {@link SSLSession} from the {@link ChannelPipeline} if the {@link SslHandshakeCompletionEvent}
+     * is successful and reports the result to {@link SecurityHandshakeObserver} if available.
```
daschl, line 140: same - outdated no reporting anymore?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AlpnChannelSingle.java
```diff
@@ -93,17 +97,34 @@ protected void configurePipeline(final ChannelHandlerContext ctx, final String p
...
+            if (!failSubscriber(cause, ctx.channel())) {
+                ChannelCloseUtils.close(ctx, cause);
+            }
         }
 
         @Override
         public void exceptionCaught(final ChannelHandlerContext ctx, final Throwable cause) {
+            Throwable wrapped;
+            // This unwrapping logic is copied from the parent ApplicationProtocolNegotiationHandler
+            if (cause instanceof DecoderException && ((wrapped = cause.getCause()) instanceof SSLException)) {
```
daschl, line 109: Should the handshakeFailure also be triggered if a different exception is raised?

daschl (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AlpnChannelSingle.java
idelpivnitskiy: After we moved observer notification logic to `SslHandler.handshakeFuture()`, it doesn't matter anymore. Observer will be notified anyway.
For the pipeline, any other exception will either be propagated to the subscribed (users will see it) or channel will be closed (users already got some other exception before). 
Therefore, choice is to make it as consistent as possible to its parent implementation.

PR #2721, opened 2023-10-02
Improve exceptions thrown when an error happens during proxy `CONNECT`

idelpivnitskiy:
Motivation:

`ProxyResponseException` is marked as `RetryableException`. However, not all response statuses are retryable. For example, it doesn't make sense to auto-retry if proxy responds with 4xx (proxy auth required, bad request, etc.).
Also, because proxy-related exceptions live in `servicetalk-http-netty`, gprc-api can not access them for better mapping into gRPC status code.

Modifications:

- Move `ProxyConnectException` from `http-netty` to `http-api`;
- Deprecate `ProxyResponseException`, introduce `ProxyConnectResponseException` instead in `http-api`;
- Use a retryable variant of `ProxyConnectResponseException` only for specific status codes from a proxy server are are intended to be retryable;
- Include Channel info in every thrown `ProxyConnectException`;
- Enhance `GrpcStatusException` to recognize `ProxyConnectResponseException`;
- Extract `HttpResponseStatus` -> `GrpcStatusCode` mapping into `GrpcUtils.fromHttpStatus(...)` utility, add support for `PROXY_AUTHENTICATION_REQUIRED` status;
- Enhance tests to verify expected behavior;

Results:

1. `ProxyResponseException` and `ProxyConnectResponseException` are part of `http-api` now.
2. `http-netty` is careful when proxy related exceptions should be marked as `RetryableException` and when not.
3. grpc-api recognizes `ProxyConnectResponseException` and infers better grpc-status instead of always using `UNKNOWN`.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConnectException.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConnectException.java
@@ -13,39 +13,36 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.servicetalk.http.netty;
-
-import io.servicetalk.transport.api.RetryableException;
+package io.servicetalk.http.api;
 
 import java.io.IOException;
 
 /**
  * An exception while processing
  * <a href="https://datatracker.ietf.org/doc/html/rfc9110#section-9.3.6">HTTP/1.1 CONNECT</a> request.
+ *
+ * @see SingleAddressHttpClientBuilder#proxyAddress(Object)
  */
 public class ProxyConnectException extends IOException {
 
     private static final long serialVersionUID = 4453075928788773272L;
 
-    ProxyConnectException(final String message) {
+    /**
+     * Creates a new instance.
+     *
+     * @param message the detail message
+     */
+    public ProxyConnectException(final String message) {
         super(message);
     }
 
-    ProxyConnectException(final String message, final Throwable cause) {
+    /**
+     * Creates a new instance.
+     *
+     * @param message the detail message
+     * @param cause the original cause
+     */
+    public ProxyConnectException(final String message, final Throwable cause) {
         super(message, cause);
-    }
-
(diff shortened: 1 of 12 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConnectException.java
```diff
@@ -13,39 +13,36 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.servicetalk.http.netty;
-
-import io.servicetalk.transport.api.RetryableException;
+package io.servicetalk.http.api;
```
idelpivnitskiy, line 16: This class was not released yet (introduced in #2711). Safe to move without going through deprecation process

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectChannelSingle.java
```diff
@@ -125,26 +128,35 @@ private void sendConnectRequest(final ChannelHandlerContext ctx) {
...
+                        " Received unexpected message in the pipeline of type: " + msg.getClass().getName()));
             }
         }
 
+        private static ProxyConnectResponseException unsuccessfulResponse(final Channel channel,
+                                                                          final HttpResponseMetaData response,
+                                                                          final String connectAddress) {
+            final String message = channel + " Non-successful response '" + response.status() +
+                    "' from proxy on CONNECT " + connectAddress;
+            return response.status().statusClass() == SERVER_ERROR_5XX ?
```
daschl, line 155: I wonder if this is too granular 🤔 .. i.e. retrying a 501 is likely not helping. We could go more conservative and choose a list (i.e. 500, 502 503, 504), or ultimately just do not mark anything as retryable and let the user decide?

idelpivnitskiy: good point! adjusted

daschl (approved):

PR #2722, opened 2023-10-02
Deprecate `HttpContextKeys.HTTP_TARGET_ADDRESS_BEHIND_PROXY`

idelpivnitskiy:
Motivation:

Prior to `ConnectionObserver.ProxyConnectObserver` added in #2711, the sequence of `ConnectionObserver` events was different between regular connections and connection to the proxy tunnels.
`HttpContextKeys.HTTP_TARGET_ADDRESS_BEHIND_PROXY` (#2169) was useful to distinguish when the correct callback inside `ConnectionObserver` before reporting that a new connection is ready to take traffic. After adding `ProxyConnectObserver`, there is no need in this custom key.

Modifications:

- Deprecate `HttpContextKeys.HTTP_TARGET_ADDRESS_BEHIND_PROXY` and describe what is an appropriate replacement for users;

Result:

We will be able to remove
`HttpContextKeys.HTTP_TARGET_ADDRESS_BEHIND_PROXY` in the future releases.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
@@ -49,8 +49,8 @@ public final class HttpContextKeys {
      * @deprecated Use {@link TransportObserverConnectionFactoryFilter} to configure {@link TransportObserver} and then
      * listen {@link ConnectionObserver#onProxyConnect(Object)} callback to distinguish between a regular connection and
      * a connection to the secure HTTP proxy tunnel. For clear text HTTP proxies, consider installing a custom client
-     * filter that will populate {@link HttpRequestMetaData#context()} with a similar key or reach out to us to discuss
-     * ideas.
+     * filter that will populate {@link HttpRequestMetaData#context()} with a similar key or reach out to the
+     * ServiceTalk developers to discuss ideas.
      */
     @Deprecated
     @SuppressWarnings("DeprecatedIsStillUsed")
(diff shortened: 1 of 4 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
```diff
@@ -43,8 +46,15 @@ public final class HttpContextKeys {
      * HTTP proxy tunneling</a> and a clear text HTTP proxy, check presence of {@link ConnectionInfo#sslConfig()}.
      *
      * @see SingleAddressHttpClientBuilder#proxyAddress(Object)
+     * @deprecated Use {@link TransportObserverConnectionFactoryFilter} to configure {@link TransportObserver} and then
+     * listen {@link ConnectionObserver#onProxyConnect(Object)} callback to distinguish between a regular connection and
+     * a connection to the secure HTTP proxy tunnel. For clear text HTTP proxies, consider installing a custom client
+     * filter that will populate {@link HttpRequestMetaData#context()} with a similar key or reach out to us to discuss
```
daschl, line 52: suggestion: `or reach out to the ServiceTalk developers` vs `us`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectConnectionFactoryFilter.java
```diff
@@ -70,6 +70,7 @@ private ProxyFilter(final ConnectionFactory<ResolvedAddress, C> delegate) {
         }
 
         @Override
+        @SuppressWarnings("deprecation")
```
daschl, line 73: Should we mark the filter as a whole as deprecated too?

daschl (approved):

PR #2723, opened 2023-10-04
Add Publisher.onCompleteError

Scottmitch:
Motivation:
There are some scenarios where a stream terminating with onComplete is not expected, and translating to an error simplifies recovery.

### DIFF
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/OnCompleteErrorPublisherTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/OnCompleteErrorPublisherTest.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright © 2023 Apple Inc. and the ServiceTalk project authors
+ *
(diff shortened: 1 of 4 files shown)


servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/OnCompleteErrorPublisherTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @ParameterizedTest(name = "{displayName} [{index}] shouldThrow={0}")
+    @ValueSource(booleans = {false, true})
+    void completeToError(boolean shouldThrow) {
+        toSource(from(1)
+                .onCompleteError(() -> {
+                    if (shouldThrow) {
+                        throw DELIBERATE_EXCEPTION;
+                    }
+                    return DELIBERATE_EXCEPTION;
```
idelpivnitskiy, line 58: Consider returning `new DeliberateException` here to test that an exception thrown by `shouldThrow` branch is propagated

idelpivnitskiy (approved):

PR #2726, opened 2023-10-06
ConnectionObserver: provide ConnectionInfo on transport handshake

idelpivnitskiy:
Motivation:

If connection fails with an exception before one of the "established" methods were invoked, users don't have access to meaningful information, like `SslConfig`, SocketOptions, channelId (for correlation with wire logs), local address, etc.

Modifications:

- Add `ConnectionObserver.onTransportHandshakeComplete(ConnectionInfo)` callback that gives users a ST view of the netty's `Channel`;
- Deprecate pre-existing `ConnectionObserver.onTransportHandshakeComplete()`;

Results:

1. Users can collect more information about a connection if it's failed before "established".
2. Users can get Channel's ID for reporting `ProxyConnectObserver` events and security handshake failures.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
@@ -29,21 +29,6 @@ import javax.net.ssl.SSLSession;
 public interface ConnectionObserver {
 
     /**
-     * Callback when a connection starts initialization and {@link ConnectionInfo} becomes available.
-     * <p>
-     * This callback is invoked before any other callback on this observer.
-     *
-     * @param info {@link ConnectionInfo} for the connection that starts initialization. Note that the
-     * {@link ConnectionInfo#sslSession()} will always return {@code null} since it is called before the
-     * {@link ConnectionObserver#onSecurityHandshake() security handshake} is performed (and as a result no SSL session
-     * has been established). Also, {@link ConnectionInfo#protocol()} will return L4 (transport) protocol.
-     * Finalized {@link ConnectionInfo} will be available via {@link #connectionEstablished(ConnectionInfo)} or
-     * {@link #multiplexedConnectionEstablished(ConnectionInfo)} callbacks.
-     */
-    default void onConnectionInitialization(ConnectionInfo info) {  // FIXME: 0.43 - consider removing default impl
-    }
-
-    /**
      * Callback when {@code size} bytes are read from the connection.
      *
      * @param size size of the data chunk read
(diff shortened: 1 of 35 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -28,6 +28,21 @@
...
+     * This callback is invoked before any other callback on this observer.
+     *
+     * @param info {@link ConnectionInfo} for the connection that starts initialization. Note that the
+     * {@link ConnectionInfo#sslSession()} will always return {@code null} since it is called before the
+     * {@link ConnectionObserver#onSecurityHandshake() security handshake} is performed (and as a result no SSL session
+     * has been established). Also, {@link ConnectionInfo#protocol()} will return L4 (transport) protocol.
+     * Finalized {@link ConnectionInfo} will be available via {@link #connectionEstablished(ConnectionInfo)} or
+     * {@link #multiplexedConnectionEstablished(ConnectionInfo)} callbacks.
+     */
+    default void onConnectionInitialization(ConnectionInfo info) {  // FIXME: 0.43 - consider removing default impl
```
idelpivnitskiy, line 43: Open for better name ideas

idelpivnitskiy: The trick is that at this point `channel.localAddress()` and `channel.remoteAddress()` return `null`. This is not aligned with `ConnectionInfo` API (non-nullable methods).  Another option is to provide `onTransportHandshakeComplete(ConnectionInfo)` instead of `onTransportHandshakeComplete()`. When Channel becomes active, addresses are available.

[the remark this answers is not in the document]
idelpivnitskiy: Context for others from our offline discussion:

> I thought about that as well, but because `EarlyConnectionAcceptor` already uses `ConnectionInfo` and it's a `FunctionalInterface`, didn't want to create inconsistency. Propagation of `ExecutionContext` makes it harder indeed. I was less worried, bcz officially it's "internal" API, changes for initializers should not be visible for users in practice.

idelpivnitskiy: As we discussed offline, changing to `onTransportHandshakeComplete(ConnectionInfo)`

idelpivnitskiy:
@daschl, resolved all comments based on our discussion. I will merge this bcz I need it as a snapshot. Please review the final changes, I will submit a follow-up PR if you have comments.

daschl:
> @daschl, resolved all comments based on our discussion. I will merge this bcz I need it as a snapshot. Please review the final changes, I will submit a follow-up PR if you have comments.

LGTM 👍

PR #2742, opened 2023-11-02
Example with jax-rs error mapper

tkountis:
Motivation:

Add an example of how error mapping can be done with jax-rs to help users debug development and/or not issues.

Results:

Better example base

### DIFF
--- a/servicetalk-examples/http/jaxrs/src/main/java/io/servicetalk/examples/http/jaxrs/HelloWorldExceptionMapper.java
+++ b/servicetalk-examples/http/jaxrs/src/main/java/io/servicetalk/examples/http/jaxrs/HelloWorldExceptionMapper.java
@@ -21,7 +21,10 @@ import javax.ws.rs.ext.ExceptionMapper;
 public class HelloWorldExceptionMapper implements ExceptionMapper<Throwable> {
     @Override
     public Response toResponse(final Throwable exception) {
-        exception.printStackTrace();
+        if (exception instanceof IllegalStateException) {
+            return Response.status(Response.Status.BAD_REQUEST).build();
+        }
+
         return Response.serverError().build();
     }
 }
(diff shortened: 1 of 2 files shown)


servicetalk-examples/http/jaxrs/src/main/java/io/servicetalk/examples/http/jaxrs/HelloWorldExceptionMapper.java
```diff
@@ -0,0 +1,27 @@
...
+package io.servicetalk.examples.http.jaxrs;
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+
+public class HelloWorldExceptionMapper implements ExceptionMapper<Throwable> {
+    @Override
+    public Response toResponse(final Throwable exception) {
+        exception.printStackTrace();
+        return Response.serverError().build();
```
idelpivnitskiy, line 25: They way current example is placed is hard for users to run it and see the difference. Consider adding 2 endpoints in `HelloWorldJaxRsResource` that throw different exception types. Add proper javadoc to tell users how to run a `curl` command for each endpoint and what is expected result. Then adjust `HelloWorldExceptionMapper` to target only one of these 2 exceptions and map it to a different status code compare to the default Jersey behavior. That way users can execute 2 curl commands and observe the difference.

idelpivnitskiy (changes requested):

servicetalk-examples/http/jaxrs/src/main/java/io/servicetalk/examples/http/jaxrs/HelloWorldJaxRsResource.java
```diff
@@ -145,6 +145,28 @@ public CompletionStage<String> slowHello(@DefaultValue("world") @QueryParam("who
...
+     * @param mapped whether the exception is mapped or not.
+     * @param ctx the {@link ConnectionContext}.
+     * @return future greetings as a {@link CompletionStage} of {@link String}.
+     */
+    @GET
+    @Path("error-hello")
+    @Produces(TEXT_PLAIN)
+    public CompletionStage<String> errorHello(@DefaultValue("true") @QueryParam("mapped") final boolean mapped,
+                                              @Context final ConnectionContext ctx) {
+        return CompletableFuture.failedFuture(mapped ? new IllegalStateException() : new Exception());
```
idelpivnitskiy, line 167: `failedFuture` does not exist in Java 8

idelpivnitskiy (approved):
Everything else LGTM

PR #2744, opened 2023-11-08
Add initializer API for headers of HTTP proxy CONNECT request

idelpivnitskiy:
Motivation:

After https://github.com/apple/servicetalk/pull/2697 moved HTTP proxy `CONNECT` logic before user-defined
`ConnectionFactoryFilter`s, users lost the ability to intercept
`CONNECT` requests for the purpose of adding custom headers, like auth,
tracing, etc.

Modifications:

- Introduce `ProxyConfig` and `ProxyConfigBuilder` in `http-api` that
can be used to provide additional options for proxy behavior;
- Add `SingleAddressHttpClientBuilder.proxyConfig(...)` overload that
takes `ProxyConfig`;
- Deprecate pre-existing `SingleAddressHttpClientBuilder.proxyAddress()`
method, recommend switching to new API;
- Enhance `HttpsProxyTest` to verify that new API can be used to send
`Proxy-Authorization` header;

Result:

Users have explicit API to alter HTTP `CONNECT` request headers, if
necessary.

### DIFF
--- a/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcProxyTunnelTest.java
+++ b/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcProxyTunnelTest.java
@@ -22,6 +22,7 @@ import io.servicetalk.grpc.api.GrpcClientMetadata;
 import io.servicetalk.grpc.api.GrpcStatusCode;
 import io.servicetalk.grpc.api.GrpcStatusException;
 import io.servicetalk.http.api.HttpResponseStatus;
+import io.servicetalk.http.api.ProxyConfig;
 import io.servicetalk.http.api.ProxyConnectResponseException;
 import io.servicetalk.http.api.StreamingHttpConnectionFilter;
 import io.servicetalk.http.api.StreamingHttpRequest;
@@ -89,7 +90,7 @@ class GrpcProxyTunnelTest {
                 .listenAndAwait((Greeter.BlockingGreeterService) (ctx, request) ->
                         HelloReply.newBuilder().setMessage(GREETING_PREFIX + request.getName()).build());
         client = GrpcClients.forAddress(serverHostAndPort(serverContext))
-                .initializeHttp(httpBuilder -> httpBuilder.proxyAddress(proxyAddress)
+                .initializeHttp(httpBuilder -> httpBuilder.proxyConfig(ProxyConfig.of(proxyAddress))
                         .sslConfig(new ClientSslConfigBuilder(DefaultTestCerts::loadServerCAPem)
                                 .peerHost(serverPemHostname()).build())
                         .appendConnectionFilter(connection -> new StreamingHttpConnectionFilter(connection) {
--- a/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcProxyTunnelTest.java
+++ b/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcProxyTunnelTest.java
@@ -22,7 +22,6 @@ import io.servicetalk.grpc.api.GrpcClientMetadata;
 import io.servicetalk.grpc.api.GrpcStatusCode;
 import io.servicetalk.grpc.api.GrpcStatusException;
 import io.servicetalk.http.api.HttpResponseStatus;
-import io.servicetalk.http.api.ProxyConfig;
 import io.servicetalk.http.api.ProxyConnectResponseException;
 import io.servicetalk.http.api.StreamingHttpConnectionFilter;
 import io.servicetalk.http.api.StreamingHttpRequest;
@@ -57,6 +56,7 @@ import static io.servicetalk.http.api.HttpResponseStatus.BAD_REQUEST;
 import static io.servicetalk.http.api.HttpResponseStatus.INTERNAL_SERVER_ERROR;
 import static io.servicetalk.http.api.HttpResponseStatus.PROXY_AUTHENTICATION_REQUIRED;
 import static io.servicetalk.http.api.HttpResponseStatus.SERVICE_UNAVAILABLE;
+import static io.servicetalk.http.api.ProxyConfig.forAddress;
 import static io.servicetalk.test.resources.DefaultTestCerts.serverPemHostname;
 import static io.servicetalk.transport.netty.internal.AddressUtils.localAddress;
 import static io.servicetalk.transport.netty.internal.AddressUtils.serverHostAndPort;
@@ -90,7 +90,7 @@ class GrpcProxyTunnelTest {
                 .listenAndAwait((Greeter.BlockingGreeterService) (ctx, request) ->
                         HelloReply.newBuilder().setMessage(GREETING_PREFIX + request.getName()).build());
         client = GrpcClients.forAddress(serverHostAndPort(serverContext))
-                .initializeHttp(httpBuilder -> httpBuilder.proxyConfig(ProxyConfig.of(proxyAddress))
+                .initializeHttp(httpBuilder -> httpBuilder.proxyConfig(forAddress(proxyAddress))
                         .sslConfig(new ClientSslConfigBuilder(DefaultTestCerts::loadServerCAPem)
                                 .peerHost(serverPemHostname()).build())
                         .appendConnectionFilter(connection -> new StreamingHttpConnectionFilter(connection) {
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfig.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfig.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright © 2023 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.api;
+
+import io.servicetalk.transport.api.ClientSslConfig;
+
+import java.util.function.Consumer;
+
+/**
+ * Configuration for a proxy.
+ *
+ * @param <A> the type of address
+ */
+public interface ProxyConfig<A> {
+
+    /**
+     * Address of the proxy.
+     * <p>
+     * Usually, this is an unresolved proxy address and its type must match the type of address before resolution used
+     * by {@link SingleAddressHttpClientBuilder}. However, if the client builder was created for a resolved address,
+     * this address must also be already resolved. Otherwise, a runtime exceptions will occur.
+     *
+     * @return address of the proxy
+     */
+    A address();
+
+    /**
+     * An initializer for {@link HttpHeaders} related to
+     * <a href="https://datatracker.ietf.org/doc/html/rfc9110#section-9.3.6">HTTP/1.1 CONNECT</a> request.
+     * <p>
+     * When this {@link ProxyConfig} is used for secure proxy tunnels (when
+     * {@link SingleAddressHttpClientBuilder#sslConfig(ClientSslConfig) ClientSslConfig} is configured) the tunnel is
+     * always initialized using {@code HTTP/1.1 CONNECT} request. This {@link Consumer} can be used to set custom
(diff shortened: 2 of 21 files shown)


servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcProxyTunnelTest.java
```diff
@@ -89,7 +90,7 @@ class GrpcProxyTunnelTest {
                 .listenAndAwait((Greeter.BlockingGreeterService) (ctx, request) ->
                         HelloReply.newBuilder().setMessage(GREETING_PREFIX + request.getName()).build());
         client = GrpcClients.forAddress(serverHostAndPort(serverContext))
-                .initializeHttp(httpBuilder -> httpBuilder.proxyAddress(proxyAddress)
+                .initializeHttp(httpBuilder -> httpBuilder.proxyConfig(ProxyConfig.of(proxyAddress))
```
idelpivnitskiy, line 93: I debated between `ProxyConfig.forAddress(proxyAddress)` and `ProxyConfig.of(proxyAddress)`. Went with `of` only because wanted to be consistent with similar static method on other interfaces/classes, like `HostAndPort`, `HttpRequestMethod`, etc. If `of` doesn't read well for this use-case, happy to rename it by your recommendations.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfig.java
```diff
@@ -0,0 +1,68 @@
...
+import io.servicetalk.transport.api.ClientSslConfig;
+
+import java.util.function.Consumer;
+
+/**
+ * Configuration for a proxy.
+ *
+ * @param <A> the type of address
+ */
+public interface ProxyConfig<A> {
```
idelpivnitskiy, line 27: For others who review: this may look like an overkill to add a new "config" API while we need only one more config parameter, which could be a 2nd argument for original `proxyAddress(...)` method, like it was implemented in #2698. However, we discussed with @daschl and agreed that having a config object now (similar to `RedirectConfig`) will minimize API changes on `SingleAddressHttpClientBuilder ` in the future, when we may need to add more options, like:
1. Timeout for `CONNECT` request, similar to what Netty does: https://github.com/netty/netty/blob/7e18e8344819428a5560064ac15a2f035b4728fd/handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java#L57
2. SOCKS proxy config options

Note that `SingleAddressHttpClientBuilder` is used by users when they rely on `HttpProviders`. So, it's important to minimize API friction for them.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfigBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+            if (!(o instanceof DefaultProxyConfig)) {
+                return false;
+            }
+
+            final DefaultProxyConfig<?> that = (DefaultProxyConfig<?>) o;
+            if (!address.equals(that.address)) {
+                return false;
+            }
+            return connectRequestHeadersInitializer.equals(that.connectRequestHeadersInitializer);
```
bryce-anderson, line 110:
```suggestion
            return address.equals(that.address) &&  
            connectRequestHeadersInitializer.equals(that.connectRequestHeadersInitializer);
```
ffti: this is a bit more concise and fits the patter of how you would describe equality verbally. ("field1 are equal AND field2 are equal ...)

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfigBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+            return address;
+        }
+
+        @Override
+        public Consumer<HttpHeaders> connectRequestHeadersInitializer() {
+            return connectRequestHeadersInitializer;
+        }
+
+        @Override
+        public boolean equals(final Object o) {
```
bryce-anderson, line 98: Because one of the fields is a `Consumer`, which is going to often be set by anonymous lambda, I think notions of equality and hash-code are going to be error prone. Does this class need to have such notions?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectChannelSingle.java
```diff
@@ -193,7 +195,7 @@ public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt
             connectObserver.proxyConnectComplete(response);
             ctx.pipeline().remove(this);
             final Channel channel = ctx.channel();
-            LOGGER.debug("{} Received successful response from proxy on CONNECT {}", channel, connectAddress);
+            LOGGER.debug("{} Received successful response from proxy on CONNECT {}", channel, proxyConfig.address());
```
bryce-anderson, line 198: Does it make sense to include the name of the headers initializer? For really adversarially defined initializers I can see that leading to the case of logging headers so maybe it's not a good idea.

bryce-anderson (approved):
Other than the open question wrt `equals` and `hashCode` it looks good.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfigBuilder.java
idelpivnitskiy: this is auto-generated code by intellij idea

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfigBuilder.java
idelpivnitskiy: I think it's fine to keep best effort and we can rely on users to care about equality for consumer. If they don't have, having different lambdas will make configs different. if they care, they must provide the same instance of Consumer or override equals/hashCode for consumer.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ProxyConnectChannelSingle.java
idelpivnitskiy: No, here we care only for what address we failed or were successful. The rest of the context either should not matter or users may infer it from their codebase

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfig.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Address of the proxy.
+     * <p>
+     * Usually, this is an unresolved proxy address and its type must match the type of address before resolution used
+     * by {@link SingleAddressHttpClientBuilder}. However, if the client builder was created for a resolved address,
+     * this address must also be already resolved. Otherwise, a runtime exceptions will occur.
```
daschl, line 34:
```suggestion
     * this address must also be already resolved. Otherwise, a runtime exception will occur.
```

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfig.java
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * An initializer for {@link HttpHeaders} related to
+     * <a href="https://datatracker.ietf.org/doc/html/rfc9110#section-9.3.6">HTTP/1.1 CONNECT</a> request.
+     * <p>
+     * When this {@link ProxyConfig} is used for secure proxy tunnels (when
+     * {@link SingleAddressHttpClientBuilder#sslConfig(ClientSslConfig) ClientSslConfig} is configured) the tunnel is
+     * always initialized using {@code HTTP/1.1 CONNECT} request. This {@link Consumer} can be used to set custom
+     * {@link HttpHeaders} for that request, like {@link HttpHeaderNames#PROXY_AUTHORIZATION proxy-authorization},
+     * tracing, or any other header.
+     *
```
daschl, line 49: If (are we?) doing any retries on the connect request, this comment I think should clarified if it is called N times or just once.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfig.java
```diff
@@ -0,0 +1,68 @@
...
+     * Returns a {@link ProxyConfig} for the specified {@code address}.
+     * <p>
+     * All other {@link ProxyConfig} options will use their default values applied by {@link ProxyConfigBuilder}.
+     *
+     * @param address Address of the proxy
+     * @param <A> the type of address
+     * @return a {@link ProxyConfig} for the specified {@code address}
+     * @see ProxyConfigBuilder
+     */
+    static <A> ProxyConfig<A> of(A address) {
```
daschl, line 65: "forAddress" I think reads easier (since you are creating a proxy config for an address, not of an address), but as you mentioned using of aligns it with HostAndPort. I'm ok either way and it's not a big deal.

daschl: Another option would be to not have this static method at all and let users go through the builder all the time?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfigBuilder.java
daschl: If that's an invariant which is important for the semantics of the executed code, I think we should point this out in a javadoc somewhere.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
```diff
@@ -65,10 +65,40 @@ public interface SingleAddressHttpClientBuilder<U, R> extends HttpClientBuilder<
...
      * {@code proxyAddress} should also be already resolved – otherwise runtime exceptions may occur.
      * @return {@code this}.
+     * @deprecated Use {@link #proxyConfig(ProxyConfig)} with {@link ProxyConfig#of(Object)}.
      */
-    default SingleAddressHttpClientBuilder<U, R> proxyAddress(U proxyAddress) { // FIXME: 0.43 - remove default impl
-        throw new UnsupportedOperationException("Setting proxy address is not yet supported by "
-                + getClass().getName());
+    // FIXME: 0.43 - remove deprecated method
+    @Deprecated
+    @SuppressWarnings("DeprecatedIsStillUsed")
```
daschl, line 72: Should we keep this method for convenience? Related to my suggestions of the naming of the factory method: if we force them to go through the builder I would keep this convenience method, if we have the static factory I guess there is no harm in removing this since its going to be very short anyways.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfigBuilder.java
idelpivnitskiy: It feels kind of obvious that if users pass 2 different lambdas, the resulting "pojo" will be different. Note that there is no use-case for equals/hashCode right now to really worry about it. This is a pkg-private class, which makes it harder to document impl details. Not sure if we need to put this info on interface bcz it's not required right now.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfig.java
idelpivnitskiy: No, there are no retries specifically for a `CONNECT` request. In case of a general retry, it will go through process of opening a new connection and then it's not different from any other new connection event, regardless of if it was retried or not.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
idelpivnitskiy: Usually, users have more confusion when there are multiple overloads on the builder for the same functionality. Especially when they override each other. People don't really read javadoc and as a result, there is a high chance of confusing their understanding of runtime behavior when they use both of them:

```java
builder.proxyAddress(address)
    .proxyConfig(config)
```

Another use-case is `HttpProviders`. It makes it harder to implement a provider that need to track 2 different methods for the same functionality. As a result, I proposed a static factory.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/ProxyConfig.java
idelpivnitskiy: I had it originally and it looked quite verbose

idelpivnitskiy: I will rename to `forAddress` to make it more readable when someone uses a static import.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/SingleAddressHttpClientBuilder.java
daschl: good point, makes sense. thanks!

daschl (approved):

Issue #1217, opened 2020-11-17
HTTP 413 - when compressed payloads are bigger than the allowed limit


tkountis:
When a compressed payload is bigger than the acceptable limit, we should surface the error to the client by returning a 413

see. https://tools.ietf.org/html/rfc7231#section-6.5.11

yanisderbikov:
Hi. 
So it sounds interesting. Do you know how to find an server code for take a look and find the possible problems?

PR #2752, opened 2023-11-12
Disable `decoderEnforceMaxRstFramesPerWindow` for HTTP/2 clients

idelpivnitskiy:
Motivation:

Netty 4.1.100 introduced a new feature for HTTP/2 to protect from DDoS vector, and we added system properties to control its settings in #2728. In version 4.1.101 Netty disabled this protection for clients by default, but due to how we use our system properties, that change won't automatically apply to ServiceTalk.

Modifications:

- `OptimizedHttp2FrameCodecBuilder.decoderEnforceMaxRstFramesPerWindow` now considers `isServer` flag to decide if the inferred values should be applied or not;

Result:

`decoderEnforceMaxRstFramesPerWindow` is enabled only for HTTP/2 servers.


yanisderbikov:
Could I be helpful?

daschl (approved):
LGTM, but consider modifying the inline javadoc for the properties to mention that it only applies to the server, not the clients

idelpivnitskiy:
Hi @yanisderbikov,
Thanks for your interest in this project. I checked our [good first issues](https://github.com/apple/servicetalk/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) list and TBH, it's not in a good shape right now. #2502 is already taken. If you like digging, consider exploring #1217. Our team will try to add more issues with this label in the following weeks.

PR #2753, opened 2023-11-13
Enhance PMD rules

idelpivnitskiy:
Motivation:

Currently, `pmd/basic.xml` defines explicit set of rules to validate. With this
strategy we miss all new rules that PMD adds overtime.

Modifications:

- Rework `pmd/basic.xml` to include all rules by default but exclude those
that are not providing much value;
- Correct our codebase to satisfy all newly activated rules;

Result:

More coverage during automated code validation that will reduce nit-picking
comments during PR reviews.


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpsProxyTest.java
```diff
@@ -153,6 +153,7 @@ void tearDown() throws Exception {
         safeClose(proxyTunnel);
     }
 
+    @SuppressWarnings("PMD.AvoidPrintStackTrace")
```
bryce-anderson, line 156: Is this necessary? It seems to do the right thing.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/RetryingHttpRequesterFilterTest.java
```diff
@@ -171,7 +171,7 @@ public HttpResponse handle(final HttpServiceContext ctx, final HttpRequest reque
                              .build())
                      .appendClientFilter(new Builder()
                              .waitForLoadBalancer(true)
-                             .responseMapper(resp -> resp.status().statusClass().equals(SERVER_ERROR_5XX) ?
+                             .responseMapper(resp -> resp.status().statusClass() == SERVER_ERROR_5XX ?
```
bryce-anderson, line 174: ffti: above in [ConnectablePayloadWriter.java](https://github.com/apple/servicetalk/pull/2753/files#diff-4533b6b425f43c7ef436832242a1dcf077631fd7cb800b1f187c55afb91b012e) you used a style something like this, although I don't know that it matters.
```suggestion
                             .responseMapper(resp ->SERVER_ERROR_5XX.equals(resp.status().statusClass()) ?
```

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamObserverTest.java
```diff
@@ -124,6 +124,7 @@ void teardown() throws Exception {
         safeClose(client);
     }
 
+    @SuppressWarnings("PMD.AvoidPrintStackTrace")
```
bryce-anderson, line 127: Any reason not to use a logger below?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/TestServiceStreaming.java
```diff
@@ -75,6 +75,7 @@ final class TestServiceStreaming implements StreamingHttpService {
     }
 
     @Override
+    @SuppressWarnings("PMD.ImplicitSwitchFallThrough")
```
bryce-anderson, line 78: Is it complaining about no break in the `default` case?

servicetalk-http-router-jersey-internal/src/main/java/io/servicetalk/http/router/jersey/internal/BufferPublisherInputStream.java
```diff
@@ -147,12 +147,11 @@ public static <T> T handleEntityStream(final InputStream entityStream,
         return inputStreamHandler.apply(wrappedStream, allocator);
     }
 
-    @Nullable
     private static byte[] getBytes(final Buffer content) {
         final int readableBytes = content.readableBytes();
 
         if (readableBytes == 0) {
-            return null;
+            return EMPTY_ARRAY;
```
bryce-anderson, line 154: This is an interesting change. It seems right to me, but how did it survive this long? Maybe it's just never called with an empty buffer?

bryce-anderson (approved):
Seems good to me but you should get a second opinion on the rules you've selected.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/NoopAsyncContextProvider.java
```diff
@@ -252,7 +252,7 @@ public ContextMap copy() {
 
         @Override
         public int hashCode() {
-            return super.hashCode();
+            return System.identityHashCode(this);
```
daschl, line 255: Not a big thing, but I wonder why calling `identityHashCode` is better than `hashCode` on the parent in those cases?

servicetalk-http-api/src/test/java/io/servicetalk/http/api/AbstractHttpHeadersTest.java
```diff
@@ -657,25 +658,25 @@ void testToString() {
         headers.add("name2", "value3");
         String result = headers.toString((name, value) -> value);
         assertTrue(result.startsWith(headers.getClass().getSimpleName() + "["), result);
-        assertTrue(result.toLowerCase().contains("name1: value1"), result);
-        assertTrue(result.toLowerCase().contains("name1: value2"), result);
-        assertTrue(result.toLowerCase().contains("name2: value3"), result);
+        assertTrue(result.toLowerCase(Locale.ENGLISH).contains("name1: value1"), result);
```
daschl, line 661: Do we need to pin this to the english locale, or can we use `Locale.ROOT`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1Utils.java
```diff
@@ -249,7 +250,7 @@ static Http2Headers h1HeadersToH2Headers(HttpHeaders h1Headers) {
         for (Map.Entry<CharSequence, CharSequence> h1Entry : h1Headers) {
             // header field names MUST be converted to lowercase prior to their encoding in HTTP/2
             // https://tools.ietf.org/html/rfc7540#section-8.1.2
-            http2Headers.add(h1Entry.getKey().toString().toLowerCase(), h1Entry.getValue());
+            http2Headers.add(h1Entry.getKey().toString().toLowerCase(Locale.ENGLISH), h1Entry.getValue());
```
daschl, line 253: I wonder how this changes the behavior compared to the original method which picked the default locale?

daschl:
Other than the comments, looks good to me. Admittedly I did not read every exclude rule in detail, but I think if we want to enable more we should do in follow-up PRs since this one is quite big already and naturally touches a lot of files.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpsProxyTest.java
idelpivnitskiy: Good catch, I did these pmd changes long time ago, found out on the plain that I forgot to push them. Some other change switched from print-stack-trace to a logger. Not necessary to have this suppression anymore.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/RetryingHttpRequesterFilterTest.java
idelpivnitskiy: Bytecode checkers recommend using `==` for enums (`StatusClass` is enum) and `equals` for objects (`ConnectablePayloadWriter` uses `IOException`)

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/TestServiceStreaming.java
idelpivnitskiy: Yes, good eye. Removed suppression and added a `break`

servicetalk-http-router-jersey-internal/src/main/java/io/servicetalk/http/router/jersey/internal/BufferPublisherInputStream.java
idelpivnitskiy: yes, looks like that, glad it was caught

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/NoopAsyncContextProvider.java
idelpivnitskiy: It was complaining with `UselessOverridingMethod` warning when it used `super.hashCode()` but if I remove it, it was complaining that this class overrides `equals` but does not override `hashCode`. So, I had to use `System.identityHashCode(this)` to keep the override but fool the checker.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1Utils.java
idelpivnitskiy: Should not impact anything because header names are defined by RFC as:
```
header-field = field-name ":" OWS field-value OWS
field-name = token
token = 1*tchar
tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
ALPHA =  %x41-5A / %x61-7A   ; A-Z / a-z
DIGIT =  %x30-39 ; 0-9
```
So, it's only basic visible ASCII characters that are identical for all locales.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/StreamObserverTest.java
idelpivnitskiy: This method was reused in many different classes, it was awkward to use a logger from `StreamObserverTest` for all those cases. I moved both "safe*" methods in a new test fixture: https://github.com/apple/servicetalk/pull/2753/commits/3667b92e766b4b3563aa8bc203ab6aaae82e59ff

idelpivnitskiy:
> if we want to enable more we should do in follow-up PRs since this one is quite big already and naturally touches a lot of files

Those that are left excluded were quite noisy and didn't provide much value. I'm happy to exclude more rules and revert changes if anything looks unreasonable.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/NoopAsyncContextProvider.java
daschl: Ah I see - still weird but ok heh.

daschl (approved):

PR #2759, opened 2023-11-14
Fix bug in http/grpc logging observer that prevented request Throwables to be logged

tflobbe:
Motivation:

LoggingHttpLifecycleObserver and LoggingGrpcLifecycleObserver included a bug that prevented request exceptions to be logged.

Modifications:

- Update LoggingHttpLifecycleObserver and LoggingGrpcLifecycleObserver to include the Throwable as the last argument in the logger call
- Add tests for the two components

Result:

The Logger object will be called with a Throwable as last argument instead of null in case of request errors. Most logging frameworks implementing slf4j will print the full stacktraces in this situation.

### DIFF
--- a/servicetalk-grpc-utils/build.gradle
+++ b/servicetalk-grpc-utils/build.gradle
@@ -27,6 +27,7 @@ dependencies {
   implementation "com.google.code.findbugs:jsr305"
 
   testImplementation project(":servicetalk-test-resources")
+  testImplementation testFixtures(project(":servicetalk-concurrent-internal"))
   testImplementation "org.junit.jupiter:junit-jupiter-api"
   testImplementation "org.hamcrest:hamcrest:$hamcrestVersion"
   testImplementation "org.mockito:mockito-core:$mockitoCoreVersion"
--- a/servicetalk-grpc-utils/build.gradle
+++ b/servicetalk-grpc-utils/build.gradle
@@ -18,10 +18,19 @@ apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-library"
 
 dependencies {
   implementation platform(project(":servicetalk-dependencies"))
+  testImplementation enforcedPlatform("org.junit:junit-bom:$junit5Version")
   api project(":servicetalk-grpc-api")
 
   implementation project(":servicetalk-annotations")
   implementation project(":servicetalk-logging-slf4j-internal")
   implementation project(":servicetalk-utils-internal")
   implementation "com.google.code.findbugs:jsr305"
+
+  testImplementation project(":servicetalk-test-resources")
+  testImplementation "org.junit.jupiter:junit-jupiter-api"
+  testImplementation "org.hamcrest:hamcrest:$hamcrestVersion"
+  testImplementation "org.mockito:mockito-core:$mockitoCoreVersion"
+  testImplementation "org.mockito:mockito-junit-jupiter:$mockitoCoreVersion"
 }
+
+spotbugsTest.enabled = false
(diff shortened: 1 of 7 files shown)


servicetalk-grpc-utils/build.gradle
```diff
@@ -18,10 +18,19 @@ apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-library"
...
   implementation "com.google.code.findbugs:jsr305"
+
+  testImplementation project(":servicetalk-test-resources")
+  testImplementation "org.junit.jupiter:junit-jupiter-api"
+  testImplementation "org.hamcrest:hamcrest:$hamcrestVersion"
+  testImplementation "org.mockito:mockito-core:$mockitoCoreVersion"
+  testImplementation "org.mockito:mockito-junit-jupiter:$mockitoCoreVersion"
 }
+
+spotbugsTest.enabled = false
```
idelpivnitskiy, line 36: Why did you disable spotbugs for this module?
If it has any false positives, please add `servicetalk-grpc-utils/gradle/spotbugs/test-exclusions.xml`. The format can be copied from any other test-exclusions.xml we already have.

servicetalk-grpc-utils/src/test/java/io/servicetalk/grpc/utils/LoggingGrpcLifecycleObserverTest.java
```diff
@@ -0,0 +1,159 @@
...
+    @Test
+    void testFactoryMethod() {
+        assertThat(observer, isA(GrpcLifecycleObserver.GrpcRequestObserver.class));
+        assertThat(observer, isA(GrpcLifecycleObserver.GrpcResponseObserver.class));
+    }
+
+    @Test
+    void testOnRequestError() {
+        observer.onConnectionSelected(mock(ConnectionInfo.class));
+        Throwable requestError = new RuntimeException("fake request error");
```
idelpivnitskiy, line 77: For purposes like this one we have `DeliberateException`. You will need to add
```
testImplementation testFixtures(project(":servicetalk-concurrent-internal"))
```
dependency to access it.

servicetalk-http-utils/src/test/java/io/servicetalk/http/utils/LoggingHttpLifecycleObserverTest.java
```diff
@@ -0,0 +1,158 @@
...
+    @Test
+    void testFactoryMethod() {
+        assertThat(observer, isA(HttpLifecycleObserver.HttpRequestObserver.class));
+        assertThat(observer, isA(HttpLifecycleObserver.HttpResponseObserver.class));
+    }
+
+    @Test
+    void testOnRequestError() {
+        observer.onConnectionSelected(mock(ConnectionInfo.class));
+        Throwable requestError = new RuntimeException("fake request error");
```
idelpivnitskiy, line 77: Same in this test class

idelpivnitskiy (commented):
Thanks a lot for contributing a fix!
Some minor comments related to tests:

servicetalk-grpc-utils/build.gradle
tflobbe: Ah, makes sense. The reason I ignored spotbugs is that it was complaining about `Non-null field [...] not initialized` in the constructor for all the mocks. I'm initializing the mocks in the `@BeforeEach/setUp` code, so it doesn't belong on the constructor. I noticed other modules disabling spotbugs with `spotbugsTestProps.enabled = false` and I thought that was the reason I was seeing the issue in the `grpc` module and not the `http` one. Now that you mentioned the exclusions, it makes more sense. Adding that instead and removing the disabling line

idelpivnitskiy (approved):

idelpivnitskiy:
Thank you,
That was a good find!

PR #2772, opened 2023-12-07
Disable RRLB health-checking for `DiscoveryStrategy.ON_NEW_CONNECTION`

idelpivnitskiy:
Motivation:

When `DiscoveryStrategy.ON_NEW_CONNECTION` mode is used, LB will have a single unresolved entry, the actual IP address will be resolved by a `ConnectionFactory`. We should not run health-checking logic in this case because it can lead to "all unhealthy" state. We should run a new connection attempt all the time and fail with an exception from transport.

Modifications:

- Configure a custom `loadBalancerFactory` on a client builder when `DiscoveryStrategy.ON_NEW_CONNECTION` is used;

Result:

No health-checking for an unresolved address when
`DiscoveryStrategy.ON_NEW_CONNECTION` is used.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpClients.java
@@ -28,6 +28,7 @@ import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
 import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.http.api.DelegatingSingleAddressHttpClientBuilder;
+import io.servicetalk.http.api.FilterableStreamingHttpLoadBalancedConnection;
 import io.servicetalk.http.api.HttpClient;
 import io.servicetalk.http.api.HttpHeaderNames;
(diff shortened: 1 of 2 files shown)


bryce-anderson:
Code change looks good. Is this worth adding a test?

It looks like the load balancer id is not unique which is the same problem for all load balancers generated by `fromDefaults()`. Is that something we should address later?

idelpivnitskiy:
> It looks like the load balancer id is not unique which is the same problem for all load balancers generated by `fromDefaults()`. Is that something we should address later?

The original intend was that `id` represents who is creating a load balancer factory. When `fromDefaults()` is used, it means that ST creates it with default settings. When users create their custom LB, they suppose to provide a different `id` so that providers will see the difference between the "default one" and a "custom one". 

I'm agreed that there could be use-cases, when users need to change an LB config for a specific client, even if it was `fromDefaults()` originally. In this case, we need something more than just `id`, maybe a bundle of `id` and `targetResource`. We can keep this in mind for the new `LoadBalancerBuilder` you are working on. Not sure if there is currently a use-case for `RoundRobinLoadBalancer`.

> Is this worth adding a test?

Good call, will add one.

bryce-anderson (approved):

PR #2776, opened 2023-12-08
Handle DNS SERVFAIL differently than NXDOMAIN

tkountis:
Motivation:
An `UnknownHostException` can have multiple root causes, each with a different application logic.
By far the most prominent is due to an NXDOMAIN from the nameserver, meaning that the domain in the query doesn't not exist. Other reasons could be a SERVFAIL or timeouts during resolutions. Due to insufficient information in the error itself, the dns-client has no obvious way of distinctly handling the various causes. 

Modification:
Netty will expose that information as an `init-cause`. ST can leverage the enrichment to handle the different scenarios.

Results:
Better behavior.
As part of this effort a behavior change is expected. NXDOMAIN errors will now cause a DNS state invalidation. To retain previous behavior use the property `io.servicetalk.dns.discovery.nxdomain.invalidation` set to `false`.

Depends on https://github.com/netty/netty/pull/13721
Depends on https://github.com/netty/netty/pull/13850


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -968,6 +970,14 @@ private static <T, A> Publisher<? extends Collection<ServiceDiscovererEvent<T>>>
         });
     }
 
+    private static boolean shouldRevokeState(final Throwable t) {
+        // ISE => Subscriber exceptions (downstream of retry)
+        return t instanceof SrvAddressRemovedException || t instanceof IllegalStateException ||
+                t instanceof ClosedDnsServiceDiscovererException || (NX_DOMAIN_INVALIDATES && t.getCause() != null &&
+                t.getCause() instanceof DnsErrorCauseException &&
+                NXDOMAIN.equals(((DnsErrorCauseException) t.getCause()).getCode()));
```
bryce-anderson, line 978:
```suggestion
        return t instanceof SrvAddressRemovedException || t instanceof IllegalStateException ||
                t instanceof ClosedDnsServiceDiscovererException || (NX_DOMAIN_INVALIDATES &&
                t.getCause() instanceof DnsErrorCauseException &&
                NXDOMAIN.equals(((DnsErrorCauseException) t.getCause()).getCode()));
```
The `instanceof` check will ensure the value is not null.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/TestRecordStore.java
```diff
@@ -47,6 +50,30 @@ final class TestRecordStore implements RecordStore {
...
+            this.name = name;
+            this.type = type;
+        }
+
+        static ServFail of(final QuestionRecord question) {
+            return new ServFail(question.getDomainName(), question.getRecordType());
+        }
+    }
+
+    private final Set<ServFail> failSet = new HashSet<>(null);
```
bryce-anderson, line 67: What is this `null` about?

bryce-anderson (commented):
I'm not an expert here but it seems good to me.

bryce-anderson (commented):
This looks good to me. I'll give it the 👍 once we can get the required netty changes landed and this PR building.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018-2020 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2018-2020, 2023 Apple Inc. and the ServiceTalk project authors
```
daschl, line 2: 2024?

bryce-anderson (approved):
This seems good to me. The apparently failing test in https://github.com/apple/servicetalk/pull/2882 is curious.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
```diff
@@ -565,23 +592,25 @@ void unknownHostDiscover() throws Exception {
     }
 
     @Test
-    void singleADiscover() throws Exception {
+    void unknownHostAfterKnownDiscover() throws Exception {
```
idelpivnitskiy, line 595: Consider adding another test with `io.servicetalk.dns.discovery.nxdomain.invalidation=false`. You can make it a pkg-private builder option with default set to the system property value, then you can make changes for each test.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -968,6 +970,14 @@ private static <T, A> Publisher<? extends Collection<ServiceDiscovererEvent<T>>>
         });
     }
 
+    private static boolean shouldRevokeState(final Throwable t) {
+        // ISE => Subscriber exceptions (downstream of retry)
+        return t instanceof SrvAddressRemovedException || t instanceof IllegalStateException ||
+                t instanceof ClosedDnsServiceDiscovererException || (NX_DOMAIN_INVALIDATES &&
+                t.getCause() instanceof DnsErrorCauseException &&
+                NXDOMAIN.equals(((DnsErrorCauseException) t.getCause()).getCode()));
```
idelpivnitskiy, line 978: By creating a hard dependency on a new netty type `DnsErrorCauseException` we limit users ability to downgrade Netty version independently from ST. It usually creates a lot of classpath pain, especially if users classpath is messy. Consider either checking the type by its string name instead of `instanceof` or use a `MethodHandler` to check if `DnsErrorCauseException` type is available at runtime as part of initializing `NX_DOMAIN_INVALIDATES` value.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
```diff
@@ -442,12 +438,47 @@ void srvRecordRemovalPropagatesError() throws Exception {
...
+        recordStore.removeFail(aFail);
+        advanceTime();
+
+        List<ServiceDiscovererEvent<InetSocketAddress>> resumeSignals = subscriber.takeOnNext(1);
+        assertHasEvent(resumeSignals, ip1, targetPort, AVAILABLE);
+    }
+
     @ParameterizedTest(name = "{displayName} [{index}] srvFilterDuplicateEvents={0}")
-    @ValueSource(booleans = {false, true})
+    @ValueSource(booleans = {false})
```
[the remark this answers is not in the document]
tkountis: 🦅 good catch, debugging omission.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -1101,4 +1110,13 @@ static SrvAddressRemovedException newInstance(Class<?> clazz, String method) {
             return unknownStackTrace(new SrvAddressRemovedException(), clazz, method);
         }
     }
+
+    @Nullable
+    private static boolean parseProperty(final String name, final boolean defaultValue) {
```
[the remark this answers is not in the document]
tkountis: need the default value of true.

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
```diff
@@ -565,23 +592,25 @@ void unknownHostDiscover() throws Exception {
...
 
-        // Remove the ip
-        recordStore.removeIPv4Address(domain, DEFAULT_TTL, ip);
-        subscription.request(1);
+        recordStore.removeIPv4Addresses(targetDomain1);
         advanceTime();
-        assertThat(subscriber.awaitOnError(), instanceOf(UnknownHostException.class));
+
+        signals = subscriber.takeOnNext(1);
+        assertHasEvent(signals, ip1, EXPIRED);
```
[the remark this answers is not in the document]
tkountis: they will, regardless invalidation the error is propagated, the test was missing the check after the refactoring.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -62,6 +62,8 @@ public final class DefaultDnsServiceDiscovererBuilder implements DnsServiceDisco
      */
     @Deprecated // FIXME: 0.43 - consider removing this system property
     private static final String SKIP_BINDING_PROPERTY = "io.servicetalk.dns.discovery.netty.skipBinding";
+    private static final String NX_DOMAIN_INVALIDATES_PROPERTY = "io.servicetalk.dns.discovery.nxdomain.invalidation";
```
idelpivnitskiy, line 65: Should we mark this one as deprecated, like `SKIP_BINDING_PROPERTY` above?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -974,8 +972,9 @@ private static boolean shouldRevokeState(final Throwable t) {
         // ISE => Subscriber exceptions (downstream of retry)
         return t instanceof SrvAddressRemovedException || t instanceof IllegalStateException ||
                 t instanceof ClosedDnsServiceDiscovererException || (NX_DOMAIN_INVALIDATES &&
-                t.getCause() instanceof DnsErrorCauseException &&
-                NXDOMAIN.equals(((DnsErrorCauseException) t.getCause()).getCode()));
+                // string matching is done on purpose to avoid the hard Netty dependency
+                (t.getCause() != null && t.getCause().getClass().getSimpleName().contains("DnsErrorCauseException")) &&
```
idelpivnitskiy, line 975: Consider using the full name match (not simple) with `equals` instead of `contains`

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
```diff
@@ -593,26 +594,36 @@ void unknownHostDiscover() throws Exception {
...
-    void unknownHostAfterKnownDiscover() throws Exception {
-        setup();
-        final String targetDomain1 = "sd.domain.com";
-        final String ip1 = nextIp();
+    @ParameterizedTest(name = "{displayName} [{index}] cache={0}")
+    @ValueSource(booleans = {false, true})
+    void unknownHostAfterKnownDiscover(boolean nxInvalidation) throws Exception {
+        boolean restore = NX_DOMAIN_INVALIDATES;
+        try {
+            NX_DOMAIN_INVALIDATES = nxInvalidation;
```
idelpivnitskiy, line 602: Ideally, we should have all constants as `final`. Alternatively, consider adding a pkg-private builder method that will modify the value and then pass it to `DefaultDnsClient` as a ctor argument instead of direct dependency on a non-final constant from the builder

servicetalk-dns-discovery-netty/gradle/spotbugs/test-exclusions.xml
```diff
@@ -43,4 +43,8 @@
     <Source name="~.*Test\.java"/>
     <Bug pattern="THROWS_METHOD_THROWS_CLAUSE_THROWABLE"/>
   </Match>
+  <Match>
+    <Source name="~.*Test\.java"/>
+    <Bug pattern="ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD"/>
```
idelpivnitskiy, line 48: If you follow the above recommendation, spotbugs won't rise this error and you won't need `@SuppressWarnings("PMD.MutableStaticState")`

servicetalk-dns-discovery-netty/gradle/checkstyle/suppressions.xml
```diff
@@ -0,0 +1,24 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<!DOCTYPE suppressions PUBLIC
+    "-//Checkstyle//DTD SuppressionFilter Configuration 1.2//EN"
+    "https://checkstyle.org/dtds/suppressions_1_2.dtd">
+
+<suppressions>
+  <suppress checks="StaticVariableName"
+            files="io[\\/]servicetalk[\\/]dns[\\/]discovery[\\/]netty[\\/]DefaultDnsServiceDiscovererBuilder.java"/>
```
idelpivnitskiy, line 23: What's wrong with `StaticVariableName`?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -62,6 +62,9 @@ public final class DefaultDnsServiceDiscovererBuilder implements DnsServiceDisco
...
+    @SuppressWarnings("PMD.MutableStaticState")
+    static boolean NX_DOMAIN_INVALIDATES = parseProperty(NX_DOMAIN_INVALIDATES_PROPERTY, true);
```
idelpivnitskiy, line 67: The only thing that worries me is that if DNS server had a blip or if someone removed entries before adding new entries, DNS client could receive `NXDOMAIN`, then invalidate all entries, and then end up in an awkward state without any addresses until another resolution happens. Could you please elaborate a bit more on why it is important to invalidate addresses in this case? Can we just wait for another result set (keep using pre-existing IPs) while we retry `UnknownHostException`?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
tkountis: It's not meant to be deprecatable, we have no good understanding how this will be used in the wild just yet, so i think its too early to make that decision.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -319,6 +313,11 @@ DefaultDnsServiceDiscovererBuilder srvHostNameRepeatDelay(
             Duration initialDelay, Duration jitter) {
         this.srvHostNameRepeatInitialDelay = requireNonNull(initialDelay);
         this.srvHostNameRepeatJitter = requireNonNull(jitter);
+        ensurePositive(srvHostNameRepeatInitialDelay, "srvHostNameRepeatInitialDelay");
+        ensurePositive(srvHostNameRepeatJitter, "srvHostNameRepeatJitter");
```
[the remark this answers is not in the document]
bryce-anderson: This sounds like a bug in `repeatWithConstantBackoffDeltaJitter`.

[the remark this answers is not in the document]
bryce-anderson: I put up a PR to address it: https://github.com/apple/servicetalk/pull/2888

idelpivnitskiy (approved):

servicetalk-dns-discovery-netty/src/test/java/io/servicetalk/dns/discovery/netty/DefaultDnsClientTest.java
```diff
@@ -354,16 +354,21 @@ void srvCNAMEDuplicateAddresses() throws Exception {
                 createSrvRecord(domain, targetDomain2, targetPort, ttl));
 
         advanceTime();
-        signals = subscriber.takeOnNext(2);
-        assertHasEvent(signals, ip1, targetPort, EXPIRED);
-        assertHasEvent(signals, ip2, targetPort, EXPIRED);
+        if (DEFAULT_NX_DOMAIN_INVALIDATES) {
```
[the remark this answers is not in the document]
bryce-anderson: I suspect this is forward looking for if/when we change the default to `true` so we don't have to re-discovery all the new conditions to check. If that's the case, I personally think it's fine.

bryce-anderson (approved):

PR #2778, opened 2023-12-08
Additional fix for IPv6-derived SNI hostnames

chrisvest:
Motivation:
PR #2575 replaced colons with dots, but we may also see percentage-sign characters – `%` – in the IPv6 text representation, which would give us errors like the following:

```
java.security.cert.CertificateException: Illegal given domain name: localhost-0.0.0.0.0.0.0.0%0
	at java.base/sun.security.util.HostnameChecker.matchDNS(HostnameChecker.java:193)
	at java.base/sun.security.util.HostnameChecker.match(HostnameChecker.java:103)
	at java.base/sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:455)
	at java.base/sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:429)
	at java.base/sun.security.ssl.X509TrustManagerImpl.checkTrusted(X509TrustManagerImpl.java:283)
	at java.base/sun.security.ssl.X509TrustManagerImpl.checkServerTrusted(X509TrustManagerImpl.java:141)
	at io.netty.handler.ssl.EnhancingX509ExtendedTrustManager.checkServerTrusted(EnhancingX509ExtendedTrustManager.java:69)
	...
Caused by: java.lang.IllegalArgumentException: Contains non-LDH ASCII characters
	at java.base/java.net.IDN.toASCIIInternal(IDN.java:296)
	at java.base/java.net.IDN.toASCII(IDN.java:122)
	at java.base/javax.net.ssl.SNIHostName.<init>(SNIHostName.java:99)
	at java.base/sun.security.util.HostnameChecker.matchDNS(HostnameChecker.java:191)
	... 36 more
```

Modifications:
Add an additional `replace(char, char)` call, to replace the colons with underscores. The `replace(char, char)` method is well-optimised in the JDK, so I don't think we can do better with a single loop over an extracted char-array.

Tests were added using a variety of synthetic inet addresses, and all available host addresses, verifying that they can all be converted to valid SNI hostnames.

Result:
Users see expected `CertificateException`: No subject alternative DNS name matching <...> found.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/StreamingConnectionFactory.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/StreamingConnectionFactory.java
@@ -120,13 +120,13 @@ final class StreamingConnectionFactory {
         return config;
     }
 
-    private static String toHostAddress(final InetAddress address) {
+    static String toHostAddress(final InetAddress address) {
         final String hostAddress = address.getHostAddress();
-        // Replace colons with dots to satisfy SNIHostName validation
-        return address instanceof Inet6Address ? hostAddress.replace(':', '.') : hostAddress;
+        // Replace colons with dots, and percentages with hyphens, to satisfy SNIHostName validation
+        return address instanceof Inet6Address ? hostAddress.replace(':', '.').replace('%', '-') : hostAddress;
     }
 
-    private static boolean isValidSniHostname(String peerHost) {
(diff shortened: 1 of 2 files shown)


chrisvest:
@idelpivnitskiy Updated

idelpivnitskiy (approved):
Thank you!

PR #2798, opened 2024-01-11
Add `SslConfig.cipherSuiteFilter()`

idelpivnitskiy:
Motivation:

Sometimes users can configure ciphers that are not supported by current `SSLEngine`. As a result, they are getting a runtime exception.

Modifications:
- Add new API `SslConfig.cipherSuiteFilter()` that allows users to configure how Netty should consume the configured ciphers: take them all as-is or filter out unsupported ciphers from the passed list;
- Test new behavior;

Result:

Users can configure `SslConfig` to filter out unsupported ciphers.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
@@ -112,9 +112,20 @@ public interface SslConfig {
      * Get the cipher suites to enable, in the order of preference.
      *
      * @return the cipher suites to enable, in the order of preference.
+     * @see #cipherSuiteFilter()
      */
     @Nullable
     List<String> ciphers();
+
+    /**
+     * Defines filtering behavior for {@link #ciphers() ciphers suites}.
+     *
+     * @return filtering behavior for {@link #ciphers() ciphers suites}.
+     * @see #ciphers()
+     */
+    default CipherSuiteFilter cipherSuiteFilter() {
+        return CipherSuiteFilter.IDENTITY;
+    }
 
     /**
      * Get the size of the cache used for storing SSL session objects.
@@ -185,5 +196,20 @@ public interface SslConfig {
     // FIXME 0.43 - remove default implementation
     default int maxCertificateListBytes() {
         return 0;
+    }
+
+    /**
+     * Defines filtering logic for {@link #ciphers() ciphers suites}.
+     */
+    enum CipherSuiteFilter {
+        /**
+         * Will take all requested ciphers suites as-is without any filtering.
+         */
+        IDENTITY,
+
+        /**
+         * Will filter all requested ciphers suites out that are not supported by the current {@link SSLEngine}.
+         */
+        SUPPORTED
     }
 }
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
@@ -124,7 +124,7 @@ public interface SslConfig {
      * @see #ciphers()
      */
     default CipherSuiteFilter cipherSuiteFilter() {
-        return CipherSuiteFilter.IDENTITY;
+        return CipherSuiteFilter.PROVIDED;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
```diff
@@ -186,4 +197,19 @@ default Duration handshakeTimeout() {   // FIXME 0.43 - remove default implement
...
+        /**
+         * Will take all requested ciphers suites as-is without any filtering.
+         */
+        IDENTITY,
+
+        /**
+         * Will filter all requested ciphers suites out that are not supported by the current {@link SSLEngine}.
+         */
+        SUPPORTED
+    }
```
bryce-anderson, line 214: I think this is nice and simple, simpler than the Netty implementation, but do we want to lock ourselves out of doing something more generic in the future without breaking API? Like Netty, we can just provide these two options as default.

idelpivnitskiy: Based on Netty experience, there is no really a need to implement any other filtering strategies besides these 2 options. I picked enum instead of just a boolean to leave us a room to add more strategies later. 

If there is a well-defined need to provide the full filter API in the future, it can be additive to this enum. Some users will keep using the enum, others may opt-in for more advance API. It's just hard to come up with a use-case that requires something more than these 2.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
```diff
@@ -186,4 +197,19 @@ default Duration handshakeTimeout() {   // FIXME 0.43 - remove default implement
...
     }
+
+    /**
+     * Defines filtering logic for {@link #ciphers() ciphers suites}.
+     */
+    enum CipherSuiteFilter {
+        /**
+         * Will take all requested ciphers suites as-is without any filtering.
+         */
+        IDENTITY,
```
daschl, line 208: Consider using a different name here, I know `identity` is technically correct, but maybe something like `AS_PROVIDED` or `ALL_CONFIGURED` is more descriptive of which ones are used.

daschl: (also since the user is not providing a function as an argument, related to the `identity function`)

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
bryce-anderson: Can you give an example of what you mean by additive? I interpret that is there being a parallel method, something like `CipherSuiteFilterFunction cipherSuiteFilterV2()` (perhaps more clever names but of this form) which seems unpleasant.

idelpivnitskiy: Correct, it will be another method on `SslConfig` that returns a filtering function instead of an enum, and we will deprecate the pre-existing method. The enum will be able to stay and work as a holder for default filter implementations with an extra method that returns a function. 

I'm agreed that it will be unpleasant situation, but the likelihood of that situation is quite low. I can not imagine other way to filter ciphers except these two. Even if there will a 3rd way in the future, we can just add a new enum value instead of exposing a function.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
idelpivnitskiy: How about `PROVIDED` (with updated javadoc using similar wording) and `SUPPORTED`?
```
* Will take all {@link SslConfig#cipers() provided ciphers suites} as-is without any filtering.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
bryce-anderson: Sticking with an enum makes sense so long as we're quite certain we'll never need a custom filter. 👍

bryce-anderson (approved):

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/SslConfig.java
daschl: 👍

daschl (approved):

PR #2799, opened 2024-01-11
h2: disable auto read for stream channels

Scottmitch:
Motivation:
HTTP/2 stream/child channel use cases don't disable autoRead explicitly. This means we may read and queue more data than desirable.

Modifications:
- Disable auto read for h2 stream/child channels.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
@@ -343,6 +343,12 @@ public final class DefaultNettyConnection<Read, Write> extends NettyChannelListe
             Predicate<Object> shouldWait, UnaryOperator<Throwable> enrichProtocolError) {
         assert parent == null || parent.executionContext() == executionContext;
         assert channel.eventLoop() == toEventLoopAwareNettyIoExecutor(executionContext.ioExecutor()).eventLoopGroup();
+
+        // For h2 the parent channel must use auto read because control frames and flow controlled frames are on the
+        // same socket, and we must read in timely manner to avoid deadlock. Child channel should not use auto
+        // read as read is explicitly called by NettyChannelPublisher according to the Subscription.request(n) demand.
+        channel.config().setAutoRead(false);
+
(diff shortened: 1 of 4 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
```diff
@@ -921,6 +929,12 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
             LOGGER.debug("{} Received a user event: {}", ctx.channel(), evt);
             if (evt == CloseHandler.InboundDataEndEvent.INSTANCE) {
                 connection.nettyChannelPublisher.channelOnComplete();
+                if (connection.readOnChannelComplete) {
+                    // h2 child channel may require read() to notify of channel closure. This constraint is lifted
+                    // with netty enhancement [1] but keep this for some time to give folks time to update Netty.
+                    // [1] https://github.com/netty/netty/pull/13779
```
Scottmitch, line 935: related to https://github.com/netty/netty/pull/13779

tkountis (commented):
LGTM, please add a test to catch any re-occurrence in the future. 

Also consider adding a comment in `NettyChannelPublisher` to let the readers know that incoming flow is driven by demand and no extra guards/limits are required for the `pending` buffer.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/DefaultNettyConnection.java
idelpivnitskiy: Consider prefixing this comments with `// FIXME 0.43` to make sure we won't forget to remove it in the future

idelpivnitskiy (commented):
LGTM after @tkountis comments addressed

gradle.properties
```diff
@@ -30,7 +30,7 @@ issueManagementUrl=https://github.com/apple/servicetalk/issues
 ciManagementUrl=https://github.com/apple/servicetalk/actions
 
 # dependency versions
-nettyVersion=4.1.104.Final
+nettyVersion=4.1.105.Final-SNAPSHOT
```
Scottmitch, line 33: depends upon https://github.com/netty/netty/pull/13779 to avoid SOOE

PR #2804, opened 2024-01-15
ClientEffectiveStrategyTest: Attempt at fixing flakyness

daschl:
Based on initial analysis the ClientInvokingThreadRecorder's state gets written and read from multiple threads, but not all of it is properly thread-safe. The invokingThreads and errors use thread-safe datastructures, but the offloadPoints do not.

Since the offloadPoints are on the critical path for the flaky test failures, let's start by making them thread safe and check if the errors still show up.

If they still persist, there might be more coordination needed between the recording and checking threads as well.


daschl:
Linking #2245

bryce-anderson (approved):
Worth a try.

PR #2816, opened 2024-01-26
LoadBalancer wire the tracker on the host layers with the request flow

tkountis:
Motivation

In order for the new load balancer features to work, the health indicators need to be wired as request trackers on the request flow. 

Modification

- Added additional configuration options in the DefaultHttpLoadBalancerFactory to support error classifications. Feature only useful to load-balancers that support it.
- When the Context is available and a request-tracker is present in it, the request flow is now enriched with additional logic to track the state and feed it to the request-tracker.


servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/DefaultLoadBalancerBuilder.java
```diff
@@ -148,6 +160,126 @@ public LoadBalancerFactory<ResolvedAddress, C> build() {
                 loadBalancerObserver, healthCheckerSupplier);
     }
 
+    public static final class DefaultHttpLoadBalancerFactory<ResolvedAddress> implements HttpLoadBalancerFactory<ResolvedAddress> {
```
bryce-anderson, line 163: In my minds eye, making a new `DefaultHttpLoadBalancerFactory` makes it tough for users to switch to the new load balancer. The `HttpClients` class uses the `DefaultHttpLoadBalancerFactory` internally: could we add the piping to that class? I imagine we would need to massage both `toLoadBalancedConnection` methods at that point but it seems fine to do. I think we'd need to open up a few more types to being public.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/LoadBalancers.java
```diff
@@ -23,7 +23,7 @@
 /**
  * A factory to create {@link DefaultLoadBalancer DefaultLoadBalancers}.
  */
-final class LoadBalancers {
+public final class LoadBalancers {
```
bryce-anderson, line 26: I think we need to do this eventually but I'm not seeing how it's related to this change.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/DefaultHttpLoadBalancedConnection.java
```diff
@@ -0,0 +1,229 @@
...
+    public Completable closeAsync() {
+        return delegate.closeAsync();
+    }
+
+    @Override
+    public StreamingHttpRequest newRequest(final HttpRequestMethod method, final String requestTarget) {
+        return delegate.newRequest(method, requestTarget);
+    }
+
+    private static final class StreamingHttpResponseSubscriber implements Subscriber<StreamingHttpResponse> {
```
bryce-anderson, line 169: @idelpivnitskiy pointed out that we're not fully tracking the response: we're only getting the head of the response. He showed me the `BeforeFinallyHttpOperator` which has the property of waiting until the body of the response has been received before finishing its operations, but it doesn't have a way to inspect the response such as in the case of a 5xx response. We don't have to make it perfect right now but we should keep it in mind at a minimum.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RootSwayingLeafRequestTracker.java
```diff
@@ -15,22 +15,31 @@
...
+
 import static java.util.Objects.requireNonNull;
 
 /**
  * A two-level request tracker, namely root and leaf.
  * Each tracking interaction influences both levels, but reporting operations will only
  * consult the leaf.
  */
-final class RootSwayingLeafRequestTracker implements RequestTracker {
+final class RootSwayingLeafRequestTracker implements RequestTracker, ScoreSupplier {
```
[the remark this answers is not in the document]
tkountis: Discussed offline, will keep this effort targeted to the Host layer, and we can come back to the connection layer as a separate PR.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/DefaultHost.java
```diff
@@ -450,6 +456,11 @@ public Completable onClosing() {
         return closeable.onClosing();
     }
 
+    @Nullable
+    HealthIndicator healthIndicator() {
+        return healthIndicator;
+    }
+
```
bryce-anderson, line 463: Is this still necessary?

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/DefaultHttpLoadBalancedConnection.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.function.Function;
+import javax.annotation.Nullable;
+
+import static io.servicetalk.http.api.HttpApiConversions.toReservedBlockingConnection;
+import static io.servicetalk.http.api.HttpApiConversions.toReservedBlockingStreamingConnection;
+import static io.servicetalk.http.api.HttpApiConversions.toReservedConnection;
+import static io.servicetalk.loadbalancer.ErrorClass.LOCAL_ORIGIN_REQUEST_FAILED;
+import static java.util.concurrent.atomic.AtomicIntegerFieldUpdater.newUpdater;
+
+final class DefaultHttpLoadBalancedConnection
```
bryce-anderson, line 49: This seems like broken/dead code. It doesn't seem to compile and it looks like it is the embodiment of the changes in the same named static class in DefaultHttpLoadBalancerFactory.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -199,19 +371,44 @@ public Completable closeAsync() {
...
+            private final RequestTracker original;
+            private volatile int done;
+
+            private AtMostOnceDeliveryRequestTracker(final RequestTracker original) {
+                this.original = original;
+            }
+
+            @Override
+            public long beforeStart() {
+                if (doneUpdater.compareAndSet(this, 0, 1)) {
```
bryce-anderson, line 393: Is this right? It seems like we can only call one of the three methods but we really need to call `beforeStart()` once then one of the other two exactly once after that.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -171,7 +309,41 @@ public <T> Publisher<? extends T> transportEventStream(final HttpEventKey<T> eve
...
+
+                return delegate.request(request)
+                        .flatMap(response -> {
+                            final ErrorClass eClass = peerResponseErrorClassifier.apply(response);
+                            if (eClass != null) {
+                                // The onError is triggered before the body is actually consumed.
+                                theTracker.onError(startTime, eClass);
+                            }
+                            return Single.succeeded(response);
+                        })
```
bryce-anderson, line 328: I like this combining the 'once' behavior to do the classification and also correctly get the terminal events.

Not a correctness thing, but would a `.map` operator be slightly more efficient than `.flatMap`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
tkountis: yeah it was late :-) 
I fixed it but never pushed.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/DefaultHost.java
```diff
@@ -218,59 +218,65 @@ public boolean markExpired() {
...
-                                            + " could not be reserved.",
-                                    RoundRobinLoadBalancer.class, "selectConnection0(...)")))
-                            .shareContextOnSubscribe();
-                }
+                                            + " was rejected by the selection filter.",
+                                    RoundRobinLoadBalancer.class, "selectConnection0(...)"));
 
-                // Invoke the selector before adding the connection to the pool, otherwise, connection can be
-                // used concurrently and hence a new connection can be rejected by the selector.
-                if (!selector.test(newCnx)) {
```
idelpivnitskiy, line 249: Why this part was removed?
Consider adding a comment in the code clarifying why we do not test `newCnx` now

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -74,10 +105,29 @@ public LoadBalancer<FilterableStreamingHttpLoadBalancedConnection> newLoadBalanc
         return rawFactory.newLoadBalancer(eventPublisher, connectionFactory, targetResource);
     }
 
-    @Override   // FIXME: 0.43 - remove deprecated method
+    @Override
     public FilterableStreamingHttpLoadBalancedConnection toLoadBalancedConnection(
-            final FilterableStreamingHttpConnection connection) {
```
idelpivnitskiy, line 79: Please revert, this is ABI incompatible change:
```
Comparing binary compatibility of servicetalk-http-netty-0.42.42-SNAPSHOT.jar against servicetalk-http-netty-0.42.41.jar
***! MODIFIED CLASS: PUBLIC FINAL io.servicetalk.http.netty.DefaultHttpLoadBalancerFactory  (not serializable)
        ===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
        ===  UNCHANGED SUPERCLASS: java.lang.Object (<- java.lang.Object)
        ---! REMOVED METHOD: PUBLIC(-) io.servicetalk.http.api.FilterableStreamingHttpLoadBalancedConnection toLoadBalancedConnection(io.servicetalk.http.api.FilterableStreamingHttpConnection)
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -74,10 +105,29 @@ public LoadBalancer<FilterableStreamingHttpLoadBalancedConnection> newLoadBalanc
...
-        return new DefaultFilterableStreamingHttpLoadBalancedConnection(connection);
+            final FilterableStreamingHttpConnection connection,
+            final ReservableRequestConcurrencyController concurrencyController,
+            @Nullable final ContextMap context) {
+
+        RequestTracker hostHealthIndicator = null;
+        if (context == null) {
+            LOGGER.debug("Context is null. In order for " + DefaultHttpLoadBalancerFactory.class.getSimpleName() +
```
idelpivnitskiy, line 116: 1. Consider adding assertion in case this can be surfaced by a test.
2. Consider increasing the log level to info. If it's unexpected, we should know about it.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -74,10 +105,29 @@ public LoadBalancer<FilterableStreamingHttpLoadBalancedConnection> newLoadBalanc
...
+                    ":toLoadBalancedConnection to get access to the " + RequestTracker.class.getSimpleName() +
+                    ", health-monitor of this connection, the context must not be null.");
+        } else {
+            hostHealthIndicator = context.get(REQUEST_TRACKER_KEY);
+            if (hostHealthIndicator == null) {
+                LOGGER.debug(REQUEST_TRACKER_KEY.name() + " is not set in context. " +
```
idelpivnitskiy, line 122: same comments here

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -102,13 +157,48 @@ private Builder(
...
+         * {@link ErrorClass}.
+         * <p>
+         * <strong>Note: </strong>Enabling this feature has no effect to {@link LoadBalancer}s that are not evaluating
+         * it.
+         *
+         * @param errorClassifier A {@link Function} to check if a received response should be considered an
+         * error.
+         * @return {@code this}.
+         */
+        public Builder<ResolvedAddress> errorClassifier(
```
idelpivnitskiy, line 188: Any chance we can defer adding these methods until we have more confidence in new API? If we add them now, it will be harder to remove/change them later.

I'm ok to consider a dirty hack of making them pkg private. If you will ever need it in any use-case, you can create an external utility for `io.servicetalk.http.netty` package to get access to these methods.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -171,7 +309,41 @@ public <T> Publisher<? extends T> transportEventStream(final HttpEventKey<T> eve
...
+            return Single.defer(() -> {
+                if (tracker == null) {
+                    return delegate.request(request).shareContextOnSubscribe();
+                }
+
+                final RequestTracker theTracker = new AtMostOnceDeliveryRequestTracker(tracker);
+                final long startTime = theTracker.beforeStart();
+
+                return delegate.request(request)
+                        .map(response -> {
```
idelpivnitskiy, line 321: `map` should be applied after `BeforeFinallyHttpOperator`, see `AbstractLifecycleObserverHttpFilter` and borrow a comment from there.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -171,7 +309,41 @@ public <T> Publisher<? extends T> transportEventStream(final HttpEventKey<T> eve
...
+                            final ErrorClass eClass = peerResponseErrorClassifier.apply(response);
+                            if (eClass != null) {
+                                // The onError is triggered before the body is actually consumed.
+                                theTracker.onError(startTime, eClass);
+                            }
+                            return response;
+                        })
+                        .liftSync(new BeforeFinallyHttpOperator(new TerminalSignalConsumer() {
```
idelpivnitskiy, line 329: If you use `BeforeFinallyHttpOperator(TerminalSignalConsumer, true)`, there is no need to wrap tracker with `AtMostOnceDeliveryRequestTracker`

idelpivnitskiy (changes requested):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -171,7 +309,41 @@ public <T> Publisher<? extends T> transportEventStream(final HttpEventKey<T> eve
 
         @Override
         public Single<StreamingHttpResponse> request(final StreamingHttpRequest request) {
-            return delegate.request(request);
+            return Single.defer(() -> {
```
idelpivnitskiy, line 312: Btw, if you don't need public API for error classifiers at the first iteration, there might be even better place for this logic. Consider `LoadBalancedStreamingHttpClient` that already uses `BeforeFinallyHttpOperator` that you can take advantage of and there is also no need for `defer`.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/DefaultHost.java
tkountis: It wasn't - diff mislead?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
tkountis: Until the new LB becomes the default, these can cause unnecessary noise if added in a higher log level. For current defaults, they aren't needed. We can do so when that changes.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
tkountis: Agreed, removed the public APIs and left the defaults for the wiring. We can add the APIs back when we are done with integration evaluation.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
tkountis: The `AtMostOnceDeliveryRequestTracker` is still needed for the `map` the successful response -> error use-case.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
tkountis: Good idea, but the new APIs (which were removed for this round) will more likely than not, be added in the near future. Which means there will be wiring involved to get these into that layer. Therefore, I chose to keep everything in the current placement.

bryce-anderson (approved):
Thanks @tkountis!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -80,6 +113,31 @@ public FilterableStreamingHttpLoadBalancedConnection toLoadBalancedConnection(
...
+            if (hostHealthIndicator == null) {
+                LOGGER.debug(REQUEST_TRACKER_KEY.name() + " is not set in context. " +
+                        "In order for " + DefaultHttpLoadBalancerFactory.class.getSimpleName() +
+                        ":toLoadBalancedConnection to get access to the " + RequestTracker.class.getSimpleName() +
+                        ", health-monitor of this connection, the context must be properly wired.");
+            }
+        }
+
+        return new DefaultHttpLoadBalancedConnection(connection, concurrencyController,
+                errorClassFunction, peerResponseErrorClassifier, hostHealthIndicator);
```
idelpivnitskiy, line 138: When there are no hostHealthIndicator, we should use `HttpLoadBalancerFactory.DefaultFilterableStreamingHttpLoadBalancedConnection` to keep the existing path as-is.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -214,4 +278,188 @@ public String toString() {
             return delegate.toString();
         }
     }
+
+    private static final class DefaultHttpLoadBalancedConnection
```
idelpivnitskiy, line 282: This new class should also have `toString()` impl. Otherwise, it removes netty's channel info from logs.
We have tests that verify it for the regular path, they didn't trigger assertions bcz they don't go through new lb.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -214,4 +278,188 @@ public String toString() {
...
+        @Override
+        public <T> Publisher<? extends T> transportEventStream(final HttpEventKey<T> eventKey) {
+            return delegate.transportEventStream(eventKey);
+        }
+
+        @Override
+        public Single<StreamingHttpResponse> request(final StreamingHttpRequest request) {
+            return Single.defer(() -> {
+                if (tracker == null) {
+                    return delegate.request(request).shareContextOnSubscribe();
```
idelpivnitskiy, line 362: bcz the presence of tracker is known upfront and never changes, consider adding this check before using `Single.defer`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
idelpivnitskiy: Another option could be to add a new pkg-private ctor for `HttpLoadBalancerFactory.DefaultFilterableStreamingHttpLoadBalancedConnection` instead of creating a new class.

idelpivnitskiy (commented):
few more comments, then lgtm:

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
tkountis: We will need to override the concurrency controller callbacks, and have null checks to keep old/new behavior. I think this looks cleaner and its easier to remove when we clean the deprecation.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -80,6 +113,35 @@ public FilterableStreamingHttpLoadBalancedConnection toLoadBalancedConnection(
...
+            if (hostHealthIndicator == null) {
+                LOGGER.debug(REQUEST_TRACKER_KEY.name() + " is not set in context. " +
+                        "In order for " + DefaultHttpLoadBalancerFactory.class.getSimpleName() +
+                        ":toLoadBalancedConnection to get access to the " + RequestTracker.class.getSimpleName() +
+                        ", health-monitor of this connection, the context must be properly wired.");
+            }
+        }
+
+        if (hostHealthIndicator == null) {
+            return HttpLoadBalancerFactory.super.toLoadBalancedConnection(connection, concurrencyController, context);
```
idelpivnitskiy, line 138: Here we need to call `new DefaultFilterableStreamingHttpLoadBalancedConnection(connection, concurrencyController)`. The super impl has more code for backward compatibility in case users have their own `HttpLoadBalancerFactory`, but here we are in final implementation.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpLoadBalancerFactory.java
```diff
@@ -80,6 +113,35 @@ public FilterableStreamingHttpLoadBalancedConnection toLoadBalancedConnection(
...
+            if (hostHealthIndicator == null) {
+                LOGGER.debug(REQUEST_TRACKER_KEY.name() + " is not set in context. " +
+                        "In order for " + DefaultHttpLoadBalancerFactory.class.getSimpleName() +
+                        ":toLoadBalancedConnection to get access to the " + RequestTracker.class.getSimpleName() +
+                        ", health-monitor of this connection, the context must be properly wired.");
+            }
+        }
+
+        if (hostHealthIndicator == null) {
+            return new DefaultFilterableStreamingHttpLoadBalancedConnection(connection);
```
idelpivnitskiy, line 138:
```suggestion
            return new HttpLoadBalancerFactory.DefaultFilterableStreamingHttpLoadBalancedConnection(connection, concurrencyController);
```

idelpivnitskiy (approved):

PR #2859, opened 2024-03-05
Support for Optional TLS negotiation for http/1 and http/2

daschl:
Motivation
----------
In certain deployment scenarios it can be useful to accept incoming traffic on the same port for both TLS and non-TLS connections. ServiceTalk does not support this at this point, so this changeset aims to add the feature.

Modifications
-------------

User-facing, a new option is introduced on the ServerSslConfig called "acceptInsecureConnections". When set to true, http/1 and http/2 connections are accepted both encrypted and unencrypted.

This still works with SNI and ALPN, since the negotiation happens when the connection is first accepted and the first bytes are checked to see if a SSL request is made or not.

Internally, the OptionalSSLHandler will do the check and afterwards the OptionalSslNegotiator will add the corresponding handlers depending on the protocol and tls/non-tls to the pipeline. At this point the other code takes over.

Tests have been added to validate the feature and avoid regressions.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
@@ -31,6 +31,7 @@ import io.servicetalk.transport.api.IoExecutor;
 import io.servicetalk.transport.api.LateConnectionAcceptor;
 import io.servicetalk.transport.api.ServerSslConfig;
 import io.servicetalk.transport.api.ServiceTalkSocketOptions;
+import io.servicetalk.transport.api.SslListenMode;
 import io.servicetalk.transport.api.TransportObserver;
 
 import java.net.SocketOption;
@@ -98,6 +99,16 @@ public interface HttpServerBuilder {
                                         int maxClientHelloLength, Duration clientHelloTimeout) {
         throw new UnsupportedOperationException(
                 "sslConfig(ServerSslConfig, Map, int, Durations) is not supported by " + getClass());
+    }
+
+    /**
+     * Allows to configure the listen mode when SSL is enabled.
+     *
+     * @param mode the mode to use.
+     * @return {@code this}.
+     */
+    default HttpServerBuilder sslListenMode(final SslListenMode mode) {
+        throw new UnsupportedOperationException("sslListenMode(SslListenMode) is not supported by " + getClass());
     }
 
     /**
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
@@ -399,7 +399,11 @@ final class DefaultHttpServerBuilder implements HttpServerBuilder {
         ReadOnlyHttpServerConfig roConfig = config.asReadOnly();
         StreamingHttpService filteredService = applyInternalFilters(service, roConfig.lifecycleObserver());
 
-        if (roConfig.tcpConfig().isAlpnConfigured()) {
+        final ServerSslConfig sslConfig = roConfig.tcpConfig().sslConfig();
+        if (sslConfig != null && sslConfig.acceptInsecureConnections()) {
+            return OptionalSslNegotiator.bind(executionContext, config, address, connectionAcceptor, filteredService,
+                    drainRequestPayloadBody, earlyConnectionAcceptor, lateConnectionAcceptor);
+        } else if (roConfig.tcpConfig().isAlpnConfigured()) {
             return DeferredServerChannelBinder.bind(executionContext, roConfig, address, connectionAcceptor,
                     filteredService, drainRequestPayloadBody, false, earlyConnectionAcceptor, lateConnectionAcceptor);
         } else if (roConfig.tcpConfig().sniMapping() != null) {
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
@@ -49,6 +49,7 @@ import io.servicetalk.transport.api.ExecutionStrategyInfluencer;
 import io.servicetalk.transport.api.IoExecutor;
 import io.servicetalk.transport.api.LateConnectionAcceptor;
 import io.servicetalk.transport.api.ServerSslConfig;
+import io.servicetalk.transport.api.SslListenMode;
 import io.servicetalk.transport.api.TransportObserver;
 import io.servicetalk.transport.netty.internal.InfluencerConnectionAcceptor;
 
@@ -218,6 +219,12 @@ final class DefaultHttpServerBuilder implements HttpServerBuilder {
     public HttpServerBuilder sslConfig(final ServerSslConfig defaultConfig, final Map<String, ServerSslConfig> sniMap,
                                        final int maxClientHelloLength, final Duration clientHelloTimeout) {
         this.config.tcpConfig().sslConfig(defaultConfig, sniMap, maxClientHelloLength, clientHelloTimeout);
+        return this;
+    }
+
+    @Override
+    public HttpServerBuilder sslListenMode(final SslListenMode mode) {
+        this.config.tcpConfig().sslListenMode(mode);
         return this;
     }
 
@@ -399,8 +406,10 @@ final class DefaultHttpServerBuilder implements HttpServerBuilder {
         ReadOnlyHttpServerConfig roConfig = config.asReadOnly();
         StreamingHttpService filteredService = applyInternalFilters(service, roConfig.lifecycleObserver());
 
-        final ServerSslConfig sslConfig = roConfig.tcpConfig().sslConfig();
-        if (sslConfig != null && sslConfig.acceptInsecureConnections()) {
+        // Important: This code needs to be kept in-sync with the similar, but not identical
+        // code inside OptionalSslNegotiator#bind
+        if (roConfig.tcpConfig().sslConfig() != null &&
+                roConfig.tcpConfig().sslListenMode() == SslListenMode.SSL_OPTIONAL) {
             return OptionalSslNegotiator.bind(executionContext, config, address, connectionAcceptor, filteredService,
                     drainRequestPayloadBody, earlyConnectionAcceptor, lateConnectionAcceptor);
         } else if (roConfig.tcpConfig().isAlpnConfigured()) {
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
@@ -16,7 +16,6 @@
 package io.servicetalk.http.netty;
 
 import io.servicetalk.concurrent.SingleSource;
-import io.servicetalk.transport.netty.internal.ChannelInitializer;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.Channel;
@@ -32,8 +31,8 @@ import javax.annotation.Nullable;
 
 final class OptionalSslChannelSingle extends ChannelInitSingle<Boolean> {
 
-    OptionalSslChannelSingle(final Channel channel, final ChannelInitializer channelInitializer) {
-        super(channel, channelInitializer);
+    OptionalSslChannelSingle(final Channel channel) {
+        super(channel, NoopChannelInitializer.INSTANCE);
     }
 
     @Override
@@ -41,7 +40,7 @@ final class OptionalSslChannelSingle extends ChannelInitSingle<Boolean> {
         return new OptionalSslHandler(subscriber);
     }
 
-    static final class OptionalSslHandler extends ByteToMessageDecoder {
+    private static final class OptionalSslHandler extends ByteToMessageDecoder {
 
         private static final Logger LOGGER = LoggerFactory.getLogger(OptionalSslHandler.class);
 
@@ -81,37 +80,7 @@ final class OptionalSslChannelSingle extends ChannelInitSingle<Boolean> {
             final SingleSource.Subscriber<? super Boolean> subscriberCopy = subscriber;
             subscriber = null;
             subscriberCopy.onSuccess(isEncrypted);
-        }
-
-        @Override
-        public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) throws Exception {
-            if (evt instanceof PipelineInitializedEvent) {
-                ctx.pipeline().remove(this);
-            } else {
-                ctx.fireUserEventTriggered(evt);
-            }
-        }
-    }
-
-    /**
-     * {@link ChannelInitializer} that does not do anything.
-     */
-    static final class NoopChannelInitializer implements ChannelInitializer {
-
-        static final ChannelInitializer INSTANCE = new NoopChannelInitializer();
-
-        private NoopChannelInitializer() {
-            // Singleton
-        }
-
-        @Override
-        public void init(final Channel channel) {
-            // NOOP
-        }
-
-        @Override
-        public ChannelInitializer andThen(final ChannelInitializer after) {
-            return after;
+            ctx.pipeline().remove(this);
         }
     }
 }
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright © 2024 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+import io.servicetalk.concurrent.SingleSource;
+import io.servicetalk.transport.netty.internal.ChannelInitializer;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.ByteToMessageDecoder;
+import io.netty.handler.ssl.SslHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import javax.annotation.Nullable;
+
+final class OptionalSslChannelSingle extends ChannelInitSingle<Boolean> {
+
+    OptionalSslChannelSingle(final Channel channel, final ChannelInitializer channelInitializer) {
+        super(channel, channelInitializer);
+    }
+
+    @Override
+    protected ChannelHandler newChannelHandler(final Subscriber<? super Boolean> subscriber) {
+        return new OptionalSslHandler(subscriber);
+    }
+
+    static final class OptionalSslHandler extends ByteToMessageDecoder {
+
+        private static final Logger LOGGER = LoggerFactory.getLogger(OptionalSslHandler.class);
+
+        /**
+         * the length of the ssl record header (in bytes)
+         */
+        private static final int SSL_RECORD_HEADER_LENGTH = 5;
+
+        @Nullable
+        SingleSource.Subscriber<? super Boolean> subscriber;
+
+        OptionalSslHandler(final SingleSource.Subscriber<? super Boolean> subscriber) {
+            this.subscriber = subscriber;
+        }
+
+        @Override
+        public void handlerAdded(final ChannelHandlerContext ctx) throws Exception {
+            if (ctx.channel().isActive()) {
+                ctx.read(); // we need to force a read to detect SSL yes/no
+            }
+            super.handlerAdded(ctx);
+        }
+
+        @Override
+        public void channelActive(final ChannelHandlerContext ctx) throws Exception {
+            ctx.read(); // we need to force a read to detect SSL yes/no
+            ctx.fireChannelActive();
+        }
+
+        @Override
+        protected void decode(final ChannelHandlerContext ctx, final ByteBuf in, final List<Object> out) {
+            if (in.readableBytes() < SSL_RECORD_HEADER_LENGTH || subscriber == null) {
+                return;
+            }
+            boolean isEncrypted = SslHandler.isEncrypted(in);
+            LOGGER.debug("{} Detected TLS for this connection: {}", ctx.channel(), isEncrypted);
+            final SingleSource.Subscriber<? super Boolean> subscriberCopy = subscriber;
+            subscriber = null;
+            subscriberCopy.onSuccess(isEncrypted);
+        }
+
+        @Override
+        public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) throws Exception {
+            if (evt instanceof PipelineInitializedEvent) {
+                ctx.pipeline().remove(this);
+            } else {
+                ctx.fireUserEventTriggered(evt);
+            }
+        }
+    }
+
+    /**
+     * {@link ChannelInitializer} that does not do anything.
+     */
+    static final class NoopChannelInitializer implements ChannelInitializer {
+
+        static final ChannelInitializer INSTANCE = new NoopChannelInitializer();
+
+        private NoopChannelInitializer() {
+            // Singleton
+        }
+
+        @Override
+        public void init(final Channel channel) {
+            // NOOP
+        }
+
+        @Override
+        public ChannelInitializer andThen(final ChannelInitializer after) {
+            return after;
+        }
+    }
+}
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
@@ -29,6 +29,9 @@ import org.slf4j.LoggerFactory;
 import java.util.List;
 import javax.annotation.Nullable;
 
+/**
+ * Part of a {@link io.netty.channel.ChannelInitializer} which negotiates SSL/non-SSL connections when SSL is enabled.
+ */
 final class OptionalSslChannelSingle extends ChannelInitSingle<Boolean> {
 
     OptionalSslChannelSingle(final Channel channel) {
@@ -80,6 +83,10 @@ final class OptionalSslChannelSingle extends ChannelInitSingle<Boolean> {
             final SingleSource.Subscriber<? super Boolean> subscriberCopy = subscriber;
             subscriber = null;
             subscriberCopy.onSuccess(isEncrypted);
+
+            // Need to make sure that when this handler is removed, there is another handler in the pipeline
+            // to pick up the read bytes from ByteToMessageDecoder when this handler is removed.
+            assert ctx.pipeline().last() != this;
             ctx.pipeline().remove(this);
         }
     }
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslNegotiator.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslNegotiator.java
@@ -17,26 +17,41 @@ package io.servicetalk.http.netty;
 
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.http.api.HttpExecutionContext;
+import io.servicetalk.http.api.HttpLifecycleObserver;
 import io.servicetalk.http.api.HttpServerContext;
 import io.servicetalk.http.api.StreamingHttpService;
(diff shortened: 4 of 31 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        @Override
+        public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) throws Exception {
+            if (evt instanceof PipelineInitializedEvent) {
+                ctx.pipeline().remove(this);
+            } else {
+                ctx.fireUserEventTriggered(evt);
+            }
+        }
```
bryce-anderson, line 93: I just want to double check my understanding of this: we are using the buffering properties of `ByteToMessageDecoder` to both aggregate data until we get a TLS header (or not) and then we keep this in the pipeline until we get the user event so that we will continue to buffer until we've fully setup the pipeline. Is that right?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslNegotiator.java
```diff
@@ -0,0 +1,125 @@
...
+                                                executionContext, roNewConfig,
+                                                new TcpServerChannelInitializer(roNewConfig.tcpConfig(),
+                                                        connectionObserver, executionContext), service,
+                                                drainRequestPayloadBody, connectionObserver);
+                                    } else {
+                                        return NettyHttpServer.initChannel(channel, executionContext, roNewConfig,
+                                                new TcpServerChannelInitializer(roNewConfig.tcpConfig(),
+                                                        connectionObserver, executionContext), service,
+                                                drainRequestPayloadBody, connectionObserver);
+                                    }
```
bryce-anderson, line 108: feel free to ignore: I know it's not often the style in java, but if you put all this in an `else` block it can prevent help bugs since it forces the `if` block to have a return statement on all paths.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
```diff
@@ -0,0 +1,117 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import javax.annotation.Nullable;
+
+final class OptionalSslChannelSingle extends ChannelInitSingle<Boolean> {
+
+    OptionalSslChannelSingle(final Channel channel, final ChannelInitializer channelInitializer) {
+        super(channel, channelInitializer);
+    }
```
bryce-anderson, line 37: Can we drop the `ChannelInitializer` parameter and pass `NoopChannelInitializer.INSTANCE` directly to `super(..)`? It looks like we never pass another initializer.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
```diff
@@ -0,0 +1,117 @@
...
+            } else {
+                ctx.fireUserEventTriggered(evt);
+            }
+        }
+    }
+
+    /**
+     * {@link ChannelInitializer} that does not do anything.
+     */
+    static final class NoopChannelInitializer implements ChannelInitializer {
```
bryce-anderson, line 99: Does it make sense to use the instance in `AlpnChannelSingle`? If we find that distasteful, we could lift it up into the package level. It looks like it's used in a number of places already that are outside of the `AlpnChannelSingle` class (although many are associated).

bryce-anderson (commented):
Nice work! Would it make sense to have some logging around the creation of insecure channels so people can track the percentages of traffic that is secure vs insecure?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
```diff
@@ -0,0 +1,117 @@
...
+    OptionalSslChannelSingle(final Channel channel, final ChannelInitializer channelInitializer) {
+        super(channel, channelInitializer);
+    }
+
+    @Override
+    protected ChannelHandler newChannelHandler(final Subscriber<? super Boolean> subscriber) {
+        return new OptionalSslHandler(subscriber);
+    }
+
+    static final class OptionalSslHandler extends ByteToMessageDecoder {
```
idelpivnitskiy, line 44: Consider making this class private

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
```diff
@@ -0,0 +1,117 @@
...
+            boolean isEncrypted = SslHandler.isEncrypted(in);
+            LOGGER.debug("{} Detected TLS for this connection: {}", ctx.channel(), isEncrypted);
+            final SingleSource.Subscriber<? super Boolean> subscriberCopy = subscriber;
+            subscriber = null;
+            subscriberCopy.onSuccess(isEncrypted);
+        }
+
+        @Override
+        public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) throws Exception {
+            if (evt instanceof PipelineInitializedEvent) {
```
idelpivnitskiy, line 88: Because this is a pkg-private singleton event, consider using `==` instead of `instanceof`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
```diff
@@ -0,0 +1,117 @@
...
+        @Override
+        protected void decode(final ChannelHandlerContext ctx, final ByteBuf in, final List<Object> out) {
+            if (in.readableBytes() < SSL_RECORD_HEADER_LENGTH || subscriber == null) {
+                return;
+            }
+            boolean isEncrypted = SslHandler.isEncrypted(in);
+            LOGGER.debug("{} Detected TLS for this connection: {}", ctx.channel(), isEncrypted);
+            final SingleSource.Subscriber<? super Boolean> subscriberCopy = subscriber;
+            subscriber = null;
+            subscriberCopy.onSuccess(isEncrypted);
```
idelpivnitskiy, line 83: Can you try doing `ctx.pipeline().remove(this);` right after this line instead of handling `userEventTriggered`?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslNegotiator.java
```diff
@@ -0,0 +1,125 @@
...
+        final ReadOnlyHttpServerConfig roConfig = config.asReadOnly();
+        final ReadOnlyTcpServerConfig roTcpConfig = roConfig.tcpConfig();
+
+        final BiFunction<Channel, ConnectionObserver, Single<NettyConnectionContext>> channelInit =
+                (channel, connectionObserver) ->
+                        new OptionalSslChannelSingle(channel, OptionalSslChannelSingle.NoopChannelInitializer.INSTANCE)
+                                .flatMap(isTls -> {
+                                    if (isTls) {
+                                        if (roTcpConfig.isAlpnConfigured()) {
+                                            return DeferredServerChannelBinder.alpnInitChannel(listenAddress, channel,
```
idelpivnitskiy, line 65: Sad that for the same series of conditions the return result is different compare to `DefaultHttpServerBuilder.doBind` . Please add a cross comment that these 2 places must always be consistent.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslNegotiator.java
```diff
@@ -0,0 +1,125 @@
...
+                                                    new TcpServerChannelInitializer(roTcpConfig, connectionObserver,
+                                                            executionContext), service,
+                                                    drainRequestPayloadBody, connectionObserver);
+                                        }
+                                    }
+
+                                    // TODO: need to duplicate all but SSL config
+                                    final HttpServerConfig newConfig = new HttpServerConfig();
+                                    if (roConfig.h1Config() != null && roConfig.h2Config() != null) {
+                                        newConfig.httpConfig().protocols(roConfig.h1Config(), roConfig.h2Config());
```
idelpivnitskiy, line 89: This won't work for plaintext, should we fallback to HTTP/1.1?
We can consider adding a log statement in `DefaultHttpServerBuilder` (either inside `listenForService` or `doBind`) to let users know that we can fallback to HTTP/1.1 for plaintext path if they have both `acceptInsecureConnections()` and 2 protocols configured.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/SniCompleteChannelSingle.java
```diff
@@ -55,6 +55,7 @@ public void handlerAdded(final ChannelHandlerContext ctx) throws Exception {
             // Force a read to get the SSL handshake started. We initialize pipeline before
             // SslHandshakeCompletionEvent will complete, therefore, no data will be propagated before we finish
             // initialization.
+            ctx.pipeline().fireUserEventTriggered(PipelineInitializedEvent.INSTANCE);
```
idelpivnitskiy, line 58: The above comment is related to `ctx.read()`, consider moving this event before the comment (same for all other places).

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/OptionalSslTest.java
```diff
@@ -0,0 +1,115 @@
...
+            ExecutionContextExtension.cached("server-io", "server-executor")
+                    .setClassLevel(true);
+    @RegisterExtension
+    static final ExecutionContextExtension CLIENT_CTX =
+            ExecutionContextExtension.cached("client-io", "client-executor")
+                    .setClassLevel(true);
+
+    @ParameterizedTest(name = "{displayName} [{index}] acceptEnabled={0}, protocol={1}")
+    @MethodSource("args")
+    void acceptsEncryptedAndNonEncryptedRequests(final boolean acceptEnabled, final HttpProtocol protocol)
```
idelpivnitskiy, line 66: Consider using `acceptInsecureConnections` name instead of `acceptEnabled`. It's easier to understand the purpose of an arg when it matches with a builder option

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/OptionalSslTest.java
```diff
@@ -0,0 +1,115 @@
...
+            throws Exception {
+        final HttpService service = (ctx, request, responseFactory) ->
+                succeeded(responseFactory.ok().payloadBody("Hello World!", textSerializerUtf8()));
```
idelpivnitskiy, line 69: It will be great if you enhance the service implementation to check `ConnectionInfo`. We should assert that when it's a secure connection, we always have `SslConfig` and `SSLSession`. When plaintext, both should be `null`. You may need to send a hint in request (different path?) to distinguish expectations.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/OptionalSslTest.java
```diff
@@ -0,0 +1,115 @@
...
+                    assertThrows(ClosedChannelException.class, () -> client.request(request));
+                }
+            }
+        }
+    }
+
+    private static Stream<Arguments> args() {
+        return Stream.of(
+                Arguments.of(true, HttpProtocol.HTTP_1),
+                Arguments.of(false, HttpProtocol.HTTP_1),
```
idelpivnitskiy, line 92: We should also test both paths (plaintext and secure) for the following scenarios:
1. Different `SslProvier`(s)
2. TLSv1.3 / TLSv1.2
3. mTLS / no mTLS
4. ALPN / no ALPN
5. SecurityHandshakeObserver - we need to verify that observer is invoked when secure client makes a request and that it's not invoked when plaintext client makes a request.

You can either add more tests here for these scenarios or go to `SslProvidersTest`, `Tls13Test`, `MutualSslTest`, `AlpnClientAndServerTest`, `SecurityHandshakeObserverTest` to enhance those in a similar way you improved `SniTest`.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SniTest.java
```diff
@@ -93,9 +95,9 @@ void sniSuccess(SslProvider serverSslProvider, SslProvider clientSslProvider,
             "{displayName} [{index}] serverSslProvider={0} clientSslProvider={1} protocol={2} useALPN={3}")
```
idelpivnitskiy, line 95: Name should be updated (same for other tests below this one)

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SniTest.java
```diff
@@ -122,11 +124,12 @@ void sniFailExpected(SslProvider serverSslProvider, SslProvider clientSslProvide
...
     void sniDefaultFallbackFailExpected(SslProvider serverSslProvider, SslProvider clientSslProvider,
-                                        HttpProtocol protocol, boolean useALPN) throws Exception {
+                                        HttpProtocol protocol, boolean useALPN, boolean acceptInsecure)
+            throws Exception {
         try (ServerContext serverContext = newServerBuilder(SERVER_CTX, protocol)
-                .sslConfig(untrustedServerConfig(serverSslProvider, alpnIds(protocol, useALPN)),
+                .sslConfig(untrustedServerConfig(serverSslProvider, acceptInsecure, alpnIds(protocol, useALPN)),
                         singletonMap("no_match" + SNI_HOSTNAME,
-                                trustedServerConfig(serverSslProvider, alpnIds(protocol, useALPN))))
+                                trustedServerConfig(serverSslProvider, acceptInsecure, alpnIds(protocol, useALPN))))
```
idelpivnitskiy, line 132: hm... because of this it makes me feel like `ServerSslConfig` might be not the right place for a new flag. API should not allow users configure "foo" SNI with optional TLS and "bar" SNI with non-optional. It just won't work,  the "optional" choice is made before SNI choice. 

Should we consider a different approach? For example, we can add `HttpServerBuilder.sslConfig` overloads that will accept a boolean to signal if it should be optional or not. 

This will also help with `HttpServiceContext` state, because `ctx.sslConfig()` will be null for plaintext connections and non-null for secure. Having `ctx.sslConfig().acceptInsecureConnections()` does not provide much value. 

WDYT?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SniTest.java
```diff
@@ -175,11 +180,12 @@ void noSniClientDefaultServerFallbackSuccess(SslProvider serverSslProvider, SslP
             "{displayName} [{index}] serverSslProvider={0} clientSslProvider={1} protocol={2} useALPN={3}")
     @MethodSource("params")
     void noSniClientDefaultServerFallbackFailExpected(SslProvider serverSslProvider, SslProvider clientSslProvider,
-                                                      HttpProtocol protocol, boolean useALPN) throws Exception {
+                                                      HttpProtocol protocol, boolean useALPN, boolean acceptInsecure)
```
idelpivnitskiy, line 183: Consider enhancing tests to also execute plaintext path, similar to what you do in `OptionalSslTest`: create a non-secure client and test that it can/cannot serve a request depending on `acceptInsecure` flag. We will need to enhance `newSslVerifyService()` to allow plaintext connections.

idelpivnitskiy (commented):
Overall looks really nice, great job!
It's important to have comprehensive testing for all possible paths, see my recommendations for enhancing test coverage.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
daschl: 🤯 it works now -- I can swear it didn't work before but I clearly missed something when developing.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
daschl: So the challenge in netty is that this handler has data in the `in` buffer, and when the handler is removed anything thats in that buffer is sent up the pipeline to the next handler. This means that the pipeline must be set up (at least the next one in the pipeline must be there), otherwise those bytes go into nowhere and we are not making any more progress.

I had introduced the event propagation since I couldn't make it work, Idel suggested another time when to remove the handler (after the subscription completed). In the current implementation this works, so we can get rid of the custom event.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
daschl: good idea!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
daschl: 👍

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslNegotiator.java
daschl: good idea, thanks!

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SniTest.java
daschl: That is definitely a good point, it feels like in the wrong spot inside the SslConfig even though it is clearly related. I'm not sure I like the overload proposal because we already have 4 of them and another one is probably not going to help with discoverability.

As an alternative proposal in my latest PR I added another option to the builder which should hopefully make it clearer. I'm not 100% convinced with the naming though, so any input would be appreciated.

I know I still need to fix the tests and polish checkstyle to make the CI happy, but I wanted to get to a conclusion of this first before I spread out the API wider.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslNegotiator.java
daschl: We can consider something like this, but the TODO above is a bit wider. Do you know how we could duplicate a server config easily modulo the ssl bits? Or somehow duplicate the builder and then convert to readonly.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
bryce-anderson: This pattern relies on the `Subscriber` to synchronously triggering all pipeline modifications. While this is true for the code as it stands now, it relies on a coupling between the execution models of the netty pipeline and the RS implementation and the consumption of the signal itself and there is no guarantee that the downstream consumers of the `Single` will do their pipeline transformations synchronously forever. That is why I kind of liked having the event that acted as a bridge between the two threading models. If we want to rely on the implicit coupling we should make it it very clear (maybe via asserts?) that the pipeline was modified after calling the subscriber and before we remove this handler so we know someone will catch the bytes.

daschl: I'll see if I can add some asserts (i.e. assert that the pipeline is not empty before removing the handler) -- the event approach had the problem that there wasn't a central spot where the event got fired, but I had to add it "ad-hoc" in all kinds of places depending on the code path.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
```diff
@@ -52,6 +53,7 @@ public final class TcpServerConfig extends AbstractTcpConfig<ServerSslConfig> {
     private TransportObserver transportObserver = NoopTransportObserver.INSTANCE;
     @Nullable
     private Map<String, ServerSslConfig> sniConfig;
+    private SslListenMode sslListenMode;
```
bryce-anderson, line 56: Does this need a default value, presumably `SSL_REQUIRED`?

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
```diff
@@ -128,6 +134,11 @@ public TcpServerConfig sslConfig(ServerSslConfig defaultSslConfig, Map<String, S
         return this;
     }
 
+    public TcpServerConfig sslListenMode(final SslListenMode mode) {
+        this.sslListenMode = mode;
```
bryce-anderson, line 138: `requireNonNull(..)` Could also use some javadocs making it clear that this only has meaning if the `ServerSslConfig` is also set, otherwise it is ignored.

bryce-anderson (commented):
I thought the configuration fit best on the `SslServerConfig` but the SNI compatibility clearly throws a wrench in it. It could be that it's only matters in the default `ServerConfig` since as noted if we're selecting based on SNI name we're already doing TLS and thus the point is moot. However, that could definitely be confusing and lifting it up a level seems fine.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
bryce-anderson: Bridging the two models is always going to be a bit messy. The cleanest way to do this is to put the initialization all in channel handlers since then you're only dealing with a single execution model but that ship has sailed based on the existing `*ChannelSingle` implementations.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
daschl: 👍

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/ReadOnlyHttpServerConfig.java
```diff
@@ -20,50 +20,20 @@
 
 import javax.annotation.Nullable;
 
-final class ReadOnlyHttpServerConfig {
+interface ReadOnlyHttpServerConfig {
```
daschl, line 23: Had to refactor the readonly configs into interfaces so that I could add a delegate in the OptionalSslNegotiator.

NOTE: I know that this is a little incomplete (the client is not converted, and javadocs are missing). I can do some more polish and cleanup in a follow up commit.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/DefaultReadOnlyTcpServerConfig.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.Map.Entry;
+import javax.annotation.Nullable;
+
+import static io.servicetalk.transport.api.TransportObservers.asSafeObserver;
+import static io.servicetalk.transport.netty.internal.SslContextFactory.forServer;
+
+/**
+ * Read only view of {@link TcpServerConfig}.
+ */
+public final class DefaultReadOnlyTcpServerConfig extends AbstractReadOnlyTcpConfig<ServerSslConfig>
```
daschl, line 39: Identical to the old class of ReadOnlyTcpServerConfig, just added a Default in the name

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SecurityHandshakeObserverTest.java
```diff
@@ -204,4 +228,61 @@ private static void verifyObservers(InOrder order, TransportObserver transportOb
...
+    @Test
+    void optionalSslWithPlaintextDoesNotTriggerSecurityHandshake() throws Exception {
+        final CountDownLatch done = new CountDownLatch(1);
+        final AtomicBoolean securityInvoked = new AtomicBoolean(false);
+
+        try (ServerContext serverContext = BuilderUtils.newServerBuilder(SERVER_CTX)
+                .sslConfig(new ServerSslConfigBuilder(
+                        DefaultTestCerts::loadServerPem, DefaultTestCerts::loadServerKey).build())
+                .sslListenMode(SslListenMode.SSL_OPTIONAL)
+                .transportObserver((localAddress, remoteAddress) -> new ConnectionObserver() {
```
idelpivnitskiy, line 241: Consider using already initialized mock `serverTransportObserver` and then verify that `serverConnectionObserver.onSecurityHandshake()` was not invoked and there were no interactions with `serverSecurityHandshakeObserver`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/OptionalSslTest.java
```diff
@@ -0,0 +1,140 @@
...
+import static io.servicetalk.test.resources.DefaultTestCerts.serverPemHostname;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * Verifies the behavior and functionality of {@link ReadOnlyTcpServerConfig#sslListenMode()}.
+ */
+final class OptionalSslTest {
```
idelpivnitskiy, line 60: Thanks a lot for adding more tests! 
The last one we need to enhance is `AlpnClientAndServerTest`, to make sure ALPN works as expected after optional step.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/ReadOnlyTcpServerConfig.java
```diff
@@ -16,75 +16,22 @@
...
-            alpnConfigured = !sslContext.applicationProtocolNegotiator().protocols().isEmpty();
-        } else {
-            sslContext = null;
-            sniMapping = null;
-            alpnConfigured = false;
-        }
-        sniMaxClientHelloLength = from.sniMaxClientHelloLength();
-        sniClientHelloTimeout = from.sniClientHelloTimeout();
-    }
+public interface ReadOnlyTcpServerConfig extends ReadOnlyTcpConfig<ServerSslConfig> {
```
idelpivnitskiy, line 34: Even though this one is in `-internal` module, it was still public class. We can not make such changes in 0.42.x branch.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslNegotiator.java
```diff
@@ -0,0 +1,248 @@
...
+
+    static Single<HttpServerContext> bind(final HttpExecutionContext executionContext,
+                                          final HttpServerConfig config,
+                                          final SocketAddress listenAddress,
+                                          @Nullable final InfluencerConnectionAcceptor connectionAcceptor,
+                                          final StreamingHttpService service,
+                                          final boolean drainRequestPayloadBody,
+                                          @Nullable final EarlyConnectionAcceptor earlyConnectionAcceptor,
+                                          @Nullable final LateConnectionAcceptor lateConnectionAcceptor) {
+        final ReadOnlyHttpServerConfig roConfig = config.asReadOnly();
```
idelpivnitskiy, line 71: The caller of this `bind` method already has `roConfig`, consider passing that instead of creating yet another copy

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -399,7 +406,13 @@ private Single<HttpServerContext> doBind(final HttpExecutionContext executionCon
         ReadOnlyHttpServerConfig roConfig = config.asReadOnly();
         StreamingHttpService filteredService = applyInternalFilters(service, roConfig.lifecycleObserver());
 
-        if (roConfig.tcpConfig().isAlpnConfigured()) {
+        // Important: This code needs to be kept in-sync with the similar, but not identical
+        // code inside OptionalSslNegotiator#bind
+        if (roConfig.tcpConfig().sslConfig() != null &&
+                roConfig.tcpConfig().sslListenMode() == SslListenMode.SSL_OPTIONAL) {
+            return OptionalSslNegotiator.bind(executionContext, config, address, connectionAcceptor, filteredService,
```
idelpivnitskiy, line 413: Instead of doing refactoring of `ReadOnlyTcpServerConfig`, consider following an approach we have for `HttpClientConfig`. It has a copy constructor, that invokes a copy ctor of `TcpClientConfig` and `HttpConfig`.

If we add similar copy ctors for `HttpServerConfig` and `TcpServerConfig`, we will be able to make a full copy before invoking `OptionalSslNegotiator.bind`, then reset `SslConfig` (we can make `AbstractTcpConfig.sslConfig(...) param `@Nullable`), and create a read-only version without `SslConfig` upfront. Both configs will be passed to `bind`. A config copy will be created only once per server, not per each new connection.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -100,6 +101,16 @@ default HttpServerBuilder sslConfig(ServerSslConfig defaultConfig, Map<String, S
                 "sslConfig(ServerSslConfig, Map, int, Durations) is not supported by " + getClass());
     }
 
+    /**
+     * Allows to configure the listen mode when SSL is enabled.
+     *
+     * @param mode the mode to use.
+     * @return {@code this}.
+     */
+    default HttpServerBuilder sslListenMode(final SslListenMode mode) {
```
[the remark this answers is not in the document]
daschl: I'm going to go with the first approach, because with the second one it's again ambiguous (the default config does not have the acceptInsecureConnections).

PR #2862, opened 2024-03-06
Test gzipCompression encoding for hello server

dramaticlly:
Start test server `HelloWorldJaxRsServer`

send following curl and get the results like 

- good
```
curl http://localhost:8080/greetings/hello
hello world%
curl http://localhost:8080/greetings/hello\?who\=turnip
hello turnip%
```

- bad
```
✗ curl http://localhost:8080/greetings/hello -H 'Accept-Encoding: gzip'
curl: (18) transfer closed with 11 bytes remaining to read
✗ curl --compressed http://localhost:8080/greetings/hello -H 'Accept-Encoding: gzip'
curl: (18) transfer closed with 11 bytes remaining to read
```

PR #2863, opened 2024-03-06
Fix Content-Length header removal after content compression

chrisvest:
Motivation:
The ContentEncodingHttpServiceFilter is meant to remove the `Content-Length` header if it does compression, because any such header from earlier stages will be incorrect after compression. An incorrect `Content-Length` header will then cause down-stream clients to be confused and potentially stall, timeout, or produce an unexpected close error, when they don't receive their promised number of bytes (most likely fewer than promised, due to compression).

Modifications:
The ContentEncodingHttpServiceFilter incorrectly removed the `Content-Length` header from the _request_ object, and this has been fixed so it now correctly removes the header from the _response_ object. There was a test for this behavior, but the test wasn't actually hitting this code because it used a client that did not advertise support for compressed responses. The test has also been fixed, adding the case where the client supports compressed responses. We now test both that the `Content-Length` header is removed when the response is compressed, and that the header stays when the response is not compressed.

Result:
This fixes the issue showcased by https://github.com/apple/servicetalk/pull/2862


idelpivnitskiy (approved):
Great catch!

dramaticlly (approved):

Scottmitch (approved):

PR #2871, opened 2024-03-13
Add support for Optional TLS negotiation for http/1 and http/2

daschl:
Motivation
----------
In certain deployment scenarios it can be useful to accept incoming traffic on the same port for both TLS and non-TLS connections. ServiceTalk does not support this at this point, so this changeset aims to add the feature.

Modifications
-------------
User-facing, a new overload option for the sslConfig is introduced which allows the user to specify that insecure connections should also be accepted. When set to true, both http/1 and http/2 protocols are supported, although ServiceTalk does not support h2c cleartext upgrades so this combination is still not supported.

This also works with ALPN and SNI, since the negotiation happens before the TLS connection is fully established.

Internally, the OptionalSslNegotiator will determine if the incoming first bytes of the connection signify a TLS connection or not and will hand it to the right channel initializer afterwards.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
@@ -68,6 +68,18 @@ public interface HttpServerBuilder {
     HttpServerBuilder sslConfig(ServerSslConfig config);
 
     /**
+     * Set the SSL/TLS configuration and allows to specify if insecure connections should also be allowed.
+     *
+     * @param config The configuration to use.
+     * @param acceptInsecureConnections if non-TLS connections are accepted on the same socket.
+     * @return {@code this}.
+     */
+    default HttpServerBuilder sslConfig(ServerSslConfig config, boolean acceptInsecureConnections) {
+        throw new UnsupportedOperationException(
+                "sslConfig(ServerSslConfig, boolean) is not supported by " + getClass());
+    }
+
+    /**
      * Set the SSL/TLS and <a href="https://tools.ietf.org/html/rfc6066#section-3">SNI</a> configuration.
      *
      * @param defaultConfig The configuration to use if the client certificate's SNI extension isn't present or the
@@ -93,11 +105,37 @@ public interface HttpServerBuilder {
      * Implementations can round the specified {@link Duration} to full time units, depending on their time granularity.
      * {@link Duration#ZERO Zero (0)} disables timeout.
      * @return {@code this}.
+     * @deprecated use {@link #sslConfig(ServerSslConfig, Map, int, Duration, boolean)} instead.
      */
+    @Deprecated
     default HttpServerBuilder sslConfig(ServerSslConfig defaultConfig, Map<String, ServerSslConfig> sniMap,
                                         int maxClientHelloLength, Duration clientHelloTimeout) {
         throw new UnsupportedOperationException(
-                "sslConfig(ServerSslConfig, Map, int, Durations) is not supported by " + getClass());
+                "sslConfig(ServerSslConfig, Map, int, Duration) is not supported by " + getClass());
+    }
+
+    /**
+     * Set the SSL/TLS and <a href="https://tools.ietf.org/html/rfc6066#section-3">SNI</a> configuration.
+     *
+     * @param defaultConfig The configuration to use if the client certificate's SNI extension isn't present or the
+     * SNI hostname doesn't match any values in {@code sniMap}.
+     * @param sniMap A map where the keys are matched against the client certificate's SNI extension value in order
+     * to provide the corresponding {@link ServerSslConfig}.
+     * @param maxClientHelloLength The maximum length of a
+     * <a href="https://www.rfc-editor.org/rfc/rfc5246#section-7.4.1.2">ClientHello</a> message in bytes, up to
+     * {@code 2^24 - 1} bytes. Zero ({@code 0}) disables validation.
+     * @param clientHelloTimeout The timeout for waiting until
+     * <a href="https://www.rfc-editor.org/rfc/rfc5246#section-7.4.1.2">ClientHello</a> message is received.
+     * Implementations can round the specified {@link Duration} to full time units, depending on their time granularity.
+     * @param acceptInsecureConnections if non-TLS connections are accepted on the same socket.
+     * {@link Duration#ZERO Zero (0)} disables timeout.
+     * @return {@code this}.
+     */
+    default HttpServerBuilder sslConfig(ServerSslConfig defaultConfig, Map<String, ServerSslConfig> sniMap,
+                                        int maxClientHelloLength, Duration clientHelloTimeout,
+                                        boolean acceptInsecureConnections) {
+        throw new UnsupportedOperationException(
+                "sslConfig(ServerSslConfig, Map, int, Duration, boolean) is not supported by " + getClass());
     }
 
     /**
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
@@ -204,20 +204,36 @@ final class DefaultHttpServerBuilder implements HttpServerBuilder {
 
     @Override
     public HttpServerBuilder sslConfig(final ServerSslConfig config) {
-        this.config.tcpConfig().sslConfig(config);
+        this.config.tcpConfig().sslConfig(requireNonNull(config));
         return this;
     }
 
     @Override
     public HttpServerBuilder sslConfig(final ServerSslConfig defaultConfig, final Map<String, ServerSslConfig> sniMap) {
-        this.config.tcpConfig().sslConfig(defaultConfig, sniMap);
+        this.config.tcpConfig().sslConfig(requireNonNull(defaultConfig), requireNonNull(sniMap));
         return this;
     }
 
     @Override
     public HttpServerBuilder sslConfig(final ServerSslConfig defaultConfig, final Map<String, ServerSslConfig> sniMap,
                                        final int maxClientHelloLength, final Duration clientHelloTimeout) {
-        this.config.tcpConfig().sslConfig(defaultConfig, sniMap, maxClientHelloLength, clientHelloTimeout);
+        this.config.tcpConfig().sslConfig(requireNonNull(defaultConfig), requireNonNull(sniMap), maxClientHelloLength,
+                requireNonNull(clientHelloTimeout));
+        return this;
+    }
+
+    @Override
+    public HttpServerBuilder sslConfig(final ServerSslConfig config, final boolean acceptInsecureConnections) {
+        this.config.tcpConfig().sslConfig(requireNonNull(config), acceptInsecureConnections);
+        return this;
+    }
+
+    @Override
+    public HttpServerBuilder sslConfig(final ServerSslConfig defaultConfig, final Map<String, ServerSslConfig> sniMap,
+                                       final int maxClientHelloLength, final Duration clientHelloTimeout,
+                                       final boolean acceptInsecureConnections) {
+        this.config.tcpConfig().sslConfig(requireNonNull(defaultConfig), requireNonNull(sniMap), maxClientHelloLength,
+                requireNonNull(clientHelloTimeout), acceptInsecureConnections);
         return this;
     }
 
@@ -399,7 +415,13 @@ final class DefaultHttpServerBuilder implements HttpServerBuilder {
         ReadOnlyHttpServerConfig roConfig = config.asReadOnly();
         StreamingHttpService filteredService = applyInternalFilters(service, roConfig.lifecycleObserver());
 
-        if (roConfig.tcpConfig().isAlpnConfigured()) {
+        if (roConfig.tcpConfig().sslConfig() != null && roConfig.tcpConfig().acceptInsecureConnections()) {
+            config.tcpConfig().sslConfig(null, null);
+            ReadOnlyHttpServerConfig roConfigWithoutSsl = config.asReadOnly();
+            return OptionalSslNegotiator.bind(executionContext, roConfig, roConfigWithoutSsl, address,
+                    connectionAcceptor, service, drainRequestPayloadBody, earlyConnectionAcceptor,
+                    lateConnectionAcceptor);
+        } else if (roConfig.tcpConfig().isAlpnConfigured()) {
             return DeferredServerChannelBinder.bind(executionContext, roConfig, address, connectionAcceptor,
                     filteredService, drainRequestPayloadBody, false, earlyConnectionAcceptor, lateConnectionAcceptor);
         } else if (roConfig.tcpConfig().sniMapping() != null) {
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright © 2024 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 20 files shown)


daschl:
supersedes #2859

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -399,7 +415,13 @@ private Single<HttpServerContext> doBind(final HttpExecutionContext executionCon
         ReadOnlyHttpServerConfig roConfig = config.asReadOnly();
         StreamingHttpService filteredService = applyInternalFilters(service, roConfig.lifecycleObserver());
 
-        if (roConfig.tcpConfig().isAlpnConfigured()) {
+        if (roConfig.tcpConfig().sslConfig() != null && roConfig.tcpConfig().acceptInsecureConnections()) {
+            config.tcpConfig().sslConfig(null, null);
```
idelpivnitskiy, line 419: We should not modify user-provided config. Consider using `new HttpServerConfig(config)` copy ctor that you added and then modify and create `roConfigWithoutSsl`.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
```diff
@@ -95,9 +126,10 @@ public void transportObserver(final TransportObserver transportObserver) {
      * found the corresponding {@link ServerSslConfig} is used.
      * @return {@code this}
      */
-    public TcpServerConfig sslConfig(ServerSslConfig defaultSslConfig, Map<String, ServerSslConfig> sniConfig) {
```
idelpivnitskiy, line 98: This is a public API change. Consider deprecating pre-existing `sslConfig(SslConfig)` method before adding `sslConfig(SslConfig, boolean)`

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
```diff
@@ -128,6 +160,30 @@ public TcpServerConfig sslConfig(ServerSslConfig defaultSslConfig, Map<String, S
...
+     * <a href="https://www.rfc-editor.org/rfc/rfc5246#section-7.4.1.2">ClientHello</a> message is received.
+     * Implementations can round the specified {@link Duration} to full time units, depending on their time granularity.
+     * @param acceptInsecureConnections if non-TLS connections are accepted on the same socket.
+     * {@link Duration#ZERO Zero (0)} disables timeout.
+     * @return {@code this}
+     */
+    public TcpServerConfig sslConfig(ServerSslConfig defaultSslConfig, Map<String, ServerSslConfig> sniConfig,
+                                     int maxClientHelloLength, Duration clientHelloTimeout,
+                                     boolean acceptInsecureConnections) {
+        sslConfig(defaultSslConfig, sniConfig, maxClientHelloLength, clientHelloTimeout);
```
idelpivnitskiy, line 182: For consistency, consider deprecating `sslConfig(defaultSslConfig, sniConfig, maxClientHelloLength, clientHelloTimeout)` and more all move to use this new overload.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -204,20 +204,36 @@ public HttpServerBuilder protocols(final HttpProtocolConfig... protocols) {
...
+        this.config.tcpConfig().sslConfig(requireNonNull(defaultConfig), requireNonNull(sniMap));
         return this;
     }
 
     @Override
     public HttpServerBuilder sslConfig(final ServerSslConfig defaultConfig, final Map<String, ServerSslConfig> sniMap,
                                        final int maxClientHelloLength, final Duration clientHelloTimeout) {
-        this.config.tcpConfig().sslConfig(defaultConfig, sniMap, maxClientHelloLength, clientHelloTimeout);
+        this.config.tcpConfig().sslConfig(requireNonNull(defaultConfig), requireNonNull(sniMap), maxClientHelloLength,
+                requireNonNull(clientHelloTimeout));
```
idelpivnitskiy, line 221: Consider moving `requireNonNull` checks inside `TcpServerConfig` method

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -204,20 +204,36 @@ public HttpServerBuilder protocols(final HttpProtocolConfig... protocols) {
...
+    public HttpServerBuilder sslConfig(final ServerSslConfig config, final boolean acceptInsecureConnections) {
+        this.config.tcpConfig().sslConfig(requireNonNull(config), acceptInsecureConnections);
+        return this;
+    }
+
+    @Override
+    public HttpServerBuilder sslConfig(final ServerSslConfig defaultConfig, final Map<String, ServerSslConfig> sniMap,
+                                       final int maxClientHelloLength, final Duration clientHelloTimeout,
+                                       final boolean acceptInsecureConnections) {
+        this.config.tcpConfig().sslConfig(requireNonNull(defaultConfig), requireNonNull(sniMap), maxClientHelloLength,
```
idelpivnitskiy, line 235: ^ that will help to avoid need for another check here

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
```diff
@@ -93,11 +105,37 @@ public interface HttpServerBuilder {
...
+     * @param clientHelloTimeout The timeout for waiting until
+     * <a href="https://www.rfc-editor.org/rfc/rfc5246#section-7.4.1.2">ClientHello</a> message is received.
+     * Implementations can round the specified {@link Duration} to full time units, depending on their time granularity.
+     * @param acceptInsecureConnections if non-TLS connections are accepted on the same socket.
+     * {@link Duration#ZERO Zero (0)} disables timeout.
+     * @return {@code this}.
+     */
+    default HttpServerBuilder sslConfig(ServerSslConfig defaultConfig, Map<String, ServerSslConfig> sniMap,
+                                        int maxClientHelloLength, Duration clientHelloTimeout,
+                                        boolean acceptInsecureConnections) {
```
idelpivnitskiy, line 136: The new API lgtm! Let me just finish [discussion](https://github.com/apple/servicetalk/pull/2859#discussion_r1522546121) to make sure we don't miss anything.

> I'm going to go with the first approach, because with the second one it's again ambiguous (the default config does not have the acceptInsecureConnections).

Agreed, moving `acceptInsecureConnections` under `SniConfig` was likely confusing. WDYT about:

```java
HttpServerBuilder sslConfig(ServerSslConfig config);
HttpServerBuilder sslConfig(ServerSslConfig config, boolean acceptInsecureConnections);
HttpServerBuilder sslConfig(ServerSslConfig defaultConfig, SniConfig sniConfig);
HttpServerBuilder sslConfig(ServerSslConfig defaultConfig, boolean acceptInsecureConnections, SniConfig sniConfig);
@Deprecated
HttpServerBuilder sslConfig(ServerSslConfig defaultConfig, Map<String, ServerSslConfig> sniMap);
@Deprecated
HttpServerBuilder sslConfig(ServerSslConfig defaultConfig, Map<String, ServerSslConfig> sniMap, int maxClientHelloLength, Duration clientHelloTimeout);
```
My worry is that there could be more options for SNI in the future and it will force us to change API of `HttpServerBuilder` again and again. 

Another option to speed up this PR is to add only 
```java
HttpServerBuilder sslConfig(ServerSslConfig config, boolean acceptInsecureConnections);
```
And skip SNI for the follow-up, if we decide to go forward with `SniConfig` because it will require quite a bit of refactoring around existing SNI logic.

idelpivnitskiy: @bryce-anderson we discussed offline with @daschl and agreed to add only 
```
HttpServerBuilder sslConfig(ServerSslConfig config, boolean acceptInsecureConnections);
``` 
in this PR. Could you please look at this thread from the [beginning](https://github.com/apple/servicetalk/pull/2859#discussion_r1522546121) and share your opinion? What will be a good API for SNI configuration?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/AlpnClientAndServerTest.java
```diff
@@ -91,50 +94,56 @@ void beforeEach() {
...
-                        HttpProtocolVersion.HTTP_1_1, null, null),
-                Arguments.of(singletonList(HTTP_1_1), singletonList(HTTP_2),
-                        null, ClosedChannelException.class, null),
-                Arguments.of(singletonList(HTTP_1_1), singletonList(HTTP_1_1),
-                        HttpProtocolVersion.HTTP_1_1, null, null));
+                        HttpProtocolVersion.HTTP_1_1, null, null, false)
+        ));
+        for (boolean acceptInsecureConnections : asList(true, false)) {
+            arguments.addAll(Arrays.asList(
+                    Arguments.of(singletonList(HTTP_2), asList(HTTP_2, HTTP_1_1),
```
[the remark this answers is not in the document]
daschl: Will be done as a follow-up.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
```diff
@@ -95,9 +126,10 @@ public void transportObserver(final TransportObserver transportObserver) {
      * found the corresponding {@link ServerSslConfig} is used.
      * @return {@code this}
      */
-    public TcpServerConfig sslConfig(ServerSslConfig defaultSslConfig, Map<String, ServerSslConfig> sniConfig) {
+    public TcpServerConfig sslConfig(@Nullable ServerSslConfig defaultSslConfig,
+                                     @Nullable Map<String, ServerSslConfig> sniConfig) {
         sslConfig(defaultSslConfig);
-        this.sniConfig = requireNonNull(sniConfig);
+        this.sniConfig = sniConfig;
         return this;
```
bryce-anderson: What happens if we call `.sslConfig(config1, true).sslConfig(config2)`? I would expect to find a final state of `.sslConfig() == config2 && .acceptInsecureConnections() == false`, but I think we'll find `.sslConfig() == config2 && .acceptInsecureConnections() == true` right now. Since `DefaultHttpServerBuilder` defers to `TcpServerConfig`, I think it inherits this behavior and is probably a security foot-gun.

bryce-anderson (commented):
The implementation looks good to me but there are still a few API things to consider. By and large I think the API itself is fine but the documentation and implementation of it probably need a few more tweaks.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
```diff
@@ -87,6 +105,19 @@ public void transportObserver(final TransportObserver transportObserver) {
...
+    /**
+     * Set the SSL/TLS configuration and allows to specify if insecure connections should also be allowed.
+     *
+     * @param config The configuration to use.
+     * @param acceptInsecureConnections if non-TLS connections are accepted on the same socket.
+     * @return {@code this}.
+     */
+    public TcpServerConfig sslConfig(ServerSslConfig config, boolean acceptInsecureConnections) {
+        sslConfig(config);
+        this.acceptInsecureConnections = acceptInsecureConnections;
```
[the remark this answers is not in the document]
bryce-anderson: looks like this is outdated.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpServerBuilder.java
idelpivnitskiy: Correct, `TcpServerConfig` is public only due to packaging reasons. Unfortunately, we can not control how users use internal public API and therefore, must play a safe game for 0.42.x backward compatibility. No use-cases comes to my mind immediately except someone using a TCP server or building something on top of our TCP stack. 

> I would have thought `HttpServerBuilder sslConfig(ServerSslConfig defaultConfig, Map<String, ServerSslConfig> sniMap, boolean acceptInsecureConnections)` would have been the go-to setter for most people.


What about users who need 2 other options for SNI: `maxClientHelloLength` and `clientHelloTimeout`? Are you ok with yet another overload `HttpServerBuilder sslConfig(ServerSslConfig defaultConfig, Map<String, ServerSslConfig> sniMap, int maxClientHelloLength, Duration clientHelloTimeout, boolean acceptInsecureConnections)`? Or do you prefer to move `sniMap`, `maxClientHelloLength` and `clientHelloTimeout` under a new `SniConfig` with assumption that we will be able to add more props there if needed without changing `HttpServerBuilder` API again.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
```diff
@@ -0,0 +1,94 @@
...
+        @Override
+        protected void decode(final ChannelHandlerContext ctx, final ByteBuf in, final List<Object> out) {
+            if (in.readableBytes() < SSL_RECORD_HEADER_LENGTH || subscriber == null) {
+                return;
+            }
+            boolean isEncrypted = SslHandler.isEncrypted(in);
+            LOGGER.debug("{} Detected TLS for this connection: {}", ctx.channel(), isEncrypted);
+            final SingleSource.Subscriber<? super Boolean> subscriberCopy = subscriber;
+            subscriber = null;
+            subscriberCopy.onSuccess(isEncrypted);
```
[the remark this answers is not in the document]
idelpivnitskiy: this is a pattern we use in many other places (search for `subscriberCopy`), I would probably be in favor of consistency... but either way is fine

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
idelpivnitskiy: Good point! I would expect the same.
I think it should be backward compatible if we remove `sslConfig` method from `AbstractTcpConfig` (because it's a pkg-private abstract class) and implement it on `TcpServerConfig/TcpClientConfig`, where `TcpServerConfig` will reset the flag to `false`.

@daschl this is how you can validate that with japicmp in your branch:
```
./gradlew assemble
./scripts/japicmp.sh 0.42.42 local
```

idelpivnitskiy (approved):
LGTM after @bryce-anderson comments addressed

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptionalSslChannelSingle.java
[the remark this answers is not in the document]
bryce-anderson: Consistency is a good so unless someone feels real strongly about switching to a boolean we should leave it as is.

bryce-anderson (dismissed):

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
```diff
@@ -87,6 +118,29 @@ public void transportObserver(final TransportObserver transportObserver) {
...
+    /**
+     * Set the SSL/TLS configuration and allows to specify if insecure connections should also be allowed.
+     *
+     * @param config The configuration to use.
+     * @param acceptInsecureConnections if non-TLS connections are accepted on the same socket.
+     * @return {@code this}.
+     */
+    public TcpServerConfig sslConfig(ServerSslConfig config, boolean acceptInsecureConnections) {
+        sslConfig(requireNonNull(config));
+        this.acceptInsecureConnections = acceptInsecureConnections;
```
bryce-anderson, line 140: I think we still have a few problems with these overloads.
1. This is the only signature enforcing that `ServerSslConfig != null` but I think all but the single param version expect it to be non null.
2. We're still keeping a false memory of the `maxClientHelloLength` and `clientHelloTimeout` settings, although this is not your doing.

Maybe we should funnel everything through a `private setSslConfig(*all params*)` method and make each overload to defer to it and fill in missing params with defaults. Since this predates this PR (although the proliferation makes it worse) there is a good argument to do it as a follow up.

bryce-anderson (approved):
I'll defer to your judgement whether we should cleanup `TcpServerConfig` now or later.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerConfig.java
daschl: 👍 I will do this as a follow up, thanks!

PR #2877, opened 2024-03-15
Optimize protocol negotiation with Optional TLS

daschl:
Motivation
----------
The initial support for Optional TLS did only allow insecure connections if either H1 or H2 is enabled, but not protocols at the same time. This is a limitation when used together with ALPN on secure connections and can be improved.

Modifications
-------------
This changeset makes sure that ALPN support still works when optional TLS is configured. In the case where both H1 and H2 are enabled on the server side and a insecure connection has been negotiated, the code now falls back to H1 support only. This means that h2c cleartext upgrades are not supported, but they haven't been supported inside ServiceTalk anyways.

Result
------
Best possible ALPN support at the moment in combination with Optional TLS.


bryce-anderson (approved):
Within the limits of the current implementation this looks good. I do think we can support H2 later: there is no reason we can't do what you're doing at the TLS layer at the protocol layer and detect a cleartext H2 prior knowledge session but I think that would be best left for a followup.

idelpivnitskiy (approved):

idelpivnitskiy:
Instead of a proper cleartext upgrade checking the first bytes to detect if it's a preface or HTTP/1.x initial line? Nice idea!
Don't have a use-case in mind for now, but we can make it if someone is interested

PR #2891, opened 2024-04-12
Lower initial delay for ServiceDiscovery retries from 8 to 2 seconds

idelpivnitskiy:
Motivation:

Experiments show that majority of ServiceDiscovery queries take only a few milliseconds. Jitter up to 8 seconds causes too long delay before the next retry. Based on observed numbers we can lower the value down to 2 seconds.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
@@ -104,7 +104,7 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
     private static final StreamingHttpConnectionFilterFactory DEFAULT_IDLE_TIMEOUT_FILTER =
             new IdleTimeoutConnectionFilter(ofMinutes(5));
 
-    static final Duration SD_RETRY_STRATEGY_INIT_DURATION = ofSeconds(8);
+    static final Duration SD_RETRY_STRATEGY_INIT_DURATION = ofSeconds(2);
     static final Duration SD_RETRY_STRATEGY_MAX_DELAY = ofSeconds(256);
 
     @Nullable


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -104,7 +104,7 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
     private static final StreamingHttpConnectionFilterFactory DEFAULT_IDLE_TIMEOUT_FILTER =
             new IdleTimeoutConnectionFilter(ofMinutes(5));
 
-    static final Duration SD_RETRY_STRATEGY_INIT_DURATION = ofSeconds(8);
+    static final Duration SD_RETRY_STRATEGY_INIT_DURATION = ofSeconds(2);
     static final Duration SD_RETRY_STRATEGY_MAX_DELAY = ofSeconds(256);
```
idelpivnitskiy, line 108: Should we also lower the max cap down to 128 seconds (~2min)?

bryce-anderson: That seems intuitively reasonable to me but some specific data would be helpful.

bryce-anderson (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
idelpivnitskiy: what kind of data? I can only use rational thinking and ~5 min seems quite a long delay 😄

tkountis (approved):

daschl (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
idelpivnitskiy: I will merge as-is and we can discuss the max delay separately

bryce-anderson: Sorry, I didn't see the reply. Data that would potentially be interesting would be what the conditions were around DNS overload scenarios, but I have a hard time imagining this would be a contributor to an otherwise healthy system falling over.

idelpivnitskiy: https://github.com/apple/servicetalk/pull/2895

PR #2908, opened 2024-05-06
`DnsServiceDiscoverer`: add total resolution timeout

idelpivnitskiy:
Motivation:

The existing `DnsServiceDiscovererBuilder.queryTimeout` is applied per each UDP/TCP query. If resolution requires multiple CNAME resolutions, retries on a different NS server, or applies a list of search domains, the total time is not capped. As a result, resolution can take significantly longer than originally anticipated.

Modifications:

- Add `DnsServiceDiscovererBuilder.resolutionTimeout` option that can be used to cap total resolution duration with default value set to `2 x queryTimeout`;
- Add tests for `queryTimeout` and `resolutionTimeout`;

Result:

Users can configure the total resolution timeout.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -437,11 +437,9 @@ final class DefaultDnsClient implements DnsClient {
                     final Future<List<DnsRecord>> resolveFuture =
                             resolver.resolveAll(new DefaultDnsQuestion(name, SRV));
                     final Future<?> timeoutFuture = resolutionTimeoutMillis == 0L ? null : eventLoop.schedule(() -> {
-                        if (promise.isDone()) {
-                            return;
-                        }
-                        if (promise.tryFailure(DnsNameResolverTimeoutException.newInstance(name,
-                                resolutionTimeoutMillis, SRV.toString(), SrvRecordPublisher.class, "doDnsQuery"))) {
+                        if (!promise.isDone() && promise.tryFailure(DnsNameResolverTimeoutException.newInstance(
+                                name, resolutionTimeoutMillis, SRV.toString(),
+                                SrvRecordPublisher.class, "doDnsQuery"))) {
                             resolveFuture.cancel(true);
                         }
                     }, resolutionTimeoutMillis, MILLISECONDS);
@@ -527,12 +525,9 @@ final class DefaultDnsClient implements DnsClient {
                     final Promise<DnsAnswer<InetAddress>> dnsAnswerPromise = eventLoop.newPromise();
                     final Future<List<InetAddress>> resolveFuture = resolver.resolveAll(name);
                     final Future<?> timeoutFuture = resolutionTimeoutMillis == 0L ? null : eventLoop.schedule(() -> {
-                        if (dnsAnswerPromise.isDone()) {
-                            return;
-                        }
-                        if (dnsAnswerPromise.tryFailure(DnsNameResolverTimeoutException.newInstance(
-                                name, resolutionTimeoutMillis, toRecordTypeNames(addressTypes),
-                                ARecordPublisher.class, "doDnsQuery"))) {
+                        if (!dnsAnswerPromise.isDone() && dnsAnswerPromise.tryFailure(
+                                DnsNameResolverTimeoutException.newInstance(name, resolutionTimeoutMillis,
+                                        toRecordTypeNames(addressTypes), ARecordPublisher.class, "doDnsQuery"))) {
                             resolveFuture.cancel(true);
                         }
                     }, resolutionTimeoutMillis, MILLISECONDS);
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -100,6 +100,7 @@ import static io.servicetalk.dns.discovery.netty.DnsClients.mapEventList;
 import static io.servicetalk.dns.discovery.netty.DnsResolverAddressTypes.IPV4_PREFERRED;
 import static io.servicetalk.dns.discovery.netty.DnsResolverAddressTypes.IPV6_PREFERRED;
 import static io.servicetalk.dns.discovery.netty.DnsResolverAddressTypes.preferredAddressType;
+import static io.servicetalk.dns.discovery.netty.DnsResolverAddressTypes.toRecordTypeNames;
 import static io.servicetalk.dns.discovery.netty.ServiceDiscovererUtils.calculateDifference;
 import static io.servicetalk.transport.netty.internal.BuilderUtils.datagramChannel;
 import static io.servicetalk.transport.netty.internal.BuilderUtils.socketChannel;
@@ -113,6 +114,7 @@ import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonList;
 import static java.util.Comparator.comparing;
 import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static java.util.function.Function.identity;
@@ -129,6 +131,7 @@ final class DefaultDnsClient implements DnsClient {
     private final MinTtlCache ttlCache;
     private final long maxTTLNanos;
     private final long ttlJitterNanos;
+    private final long resolutionTimeoutMillis;
     private final ListenableAsyncCloseable asyncCloseable;
     @Nullable
     private final DnsServiceDiscovererObserver observer;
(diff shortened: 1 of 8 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -222,6 +226,10 @@ final class DefaultDnsClient implements DnsClient {
             builder.nameServerProvider(toNettyType(dnsServerAddressStreamProvider));
         }
         resolver = builder.build();
+        this.resolutionTimeoutMillis = resolutionTimeout != null ? resolutionTimeout.toMillis() :
+                // Default value is chosen based on a combination of default "timeout" and "attempts" options of
+                // /etc/resolv.conf: https://man7.org/linux/man-pages/man5/resolv.conf.5.html
+                resolver.queryTimeoutMillis() * 2;
```
idelpivnitskiy, line 232: Open for alternative suggestions. Originally, I planned to use `resolver.queryTimeoutMillis() * resolver.maxQueriesPerResolve()`, but Nett's default values are:
```
queryTimeout = 5 sec
maxQueriesPerResolve = 16
```
which results in 80 sec and seems unreasonably high. 

Decided to go with x2 because https://man7.org/linux/man-pages/man5/resolv.conf.5.html because the linux default for attempts (maxQueriesPerResolve) is 2.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -424,9 +432,20 @@ protected AbstractDnsSubscription newSubscription(
...
                 protected Future<DnsAnswer<HostAndPort>> doDnsQuery(final boolean scheduledQuery) {
-                    Promise<DnsAnswer<HostAndPort>> promise = nettyIoExecutor.eventLoopGroup().next().newPromise();
+                    final EventLoop eventLoop = nettyIoExecutor.eventLoopGroup().next();
+                    final Promise<DnsAnswer<HostAndPort>> promise = eventLoop.newPromise();
+                    final Future<?> timeoutFuture = resolutionTimeoutMillis == 0L ? null : eventLoop.schedule(() -> {
+                        if (promise.isDone()) {
+                            return;
+                        }
+                        promise.tryFailure(DnsNameResolverTimeoutException.newInstance(
+                                name, resolutionTimeoutMillis, SRV.toString(), SrvRecordPublisher.class, "doDnsQuery"));
```
bryce-anderson, line 442: wdyt about cancelling the underlying resolution if we succeed in failing `promise`? It looks like it won't be properly acted upon by netty, but otoh it's good practice in case it ever is honored.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
bryce-anderson: Fwiw, this seems reasonable to me.

bryce-anderson (approved):

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
idelpivnitskiy: good idea, please check that I didn't mess up with anything: https://github.com/apple/servicetalk/pull/2908/commits/c86221fab6c80d5dc2890c357700e72d0fc98b59

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -424,9 +432,23 @@ protected AbstractDnsSubscription newSubscription(
...
+                    final Future<List<DnsRecord>> resolveFuture =
+                            resolver.resolveAll(new DefaultDnsQuestion(name, SRV));
+                    final Future<?> timeoutFuture = resolutionTimeoutMillis == 0L ? null : eventLoop.schedule(() -> {
+                        if (promise.isDone()) {
+                            return;
+                        }
+                        if (promise.tryFailure(DnsNameResolverTimeoutException.newInstance(name,
+                                resolutionTimeoutMillis, SRV.toString(), SrvRecordPublisher.class, "doDnsQuery"))) {
+                            resolveFuture.cancel(true);
+                        }
```
bryce-anderson, line 446: ffti: it may not even change the bytecode but this could be a bit cleaner as
```
if (!promise.isDone() && promise.tryFailure(... )) {
  ...
}
```
Same below.

bryce-anderson (approved):

bryce-anderson (approved):

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
```diff
@@ -258,16 +261,23 @@ public DefaultDnsServiceDiscovererBuilder ndots(final int ndots) {
...
     @Override
-    public DefaultDnsServiceDiscovererBuilder queryTimeout(final Duration queryTimeout) {
-        this.queryTimeout = queryTimeout;
+    public DefaultDnsServiceDiscovererBuilder queryTimeout(final @Nullable Duration queryTimeout) {
+        this.queryTimeout = queryTimeout == null ? null : DurationUtils.ensureNonNegative(queryTimeout, "queryTimeout");
+        return this;
+    }
+
+    @Override
+    public DefaultDnsServiceDiscovererBuilder resolutionTimeout(final @Nullable Duration resolutionTimeout) {
```
daschl, line 270: IMO the name between `resolution` and `query` is quite close and not directly obvious how they are different. WDYT about using "total" in the name to make it more clear - i.e. `totalResolutionTimeout` in addition to the javadoc clarification you have in place?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
daschl: Seems reasonable to me as well 👍

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererBuilder.java
```diff
@@ -182,19 +182,46 @@ DnsServiceDiscovererBuilder dnsServerAddressStreamProvider(
 
     /**
      * Set the number of dots which must appear in a name before an initial absolute query is made.
+     * <p>
+     * If not set, the default value is read from {@code ndots} option of {@code /etc/resolv.conf}).
```
daschl, line 186: what if it is not present in the resolv.conf?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererBuilder.java
```diff
@@ -182,19 +182,46 @@ DnsServiceDiscovererBuilder dnsServerAddressStreamProvider(
...
+     * @see #resolutionTimeout(Duration)
+     */
+    DnsServiceDiscovererBuilder queryTimeout(@Nullable Duration queryTimeout);
+
+    /**
+     * Sets the total timeout of each DNS resolution performed by this service discoverer.
+     * <p>
+     * Each resolution may execute one or more DNS queries, like following multiple CNAME(s) or trying different search
+     * domains. This is the total timeout for all intermediate queries involved in a single resolution request. Note,
+     * that <a href="https://tools.ietf.org/html/rfc2782">SRV</a> resolutions may generate independent resolutions for
```
daschl, line 210: this is a slightly different conversations, but I wonder if this is the right behavior. Even if technically it's two different resolutions, from a user perspective I still want to apply a "total" timeout, right? Since in the end they are just individual queries against a nameserver where we need a total timeout to cap?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererBuilder.java
idelpivnitskiy: docs for resolv.conf clarify the behavior if value is not specified in the file: https://man7.org/linux/man-pages/man5/resolv.conf.5.html
netty follows the same behavior. I would prefer to avoid documenting the behavior ST does not control directly, bcz either linux or netty behavior can change and it's hard to track to keep docs consistent.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererBuilder.java
idelpivnitskiy: agreed, I don't like this fact either but it's what it's. the way SRV resolutions are implemented right now makes it too complicated to apply a timeout at SRV+all_A_resolutions level. requires a big refactoring. 
taking into account, they are less frequent and used only for `BACKGROUND` discovery strategy, the need for a proper total timeout is less.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsServiceDiscovererBuilder.java
idelpivnitskiy: I debated between adding "total" or not in the name as well. Decided to drop it for 2 reasons:
1. Consistency with naming `DnsServiceDiscovererObserver.DnsResolutionObserver`. This timeouts starts when `onNewResolution` is called and stops when `DnsResolutionObserver` terminates.
2. Bcz of the SRV records, it's not really "total".

idelpivnitskiy:
@daschl I will merge to avoid conflicts with #2918. Lmk if you have more comments and I will open a follow-up

PR #2916, opened 2024-05-09
`RetryingHttpRequesterFilter`: add on request retry callback

idelpivnitskiy:
Motivation:

There are use-cases when users need to perform an action before the retry, like updating meta-data or logging/metrics. Because all retry functions return `BackoffPolicy`, it's still not known inside the function of `BackoffPolicy` will retry or not because of the retry counts limit.

Modifications:
- Add `RetryingHttpRequesterFilter.Builder.onRequestRetry(BiConsumer)` that users can use to intercept every retry attempt;
- Test that the new callback works for request retries;

Result:

Users can see when the retry actually happens, after backoff time.

If there will be a similar use-case for `reserveConnection` in the future, we will add a separate callback.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2021 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2021-2024 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -36,6 +36,7 @@ import io.servicetalk.http.api.HttpHeaderNames;
 import io.servicetalk.http.api.HttpRequestMetaData;
 import io.servicetalk.http.api.HttpResponseMetaData;
 import io.servicetalk.http.api.HttpResponseStatus;
+import io.servicetalk.http.api.StreamingHttpClient;
 import io.servicetalk.http.api.StreamingHttpClientFilter;
 import io.servicetalk.http.api.StreamingHttpClientFilterFactory;
 import io.servicetalk.http.api.StreamingHttpRequest;
@@ -48,6 +49,7 @@ import io.servicetalk.transport.api.RetryableException;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.UnaryOperator;
@@ -90,10 +92,10 @@ public final class RetryingHttpRequesterFilter
     static final int DEFAULT_MAX_TOTAL_RETRIES = 4;
     private static final RetryingHttpRequesterFilter DISABLE_AUTO_RETRIES =
             new RetryingHttpRequesterFilter(true, false, false, 1, null,
-                    (__, ___) -> NO_RETRIES);
+                    (__, ___) -> NO_RETRIES, null);
     private static final RetryingHttpRequesterFilter DISABLE_ALL_RETRIES =
             new RetryingHttpRequesterFilter(false, true, false, 0, null,
-                    (__, ___) -> NO_RETRIES);
+                    (__, ___) -> NO_RETRIES, null);
 
     private final boolean waitForLb;
     private final boolean ignoreSdErrors;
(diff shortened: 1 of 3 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -842,6 +872,22 @@ public Builder retryOther(
...
+         * <p>
+         * This can be used to track when {@link BackOffPolicy} actually decides to retry a request, to update
+         * {@link HttpRequestMetaData request meta-data} before a retry, or implement logging/metrics.
+         *
+         * @param onRequestRetry {@link BiConsumer} with {@link HttpRequestMetaData request meta-data} and
+         * {@link Throwable cause} that is invoked before every
+         * {@link StreamingHttpClient#request(StreamingHttpRequest) request} retry attempt
+         * @return {@code this}
+         */
+        public Builder onRequestRetry(final BiConsumer<HttpRequestMetaData, Throwable> onRequestRetry) {
```
daschl, line 886: Would it be useful if the number of retries on the current attempt are also made available to the consumer somehow? I know one can do this manually by setting it on the context and tracking it on their own, but I would assume that the number of attempts is something a lot of people need, if only for logging.

idelpivnitskiy: Good point. I will have to come up with a new interface. Will do that, thanks!

idelpivnitskiy: Any suggestion for the name? Assuming we may need a similar feature for `reserveConnection` retries

tkountis: Its not very apparent from the name at which stage the flow runs other than the javadoc. WDYT about including such information as part of the name? Something like `onBeforeRequestRetry`

tkountis: Regarding the interface my 2c would be `BeforeRetryConsumer` or `BeforeRetryInfluencer` to highlight the ability to change things in the flow.

tkountis (approved):
lgtm - consider altering the examples to include a case where you alter a header of the request with the retry attempt count. Often useful.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
idelpivnitskiy: I will probably go with `beforeRequestRetry(BeforeRetryCallback)` if no objections

idelpivnitskiy: I ended up keeping the builder name as-is, but clarifying "before" as part of the `RetryCallbacks` contract.

Naming is hard, open to other suggestions in the context of a new interface.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -642,6 +672,22 @@ public interface DelayedRetry {
...
     }
 
+    /**
+     * Callbacks invoked on a retry attempt.
+     */
+    @FunctionalInterface
+    public interface RetryCallbacks {
+
+        /**
+         * Called before a retry is performed.
```
daschl, line 682: Wdyt about clarifying the semantics a bit more - i..e `Called after a retry decision has been made, but before the retry is performed`. To make sure that the user does not try to influence the retry decision as part of that callback.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -642,6 +672,22 @@ public interface DelayedRetry {
...
+         *
+         * @param retryCount a current retry counter value for this attempt
+         * @param requestMetaData {@link HttpRequestMetaData} that is being retried
+         * @param cause {@link Throwable} cause for the retry
+         */
+        void beforeRetry(int retryCount, HttpRequestMetaData requestMetaData, Throwable cause);
```
daschl, line 688: lgtm

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -642,6 +672,22 @@ public interface DelayedRetry {
         Duration delay();
     }
 
+    /**
+     * Callbacks invoked on a retry attempt.
+     */
+    @FunctionalInterface
+    public interface RetryCallbacks {
```
daschl, line 679: Are you envisioning more methods to be added here in the future since the name is plural?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingHttpRequesterFilter.java
```diff
@@ -842,6 +887,21 @@ public Builder retryOther(
             return this;
         }
 
+        /**
+         * Callback invoked on every {@link StreamingHttpClient#request(StreamingHttpRequest) request} retry attempt.
+         * <p>
+         * This can be used to track when {@link BackOffPolicy} actually decides to retry a request, to update
+         * {@link HttpRequestMetaData request meta-data} before a retry, or implement logging/metrics.
```
daschl, line 894: consider also clarifying here that the implementor can't influence the retry decision at this point and needs to use other methods to do that.

daschl (approved):

tkountis (approved):

PR #2927, opened 2024-05-14
Move `DelayedRetry` to `servicetalk-client-api`

idelpivnitskiy:
Motivation:

Currently, `servicetalk-traffic-resilience-http` module must define `servicetalk-http-netty` as API dependency only to access `RetryingHttpRequesterFilter.DelayedRetry` interface. This can be avoided if we promote this interface to `servicetalk-client-api`.

Modifications:

- Introduce `io.servicetalk.client.api.DelayedRetry` interface;
- Deprecate `io.servicetalk.http.netty.RetryingHttpRequesterFilter.DelayedRetry`;
- Introduce `io.servicetalk.http.netty.RetryingHttpRequesterFilter.Builder.retryDelayedRetryExceptions(BiFunction)`;
- Deprecate `io.servicetalk.http.netty.RetryingHttpRequesterFilter.Builder.retryDelayedRetries(BiFunction)`;
- Update `DelayedRetryRequestDroppedException` to implement `io.servicetalk.client.api.DelayedRetry` instead of `io.servicetalk.http.netty.RetryingHttpRequesterFilter.DelayedRetry`;
- Update javadoc;

Result:

`servicetalk-traffic-resilience-http` doesn't have to define `servicetalk-http-netty` as API dependency.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelayedRetry.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelayedRetry.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright © 2024 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.client.api;
+
+import java.time.Duration;
+
+/**
+ * An interface that enhances any {@link Exception} to provide a constant {@link Duration delay} to be applied when
+ * it needs to be retried.
+ */
+public interface DelayedRetry {
+
+    /**
+     * A constant delay to apply.
+     *
+     * @return The {@link Duration} to apply as constant delay when retrying
+     */
+    Duration delay();
+
+    /**
+     * Returns original {@link Throwable} this {@link DelayedRetry} represents.
+     *
+     * @return the original {@link Throwable}
+     */
+    Throwable throwable();
+}
--- a/servicetalk-traffic-resilience-http/src/main/java/io/servicetalk/traffic/resilience/http/DelayedRetryRequestDroppedException.java
+++ b/servicetalk-traffic-resilience-http/src/main/java/io/servicetalk/traffic/resilience/http/DelayedRetryRequestDroppedException.java
@@ -16,7 +16,7 @@
 package io.servicetalk.traffic.resilience.http;
 
 import io.servicetalk.capacity.limiter.api.RequestDroppedException;
-import io.servicetalk.client.api.DelayedRetry;
+import io.servicetalk.client.api.DelayedRetryException;
 import io.servicetalk.http.api.HttpResponseStatus;
 
 import java.time.Duration;
@@ -32,7 +32,8 @@ import static java.util.Objects.requireNonNull;
  * its up to the application to declare whether a {@link HttpResponseStatus#TOO_MANY_REQUESTS} is a safe-to-retry
  * response, and if so after how much {@link #delay()}.
  */
-public final class DelayedRetryRequestDroppedException extends RequestDroppedException implements DelayedRetry {
+public final class DelayedRetryRequestDroppedException extends RequestDroppedException
+        implements DelayedRetryException {
 
     private static final long serialVersionUID = -7933994513110803151L;
     private final Duration delay;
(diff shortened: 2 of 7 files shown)


servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelayedRetry.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package io.servicetalk.client.api;
+
+import java.time.Duration;
+
+/**
+ * An interface that enhances any {@link Exception} to provide a constant {@link Duration delay} to be applied when
+ * it needs to be retried.
+ */
+public interface DelayedRetry {
```
idelpivnitskiy, line 24: To align with `RetryableException` marker interface and highlight that it's related to exceptions, should we name it `DelayedRetryException`?

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelayedRetry.java
```diff
@@ -0,0 +1,39 @@
...
+     * @return The {@link Duration} to apply as constant delay when retrying
+     */
+    Duration delay();
+
+    /**
+     * Returns original {@link Throwable} this {@link DelayedRetry} represents.
+     *
+     * @return the original {@link Throwable}
+     */
+    Throwable throwable();
```
idelpivnitskiy, line 38: I found it useful for users of `retryDelayedRetryExceptions` function as a way to access exception and inspect its message/cause instead of casting the marker interface to `Throwable`. Lmk wdyt.

bryce-anderson: Seems fine to me although it seems unhelpful for now considering the default implementation in `RetryingHttpRequesterFilter.DelayedRetry`.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelayedRetry.java
idelpivnitskiy: My understanding is that it's a marker only for exceptions. @tkountis ?

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelayedRetry.java
idelpivnitskiy: I had a case like that in `retryRetryableExceptions` function with `RetryableException` where I had to inspect the underlying exception type but it was not possible without a cast

servicetalk-traffic-resilience-http/src/main/java/io/servicetalk/traffic/resilience/http/DelayedRetryRequestDroppedException.java
```diff
@@ -16,22 +16,23 @@
...
+ * A {@link RequestDroppedException} to indicate that a request was dropped by a server due to capacity constraints and
+ * requires a client to delay its retry.
+ * <p>
  * This error reflects the client side application logic and its interpretation of a service response; meaning that
  * its up to the application to declare whether a {@link HttpResponseStatus#TOO_MANY_REQUESTS} is a safe-to-retry
  * response, and if so after how much {@link #delay()}.
  */
-public final class DelayedRetryRequestDroppedException extends RequestDroppedException
-        implements RetryingHttpRequesterFilter.DelayedRetry {
+public final class DelayedRetryRequestDroppedException extends RequestDroppedException implements DelayedRetry {
```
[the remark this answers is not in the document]
idelpivnitskiy: This is a new code that was not released yet. The goal is to adjust API before the release

idelpivnitskiy:
> I like the sentiment but let's see if the ABI checker is okay with it before merging.

The only API changes in this PR are highlighted in the PR description. Everything else was not released yet. I already used `japicmp.sh` to validate API compatibility for `servicetalk-http-netty` - no warnings.

servicetalk-traffic-resilience-http/src/main/java/io/servicetalk/traffic/resilience/http/DelayedRetryRequestDroppedException.java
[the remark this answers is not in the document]
bryce-anderson: My mistake: this exception hasn't been released so we can do whatever we want with it.

bryce-anderson (approved):

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelayedRetry.java
tkountis: Yup, exceptions only.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelayedRetry.java
bryce-anderson: What I'm saying is that in the face of a default implementation that just throws it's dangerous to use that method so the safe thing to do is probably cast. You could make the default method attempt a downcast which should work for the intended use case:
```
@Override
        default Throwable throwable() {
            if (this instanceof Throwable) {
                return (Throwable) this;
            }
            throw new UnsupportedOperationException("DelayedRetry#throwable() is not supported by " + getClass());
        }
        ```

tkountis (approved):

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelayedRetry.java
idelpivnitskiy: Renamed: https://github.com/apple/servicetalk/pull/2927/commits/8a1e5d435b8392c552d4c646ee47339013d43f38

PR #2938, opened 2024-05-19
Convert scheme to lower case when parsing URI

### LINKED ISSUE #1423: HttpRequestMetaData scheme case sensitivity
`HttpRequestMetaData#scheme()` javadocs state the result will be lowercase, but this is not always the case. `DefaultMultiAddressUrlHttpClientBuilder` has a few instances (`CachingKeyFactory`, `UrlKey`) which rely upon the lowercase behavior describe in `HttpRequestMetaData#scheme()`. In general the comparisons should be done in a case insensitive fashion [1].

[1] https://tools.ietf.org/html/rfc3986#section-3.1
> Although schemes are case-
   insensitive, the canonical form is lowercase and documents that
   specify schemes must do so with lowercase letters.  An implementation
   should accept uppercase letters as equivalent to lowercase in scheme
   names (e.g., allow "HTTP" as well as "http") for the sake of
   robustness but should only produce lowercase scheme names for
   consistency.

original context: https://github.com/apple/servicetalk/pull/1387#discussion_r589684620

0x1306e6d:
Motivation:

`HttpRequestMetaData#scheme()` javadocs state the result will be lowercase, but this is not always the case because `Uri3986` parses the scheme as-is. As a result, `DefaultMultiAddressUrlHttpClientBuilder` may have more than 1 client instance for the same address if users use URI with different scheme letters case. 
See https://github.com/apple/servicetalk/issues/1423 for details.

Modifications:

- Convert scheme to lower case when parsing `Uri3986` reusing well-known schemes ("http" and "https")

Result:

Fixes #1423.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Uri3986.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Uri3986.java
@@ -173,7 +173,7 @@ final class Uri3986 implements Uri {
                 if (i == 0) {
                     throw new IllegalArgumentException("Invalid URI format: no scheme before colon (':')");
                 }
-                parsedScheme = getLowercaseScheme(uri.substring(0, i));
+                parsedScheme = getLowerCaseScheme(uri.substring(0, i));
                 begin = ++i;
                 // We don't enforce the following, browsers still generate these types of requests.
                 // https://tools.ietf.org/html/rfc3986#section-3.3
@@ -200,7 +200,7 @@ final class Uri3986 implements Uri {
         this.uri = uri;
     }
 
-    private static String getLowercaseScheme(String scheme) {
+    private static String getLowerCaseScheme(String scheme) {
         if (regionMatches(scheme, true, 0, HTTP_SCHEME, 0, HTTP_SCHEME.length())) {
             if (scheme.length() == 4) {
                 return HTTP_SCHEME;
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Uri3986.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Uri3986.java
@@ -17,9 +17,11 @@
 package io.servicetalk.http.api;
 
 import java.nio.charset.Charset;
+import java.util.Locale;
 import java.util.Objects;
 import javax.annotation.Nullable;
 
+import static io.servicetalk.buffer.api.CharSequences.regionMatches;
 import static io.servicetalk.http.api.UriComponentType.FRAGMENT;
 import static io.servicetalk.http.api.UriComponentType.HOST_NON_IP;
 import static io.servicetalk.http.api.UriComponentType.PATH;
@@ -38,6 +40,8 @@ import static io.servicetalk.http.api.UriUtils.parsePort;
 final class Uri3986 implements Uri {
     @SuppressWarnings({"StringOperationCanBeSimplified", "PMD.StringInstantiation"})
     private static final String NULL_COMPONENT = new String(""); // instance equality required!
+    private static final String HTTP_SCHEME = "http";
+    private static final String HTTPS_SCHEME = "https";
     private final String uri;
     @Nullable
     private final String scheme;
@@ -169,7 +173,7 @@ final class Uri3986 implements Uri {
                 if (i == 0) {
                     throw new IllegalArgumentException("Invalid URI format: no scheme before colon (':')");
                 }
-                parsedScheme = uri.substring(0, i);
+                parsedScheme = getLowercaseScheme(uri.substring(0, i));
                 begin = ++i;
                 // We don't enforce the following, browsers still generate these types of requests.
                 // https://tools.ietf.org/html/rfc3986#section-3.3
@@ -194,6 +198,18 @@ final class Uri3986 implements Uri {
         port = parsedPort;
         path = parsedPath;
         this.uri = uri;
+    }
+
+    private static String getLowercaseScheme(String scheme) {
+        if (regionMatches(scheme, true, 0, HTTP_SCHEME, 0, HTTP_SCHEME.length())) {
+            if (scheme.length() == 4) {
+                return HTTP_SCHEME;
+            }
+            if (scheme.length() == 5 && (scheme.charAt(4) == 's' || scheme.charAt(4) == 'S')) {
+                return HTTPS_SCHEME;
+            }
+        }
+        return scheme.toLowerCase(Locale.ENGLISH);
     }
 
     @Override
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Uri3986.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Uri3986.java
@@ -21,7 +21,6 @@ import java.util.Locale;
 import java.util.Objects;
 import javax.annotation.Nullable;
 
-import static io.servicetalk.buffer.api.CharSequences.regionMatches;
 import static io.servicetalk.http.api.UriComponentType.FRAGMENT;
 import static io.servicetalk.http.api.UriComponentType.HOST_NON_IP;
 import static io.servicetalk.http.api.UriComponentType.PATH;
@@ -173,7 +172,7 @@ final class Uri3986 implements Uri {
                 if (i == 0) {
                     throw new IllegalArgumentException("Invalid URI format: no scheme before colon (':')");
                 }
-                parsedScheme = getLowerCaseScheme(uri.substring(0, i));
+                parsedScheme = getLowerCaseScheme(uri, i);
                 begin = ++i;
                 // We don't enforce the following, browsers still generate these types of requests.
                 // https://tools.ietf.org/html/rfc3986#section-3.3
@@ -200,16 +199,13 @@ final class Uri3986 implements Uri {
         this.uri = uri;
     }
 
-    private static String getLowerCaseScheme(String scheme) {
-        if (regionMatches(scheme, true, 0, HTTP_SCHEME, 0, HTTP_SCHEME.length())) {
-            if (scheme.length() == 4) {
-                return HTTP_SCHEME;
-            }
-            if (scheme.length() == 5 && (scheme.charAt(4) == 's' || scheme.charAt(4) == 'S')) {
-                return HTTPS_SCHEME;
-            }
+    private static String getLowerCaseScheme(final String uri, final int endIndex) {
+        if (endIndex == 4 && uri.regionMatches(true, 0, HTTP_SCHEME, 0, 4)) {
+            return HTTP_SCHEME;
+        } else if (endIndex == 5 && uri.regionMatches(true, 0, HTTPS_SCHEME, 0, 5)) {
+            return HTTPS_SCHEME;
         }
-        return scheme.toLowerCase(Locale.ENGLISH);
+        return uri.substring(0, endIndex).toLowerCase(Locale.ENGLISH);
     }
 
     @Override
--- a/servicetalk-http-api/src/test/java/io/servicetalk/http/api/Uri3986Test.java
+++ b/servicetalk-http-api/src/test/java/io/servicetalk/http/api/Uri3986Test.java
@@ -31,7 +31,11 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.regex.Pattern.compile;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalToIgnoringCase;
 import static org.hamcrest.Matchers.lessThan;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.sameInstance;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
@@ -162,8 +166,24 @@ class Uri3986Test {
     }
 
     @Test
-    void schemeParsedCaseSensative() {
-        verifyUri3986("HTTP://test.com", "HTTP", null, "test.com", -1, "", "", null, null, null);
+    void schemeParsedCaseInsensitive() {
+        verifyUri3986("HTTP://test.com", "http", null, "test.com", -1, "", "", null, null, null);
+        verifyUri3986("HTTPS://test.com", "https", null, "test.com", -1, "", "", null, null, null);
+        verifyUri3986("ANY://test.com", "any", null, "test.com", -1, "", "", null, null, null);
+    }
+
+    @Test
+    void reuseWellKnownScheme() {
+        String http1 = new Uri3986("http://test.com").scheme();
+        String http2 = new Uri3986("HTTP://test.com").scheme();
+        assertThat("different http", http1, sameInstance(http2));
+
+        String https1 = new Uri3986("https://test.com").scheme();
+        String https2 = new Uri3986("HTTPS://test.com").scheme();
+        assertThat("different https", https1, sameInstance(https2));
+
+        String httpss = new Uri3986("HTTPSS://test.com").scheme();
+        assertThat("same https and httpss", https1, not(sameInstance(httpss)));
     }
 
     @Test
@@ -637,7 +657,9 @@ class Uri3986Test {
         // Validate against the RFC's regex
         Matcher matcher = VALID_PATTERN.matcher(expectedUri);
         assertThat(true, is(matcher.matches()));
-        assertThat("invalid scheme()", uri.scheme(), is(matcher.group(2)));
+        String matcherScheme = matcher.group(2);
+        assertThat("invalid scheme()", uri.scheme(),
+                matcherScheme == null ? is(nullValue()) : equalToIgnoringCase(matcherScheme));
         assertThat("invalid authority()", uri.authority(), is(matcher.group(4)));
         assertThat("invalid path()", uri.path(), is(matcher.group(5)));
         assertThat("invalid query()", uri.query(), is(matcher.group(7)));
--- a/servicetalk-http-api/src/test/java/io/servicetalk/http/api/Uri3986Test.java
+++ b/servicetalk-http-api/src/test/java/io/servicetalk/http/api/Uri3986Test.java
@@ -33,7 +33,6 @@ import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalToIgnoringCase;
 import static org.hamcrest.Matchers.lessThan;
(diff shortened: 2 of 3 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2SchemeTest.java
```diff
@@ -158,7 +157,7 @@ private SchemeValidatorHandler(final HttpScheme expectedScheme) {
         public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
             if (msg instanceof Http2HeadersFrame) {
                 Http2Headers headers = ((Http2HeadersFrame) msg).headers();
-                if (!expectedScheme.name().equals(headers.scheme())) {
+                if (!expectedScheme.name().contentEquals(headers.scheme())) {
```
0x1306e6d, line 160: To prevent a possible backward compatibility breaking, it might be better to return `AsciiString` on `HttpScheme#name()` like Netty's `HttpScheme#name()`.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MultiAddressUrlHttpClientTest.java
```diff
@@ -191,7 +191,7 @@ void requestWithRequestTargetWithoutScheme() {
         StreamingHttpRequest request = client.get(format("%s/200?param=value#tag", hostHeader));
         // no host header
         toSource(client.request(request)).subscribe(subscriber);
-        assertThat(subscriber.awaitOnError(), is(instanceOf(MalformedURLException.class)));
+        assertThat(subscriber.awaitOnError(), is(instanceOf(IllegalArgumentException.class)));
```
0x1306e6d, line 194: `IllegalArgumentException` is ok?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpScheme.java
```diff
@@ -0,0 +1,113 @@
...
+     */
+    public static HttpScheme of(String scheme) {
+        requireNonNull(scheme, "scheme");
+        if (http().name().equalsIgnoreCase(scheme)) {
+            return http();
+        }
+        if (https().name().equalsIgnoreCase(scheme)) {
+            return https();
+        }
+        throw new IllegalArgumentException("Invalid HTTP scheme: " + scheme);
```
idelpivnitskiy, line 71: If in the future we add support for WebSockets, we can definitely add more schemes as constants, but what if users use this HTTP client for other protocols, like WebDAV extensions that use `dav` scheme?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpScheme.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        final HttpScheme that = (HttpScheme) o;
+        return defaultPort == that.defaultPort && Objects.equals(name, that.name);
```
idelpivnitskiy, line 101: Reminder for myself: if in other iterations there will be a way to create a new `HttpScheme`, either this comparison should be case-insensitive or ctor should always lowercase the passed `String`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpScheme.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, defaultPort);
```
idelpivnitskiy, line 106: We try to avoid using `Objects.hash(varargs)` because it always allocates an array

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequestMetaData.java
```diff
@@ -140,7 +142,14 @@ public HttpRequestMetaData requestTarget(final String requestTarget, final Chars
     @Nullable
     @Override
     public final String scheme() {
-        return lazyParseRequestTarget().scheme();
+        if (scheme == null) {
+            final String rawScheme = lazyParseRequestTarget().scheme();
+            if (rawScheme == null) {
+                return null;
+            }
+            scheme = HttpScheme.of(rawScheme);
```
idelpivnitskiy, line 150: The biggest risk is here, if another scheme is used, this method will start throwing `IllegalArgumentException`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpScheme.java
```diff
@@ -0,0 +1,113 @@
...
+package io.servicetalk.http.api;
+
+import java.util.Objects;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Represents HTTP protocol schemes.
+ */
+public final class HttpScheme {
```
idelpivnitskiy, line 25: What is the value of adding a new public class if it's not used for `HttpRequestMetaData` API?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
```diff
@@ -222,6 +223,45 @@ void internalClientsUseDifferentExecutionContextWhenConfigured() throws Exceptio
...
+                .executionStrategy(offloadNone())
+                .listenStreamingAndAwait((ctx, request, responseFactory) -> succeeded(responseFactory.ok()));
+             BlockingHttpClient blockingHttpClient = HttpClients.forMultiAddressUrl(getClass().getSimpleName())
+                     .initializer((scheme, address, builder) ->
+                             builder.appendClientFilter(
+                                     client -> {
+                                         counter.incrementAndGet();
+                                         return new StreamingHttpClientFilter(client) {
+                                         };
+                                     }))
```
idelpivnitskiy, line 246: This can be replaced by 
```
.initializer((scheme, address, builder) -> counter.incrementAndGet())
```

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
```diff
@@ -222,6 +223,45 @@ void internalClientsUseDifferentExecutionContextWhenConfigured() throws Exceptio
         }
     }
 
+    @Test
+    void internalClientSchemeShouldBeCaseInsensitive() throws Exception {
+        HttpExecutionContext expectedCtx = new HttpExecutionContextBuilder()
```
idelpivnitskiy, line 228: For this specific test, there is no need to customize the context. Just skip all these methods on the client builder

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
```diff
@@ -222,6 +223,45 @@ void internalClientsUseDifferentExecutionContextWhenConfigured() throws Exceptio
...
+                .ioExecutor(CTX.ioExecutor())
+                .executor(CTX.executor())
+                .bufferAllocator(PREFER_DIRECT_ALLOCATOR)
+                .executionStrategy(offloadNone())
+                .build();
+        AtomicInteger counter = new AtomicInteger();
+
+        try (ServerContext serverContext = HttpServers.forAddress(localAddress(0))
+                .executionStrategy(offloadNone())
```
idelpivnitskiy, line 237: You can also skip this line

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
```diff
@@ -222,6 +223,45 @@ void internalClientsUseDifferentExecutionContextWhenConfigured() throws Exceptio
...
+
+            HttpResponse response = blockingHttpClient.request(
+                    blockingHttpClient.get("http://" + serverHostAndPort(serverContext)));
+            assertThat(response.status(), is(OK));
+
+            response = blockingHttpClient.request(
+                    blockingHttpClient.get("HTTP://" + serverHostAndPort(serverContext)));
+            assertThat(response.status(), is(OK));
+
+            assertThat(counter.get(), is(1));
```
idelpivnitskiy, line 261: Thanks for the reproducer! An alternative (minimal) way to fix the issue without introducing new types will be:
```patch
Index: servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java	(revision f4d7153d5606531fbdd6640d29502970b19e8c97)
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java	(date 1716514112838)
@@ -58,6 +58,7 @@
 
 import java.net.MalformedURLException;
+import java.util.Locale;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
@@ -165,15 +166,16 @@
                                 ", expected absolute-form URL");
             }
 
-            final String scheme = metaData.scheme();
-            if (scheme == null) {
+            final String originalScheme = metaData.scheme();
+            if (originalScheme == null) {
                 throw new MalformedURLException("Request-target does not contains scheme: " +
                         metaData.requestTarget() + ", expected absolute-form URL");
             }
+            final String scheme = originalScheme.toLowerCase(Locale.ENGLISH);
 
             final int parsedPort = metaData.port();
             final int port = parsedPort >= 0 ? parsedPort :
-                    (HTTPS_SCHEME.equalsIgnoreCase(scheme) ? defaultHttpsPort : defaultHttpPort);
+                    (HTTPS_SCHEME.equals(scheme) ? defaultHttpsPort : defaultHttpPort);
             setHostHeader(metaData);
             metaData.requestTarget(absoluteToRelativeFormRequestTarget(metaData.requestTarget(), scheme, host));
```
Another option is to always lowercase inside `DefaultHttpRequestMetaData.scheme()` implementation to aligs with its javadoc. 

To reduce memory allocations in both cases, we can enhance the code to guess the standard schemes first before doing lower casing.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/MultiAddressUrlHttpClientTest.java
idelpivnitskiy: No, this is a change in behavior

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1ClientDuplexHandler.java
```diff
@@ -63,7 +63,7 @@ final class H2ToStH1ClientDuplexHandler extends AbstractH2DuplexHandler {
     H2ToStH1ClientDuplexHandler(boolean sslEnabled, BufferAllocator allocator, HttpHeadersFactory headersFactory,
                                 CloseHandler closeHandler, StreamObserver observer) {
         super(allocator, headersFactory, closeHandler, observer);
-        this.scheme = sslEnabled ? HttpScheme.HTTPS : HttpScheme.HTTP;
+        this.scheme = sslEnabled ? HttpScheme.https() : HttpScheme.http();
```
idelpivnitskiy, line 66: It's preferable to keep Netty's `HttpScheme` here to take advantage of `AsciiString`

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2SchemeTest.java
idelpivnitskiy: We can not expose Netty types in our public API, we can only do this for internal utilities.

idelpivnitskiy (commented):
Hi Gihwan,
Thanks for the interest in our project and sorry for the delayed review. 
Please see my comments below:

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
idelpivnitskiy: After thinking about it more, it feels like the most minimalistic and efficient way is to enhance this line:
https://github.com/apple/servicetalk/blob/b8149d36ba1d4cddf4b034b18f3cdb4c5a358adc/servicetalk-http-api/src/main/java/io/servicetalk/http/api/Uri3986.java#L172

We can have an internal static function that will return a lowercase scheme. For well-known and expected schemes (http and https) it can even avoid allocation of a new string by checking the first 4-5 chars of `uri` and returning a corresponding constant. 

WDYT?

0x1306e6d: Thank you for your kind review, @idelpivnitskiy!

I also prefer the second one and apply it. It becomes much more minimal.

nit: `Locale.ROOT` instead of `Locale.ENGLISH`?

> To obtain correct results for locale insensitive strings, use {@code toLowerCase(Locale.ROOT)}.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/Uri3986Test.java
```diff
@@ -162,8 +166,24 @@ void emptyUserInfo() {
...
+        String http1 = new Uri3986("http://test.com").scheme();
+        String http2 = new Uri3986("HTTP://test.com").scheme();
+        assertThat("different http", http1, sameInstance(http2));
+
+        String https1 = new Uri3986("https://test.com").scheme();
+        String https2 = new Uri3986("HTTPS://test.com").scheme();
+        assertThat("different https", https1, sameInstance(https2));
+
+        String httpss = new Uri3986("HTTPSS://test.com").scheme();
+        assertThat("same https and httpss", https1, not(sameInstance(httpss)));
```
idelpivnitskiy, line 186: It's expected because `https` != `httpss`. Doesn't look like there is much value in the last 2 lines of this test, consider verifying only `http` and `https`.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
idelpivnitskiy: idk, feels more predictable on any OS. based on IANA list of registered schemes, all of them are in English: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Uri3986.java
```diff
@@ -196,6 +200,18 @@ final class Uri3986 implements Uri {
         this.uri = uri;
     }
 
+    private static String getLowerCaseScheme(String scheme) {
+        if (regionMatches(scheme, true, 0, HTTP_SCHEME, 0, HTTP_SCHEME.length())) {
+            if (scheme.length() == 4) {
```
idelpivnitskiy, line 205: To avoid expensive `regionMatches` operation in case of unknown scheme, consider first checking the length. If it's 4, run `regionMatches` for HTTP, and if it's 5 - for HTTPS. Consider using `scheme.regionMatches(...)` instead of `CharSequences`, because we know both are `String`s -> no need to go through an extra `instanceof` check.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Uri3986.java
```diff
@@ -169,7 +173,7 @@ final class Uri3986 implements Uri {
                 if (i == 0) {
                     throw new IllegalArgumentException("Invalid URI format: no scheme before colon (':')");
                 }
-                parsedScheme = uri.substring(0, i);
+                parsedScheme = getLowerCaseScheme(uri.substring(0, i));
```
idelpivnitskiy, line 176: If you are using `regionMatches`, there is an opportunity to skip allocation of a `substring` for the happy path if you pass `uri` and `i` as separate arguments.

idelpivnitskiy (commented):
Looks much better, thanks for refactoring!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Uri3986.java
0x1306e6d: Indeed!
Additionally, I think we can reduce the allocation of the substring:

```java
    private static String getLowerCaseScheme(String uri, int endIndex) {
        if (endIndex == 4 && uri.regionMatches(true, 0, HTTP_SCHEME, 0, 4)) {
            return HTTP_SCHEME;
        } else if (endIndex == 5 && uri.regionMatches(true, 0, HTTPS_SCHEME, 0, 5)) {
            return HTTPS_SCHEME;
        }
        final char[] scheme = new char[endIndex];
        for (int i = 0; i < scheme.length; ++i) {
            scheme[i] = Character.toLowerCase(uri.charAt(i));
        }
        return new String(scheme);
    }
```

What do you think it?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilderTest.java
0x1306e6d: Thank you, I see.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/Uri3986.java
idelpivnitskiy: Taking into account that this is an unlikely fallback, I think `substring` followed by `toLoweCase` will be more readable and likely more efficient. `new String` does yet another allocation and attempts to do a compact string, while `substring` already knows if the original `uri` is compact or not and simply copies a subrange. Then `toLowerCase` has a non-allocating path if everything is already in lowercase.

idelpivnitskiy (approved):
Many thanks! Appreciate your contribution

PR #2947, opened 2024-05-29
Reduce default TLS handshake timeout from 10 to 5 seconds

idelpivnitskiy:
Motivation:

10 seconds feels like too high value to hold a connection attempt by waiting for TLS handshake to complete. Because TLS handshake happens before ST writes a request, it's always marked as `RetryableException` and is retried up to 3 times.

Modifications:

- Reduce `DEFAULT_HANDSHAKE_TIMEOUT` from 10 to 5 seconds;

Result:

Less wait time for TLS handshake process that hangs longer than expected.


tkountis (approved):
Still feels high, but in lack of data to drive a better default lgtm

bryce-anderson (approved):

daschl (approved):

PR #2949, opened 2024-06-04
Fallback to `readChunkSize` if `InputStream.available()` not implemented

idelpivnitskiy:
Motivation:

`FromInputStreamPublisher` attempts to read only 1 byte when `InputStream.available()` is not implemented. As a result, users of blocking streaming API with payload body as `InputStream` may write and flush by 1 byte if their `InputStream` implementation always returns 0 available bytes (default).

Modifications:

- Consider `available()` as a best effort to avoid blocking. If it returns 0 bytes, attempt to read up to `readChunkSize` bytes;

Result:

Improved efficiency of writing `InputStream` data for outgoing requests.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2018-2021, 2023-2024 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
```diff
@@ -171,7 +171,7 @@ private void readAndDeliver(final Subscriber<? super byte[]> subscriber) {
                     int available = stream.available();
                     if (available == 0) {
                         // Work around InputStreams that don't strictly honor the 0 == EOF contract.
-                        available = buffer != null ? buffer.length : 1;
+                        available = buffer != null ? buffer.length : readChunkSize;
```
bryce-anderson, line 174: As you mentioned offline, the default value is 64kB which could be rough on GC in adversarial cases. Do you plan to change that as well in a different PR, or are we fine trying it as-is?

bryce-anderson (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
idelpivnitskiy: Yes, I already played with different values and wire-logging enabled, 64Kb is not a good default.
Still need to run some benchmarks and will update the value in a separate PR.

daschl (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
daschl: As a data point, in AHC the `InputStreamEntity` uses `OUTPUT_BUFFER_SIZE = 4096;` as the chunk size for this.

PR #2951, opened 2024-06-04
Reduce allocations for `ServiceTalkBufferAllocator.wrap(byte[], int, int)`

idelpivnitskiy:
Motivation:

Default impl in `BufferAllocator` takes a slice of a wrapped original byte-array. We can avoid double wrapping with `NettyBuffer` if we implement this method in `ServiceTalkBufferAllocator`.

Modifications:

- Override `BufferAllocator.wrap(byte[], int, int)` to wrap with `NettyBuffer` only one time;
- Enhance `BufferAllocatorsTest` to validate wrapping and ensure the wrapped `Buffer` always has readerIndex set to 0;

Result:

Minus 2 object allocations when users wrap a byte-array partially.

### DIFF
--- a/servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ServiceTalkBufferAllocator.java
+++ b/servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ServiceTalkBufferAllocator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018-2021, 2024 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2018-2020 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ServiceTalkBufferAllocator.java
+++ b/servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ServiceTalkBufferAllocator.java
@@ -161,8 +161,10 @@ final class ServiceTalkBufferAllocator extends AbstractByteBufAllocator implemen
 
     @Override
     public Buffer wrap(final byte[] bytes, final int offset, final int len) {
-        return bytes.length == 0 ? EMPTY_BUFFER :
-                new NettyBuffer<>(new UnreleasableHeapByteBuf(this, bytes, bytes.length).slice(offset, len));
+        if (offset == 0 && len == bytes.length) {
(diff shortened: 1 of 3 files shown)


servicetalk-buffer-netty/src/test/java/io/servicetalk/buffer/netty/BufferAllocatorsTest.java
```diff
@@ -76,6 +84,42 @@ void testReadOnlyDirectBuffer(BufferAllocator allocator) {
         assertBuffer(allocator.wrap(ByteBuffer.allocateDirect(16).asReadOnlyBuffer()), true);
     }
```
daschl, line 86: consider also testing negative case - i.e length is negative and/or offset is more than bytes length.

daschl (commented):
minor comment about test coverage, optimization looks good!

servicetalk-buffer-netty/src/main/java/io/servicetalk/buffer/netty/ServiceTalkBufferAllocator.java
```diff
@@ -159,6 +159,17 @@ public Buffer wrap(byte[] bytes) {
...
     }
 
+    @Override
+    public Buffer wrap(final byte[] bytes, final int offset, final int len) {
+        if (len == 0) {
+            return EMPTY_BUFFER;
+        }
+        final Buffer b = wrap(bytes);
+        b.readerIndex(offset);
+        b.writerIndex(offset + len);
```
bryce-anderson, line 169: In my minds eye this might be an unexpected behavior change: the resulting buffer will have different reader and writer indexes than expected as if they had sliced the `byte[]` themselves beforehand. This could be a trip hazard as users often will reset the reader index to 0 to reuse the buffer efficiently.

daschl (approved):

bryce-anderson (approved):
Looks good after fixing style check. I think you can still eliminate a `NettyBuffer` allocation if you override the `wrap(byte[], int, int)` method to slice the netty buffer before wrapping with `NettyBuffer`, although it may be that it was stack allocated all along.

servicetalk-buffer-netty/src/test/java/io/servicetalk/buffer/netty/BufferAllocatorsTest.java
idelpivnitskiy: Added `testWrapPartialByteArrayWithIllegalArguments`

idelpivnitskiy (commented):
Forgot to publish my reply comments y'day

bryce-anderson (approved):

PR #2972, opened 2024-06-16
Multi-address client: don't leak full request target in exception msg

idelpivnitskiy:
Motivation:

Request target may contain sensitive information, like secret query parameters or tokens in path. We should not include it as-is when `MalformedURLException` is thrown.

Modifications:

- Change how `MalformedURLException` is generated to include only small substring of the request target in the exception message.

Result:

No risk to leak sensitive data in exception logs.


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -158,31 +159,34 @@ private static final class CachingKeyFactory implements AsyncCloseable {
...
             final UrlKey urlKey = urlKeyCache.get(key);
             return urlKey != null ? urlKey : urlKeyCache.computeIfAbsent(key, ignore ->
                     new UrlKey(scheme, HostAndPort.of(host, port)));
         }
 
+        private static String ensureUrlComponentNonNull(@Nullable final String value,
+                                                        final String name,
+                                                        final String requestTarget) throws MalformedURLException {
+            if (value == null) {
+                // 8 characters should be enough to give users an idea of what's wrong with the passed URL without
```
daschl, line 181: Since we are only checking for scheme and host, what about omitting the path completely as part of the error?

daschl (approved):

PR #2975, opened 2024-06-17
Multi-address client: take advantage of lowercase scheme

idelpivnitskiy:
Motivation:

In https://github.com/apple/servicetalk/pull/2938 we made it guaranteed that `HttpRequestMetaData#scheme()` result is always in lowercase. We can use that to simplify scheme checks inside `DefaultMultiAddressUrlHttpClientBuilder`.

Modifications:
- Replace all scheme checks from `equalsIgnoreCase` to `equals`;

Result:

No need to worry about cases for scheme check.

### DIFF
--- a/servicetalk-examples/http/http2/src/main/java/io/servicetalk/examples/http/http2/alpn/HttpUrlClientWithAlpn.java
+++ b/servicetalk-examples/http/http2/src/main/java/io/servicetalk/examples/http/http2/alpn/HttpUrlClientWithAlpn.java
@@ -39,7 +39,7 @@ public final class HttpUrlClientWithAlpn {
         try (BlockingHttpClient client = HttpClients.forMultiAddressUrl().initializer((scheme, address, builder) -> {
             // If necessary, users can also take `address` into account for setting distinct protocols or TLS
             // configurations for various server addresses.
-            if ("https".equalsIgnoreCase(scheme)) {
+            if ("https".equals(scheme)) {
                 builder.protocols(h2Default(), h1Default()) // Configure support for HTTP/2 and HTTP/1.1 protocols
                 // Note: DefaultTestCerts contains self-signed certificates that may be used only for local testing.
                 // or demonstration purposes. Never use those for real use-cases.
--- a/servicetalk-examples/http/http2/src/main/java/io/servicetalk/examples/http/http2/alpn/HttpUrlClientWithAlpn.java
+++ b/servicetalk-examples/http/http2/src/main/java/io/servicetalk/examples/http/http2/alpn/HttpUrlClientWithAlpn.java
@@ -39,7 +39,7 @@ public final class HttpUrlClientWithAlpn {
         try (BlockingHttpClient client = HttpClients.forMultiAddressUrl().initializer((scheme, address, builder) -> {
             // If necessary, users can also take `address` into account for setting distinct protocols or TLS
             // configurations for various server addresses.
-            if ("https".equals(scheme)) {
+            if ("https".equalsIgnoreCase(scheme)) {
                 builder.protocols(h2Default(), h1Default()) // Configure support for HTTP/2 and HTTP/1.1 protocols
                 // Note: DefaultTestCerts contains self-signed certificates that may be used only for local testing.
                 // or demonstration purposes. Never use those for real use-cases.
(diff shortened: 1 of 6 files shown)


servicetalk-examples/http/http2/src/main/java/io/servicetalk/examples/http/http2/alpn/HttpUrlClientWithAlpn.java
```diff
@@ -39,7 +39,7 @@ public static void main(String[] args) throws Exception {
         try (BlockingHttpClient client = HttpClients.forMultiAddressUrl().initializer((scheme, address, builder) -> {
             // If necessary, users can also take `address` into account for setting distinct protocols or TLS
             // configurations for various server addresses.
-            if ("https".equalsIgnoreCase(scheme)) {
+            if ("https".equals(scheme)) {
```
idelpivnitskiy, line 42: How do you feel about changing examples and clarifying the contract in `MultiAddressHttpClientBuilder` javadoc?
The only risk is that examples can be copy pasted and if users messed up and used a non-lowercase constant on the left hand side, they will have a problem. It might be safer to teach them to use `equalsIgnoreCase`.

bryce-anderson: Good observation re copy pasting problems. There is a reasonable chance people are using constants and they don't actually know what the string is. If I had to pick, I'd say leave it as `.equalsIgnoreCase` since as you noted, it's safer and still correct, if perhaps redundant.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -96,7 +100,11 @@ final class DefaultMultiAddressUrlHttpClientBuilder
 
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultMultiAddressUrlHttpClientBuilder.class);
 
-    private static final String HTTPS_SCHEME = HTTPS.toString();
+    // Use HttpRequestMetaData to access "https" constant used by Uri3986 class to optimize "equals" check to be a
+    // trivial reference check.
```
bryce-anderson, line 104: Clever. 😄

bryce-anderson (approved):
I'm not strongly opinionated about the examples so feel free to change or ship as-is.

servicetalk-examples/http/http2/src/main/java/io/servicetalk/examples/http/http2/alpn/HttpUrlClientWithAlpn.java
idelpivnitskiy: Ok, I will revert examples then

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
```diff
@@ -44,7 +44,7 @@ interface SingleAddressInitializer<U, R> {
         /**
          * Configures the passed {@link SingleAddressHttpClientBuilder} for the given {@code scheme} and
          * {@code address}.
-         * @param scheme The scheme parsed from the request URI.
+         * @param scheme The scheme parsed from a {@link HttpRequestMetaData#requestTarget() request URI} in lowercase.
```
idelpivnitskiy, line 47: What about this? Should we freeze this contract in javadoc or also leave us some flexibility? 
The javadoc for `HttpRequestMetaData#schem()` already says it's always in lowercase.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
idelpivnitskiy: The `selectClient` was about 10% of CPU on the flame graph. There will be a few more follow-ups in this area, I'm just trying to isolate independent changes in separate PRs

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiAddressHttpClientBuilder.java
idelpivnitskiy: Decided to keep this PR purely internal. If you feel we need to update javadoc, I will open a follow-up

PR #2979, opened 2024-06-18
Multi-address client: optimize client selection

idelpivnitskiy:
Motivation:

`selectClient` attracted my attention after I saw it can take up to 10% on the flame graph.

Modifications:
- Change `CachingKeyFactory` to `UrlKeyFactory`;
- Remove an extra layer of caching for `UrlKey`s;
- Eagerly compute `UrlKey.hashCode` because it's used >1 time;
- Remove allocation of `HostAndPort` from the hot path because it's required only to create a new client once;

Result:

Client selection reduced from `63.788 ± 1.268  ns/op` to `14.512 ± 0.233  ns/op`.

---

I tried a few different approaches. All of them are shown here as independent commits.
```
Benchmark                                                    (lowercase)  Mode  Cnt   Score   Error  Units
Original                                                            true  avgt    5  63.788 ± 1.268  ns/op
Original                                                           false  avgt    5  64.920 ± 0.517  ns/op
v1: UrlKey caches hashCode                                          true  avgt    5  44.918 ± 0.685  ns/op
v1: UrlKey caches hashCode                                         false  avgt    5  45.352 ± 0.510  ns/op
v2: remove urlKeyCache, key as a string                             true  avgt    5  81.459 ± 1.087  ns/op
v2: remove urlKeyCache, key as a string                            false  avgt    5  82.457 ± 1.015  ns/op
v3: remove urlKeyCache, key as UrlKey                               true  avgt    5  19.767 ± 0.350  ns/op
v3: remove urlKeyCache, key as UrlKey                              false  avgt    5  19.813 ± 0.048  ns/op
v4: remove urlKeyCache, key as UrlKey, don't allocate HostAndPort   true  avgt    5  14.512 ± 0.233  ns/op
v4: remove urlKeyCache, key as UrlKey, don't allocate HostAndPort  false  avgt    5  14.808 ± 0.324  ns/op
```


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -213,47 +202,38 @@ private static String absoluteToRelativeFormRequestTarget(final String requestTa
...
             if (this == o) {
                 return true;
             }
-            if (o == null || getClass() != o.getClass()) {
-                return false;
-            }
-
             final UrlKey urlKey = (UrlKey) o;
-            return scheme.equals(urlKey.scheme) && hostAndPort.equals(urlKey.hostAndPort);
+            return port == urlKey.port && scheme.equals(urlKey.scheme) && host.equalsIgnoreCase(urlKey.host);
```
bryce-anderson, line 231: ffti: in my minds eye, the optimal ordering of checks would be `port -> host -> scheme`. It would seem uncommon for the host and port to match but not the scheme other than in TLS-detection scenarios which are uncommon and typically transient. I imagine in real scenarios the host will be the overwhelmingly highest cardinality of the three, but it's so cheap to check the port that it makes sense to do it first.

bryce-anderson (approved):
I like the simplification, that alone is worth it for me.

I am curious about the benchmarks: are they already committed somewhere or are they local? I'm specifically curious whether a lot of the benefit comes from the port being the first thing checked if the ports are set as random numbers vs the most common 80/443.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
idelpivnitskiy: Because the scheme comes from `Uri3986`, it's always a trivial ref check for http and https. I changed the order anyway, bcz agreed that scheme has less variance.

idelpivnitskiy:
> I am curious about the benchmarks: are they already committed somewhere or are they local?

See commits in this PR. I added the benchmark and removed it later because it required some internal changes for multi-address client builder. 

> I'm specifically curious whether a lot of the benefit comes from the port being the first thing checked if the ports are set as random numbers vs the most common 80/443.

I used a fixed port bcz int check seems trivial to make any difference.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -213,47 +202,38 @@ private static String absoluteToRelativeFormRequestTarget(final String requestTa
...
+        private final int hashCode;
 
-        UrlKey(final String scheme, final HostAndPort hostAndPort) {
+        UrlKey(final String scheme, final String host, final int port) {
             this.scheme = scheme;
-            this.hostAndPort = hostAndPort;
+            this.host = host;
+            this.port = port;
+            // hashCode is required at least once, but may be necessary multiple times for a single selectClient run
+            this.hashCode = 31 * (caseInsensitiveHashCode(host) + port) + scheme.hashCode();
```
idelpivnitskiy, line 218: I also tried removing `31 * ` from here, but it doesn't make any difference 😄

bryce-anderson (approved):
Looks good after you address the spotbugs complaints regarding the omission of the type and null checking in `UrlKey.equals(..)`. (I don't care if they're ignored or added back.)

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
bryce-anderson: I'd leave it for consistency.

idelpivnitskiy:
Verified with the benchmark that it doesn't make any difference, added them back

bryce-anderson (approved):

daschl (approved):

PR #2983, opened 2024-06-18
Add Gradle Lock files

idelpivnitskiy:
Motivation:

Lock files will help us see what dependencies are used by every module. When we bump any dependency, it will also show what transitive dependencies are upgraded automatically.
See: https://docs.gradle.org/current/userguide/dependency_locking.html 

Modifications:

- Update main `build.gradle` to activate dependency locking;
- Add `check-lockfiles.sh` script for PRQ validation;
- Add `ci-prq.yml` step to run `check-lockfiles.sh`;
- Generate initial lockfiles;

Result:

We will be able to observe how dependency updates change all transitive dependencies.


build.gradle
```diff
@@ -67,4 +67,19 @@ subprojects {
       }
     }
   }
+
+  configurations {
```
daschl, line 71: Do we have configurations which we do not want to lock?

```
dependencyLocking {
    lockAllConfigurations()
}
```

might be simpler if not.

gradlew
```diff
@@ -246,4 +246,5 @@ eval "set -- $(
         tr '\n' ' '
     )" '"$@"'
 
-exec "$JAVACMD" "$@"
+# Re-generate locks on every task
+exec "$JAVACMD" "$@" --write-locks
```
daschl, line 250: Is this something we _want_ on every task or should it be done on purpose when changing a dependency?

build.gradle
idelpivnitskiy: I tried that first and gradle discovered 20+ configurations, most related to tests, quality tasks, and plugins. We only need to keep an eye on the dependencies we bring to us and users.

gradlew
idelpivnitskiy: I borrowed this trick from another project. The problem is that ppl will always forget to re-generate the lock

AkshDesai04 (approved):

daschl (approved):

PR #2985, opened 2024-06-25
Set `SslContextBuilder.endpointIdentificationAlgorithm` for Netty 4.2.x

idelpivnitskiy:
Motivation:

In version 4.2.x, Netty introduced new API
`SslContextBuilder.endpointIdentificationAlgorithm(String)` and flipped the default value to `HTTPS`.

Modifications:
- Detect presence of this new method and initialize it with the value from `config.hostnameVerificationAlgorithm()`.
- Verify `isOptionSupported` method works as expected before initializing its `MethodHandle`.

Result:

Hostname verification algorithm is enabled/disabled as expected for ST users with Netty 4.2.x in the classpath.


daschl (approved):

chrisvest (approved):
The `SslUtils.newClientSslHandler` will still set the endpoint verification algorithm, which shouldn't be a problem. We're just doing the work to configure it twice.

idelpivnitskiy:
Correct, I left that intentionally as a fallback for Netty 4.1.x

PR #2988, opened 2024-06-26
`StreamingConnectionFactory` should set hostnameVerificationAlgorithm to an empty string

idelpivnitskiy:
Motivation:

There are cases when default algorithm `HTTPS` can be reset if there are no `sniHostname` and no `peerHost`. See
`GrpcSslAndNonSslConnectionsTest` as a reproducer. This flow works well with Netty 4.1, but breaks with Netty 4.2 because Netty uses `HTTPS` by default. If we set it back to `null`, `sun.security.ssl.SSLEngineImpl` ignores it. To disable it later, we should use an empty string.

Modifications:

- Update `StreamingConnectionFactory.withSslConfigPeerHost` to use an empty string instead of `null` when it needs to reset `hostnameVerificationAlgorithm`;

Result:

We can override the previously set endpoint identification algorithm.


idelpivnitskiy:
I will pull this in to unblock testing with 4.2 snapshots

PR #2989, opened 2024-06-26
Add `Publisher.fromInputStream(InputStream, ByteArrayMapper)`

idelpivnitskiy:
Motivation:

Existing `Publisher.fromInputStream(InputStream)` contract has a side effect of an extra allocation in case the pre-allocated byte-array was not completely full of data.

Modifications:

- Add `Publisher.fromInputStream(InputStream, ByteArrayMapper)` that allows users to decide how to use the buffer region with data;
- Deprecate pre-existing `Publisher.fromInputStream(InputStream)` and `Publisher.fromInputStream(InputStream, int)` overloads;
- Update all existing use-cases to use `BufferAllocator.wrap` as a method reference as a `ByteArrayMapper`;

Result:

No impact on throughput and latency for `BlockingStreamingHttpClient` requests with `InputStream` payload (8Kb, 16Kb, 24Kb), but significant reduction in memory allocations and number of GC runs per benchmark:

Allocations: ~2Gb/s -> ~1.4Gb/s
Young Collection GC Count: 29 -> 23
Alloc Outside TLABs: 7.33 GiB -> 5.04 GiB

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ByteArrayMapper.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ByteArrayMapper.java
@@ -31,7 +31,7 @@ public interface ByteArrayMapper<T> {
     /**
      * Maps a specified {@code byte[]} buffer region into a {@code T}.
      * <p>
-     * The mapper can operate only with the specified region of the {@code buffer}, which can be safely used without a
+     * The mapper can operate only within the specified region of the {@code buffer}, which can be safely used without a
      * need to copy data. Access to other parts of the buffer may lead to unexpected results and due care should be
      * taken to avoid leaking that data through the returned type {@code T}.
      *
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ByteArrayMapper.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ByteArrayMapper.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright © 2024 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.concurrent.api;
+
+import io.servicetalk.concurrent.api.FromInputStreamPublisher.ToByteArrayMapper;
+
+import static io.servicetalk.concurrent.api.FromInputStreamPublisher.DEFAULT_MAX_BUFFER_SIZE;
+import static io.servicetalk.concurrent.api.FromInputStreamPublisher.ToByteArrayMapper.DEFAULT_TO_BYTE_ARRAY_MAPPER;
+
+/**
+ * A mapper to transform {@code byte[]} buffer regions into a desired type {@code T}.
+ *
+ * @param <T> Type of the result of this mapper
+ */
+@FunctionalInterface
+public interface ByteArrayMapper<T> {
+
+    /**
+     * Maps a specified {@code byte[]} buffer region into a {@code T}.
+     * <p>
+     * The mapper can operate only with the specified region of the {@code buffer}. Access to other parts of the buffer
+     * may lead to unexpected results. The specified region can be safely used by the mapper without a need to copy
+     * data.
+     *
+     * @param buffer {@code byte[]} buffer with data
+     * @param offset the offset of the region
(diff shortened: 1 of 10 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ByteArrayMapper.java
```diff
@@ -0,0 +1,79 @@
...
+import static io.servicetalk.concurrent.api.FromInputStreamPublisher.DEFAULT_MAX_BUFFER_SIZE;
+import static io.servicetalk.concurrent.api.FromInputStreamPublisher.ToByteArrayMapper.DEFAULT_TO_BYTE_ARRAY_MAPPER;
+
+/**
+ * A mapper to transform {@code byte[]} buffer regions into a desired type {@code T}.
+ *
+ * @param <T> Type of the result of this mapper
+ */
+@FunctionalInterface
+public interface ByteArrayMapper<T> {
```
idelpivnitskiy, line 29: We need this interface bcz concurrent-api doesn't have a dependency on buffer-api module.
Eventually, I think it's worth it. We can consolidate configuration parameters for how to read data from `InputStream` as default methods on this interface instead of adding more and more overloads for `Publisher.fromInputStream`. `Publisher` is already quite a lengthy class.

bryce-anderson (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ByteArrayMapper.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Maps a specified {@code byte[]} buffer region into a {@code T}.
+     * <p>
+     * The mapper can operate only with the specified region of the {@code buffer}. Access to other parts of the buffer
+     * may lead to unexpected results. The specified region can be safely used by the mapper without a need to copy
+     * data.
```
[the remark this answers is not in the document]
idelpivnitskiy: It says "Access to other parts of the buffer may lead to unexpected results.". Do you have a recommendation on how to rephrase it to highlight what you suggested?

> It's pretty common for people to wrap with a ByteBuffer and set limits, but still allow access to the underlying array.

It's possible to extract the underlying array from the `HeapByteBuffer`. However, users typically respect offset and length based on ByteBuffer info available via getters. Otherwise, it's always a bug.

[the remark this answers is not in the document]
bryce-anderson: > It says "Access to other parts of the buffer may lead to unexpected results.". Do you have a recommendation on how to rephrase it to highlight what you suggested?

Maybe something to the tune of "Access to other parts of the buffer may lead to unexpected results and due care should be taken to avoid leaking that data through the result type T." ?

> It's possible to extract the underlying array from the HeapByteBuffer. However, users typically respect offset and length based on ByteBuffer info available via getters. Otherwise, it's always a bug.

The array can't be accessed if the HeapByteBuffer is read-only, but I agree that it's uncommon to do that. This was just an example to justify making the warning a bit more explicit that people should be careful to avoid leaking the buffer through their type T.

bryce-anderson (approved):

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ByteArrayMapper.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @param <T> Type of the result of this mapper
+ */
+@FunctionalInterface
+public interface ByteArrayMapper<T> {
+
+    /**
+     * Maps a specified {@code byte[]} buffer region into a {@code T}.
+     * <p>
+     * The mapper can operate only with the specified region of the {@code buffer}, which can be safely used without a
```
daschl, line 34: with/within?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/ByteArrayMapper.java
daschl: 👍

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
```diff
@@ -306,4 +291,32 @@ private boolean trySetTerminalSent() {
...
+        ToByteArrayMapper(final int maxBufferSize) {
+            this.maxBufferSize = ensurePositive(maxBufferSize, "maxBufferSize");
+        }
+
+        @Override
+        public byte[] map(final byte[] buffer, final int offset, final int length) {
+            if (offset == 0 && length == buffer.length) {
+                return buffer;
+            } else {
+                final byte[] partial = new byte[length];
```
daschl, line 311: since we are also allocating a new array like with the old code, how does that improve the GC performance in the default cases? (not the ones with a custom ByteArrayMapper)

idelpivnitskiy: This exists only for backward compatibility when users already depend on T as byte-array and need to migrate from deprecated methods. All internal use-cases were changed to use `BufferAllocator.wrap` that helps to avoid this path

daschl (approved):

PR #2995, opened 2024-07-03
Align Set-Cookie attribute names with RFC6265 syntax

daschl:
This changeset aligns the "Set-Cookie" property names with the RFC6265 in the sense that according to the spec (see https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#section-4.1.1) where property names are using camel-case syntax instead of all lower case.

NOTE: The name of an attribute-value pair is not case sensitive. So while they are presented here in CamelCase, such as "HttpOnly" or "SameSite", any case is accepted. E.x.: "httponly", "Httponly", "hTTPoNLY", etc.

This means that this change is purely cosmetic and helps with consistency with other implementations, i.e. Netty.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
@@ -34,13 +34,13 @@ import static java.lang.Integer.toHexString;
  * Default implementation of {@link HttpSetCookie}.
  */
 public final class DefaultHttpSetCookie implements HttpSetCookie {
-    private static final String ENCODED_LABEL_DOMAIN = "; domain=";
-    private static final String ENCODED_LABEL_PATH = "; path=";
-    private static final String ENCODED_LABEL_EXPIRES = "; expires=";
-    private static final String ENCODED_LABEL_MAX_AGE = "; max-age=";
-    private static final String ENCODED_LABEL_HTTP_ONLY = "; httponly";
-    private static final String ENCODED_LABEL_SECURE = "; secure";
-    private static final String ENCODED_LABEL_SAMESITE = "; samesite=";
+    private static final String ENCODED_LABEL_DOMAIN = "; Domain=";
+    private static final String ENCODED_LABEL_PATH = "; Path=";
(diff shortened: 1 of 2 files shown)


bryce-anderson (approved):

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpSetCookie.java
```diff
@@ -34,13 +34,13 @@
...
-    private static final String ENCODED_LABEL_EXPIRES = "; expires=";
-    private static final String ENCODED_LABEL_MAX_AGE = "; max-age=";
-    private static final String ENCODED_LABEL_HTTP_ONLY = "; httponly";
-    private static final String ENCODED_LABEL_SECURE = "; secure";
-    private static final String ENCODED_LABEL_SAMESITE = "; samesite=";
+    private static final String ENCODED_LABEL_DOMAIN = "; Domain=";
+    private static final String ENCODED_LABEL_PATH = "; Path=";
+    private static final String ENCODED_LABEL_EXPIRES = "; Expires=";
+    private static final String ENCODED_LABEL_MAX_AGE = "; Max-Age=";
+    private static final String ENCODED_LABEL_HTTP_ONLY = "; HttpOnly";
```
idelpivnitskiy, line 41: The spec doesn't have `MUST` in its wording and clarifies that attribute names are case insensitive:

> NOTE: The name of an attribute-value pair is not case sensitive. So while they are presented here in CamelCase, such as "HttpOnly" or "SameSite", any case is accepted. E.x.: "httponly", "Httponly", "hTTPoNLY", etc.

I don't mind either way. Netty uses CamelCase: https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieHeaderNames.java#L19-L33
We can align with spec and Netty.
Note that it has discrepancy for `HTTPOnly`.

idelpivnitskiy (approved):
Please update title, PR description and commit message to clarify that it's not a bug or misbehavior, we only align with spec formatting without impact on anyone. We don't want to mislead anyone in release notes.

PR #3002, opened 2024-07-12
Enhance clarifications for `ServiceDiscoverer` contract

idelpivnitskiy:
Motivation:

Some assumptions that are currently in place are not clarified in documentation for new implementations.

Modifications:

- Clarify that the returned publisher should never complete and must support re-subscribes to handle failures and cancellations.
- Clarify motivation for always starting with the "state of the world" for interpreting `ServiceDiscovererEvent`(s).
- Add recommendation to implement a `ServiceDiscoverer` using a single thread.

Result:

More details for expected behavior of a `ServiceDiscoverer` implementation.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
@@ -15,6 +15,7 @@
  */
 package io.servicetalk.client.api;
 
+import io.servicetalk.concurrent.PublisherSource.Subscriber;
 import io.servicetalk.concurrent.PublisherSource.Subscription;
 import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
 import io.servicetalk.concurrent.api.Publisher;
@@ -25,6 +26,10 @@ import java.util.Collection;
  * Represents the interaction pattern with a service discovery system. It is assumed that once {@link #discover(Object)}
  * is called that the service discovery system will push data updates or implementations of this interface will poll for
  * data updates. Changes in the available addresses will be communicated via the resulting {@link Publisher}.
+ * <p>
+ * Because typically a {@link ServiceDiscoverer} implementation runs in the background and doesn't require many compute
+ * resources, it's recommended (but not required) to run it and deliver updates on a single thread either for all
+ * discoveries or at least for all {@link Subscriber Subscribers} to the same {@link Publisher}.
  * <p>
  * See {@link ServiceDiscovererEvent} for documentation regarding the interpretation of events.
  *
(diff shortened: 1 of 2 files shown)


servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
```diff
@@ -39,8 +44,12 @@ public interface ServiceDiscoverer<UnresolvedAddress, ResolvedAddress,
      * {@code address}.
      * <p>
      * In general a call to this method will continue to discover changes related to {@code address} until the
-     * {@link Subscription} corresponding to the return value is cancelled via {@link Subscription#cancel()} or there
-     * are no more changes to be published.
+     * {@link Subscription Subscription} corresponding to the return value is {@link Subscription#cancel() cancelled} or
```
bryce-anderson, line 47:
```suggestion
     * {@link Subscription} corresponding to the return value is {@link Subscription#cancel() cancelled} or
```

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
```diff
@@ -26,6 +27,10 @@
  * is called that the service discovery system will push data updates or implementations of this interface will poll for
  * data updates. Changes in the available addresses will be communicated via the resulting {@link Publisher}.
  * <p>
+ * Because typically a {@link ServiceDiscoverer} implementation runs in the background and doesn't require many compute
+ * resources, it's recommended (but not required) to run it and deliver updates on a single thread either for all
+ * discoveries or at least for all {@link Subscriber Subscribers} to the same {@link Publisher}.
```
bryce-anderson, line 32: fwiw, it's not clear what this buys the implementors.

bryce-anderson (approved):

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
idelpivnitskiy: This is intentional bcz without an alias javadoc puts a full class name, which breaks readability of the javadoc

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
idelpivnitskiy: I don't want to put it as a requirement, it's just a nice-to-have recommendation bcz most ppl don't think about it when they implement.
For RRLB it will help with this case: https://github.com/apple/servicetalk/pull/3005#discussion_r1676212131
But we are moving away from RRLB anyway.

In general, it makes debugging easier bcz you can track all events from a single thread and logs will represent the exact serial behavior. If they use multiple thread, serial behavior is still guaranteed by RS, but logs may print events between Subscribers in any order. 

Do you have a recommendation how to rephrase?

bryce-anderson, line 32:
```suggestion
 * discoveries or at least for all {@link Subscriber Subscribers} to the same {@link Publisher}. This will make debugging
   easier as discovery events and the logs they generate will be less susceptible to reordering.
```

bryce-anderson: My best suggestion is basically what you said in the reply. 🙂 I added as a suggestion but I'm also not strongly opinionated about it. If you prefer to keep it short and not try to sell it then that's perfectly fine with me.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
bryce-anderson: Ah, I see.

servicetalk-client-api/src/main/java/io/servicetalk/client/api/ServiceDiscoverer.java
idelpivnitskiy: thanks, applied with minor rewording

bryce-anderson (approved):

tkountis (approved):

PR #3005, opened 2024-07-12
Avoid race b/w ServiceDiscoverer events after cancel and re-subscribe

idelpivnitskiy:
Motivation:

#2514 added an ability for `LoadBalancer` to cancel discovery stream and
re-subscribe in attempt to trigger a new resolution when its internal
state became `UNHEALTHY`. However, it did not account for a possibility
of new events racing with cancellation or a Publisher not stopping
updates immediately. As a result, it's possible to end up in a corrupted
state if the old `EventSubscriber` receives new events after a new
`EventSubscriber` received a new "state of the world".

Modifications:

- Track the current `EventSubscriber` and process events only if it's
current.
- Enhance `resubscribeToEventsWhenAllHostsAreUnhealthy()` to simulate
the described behavior.

Result:

New "state of the world" can not be corrupted by events emitted for the
old `EventSubscriber`.

---
Depends on #3004, review only the 2nd commit: https://github.com/apple/servicetalk/pull/3005/commits/d48d8f59053a567670f803b82ac6e23d4719e39f

### DIFF
--- a/servicetalk-loadbalancer-experimental/src/main/java/io/servicetalk/loadbalancer/DefaultLoadBalancer.java
+++ b/servicetalk-loadbalancer-experimental/src/main/java/io/servicetalk/loadbalancer/DefaultLoadBalancer.java
@@ -90,6 +90,8 @@ final class DefaultLoadBalancer<ResolvedAddress, C extends LoadBalancedConnectio
     private static final long RESUBSCRIBING = -1L;
 
     private volatile long nextResubscribeTime = RESUBSCRIBING;
+    @Nullable
+    private volatile EventSubscriber currentSubscriber;
 
     // writes are protected by `sequentialExecutor` but the field can be read by any thread.
     private volatile HostSelector<ResolvedAddress, C> hostSelector;
@@ -181,10 +183,13 @@ final class DefaultLoadBalancer<ResolvedAddress, C extends LoadBalancedConnectio
         // This method is invoked only when we are in RESUBSCRIBING state. Only one thread can own this state.
         assert nextResubscribeTime == RESUBSCRIBING;
         if (resubscribe) {
+            assert healthCheckConfig != null : "Resubscribe can happen only when health-checking is configured";
             LOGGER.debug("{}: resubscribing to the ServiceDiscoverer event publisher.", this);
             discoveryCancellable.cancelCurrent();
         }
-        toSource(eventPublisher).subscribe(new EventSubscriber(resubscribe));
+        final EventSubscriber eventSubscriber = new EventSubscriber(resubscribe);
+        this.currentSubscriber = eventSubscriber;
+        toSource(eventPublisher).subscribe(eventSubscriber);
         if (healthCheckConfig != null) {
             assert healthCheckConfig.executor instanceof NormalizedTimeSourceExecutor;
             nextResubscribeTime = nextResubscribeTime(healthCheckConfig, this);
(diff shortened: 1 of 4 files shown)


servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
```diff
@@ -290,6 +281,15 @@ public void onNext(@Nullable final Collection<? extends ServiceDiscovererEvent<R
                 LOGGER.debug("{}: unexpectedly received null instead of events.", RoundRobinLoadBalancer.this);
                 return;
             }
+            // According to Reactive Streams Rule 1.8
+            // (https://github.com/reactive-streams/reactive-streams-jvm?tab=readme-ov-file#1.8) new events will
+            // stop eventually but not guaranteed to stop immediately after cancellation or could race with cancel.
+            // Therefore, we should check that this is the current Subscriber before processing new events.
+            if (healthCheckConfig != null && currentSubscriber != this) {
```
idelpivnitskiy, line 288: Note that for the `RoundRobinLoadBalancer` there is still a possibility for a race if `ServiceDiscoverer` implementation doesn't follow the "use a single thread" recommendation: https://github.com/apple/servicetalk/pull/3002/files#diff-994a09bcaa716664549357bca8bfa8288ee0dd98b0280b139fd30b15cd69bd8eR30-R32

However, taking into account that:
1. we are close to switching to `DefaultLoadBalancer`;
2. the default DNS `ServiceDiscoverer` is single-threaded, and for other implementations it's a trivial fix;
3. possibility for that race is significantly reduced by the current check;

I don't see a reason to significantly overcomplicate this implementation unless someone has a simple suggestion.

Example of a race:
1. SD Thread#1 delivers events for old `EventSubscriber`, crosses this check and OS doesn't execute it further.
2. Caller Thread#2 triggers cancellation, resubscribes and flips the current subscriber.
3. SD Thread#3 delivers "state of the world" for a new `EventSubscriber`.
4. OS continues execution of Thread#1 and it corrupts the `usedHosts` state.

servicetalk-loadbalancer-experimental/src/main/java/io/servicetalk/loadbalancer/DefaultLoadBalancer.java
```diff
@@ -90,6 +90,8 @@ final class DefaultLoadBalancer<ResolvedAddress, C extends LoadBalancedConnectio
     private static final long RESUBSCRIBING = -1L;
 
     private volatile long nextResubscribeTime = RESUBSCRIBING;
+    @Nullable
```
bryce-anderson, line 93: In reality this is never null after initialization, much like every other field. Could we ignore it, or give it a non-null initial value that we just discard?

servicetalk-loadbalancer-experimental/src/main/java/io/servicetalk/loadbalancer/DefaultLoadBalancer.java
```diff
@@ -287,16 +278,27 @@ public void onSubscribe(final Subscription s) {
...
+            if (isClosed) {
+                // nothing to do if the load balancer is closed.
+                return;
+            }
+
+            // According to Reactive Streams Rule 1.8
+            // (https://github.com/reactive-streams/reactive-streams-jvm?tab=readme-ov-file#1.8) new events will
+            // stop eventually but not guaranteed to stop immediately after cancellation or could race with cancel.
+            // Therefore, we should check that this is the current Subscriber before processing new events.
+            if (healthCheckConfig != null && currentSubscriber != this) {
```
bryce-anderson, line 299: This is my only real concern with the PR.
```suggestion
            if (currentSubscriber != this) {
```
It seems to me that it shouldn't matter for what reason we switched our subscription: this one is simply no longer active so we must ignore updates or risk clobbering updates from the active subscription.
Same in RoundRobinLoadBalancer.

servicetalk-loadbalancer/src/main/java/io/servicetalk/loadbalancer/RoundRobinLoadBalancer.java
bryce-anderson: I tend to agree with your assessment that we need not address it: this race has been here for a long time and we've survived until now so a little longer wont hurt. However, if you felt like fixing it I think it could be done with a lock protecting all `EventSubscriber.onNext` calls.

bryce-anderson (commented):
Just one concern.

servicetalk-loadbalancer-experimental/src/main/java/io/servicetalk/loadbalancer/DefaultLoadBalancer.java
idelpivnitskiy: We only need to check for "current". So, it doesn't matter if it's nullable or not because we don't access any fields or methods of `currentSubscriber` -> NPE is not possible. 
To make it non-null I will have to add a `NoopSubscriber` in every implementation, which seems like a boilerplate for this use-case.

bryce-anderson: I was thinking you could give it an idle `new EventSubscriber()` during the declaration but I'll admit that has its own drawbacks so I suspect you're right that it's fine to just give it `@Nullable` for now.

bryce-anderson (approved):

PR #1949, opened 2021-11-10
Removing deprecated `ServiceDiscoveryRetryStrategy`

chemicL:
Motivation:

Recently, `ServiceDiscoveryRetryStrategy` was deprecated. For 0.42
release we intend to remove it.

Modifications:

`ServiceDiscoveryRetryStrategy` and
`DefaultServiceDiscoveryRetryStrategy` have been removed.

Result:

Only the standard retry mechanism is available for `ServiceDiscoverer`
errors in `SingleAddressHttpClientBuilder` and in
`PartitionedHttpClientBuilder`.

PR #3006, opened 2024-07-12
Incorrect state possible after retrying `ServiceDiscoverer` events

idelpivnitskiy:
Motivation:

Clients have a configurable `serviceDiscovererRetryStrategy` to guarantee a steady stream of events to the `LoadBalancer` that never fails. It's necessary at the client level to avoid hanging requests indefinitely and let requests observe failures from ServiceDiscoverer. Also, for `PartitionedHttpClient` it's necessary to guarantee that `GroupedPublisher` never fails.

Retry is effectively a re-subscribe. According to `ServiceDiscoverer` contract (clarified in #3002), each `Subscriber` receives a "state of the world" as the first collection of events. The problem is that the state may change significantly between retries. As a result, unavailable addresses can remain inside the `LoadBalancer` forever. Example:

T1. SD delivers [a,b]
T1. LB receives [a,b]
T1. SD delivers error
T2. SD info changed ("a" got revoked)
T3. Client retries SD
T3. SD delivers [b]
T3. LB receives [b] (but still holds "a")

When we retry `ServiceDiscoverer` errors, we should keep pushing deltas downstream or purge events that are not present in the new "state of the world".

We previously had this protection but it was mistakenly removed in #1949 as part of a broader refactoring around `ServiceDiscoverer` <-> `LoadBalancer` contract.

Modifications:

- Add `RetryingServiceDiscoverer` that handles retries and keeps the state between retries.
- Use it in `DefaultSingleAddressHttpClientBuilder` and `DefaultPartitionedHttpClientBuilder`.
- Use `CastedServiceDiscoverer` to allow modifications for `ServiceDiscovererEvent` after we started to use a wildcard type in #2379.
- Pass consistent `targetResource` identifier to both `RetryingServiceDiscoverer` and `LoadBalancerFactory` to allow state correlation when inspecting heap dump.

Result:

Client keeps pushing deltas to `LoadBalancer` after retrying `ServiceDiscoverer` errors, keeping its state consistent with `ServiceDiscoverer`.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
@@ -19,6 +19,7 @@ import io.servicetalk.buffer.api.BufferAllocator;
 import io.servicetalk.buffer.api.CharSequences;
 import io.servicetalk.client.api.ConnectionFactory;
 import io.servicetalk.client.api.ConnectionFactoryFilter;
+import io.servicetalk.client.api.DefaultServiceDiscovererEvent;
 import io.servicetalk.client.api.DelegatingServiceDiscoverer;
 import io.servicetalk.client.api.LoadBalancer;
 import io.servicetalk.client.api.ServiceDiscoverer;
@@ -66,17 +67,17 @@ import org.slf4j.LoggerFactory;
 
 import java.net.InetSocketAddress;
 import java.net.SocketOption;
-import java.time.Duration;
 import java.util.Collection;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BooleanSupplier;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import javax.annotation.Nullable;
 
 import static io.netty.util.NetUtil.toSocketAddressString;
+import static io.servicetalk.client.api.ServiceDiscovererEvent.Status.UNAVAILABLE;
 import static io.servicetalk.concurrent.api.AsyncCloseables.newCompositeCloseable;
 import static io.servicetalk.concurrent.api.Processors.newCompletableProcessor;
-import static io.servicetalk.concurrent.api.RetryStrategies.retryWithExponentialBackoffFullJitter;
 import static io.servicetalk.http.api.HttpExecutionStrategies.defaultStrategy;
 import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
 import static io.servicetalk.http.api.HttpProtocolVersion.HTTP_1_1;
@@ -86,7 +87,6 @@ import static io.servicetalk.http.netty.StrategyInfluencerAwareConversions.toCon
 import static io.servicetalk.http.netty.StrategyInfluencerAwareConversions.toConditionalConnectionFilterFactory;
 import static java.lang.Integer.parseInt;
 import static java.time.Duration.ofMinutes;
-import static java.time.Duration.ofSeconds;
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -105,9 +105,7 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
             new RetryingHttpRequesterFilter.Builder().build();
     private static final StreamingHttpConnectionFilterFactory DEFAULT_IDLE_TIMEOUT_FILTER =
             new IdleTimeoutConnectionFilter(ofMinutes(5));
-
-    static final Duration SD_RETRY_STRATEGY_INIT_DURATION = ofSeconds(2);
-    static final Duration SD_RETRY_STRATEGY_MAX_DELAY = ofSeconds(128);
+    private static final AtomicInteger CLIENT_ID = new AtomicInteger();
 
     private final U address;
     @Nullable
@@ -116,7 +114,7 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
     final HttpExecutionContextBuilder executionContextBuilder;
     private final ClientStrategyInfluencerChainBuilder strategyComputation;
     private HttpLoadBalancerFactory<R> loadBalancerFactory;
-    private ServiceDiscoverer<U, R, ? extends ServiceDiscovererEvent<R>> serviceDiscoverer;
+    private ServiceDiscoverer<U, R, ServiceDiscovererEvent<R>> serviceDiscoverer;
     private Function<U, CharSequence> hostToCharSequenceFunction =
             DefaultSingleAddressHttpClientBuilder::toAuthorityForm;
     private boolean addHostHeaderFallbackFilter = true;
@@ -142,8 +140,7 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
         executionContextBuilder = new HttpExecutionContextBuilder();
         strategyComputation = new ClientStrategyInfluencerChainBuilder();
         this.loadBalancerFactory = defaultLoadBalancer();
-        this.serviceDiscoverer = requireNonNull(serviceDiscoverer);
-
+        this.serviceDiscoverer = new CastedServiceDiscoverer<>(serviceDiscoverer);
         clientFilterFactory = appendFilter(clientFilterFactory, HttpMessageDiscardWatchdogClientFilter.CLIENT_CLEANER);
     }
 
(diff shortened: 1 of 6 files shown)


idelpivnitskiy:
Marking it as "draft" because I still need to implement tests for `RetryingServiceDiscoverer`, but want to get earlier feedback for the main part before I work on tests.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingServiceDiscoverer.java
```diff
@@ -0,0 +1,149 @@
...
+    public Publisher<Collection<E>> discover(final U address) {
+        // The returned publisher is guaranteed to never fail because we retry all errors here. However, LoadBalancer
+        // can still cancel and re-subscribe in attempt to recover from unhealthy state. In this case, we need to
+        // re-initialize the ServiceDiscovererEventsCache and restart from an empty state.
+        return Publisher.defer(() -> {
+            final ServiceDiscovererEventsCache<R, E> eventsCache =
+                    new ServiceDiscovererEventsCache<>(targetResource, makeUnavailable);
+            return delegate().discover(address)
+                    .map(eventsCache::consumeAndFilter)
+                    .beforeOnError(eventsCache::errorSeen)
```
idelpivnitskiy, line 79: Alternative approach could be to use `scanWithMapper` or `liftSync` instead of (`defer` + `map` + `beforeOnError`). I picked this one for simplicity bcz internals of `scanWithMapper` look a bit complicated for this task and `liftSync` is too low level API.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingServiceDiscoverer.java
```diff
@@ -0,0 +1,149 @@
...
+            LOGGER.debug("{} observed an error from ServiceDiscoverer", targetResource, t);
+        }
+
+        Collection<E> consumeAndFilter(final Collection<E> events) {
+            if (retainedState == NONE_RETAINED) {
+                for (E event : events) {
+                    if (UNAVAILABLE.equals(event.status())) {
+                        currentState.remove(event.address());
+                    } else {
+                        currentState.put(event.address(), event);
```
idelpivnitskiy, line 113: Note that even though this will retain only the last event for the `address`, it should not matter because the `retainedState` will be used only to propagate `UNAVAILABLE` state that will simply remove it from LB. Any other associated meta-data (if any) doesn't matter for `UNAVAILABLE` state.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingServiceDiscoverer.java
```diff
@@ -0,0 +1,149 @@
...
+            assert currentState.isEmpty();
+            final List<E> toReturn = new ArrayList<>(events.size() + retainedState.size());
+            for (E event : events) {
+                final R address = event.address();
+                toReturn.add(event);
+                retainedState.remove(address);
+                if (!UNAVAILABLE.equals(event.status())) {
+                    currentState.put(address, event);
+                }
+            }
```
bryce-anderson, line 133: I think we want to avoid emitting events that are already known state. In theory we know the status isn't unavailable so we just need to compare the status and metadata but we don't actually know about metadata. Since the events are coming from the same source simple equality should work to capture metadata on a refined `ServiceDiscovererEvent`.
```suggestion
            for (E event : events) {
                final R address = event.address();
                if (UNAVAILABLE.equals(event.status())) {
                    LOGGER.warn("When recovering from ServiceDiscoverer error an UNAVAILABLE event was observed " +
                            "which may indicate that the first event was not a state-of-the-world event and " +
                            "therefore is a ServiceDiscoverer bug.");
                    continue;
                }
                currentState.put(address, event);
                E oldEvent = retainedState.remove(address);
                if (!oldEvent.equals(event)) {
                    toReturn.add(event);
                }
            }
```

bryce-anderson (commented):
I only got partially through this and will continue in another session.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/RetryingServiceDiscovererTest.java
```diff
@@ -0,0 +1,304 @@
...
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class RetryingServiceDiscovererTest {
```
idelpivnitskiy, line 61: Recovered all tests from https://github.com/apple/servicetalk/pull/1949/files#diff-17f80381993b8c2b74a8b78674c065228c8ef16a9320fe23a11abffdfe629eb9 and enhanced those to cover more edge cases

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingServiceDiscoverer.java
idelpivnitskiy: The `RetryingServiceDiscoverer` should not alter the original flow of events from `ServiceDiscoverer`. We only need to purge addresses after retry to enforce new "state of the world".
I've added tests that validate duplicate events are propagated as-is and clarified that in the comment above this section.

bryce-anderson: I think based on the contract of `ServiceDiscovererEvent` you're technically right, it is legal to do it, but it will seem _odd_ to the consumer (the LB) to receive a no-op message. That said it would mean we'd need to filter out empty results since perhaps the most common result would be the no-op result where nothing changed.

idelpivnitskiy: It won't receive a noop, it will always receive a new SOTW. See `RetryingServiceDiscovererTest.sameAddressPostRetry()`: https://github.com/apple/servicetalk/pull/3006/files#diff-43e3e0456e8b6d6ddfeeedc106c79ec7ab5a5c2cb13eb325ee91e2cd257ee4d8R104-R122

We don't know if the event changed any meta-data or not. So, after re-subscribe we have to propagate the state as-is. We only prune what is not present in the new state.

bryce-anderson (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -839,4 +823,47 @@ private static <ResolvedAddress> HttpLoadBalancerFactory<ResolvedAddress> defaul
                 RoundRobinLoadBalancers.<ResolvedAddress, FilterableStreamingHttpLoadBalancedConnection>builder(
                                 DefaultHttpLoadBalancerFactory.class.getSimpleName()).build());
     }
+
+    // Because of the change in https://github.com/apple/servicetalk/pull/2379, we should constrain the type back to
+    // ServiceDiscovererEvent without "? extends" to allow RetryingServiceDiscoverer to mark events as UNAVAILABLE.
```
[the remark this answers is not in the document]
idelpivnitskiy: It's not about bridging the type between SD and LB, it's about a possibility to pass a custom SD with custom event subtype to `HttpClients`. Without this, it won't be possible to use those implementations, they will have to change to use `ServiceDiscovererEvent` instead.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingServiceDiscoverer.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.Map;
+import java.util.function.UnaryOperator;
+import javax.annotation.Nullable;
+
+import static io.servicetalk.client.api.ServiceDiscovererEvent.Status.UNAVAILABLE;
+import static io.servicetalk.concurrent.api.RetryStrategies.retryWithExponentialBackoffFullJitter;
+import static java.time.Duration.ofSeconds;
+import static java.util.Collections.emptyMap;
+
+final class RetryingServiceDiscoverer<U, R, E extends ServiceDiscovererEvent<R>>
```
tkountis, line 44: should this more broadly accessible? and if-so, consider adding `shareContextOnSub` inside the defer.

tkountis (approved):
lgtm

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/RetryingServiceDiscoverer.java
idelpivnitskiy: no, I don't have a use-case in mind where we need to share this "filter". 
for now, I will keep it without `shareContextOnSubscribe()` because subscribe can be driven from any request and I don't think the background discovery context needs to be shared with that.

PR #3010, opened 2024-07-16
Add HTTP file serving examples

Scottmitch:
Motivation:
There are no examples demonstrating serving response payload body from file.

### DIFF
--- a/servicetalk-examples/http/files/src/main/java/io/servicetalk/examples/http/files/FilesServer.java
+++ b/servicetalk-examples/http/files/src/main/java/io/servicetalk/examples/http/files/FilesServer.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright © 2024 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.examples.http.files;
+
+import io.servicetalk.buffer.api.BufferAllocator;
+import io.servicetalk.http.api.StreamingHttpResponse;
+import io.servicetalk.http.netty.HttpServers;
+import io.servicetalk.transport.api.IoExecutor;
+
+import java.io.InputStream;
+
+import static io.servicetalk.concurrent.api.Publisher.from;
+import static io.servicetalk.concurrent.api.Publisher.fromInputStream;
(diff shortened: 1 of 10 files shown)


bryce-anderson (approved):

servicetalk-examples/http/files/src/main/java/io/servicetalk/examples/http/files/FilesServer.java
```diff
@@ -0,0 +1,56 @@
...
+        final HttpStreamingSerializer<byte[]> byteSerializer = bytesStreamingSerializer(requestHeaders ->
+                requestHeaders.set(CONTENT_TYPE, TEXT_PLAIN_UTF_8));
+        final Publisher<byte[]> notFoundResponse = from("file not found, please rebuild the project".getBytes(UTF_8));
+        HttpServers.forPort(8080)
+                .listenStreamingAndAwait((ctx, request, responseFactory) -> {
+                    // InputStream lifetime ownership is transferred to ServiceTalk (e.g. it will call close) because
+                    // we create a new InputStream per request and always pass it to ServiceTalk as the response payload
+                    // body (if not null).
+                    final InputStream responseStream = FilesServer.class.getClassLoader()
+                            .getResourceAsStream("response_payload.txt");
```
idelpivnitskiy, line 49: `InputStream` is a good option for `BlockingStreamingHttpService`. For async `StreamingHttpService` it's preferable to use NIO `FileChannel` via `RandomAccessFile`. For large files `file.getChannel().map(...)` may throw. In that case, we need an Iterable (or Publisher) of mapped `ByteBuffer`(s) for regions of the file.

idelpivnitskiy: It's also useful for users to clarify the purpose of this example, because there could be different strategies for different use-cases. For example, large files can be served via mapped `ByteBuffer`(s), while static web server may prefer to load all files into memory (read only direct buffers) to serve them faster (avoid access to disk).

servicetalk-examples/http/files/src/main/java/io/servicetalk/examples/http/files/FilesServer.java
```diff
@@ -0,0 +1,56 @@
...
+                    return succeeded(responseStream == null ?
+                                responseFactory.notFound().payloadBody(notFoundResponse, byteSerializer) :
+                                responseFactory.ok().payloadBody(fromInputStream(responseStream), byteSerializer));
```
idelpivnitskiy, line 52: There is no need to go through `byteSerializer` anymore after #2989. It's enough to pass `BufferAllocator::wrap` as a method reference to safe on allocations of intermediate byte-arrays.

servicetalk-examples/http/files/src/main/java/io/servicetalk/examples/http/files/FilesServer.java
Scottmitch: discussed offline. FileChannel approach generally still pull data into memory if using `trasnferTo` or related APIs. For larger files there maybe some benefit here but lets enhance the example in the future for these use cases as necessary.

bryce-anderson (approved):

PR #3018, opened 2024-07-23
Disable `fail-fast` on GitHub Actions for snapshot and release pipelines

idelpivnitskiy:
Motivation:

We always use 2 pipelines for snapshots and releases (JDK8 and JDK11) because those build and publish different set of modules. Because they are independent, we should not fail another one if any of them fail.

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -9,6 +9,7 @@ jobs:
     name: Release JDK ${{ matrix.java }}
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         java: [ 8, 11 ]
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-release.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     name: Release JDK ${{ matrix.java }}
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
daschl, line 12: for snapshots I see the point, would there be any negative side effects we release one but not the other potentially?

idelpivnitskiy: no, the set of produced modules is completely independent and because of the occasional flaky tests it affects the other pipeline

daschl (approved):

PR #3031, opened 2024-08-05
capacity-limiter-api: polish javadocs and minor code modifications

daschl:
This changeset mainly polishes javadocs as well as cleans up the implementations slightly without changing their behavior.


tkountis (approved):

daschl:
> These seem fine to me. Feel free to ignore: while you're here would you like to change the documentation for the meaning of the return values, specifically that the hints have meaning if they're non-negative? The only place I see it used in non-tests is [here](https://github.com/apple/servicetalk/blob/main/servicetalk-capacity-limiter-api/src/main/java/io/servicetalk/capacity/limiter/api/CompositeCapacityLimiter.java#L77-L88) where only negative values are ignored (eg, 0 has meaning).

Will do this as a follow-up.

PR #3036, opened 2024-08-08
`SpliceFlatStreamToMetaSingle`: propagate cancel when races with data

idelpivnitskiy:
Motivation:

Cancellation of the single may race with delivering the first element. If `cancel` wins, we already deliver `CancellationException` if someone subscribes to the payload publisher. However, if `cancel` arrives after  `dataSubscriber.onSuccess(...)` and before someone subscribes to the payload publisher, it's a strong indicator that there was a race and `dataSubscriber` is not interested in it anymore. Therefore, we should pessimistically assume that nobody will ever subscribe to the payload publisher, cancel upstream subscription, and deliver `CancellationException` is someone still subscribes.

Modifications:
- Cancel upstream subscription for both `null` and `PENDING` states if received data cancel;
- Upwrap code around `parent.packer.apply` to assert that this path happens only if `CANCELED`;
- Deliver `CancellationException` is someone still subscribes to the payload publisher after `CANCELED`;
- Fix another bug: do not deliver `dataSubscriber.onError` if `metaSeenInOnNext == true` because it's already terminated (`CANCELED` does not play any role here);
- Log more appropriate message if upstream delivers a terminal event after `payloadSubscriber` is already terminated (`EMPTY_COMPLETED_DELIVERED`);
- Remove duplication around "Duplicate Subscribers are not allowed";

Result:

Network resources receive `cancel` signal and clean up connection state.

---
Note that the original version of this operator had this behavior, but it was changed in https://github.com/apple/servicetalk/commit/38e2c67ff19f67e3e9730d56e379e726e1981aa6
We don't have a full PR for that commit with review comments. Based on the commit description I believe it was fixing a different problem and the change in cancel propagation behavior for `PENDING` state is likely a side-effect of introducing a new state.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/SpliceFlatStreamToMetaSingle.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/SpliceFlatStreamToMetaSingle.java
@@ -202,14 +202,16 @@ final class SpliceFlatStreamToMetaSingle<Data, MetaData, Payload> implements Pub
                     if (maybePayloadSubUpdater.compareAndSet(this, null, PENDING)) {
                         payload = newPayloadPublisher();
                     } else {
-                        assert maybePayloadSub == CANCELED;
-                        maybePayloadSubUpdater.compareAndSet(this, CANCELED, EMPTY_COMPLETED_DELIVERED);
+                        final Object maybePayloadSub = this.maybePayloadSub;
+                        assert maybePayloadSub == CANCELED : "Expected CANCELED but got: " + maybePayloadSub;
+                        boolean cas = maybePayloadSubUpdater.compareAndSet(this, CANCELED, EMPTY_COMPLETED_DELIVERED);
+                        assert cas : "Could not transition from CANCELED to EMPTY_COMPLETED_DELIVERED";
                         payload = Publisher.failed(newCancellationException());
                     }
                     data = parent.packer.apply(meta, payload);
                     assert data != null : "Packer function must return non-null Data";
                 } catch (Throwable t) {
-                    assert rawSubscription != null;
+                    assert rawSubscription != null : "Expected rawSubscription but got null";
                     // We know that there is nothing else that can happen on this stream as we are not sending the
                     // data to the dataSubscriber.
                     rawSubscription.cancel();
@@ -234,7 +236,7 @@ final class SpliceFlatStreamToMetaSingle<Data, MetaData, Payload> implements Pub
                     // Subscriber which is not allowed by the Reactive Streams specification.
                     newSubscriber.onSubscribe(delayedSubscription);
                     if (maybePayloadSubUpdater.compareAndSet(SplicingSubscriber.this, PENDING, newSubscriber)) {
-                        assert rawSubscription != null;
+                        assert rawSubscription != null : "Expected rawSubscription but got null";
                         delayedSubscription.delayedSubscription(rawSubscription);
                     } else {
                         // Entering this branch means either a duplicate subscriber or a stream that completed or failed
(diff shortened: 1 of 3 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/JavaNetSoTimeoutHttpConnectionFilterTest.java
```diff
@@ -67,6 +68,7 @@
 import static org.hamcrest.Matchers.startsWith;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
+@Disabled   // FIXME: remove before merging
```
idelpivnitskiy, line 71: This PR is blocked by awaiting a fix for `Single.ambWith(...)` operator

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/SpliceFlatStreamToMetaSingle.java
```diff
@@ -156,7 +156,9 @@ private SplicingSubscriber(SpliceFlatStreamToMetaSingle<Data, MetaData, Payload>
          * Publisher}&lt;{@link Payload}&gt;
          */
         private void cancelData(Subscription subscription) {
-            if (maybePayloadSubUpdater.compareAndSet(this, null, CANCELED)) {
+            final Object current = maybePayloadSubUpdater.getAndUpdate(this,
+                    curr -> curr == null || curr == PENDING ? CANCELED : curr);
+            if (current == null || current == PENDING) {
```
idelpivnitskiy, line 161: Please carefully review the updated state machine and existing tests that we have. If you see a use-case not covered by the tests, will appreciate new ideas.

idelpivnitskiy (commented):
Ready for review, but will be merged only after `Single.ambWith(...)` fix

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/SpliceFlatStreamToMetaSingle.java
```diff
@@ -196,9 +198,15 @@ public void onNext(@Nullable Object obj) {
...
-                    data = parent.packer.apply(meta, maybePayloadSubUpdater.compareAndSet(this, null, PENDING) ?
-                            newPayloadPublisher() : Publisher.failed(StacklessCancellationException.newInstance(
-                                    "Canceled prematurely from Data", SplicingSubscriber.class, "cancelData(..)")));
+                    final Publisher<Payload> payload;
+                    if (maybePayloadSubUpdater.compareAndSet(this, null, PENDING)) {
+                        payload = newPayloadPublisher();
+                    } else {
+                        assert maybePayloadSub == CANCELED;
+                        maybePayloadSubUpdater.compareAndSet(this, CANCELED, EMPTY_COMPLETED_DELIVERED);
+                        payload = Publisher.failed(newCancellationException());
```
bryce-anderson, line 207: I believe this `cas` should never fail: can we throw an AssertionError or similar if it does?

bryce-anderson (approved):
Looks good but it's complex code so I'd recommend getting one more review before merging.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/SpliceFlatStreamToMetaSingle.java
idelpivnitskiy: good idea, it should never fail

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/SpliceFlatStreamToMetaSingle.java
```diff
@@ -290,26 +305,37 @@ public void onComplete() {
...
             }
         }
+
+        private void terminateWithIllegalStateException(PublisherSource.Subscriber<Payload> subscriber) {
+            subscriber.onError(new IllegalStateException("Duplicate Subscribers are not allowed. Existing: " +
+                    subscriber + ", failed the race with a duplicate, but neither has seen onNext()"));
+        }
+
+        private static StacklessCancellationException newCancellationException() {
+            return StacklessCancellationException.newInstance(
```
Scottmitch, line 337: is this expected to be common enough where it needs to be stackless? if not, its called from 2 different locations so we should differentiate these two cases.

Scottmitch (approved):
lgtm after ignored test is brought back to working

PR #3039, opened 2024-08-09
Improve `CancellationException`

idelpivnitskiy:
Motivation:

1. `H2ClientParentConnectionContext.StacklessCancellationException` was reused by `SpliceFlatStreamToMetaSingle`, which created a false impression in logs that this exception is related to HTTP/2.
2. We should capture the Subscriber's stack-trace for `CancellationException` in `BeforeFinallyHttpOperator` instead of a netty stack-trace, it will help to debug paths that subscribe to the payload post cancel.

Modifications:

1. Make `H2ClientParentConnectionContext.StacklessCancellationException` a top-level class.
2. Wrap `CancellationException` with `Publisher.defer(...)` in `BeforeFinallyHttpOperator`.
3. Add an exception message in `DefaultBlockingIterableProcessor`.

Result:

Easier for users to reason about received `CancellationException`.


servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/StacklessCancellationException.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.utils;
+
+import io.servicetalk.concurrent.internal.ThrowableUtils;
+
+import java.util.concurrent.CancellationException;
+
+final class StacklessCancellationException extends CancellationException {
```
idelpivnitskiy, line 22: I was not sure if it's worth making it public in a module like `concurrent-internal` and created a copy for now. Can move if someone prefers.

daschl: Feels a bit confusing to have the same exception in two different places, especially since http-netty depends on utils?

idelpivnitskiy: Actually, I changed my mind. It's worth capturing the full stack-trace if we capture the Subscriber's stack-trace instead of the netty transport stack-trace. Updated PR

Scottmitch (approved):

daschl (approved):

bryce-anderson (approved):

PR #3040, opened 2024-08-12
`Amb` operators for `Single` and `Completable` should respect reactive contract 2.3

daschl:
Motivation
==========
Reactive Streams rule 2.3 mandates, that `Subscriber.onComplete() and Subscriber.onError(Throwable t) MUST NOT call any methods on the Subscription or the Publisher`.

Our current amb operator implementation also performs a cancellation on the Single that just completed, which violates the rule.

Modifications
=============
This changeset modifies the amb operator in a way that all but the one Single which got a termination signal (through `onNext` or `onError`) will get the cancellation propagated.

The test suite is enhanced to cover this scenario, both for `ambWith` as well as the static `amb` factory methods. Since the `Completable` also converts to `Single` underneath, the functionality is transitively available as well.

Result
======
The `amb*` operator variants adhere to reactive streams 2.3 rule.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
@@ -78,7 +78,6 @@ final class AmbSingles<T> extends Single<T> {
 
     static final class AmbSubscriber<T> extends DelayedCancellable implements Subscriber<T> {
         private final State<T> state;
-        private volatile boolean propagateCancel = true;
 
         AmbSubscriber(final State<T> state) {
             this.state = state;
@@ -91,21 +90,14 @@ final class AmbSingles<T> extends Single<T> {
 
         @Override
         public void onSuccess(@Nullable final T result) {
-            propagateCancel = false;
+            ignoreCancel();
             state.trySuccess(result);
         }
 
         @Override
         public void onError(final Throwable t) {
-            propagateCancel = false;
+            ignoreCancel();
             state.tryError(t);
-        }
-
-        @Override
-        public void cancel() {
-            if (propagateCancel) {
-                super.cancel();
-            }
         }
     }
 
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
@@ -78,6 +78,7 @@ final class AmbSingles<T> extends Single<T> {
 
     static final class AmbSubscriber<T> extends DelayedCancellable implements Subscriber<T> {
         private final State<T> state;
+        private volatile boolean propagateCancel = true;
 
         AmbSubscriber(final State<T> state) {
             this.state = state;
@@ -90,12 +91,21 @@ final class AmbSingles<T> extends Single<T> {
 
         @Override
         public void onSuccess(@Nullable final T result) {
+            propagateCancel = false;
             state.trySuccess(result);
         }
 
         @Override
         public void onError(final Throwable t) {
+            propagateCancel = false;
             state.tryError(t);
+        }
+
+        @Override
+        public void cancel() {
+            if (propagateCancel) {
+                super.cancel();
+            }
         }
     }
 
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/SingleAmbTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/SingleAmbTest.java
@@ -15,12 +15,16 @@
  */
 package io.servicetalk.concurrent.api;
 
+import io.servicetalk.concurrent.Cancellable;
 import io.servicetalk.concurrent.test.internal.TestSingleSubscriber;
 
-import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
 import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.ValueSource;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.function.BiFunction;
 
 import static io.servicetalk.concurrent.api.Single.amb;
@@ -183,24 +187,42 @@ class SingleAmbTest {
         first.onSuccess(2);
     }
 
-    @Test
-    void doNotCancelCompletedWithMoreSingles() {
-        final TestSingle<Integer> first = new TestSingle<>();
-        final TestSingle<Integer> second = new TestSingle<>();
-        final TestSingle<Integer> third = new TestSingle<>();
-        final TestSingle<Integer> fourth = new TestSingle<>();
+    @ParameterizedTest(name = "{displayName} [{index}] index={0}")
+    @ValueSource(ints = {0, 1, 2, 3})
+    void doNotCancelCompletedWithMoreSingles(final int index) {
+        final List<TestSingle<Integer>> singles = Arrays.asList(new TestSingle<>(), new TestSingle<>(),
+                new TestSingle<>(), new TestSingle<>());
 
         final TestSingleSubscriber<Integer> subscriber = new TestSingleSubscriber<>();
-        toSource(amb(first, second, third, fourth)).subscribe(subscriber);
+        toSource(amb(singles.get(0), singles.get(1), singles.get(2), singles.get(3))).subscribe(subscriber);
         subscriber.awaitSubscription();
 
-        third.onSuccess(1);
+        final TestSingle<Integer> terminates = singles.get(index);
+        terminates.onSuccess(1);
         assertThat("Unexpected result.", subscriber.awaitOnSuccess(), is(1));
 
-        verifyNotCancelled(third);
+        verifyNotCancelled(terminates);
+        for (int i = 0; i < singles.size(); i++) {
+            if (i != index) {
+                verifyCancelled(singles.get(i));
+            }
+        }
+    }
+
+    @ParameterizedTest(name = "{displayName} [{index}] error={0}, completeFirst={1}")
+    @CsvSource({"true,true", "true,false", "false,true", "false,false"})
+    void receiveTerminationSignalAfterCancel(final boolean error, final boolean completeFirst) {
+        toSource(amb(first, second)).subscribe(subscriber);
+        final Cancellable subscription = subscriber.awaitSubscription();
+        subscription.cancel();
+
         verifyCancelled(first);
         verifyCancelled(second);
-        verifyCancelled(fourth);
+        if (error) {
+            sendErrorToAndVerify(completeFirst ? first : second);
+        } else {
+            sendSuccessToAndVerify(completeFirst ? first : second);
+        }
     }
 
     private void sendSuccessToAndVerify(final TestSingle<Integer> source) {
@@ -213,15 +235,15 @@ class SingleAmbTest {
         assertThat("Unexpected error result.", subscriber.awaitOnError(), is(sameInstance(DELIBERATE_EXCEPTION)));
     }
 
-    private void verifyNotCancelled(final TestSingle<Integer> single) {
+    private static void verifyNotCancelled(final TestSingle<Integer> single) {
         final TestCancellable cancellable = new TestCancellable();
         single.onSubscribe(cancellable);
-        assertThat("Other source not cancelled.", cancellable.isCancelled(), is(false));
+        assertThat("Single cancelled when no cancellation was expected.", cancellable.isCancelled(), is(false));
     }
 
-    private void verifyCancelled(final TestSingle<Integer> single) {
+    private static void verifyCancelled(final TestSingle<Integer> single) {
         final TestCancellable cancellable = new TestCancellable();
         single.onSubscribe(cancellable);
-        assertThat("Other source not cancelled.", cancellable.isCancelled(), is(true));
+        assertThat("Single not cancelled, but cancellation was expected.", cancellable.isCancelled(), is(true));
     }
 }
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/SingleAmbTest.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/SingleAmbTest.java
@@ -17,6 +17,7 @@ package io.servicetalk.concurrent.api;
 
 import io.servicetalk.concurrent.test.internal.TestSingleSubscriber;
 
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
 
@@ -35,7 +36,6 @@ class SingleAmbTest {
     private final TestSingle<Integer> first = new TestSingle<>();
     private final TestSingle<Integer> second = new TestSingle<>();
     private final TestSingleSubscriber<Integer> subscriber = new TestSingleSubscriber<>();
-    private final TestCancellable cancellable = new TestCancellable();
 
     private enum AmbParam {
         AMB_WITH {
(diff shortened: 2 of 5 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
```diff
@@ -90,13 +91,22 @@ public void onSubscribe(final Cancellable cancellable) {
...
 
         @Override
         public void onError(final Throwable t) {
+            propagateCancel = false;
             state.tryError(t);
         }
+
+        @Override
+        public void cancel() {
+            if (propagateCancel) {
```
daschl, line 106: @idelpivnitskiy did I understand you correctly that it needs to be both ways, not just the direction it is done here?

Scottmitch: should we make the state around `propagateCancel` atomic via AtomicReferenceUpdater? otherwise will this be racy (unless we assume synchronization/sequencing is done externally)?

Scottmitch: Cancellation after terminal Subscriber method is OK.

https://github.com/reactive-streams/reactive-streams-jvm?tab=readme-ov-file#1.6
> If a Publisher signals either onError or onComplete on a Subscriber, that Subscriber’s Subscription MUST be considered cancelled.

https://github.com/reactive-streams/reactive-streams-jvm?tab=readme-ov-file#3.5
> Subscription.cancel MUST respect the responsivity of its caller by returning in a timely manner, MUST be idempotent and MUST be [thread-safe](https://github.com/apple/servicetalk/pull/3040#term_thread-safe).

idelpivnitskiy: @Scottmitch the way I read the posted rules doesn't mean that it's ok to explicitly call `cancel()` after terminal, it only says that subscription must be **considered** cancelled. 

The rule 2.3 clarifies it clearly, that we should not call `cancel()` after we got terminal.

@daschl yes, protection as needed both ways, the say Scott described in his first message. I will look around if we can reuse any pre-existing atomic.

Scottmitch: the "idempotent" from 3.5 allows for multiple calls to cancel. 3.7 also talks to this but is superseded by 3.5.

2.3 says that after the Subscriber is terminated via onComplete/onError, that it shouldn't call Subscription.cancel(). However calling `cancel()` doesn't need to strictly prevent Subscriber onComplete/onError.

https://github.com/reactive-streams/reactive-streams-jvm?tab=readme-ov-file#2.3

> Subscriber.onComplete() and Subscriber.onError(Throwable t) MUST NOT call any methods on the Subscription or the Publisher.

https://github.com/reactive-streams/reactive-streams-jvm?tab=readme-ov-file#3.7

> After the Subscription is cancelled, additional Subscription.cancel() MUST be [NOPs](https://github.com/apple/servicetalk/pull/3040#term_nop).

https://github.com/reactive-streams/reactive-streams-jvm?tab=readme-ov-file#2.8

> A Subscriber MUST be prepared to receive one or more onNext signals after having called Subscription.cancel() if there are still requested elements pending [see [3.12](https://github.com/apple/servicetalk/pull/3040#3.12)]. Subscription.cancel() does not guarantee to perform the underlying cleaning operations immediately.

https://github.com/reactive-streams/reactive-streams-jvm?tab=readme-ov-file#3.12

> While the Subscription is not cancelled, Subscription.cancel() MUST request the Publisher to eventually stop signaling its Subscriber. The operation is NOT REQUIRED to affect the Subscription immediately.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
```diff
@@ -90,13 +91,22 @@ public void onSubscribe(final Cancellable cancellable) {
 
         @Override
         public void onSuccess(@Nullable final T result) {
+            propagateCancel = false;
             state.trySuccess(result);
         }
 
         @Override
         public void onError(final Throwable t) {
+            propagateCancel = false;
```
Scottmitch, line 100: this is what I had in mind (not tested). @idelpivnitskiy does that make sense?

```diff
diff --git a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
index 184fe6c2b..2217dfa7c 100644
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
@@ -77,7 +77,14 @@ final class AmbSingles<T> extends Single<T> {
     }
 
     static final class AmbSubscriber<T> extends DelayedCancellable implements Subscriber<T> {
+        private static final int INIT = 0;
+        private static final int TERMINATED = 1;
+        private static final int CANCELLED = 2;
+        @SuppressWarnings("rawtypes")
+        private static final AtomicIntegerFieldUpdater<AmbSubscriber> doneUpdater = newUpdater(AmbSubscriber.class, "done");
+
         private final State<T> state;
+        private volatile int done;
 
         AmbSubscriber(final State<T> state) {
             this.state = state;
@@ -90,12 +97,30 @@ final class AmbSingles<T> extends Single<T> {
 
         @Override
         public void onSuccess(@Nullable final T result) {
-            state.trySuccess(result);
+            // GAS allows for terminal after cancel. Downstream knows best how to process state after cancel.
+            if (doneUpdater.getAndSet(this, TERMINATED) != TERMINATED) {
+                state.trySuccess(result);
+            }
         }
 
         @Override
         public void onError(final Throwable t) {
-            state.tryError(t);
+            // GAS allows for terminal after cancel. Downstream knows best how to process state after cancel.
+            if (doneUpdater.getAndSet(this, TERMINATED) != TERMINATED) {
+                state.tryError(t);
+            }
+        }
+
+        @Override
+        public void cancel() {
+            for (;;) {
+                final int currDone = done;
+                if (currDone != INIT) { // we could let cancelled go through here if we wanted too
+                    break;
+                } else if (doneUpdater.compareAndSet(this, INIT, CANCELLED)) {
+                    super.cancel();
+                }
+            }
         }
     }
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
idelpivnitskiy: > However calling cancel() doesn't need to strictly prevent Subscriber onComplete/onError (see 2.8 and 3.12).

💯, thanks for clarifying and bringing these points.
So, we should only prevent cancel after terminal but not terminal after cancel.

idelpivnitskiy: @daschl worth adding a test if don't have one already to make sure we can still receive complete/error (from first/second) after the result on amb operation is cancelled.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
idelpivnitskiy: yes, this makes sense, but this is not much different from what @daschl already has in this PR and his version looks simpler. We don't need to prevent 2 terminal events at this level, we only need to prevent cancel after terminal

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
```diff
@@ -78,6 +78,7 @@ protected void handleSubscribe(final Subscriber<? super T> subscriber) {
 
     static final class AmbSubscriber<T> extends DelayedCancellable implements Subscriber<T> {
         private final State<T> state;
+        private volatile boolean propagateCancel = true;
```
idelpivnitskiy, line 81: We have an opportunity to avoid the second `volatile` state inside `AmbSubscriber` if we consolidate it under current `Cancellable` from `DelayedCancellable`. We can do this either by adding a new `protected` method `DelayedCancellable.ignoreCancel()` that will set `current` to `IGNORE_CANCEL` or (if we don't want more API on `DelayedCancellable`) we can remove `extends DelayedCancellable` from `AmbSubscriber` and repeat its logic here with a new feature to ignore cancel post termination.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/SingleAmbTest.java
```diff
@@ -72,6 +72,7 @@ private void init(final AmbParam ambParam) {
     void successFirst(final AmbParam ambParam) {
         init(ambParam);
         sendSuccessToAndVerify(first);
+        verifyNotCancelled(first);
```
idelpivnitskiy, line 75: Consider adding similar assertions in `CompletableAmbTest`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
daschl: Thanks for the clarifications! I updated the PR to reuse the volatile from `DelayedCancellable` and added a test to make sure we still see a terminal signal after the cancellation.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/AmbSingles.java
```diff
@@ -90,11 +90,13 @@ public void onSubscribe(final Cancellable cancellable) {
 
         @Override
         public void onSuccess(@Nullable final T result) {
+            ignoreCancel();
```
Scottmitch, line 93: There are 2 cases to consider consider:

1. Concurrency between Subscriber and Subscription for one AmbSubscriber object.
2. Concurrency between two different AmbSubscriber when `State` invokes cancel.

for (1) the following sequence is possible:

```
T1 - Subscription.cancel(), DelayedCancellable.GAS currentUpdater reads Cancellable into oldCancellable
T2 - AmbSubscriber.onSuccess(T), DelayedCancellable.ignoreCancel(), state.trySuccess, target.onSuccess
T1 - oldCancellable.cancel()
```

Because they are on independent threads this could happen even with atomic operations (assuming we want to always let Subscriber methods propagate even after cancel) but worth adding a comment to clarify. Can you confirm if this would be a problem for this issue you are observing?

for (2) the following sequence is possible:

```
T1 - AmbSubscriber1.onSuccess, ignoreCancel, state.trySuccess, doneUpdater.GAS, CompositeCancellable.cancel()
T2 - AmbSubscriber2.onSuccess
T1 - AmbSubscriber2.cancel() // cancel already invoked, upstream but already delivered onSuccess
T2 - AmbSubscriber2.ignoreCancel() // cancel already invoked, this won't have any effect
```

Again because each Single can have a different thread for Subscriber and Subscription, and we want to give precedence to the Subscriber, I'm not sure we can prevent this. However worth noting (and add a comment) to confirm this won't cause an issue with your use case.

I suggest clarifying in comments the expectations and constraints:
1. Let Subscriber take precedence - If Subscriber terminates after cancel(), we should propagate as we don't want to suppress work that is already complete and downstream may need to clean up.
2. Best effort enforcement of 2.3 - Subscriber and Subscription can be on different threads, and because of (1) it is possible cancel may happen "after" Subscriber methods are invoked if concurrency is involved in the following scenarios:
 - Concurrency between Subscriber and Subscription for one AmbSubscriber object.
 - Concurrency between two different AmbSubscriber when `State` invokes cancel.

idelpivnitskiy: Thank you for the detailed description. Imho, both example are legit races that are true for other operators as well. For that type of a race, @bryce-anderson opened #3038. 

Rule 2.3 says:
> `Subscriber.onComplete()` and `Subscriber.onError(Throwable t)` MUST NOT call any methods on the Subscription or the Publisher.

So, in this PR we only try to prevent a case when `onComplete/onError` invokes `cancel()` on the same thread after delivering the terminal. This helps to prevent a cycle.

idelpivnitskiy: The assumption we currently have in all other places is that if cancel **races** with a terminal on different threads, we should do our best to propagate terminal down, but because we know that we can only do 1 terminal we should pessimistically assume that some other `onError` can win in a chain of operators and the `onSuccess` will be discarded. For this reason, we should always propagate a racy `cancel` up to the originator to make sure it can clean up the state, if necessary.

idelpivnitskiy (approved):

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedCancellable.java
```diff
@@ -52,4 +52,11 @@ public void cancel() {
             oldCancellable.cancel();
         }
     }
+
+    /**
+     * Ignores any subsequent calls to {@link #cancel()}, preventing propagating the cancellation further up the stream.
+     */
+    protected void ignoreCancel() {
```
bryce-anderson, line 59:
```suggestion
    protected final void ignoreCancel() {
```

bryce-anderson, line 59: This is going to behave odd wrt ordering of `delayedCancellable` calls. In my minds eye, calling `ignoreCancel()` before or after `delayedCancellable` should have the same behavior, namely that cancellation will be ignored whereas right now if you call `ignoreCancel()` first it's going to result in immediate cancellation of the cancellable passed in through `delayedCancellable(..)`.

bryce-anderson (commented):
I agree that this is fine wrt the intrinsic races: the only goal here is to prevent breaking rule 2.3.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedCancellable.java
idelpivnitskiy: I would say that the current behavior is correct. We can only ignore if we know that this `Subscription` is already terminated. And in this case it's all correct, because terminal events can come only after `onSubscribe(Cancellable)`. We alway ignore after terminal.
If in some other use-case in the future, there is a path that calls `ignoreCancel()` before `delayedCancellable`, we should immediately cancel any new `Cancellable` because it may hold some resource that should be cleaned up.

bryce-anderson: At a minimum we need to clarify the behavior in the API docs because it reads as if it will do what I was describing.

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/DelayedCancellable.java
```diff
@@ -52,4 +52,11 @@ public void cancel() {
             oldCancellable.cancel();
         }
     }
+
+    /**
+     * Ignores any subsequent calls to {@link #cancel()}, preventing propagating the cancellation further up the stream.
+     */
```
bryce-anderson, line 58:
```suggestion
    /**
     * Ignores any subsequent calls to {@link #cancel()}, preventing propagating the cancellation further up the stream.
     * Note: if {@link DelayedCancellable#delayedCancellable(Cancellable)} is called after this method the provided
     * {@link Cancellable} will be immediately cancelled to prevent leakage of resources.
     */
```

bryce-anderson (approved):
Since this is package private right now I'm fine with just making the API contract more clear in the documentation.

daschl:
I'll follow up with the docs in a separate pr.

PR #3046, opened 2024-08-20
Subscribable sources: try-catch `onSubscribe` and cleanup state if any

idelpivnitskiy:
Motivation:

All original sources that implement `handleSubscribe` method should protect from a buggy `Subscriber` that may unexpectedly throw from `onSubscribe`. If they have any state or resource, it should be cleaned up.

Modifications:
- Find all `Publisher`/`Single`/`Completable` that implement `handleSubscribe` and use `handleExceptionFromOnSubscribe` utility;

The following sources clean up their state on exception:
- `FromInputStreamPublisher` closes `InputStream`;
- `SpliceFlatStreamToMetaSingle` cancels upstream `Subscription`;
- `TcpConnector` signals `ConnectionObserver.connectionClosed(t)`;
- `TcpServerBinder` cancels bind `Future` and closes `Channel` if necessary;
- `NettyChannelPublisher` registers `fatalError` and closes `Channel`;
- `DefaultNettyConnection`, `ChannelInitSingle`, `H2ClientParentConnectionContext`, `H2ServerParentConnectionContext` close `Channel`;

Results:

1. Best effort to propagate exception via reactive flow.
2. Cleaning up resources.

### DIFF
--- a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
+++ b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
@@ -61,6 +61,7 @@ import static io.servicetalk.transport.netty.internal.ChannelCloseUtils.close;
 import static io.servicetalk.transport.netty.internal.CopyByteBufHandlerChannelInitializer.POOLED_ALLOCATOR;
 import static io.servicetalk.transport.netty.internal.EventLoopAwareNettyIoExecutors.toEventLoopAwareNettyIoExecutor;
 import static io.servicetalk.transport.netty.internal.ExecutionContextUtils.channelExecutionContext;
+import static io.servicetalk.utils.internal.ThrowableUtils.addSuppressed;
 import static java.util.Objects.requireNonNull;
 
 /**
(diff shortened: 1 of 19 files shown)


servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
```diff
@@ -183,7 +184,13 @@ protected void initChannel(final Channel channel) {
         return new SubscribableSingle<ServerContext>() {
             @Override
             protected void handleSubscribe(Subscriber<? super ServerContext> subscriber) {
-                subscriber.onSubscribe(() -> future.cancel(true));
+                try {
+                    subscriber.onSubscribe(() -> future.cancel(true));
+                } catch (Throwable t) {
+                    handleExceptionFromOnSubscribe(subscriber, t);
+                    future.cancel(true);
```
bryce-anderson, line 191: I don't know if this is enough: I think we also need to add a listener to the future in the case that cancel isn't effective to cleanup the resource once it resolves.

bryce-anderson (approved):
Only one change necessary in `TcpServerBinder` and then ship it.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpServerBinder.java
idelpivnitskiy: Great catch!
Added: https://github.com/apple/servicetalk/pull/3046/commits/ca099b70896305b1b7b381c18c95a9837146236e

bryce-anderson (approved):

PR #3047, opened 2024-08-20
`ConnectablePayloadWriter`: avoid static exception that may leak memory

idelpivnitskiy:
Motivation:

`ConnectablePayloadWriter` uses a static `IOException` instance as a state and can propagate it to `Subscriber`. This opens a possibility for a memory leak if `Subscriber` or intermediate operators attach a suppressed exception to this static instance.

Modifications:
- Create `StacklessCancelledIOException` and always use a new instance in case of cancellation;
- Keep `cancel()` noop if `outer.closed != null`;

Result:

`ConnectablePayloadWriter` can not leak memory via static exception instance.

### DIFF
--- a/servicetalk-concurrent-api-internal/src/main/java/io/servicetalk/concurrent/api/internal/ConnectablePayloadWriter.java
+++ b/servicetalk-concurrent-api-internal/src/main/java/io/servicetalk/concurrent/api/internal/ConnectablePayloadWriter.java
@@ -20,6 +20,7 @@ import io.servicetalk.concurrent.PublisherSource.Subscription;
 import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.concurrent.internal.DuplicateSubscribeException;
 import io.servicetalk.concurrent.internal.TerminalNotification;
+import io.servicetalk.concurrent.internal.ThrowableUtils;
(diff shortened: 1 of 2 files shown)


servicetalk-concurrent-api-internal/src/main/java/io/servicetalk/concurrent/api/internal/ConnectablePayloadWriter.java
```diff
@@ -329,7 +327,11 @@ public void request(final long n) {
 
                     @Override
                     public void cancel() {
-                        if (closedUpdater.compareAndSet(outer, null, error(CONNECTED_PUBLISHER_CANCELLED))) {
+                        if (outer.closed != null) {
```
daschl, line 330: is that changing the semantics compared to not checking previously?

idelpivnitskiy: I don't think so. The code below is doing something only if `outer.closed == null`. Otherwise, it's noop. Because previously CAS attempt did not allocate anything, this check was not necessary. Now it helps to avoid unnecessary allocation if the state is not null anymore -> we know CAS will fail.

bryce-anderson (approved):

PR #3053, opened 2024-09-03
gardening: parameterize HttpObjectDecoderTest and subclasses

daschl:
This changeset refactors the `HttpObjectDecoderTest` and its related classes to use the `ParameterizedTest` instead of manually delegating.


idelpivnitskiy (approved):
Thank you!

PR #3055, opened 2024-09-09
Stop using `servicetalk-dependencies` internally

idelpivnitskiy:
Motivation:

Gradle interprets `*.module` files differently than the `pom.xml`. Because every module currently depends on `servicetalk-dependencies`, users pulled its `dependencyConstraints` in their build configurations with any servicetalk module as transitive dependency and got unintended changes for other dependencies.

For example, `servicetalk-annotations` that has only jsr305 as a 3-party dependency still forces upgrade of jackson, protobuf, log4j, netty, and jersey for users because it depends on `servicetalk-dependencies`.

Modifications:

- Remove `servicetalk-dependencies` from every module, instead define every necessary dependency by its version number or import only required boms;
- Update `servicetalk-dependencies` to include missed dependencies: opentelemetry, instrumentation, resilience4j;
- Correct dependencies for all `-jersery3` modules to keep only what they use;
- Move all `platform` dependencies from the top to their corresponding scope;
- Regenerate lock files;

Result:

Users don't get unintended dependency upgrades because of the `*.module` file handling.


servicetalk-benchmarks/build.gradle
```diff
@@ -33,9 +32,9 @@ dependencies {
   implementation project(":servicetalk-transport-netty-internal")
   implementation project(":servicetalk-loadbalancer")
   implementation project(":servicetalk-data-jackson")
-  implementation "com.google.code.findbugs:jsr305"
+  implementation "com.google.code.findbugs:jsr305:$jsr305Version"
   implementation "io.netty:netty-codec-http"
```
daschl, line 36: since there is only one netty dependency, would it make sense to also just specify it with $nettyVersion and not pull in the bom for consistency?

servicetalk-annotations/build.gradle
```diff
@@ -17,7 +17,5 @@
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-library"
 
 dependencies {
-  implementation platform(project(":servicetalk-dependencies"))
-
-  implementation "com.google.code.findbugs:jsr305"
+  implementation "com.google.code.findbugs:jsr305:$jsr305Version"
```
daschl, line 20: I know this is slightly offtopic, but would it make sense to expose this as api instead of implementation? then I think in i.e. buffer-api we do not need to repeat `implementation "com.google.code.findbugs:jsr305:$jsr305Version"` if the annotations are already imported (and we expose it as api, not an implementation detail)

servicetalk-buffer-netty/build.gradle
```diff
@@ -17,17 +17,16 @@
 apply plugin: "io.servicetalk.servicetalk-gradle-plugin-internal-library"
 
 dependencies {
-  api platform(project(":servicetalk-dependencies"))
-  testImplementation enforcedPlatform("org.junit:junit-bom:$junit5Version")
-
+  api platform("io.netty:netty-bom:$nettyVersion")
```
daschl, line 20: what's the benefit of using the netty-bom vs. explicitly using the version for each of the netty dependencies? (thought: keeps all dependency lines/versions consistent in the approach)

daschl (commented):
Overall lgtm, couple small questions

servicetalk-benchmarks/build.gradle
idelpivnitskiy: when a library publishes a bom, it's always best to use bom even for 1 dependency because it helps build tools to align version between other modules when final applications brings more modules

servicetalk-annotations/build.gradle
idelpivnitskiy: yes, you are right. I will do a follow-up for this suggestion

servicetalk-benchmarks/build.gradle
idelpivnitskiy: plus netty-bom helps to bring correct tcnative dependency, if we manage it manually we will always have to check what tcnative version is compatible with the current netty version

servicetalk-annotations/build.gradle
daschl: 👍

servicetalk-benchmarks/build.gradle
daschl: ok that's a good second point, I'll take it 😅

daschl (approved):

servicetalk-annotations/build.gradle
idelpivnitskiy: See #3056

PR #3059, opened 2024-09-19
Allow null value for query parameter setter methods

Scottmitch:
Motivation:
041f3dc9ce3fd0e423263a88e0df8cab7fda14ec replace empty string for null during parsing. However the setter/accessor methods don't consistently allow null values.

Modifications:
- Allow null values from add & set queryParameter methods.


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpQuery.java
```diff
@@ -115,7 +115,7 @@ public HttpQuery add(final String key, final String... values) {
     }
 
     public HttpQuery set(final String key, final String value) {
```
bryce-anderson, line 117: Do we need the `@Nullable` annotation on `value`? Samve with `add` above.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpQuery.java
```diff
@@ -115,7 +115,7 @@ public HttpQuery add(final String key, final String... values) {
...
-        validateQueryParam(key, value);
+        validateQueryParam(key);
         final ArrayList<String> list = new ArrayList<>(DEFAULT_LIST_SIZE);
         list.add(value);
```
bryce-anderson, line 120: I can see this going either way but I wanted to ask the question: do we want to add `null` to the list? It seems valuable if someone wants to do something like `add(a, null); add(a, "b");` but otoh, that means we will also start emitting an `Iterable` that may contain null values which feels like a footgun for users.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpQuery.java
Scottmitch: yes good call, I realized this too after I pushed the PR. done now.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpQuery.java
Scottmitch: bcz the user can add multiple params with the same key (e.g. `foo&foo=&foo=value`) I think we need some value so we can restore the original query string and folks know there was an empty value. I updated `testNullQueryParamValue` to include this case.

bryce-anderson: That's a good point. It's still a footgun, but I do agree with you that we need to represent that case somehow and I don't see a better way.

bryce-anderson (approved):
lgtm so long as CI is happy.

daschl (approved):

PR #3073, opened 2024-10-02
Add experimental system properties for Netty `DnsNameResolver`

idelpivnitskiy:
Motivation:

Netty recently introduced 2 more DNS config options that help to workaround known infrastructure issues:
- TCP fallback on timeout;
- `DnsNameResolverChannelStrategy`;

Modifications:

Add 2 new experimental system properties:
1. `io.servicetalk.dns.discovery.netty.experimental.tcpFallbackOnTimeout`
2. `io.servicetalk.dns.discovery.netty.experimental.datagramChannelStrategy`

Result:

We can test these new features without commiting to them in our public API.


daschl (approved):
LGTM - I assume this is validated manually that the properties are applied correctly?

idelpivnitskiy:
I validated that they are applied when required Netty version is in classpath. There is no easy way to validate that nothing breaks if older Netty version is in classpath. However, this use-case is rare and we can fix if there is a reported problem.

bryce-anderson (approved):

PR #3074, opened 2024-10-02
Add SslConfig to onSecurityHandshake

daschl:
Motivation
----------
This changeset adds the `SslConfig` if available to the `onSecurityHandshake`, similar to a change made to
`onTransportHandshakeComplete` where the full `ConnectionInfo` gets passed in. This allows to get
information about the config from the argument itself, without having to get it as an instance variable
from `onTransportHandshakeComplete`.

### DIFF
--- a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java
+++ b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java
@@ -95,7 +95,7 @@ public class TcpClientChannelInitializer implements ChannelInitializer {    // F
                             // ExecutionContext can be null if users used deprecated ctor
                             executionContext == null ? null : channelExecutionContext(channel, executionContext),
                             sslConfig, config.idleTimeoutMs()),
-                    sslConfig != null && !deferSslHandler, true));
+                    sslConfig != null && !deferSslHandler, true, sslConfig));
         }
 
         if (config.idleTimeoutMs() > 0L) {
--- a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java
+++ b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java
@@ -94,8 +94,7 @@ public class TcpClientChannelInitializer implements ChannelInitializer {    // F
                     channel -> new TcpConnectionInfo(channel,
                             // ExecutionContext can be null if users used deprecated ctor
                             executionContext == null ? null : channelExecutionContext(channel, executionContext),
-                            sslConfig, config.idleTimeoutMs()),
-                    sslConfig != null && !deferSslHandler, true, sslConfig));
+                            sslConfig, config.idleTimeoutMs()), true, !deferSslHandler ? sslConfig : null));
         }
 
         if (config.idleTimeoutMs() > 0L) {
--- a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java
+++ b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java
@@ -94,7 +94,7 @@ public class TcpClientChannelInitializer implements ChannelInitializer {    // F
                     channel -> new TcpConnectionInfo(channel,
                             // ExecutionContext can be null if users used deprecated ctor
                             executionContext == null ? null : channelExecutionContext(channel, executionContext),
-                            sslConfig, config.idleTimeoutMs()), true, !deferSslHandler ? sslConfig : null));
+                            sslConfig, config.idleTimeoutMs()), true, deferSslHandler ? null : sslConfig));
         }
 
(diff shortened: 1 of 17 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -106,8 +106,35 @@ default ProxyConnectObserver onProxyConnect(Object connectMsg) {    // FIXME: 0.
...
+     *     rejected by the server.</li>
+     *     <li>For a proxy connections, the handshake may happen after an observer returned by
+     *     {@link #onProxyConnect(Object)} completes successfully.</li>
+     * </ol>
+     *
+     * @param sslConfig the {@link SslConfig} used when performing the security handshake.
+     * @return a new {@link SecurityHandshakeObserver} that provides visibility into security handshake events
+     * @see <a href="https://datatracker.ietf.org/doc/html/rfc7413">RFC7413: TCP Fast Open</a>
+     */
+    default SecurityHandshakeObserver onSecurityHandshake(SslConfig sslConfig) {
```
idelpivnitskiy, line 135: The new method should also be implemented in `transport.api.NoopTransportObserver`, `transport.netty.internal.NoopTransportObserver`.
In `SslCertificateCompressionTest` and `HttpTransportObserverAsyncContextTest` it's better to migrate from deprecated to new method.

The trip I use: first add a new method on interface without `default` impl. Implement it everywhere. Then add `default` impl.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ConnectionObserverInitializer.java
```diff
@@ -75,15 +78,37 @@ public ConnectionObserverInitializer(final ConnectionObserver observer,
...
+     * @param observer {@link ConnectionObserver} to report network events
+     * @param connectionInfoFactory {@link Function} that creates {@link ConnectionInfo} from the provided
+     * {@link Channel} to report {@link ConnectionObserver#onTransportHandshakeComplete(ConnectionInfo)}
+     * @param handshakeOnActive {@code true} if the observed connection is secure
+     * @param client {@code true} if this initializer is used on the client-side
+     * @param sslConfig the {@link SslConfig} to supply to the observer on handshake.
+     */
+    public ConnectionObserverInitializer(final ConnectionObserver observer,
+                                         final Function<Channel, ConnectionInfo> connectionInfoFactory,
+                                         final boolean handshakeOnActive,
```
idelpivnitskiy, line 104: `handshakeOnActive` is effectively a `config.sslConfig() != null` check. Consider using `sslConfig` instead of yet another ctor argument. For `TcpClientChannelInitializer` we will pass `sslConfig` only if `!deferSslHandler`.

idelpivnitskiy (commented):
Couple comments then lgtm:

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ConnectionObserverInitializer.java
daschl: In the client case it is `sslConfig != null && !deferSslHandler`, an in the deprecation overrides we can only set null so it will always be false. The second case might be ok, but how would we deal with the deferSslHandler check unless adding another parameter too?

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/TcpClientChannelInitializer.java
```diff
@@ -94,8 +94,7 @@ public TcpClientChannelInitializer(final ReadOnlyTcpClientConfig config,
                     channel -> new TcpConnectionInfo(channel,
                             // ExecutionContext can be null if users used deprecated ctor
                             executionContext == null ? null : channelExecutionContext(channel, executionContext),
-                            sslConfig, config.idleTimeoutMs()),
-                    sslConfig != null && !deferSslHandler, true));
+                            sslConfig, config.idleTimeoutMs()), true, !deferSslHandler ? sslConfig : null));
```
idelpivnitskiy, line 97: To avoid the boolean flip:
```suggestion
                            sslConfig, config.idleTimeoutMs()), true, deferSslHandler ? null : sslConfig));
```

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ConnectionObserverInitializer.java
```diff
@@ -73,17 +76,37 @@ public ConnectionObserverInitializer(final ConnectionObserver observer,
...
+     * @param sslConfig the {@link SslConfig} to supply to the observer on handshake.
+     */
+    public ConnectionObserverInitializer(final ConnectionObserver observer,
+                                         final Function<Channel, ConnectionInfo> connectionInfoFactory,
+                                         final boolean client,
+                                         @Nullable final SslConfig sslConfig) {
         this.observer = requireNonNull(observer);
         this.connectionInfoFactory = requireNonNull(connectionInfoFactory);
-        this.handshakeOnActive = handshakeOnActive;
+        this.handshakeOnActive = sslConfig != null;
```
idelpivnitskiy, line 107: We can completely remove `handshakeOnActive` and make the class state smaller

idelpivnitskiy (approved):
Minors nits:

PR #3081, opened 2024-10-15
traffic-resilience-http: extend connect timeout for CI even further

idelpivnitskiy:
Motivation:

Even 4 seconds added in #3077 seems too low for CI. Trying higher value. Also, we can reuse already existing `CI` constant.


bryce-anderson:
I don't mind extending it further, but it looks like that also didn't work so it might not be a timeout thing.

idelpivnitskiy:
Looks like increasing the timeout made `TrafficResilienceHttpServiceFilterTest` even more likely to fails

idelpivnitskiy:
I will close this radar, opened #3093 for minor corrections

PR #3083, opened 2024-10-30
More flexibility for AbstractCloseableIteratorAsInputStream subclasses

daschl:
This changeset removes the final modifier for the two read methods as well as makes one method protected (from private) in order to allow any subclass more flexibility if they want to perform additional logic on read and/or decide what to do if the class has been closed already.

### DIFF
--- a/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/AbstractCloseableIteratorAsInputStream.java
+++ b/servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/AbstractCloseableIteratorAsInputStream.java
@@ -91,7 +91,7 @@ public abstract class AbstractCloseableIteratorAsInputStream<T> extends InputStr
     protected abstract boolean isClosed();
 
     @Override
-    public final int read(final byte[] b, int off, int len) throws IOException {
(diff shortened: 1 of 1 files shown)


idelpivnitskiy (approved):

servicetalk-concurrent-internal/src/main/java/io/servicetalk/concurrent/internal/AbstractCloseableIteratorAsInputStream.java
```diff
@@ -164,9 +164,13 @@ public final int read() throws IOException {
...
     }
 
-    private void checkAlreadyClosed() throws IOException {
+    protected void checkAlreadyClosed() throws IOException {
         if (isClosed()) {
             throw new IOException("Stream is already closed.");
         }
     }
+
+    protected boolean hasNext() {
```
idelpivnitskiy, line 173: Btw, 2 implementations that we currently have don't hold a reference to `iterator`. Should we pass it there as an argument. I see you had it originally but then removed, why?

daschl: Actually, turns out, we don't need to override `hasNext` I think so I will clean this change up a bit.

PR #3090, opened 2024-11-07
Migrate the javapoet dependency to a supported project

mgodave:
### Motivation

The dependency https://github.com/square/javapoet hasn't been updated in years and is no longer maintained by the original authors. A new fork, https://github.com/palantir/javapoet, has been made by a third party and is currently supported.

### Changes

Move the dependency on javapoet from com.squareup to com.palantir.javapoet and update the code dependencies to reflect the change in packaging between the two dependencies.


Scottmitch (approved):

mgodave:
Javapoet is compiled with Java17 so any of our builds with JDKS lower than this will fail.

```
bad class file: /home/runner/.gradle/caches/modules-2/files-2.1/com.palantir.javapoet/javapoet/0.5.0/a7b8f3e3797ed8d83d0fec2a723dd6b39994a746/javapoet-0.5.0.jar(/com/palantir/javapoet/TypeSpec.class)
    class file has wrong version 61.0, should be 55.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
/home/runner/work/servicetalk/servicetalk/servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/ServiceClassBuilder.java:21: error: package TypeSpec does not exist
```

mgodave:
Due to the class file version issue stated above I am going to close this and look into moving away from javapoet altogether.

PR #3092, opened 2024-11-07
Preserve default `Http2Settings` when `Http2SettingsBuilder` is used

idelpivnitskiy:
Motivation:

By default we set values for `MAX_HEADER_LIST_SIZE` and `INITIAL_WINDOW_SIZE` settings at `H2ProtocolConfigBuilder` level. However, if users want to change only one setting (for example, `MAX_CONCURRENT_STREAMS`), they lose our defaults when they use `Http2SettingsBuilder`.

Modifications:

- Move defaults initialization from `H2ProtocolConfigBuilder` to `Http2SettingsBuilder`.

Result:

Users always start with consistent default `Http2Settings` and can use builder methods to override what they want.


daschl (approved):

mgodave (approved):
This makes sense in the expected behavior that the user would expect.

PR #3093, opened 2024-11-07
Minor corrections for tests

idelpivnitskiy:
Modifications:
- Reuse `CI` constant.
- Use `localAddress` utility instead of `forPort` to avoid binding to non-locahost address that can be unavailable in some CI envs.


bryce-anderson (approved):

bryce-anderson:
🚢  after fixing the style checks.

PR #3096, opened 2024-11-09
Detect unexpected response leaks for multi-address client instances

idelpivnitskiy:
Motivation:

We already apply `HttpMessageDiscardWatchdogClientFilter` to single-address clients to detect response leaks that could be caused by unhandled exceptions in the filter chain. Multi-address client adds more logic around a client group and therefore also could leak responses if case on unexpected exceptions.

Modifications:

- Apply `HttpMessageDiscardWatchdogClientFilter` to multi-address client instances;
- Enhance `HttpMessageDiscardWatchdogClientFilter` to include unhandled exception in the logs to help narrow down the leak cause;
- `CleanerStreamingHttpClientFilterFactory`: replace `onErrorResume` with `whenOnError` to reduce RS overhead (avoids subscribe sequence for returned failed single);

Result:

Users will see warn message if their multi-address client leaks responses.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
@@ -135,6 +135,9 @@ final class DefaultMultiAddressUrlHttpClientBuilder
             urlClient = redirectConfig == null ? urlClient :
                     new RedirectingHttpRequesterFilter(redirectConfig).create(urlClient);
 
+            // Detect leaks that can be caused by unexpected exceptions
+            urlClient = HttpMessageDiscardWatchdogClientFilter.CLIENT_CLEANER.create(urlClient);
+
             LOGGER.debug("Multi-address client created with base strategy {}", executionContext.executionStrategy());
             return new FilterableClientToClient(urlClient, executionContext);
         } catch (final Throwable t) {
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
@@ -115,7 +115,7 @@ final class HttpMessageDiscardWatchdogClientFilter implements StreamingHttpConne
                                     LOGGER.warn("Discovered un-drained HTTP response message body which has " +
                                             "been dropped by user code - this is a strong indication of a bug " +
                                             "in a user-defined filter. Response payload (message) body must " +
-                                            "be fully consumed before discarding.");
+                                            "be fully consumed before discarding.", cause);
                                 }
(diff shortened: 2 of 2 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultMultiAddressUrlHttpClientBuilder.java
```diff
@@ -135,6 +135,9 @@ public StreamingHttpClient buildStreaming() {
             urlClient = redirectConfig == null ? urlClient :
                     new RedirectingHttpRequesterFilter(redirectConfig).create(urlClient);
 
+            // Detect leaks that can be caused by unexpected exceptions
+            urlClient = HttpMessageDiscardWatchdogClientFilter.CLIENT_CLEANER.create(urlClient);
```
bryce-anderson, line 139: I think this results in 2 cleaners. Maybe it's not a big deal, but just wanted to be sure I understood what the structure will be.

idelpivnitskiy: Correct, because from a single address client we don't know if it's a standalone instance or part of the multi-address client group we can not remove it from there. So, for multi-address flow we will have 2 cleaners for `onError` case:
1. Last one before returning from single-address level.
2. Last one before returning from multi-address level to the users.

This should not be a big deal because the watchdog overhead mostly comes from the connection level filter that instantiates `AtomicReference` for each request and transforms response message body for each response. The overhead of `CleanerStreamingHttpClientFilterFactory` is negligible bcz `onError` handling is super cheap and happens only in case of exceptions.

idelpivnitskiy: Sent a new commit to also replace `onErrorResume` with `whenOnError`. It will help to reduce overhead (avoids subscribe sequence for returned failed single).

daschl: I was suggesting we could make the Single client aware if it gets instantiated from a multi client, but your explanation makes me think it's probably not a big deal. Let's be sure and check the perf stats after merge.

daschl (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpMessageDiscardWatchdogClientFilter.java
```diff
@@ -107,7 +107,7 @@ protected Single<StreamingHttpResponse> request(final StreamingHttpRequester del
                                                                 final StreamingHttpRequest request) {
                     return delegate
                             .request(request)
-                            .onErrorResume(cause -> {
+                            .whenOnError(cause -> {
```
bryce-anderson, line 110: ffti: if we want to preserve the 'sequencing' of events, maybe we should be explicit and use `.beforeOnError`. I recognize that is what `.whenOnError` forwards too, but technically the `whenOnError` behavior is undefined and supposedly could change.

bryce-anderson (approved):

PR #3097, opened 2024-11-12
Update Netty 4.1.114 -> 4.1.115

idelpivnitskiy:

PR #3102, opened 2024-11-13
Discard new writes without closing connection on `AbortWritesEvent`

idelpivnitskiy:
Motivation:

Behavior of `SslHandler` in Netty 4.1.115 changed. Now it can wrap data and flush them to the network even if flush was not requested. In result, it affected our test `ConnectionCloseHeaderHandlingTest.PipelinedRequestsTest.serverCloseSecondPipelinedRequestWriteAborted` that we had to disable in #3097 to proceed with the upgrade.
Debugging showed that when we receive `AbortWritesEvent`, the following behavior of `WriteStreamSubscriber` depends on how many `activeWrites` we have. With 4.1.115 netty will flush data, there will be 0 `activeWrites` and it will close (with reset) the outbound. In result, we lose data of the previous response in the pipeline that we haven't read yet.

Modifications:
- Use `listenerDiscard` instead of `channelClosed` that will make sure new writes will be discarded without prematurely affecting the connection state regardless of the `activeWrites` count.
- Unskip the test.

Result:

Graceful handling of `connection: close` header does not depend on flushing behavior.


chrisvest (approved):
Sounds good from the description. I can't tell if the close outbound is delayed, or just doesn't happen, though.

Scottmitch (approved):

PR #3103, opened 2024-11-14
Reset flush strategy after client request is written

idelpivnitskiy:
Motivation:

Currently we reset the flush strategy after a request is written and its response is received. If client starts writing another pipelined request on the same connection, it ends up using an overridden flush strategy of the previous request that is still waiting for its response.

Modifications:
- For client connections (pipelined and non-pipelined) move flush strategy reset from "after full request-response completion" to "before completion of the request write". Using `beforeFinally` we guarantee that reset will happen before any retry/repeat operator initiates a new request on the same connection. It's safe to use "before" because for in-flight request the effective `FlushStrategy` is captured at the beginning of the write and doesn't change regardless of the reset. Only a new request will see the effect of the reset.
- Reproduce described scenario in `FlushStrategyOnClientTest`.
- Refactor `FlushStrategyOnServerTest` to use `TransportObserver` to observe flushes. This helps to use a real `HttpServerContext` instead of faking a `ServerContext` and reuse code for `FlushStrategyOnClientTest`.
- Rename `Cancellable` in `NettyHttpServer` to highlight it's responsible for reset of FlushStrategy.

Result:

Second pipelined request on the same connection uses its own `FlushStrategy`.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnClientTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnClientTest.java
@@ -88,13 +88,13 @@ class FlushStrategyOnClientTest {
         }
     }
 
-    private StreamingHttpRequest newRequest(StreamingHttpConnection connection, String path) {
+    private static StreamingHttpRequest newRequest(StreamingHttpConnection connection, String path) {
         BufferAllocator alloc = connection.executionContext().bufferAllocator();
         return connection.post(path)
                 .payloadBody(Publisher.from(alloc.fromAscii("foo"), alloc.fromAscii("bar")));
     }
 
-    private void assertResponse(Future<StreamingHttpResponse> responseFuture) throws Exception {
+    private static void assertResponse(Future<StreamingHttpResponse> responseFuture) throws Exception {
         StreamingHttpResponse response = responseFuture.get();
         assertThat(response.status(), is(OK));
         Buffer payload = response.toResponse().toFuture().get().payloadBody();
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnClientTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnClientTest.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright © 2024 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.netty;
+
+import io.servicetalk.buffer.api.Buffer;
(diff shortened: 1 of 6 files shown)


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnClientTest.java
```diff
@@ -0,0 +1,103 @@
...
+            assertFlushOnEach(interceptor);
+            canReturnResponse.countDown();
+            assertThat(receivedRequests.take().requestTarget(), is("/second"));
+
+            assertResponse(firstResponse);
+            assertResponse(secondResponse);
+        }
+    }
+
+    private StreamingHttpRequest newRequest(StreamingHttpConnection connection, String path) {
```
daschl, line 91: nit: can be static

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnClientTest.java
```diff
@@ -0,0 +1,103 @@
...
+        BufferAllocator alloc = connection.executionContext().bufferAllocator();
+        return connection.post(path)
+                .payloadBody(Publisher.from(alloc.fromAscii("foo"), alloc.fromAscii("bar")));
+    }
+
+    private void assertResponse(Future<StreamingHttpResponse> responseFuture) throws Exception {
```
daschl, line 97: nit: can be static

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnClientTest.java
```diff
@@ -0,0 +1,103 @@
...
+        try (HttpServerContext serverContext = BuilderUtils.newServerBuilder(SERVER_CTX)
+                .listenBlockingAndAwait(((ctx, request, responseFactory) -> {
+                    receivedRequests.add(request);
+                    canReturnResponse.await();
+                    return responseFactory.ok();
+                }));
+             StreamingHttpClient client = BuilderUtils.newClientBuilder(serverContext, CLIENT_CTX)
+                     .appendConnectionFactoryFilter(new TransportObserverConnectionFactoryFilter<>(interceptor))
+                     .buildStreaming();
+             ReservedStreamingHttpConnection connection = client.reserveConnection(client.get("/")).toFuture().get()) {
```
daschl, line 71: should the connection be closed after the tests complete?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
```diff
@@ -79,8 +67,6 @@ class FlushStrategyOnServerTest {
     private static final String USE_EMPTY_RESP_BODY = "empty-resp-body";
 
     private OutboundWriteEventsInterceptor interceptor;
-    private HttpHeadersFactory headersFactory;
-
     private ServerContext serverContext;
```
daschl, line 70: nit: you added a non null check in `tearDown`, should these be marked as Nullable to make IJ happier?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/FlushStrategyOnServerTest.java
```diff
@@ -328,5 +301,44 @@ int takeWritesTillFlush() throws Exception {
...
+        @Override
+        public void connectionClosed(final Throwable error) {
+        }
+
+        @Override
+        public void connectionClosed() {
+        }
+    }
+
+    private static final class NoFlushStrategy implements FlushStrategy {
```
daschl, line 333: Unused?

idelpivnitskiy: leftover from an intermediate experiment

PR #3104, opened 2024-11-14
Preserve full request context while redirecting

idelpivnitskiy:
Motivation:

Redirect handling creates new requests, but copies only `HTTP_EXECUTION_STRATEGY_KEY`. As the result, next filters after redirect don't see a state associated with the request flow.

Modifications:

- Copy full `ContextMap` from each previous request to the next one;

Result:

Any context information is preserved for all redirected requests, like UUID, counters, execution strategy, etc.


daschl (approved):
Change looks good to me, just to double check: I assume there is no context that we do _not_ want to carry over after a redirect?

idelpivnitskiy:
@daschl I don't have a use-case in mind when ppl won't want some context to propagate forward. Context is local to the app and does not leak over wire. Should be fine to copy all. If there is a use-case, users can always use request transformer to clean some parts of the context they don't want to propagate or if they want to reset something there.

PR #3108, opened 2024-11-15
Publisher#flatMapConcatIterable may skip emitting items

Scottmitch:
Motivation:
Publisher#flatMapConcatIterable may not emit some items due to race conditions and visibility issues. The iterator state is written to outside the scope of holding the lock. After a drain loop completes we may request 1 more iterator. However it is possible the thread emitting holds the lock while another thread invokes onNext(t). The emitting thread may not see the iterator, and instead see `EmptyIterator.instance()` and cause it to request 1 more item, but then the not-visible iterator contents won't be emitted.

Modifications:
- Make FlatMapIterableSubscriber iterator state volatile and atomically update it. There is only ever 1 valid iterator because only 1 outstanding demand is issued only after the current iterator `!hasNext()`. The iterator state is re-read on each drain loop, and the terminal condition must atomically set to EmptyIterator.


idelpivnitskiy (approved):
2 Fast 2 Furious 😄

bryce-anderson:
lgtm as well. Nice find.

mgodave:
How did we find this? Was it reported by a user?

idelpivnitskiy:
@mgodave yes

PR #3110, opened 2024-11-15
[grpc-protoc] Add an option to generate default service methods

mgodave:
### Motivation

When evolving a service definition that has multiple implementations it is ideal not to break these implementations. Therefore it would be nice if the protoc generator was able to generate default implementations on the service interface to ensure implementations always conform. 

### Changes

Add an option, `defaultServiceMethods` to the grpc-protoc stub compiler to generate these default implementation. The result is that service interfaces will provide a default, throwing, implementation of all service RPC interfaces that they implement. 

test_service.proto: https://gist.github.com/mgodave/d9064c67da9f365de70ef7526c9eea72


servicetalk-examples/grpc/protoc-options/src/main/java/io/servicetalk/examples/grpc/protocoptions/BlockingProtocOptionsServer.java
```diff
@@ -23,8 +25,12 @@
 public final class BlockingProtocOptionsServer {
     public static void main(String[] args) throws Exception {
         GrpcServers.forPort(8080)
-                .listenAndAwait((BlockingGreeterService) (ctx, request) ->
-                        HelloReply.newBuilder().setMessage("Hello " + request.getName()).build())
+                .listenAndAwait(new BlockingGreeterService() {
```
idelpivnitskiy, line 28: Was this change required because `BlockingGreeterService` is no longer a `@FunctionalInterface`?

servicetalk-examples/grpc/protoc-options/build.gradle
```diff
@@ -74,6 +74,7 @@ protobuf {
           outputSubDir = "java"
           // Option to append a suffix to type names to avoid naming collisions with other code generation.
           option 'typeNameSuffix=St'
+          option 'defaultServiceMethods=true'
```
idelpivnitskiy, line 77: My expectation was that we can add this behavior without introducing a new property, but if that caused a change in `BlockingProtocOptionsServer` (services with a single RPC that were used as lambda) then it's good to protect by an option.

Please add a comment above clarifying what this option does.

servicetalk-examples/grpc/protoc-options/src/main/proto/test_service.proto
```diff
@@ -0,0 +1,37 @@
...
+
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "io.servicetalk.grpc.netty";
+option java_outer_classname = "TesterProto";
+
+package grpc.netty;
+
+service Tester {
```
idelpivnitskiy, line 24: Reminder to remove this file before merging PR

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Main.java
```diff
@@ -149,6 +171,7 @@ private static CodeGeneratorResponse generate(final CodeGeneratorRequest request
         }
         final String typeSuffixValue = optionsMap.get(TYPE_NAME_SUFFIX_OPTION);
         final boolean printJavaDocs = parseBoolean(optionsMap.getOrDefault(PRINT_JAVA_DOCS_OPTION, "true"));
+        final boolean defaultServiceMethods = parseBoolean(optionsMap.getOrDefault(DEFAULT_SERVICE_METHODS, "false"));
```
idelpivnitskiy, line 174: Please add a comment: `// FIXME: 0.43 - consider changing default value to true`

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -1465,6 +1467,33 @@ private TypeSpec newServiceInterfaceSpec(final State state, final boolean blocki
...
+                final String methodName = sanitizeIdentifier(methodProto.getName(), true);
+                final int methodIndex = i;
+                interfaceSpecBuilder.addMethod(newRpcMethodSpec(inClass, outClass, methodName,
+                        methodProto.getClientStreaming(),
+                        methodProto.getServerStreaming(),
+                        !blocking ? EnumSet.of(INTERFACE) : EnumSet.of(INTERFACE, BLOCKING),
+                        printJavaDocs, (__, c) -> {
+                            c.addModifiers(DEFAULT)
+                                    .addParameter(GrpcServiceContext, ctx)
+                                    .addStatement("throw new UnsupportedOperationException(\"Method " + methodName +
```
idelpivnitskiy, line 1485: 1. gRPC services should use `GrpcStatusException` with `UNIMPLEMENTED` status, similar to official grpc-java to be compatible. Consider adding a test in `ProtocolCompatibilityTest` to make sure our default behavior is the same as grpc-java default behavior. In grpc-netty tests we can always use this new option, no need to create a new test module.
2. Async RPC that return `Single` or `Publisher` can not throw, they should return a failed async source, like `Single.failed(...)` or `Publisher.failed(...)`.
3. grpc-java uses `fullMethodName()` from method descriptor for exception message.

idelpivnitskiy (changes requested):

servicetalk-examples/grpc/protoc-options/src/main/java/io/servicetalk/examples/grpc/protocoptions/BlockingProtocOptionsServer.java
mgodave: That is correct, in fact it was never officially an enforced `@FunctionalInterface`

servicetalk-examples/grpc/protoc-options/build.gradle
mgodave: Yeah, this was put behind an option because it would break code that treated these interfaces as functional, in cases where it was able to.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
mgodave: Sounds good. I hammered this together to gather feedback on the option. I'll clean things up.

mgodave: FWIW, `fullMethodName` is an internal concoction by the grpc library. I was able to track down the source and am able to replicate.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -1639,6 +1645,43 @@ private TypeSpec newServiceInterfaceSpec(final State state, final boolean blocki
...
+                final ClassName outClass = messageTypesMap.get(methodProto.getOutputType());
+                final String methodName = sanitizeIdentifier(methodProto.getName(), true);
+                final String serviceName = isNullOrEmpty(state.serviceProto.getName()) ?
+                        "" : state.serviceProto.getName();
+                final String fullMethodName = serviceName + "/" + methodName;
+                final int methodIndex = i;
+                interfaceSpecBuilder.addMethod(newRpcMethodSpec(inClass, outClass, methodName,
+                        methodProto.getClientStreaming(),
+                        methodProto.getServerStreaming(),
+                        !blocking ? EnumSet.of(INTERFACE) : EnumSet.of(INTERFACE, BLOCKING),
```
idelpivnitskiy, line 1663: The `BlockingTesterService` doesn't have default implementations for the following 2 method:
1.
```java
        @Override
        void testBiDiStream(GrpcServiceContext ctx, BlockingIterable<TestRequest> request, BlockingStreamingGrpcServerResponse<TestResponse> response) throws Exception;
```

2.
```java
        @Override
        void testResponseStream(GrpcServiceContext ctx, TestRequest request, BlockingStreamingGrpcServerResponse<TestResponse> response) throws Exception;
```

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -1639,6 +1645,43 @@ private TypeSpec newServiceInterfaceSpec(final State state, final boolean blocki
...
+            for (int i = 0; i < methodList.size(); i++) {
+                final MethodDescriptorProto methodProto = methodList.get(i);
+                final ClassName inClass = messageTypesMap.get(methodProto.getInputType());
+                final ClassName outClass = messageTypesMap.get(methodProto.getOutputType());
+                final String methodName = sanitizeIdentifier(methodProto.getName(), true);
+                final String serviceName = isNullOrEmpty(state.serviceProto.getName()) ?
+                        "" : state.serviceProto.getName();
+                final String fullMethodName = serviceName + "/" + methodName;
+                final int methodIndex = i;
+                interfaceSpecBuilder.addMethod(newRpcMethodSpec(inClass, outClass, methodName,
```
idelpivnitskiy, line 1660: On the other hand, `BlockingTesterService` has 2 methods that should be there only when `slipDeprecated=false`

1.
```java
        @Deprecated
        default void testBiDiStream(GrpcServiceContext ctx, BlockingIterable<TestRequest> request,
                GrpcPayloadWriter<TestResponse> responseWriter) throws Exception {
            throw new GrpcStatusException(new GrpcStatus(GrpcStatusCode.UNIMPLEMENTED, "Method Tester/testBiDiStream is unimplemented"));
        }
```

2.
```java
        @Deprecated
        default void testResponseStream(GrpcServiceContext ctx, TestRequest request,
                GrpcPayloadWriter<TestResponse> responseWriter) throws Exception {
            throw new GrpcStatusException(new GrpcStatus(GrpcStatusCode.UNIMPLEMENTED, "Method Tester/testResponseStream is unimplemented"));
        }
```

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -1639,6 +1645,43 @@ private TypeSpec newServiceInterfaceSpec(final State state, final boolean blocki
...
+                        methodProto.getClientStreaming(),
+                        methodProto.getServerStreaming(),
+                        !blocking ? EnumSet.of(INTERFACE) : EnumSet.of(INTERFACE, BLOCKING),
+                        printJavaDocs, (__, c) -> {
```
idelpivnitskiy, line 1664: All methods at service level (`TesterService` and `BlockingTesterService`) should have `@Override` annotation

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -1639,6 +1645,43 @@ private TypeSpec newServiceInterfaceSpec(final State state, final boolean blocki
                 .addStatement("return $L", blocking ? BLOCKING_METHOD_DESCRIPTORS : ASYNC_METHOD_DESCRIPTORS)
                 .build());
 
+        // generate default service methods
```
idelpivnitskiy, line 1648: Consider adding a test in `ProtocolCompatibilityTest` to make sure our default behavior is the same as grpc-java default behavior. In grpc-netty tests we can always use this new option `defaultServiceMethods=true`, bcz we will likely make it new default in the next "major" version.

idelpivnitskiy (changes requested):
New findings:

bryce-anderson (approved):
Take my review with a grain of salt, but it seems good to me.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -1639,6 +1645,57 @@ private TypeSpec newServiceInterfaceSpec(final State state, final boolean blocki
...
+                            } else {
+                                if (!skipDeprecated && methodProto.getServerStreaming()) {
+                                    spec.addStatement("$T.super.$L(ctx, request, response)",
+                                            rpcInterface.className, methodName);
+                                } else {
+                                    spec.addStatement("throw new $T(new $T($T.UNIMPLEMENTED, $L))",
+                                            GrpcStatusException, GrpcStatus, GrpcStatusCode, errorMessage);
+                                }
+                            }
+                            if (printJavaDocs) {
```
idelpivnitskiy, line 1682: I realized that in this case we actually don't need to generate javadoc because these methods are overrides and javadoc is automatically inherited from the `*Rpc` interfaces:
```patch
Index: servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java b/servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
--- a/servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java	(revision 9120c33cf95e6a1cf2cd8f482de3b51b108f13cc)
+++ b/servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java	(date 1732233569524)
@@ -1662,7 +1662,7 @@
                         methodProto.getClientStreaming(),
                         methodProto.getServerStreaming(),
                         !blocking ? EnumSet.of(INTERFACE) : EnumSet.of(INTERFACE, BLOCKING, SERVER_RESPONSE),
-                        printJavaDocs, (__, spec) -> {
+                        false, (__, spec) -> {
                             final String errorMessage = "\"Method " + methodPath + " is unimplemented\"";
                             spec.addModifiers(DEFAULT).addParameter(GrpcServiceContext, ctx);
                             spec.addAnnotation(Override.class);
@@ -1679,11 +1679,6 @@
                                             GrpcStatusException, GrpcStatus, GrpcStatusCode, errorMessage);
                                 }
                             }
-                            if (printJavaDocs) {
-                                extractJavaDocComments(state, methodIndex, spec);
-                                spec.addJavadoc(JAVADOC_PARAM + ctx +
-                                        " context associated with this service and request." + lineSeparator());
                             return spec;
                         });

```

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -1639,6 +1645,57 @@ private TypeSpec newServiceInterfaceSpec(final State state, final boolean blocki
...
+                                spec.addJavadoc(JAVADOC_PARAM + ctx +
+                                        " context associated with this service and request." + lineSeparator());
+                            }
+                            return spec;
+                        });
+
+                final boolean isDeprecated = methodSpec.annotations.contains(
+                        AnnotationSpec.builder(Deprecated.class).build());
+
+                if (!isDeprecated || !skipDeprecated) {
```
idelpivnitskiy, line 1693: `skipDeprecated=true` case works perfect for `defaultServiceMethods=false|true` 👍 

`skipDeprecated=false` has a problem: currently it overrides non-deprecated methods that already have a default impl (calling their deprecated variant). The deprecated methods don't. In result, users will still have compilation issues when they add a new "response streaming" method. Expectation is that we add default impl override only for those methods that don't already have a default impl in their parent interface.

Way to reproduce:
```java
public final class BlockingProtocOptionsServer {
    public static void main(String[] args) throws Exception {
        GrpcServers.forPort(8080)
                .listenAndAwait(new TesterSt.BlockingTesterService() { /* noop */ })
                .awaitShutdown();
    }
}
```
Result:
```
Class 'Anonymous class derived from BlockingTesterService' must implement abstract method 'testBiDiStream(GrpcServiceContext, BlockingIterable<TestRequest>, GrpcPayloadWriter<TestResponse>)' in 'BlockingTestBiDiStreamRpc'
```

If I change `BlockingTesterService` to `TesterService`, the compilation error goes away. It should be the same behavior for `BlockingTesterService`.

Current output:
```java
    public interface BlockingTesterService extends GrpcBindableService<TesterService>, BlockingTestRpc, BlockingTestBiDiStreamRpc, BlockingTestResponseStreamRpc, BlockingTestRequestStreamRpc {
        /**
         * Makes a {@link ServiceFactory} bound to this instance implementing {@link BlockingTesterService}
         */
        @Override
        default ServiceFactory bindService() {
            return new ServiceFactory(this);
        }

        @Override
        default Collection<MethodDescriptor<?, ?>> methodDescriptors() {
            return BLOCKING_METHOD_DESCRIPTORS;
        }

        @Override
        default TestResponse test(GrpcServiceContext ctx, TestRequest request) throws Exception {
            throw new GrpcStatusException(new GrpcStatus(GrpcStatusCode.UNIMPLEMENTED, "Method grpc.netty.Tester/test is unimplemented"));
        }

        @Override
        default void testBiDiStream(GrpcServiceContext ctx, BlockingIterable<TestRequest> request,
                BlockingStreamingGrpcServerResponse<TestResponse> response) throws Exception {
            BlockingTestBiDiStreamRpc.super.testBiDiStream(ctx, request, response);
        }

        @Override
        default void testResponseStream(GrpcServiceContext ctx, TestRequest request,
                BlockingStreamingGrpcServerResponse<TestResponse> response) throws Exception {
            BlockingTestResponseStreamRpc.super.testResponseStream(ctx, request, response);
        }

        @Override
        default TestResponse testRequestStream(GrpcServiceContext ctx,
                BlockingIterable<TestRequest> request) throws Exception {
            throw new GrpcStatusException(new GrpcStatus(GrpcStatusCode.UNIMPLEMENTED, "Method grpc.netty.Tester/testRequestStream is unimplemented"));
        }
    }
```

Expected output:
```java
    public interface BlockingTesterService extends GrpcBindableService<TesterService>, BlockingTestRpc, BlockingTestBiDiStreamRpc, BlockingTestResponseStreamRpc, BlockingTestRequestStreamRpc {
        /**
         * Makes a {@link ServiceFactory} bound to this instance implementing {@link BlockingTesterService}
         */
        @Override
        default ServiceFactory bindService() {
            return new ServiceFactory(this);
        }

        @Override
        default Collection<MethodDescriptor<?, ?>> methodDescriptors() {
            return BLOCKING_METHOD_DESCRIPTORS;
        }

        @Override
        default TestResponse test(GrpcServiceContext ctx, TestRequest request) throws Exception {
            throw new GrpcStatusException(new GrpcStatus(GrpcStatusCode.UNIMPLEMENTED, "Method grpc.netty.Tester/test is unimplemented"));
        }

        @Override
        @Deprecated
        default void testBiDiStream(GrpcServiceContext ctx, BlockingIterable<TestRequest> request,
                                    GrpcPayloadWriter<TestResponse> responseWriter) throws Exception {
            throw new GrpcStatusException(new GrpcStatus(GrpcStatusCode.UNIMPLEMENTED, "Method grpc.netty.Tester/testBiDiStream is unimplemented"));
        }

        @Override
        @Deprecated
        default void testResponseStream(GrpcServiceContext ctx, TestRequest request,
                                        GrpcPayloadWriter<TestResponse> responseWriter) throws Exception {
            throw new GrpcStatusException(new GrpcStatus(GrpcStatusCode.UNIMPLEMENTED, "Method grpc.netty.Tester/testResponseStream is unimplemented"));
        }

        @Override
        default TestResponse testRequestStream(GrpcServiceContext ctx,
                BlockingIterable<TestRequest> request) throws Exception {
            throw new GrpcStatusException(new GrpcStatus(GrpcStatusCode.UNIMPLEMENTED, "Method grpc.netty.Tester/testRequestStream is unimplemented"));
        }
    }
```

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
```diff
@@ -1639,6 +1645,57 @@ private TypeSpec newServiceInterfaceSpec(final State state, final boolean blocki
...
+                            final String errorMessage = "\"Method " + methodPath + " is unimplemented\"";
+                            spec.addModifiers(DEFAULT).addParameter(GrpcServiceContext, ctx);
+                            spec.addAnnotation(Override.class);
+                            if (!blocking) {
+                                final ClassName returnType = methodProto.getServerStreaming() ? Publisher : Single;
+                                spec.addStatement("return $T.failed(new $T(new $T($T.UNIMPLEMENTED, $L)))",
+                                        returnType, GrpcStatusException, GrpcStatus, GrpcStatusCode, errorMessage);
+                            } else {
+                                if (!skipDeprecated && methodProto.getServerStreaming()) {
+                                    spec.addStatement("$T.super.$L(ctx, request, response)",
```
idelpivnitskiy, line 1675: We should always throw `UNIMPLEMENTED` exception instead of calling `super`.

Instead of
```java
        @Override
        default void testBiDiStream(GrpcServiceContext ctx, BlockingIterable<TestRequest> request,
                BlockingStreamingGrpcServerResponse<TestResponse> response) throws Exception {
            BlockingTestBiDiStreamRpc.super.testBiDiStream(ctx, request, response);
        }

        @Override
        default void testResponseStream(GrpcServiceContext ctx, TestRequest request,
                BlockingStreamingGrpcServerResponse<TestResponse> response) throws Exception {
            BlockingTestResponseStreamRpc.super.testResponseStream(ctx, request, response);
        }
```

We should generate:
```java
        @Override
        @Deprecated
        default void testBiDiStream(GrpcServiceContext ctx, BlockingIterable<TestRequest> request,
                                    GrpcPayloadWriter<TestResponse> responseWriter) throws Exception {
            throw new GrpcStatusException(new GrpcStatus(GrpcStatusCode.UNIMPLEMENTED, "Method grpc.netty.Tester/testBiDiStream is unimplemented"));
        }

        @Override
        @Deprecated
        default void testResponseStream(GrpcServiceContext ctx, TestRequest request,
                                        GrpcPayloadWriter<TestResponse> responseWriter) throws Exception {
            throw new GrpcStatusException(new GrpcStatus(GrpcStatusCode.UNIMPLEMENTED, "Method grpc.netty.Tester/testResponseStream is unimplemented"));
        }
```

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
```diff
@@ -610,6 +611,50 @@ void clientH2ReturnStatus() throws Exception {
         }
     }
 
+    @Test
+    void grpcJavaToServiceTalkUnimplementedService() throws Exception {
+        try (TestServerContext server = serviceTalkServerBlocking(ErrorMode.STATUS, false, null, null);
+             CompatClient client = grpcJavaClient(server.listenAddress(), null, false, null)) {
+            final Single<CompatResponse> response =
+                    client.unimplementedServerCall(CompatRequest.newBuilder().setId(1).build());
```
idelpivnitskiy, line 619: Instead of adding a new `unimplementedServerCall` in the proto, consider adding service parameter for `serviceTalkServerBlocking`, `serviceTalkServer`, and `grpcJavaServer`. There you can pass an unimplemented service like `new BlockingCompatService() { }`. That way you can validate all 4 API flavors in every test method.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
```diff
@@ -610,6 +611,50 @@ void clientH2ReturnStatus() throws Exception {
...
+            validateGrpcErrorInResponse(response.toFuture(), false, UNIMPLEMENTED,
+                    "Method grpc.netty.Compat/unimplementedServerCall is unimplemented");
+        }
+    }
+
+    @Test
+    void serviceTalkToServiceTalkUnimplementedService() throws Exception {
+        try (TestServerContext server = serviceTalkServerBlocking(ErrorMode.NONE, false, null, null);
```
idelpivnitskiy, line 649: For `grpcJavaToServiceTalkUnimplementedService` and `serviceTalkToServiceTalkUnimplementedService` consider making them `@ParameterizedTest` to also run with `serviceTalkServer(..., new CompatService() { })`

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
mgodave: I would argue that this is not broken, it works as it should with the combination of flags. I definitely see your point about not redefining a default method when the base interface already has one. That is potentially easier said than done with javapoet.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
mgodave: I see the error now with a server streaming response, I'll replicate the full combination of methods as unimplemented.

mgodave: I see what is going on here. I need to somehow grab a hold of the super interface specs so I can query them for any abstract methods that I can then implement... I have no idea how to do this at the moment...

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
idelpivnitskiy: It will be ok to override both variants (deprecated and not) of `testBiDiStream` and `testResponseStream` and always throw `UNIMPLEMENTED` (no call to super). Then users will be able to override only non-deprecated variant and won't see any compilation problems asking to implement the deprecated one.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
idelpivnitskiy: As replied in another thread, it will be ok to implement just all `*Rpc` methods and don't bother with skipping those that already have a default impl.

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
mgodave: Always throwing will still break the default behavior, especially in cases where the user has implemented the `GrpcResponseWriter` variant.

idelpivnitskiy: Looks like a reply to a wrong thread, but you are right. We should still skip those that already have a default.

servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
idelpivnitskiy: Figured it 🎉 
I reverted your last 2 commits to go back to the original approach you had and then made the following changes:
```patch
Index: servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java b/servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java
--- a/servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java	(revision af1dfe3dc11881a0cf5cd0b18f952d8cda2c5bad)
+++ b/servicetalk-grpc-protoc/src/main/java/io/servicetalk/grpc/protoc/Generator.java	(date 1732316010356)
@@ -1650,49 +1650,33 @@
             final List<RpcInterface> interfaces = state.serviceRpcInterfaces.stream()
                     .filter(intf -> intf.blocking == blocking)
                     .collect(toList());
-            for (int i = 0; i < interfaces.size(); ++i) {
-                final RpcInterface rpcInterface = interfaces.get(i);
+            for (final RpcInterface rpcInterface : interfaces) {
                 final MethodDescriptorProto methodProto = rpcInterface.methodProto;
                 final ClassName inClass = messageTypesMap.get(methodProto.getInputType());
                 final ClassName outClass = messageTypesMap.get(methodProto.getOutputType());
                 final String methodName = sanitizeIdentifier(methodProto.getName(), true);
                 final String methodPath = context.methodPath(state.serviceProto, methodProto).substring(1);
-                final int methodIndex = i;
                 final MethodSpec methodSpec = newRpcMethodSpec(inClass, outClass, methodName,
                         methodProto.getClientStreaming(),
                         methodProto.getServerStreaming(),
-                        !blocking ? EnumSet.of(INTERFACE) : EnumSet.of(INTERFACE, BLOCKING, SERVER_RESPONSE),
-                        printJavaDocs, (__, spec) -> {
+                        !blocking ? EnumSet.of(INTERFACE) : (skipDeprecated ?
+                                EnumSet.of(INTERFACE, BLOCKING, SERVER_RESPONSE) : EnumSet.of(INTERFACE, BLOCKING)),
+                        false, (__, spec) -> {
+                            spec.addAnnotation(Override.class);
+                            spec.addModifiers(DEFAULT).addParameter(GrpcServiceContext, ctx);
                             final String errorMessage = "\"Method " + methodPath + " is unimplemented\"";
-                            spec.addModifiers(DEFAULT).addParameter(GrpcServiceContext, ctx);
-                            spec.addAnnotation(Override.class);
                             if (!blocking) {
                                 final ClassName returnType = methodProto.getServerStreaming() ? Publisher : Single;
                                 spec.addStatement("return $T.failed(new $T(new $T($T.UNIMPLEMENTED, $L)))",
                                         returnType, GrpcStatusException, GrpcStatus, GrpcStatusCode, errorMessage);
                             } else {
-                                if (!skipDeprecated && methodProto.getServerStreaming()) {
-                                    spec.addStatement("$T.super.$L(ctx, request, response)",
-                                            rpcInterface.className, methodName);
-                                    spec.addStatement("throw new $T(new $T($T.UNIMPLEMENTED, $L))",
-                                            GrpcStatusException, GrpcStatus, GrpcStatusCode, errorMessage);
+                                spec.addStatement("throw new $T(new $T($T.UNIMPLEMENTED, $L))",
+                                        GrpcStatusException, GrpcStatus, GrpcStatusCode, errorMessage);
                             }
-                            if (printJavaDocs) {
-                                extractJavaDocComments(state, methodIndex, spec);
-                                spec.addJavadoc(JAVADOC_PARAM + ctx +
-                                        " context associated with this service and request." + lineSeparator());
                             return spec;
                         });
 
-                final boolean isDeprecated = methodSpec.annotations.contains(
-                        AnnotationSpec.builder(Deprecated.class).build());
-
-                if (!isDeprecated || !skipDeprecated) {
-                    interfaceSpecBuilder.addMethod(methodSpec);
-                }
+                interfaceSpecBuilder.addMethod(methodSpec);
             }
         }
 
```
Now all default methods look as expected. We just need to add tests to cover them all.

idelpivnitskiy (approved):
Pushed minor style corrections into your branch: https://github.com/apple/servicetalk/pull/3110/commits/80b53a49ddad9c21b4845c3f03931407a27741b0
Everything looks good, great work!

PR #3111, opened 2024-11-18
Server-side `AsyncContext` initialized in lifecycle observer is lost

idelpivnitskiy:
Motivation:

If users put any data into `AsyncContext` inside `HttpLifecycleObserver` configured via `HttpServerBuilder.lifecycleObserver(...)` it won't be visible for filters and service because
`ClearAsyncContextHttpServiceFilter` is appended after `HttpLifecycleObserverServiceFilter` inside `applyInternalFilters`.

Modifications:
1. Move `ClearAsyncContextHttpServiceFilter` from the beginning of `noOffloadServiceFilters` to `applyInternalFilters`. However, this move discovered another bug in path that handles
`noOffloadServiceFilters.isEmpty()` case because it never adds `OffloadingFilter`.
2. Refactor `listenForService` method to always make a copy of filters lists, then prepend/append internal filters to those copies in easy to read/understand way, and construct a single filters `Stream` for `buildService` method.
3. Refactor `OffloadingFilter` to act as a regular filter factory instead of taking all further `serviceFilters` as pre-built factory.
4. Rename `ClearAsyncContextHttpServiceFilter` singleton instance.
5. Fix another bug when `OptionalSslNegotiator` binder takes a raw service instead of `filteredService` (user-defined filters were never applied for this path).
6. Modify `AbstractHttpServiceAsyncContextTest` to test `AsyncContext` propagation from lifecycle observer and non-offloading filters.
7. Modify `AbstractHttpServiceAsyncContextTest` to test `AsyncContext` initialized by early/late connection acceptor is not visible at request path.
8. Add `BlockingHttpServiceAsyncContextTest` and `HttpServiceAsyncContextTest` to make sure `AsyncContext` propagation for blocking/async aggregated services is also tested.
9. Use `ParameterizedTest` where possible.

Results:

1. `AsyncContext` initialized inside server's lifecycle observer is visible through filter chain and services.
2. Construction of server-side filter chain is consistent and sequential.
3. Filters are applied for servers with optional TLS.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
@@ -73,6 +73,7 @@ import static io.servicetalk.http.api.HttpExecutionStrategies.defaultStrategy;
 import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
 import static io.servicetalk.http.netty.StrategyInfluencerAwareConversions.toConditionalServiceFilterFactory;
 import static io.servicetalk.transport.api.ConnectionAcceptor.ACCEPT_ALL;
+import static java.util.Collections.unmodifiableList;
 import static java.util.Objects.requireNonNull;
 
 final class DefaultHttpServerBuilder implements HttpServerBuilder {
@@ -96,14 +97,7 @@ final class DefaultHttpServerBuilder implements HttpServerBuilder {
 
     // Do not use this ctor directly, HttpServers is the entry point for creating a new builder.
     DefaultHttpServerBuilder(SocketAddress address) {
-        appendNonOffloadingServiceFilter(ClearAsyncContextHttpServiceFilter.CLEAR_ASYNC_CONTEXT_HTTP_SERVICE_FILTER);
         this.address = address;
-    }
-
-    private static StreamingHttpServiceFilterFactory buildFactory(List<StreamingHttpServiceFilterFactory> filters) {
-        return filters.stream()
-                .reduce((prev, filter) -> service -> prev.create(filter.create(service)))
-                .orElse(StreamingHttpServiceFilter::new); // unfortunate that we need extra layer
     }
 
     private static StreamingHttpService buildService(Stream<StreamingHttpServiceFilterFactory> filters,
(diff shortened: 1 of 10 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -300,28 +300,65 @@ public HttpServerBuilder bufferAllocator(final BufferAllocator allocator) {
...
+        // Append internal filters:
+        // 1. The watchdog sits at the very beginning of the response flow (the end of the filter pipeline) so that any
+        // payload coming from the service is ensured to be tracked before subsequent filters get a chance to drop it
+        // without being accounted for.
+        filters.add(HttpMessageDiscardWatchdogServiceFilter.INSTANCE);
+
+        return unmodifiableList(filters);
+    }
+
+    private List<StreamingHttpServiceFilterFactory> initNonOffloadsServiceFilters(
```
daschl, line 345: I would suggest renaming this method to something that includes the word `default` so its understood that those are the default ones always added

daschl (approved):
Change looks good to me, I think it is an improvement also in follow-ability through the filter bootstrapping of the server. Just one nit, feel free to ignore.

One separate thing @idelpivnitskiy - since it changes the code quite a bit, just wanted to make sure that it is double checked that the new order is good and nothing ended up accidentally in the wrong spot of the filter chain. Doesn't look like it to me, just checking.

idelpivnitskiy:
> just wanted to make sure that it is double checked that the new order is good and nothing ended up accidentally in the wrong spot of the filter chain. Doesn't look like it to me, just checking.

There are existing tests to validate most of the features and ordering:
- `HttpServerFilterOrderTest` to test the ordering of user-added filters;
- `ServerEffectiveStrategyTest` to test `OffloadingFilter` (it started to fail when I moved `ClearAsyncContextHttpServiceFilter`);
- `AbstractHttpServiceAsyncContextTest` to test `ClearAsyncContextHttpServiceFilter`;
- `HttpLifecycleObserverTest` to test observer;
- `ConnectionCloseHeaderHandlingTest` to test `KeepAliveServiceFilter`;
- There are plenty if tests with unhandled exceptions for `HttpExceptionMapperServiceFilter`;
- `HttpMessageDiscardWatchdogServiceFilterTest` (reenabled locally to test);

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
idelpivnitskiy: Because both methods do 2 jobs (add default filters and copy user filters to avoid side-effects), I couldn't fit "default" nicely in a new name. "init" is abstract enough, methods are short and have extensive comments about what they do.

PR #3112, opened 2024-11-18
`Single`/`Completable` `subscribe` overloads to handle failures

idelpivnitskiy:
Motivation:

Currently, users can consume the result or execute a `Runnable` for successful termination of a `Single`/`Completable`. Failures will be logged at debug level. Users can only use `whenOnError` operator to intercept the error, but it may result in duplicated logging or users simply forget to handle the error case.

Modifications:

- Add `Single.subscribe(Consumer<? super T>, Consumer<? super Throwable>)`;
- Add `Completable.subscribe(Runnable, Consumer<? super Throwable>)`;
- Modify `SimpleSingleSubscriber/SimpleCompletableSubscriber` to handle error when `errorConsumer` is provided.

Result:

Users can easily handle error cases when they subscribe to `Single` or `Completable`.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
@@ -1879,16 +1879,30 @@ public abstract class Single<T> {
     }
 
     /**
-     * Subscribe to this {@link Single}, emits the result to the passed {@link Consumer} and log any
-     * {@link Subscriber#onError(Throwable)}.
+     * Subscribe to this {@link Single}, emit the result to the passed {@link Consumer} and log any
+     * {@link Subscriber#onError(Throwable)} at debug level.
      *
-     * @param resultConsumer {@link Consumer} to accept the result of this {@link Single}.
-     *
+     * @param resultConsumer {@link Consumer} to accept the result of this {@link Single} if it succeeds.
      * @return {@link Cancellable} used to invoke {@link Cancellable#cancel()} on the parameter of
      * {@link Subscriber#onSubscribe(Cancellable)} for this {@link Single}.
      */
     public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
         SimpleSingleSubscriber<T> subscriber = new SimpleSingleSubscriber<>(resultConsumer);
+        subscribeInternal(subscriber);
+        return subscriber;
+    }
+
+    /**
+     * Subscribe to this {@link Single}, emit the result or error to one of the passed {@link Consumer}s.
+     *
+     * @param resultConsumer {@link Consumer} to accept the result of this {@link Single} if it succeeds.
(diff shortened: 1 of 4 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1893,6 +1892,21 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
...
 
+    /**
+     * Subscribe to this {@link Single}, emit the result or error to one of the passed {@link Consumer}s.
+     *
+     * @param resultConsumer {@link Consumer} to accept the result of this {@link Single} if it succeeds.
+     * @param errorConsumer {@link Consumer} to accept the error of this {@link Single} if it fails.
+     * @return {@link Cancellable} used to invoke {@link Cancellable#cancel()} on the parameter of
+     * {@link Subscriber#onSubscribe(Cancellable)} for this {@link Single}.
+     */
+    public final Cancellable subscribe(Consumer<? super T> resultConsumer,
```
idelpivnitskiy, line 1903: 2 consumers vs bi-consumer: I picked the way that exists in both rxjava (for [Single](https://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/core/Single.html#subscribe-io.reactivex.rxjava3.functions.Consumer-io.reactivex.rxjava3.functions.Consumer-) and [Completable](https://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/core/Completable.html#subscribe-io.reactivex.rxjava3.functions.Action-io.reactivex.rxjava3.functions.Consumer-)) and reactor (for [Mono](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#subscribe-java.util.function.Consumer-java.util.function.Consumer-)) to be as close as possible to other implementations. We can add `BiConsumer` later if there is an ask.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
```diff
@@ -1893,6 +1892,21 @@ public final Cancellable subscribe(Consumer<? super T> resultConsumer) {
...
+                                       Consumer<? super Throwable> errorConsumer) {
```
bryce-anderson, line 1904: Do you want to enforce that this helper doesn't receive a null `errorConsumer`? If not, we might as well funnel the `subscribe(Consumer<? super T>)` variant through this method.
The same question applies to `Completable`.

bryce-anderson (approved):
I personally prefer the bifunction version, but I agree that following precedent is probably more important.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Single.java
idelpivnitskiy: Good idea, changed ctors to use `requireNonNull` when we expect `errorConsumer`

PR #3121, opened 2024-11-21
Ignore GitHub Actions for non-relevant files

idelpivnitskiy:
Motivation:

Currently, GitHub Actions workflows run on every change in the PR, even if it's known the file won't affect the output.

Modifications:
- For each workflow, ignore files that won't affect its result.

Result:

Less workflows to run for small PRs.


Scottmitch (approved):
if changes don't take effect until after merge, lets merge

PR #3128, opened 2024-12-05
Do not cancel subscription on BlockingIterable#hasNext(long, TimeUnit)

daschl:
This makes sure that if this method, or indirectly `next(long, TimeUnit)` is called and times out, the caller is able to retry the operation and also the upstream source will not be cancelled.

In the context of a `BlockingStreaming` server, this means that if a timeout is thrown on the incoming request, the outgoing response can still be modified since the underlying socket will not be immediately closed.

It also aligns the semantics with `Single#toFuture` where a blocking get with a timeout on the future also does not cancel the upstream `Single`.

A (temporary) system property is introduced which allows to fall back to the old behavior should incompatibilities be discovered in the wild.

A note for the reader who wonders how to close the subscription now: the close() method always did cancel the subscription and continues to do so. To make sure `forEach` also adheres to the iterator contract it uses it is now wrapped in `try-with-resource`.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherAsBlockingIterable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherAsBlockingIterable.java
@@ -76,8 +76,8 @@ final class PublisherAsBlockingIterable<T> implements BlockingIterable<T> {
          * Allows to re-enable cancelling the subscription on {@link #hasNext(long, TimeUnit)} timeout. This flag
          * will be removed after a couple releases and no issues identified with the new behavior.
          */
-        private static final boolean CANCEL_SUBSCRIPTION_ON_HAS_NEXT_TIMEOUT = Boolean.parseBoolean(
-                System.getProperty("io.servicetalk.concurrent.api.cancelSubscriptionOnHasNextTimeout", "false"));
+        private static final boolean CANCEL_SUBSCRIPTION_ON_HAS_NEXT_TIMEOUT = Boolean
+                .getBoolean("io.servicetalk.concurrent.api.cancelSubscriptionOnHasNextTimeout");
         private static final Logger LOGGER = LoggerFactory.getLogger(SubscriberAndIterator.class);
         private static final Object CANCELLED_SIGNAL = new Object();
         private static final TerminalNotification COMPLETE_NOTIFICATION = complete();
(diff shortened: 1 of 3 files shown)


servicetalk-concurrent/src/main/java/io/servicetalk/concurrent/BlockingIterable.java
```diff
@@ -64,6 +64,7 @@ public interface BlockingIterable<T> extends CloseableIterable<T> {
     default void forEach(Consumer<? super T> action, LongSupplier timeoutSupplier, TimeUnit unit)
             throws TimeoutException {
         requireNonNull(action);
+        // TODO: should we here try/catch with close explicitly now?
```
daschl, line 67: @idelpivnitskiy wdyt 🤔

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/PublisherAsBlockingIterable.java
```diff
@@ -72,6 +72,12 @@ public BlockingIterator<T> iterator() {
     }
 
     private static final class SubscriberAndIterator<T> implements Subscriber<T>, BlockingIterator<T> {
+        /**
+         * Allows to re-enable cancelling the subscription on {@link #hasNext(long, TimeUnit)} timeout. This flag
+         * will be removed after a couple releases and no issues identified with the new behavior.
+         */
+        private static final boolean CANCEL_SUBSCRIPTION_ON_HAS_NEXT_TIMEOUT = Boolean.parseBoolean(
+                System.getProperty("io.servicetalk.concurrent.api.cancelSubscriptionOnHasNextTimeout", "false"));
```
idelpivnitskiy, line 80: Can be simplified as `Boolean.getBoolean(name)`

servicetalk-concurrent/src/main/java/io/servicetalk/concurrent/BlockingIterable.java
idelpivnitskiy: Discussed offline: bcz in this case we own `Iterator` instance, we should keep it within try-with-resources block for all `forEach` overloads, even the one without timeout args.

idelpivnitskiy (approved):

mgodave (approved):
You presented this offline, I think this is the correct behavior.

PR #3140, opened 2024-12-14
loadbalancer-experimental: adjust `LoadBalancerObserver` API

idelpivnitskiy:
Modifications:

- Remove default implementation for `onHostSetChanged(Collection)`;
- Change `onNoHostsAvailable()` to `onNoAvailableHostException(NoAvailableHostException)`;
- Rename `onNoActiveHostsAvailable` to `onNoActiveHostException`;

### DIFF
--- a/servicetalk-loadbalancer-experimental-provider/src/main/java/io/servicetalk/loadbalancer/experimental/DefaultLoadBalancerObserver.java
+++ b/servicetalk-loadbalancer-experimental-provider/src/main/java/io/servicetalk/loadbalancer/experimental/DefaultLoadBalancerObserver.java
@@ -16,6 +16,7 @@
 package io.servicetalk.loadbalancer.experimental;
 
 import io.servicetalk.client.api.NoActiveHostException;
+import io.servicetalk.client.api.NoAvailableHostException;
 import io.servicetalk.client.api.ServiceDiscovererEvent;
 import io.servicetalk.loadbalancer.LoadBalancerObserver;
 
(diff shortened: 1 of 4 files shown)


servicetalk-loadbalancer-experimental-provider/src/main/java/io/servicetalk/loadbalancer/experimental/DefaultLoadBalancerObserver.java
```diff
@@ -42,23 +43,13 @@ public HostObserver hostObserver(Object resolvedAddress) {
         return new HostObserverImpl(resolvedAddress);
     }
 
-    @Override
-    public void onNoHostsAvailable() {
-        LOGGER.debug("{}- onNoHostsAvailable()", lbDescription);
-    }
-
     @Override
     public void onServiceDiscoveryEvent(Collection<? extends ServiceDiscovererEvent<?>> events, int oldHostSetSize,
```
idelpivnitskiy, line 47: Moving discussion to this PR:

I see that `onServiceDiscoveryEvent` is invoked after `hostObserver` callback because we need to calculate `nextHosts.size()` first. However, it’s a bit harder to correlate `hostObserver` events to SD-event if it happens later. Will it make sense if we create another nested interface, like `ServiceDiscoveryEventObserver`, which will report a Collection of event, then inside that interface we will invoke `hostObserver` callback, then there will be an event to report `(int oldHostSetSize, int newHostSetSize)` and/or `onHostSetChanged`?

Alternatively, we can keep the flat structure, but remove `(int oldHostSetSize, int newHostSetSize)` from `onServiceDiscoveryEvent`, invoke it as soon as we receive an event, then at the end, after all `hostObserver` occurrences reported we can invoke either `onHostSetChanged` with before/after collection or a new method `onHostSetNotChanged` with `int`s only?

bryce-anderson: You're right: it would be good to preserve the ordering of 'cause -> event'.

Let me think on it a bit, but I'm wondering if we want a generic notion of a `ServiceDiscovererEventObserver` that isn't tied to a specific implementation. We have a `DnsServiceDiscovererObserver` but it looks like it has callbacks for internal details but not what is ultimately emitted.

idelpivnitskiy: I still think it makes sense in LB to show what it received bcz the output of SD may change on the way. We already have `ServiceDiscovererEventsCache` and users may apply their own filtering (by wrapping SD). 
It seems like having a flat structure where `onServiceDiscoveryEvent` is the first one might be less tight to a specific impl bcz we know that in order for LB to change the state it should first receive an event from SD.

bryce-anderson (approved):

servicetalk-loadbalancer-experimental-provider/src/main/java/io/servicetalk/loadbalancer/experimental/DefaultLoadBalancerObserver.java
idelpivnitskiy: Discussed with @bryce-anderson offline, he will follow-up with a new PR for these adjustments

PR #3148, opened 2024-12-19
wip: trying to debug why we have an extra data frame in the client streaming st case

mgodave:
### Problem

The blocking servicetalk server sends an extra empty data frame on failure even when there aren't any trailers. This causes some issues for the grpc-java client.

### Changes
I added a filter and a context key to signal when we have a case where we can avoid sending the extra empty data frame. In the "BROKEN AFTER CHANGE" case below I expect not to see the extra data frame as I am in the previous case but I assume I'm missing a detail about how ST works to be able to be effective in debugging this rn. 

### Results
Before Change:
```
OUTBOUND HEADERS: streamId=3 headers=GrpcHttp2OutboundHeaders[:authority: 127.0.0.1:62451, :path: /grpc.netty.Compat/clientStreamingCall, :method: POST, :scheme: http, content-type: application/grpc, te: trailers, user-agent: grpc-java-netty/1.64.1, grpc-accept-encoding: gzip] streamDependency=0 weight=16 exclusive=false padding=0 endStream=false
OUTBOUND DATA: streamId=3 padding=0 endStream=false length=7 bytes=00000000020801
OUTBOUND DATA: streamId=3 padding=0 endStream=true length=0 bytes=
INBOUND HEADERS: streamId=3 headers=GrpcHttp2ResponseHeaders[:status: 200, server: servicetalk-grpc/, content-type: application/grpc+proto, grpc-status: 12, grpc-message: Method grpc.netty.Compat/clientStreamingCall is unimplemented] padding=0 endStream=false
INBOUND DATA: streamId=3 padding=0 endStream=true length=0 bytes=
```

After Change:
```
### WORKING AFTER CHANGE (grpc -> ST blocking)
OUTBOUND HEADERS: streamId=3 headers=GrpcHttp2OutboundHeaders[:authority: 127.0.0.1:62421, :path: /grpc.netty.Compat/clientStreamingCall, :method: POST, :scheme: http, content-type: application/grpc, te: trailers, user-agent: grpc-java-netty/1.64.1, grpc-accept-encoding: gzip] streamDependency=0 weight=16 exclusive=false padding=0 endStream=false
OUTBOUND DATA: streamId=3 padding=0 endStream=false length=7 bytes=00000000020801
OUTBOUND DATA: streamId=3 padding=0 endStream=true length=0 bytes=
INBOUND HEADERS: streamId=3 headers=GrpcHttp2ResponseHeaders[:status: 200, server: servicetalk-grpc/, content-type: application/grpc+proto, grpc-status: 12, grpc-message: Method grpc.netty.Compat/clientStreamingCall is unimplemented, content-length: 0] padding=0 endStream=true
```


servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpResponse.java
```diff
@@ -37,7 +37,7 @@ final class DefaultHttpResponse extends AbstractDelegatingHttpResponse
                         @Nullable final HttpHeaders trailers) {
         super(original);
         this.payloadBody = payloadBody;
-        this.trailers = trailers;
+        this.trailers = (trailers != null && trailers.isEmpty()) ? null : trailers;
```
mgodave, line 40: This treats and empty set of trailers as the same as `null`. By doing this on construction we are able to let `toStreamingResponse()` and `trailers()` behave as normal.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
```diff
@@ -68,6 +68,9 @@ public final class HttpContextKeys {
     public static final Key<Boolean> HTTP_FORCE_NEW_CONNECTION =
             newKey("HTTP_FORCE_NEW_CONNECTION", Boolean.class);
 
+    public static final Key<Boolean> HTTP_OPTIMIZE_ERROR_STREAM =
```
mgodave, line 71: I probably need a better name for this

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcTrailersOptimizationServiceFilter.java
```diff
@@ -0,0 +1,64 @@
...
+    public StreamingHttpServiceFilter create(StreamingHttpService service) {
+        return new StreamingHttpServiceFilter(service) {
+            @Override
+            public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
+                                                        final StreamingHttpRequest request,
+                                                        final StreamingHttpResponseFactory responseFactory) {
+                return super.handle(ctx, request, responseFactory).flatMap(response -> {
+                    Single<StreamingHttpResponse> mappedResponse;
+                    if (Boolean.TRUE.equals(response.context().get(HTTP_OPTIMIZE_ERROR_STREAM))) {
+                        mappedResponse = response.payloadBody(Publisher.empty()).toResponse()
```
idelpivnitskiy, line 49: You can not override the payload body bcz if you discard previous publisher you interrupt RS chain and if there are any listeners on the previous publisher those won't be invoked.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpResponse.java
idelpivnitskiy: We can not drop pre-allocated trailers object on every aggregation bcz users can do aggregation in place inside their controller and still use old reference to `HttpHeaders` to modify trailers state

mgodave: ugh, I was afraid of this...

mgodave:
Closed in favor or 3152

PR #3151, opened 2024-12-19
`BlockingStreamingHttpService`: drop trailers if users didn't create any

idelpivnitskiy:
Motivation:

Behavior was discovered as part of debugging #3148.
Currently, `BlockingStreamingHttpService` allocates trailers upfront and concatenates them after the payload body when protocol allows. In result, aggregation of `BlockingStreamingHttpResponse` does not lead to a single HTTP/2 frame because we always expect to receive trailers. However, adding trailers after users close payload writer is race. There are no guarantees that new trailers will be written to the network after close. Therefore, we can inspect the state after close and decide if we need to append trailers or not.

Modifications:

- Use `scanWithMapper` inside `BlockingStreamingHttpService` instead of always concatenating payload body with trailers;
- Create trailers on demand inside `BufferHttpPayloadWriter`, only when users touch them;
- Clarify behavior in `HttpPayloadWriter`'s javadoc for trailers-related methods;
- Enhance `BlockingStreamingToStreamingServiceTest` to assert new expectations;

Result:

In `BlockingStreamingHttpService`, trailers are allocated and attached only if users touch trailers before closing `HttpPayloadWriter`.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
@@ -21,12 +21,16 @@ import io.servicetalk.concurrent.PublisherSource.Subscription;
 import io.servicetalk.concurrent.SingleSource.Subscriber;
 import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.Publisher;
+import io.servicetalk.concurrent.api.ScanMapper;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.concurrent.api.internal.ConnectablePayloadWriter;
 import io.servicetalk.concurrent.internal.ThreadInterruptingCancellable;
 
 import java.io.IOException;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+import javax.annotation.Nullable;
 
 import static io.servicetalk.concurrent.api.Processors.newCompletableProcessor;
 import static io.servicetalk.concurrent.api.SourceAdapters.fromSource;
@@ -76,7 +80,7 @@ final class BlockingStreamingToStreamingService extends AbstractServiceAdapterHo
                 // still be propagated.
                 final Processor exceptionProcessor = newCompletableProcessor();
                 final BufferHttpPayloadWriter payloadWriter = new BufferHttpPayloadWriter(
-                        ctx.headersFactory().newTrailers());
+                        () -> ctx.headersFactory().newTrailers());
                 DefaultBlockingStreamingHttpServerResponse response = null;
                 try {
                     final Consumer<DefaultHttpResponseMetaData> sendMeta = (metaData) -> {
@@ -102,7 +106,7 @@ final class BlockingStreamingToStreamingService extends AbstractServiceAdapterHo
                             Publisher<Object> messageBody = fromSource(exceptionProcessor)
                                     .merge(payloadWriter.connect());
                             if (addTrailers) {
-                                messageBody = messageBody.concat(succeeded(payloadWriter.trailers()));
+                                messageBody = messageBody.scanWithMapper(() -> new TrailersMapper(payloadWriter));
                             }
                             messageBody = messageBody.beforeSubscription(() -> new Subscription() {
                                 @Override
@@ -169,10 +173,12 @@ final class BlockingStreamingToStreamingService extends AbstractServiceAdapterHo
 
     private static final class BufferHttpPayloadWriter implements HttpPayloadWriter<Buffer> {
         private final ConnectablePayloadWriter<Buffer> payloadWriter = new ConnectablePayloadWriter<>();
-        private final HttpHeaders trailers;
+        private final Supplier<HttpHeaders> trailersFactory;
+        @Nullable
+        private HttpHeaders trailers;
 
-        BufferHttpPayloadWriter(final HttpHeaders trailers) {
-            this.trailers = trailers;
+        BufferHttpPayloadWriter(final Supplier<HttpHeaders> trailersFactory) {
+            this.trailersFactory = trailersFactory;
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpPayloadWriter.java
```diff
@@ -27,24 +27,52 @@
  */
 public interface HttpPayloadWriter<T> extends PayloadWriter<T>, TrailersHolder {
 
+    /**
+     * {@inheritDoc}
```
idelpivnitskiy, line 31: Does anyone know what's wrong with JDK21? Its output:
```
/home/runner/work/servicetalk/servicetalk/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpPayloadWriter.java:36: warning: @inheritDoc used but trailers() does not override or implement any method.
    HttpHeaders trailers();
                ^
error: warnings found and -Werror specified
1 error
15 warnings

> Task :servicetalk-http-api:javadoc FAILED
```

JDK 8, 11, and 17 are happy with this javadoc.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/BlockingStreamingToStreamingServiceTest.java
```diff
@@ -93,14 +98,26 @@ void setup() {
...
-    @Test
-    void defaultResponseStatusNoPayload() throws Exception {
-        BlockingStreamingHttpService syncService = (ctx, request, response) -> response.sendMetaData().close();
+    @ParameterizedTest(name = "{displayName} [{index}] withEmptyTrailers={0}")
+    @ValueSource(booleans = {false, true})
+    void defaultResponseStatusNoPayload(boolean withEmptyTrailers) throws Exception {
+        BlockingStreamingHttpService syncService = (ctx, request, response) -> {
+            HttpPayloadWriter<Buffer> writer = response.sendMetaData();
+            if (withEmptyTrailers) {
+                writer.trailers();  // accessing trailers before close should preserve trailers in message body
```
daschl, line 107: What's the use-case of empty trailers?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -197,11 +202,61 @@ public void close(final Throwable cause) throws IOException {
...
+
+        @Nullable
+        @Override
+        public Object onNext() {
+            return payloadWriter.trailers0();
+        }
+
+        @Override
+        public boolean onNextValid() {
+            return payloadWriter.trailers0() != null;
```
daschl, line 253: see below question on the use-case for emptry trailers. If not, we could also check here that it's not null and also not empty?

servicetalk-http-api/src/test/java/io/servicetalk/http/api/BlockingStreamingToStreamingServiceTest.java
idelpivnitskiy: Good question. Looking at it with a fresh morning look. 

My original thinking was that if users didn't touch trailers at all, then it's absolutely safe to drop them. But if they did, they could do something like:
```java
HttpHeaders trailers = writer.trailers();
....
trailers.add(name, value);
```
However, if they did that before `close()`, then we will see that. If they added smth after `close()`, it's anyway racy and not guaranteed. 

Another thing is how we treat trailers in all other places:
1. DefaultHttpRequest.toStreamingRequest() or DefaultHttpResponse.toStreamingResponse()

https://github.com/apple/servicetalk/blob/6fc79877e278769c7c1fe758c2e7f04ce9c2d83b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequest.java#L277-L283

2. HttpDataSourceTransformations.aggregatePayloadAndTrailers:

https://github.com/apple/servicetalk/blob/6fc79877e278769c7c1fe758c2e7f04ce9c2d83b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpDataSourceTransformations.java#L340-L348

Those account for the case when users can grab a reference to trailers, then convert the response and still use that reference to add trailers. Something like:
```java
HttpResponse aggregatedResponse = ...;
HttpHeaders trailers = aggregatedResponse.trailers();
StreamingHttpResponse streamingResponse = aggregatedResponse.toStreamingResponse();
trailers.add(name, value);
return Single.succeeded(streamingResponse);
```
However, now I see that `BlockingStreamingHttpServerResponse` doesn't have conversion methods at all, and this trick won't be possible with `HttpPayloadWriter` anyway bcz we already sent meta-data.

The only real reason to keep it as a null check instead of "null or empty" is to keep it consistent with the above 2 cases.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpPayloadWriter.java
idelpivnitskiy: For not I just removed `@inheritDoc` and kept only new note. Should I copy the original javadoc from parent interface for all methods?

servicetalk-http-api/src/test/java/io/servicetalk/http/api/BlockingStreamingToStreamingServiceTest.java
mgodave: I think keeping the cases consistent makes sense but if there is a way to simplify I would prefer that.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequest.java
```diff
@@ -245,7 +245,7 @@ public <T> HttpRequest payloadBody(final T pojo, final HttpSerializer2<T> serial
     public HttpHeaders trailers() {
         if (trailers == null) {
             trailers = original.payloadHolder().headersFactory().newTrailers();
-            original.transform(this);
+            original.transform(this);   // Invoke "transform" to set PayloadInfo.mayHaveTrailers() flag
```
mgodave, line 248: this seemed odd to me and was one thing I had trouble grokking when I was trying to track this down.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -197,11 +202,61 @@ public void close(final Throwable cause) throws IOException {
 
         @Override
         public HttpHeaders trailers() {
```
mgodave, line 204: A "getter" method that also creates seems off to me. IDK if there is a better naming scheme where `trailers0` is basically just `trailers` and `trailers` is `getOrCreateTrailers`.

mgodave (approved):

servicetalk-http-api/src/test/java/io/servicetalk/http/api/BlockingStreamingToStreamingServiceTest.java
idelpivnitskiy: +1 for starting in a consistent way and if we feel different, we can update them all together

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpRequest.java
idelpivnitskiy: I agree 😞 
Unfortunately, we currently don't have a nice way to update `PayloadInfo` flags without invoking transformation that will do it for us. Maybe we refactor in the future.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
idelpivnitskiy: It's "lazy get" 😄 
We use the same trick in `AbstractHttpMetaData.context()`. When API is written in stone but we need to defer action. 
From users point of view it will be non-visible change.

servicetalk-http-api/src/test/java/io/servicetalk/http/api/BlockingStreamingToStreamingServiceTest.java
daschl: Thank you for the explanation, makes sense to me to keep it consistent for now 👍

daschl (approved):

PR #3152, opened 2024-12-20
gRPC `Trailers-Only` responses must be a single `HEADERS` frame

mgodave:
Replaces #3148

### Motivation

When a blocking gRPC endpoint with streaming response encounters an exception, it attempts to send `Trailers-Only` response (grpc-status is returned in headers rather than trailers). However, the HTTP/2 layer sends two frames instead of one: `HEADERS with endStream=false` and `DATA with endStream=true`. Official grpc-java implementation doesn't like that, it wants a single `HEADERS` frame with `endStream=true` for `Trailers-Only` responses.

### Changes

I added a filter and a context key to signal when we have a case where `Trailers-Only` response is sent. The filter will aggregate the response stream, observe there are no trailers, drop them, and will only return a single `HEADERS` frame with `endStream=true`.
This behavior can be tested by `ProtocolCompatibilityTest.unimplementedServiceError`, we unskipped tests that were failing before this fix.

### Results

Before Change:

```
OUTBOUND HEADERS: streamId=3 headers=GrpcHttp2OutboundHeaders[:authority: 127.0.0.1:62451, :path: /grpc.netty.Compat/clientStreamingCall, :method: POST, :scheme: http, content-type: application/grpc, te: trailers, user-agent: grpc-java-netty/1.64.1, grpc-accept-encoding: gzip] streamDependency=0 weight=16 exclusive=false padding=0 endStream=false
OUTBOUND DATA: streamId=3 padding=0 endStream=false length=7 bytes=00000000020801
OUTBOUND DATA: streamId=3 padding=0 endStream=true length=0 bytes=
INBOUND HEADERS: streamId=3 headers=GrpcHttp2ResponseHeaders[:status: 200, server: servicetalk-grpc/, content-type: application/grpc+proto, grpc-status: 12, grpc-message: Method grpc.netty.Compat/clientStreamingCall is unimplemented] padding=0 endStream=false
INBOUND DATA: streamId=3 padding=0 endStream=true length=0 bytes=
```

After Change:

```
OUTBOUND HEADERS: streamId=3 headers=GrpcHttp2OutboundHeaders[:authority: 127.0.0.1:62421, :path: /grpc.netty.Compat/clientStreamingCall, :method: POST, :scheme: http, content-type: application/grpc, te: trailers, user-agent: grpc-java-netty/1.64.1, grpc-accept-encoding: gzip] streamDependency=0 weight=16 exclusive=false padding=0 endStream=false
OUTBOUND DATA: streamId=3 padding=0 endStream=false length=7 bytes=00000000020801
OUTBOUND DATA: streamId=3 padding=0 endStream=true length=0 bytes=
INBOUND HEADERS: streamId=3 headers=GrpcHttp2ResponseHeaders[:status: 200, server: servicetalk-grpc/, content-type: application/grpc+proto, grpc-status: 12, grpc-message: Method grpc.netty.Compat/clientStreamingCall is unimplemented, content-length: 0] padding=0 endStream=true
```

### DIFF
--- a/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
+++ b/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
@@ -146,6 +146,8 @@ final class DefaultGrpcServerBuilder implements GrpcServerBuilder, ServerBinder 
         final ExecutionContextInterceptorHttpServerBuilder interceptor =
                 new ExecutionContextInterceptorHttpServerBuilder(httpServerBuilderSupplier.get());
 
+        interceptor.appendServiceFilter(GrpcTrailersOptimizationServiceFilter.INSTANCE);
+
         interceptor.appendNonOffloadingServiceFilter(GrpcExceptionMapperServiceFilter.INSTANCE);
 
         directCallInitializer.initialize(interceptor);
--- a/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
+++ b/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
@@ -145,9 +145,6 @@ final class DefaultGrpcServerBuilder implements GrpcServerBuilder, ServerBinder 
     private ExecutionContextInterceptorHttpServerBuilder preBuild() {
         final ExecutionContextInterceptorHttpServerBuilder interceptor =
                 new ExecutionContextInterceptorHttpServerBuilder(httpServerBuilderSupplier.get());
-
-        interceptor.appendServiceFilter(GrpcTrailersOptimizationServiceFilter.INSTANCE);
-
         interceptor.appendNonOffloadingServiceFilter(GrpcExceptionMapperServiceFilter.INSTANCE);
 
         directCallInitializer.initialize(interceptor);
@@ -155,6 +152,8 @@ final class DefaultGrpcServerBuilder implements GrpcServerBuilder, ServerBinder 
             interceptor.appendNonOffloadingServiceFilter(newGrpcDeadlineServerFilterFactory(defaultTimeout));
         }
         initializer.initialize(interceptor);
+
+        interceptor.appendServiceFilter(GrpcEnforceTrailersOnlyResponseServiceFilter.INSTANCE);
 
         return interceptor;
     }
--- a/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcTrailersOptimizationServiceFilter.java
+++ b/servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcTrailersOptimizationServiceFilter.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright © 2019, 2021 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.grpc.netty;
+
+import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.http.api.HttpExecutionStrategies;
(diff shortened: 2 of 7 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpResponse.java
```diff
@@ -130,7 +127,7 @@ public BlockingStreamingHttpResponse toBlockingStreamingResponse() {
     public HttpHeaders trailers() {
         if (trailers == null) {
             trailers = original.payloadHolder().headersFactory().newTrailers();
-            original.transform(this);   // Invoke "transform" to set PayloadInfo.mayHaveTrailers() flag
+            original.transform(this);
```
idelpivnitskiy, line 130: Please revert all changes in this file, they are undoing part of my PR

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcTrailersOptimizationServiceFilter.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public StreamingHttpServiceFilter create(StreamingHttpService service) {
+        return new StreamingHttpServiceFilter(service) {
+            @Override
+            public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
+                                                        final StreamingHttpRequest request,
+                                                        final StreamingHttpResponseFactory responseFactory) {
+                return super.handle(ctx, request, responseFactory).flatMap(response -> {
```
idelpivnitskiy, line 45: Please use `delegate()` instead of `super`, in the past we had cases when we changed super impl and it unexpectedly affected users who used `super`.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
```diff
@@ -68,6 +68,9 @@ public final class HttpContextKeys {
     public static final Key<Boolean> HTTP_FORCE_NEW_CONNECTION =
             newKey("HTTP_FORCE_NEW_CONNECTION", Boolean.class);
 
+    public static final Key<Boolean> HTTP_OPTIMIZE_ERROR_STREAM =
```
idelpivnitskiy, line 71: 1. This is an internal key, not intended for use by others. It's unfortunate that we need to set it in grpc-api and read in grpc-netty. Instead of creating a public key, consider creating a new class in `servicetalk-grpc-internal` module, something like `GrpcContextKeys`.

2. To align with gRPC protocol wording, consider naming it `TRAILERS_ONLY_RESPONSE`.

3. All new public API should have javadoc, even in internally public classes. Looks like this checkstyle rule is broken now, but adding javadoc now will simplify our live when we fix checkstyle. It will be helpful for us as well to have a note why we need this key.

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/GrpcTrailersOptimizationServiceFilter.java
```diff
@@ -0,0 +1,62 @@
...
+import io.servicetalk.http.api.StreamingHttpRequest;
+import io.servicetalk.http.api.StreamingHttpResponse;
+import io.servicetalk.http.api.StreamingHttpResponseFactory;
+import io.servicetalk.http.api.StreamingHttpService;
+import io.servicetalk.http.api.StreamingHttpServiceFilter;
+import io.servicetalk.http.api.StreamingHttpServiceFilterFactory;
+
+import static io.servicetalk.http.api.HttpContextKeys.HTTP_OPTIMIZE_ERROR_STREAM;
+
+final class GrpcTrailersOptimizationServiceFilter implements StreamingHttpServiceFilterFactory {
```
idelpivnitskiy, line 32: Consider naming it `GrpcEnforceTrailersOnlyResponseServiceFilter`

servicetalk-grpc-netty/src/main/java/io/servicetalk/grpc/netty/DefaultGrpcServerBuilder.java
```diff
@@ -146,6 +146,8 @@ private ExecutionContextInterceptorHttpServerBuilder preBuild() {
         final ExecutionContextInterceptorHttpServerBuilder interceptor =
                 new ExecutionContextInterceptorHttpServerBuilder(httpServerBuilderSupplier.get());
 
+        interceptor.appendServiceFilter(GrpcTrailersOptimizationServiceFilter.INSTANCE);
```
idelpivnitskiy, line 149: Having this filter at the very end, as close as possible to the `GrpcRouter`, will help to minimize the impact: makes sure there is no user code in between that can modify the response before aggregation and allows users to do their modifications after. Consider moving this line after invoking `initializer.initialize(interceptor);`

idelpivnitskiy (commented):
Glad that all tests passed 🎉

idelpivnitskiy (commented):
LGTM!
Will wait until my PR is merged and then will merge yours on Monday

PR #3155, opened 2024-12-23
Adding remote info to logger message so clients can be tracked down

colestanfield:
Motivation:

When we get a log entry from `HttpMessageDiscardWatchdogClientFilter`, there is little to no information to go on to start identifying which client this leak may be coming from. Adding some sort of context about the request like the connection id or remote socket address would at least help us correlate this leak to a particular client connection and eventually the code.

Modifications:

- Added `ConnectionInfo` or `request.effectiveHostAndPort()` info to the logger message.

Result:

Remote info is included in the log message.


idelpivnitskiy (approved):
Thank you!

PR #3159, opened 2025-01-04
Remove unused test dependencies and reduce scope to runtime if possible

idelpivnitskiy:
Motivation:

Reduce warnings produced by dependency-analysis-gradle-plugin for test dependencies.

Modifications:

- Remove unused test dependencies;
- Reduce scope to runtime-only for some test dependencies;
- Add exclude rules for false positives;
- Add missed `package-info.java` files for benchmarks module;

Result:

Less warnings for our dependencies.


daschl (approved):
Nice cleanup!

PR #3169, opened 2025-01-11
Remove unused `api` dependencies or move them to lower scope

idelpivnitskiy:
Motivation:

Dependency-analysis plugin detected that some `api` dependencies are unused and can be either removed or downgraded to `implementation` / `testImplementation` scopes.

Modifications:

- Remove unused `api` dependencies.
- Downgrade some to `implementation` / `testImplementation` scopes.
- Regenerate lock files.

Result:

No more warnings from dependency-analysis plugin.

Risk for users:

Minimal. They can be broken only if they use some dependency at compile time without explicitly declaring it in their build, which is their mistake.

Highlights:
- Most changes are related to `servicetalk-*` dependencies, users likely have them anyway via other modules.
- It's safe to remove `opentelemetry-api` from `opentelemetry-asynccontext` module because in #3164 we added `opentelemetry-context`.
- Removal of `proto-google-common-protos` should not be an issue because it's already an `api` dependency for `grpc-api` module.

### DIFF
--- a/servicetalk-data-protobuf/build.gradle
+++ b/servicetalk-data-protobuf/build.gradle
@@ -27,18 +27,18 @@ ideaModule.dependsOn "generateTestProto"
 
 dependencies {
   api platform("com.google.protobuf:protobuf-bom:$protobufVersion")
-  api project(":servicetalk-buffer-api")
-  api project(":servicetalk-concurrent-api")
   api project(":servicetalk-serialization-api")
   api project(":servicetalk-serializer-api")
   api "com.google.protobuf:protobuf-java"
 
   implementation project(":servicetalk-annotations")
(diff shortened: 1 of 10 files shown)


bryce-anderson (approved):
I wonder if our API exporting could be cleaned up more generally. For example, it seems like many of `-internal` shouldn't be exporting dependencies. However, that seems like a task for a different time, if at all.

idelpivnitskiy:
The only important rule to keep in mind (we will have to implement a custom script to enforce it) is that we should never add any of our `-internal` dependency as `api` for a non-internal module. Everything else should be fine. It's not a problem that `-internal` modules have some dependencies as `api`.

PR #3171, opened 2025-01-13
Change dependency-analysis plugin to `fail` mode

idelpivnitskiy:
Motivation:

All warnings from dependency-analysis plugin were addressed in prior PRs. We can enable the `fail` mode now.

Modifications:

- Change dependency-analysis plugin mode from `warn` to `fail`.

Result:

PRQ pipeline will fail if a new PR introduces warnings for dependencies.


bryce-anderson (approved):
🎉

PR #3184, opened 2025-02-05
`DefaultAsyncContextProvider`: restore previous ContextMap as-is

idelpivnitskiy:
Motivation:

To align with pre-existing behavior and with `slowPathSetContext`, the returned "fast-path" `Scope` should restore `prev` as-is, without allocating a new `ContextMap`.


idelpivnitskiy:
Fixed in #3185

PR #3186, opened 2025-02-05
Introduce the firstAndTail operator to the public API

mgodave:
We have found increasing cases where the `splice` operator can be used outside of the ST implementation and are therefore migrating it to the public API for general use. Instead of "splice" we are adopting "firstAndTail" as the public name.

Previously this operator parameterized different types for the first item and the remaining which, as a result, would attempt to perform a cast inside the operator for the first element. While migrating this operator, for general usage outside of the internal http packages, I have removed this cast and made the simplifying assumption that the head and the tail of the spliced stream are of the type of the input stream.


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2664,6 +2664,10 @@ public final Publisher<T> skipWhile(Predicate<? super T> predicate) {
         return filter(FilterPublisher.skipWhileSupplier(predicate));
     }
 
+    public final <R> Single<R> splice(BiFunction<T, Publisher<T>, R> packer) {
```
mgodave, line 2667: I'm not in love with the name `splice`. I feel like splice should be joining two things together, which perhaps in a way is happening here but this feels more like an "unpack first" or something similar... IDK, I don't have strong feelings but there's still a voice in the back of my head telling me something is off.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/StacklessCancellationException.java
```diff
@@ -0,0 +1,37 @@
+/*
```
mgodave, line 1: copied over for continuity.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SubscribablePublisher.java
```diff
@@ -0,0 +1,34 @@
+/*
```
mgodave, line 1: copied over. Can probably be inlined into the operator since it's not used anywhere else.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
bryce-anderson, line 2667: Please add a javadoc.

Wrt the name, it seems like what it does is processes the head of the publisher into something and then gives you the tail which I know as the `uncons` operator but I think that is more of an FP term. It's not a great name, but I think it's a tiny bit better since you're right: we don't actually splice anything here.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/StacklessCancellationException.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+package io.servicetalk.concurrent.api;
+
+import io.servicetalk.concurrent.internal.ThrowableUtils;
+
+import java.util.concurrent.CancellationException;
+
+final class StacklessCancellationException extends CancellationException {
+    private static final long serialVersionUID = 5434529104526587317L;
```
bryce-anderson, line 23: I think we need a new distinct version here, don't we?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/StacklessCancellationException.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright © 2024 Apple Inc. and the ServiceTalk project authors
```
bryce-anderson, line 2:
```suggestion
 * Copyright © 2025 Apple Inc. and the ServiceTalk project authors
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SubscribablePublisher.java
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
```
bryce-anderson, line 2:
```suggestion
 * Copyright © 2025 Apple Inc. and the ServiceTalk project authors
```

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SubscribablePublisher.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * A {@link Publisher} that is also a {@link PublisherSource} and hence can be subscribed.
+ * <p>
+ * Typically, this will be used to implement a {@link Publisher} that does not require an additional allocation when
+ * converting to a {@link PublisherSource} via {@link SourceAdapters#toSource(Publisher)}.
+ *
+ * @param <T> Type of the items emitted by this {@code SubscribablePublisher}.
+ */
+public abstract class SubscribablePublisher<T> extends Publisher<T> implements PublisherSource<T> {
```
bryce-anderson, line 28: Can we avoid making this part of the public API? The only use case seems to be in `SpliceFlatStreamToPackedSingle` and it seems like we can just return an anonymous `Publisher<T>` and remove this altogether.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SpliceFlatStreamToMetaSingleTckTest.java
```diff
@@ -1,45 +0,0 @@
-/*
- * Copyright © 2022 Apple Inc. and the ServiceTalk project authors
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
```
bryce-anderson, line 5: Can we adapt and move the tests to servicetalk-concurrent-api? Generally speaking, we try to have tck and unit tests for all operators.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SubscribablePublisher.java
mgodave: yes, I missed this when I moved it over, we should definitely not have this as part of the public API

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SpliceFlatStreamToMetaSingleTckTest.java
mgodave: Right. This was not meant to be deleted int he final draft...

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
mgodave: duh, I missed this... Yeah, `uncons` is exactly what I was thinking but I wasn't sure if we wanted to go down that road...

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/StacklessCancellationException.java
mgodave: I always forget the rules. I do believe it is safe to change it either way so I will.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SubscribablePublisher.java
mgodave: and yes it can be inlined

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
idelpivnitskiy: TIL about `uncons` :) 
I'm +1 for renaming `splice` to something. Reading about FP, `uncons` sometimes is also named `split`. But `split` can happen for any condition, not just the first elements like `uncons`. Some options to vote for:
- `uncons` - like in FP
- `splitFirst` - to indicate the condition
- `firstThenTail`/`firstAndTail` (or similar with `first*`) - similar to existing operators we already have: `firstOrElse` and `firstOrError`

mgodave: I think in terms of continuity with our codebase, something with `first` makes sense to me. I would place my vote with `firstAndTail`.

bryce-anderson: `firstAndTail` is fine with me. 👍

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SpliceFlatStreamToMetaSingleTest.java
```diff
@@ -1,347 +0,0 @@
...
-import static org.hamcrest.Matchers.empty;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.instanceOf;
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.notNullValue;
-import static org.hamcrest.Matchers.sameInstance;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-
-class SpliceFlatStreamToMetaSingleTest {
```
idelpivnitskiy, line 49: These tests should be preserved, move them to concurrent-api module

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2664,6 +2664,36 @@ public final Publisher<T> skipWhile(Predicate<? super T> predicate) {
         return filter(FilterPublisher.skipWhileSupplier(predicate));
     }
 
+    /**
+     * After the first element of this {@link Publisher} is emitted pass it, and a {@link Publisher} representing the
```
idelpivnitskiy, line 2668: Consider starting with `Converts this {@link Publisher} to a {@link Single} that will contain the first element ...`, similar to other `first*` operators

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2664,6 +2664,36 @@ public final Publisher<T> skipWhile(Predicate<? super T> predicate) {
...
+     *         Iterator<T> tail;
+     *     }
+     *
+     *     Iterator<T> itr = resultOfThisPublisher();
+     *     Result result = new Result(itr.next(), itr);
+     *     return result;
+     * }</pre>
+     *
+     * @param packer A function that takes the head of the input stream and processes it, along with a {@link Publisher}
+     * of the remainder of the stream.
```
idelpivnitskiy, line 2688: We should document a note that the `packer` itself and any following operator that works with `R` absolutely must take care of the tail publisher if they decide to throw an exception. Users should either drain it or subscribe and cancel. Otherwise, they may leak resources.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SpliceFlatStreamToPackedSingle.java
```diff
@@ -108,12 +99,12 @@ private static final class SplicingSubscriber<Data, MetaData, Payload>
          * non-volatile field to allow caching in register and avoid instanceof and casting on the hot path.
          */
         @Nullable
-        private PublisherSource.Subscriber<Payload> payloadSubscriber;
+        private PublisherSource.Subscriber<T> payloadSubscriber;
```
idelpivnitskiy, line 102: Consider renaming all `payload*` to `tail*`

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SpliceFlatStreamToPackedSingle.java
```diff
@@ -229,10 +218,10 @@ public void onNext(@Nullable Object obj) {
         }
 
         @Nonnull
-        private Publisher<Payload> newPayloadPublisher() {
-            return new SubscribablePublisher<Payload>() {
+        private Publisher<T> newPayloadPublisher() {
+            return new Publisher<T>() {
```
idelpivnitskiy, line 222: It will be beneficial (helps to avoid allocations when conversion is performed) if we can preserve `implements PublisherSource<T>`. Even if `SubscribablePublisher` is not available in this module, we can duplicate it here.

@bryce-anderson how do you think, from the context propagation point of view, should we use `AbstractSynchronousPublisher` here instead of `SubscribablePublisher`?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SpliceFlatStreamToPackedSingle.java
```diff
@@ -291,14 +280,14 @@ public void onError(Throwable t) {
                 payloadSubscriber.onError(t);
```
idelpivnitskiy, line 280: In lines 268-269, we should change message for `DuplicateSubscribeException`.

We should also add a note in javadoc that `tail` publisher has a constraint that users can subscribe to it only once. So, if they splice in memory publishers (like `Publisher.range(...)`) it's not a surprise for them to see `DuplicateSubscribeException`.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SpliceFlatStreamToPackedSingle.java
```diff
@@ -336,7 +325,7 @@ public void onComplete() {
         }
 
         private void ensureResultSubscriberOnSubscribe() {
-            assert !metaSeenInOnNext : "Already seen meta-data";
+            assert !firstElementSeenInOnNext : "Already seen first element";
```
idelpivnitskiy, line 328: Please update debug log in line 321 to replace "payload" with "tail"

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/SpliceFlatStreamToMetaSingleTest.java
mgodave: nevemind, apparently they didn't get added to the diff... one sec.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SpliceFlatStreamToPackedSingle.java
bryce-anderson: tbh, the structure there isn't 100% clear to me, but it looks like `AbstractSynchronousPublisher` has the behavior we want even if the name may not make sense in in this case.

mgodave: Added `AbstractSynchronousPublisher`

bryce-anderson (approved):

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/FirstAndTailToPackedSingleTest.java
```diff
@@ -46,50 +42,51 @@
...
 
-class SpliceFlatStreamToMetaSingleTest {
-    private final TestSingleSubscriber<Data> dataSubscriber = new TestSingleSubscriber<>();
-    private final TestPublisher<Object> upstream = new TestPublisher<>();
-    private final MetaData metaData = new MetaData("foo");
-    private final Payload one = new Payload();
-    private final Payload two = new Payload();
-    private final LastPayload last = new LastPayload();
+class FirstAndTailToPackedSingleTest {
+    private static final int METADATA = 0;
```
idelpivnitskiy, line 46: METADATA -> HEAD?

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2664,6 +2665,40 @@ public final Publisher<T> skipWhile(Predicate<? super T> predicate) {
         return filter(FilterPublisher.skipWhileSupplier(predicate));
     }
 
+    /**
+     * Converts this {@link Publisher} to a {@link Single} that will contain the first element of this {@link Publisher}
+     * and a {@link Publisher} representing the remainder of the stream, to a mapping function.
+     * <p>
+     * Note that either the packer function itself or any operator following this one MUST eventually take care of the
+     * tail {@link Publisher} or they risk leaking resources. This includes cases where an exception is thrown before
```
idelpivnitskiy, line 2673: Consider: they risk leaking resources or attached callbacks on the original Publisher.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2664,6 +2665,40 @@ public final Publisher<T> skipWhile(Predicate<? super T> predicate) {
...
+     * returning a result. In addition, the tail may only be subscribed to exactly one time. For instance, using memory
```
idelpivnitskiy, line 2674: Consider: where an exception is thrown instead of returning the packed object or a failed Single is returned.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/Publisher.java
```diff
@@ -2664,6 +2665,40 @@ public final Publisher<T> skipWhile(Predicate<? super T> predicate) {
...
+     * publishers such as range, will result in a {@link DuplicateSubscribeException}.
```
idelpivnitskiy, line 2675: Consider: 
```
In addition, the tail may only be subscribed to exactly one time. Subsequent {@link Subscriber}s will receive {@link DuplicateSubscribeException}, even if the original publisher supports re-subscribes.
```

idelpivnitskiy (commented):
Minor nits, then LGTM:

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FirstAndTailToPackedSingle.java
```diff
@@ -218,55 +218,55 @@ public void onNext(@Nullable T next) {
...
-                            // subscribes to the payload after cancellation.
+                            // subscribes to the tail after cancellation.
                             newSubscriber.onError(new CancellationException(
                                     "Canceled prematurely from SplicingSubscriber.cancelData(..), current state: " +
                                             maybeSubscriber));
                         } else {
                             // Existing subscriber or terminal event consumed by other subscriber (COMPLETED_DELIVERED)
                             newSubscriber.onError(new DuplicateSubscribeException(maybeSubscriber, newSubscriber,
-                                    "HTTP request/response payload can only be subscribed to once"));
+                                    "HTTP request/response tail can only be subscribed to once"));
```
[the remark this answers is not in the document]
mgodave: yeah, this is clearly what happens when you just do a search and replace.

idelpivnitskiy (approved):
🚀

PR #3189, opened 2025-02-11
Preparing for 0.42.55-SNAPSHOT development

mgodave:


gradle.properties
```diff
@@ -23,7 +23,7 @@ org.gradle.jvmargs=-Xms2g -Xmx4g -dsa -da -ea:io.servicetalk... -XX:+HeapDumpOnO
 
 # project metadata used for publications
 group=io.servicetalk
-version=0.42.54-SNAPSHOT
+version=0.42.55-SNAPSHOT
```
idelpivnitskiy, line 26: We do not open manual PRs, the release should be handled by release.sh script. If all pre-checks are good, run the following command:
```
 DRYRUN= ./scripts/release.sh skip 0.42.55-SNAPSHOT
```

idelpivnitskiy (changes requested):

PR #3192, opened 2025-02-18
Make all async sources subscribable

idelpivnitskiy:
Motivation:

If all sources also implement their corresponding `*Source` interface, they can be converted using `SourceAdapters` without allocating a wrapper.

Modifications:

- Introduce internal `SubscribableSources` in `concurrent-api`.
- Use `SubscribableSources` as a base class for all operators.
- Use `SubscribableSources` equivalent from `concurrent-api-internal` in all other modules where we missed `*Source` interface.

Result:

`SourceAdapters.toSource(...)` don't require allocations for our async primitives.


bryce-anderson (approved):
Nice cleanup. It's sad we've duplicated `SubscribablePublisher` between the `concurrent-api` and `concurrent-api-internal` packages but such is life.

PR #3193, opened 2025-02-19
Update exceptions from `AbstractNoHandleSubscribe` sources

idelpivnitskiy:
Motivation:

`AbstractNoHandleSubscribe[Completable|Single|Publisher]` were used in the past not only to enforce non-standard context propagation but also an `Executor`. Then in 2020 we removed an `Executor` reference from operators chain, but did not reword these exceptions.

Modifications:

- Align exception type to always fail subscriber with `UnsupportedOperationException`.
- Align exception message to clarify it's intended to target `CapturedContext` instead of `Executor`.
- Simplify `LiftSynchronousPublisherToSingle` by extending `AbstractNoHandleSubscribeSingle`.

Result:

Updated and consistent behavior of `AbstractNoHandleSubscribe` sources.


bryce-anderson (approved):
Nice cleanup. 🚀

PR #3196, opened 2025-02-19
SourceAdapters should convert back and forth without allocation

idelpivnitskiy:
Motivation:

`SourceAdapters.toSource(...)` currently returns a source without implementing the original type. In result, operation like `fromSource(toSource(notSource))` returns double wrapped object instead of wrapping only once.

Modifications:

- `PublisherToPublisherSource` extends `Publisher`;
- `SingleToSingleSource` extends `Single`;
- `CompletableToCompletableSource` extends `Completable`;
- Enhance `SourceAdaptersTest` to cover all transformations (via cast and via wrapping).

Result:

Less wrapping layers for back and forth conversions and better test coverage for `SourceAdapters`.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceAdapters.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceAdapters.java
@@ -19,6 +19,8 @@ import io.servicetalk.concurrent.CompletableSource;
 import io.servicetalk.concurrent.PublisherSource;
 import io.servicetalk.concurrent.SingleSource;
 
+import static io.servicetalk.concurrent.api.AbstractNoHandleSubscribeCompletable.newUnsupportedOperationException;
+import static io.servicetalk.concurrent.internal.SubscriberUtils.deliverErrorFromSource;
 import static java.util.Objects.requireNonNull;
 
 /**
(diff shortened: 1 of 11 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceAdapters.java
```diff
@@ -133,39 +135,75 @@ private static <T> SingleSource<T> uncheckedCast(final Single<T> single) {
...
 
+        @Override
+        protected void handleSubscribe(final Subscriber<? super T> subscriber) {
+            deliverErrorFromSource(subscriber, newUnsupportedOperationException(getClass()));
+        }
+
+        @Override
+        void handleSubscribe(final Subscriber<? super T> subscriber,
+                             final CapturedContext capturedContext, final AsyncContextProvider contextProvider) {
+            publisher.handleSubscribe(subscriber, capturedContext, contextProvider);
```
idelpivnitskiy, line 154: @bryce-anderson please carefully double-check that I implemented the right methods and didn't mess up anything by adding `extends`.

idelpivnitskiy (commented):
This is the last PR in this area, I promise 😄

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceAdapters.java
```diff
@@ -133,39 +135,75 @@ private static <T> SingleSource<T> uncheckedCast(final Single<T> single) {
...
 
+        @Override
+        protected void handleSubscribe(final Subscriber<? super T> subscriber) {
+            deliverErrorFromSource(subscriber, newUnsupportedOperationException(getClass()));
+        }
+
+        @Override
+        void handleSubscribe(final Subscriber<? super T> subscriber,
+                             final CapturedContext capturedContext, final AsyncContextProvider contextProvider) {
+            single.handleSubscribe(subscriber, capturedContext, contextProvider);
```
bryce-anderson, line 179: It shouldn't be part of this PR, but can we delete `delegateSubscribe(a, b, c)` in a followup? It is package private and final, only to forwarding calls to `handleSubscribe(a, b, c)` and half the time we just go straight  to the latter.

idelpivnitskiy: Good point. Actually, I didn't pay much attention to `delegateSubscribe` vs `handleSubscribe`. 
In this PR I made the opposite change and used `delegateSubscribe` instead to make it consistent with all the other places.

Somehow, I like an idea of having a clear method name that operators can call to delegate subscribe to another source, while `handleSubscribe(a, b, c)` is used only for overrides. Even though it looks like a redundant call in the stack-trace. I'm open to reconsider in the future if we decide to kill `delegateSubscribe`.

bryce-anderson (approved):
This looks correct to me, but I'll admit this code is pretty difficult to penetrate.

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/SourceAdapters.java
bryce-anderson: We can come back to it.

PR #3197, opened 2025-02-21
Prevent concurrent execution of the same mutable request object

idelpivnitskiy:
Motivation:

Our `HttpRequestMetaData` object is mutable, and we expect users to create a new request every time they need to make a new call. Sequential retries are acceptable, but concurrent execution can corrupt internal state. While these expectations are more clear for HTTP users, with gRPC it gets less obvious that they can not subscribe to the same returned `Single<Message>` concurrently.

Modifications:
- Enhance `FilterableClientToClient` to protect users from concurrent execution of the same request, while still allowing sequential retries.
- Verify concurrent execution is not allowed for HTTP and gRPC.

Result:

Users get `RejectedSubscribeException` if they subscribe to the same Single that shares underlying meta-data object concurrently. This is the best effort to let users know they misused the client.

Risk:

The change may unexpectedly break existing use-cases. To give users some time to adjust their code, we temporarily introduce a system property to opt-out from this new behavior: `-Dio.servicetalk.http.netty.skipConcurrentRequestCheck=true`.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/FilterableClientToClient.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/FilterableClientToClient.java
@@ -18,6 +18,8 @@ package io.servicetalk.http.netty;
 import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.Publisher;
 import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.concurrent.internal.RejectedSubscribeException;
+import io.servicetalk.context.api.ContextMap;
 import io.servicetalk.http.api.BlockingHttpClient;
 import io.servicetalk.http.api.BlockingStreamingHttpClient;
 import io.servicetalk.http.api.FilterableStreamingHttpClient;
@@ -34,9 +36,11 @@ import io.servicetalk.http.api.ReservedHttpConnection;
 import io.servicetalk.http.api.ReservedStreamingHttpConnection;
 import io.servicetalk.http.api.StreamingHttpClient;
 import io.servicetalk.http.api.StreamingHttpRequest;
+import io.servicetalk.http.api.StreamingHttpRequester;
 import io.servicetalk.http.api.StreamingHttpResponse;
 import io.servicetalk.http.api.StreamingHttpResponseFactory;
 
+import static io.servicetalk.context.api.ContextMap.Key.newKey;
 import static io.servicetalk.http.api.HttpApiConversions.toBlockingClient;
 import static io.servicetalk.http.api.HttpApiConversions.toBlockingStreamingClient;
 import static io.servicetalk.http.api.HttpApiConversions.toClient;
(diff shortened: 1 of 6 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/FilterableClientToClient.java
```diff
@@ -196,4 +203,65 @@ public Completable closeAsyncGracefully() {
...
+            // In general, we do not expect users to execute the same mutable request concurrently. Therefore, the cost
+            // of synchronized block should be negligible for most requests, unless they messed up with reactive streams
+            // chain and accidentally subscribed to the same request concurrently. This protection helps them avoid
+            // ambiguous runtime behavior caused by a corrupted mutable request state.
+            final Object inFlight;
+            // Note that because request.context() lazily allocates a new ContextMap, there is a risk that
+            // synchronization will happen on two different contexts. However, this is acceptable compromise because:
+            //  - Most likely users subscribe 2+ times to the same request from the same thread.
+            //  - This is the best effort protection, giving users at least one rejection should be enough to let them
+            //    know their code is incorrect and should be rewritten.
```
bryce-anderson, line 228: Would it be worth it to elevate the synchronization to the request itself?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/FilterableClientToClient.java
```diff
@@ -196,4 +203,65 @@ public Completable closeAsyncGracefully() {
...
+                    synchronized (context) {
+                        context.remove(HTTP_IN_FLIGHT_REQUEST);
+                    }
+                });
+            }
+            return response.shareContextOnSubscribe();
+        });
+    }
+
+    private static Single<StreamingHttpResponse> executeRequest(final StreamingHttpRequester requester,
```
bryce-anderson, line 255: Can we change the method name? This is so close in signature to the other that it might be beneficial to have different names.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/FilterableClientToClient.java
```diff
@@ -196,4 +203,65 @@ public Completable closeAsyncGracefully() {
...
+                                "creation with Single.defer() operator."))
+                        .shareContextOnSubscribe();
+            }
+
+            Single<StreamingHttpResponse> response = executeRequest(requester, request, context, strategy);
+            if (inFlight == null) {
+                // Remove only if we are the one who set the lock.
+                response = response.beforeFinally(() -> {
+                    synchronized (context) {
+                        context.remove(HTTP_IN_FLIGHT_REQUEST);
```
bryce-anderson, line 247: Does it make sense to assert the removed lock is the one we expected?

bryce-anderson (approved):
I don't _really_ like the synchronization so if someone has a better idea I'm interested to hear it. However, I can't think of a better solution that doesn't break behavior so I'm fine with the compromises this makes since it's so easy to back out of the change (both in terms of flag and in terms of we can easily revert and go back to the drawing board if this isn't effective enough etc).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/FilterableClientToClient.java
idelpivnitskiy: Because of gRPC meta-data we should synchronized on the context. Otherwise, underlying HTTP requests will be different.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/FilterableClientToClient.java
idelpivnitskiy: Changed to `setStrategyAndExecute`, open for a better suggestions

bryce-anderson: Nearly anything different will work. I was just lost for a moment because it looked like a recursive call before I examined the parameter types. 😂

mgodave:
> I don't _really_ like the synchronization so if someone has a better idea I'm interested to hear it. However, I can't think of a better solution that doesn't break behavior so I'm fine with the compromises this makes since it's so easy to back out of the change (both in terms of flag and in terms of we can easily revert and go back to the drawing board if this isn't effective enough etc).

My initial reaction was that I wasn't thrilled about the mutability and I didn't really have anywhere to go from there having not directly thought about this problem. I think this is probably about as good as we can achieve otherwise.

daschl (approved):

PR #3206, opened 2025-03-19
Increase visibility for unexpected HTTP/2 errors on the server-side

idelpivnitskiy:
Motivation:

HTTP/2 protocol defines a set of error codes for its exceptions. Some of those are expected and can be logged at debug level, but others could be concerning. Giving higher visibility to users may help them identify problems earlier and identify misbehaving clients.

Modifications:

- Add special rule in `ErrorLoggingHttpSubscriber` to handle `Http2Exception`s;

Result:

Users have better visibility for unexpected HTTP/2 errors.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
@@ -632,6 +632,8 @@ final class NettyHttpServer {
 
         @Override
         public void onError(final Throwable t) {
+            final boolean isOpen = connection.nettyChannel().isOpen();
+            Throwable cause = t;
             if (t instanceof CloseEventObservedException) {
                 final CloseEventObservedException ceoe = (CloseEventObservedException) t;
                 if (ceoe.event() == CHANNEL_CLOSED_INBOUND && t.getCause() instanceof ClosedChannelException) {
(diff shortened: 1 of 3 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -649,7 +663,33 @@ private static void logDecoderException(final DecoderException e,
...
+                        "{} HTTP/2 stream was cancelled by a remote peer, most likely due to timeout or lost interest",
+                        connection, e);
+            } else if (errorCode == NO_ERROR) {
+                LOGGER.debug("{} HTTP/2 stream was closed because underlying connection closed due to GO_AWAY",
+                        connection, e);
+            } else if (errorCode == SETTINGS_TIMEOUT) {
+                LOGGER.warn("{} HTTP/2 stream was closed because underlying connection did not receive SETTINGS " +
+                        "acknowledgement on time", connection, e);
+            } else if (errorCode == PROTOCOL_ERROR || errorCode == INTERNAL_ERROR || errorCode == FLOW_CONTROL_ERROR ||
+                    errorCode == STREAM_CLOSED || errorCode == FRAME_SIZE_ERROR || errorCode == COMPRESSION_ERROR ||
```
bryce-anderson: Are we expecting to receive RST_STREAM(STREAM_CLOSED) due to half closed races? Eg, maybe the client sends a RST_STREAM(CANCEL) frame which races with some DATA frames we already sent, and those materialize into RST_STREAM(STREAM_CLOSED)?

I recall seeing that condition pretty commonly in the past but I don't know if it will surface here since we may have dropped the stream already due to the initial CANCEL.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -649,7 +663,35 @@ private static void logDecoderException(final DecoderException e,
...
+                    INADEQUATE_SECURITY.equals(errorCode) || HTTP_1_1_REQUIRED.equals(errorCode)) {
+                LOGGER.warn("{} HTTP/2 stream failed with an error indicating client misbehavior", connection, e);
+            } else {
+                // REFUSED_STREAM & ENHANCE_YOUR_CALM - we don't support ServerPush, should only happen on client side
+                // CONNECT_ERROR - expected to happen only on client side
+                // Any other non-standard error code - we don't know how to handle it
+                LOGGER.warn("{} HTTP/2 stream failed with an error unexpected for the server-side", connection, e);
+            }
+            if (connection.nettyChannel().isOpen()) {
+                ChannelCloseUtils.close(connection.nettyChannel(), e);
```
daschl, line 690: I would have not expected that a method called `logHttp2Exception` also performs the side effect of closing the connection in addition to logging. Maybe consider either renaming the method or refactoring the code?

daschl (changes requested):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
idelpivnitskiy: Moved closure outside `log*` methods: https://github.com/apple/servicetalk/pull/3206/commits/2f4c7f5596a6db42807a529ac82d53baeec845b9

bryce-anderson (approved):

daschl (approved):

PR #3210, opened 2025-03-24
Protobuf4

idelpivnitskiy:


idelpivnitskiy:
Works well without any code changes on our side, but requires JDK11 as a minimum one

PR #3214, opened 2025-03-27
Server-side: AsyncContext and OTEL span propagation

idelpivnitskiy:


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -325,6 +326,9 @@ public FlushStrategy defaultFlushStrategy() {
         private Completable handleRequestAndWriteResponse(final Single<StreamingHttpRequest> requestSingle,
                                                           final boolean handleMultipleRequests) {
             final Completable exchange = requestSingle.flatMapCompletable(rawRequest -> {
+                System.err.println("--- " + Thread.currentThread().getName() + " --- handleRequestAndWriteResponse-beforeClean: " + AsyncContext.context());
+                AsyncContext.clear();
```
bryce-anderson, line 330: Ideally we'd get a whole new instances of `AsyncContext.context()` every loop, just in case the reference leaks outside of a single request chain.

idelpivnitskiy: This will be guaranteed by a new `defer` applied before repeat. The clear here is only a safeguard to make sure whatever new copy of the context is here, it should be empty.

PR #3219, opened 2025-03-28
Fix `AsyncContext` visibility for OTEL filters

idelpivnitskiy:
Motivation:

`AbstractOpenTelemetryFilter` manually subscribes to response and its payload body but does not share context on subscribe. In result, `AsyncContext` propagation is incorrect when OTEL filters are in place.

Modifications:

- Add `shareContextOnSubscribe()` before subscribing response single and its payload Publisher.
- Add a test that uses our utility to verify correct AsyncContext propagation.

Result:

`AsyncContext` state is correctly propagated through stages of response processing when OTEL filters are in place.

### DIFF
--- a/servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpServerFilterTest.java
+++ b/servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpServerFilterTest.java
@@ -45,6 +45,7 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 
 import static io.servicetalk.concurrent.api.Single.succeeded;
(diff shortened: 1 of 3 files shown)


servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpServerFilterTest.java
```diff
@@ -235,6 +236,11 @@ void testCaptureHeaders() throws Exception {
         }
     }
 
+    @Test
+    void verifyAsyncContextVisibility() throws Exception {
+        verifyServerFilterAsyncContextVisibility(new OpenTelemetryHttpServerFilter());
```
idelpivnitskiy, line 241: @bryce-anderson I found the utility I was describing to you the other day and decided to submit a PR to demonstrate how to use it. We only have it for the server-side for now. It helps assert that filter in question does not break AsyncContext propagation.

bryce-anderson (approved):

PR #3221, opened 2025-03-31
Server-side doesn't propagate AsyncContext through entire exchange

idelpivnitskiy:
Motivation:

When server-side processes HTTP exchange, users expect the same
`AsyncContext` instance to propagate through all parts of the exchange
processing (request, request payload, response, response payload) to
make sure the context is consistent across all callbacks of
`HttpLifecycleObserver`. However, due to the variety of combinations of
ordering in processing, users may get a copy in some cases, which makes
changes to `AsyncContext` invisible to other callbacks.

Modifications:

- Enhance `AbstractHttpServiceAsyncContextTest` and all its subclasses
to assert the same `AsyncContext` instance is set for all
`HttpLifecycleObserver` callbacks, each subsequent request on the same
connection has a new clean `AsyncContext` state, and all tests run with
both HTTP/1.x and HTTP/2 protocols.
- Enhance `verifyServerFilterAsyncContextVisibility` test-fixture to
assert all parts of request-response processing see the same instance of
`AsyncContext`.
- Add `shareContextOnSubscribe()` in all places where it was missing for
correct context propagation.
- Clear `AsyncContext` state in `NettyHttpServerConnection.process(...)`
instead of `ClearAsyncContextHttpServiceFilter`. It helps to clear the
context at the very beginning of the next request handling and avoids
clearing context on every HTTP/1.x request because we clear it before
each request creates its copy of `ContextMap`.

Result:

`AsyncContext` is correctly propagated through all parts of
request-response processing on the server side, regardless of ordering
of events.


bryce-anderson (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -431,11 +431,15 @@ public void onComplete() {
...
                 }
+                // We need to apply shareContextOnSubscribe() to the result of concatenation to share the same context
+                // between steps taken before service.handle(...) and inside service.handle(...).
+                return responseWrite.shareContextOnSubscribe();
             });
-            // AsyncContext is isolated between repeated exchanges by flatMapCompletable that processes a rawRequest.
-            return handleMultipleRequests ? exchange.repeat(__ -> true).ignoreElements() : exchange;
+            // For pipelined connection (repeated reads) we need to wrap each `exchange` with `defer` to isolate
+            // AsyncContext state between repeated cycles.
+            return handleMultipleRequests ? defer(() -> exchange).repeat(__ -> true).ignoreElements() : exchange;
```
idelpivnitskiy, line 442: @bryce-anderson could you have a look at the latest commit (https://github.com/apple/servicetalk/pull/3221/commits/7453de1319a9448594c7c2e8b25707f165528455)?

I don't have a good idea how to test it properly. Existing tests pass before/after only because we have `AbstractLifecycleObserverHttpFilter` as the first filter in the pipeline (either via `server.lifecycleObserver(...)` method or via `HttpMessageDiscardWatchdogServiceFilter.CLEANER`). And because `responseWrite` didn't share context, it was creating a new one that is visible for the filters chain (all assertions). The only ideas that come to my mind:
1. Add plain java `assert` inside `AbstractLifecycleObserverHttpFilter` to make sure `AsyncContext.context()` outside and inside `defer` is the same reference. The risk here is if users have their own filter that intentionally does not share context and then they have `AbstractLifecycleObserverHttpFilter` later in the pipeline, and **run their service with enabled assertions**, they will see errors. Low, but still a risk.
2. Remove `defer` from `AbstractLifecycleObserverHttpFilter` and move it to `HttpLifecycleObserverRequesterFilter`. Defer is needed only when you have retries. There are no retries by default on the server-side, `handle` method is expected to be invoked only once. However, users can implement their own filter that will cache the incoming request and apply retries for `delegate().handle(...)`. In that case, if they have `HttpLifecycleObserverServiceFilter` after their retry filter, they won't see retries. 
3. Leave as-is without testing this path and rely on comments to make sure nobody changes this flow unintentionally. 

The way I found it is logging. My patch:
```patch
Index: servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java	(revision 7453de1319a9448594c7c2e8b25707f165528455)
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java	(date 1743449324141)
@@ -16,6 +16,7 @@
 package io.servicetalk.http.netty;
 
 import io.servicetalk.buffer.api.BufferAllocator;
+import io.servicetalk.concurrent.api.AsyncContext;
 import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.Executor;
 import io.servicetalk.concurrent.api.Single;
@@ -544,6 +545,7 @@
                 public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
+                    System.err.println("--- " + Thread.currentThread().getName() + " --- KeepAliveServiceFilter.handle: " + AsyncContext.context());
                     final HttpKeepAlive keepAlive = HttpKeepAlive.responseKeepAlive(request);
                     // Don't expect any exceptions from delegate because it's already wrapped with
                     // ExceptionMapperServiceFilterFactory
Index: servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java	(revision 7453de1319a9448594c7c2e8b25707f165528455)
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/AbstractLifecycleObserverHttpFilter.java	(date 1743454968235)
@@ -17,6 +17,7 @@
 
 import io.servicetalk.buffer.api.Buffer;
 import io.servicetalk.concurrent.PublisherSource;
+import io.servicetalk.concurrent.api.AsyncContext;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.concurrent.api.TerminalSignalConsumer;
 import io.servicetalk.context.api.ContextMap;
@@ -78,7 +79,9 @@
             final Function<StreamingHttpRequest, Single<StreamingHttpResponse>> responseFunction) {
 
+        System.err.println("--- " + Thread.currentThread().getName() + " --- AbstractLifecycleObserverHttpFilter.handle: " + AsyncContext.context());
         return defer(() -> {
+            System.err.println("--- " + Thread.currentThread().getName() + " --- AbstractLifecycleObserverHttpFilter.handle.defer: " + AsyncContext.context());
             final HttpExchangeObserver onExchange = safeReport(observer::onNewExchange, observer, "onNewExchange",
                     NoopHttpExchangeObserver.INSTANCE);
             final Runnable clearContext;
Index: servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/HttpResponseExceptionHandlingServiceFilter.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/HttpResponseExceptionHandlingServiceFilter.java b/servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/HttpResponseExceptionHandlingServiceFilter.java
--- a/servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/HttpResponseExceptionHandlingServiceFilter.java	(revision 7453de1319a9448594c7c2e8b25707f165528455)
+++ b/servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/HttpResponseExceptionHandlingServiceFilter.java	(date 1743454086836)
@@ -39,7 +39,7 @@
             @Override
             public Single<StreamingHttpResponse> handle(HttpServiceContext ctx, StreamingHttpRequest request,
                                                         StreamingHttpResponseFactory responseFactory) {
-                return super.handle(ctx, request, responseFactory).onErrorResume(cause -> {
+                return delegate().handle(ctx, request, responseFactory).onErrorResume(cause -> {
                     if (cause instanceof HttpResponseException) {
                         // It's useful to include the exception message in the payload for demonstration purposes, but
                         // this is not recommended in production as it may leak internal information.
Index: servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/backends/ErrorResponseGeneratingServiceFilter.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/backends/ErrorResponseGeneratingServiceFilter.java b/servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/backends/ErrorResponseGeneratingServiceFilter.java
--- a/servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/backends/ErrorResponseGeneratingServiceFilter.java	(revision 7453de1319a9448594c7c2e8b25707f165528455)
+++ b/servicetalk-examples/http/service-composition/src/main/java/io/servicetalk/examples/http/service/composition/backends/ErrorResponseGeneratingServiceFilter.java	(date 1743454086831)
@@ -48,7 +48,7 @@
                 if (request.hasQueryParameter(SIMULATE_ERROR_QP_NAME, serviceName)) {
                     return succeeded(responseFactory.internalServerError());
                 }
-                return super.handle(ctx, request, responseFactory);
+                return delegate().handle(ctx, request, responseFactory);
             }
         };
     }
Index: servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java	(revision 7453de1319a9448594c7c2e8b25707f165528455)
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java	(date 1743454968240)
@@ -302,12 +302,14 @@
         }
 
         void process(final boolean handleMultipleRequests) {
+            System.err.println("--- " + Thread.currentThread().getName() + " --- process: " + AsyncContext.context());
             // We must clear the context before starting to read from the connection to make sure every request read
             // has an empty context
             AsyncContext.clear();
 
             final Single<StreamingHttpRequest> requestSingle =
                     connection.read().firstAndTail((head, payload) -> {
+                        System.err.println("--- " + Thread.currentThread().getName() + " --- firstAndTail: " + AsyncContext.context());
                         HttpRequestMetaData meta = (HttpRequestMetaData) head;
                         return newTransportRequest(meta.method(), meta.requestTarget(), meta.version(),
                                 meta.headers(), executionContext().bufferAllocator(), payload,
@@ -330,6 +332,7 @@
             final Completable exchange = requestSingle.flatMapCompletable(rawRequest -> {
+                System.err.println("--- " + Thread.currentThread().getName() + " --- flatMapCompletable: " + AsyncContext.context());
                 // We transform the request and delay the completion of the result flattened stream to avoid
                 // resubscribing to the NettyChannelPublisher before the previous subscriber has terminated. Otherwise
                 // we may attempt to do duplicate subscribe on NettyChannelPublisher, which will result in a connection
@@ -342,6 +345,7 @@
                         // terminated. That means we don't need to trigger the processor as completed because we don't
                         // care about processing more requests.
                         payload -> payload.afterSubscriber(() -> {
+                            System.err.println("--- " + Thread.currentThread().getName() + " --- request.payload.afterSubscriber: " + AsyncContext.context());
                             if (drainRequestPayloadBody) {
                                 payloadSubscribed.set(true);
                             }
@@ -381,6 +385,7 @@
 
                                 @Override
                                 public void onComplete() {
+                                    System.err.println("--- " + Thread.currentThread().getName() + " --- request.payload.onComplete: " + AsyncContext.context());
                                     requestCompletion.onComplete();
                                 }
                             };
@@ -398,6 +403,7 @@
                         // HttpExceptionMapperServiceFilter.
                         service.handle(this, request, streamingResponseFactory())
                         .flatMapPublisher(response -> {
+                            System.err.println("--- " + Thread.currentThread().getName() + " --- response.flatMapPublisher: " + AsyncContext.context());
                             if (responseSent != null) {
                                 // While concurrency between 100 (Continue) and the final response is handled in Netty
                                 // encoders, it's necessary to prevent generating 100 (Continue) response after the full
@@ -420,13 +426,18 @@
                         }));
 
                 if (drainRequestPayloadBody) {
-                    responseWrite = responseWrite.concat(defer(() -> (payloadSubscribed.get() ?
-                            // Discarding the request payload body is an operation which should not impact the state of
-                            // request/response processing. It's appropriate to recover from any error here.
-                            // ST may introduce RejectedSubscribeError if user already consumed the request payload body
-                            requestCompletion : request.messageBody().ignoreElements().onErrorComplete())
-                            // No need to make a copy of the context in both cases.
-                            .shareContextOnSubscribe())
+                    responseWrite = responseWrite.concat(defer(() -> {
+                        System.err.println("--- " + Thread.currentThread().getName() + " --- responseWrite.concat.defer: " + AsyncContext.context());
+                        return (payloadSubscribed.get() ?
+                                // Discarding the request payload body is an operation which should not impact the state of
+                                // request/response processing. It's appropriate to recover from any error here.
+                                // ST may introduce RejectedSubscribeError if user already consumed the request payload body
+                                requestCompletion : request.messageBody().ignoreElements()
+                                .whenOnComplete(() -> System.err.println("--- " + Thread.currentThread().getName() + " --- responseWrite.concat.defer.whenOnComplete: " + AsyncContext.context()))
+                                .onErrorComplete())
+                                // No need to make a copy of the context in both cases.
+                                .shareContextOnSubscribe();
+                    })
                             // We need to apply shareContextOnSubscribe() on deferred Completable to share the same
                             // context between concatenated Completables.
                             .shareContextOnSubscribe());
```

Output before this commit:
```
--- servicetalk-global-io-executor-1-2 --- process: CopyOnWriteContextMap@68ec3ce5:{}
--- servicetalk-global-io-executor-1-2 --- firstAndTail: CopyOnWriteContextMap@62a097c6:{}
--- servicetalk-global-io-executor-1-2 --- flatMapCompletable: CopyOnWriteContextMap@62a097c6:{}
--- servicetalk-global-io-executor-1-2 --- KeepAliveServiceFilter.handle: CopyOnWriteContextMap@62a097c6:{}
--- servicetalk-global-io-executor-1-2 --- AbstractLifecycleObserverHttpFilter.handle: CopyOnWriteContextMap@62a097c6:{}
--- servicetalk-global-io-executor-1-2 --- AbstractLifecycleObserverHttpFilter.handle.defer: CopyOnWriteContextMap@70549be1:{}
--- servicetalk-global-executor-1-1 --- response.flatMapPublisher: CopyOnWriteContextMap@70549be1:{}
--- servicetalk-global-io-executor-1-2 --- responseWrite.concat.defer: CopyOnWriteContextMap@70549be1:{}
--- servicetalk-global-io-executor-1-2 --- request.payload.afterSubscriber: CopyOnWriteContextMap@70549be1:{}
--- servicetalk-global-io-executor-1-2 --- request.payload.onComplete: CopyOnWriteContextMap@70549be1:{}
--- servicetalk-global-executor-1-1 --- responseWrite.concat.defer.whenOnComplete: CopyOnWriteContextMap@70549be1:{}
--- servicetalk-global-io-executor-1-2 --- firstAndTail: CopyOnWriteContextMap@62a097c6:{}
--- servicetalk-global-io-executor-1-2 --- flatMapCompletable: CopyOnWriteContextMap@62a097c6:{}
--- servicetalk-global-io-executor-1-2 --- KeepAliveServiceFilter.handle: CopyOnWriteContextMap@62a097c6:{}
--- servicetalk-global-io-executor-1-2 --- AbstractLifecycleObserverHttpFilter.handle: CopyOnWriteContextMap@62a097c6:{}
--- servicetalk-global-io-executor-1-2 --- AbstractLifecycleObserverHttpFilter.handle.defer: CopyOnWriteContextMap@6e21ec69:{}
--- servicetalk-global-executor-1-1 --- response.flatMapPublisher: CopyOnWriteContextMap@6e21ec69:{}
--- servicetalk-global-io-executor-1-2 --- responseWrite.concat.defer: CopyOnWriteContextMap@6e21ec69:{}
--- servicetalk-global-io-executor-1-2 --- request.payload.afterSubscriber: CopyOnWriteContextMap@6e21ec69:{}
--- servicetalk-global-io-executor-1-2 --- request.payload.onComplete: CopyOnWriteContextMap@6e21ec69:{}
--- servicetalk-global-executor-1-1 --- responseWrite.concat.defer.whenOnComplete: CopyOnWriteContextMap@6e21ec69:{}
```
^ Internal flow shares a single context between subsequent requests (62a097c6), but each `service.handle(...)` has its own context (70549be1 and 6e21ec69).

Output after this commit:
```
--- servicetalk-global-io-executor-1-2 --- process: CopyOnWriteContextMap@225af4d9:{}
--- servicetalk-global-io-executor-1-2 --- firstAndTail: CopyOnWriteContextMap@3e460f6:{}
--- servicetalk-global-io-executor-1-2 --- flatMapCompletable: CopyOnWriteContextMap@3e460f6:{}
--- servicetalk-global-io-executor-1-2 --- KeepAliveServiceFilter.handle: CopyOnWriteContextMap@3e460f6:{}
--- servicetalk-global-io-executor-1-2 --- AbstractLifecycleObserverHttpFilter.handle: CopyOnWriteContextMap@3e460f6:{}
--- servicetalk-global-io-executor-1-2 --- AbstractLifecycleObserverHttpFilter.handle.defer: CopyOnWriteContextMap@3e460f6:{}
--- servicetalk-global-executor-1-1 --- response.flatMapPublisher: CopyOnWriteContextMap@3e460f6:{}
--- servicetalk-global-io-executor-1-2 --- responseWrite.concat.defer: CopyOnWriteContextMap@3e460f6:{}
--- servicetalk-global-io-executor-1-2 --- request.payload.afterSubscriber: CopyOnWriteContextMap@3e460f6:{}
--- servicetalk-global-io-executor-1-2 --- request.payload.onComplete: CopyOnWriteContextMap@3e460f6:{}
--- servicetalk-global-executor-1-1 --- responseWrite.concat.defer.whenOnComplete: CopyOnWriteContextMap@3e460f6:{}
--- servicetalk-global-io-executor-1-2 --- firstAndTail: CopyOnWriteContextMap@5a3c5098:{}
--- servicetalk-global-io-executor-1-2 --- flatMapCompletable: CopyOnWriteContextMap@5a3c5098:{}
--- servicetalk-global-io-executor-1-2 --- KeepAliveServiceFilter.handle: CopyOnWriteContextMap@5a3c5098:{}
--- servicetalk-global-io-executor-1-2 --- AbstractLifecycleObserverHttpFilter.handle: CopyOnWriteContextMap@5a3c5098:{}
--- servicetalk-global-io-executor-1-2 --- AbstractLifecycleObserverHttpFilter.handle.defer: CopyOnWriteContextMap@5a3c5098:{}
--- servicetalk-global-executor-1-1 --- response.flatMapPublisher: CopyOnWriteContextMap@5a3c5098:{}
--- servicetalk-global-io-executor-1-2 --- responseWrite.concat.defer: CopyOnWriteContextMap@5a3c5098:{}
--- servicetalk-global-io-executor-1-2 --- request.payload.afterSubscriber: CopyOnWriteContextMap@5a3c5098:{}
--- servicetalk-global-io-executor-1-2 --- request.payload.onComplete: CopyOnWriteContextMap@5a3c5098:{}
--- servicetalk-global-executor-1-1 --- responseWrite.concat.defer.whenOnComplete: CopyOnWriteContextMap@5a3c5098:{}
```
^ each "exchange" has its own `ContextMap`.

bryce-anderson: We talked about this offline and decided to try adding some package-private filter order manipulation tools. If that doesn't work (is to unwieldy) then I'd be alright with option 3.

idelpivnitskiy: @bryce-anderson I added a hook in a way that won't allow users to sneak into the package to cast the type and call something like `prependInternalFilter(...)`: https://github.com/apple/servicetalk/pull/3221/commits/3c61bd9ebf02e74254389f7b0d061710d99e9c26

If they still do, they have to extend `DefaultHttpServerBuilder` and then providers won't see that.

idelpivnitskiy: And it worked well, thanks a lot for the idea ❤️

bryce-anderson (approved):

PR #3222, opened 2025-03-31
Use `delegate` instead of `super` when calling the next filter

idelpivnitskiy:
Motivation:

It's recommended to use `delegate` for invoking the next filter in the chain rather than `super`. In the past, we had cases when the `super` implementation changed and it affected filter behavior.

Modifications:

- Replace all calls to `super.handle(...)` with `delegate().handle(...)` and `super.request(...)` with `delegate.request(...)`.

Result:

All filters consistently use `delegate`.


daschl (approved):
🚀

bryce-anderson (approved):

PR #3223, opened 2025-04-01
Fix `AsyncContext` propagation through connection acceptors chain

idelpivnitskiy:
Motivation:

When we introduced Early and Late `ConnectionAcceptor`s (#2486) we tried
to isolate `AsyncContext` for each acceptor. However, there could be
use-cases that need to propagate state between acceptors, as all of them
are invoked sequentially. Testing uncovered that the state can propagate
forward if users use only one acceptor of each type, but if there are
multiple acceptors, behavior is incorrect: the first late acceptor sees
only the state of the first early acceptor, but the second early acceptor
state is dropped because it was not shared.

Modifications:

- Enhance `AbstractHttpServiceAsyncContextTest` to validate behavior
when all acceptor types are used once and twice (with a mix of shared/
non-shared context propagations).
- Enhance `connectionAcceptorContextDoesNotLeak` test to open a 2nd
connection and validate the `AsyncContext` state is new and empty for
each new connection. Also, respect `useImmediate` flag in acceptors.
- Adjust `TcpServerBinder.wrapConnectionAcceptors(...)` flow and
`DefaultHttpServerBuilder.build[Early|Late]ConnectionAcceptor` methods
to propagate `AsyncContext` according to test expectations.

Result:

`AsyncContext` state is propagated forward along the entire chain of
acceptors, regardless of their type and quantity. The state of initial
invocation of `accept` method is always shared. Users can control
isolation for their acceptors in the implementation by using or not
using `Completable.defer(...)` operator and `shareContextOnSubscribe()`.

---

Note: this PR is based on top of #3221, review only the last commit: https://github.com/apple/servicetalk/commit/f9d06db9303abd1e225e1d3c17f5b08fb5a60e33


bryce-anderson (approved):
I looked through the last commit (f9d06db9303abd1e225e1d3c17f5b08fb5a60e33) and it looks good to me. I'm giving a :shipit: presuming that https://github.com/apple/servicetalk/pull/3221 will be merged first, then this, without material changes to f9d06db93.

servicetalk-http-netty/src/testFixtures/java/io/servicetalk/http/netty/AsyncContextHttpFilterVerifier.java
```diff
@@ -182,8 +185,22 @@ public static <T> void assertAsyncContext(final Key<T> key, @Nullable final T ex
...
+        assertContext(sameInstance(expected), errorQueue);
+    }
+
+    /**
+     * Asserts that {@link AsyncContext#context()} is NOT the same instance as the expected one.
+     *
+     * @param expected {@link ContextMap} we do not expect
+     * @param errorQueue {@link Queue} to add an {@link AssertionError} in case the assertion fails
+     */
+    public static void assertNotSameContext(@Nullable final ContextMap expected, final Queue<Throwable> errorQueue) {
```
daschl, line 197: nit `expected` -> `notExpected`?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/AbstractHttpServiceAsyncContextTest.java
```diff
@@ -326,25 +337,139 @@ final void connectionAcceptorContextDoesNotLeak(HttpProtocol protocol, boolean u
...
 
+        Queue<Throwable> errorQueue = new ConcurrentLinkedQueue<>();
+        AtomicReference<ContextMap> currentContextMap = new AtomicReference<>();
         HttpServerBuilder builder = HttpServers.forAddress(localAddress(0)).protocols(protocol.config);
         switch (connectionAcceptorType) {
             case EARLY:
-                builder.appendEarlyConnectionAcceptor(conn -> {
-                    AsyncContext.put(K1, "v1");
+                builder.appendEarlyConnectionAcceptor(newEarly(useImmediate, conn -> {
+                    AsyncContext.put(K1, "early");
```
daschl, line 346: are these puts (for early, late and deprecated) asserted somewhere? I see you did it for ALL and ALL_TWICE in the succeeding ones, but here there is only one. I see `assertAsyncContext` calls using the key K1, but they only assert on a request ID but I am not sure how it aligns with these cases.

idelpivnitskiy: The idea of original tests was to check that whatever you put in acceptor does not leak to request controller. There is an assertion in `newEmptyAsyncContextService()` implementations, see `serverWithEmptyAsyncContextService(...)` in subclasses

daschl (approved):

PR #3226, opened 2025-04-09
Make incubator io_uring transport a compileOnly dependency

idelpivnitskiy:
Motivation:

Netty promoted io_uring transport from an incubator state to final in its 4.2.0.Final release. To avoid any conflicts or unexpected behavior for users who adopt 4.2.0.Final release, we would like to avoid bringing the incubator dependency transitively.

Modifications:

- Make `netty-incubator-transport-native-io_uring` a `compileOnly` dependency.
- Add its runtime packages for `servicetalk-http-netty` tests only.

Result:

Users won't see `netty-incubator-transport-native-io_uring` as a prematurely added transitive dependency on their classpath by default. Those users who still want to try it with Netty 4.1.x, will need to add these dependencies manually before activating the system property: `-Dio.servicetalk.transport.netty.tryIoUring=true`.

Risk:

Minimal. To our knowledge, incubator io_uring didn't get much traction. Netty makes further improvements and fixes for io_uring only in 4.2 branch.


daschl (approved):
Good call 👍 .. let's make sure we cover it in the release notes

mgodave (approved):

PR #3227, opened 2025-04-09
`servicetalk-grpc-protoc`: add `servicetalk-grpc-protobuf` transitively

idelpivnitskiy:
Motivation:

`servicetalk-grpc-protoc` generates stubs, and the generated code requires 2 dependencies to make it compilable:
`servicetalk-data-protobuf` and `servicetalk-grpc-protobuf`. `servicetalk-grpc-protoc` already adds the first one transitively, but not the second one. In our examples, it's used as a way to add required transitive dependencies for generated code and should have all required dependencies.

Modifications:

- Add `servicetalk-grpc-protobuf` as a transitive `api` dependency for `servicetalk-grpc-protoc`.
- Remove `servicetalk-grpc-protobuf` from all examples.

Result:

Users don't need to guess what dependencies they need for the generated code, they can simply add `servicetalk-grpc-protoc` as their dependency to bring everything that they need transitively.

### DIFF
--- a/servicetalk-examples/grpc/compression/build.gradle
+++ b/servicetalk-examples/grpc/compression/build.gradle
@@ -29,7 +29,6 @@ dependencies {
   implementation project(":servicetalk-encoding-netty")
   implementation project(":servicetalk-grpc-netty")
   implementation project(":servicetalk-grpc-protoc")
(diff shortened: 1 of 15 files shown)


mgodave (approved):
This makes a lot of sense.

PR #3239, opened 2025-04-20
Add delegating `AsyncCloseable` and `ListenableAsyncCloseable` classes

idelpivnitskiy:
Motivation:

It's a frequent case that users need to wrap `AsyncCloseable` objects. Having delegating implementations will be handy to avoid boilerplate.

Modifications:

- Add `DelegatingAsyncCloseable` class.
- Add `DelegatingListenableAsyncCloseable` class.

Result:

Users will have handy utilities to delegate async closeable methods.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelegatingListenableAsyncCloseable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelegatingListenableAsyncCloseable.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright © 2025 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelegatingListenableAsyncCloseable.java
```diff
@@ -0,0 +1,54 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.concurrent.api;
+
+/**
+ * {@link ListenableAsyncCloseable} that delegates all calls to another {@link ListenableAsyncCloseable}.
+ */
+public class DelegatingListenableAsyncCloseable extends DelegatingAsyncCloseable implements ListenableAsyncCloseable {
```
mgodave, line 21: Do you want to implement a `toString` here as well?

idelpivnitskiy: The inheritance from `DelegatingAsyncCloseable ` will do the trick of producing the same outcome

idelpivnitskiy:
I found why japicmp was not happy with my reverted changes in this PR, will open a follow-up PR

PR #3241, opened 2025-04-21
`BeforeFinallyHttpOperator`: restore binary compatibility

idelpivnitskiy:
Motivation:

Its binary compatibility was broken in #3231.

Modifications:

- Remove `final` modifier from `apply` method of the super class.
- Make its super class `abstract`.

Result:

`japicmp` is happy.


bryce-anderson (approved):
This is kind of surprising to me since before #3231, the only abstraction that was `BeforeFinallyHttpOperator` and it was final, so the `apply` method was also effectively final for all public API. What am I missing?

idelpivnitskiy:
I was fighting with japicmp to find the right combination. Don't have full answers bcz I fight with some similar mysteries in #3239 (see reverted commit there). The way I understood it is that if you move method to super class it's ok, but if you mark it as `final` in the super class, then it's a binary breaking change.

PR #3243, opened 2025-04-21
Use `DelegatingListenableAsyncCloseable` where possible

idelpivnitskiy:
Motivation:

#3239 added new `DelegatingListenableAsyncCloseable` class, but I was not able to use it because of some false positives from japicmp. Now I figured what was the problem and we can remove some boilerplate code.

Modifications:

- Use `DelegatingListenableAsyncCloseable` to simplify `DelegatingExecutor`, `DelegatingConnectionFactory`, `DelegatingServiceDiscoverer`, `DelegatingConnectionContext`, and `DelegatingFilterableStreamingHttpLoadBalancedConnection` implementations.

Result:

Less code to maintain delegating classes that implement `ListenableAsyncCloseable` interface.

Binary compatibility:

If you run `./gradlew assemble && ./scripts/japicmp.sh 0.42.55 local`, it won't report anything for `DelegatingExecutor`, but will report false positives for all other classes. This happens because `japicmp` compares jar-to-jar and does not see changes in other dependencies. I tested that by copying `DelegatingListenableAsyncCloseable` and `DelegatingAsyncCloseable` classes to `servicetalk-client-api` module and then warnings disappear. So, as long as users have all ServiceTalk libraries of the same version, their binary compatibility won't be broken.

### DIFF
--- a/servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelegatingConnectionFactory.java
+++ b/servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelegatingConnectionFactory.java
@@ -16,7 +16,7 @@
 package io.servicetalk.client.api;
 
 import io.servicetalk.concurrent.api.AsyncContext;
-import io.servicetalk.concurrent.api.Completable;
+import io.servicetalk.concurrent.api.DelegatingListenableAsyncCloseable;
 import io.servicetalk.concurrent.api.ListenableAsyncCloseable;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.context.api.ContextMap;
@@ -25,7 +25,6 @@ import io.servicetalk.transport.api.TransportObserver;
 import javax.annotation.Nullable;
 
 import static io.servicetalk.client.api.DeprecatedToNewConnectionFactoryFilter.CONNECTION_FACTORY_CONTEXT_MAP_KEY;
-import static java.util.Objects.requireNonNull;
 
 /**
  * A {@link ConnectionFactory} that delegates all methods to another {@link ConnectionFactory}.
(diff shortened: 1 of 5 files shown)


chrisvest (approved):

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelegatingConnectionFactory.java
```diff
@@ -43,7 +43,18 @@ public class DelegatingConnectionFactory<ResolvedAddress, C extends ListenableAs
...
+
+    /**
+     * Returns the {@link ConnectionFactory} delegate.
+     *
+     * @return Delegate {@link ConnectionFactory}.
+     */
+    @Override
+    protected final ConnectionFactory<ResolvedAddress, C> delegate() {
+        return delegate;
     }
```
bryce-anderson, line 58: Not critical, but can we avoid this noise if we make `DelegatingListenableAsyncCloseable` generic on the type of delegate?

bryce-anderson (approved):

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelegatingConnectionFactory.java
idelpivnitskiy: I'm actually looking at `DelegatingSslConfig` and thinking about the opposite. Generics quickly become complex to use, especially when you need to use `Delegating*` class as a map value. With my recent experience in gRPC codegen, I would love to limit amount of generics we use everywhere.

idelpivnitskiy: I will give it a try in a follow-up PR where we can see how it shapes. If we don't like it, we will close. Will submit shortly

idelpivnitskiy:
@daschl fyi bcz you will see japicmp warnings before generating the next ST release

servicetalk-client-api/src/main/java/io/servicetalk/client/api/DelegatingConnectionFactory.java
idelpivnitskiy: See #3245

PR #3245, opened 2025-04-22
Parametrize `Delegating[Listenable]AsyncCloseable`

idelpivnitskiy:
Motivation:

They typically serve a role of a base interface for specific closeable contracts. Having a parametrized `delegate()` method helps to avoid additional class variables in its subtypes.

Modifications:

- Add `T` parameter for `DelegatingAsyncCloseable` and `DelegatingListenableAsyncCloseable`.

Result:

Classes that extend `Delegating[Listenable]AsyncCloseable` don't need additional class variable to delegate its methods.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelegatingAsyncCloseable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelegatingAsyncCloseable.java
@@ -40,7 +40,7 @@ public class DelegatingAsyncCloseable<T extends AsyncCloseable> implements Async
      *
      * @return the {@link T} subtype of {@link AsyncCloseable} that this class delegates to.
      */
-    protected T delegate() {
+    protected T delegate() {    // FIXME: 0.43 - consider making this method `final`
         return delegate;
     }
 
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelegatingAsyncCloseable.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelegatingAsyncCloseable.java
@@ -19,26 +19,28 @@ import static java.util.Objects.requireNonNull;
 
 /**
  * {@link AsyncCloseable} that delegates all calls to another {@link AsyncCloseable}.
+ *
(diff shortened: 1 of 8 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelegatingExecutor.java
```diff
@@ -37,67 +35,67 @@ public class DelegatingExecutor extends DelegatingListenableAsyncCloseable imple
...
+     * Get the {@link Executor} that this class delegates to.
      *
-     * @return The delegate {@link Executor} used.
+     * @return the {@link Executor} that this class delegates to.
      */
     @Override
+    @SuppressWarnings("PMD.UselessOverridingMethod") // Method is overridden to preserve binary compatibility
     protected Executor delegate() {
-        return delegate;
+        return super.delegate();
```
idelpivnitskiy, line 48: I had to keep `delegate()` overrides for binary compatibility. Idk why jampcmp is mad if I remove this method. It makes no sense to me, taking into account that the original method inside `DelegatingAsyncCloseable` is not final, and we did the same trick in #3241

idelpivnitskiy (commented):
Lmk what you think. I like that we don't need to store `delegate` as a class variable in every layer, but there is a trick with removing `delegate()` overrides:

servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/DelegatingAsyncCloseable.java
```diff
@@ -19,26 +19,28 @@
...
 
     /**
-     * Get the {@link AsyncCloseable} that this class delegates to.
+     * Get the {@link T} subtype of {@link AsyncCloseable} that this class delegates to.
      *
-     * @return the {@link AsyncCloseable} that this class delegates to.
+     * @return the {@link T} subtype of {@link AsyncCloseable} that this class delegates to.
      */
-    protected AsyncCloseable delegate() {
+    protected T delegate() {
```
bryce-anderson, line 43: I know we can't do it now, but could we add a `FIXME` to make this final in a future breaking release? It seems to me that if users want to change the delegate they should do it before the pass it in.

bryce-anderson (approved):

PR #3273, opened 2025-05-21
Change `[Early|Late]ConnectionAcceptor` API to accept `ConnectionContext`

idelpivnitskiy:
Motivation:

There are use-cases, when users need to track a lifecycle of accepted connections. For example, maintaining a counter of open connections to make accept/reject decisions.

Modifications:

- Add `accept(ConnectionContext)` overload for `EarlyConnectionAcceptor` and `LateConnectionAcceptor`.
- Add default implementation to maintain backward compatibility.
- Change `TcpConnectionInfo` to `EarlyConnectionContext` and extend `NettyChannelListenableAsyncCloseable`.
- Update code references and tests to use the new method where possible.

Result:

Implementations of `EarlyConnectionAcceptor` and/or `LateConnectionAcceptor` can track connection lifecycle.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/EarlyConnectionAcceptor.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/EarlyConnectionAcceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2023 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2023, 2025 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/EarlyConnectionAcceptor.java
```diff
@@ -29,14 +29,32 @@ public interface EarlyConnectionAcceptor extends ExecutionStrategyInfluencer<Con
...
      *
-     * @param info additional information about the connection to make an acceptance decision.
-     * @return a completed (to accept) or a failed (to reject) {@link Completable}
+     * @param info {@link ConnectionInfo} to make an acceptance decision
+     * @return {@link Completable#completed()} to accept, or {@link Completable#failed(Throwable)} to reject with the
+     * cause
+     * @deprecated Implement {@link #accept(ConnectionContext)} instead
      */
+    @Deprecated // FIXME: 0.43 - swap default implementation with non-deprecated method
     Completable accept(ConnectionInfo info);
```
idelpivnitskiy, line 41: This is why I dislike `@FunctionalInterface`s, they don't allow adding default impl for a method used as lambda, which requires 2 major releases to get rid of the deprecated method. Assuming all users implemented new overload, it means in 0.43 we can swap `default` impl`, and only in `0.44` we can remove the deprecated method 😞

PR #3277, opened 2025-05-28
Tighten usage of Vary HTTP header

daschl:
Motivation:
The current behavior around the Vary HTTP header has two issues associated:

 - It is a response-only header, so the client should not send it when sending a compressed payload.
 - The Vary header describes the parts of the request message that influenced the content of the response it occurs in. This in practice means for a server to return Accept-Encoding, but we return Content-Encoding which is not correct.

Modifications:
This changeset first makes sure that the client actually never sends the vary header with its request, even if the payload is being compressed. Then, if the vary header is set by the server side it is not set to Content-Encoding but rather to Accept-Encoding, which influenced the compression decision on the response from the incoming request.

Result:
Vary header behavior which aligns better with HTTP protocol expectations.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ServiceTalkContentEncodingTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ServiceTalkContentEncodingTest.java
@@ -21,6 +21,8 @@ import io.servicetalk.encoding.api.BufferEncoder;
 import io.servicetalk.http.api.BlockingHttpClient;
 import io.servicetalk.http.api.ContentEncodingHttpRequesterFilter;
 import io.servicetalk.http.api.ContentEncodingHttpServiceFilter;
+import io.servicetalk.http.api.HttpHeaderNames;
+import io.servicetalk.http.api.HttpHeaderValues;
 import io.servicetalk.http.api.HttpResponse;
 import io.servicetalk.http.api.HttpServiceContext;
 import io.servicetalk.http.api.StreamingHttpClientFilter;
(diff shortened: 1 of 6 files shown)


mgodave (approved):

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ServiceTalkContentEncodingTest.java
```diff
@@ -123,6 +126,11 @@ protected Single<StreamingHttpResponse> request(final StreamingHttpRequester del
                 // content encoding should be stripped by the time the decoding is done.
                 assertThat(response.headers().get(CONTENT_ENCODING), nullValue());
 
+                final CharSequence varyHeader = response.headers().get(VARY);
+                if (varyHeader != null) {
+                    assertEquals(ACCEPT_ENCODING.toString(), varyHeader.toString());
```
idelpivnitskiy, line 131: Just FYI that we have `io.servicetalk.buffer.api.Matchers.contentEqualTo` that can be used with hamcrest's `assertThat` to avoid null check and `toString()` enforcement

idelpivnitskiy (approved):

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ServiceTalkContentEncodingTest.java
idelpivnitskiy:
```java
assertThat(varyHeader, contentEqualTo(ACCEPT_ENCODING));
```

daschl: Thanks for the info. In this case we can't remove the null check since there might be cases where the vary header is null (no server compression). Should we keep it as is or should I refactor the code to check on a different if condition and then use `contentEqualTo` inside?

PR #3290, opened 2025-07-03
Add utility method to calculate the total header size

daschl:


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaders.java
```diff
@@ -602,4 +602,24 @@ default HttpHeaders addSetCookie(final CharSequence name, final CharSequence val
      * @return {@code true} if at least one entry has been removed.
      */
     boolean removeSetCookies(CharSequence name, CharSequence domain, CharSequence path);
+
+    /**
+     * Returns the total length in bytes of all the header names and values combined.
+     * <p>
+     * Note that this does not include whitespace like new lines or spaces between name and value.
```
daschl, line 611: @idelpivnitskiy I checked and I don't think we can get access to the full total bytes including whitespace in a easy way unless we carry it forward from the decoder?

We could at least add 1 byte to each line to account for the `\n`

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaders.java
```diff
@@ -602,4 +604,24 @@ default HttpHeaders addSetCookie(final CharSequence name, final CharSequence val
      * @return {@code true} if at least one entry has been removed.
      */
     boolean removeSetCookies(CharSequence name, CharSequence domain, CharSequence path);
+
+    /**
+     * Returns the total length in bytes of all the header names and values combined.
```
idelpivnitskiy, line 609:
```suggestion
     * Returns the total length in characters of all the header names and values combined.
```

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaders.java
```diff
@@ -602,4 +604,24 @@ default HttpHeaders addSetCookie(final CharSequence name, final CharSequence val
...
+            return 0;
+        }
+
+        int length = 0;
+        for (final Entry<CharSequence, CharSequence> entry : this) {
+            length += entry.getKey().length();
+            length += entry.getValue().length();
+        }
+        return length;
+    }
```
idelpivnitskiy, line 626: It's good enough for default implementation, but the main motivation for this proposal was that 2 main implementations we have today (`DefaultHttpHeaders` and `NettyH2HeadersToHttpHeaders`) should keep track of the total length on every modification instead of calculating it at runtime. It's possible to do for `DefaultHttpHeaders`, but for `NettyH2HeadersToHttpHeaders` it's tricky for `set` cases bcz we don't know if `set` overrides the previous value or not. It may require a change in Netty first or some other way of being creative.

Until that, there is not much value in adding new API if we can not gain more performance. For now, it can stay as internal utility.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaders.java
idelpivnitskiy: We don't really need that whitespace if we counting length instead of real bytes. 1 optional whitespace can be skipped, but if there is more whitespaces they will become part of the value.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaders.java
daschl: I was considering that, but then I wondered if we really should pay the overhead for every mutation even though the usage of this API might be very limited. (so we would do the calculations pretty much for nothing most of the time)

idelpivnitskiy: Incrementing/decrementing an integer is quite trivial if done on every mutation, it's like updating `size` in every `Collection` implementation. If we end up needing this for every request, it will be worth it.

daschl:
will do that differently.

PR #3293, opened 2025-07-16
Add `HttpContextKeys.HTTP_ROUTE` to propagate `http.route` attribute

idelpivnitskiy:
#### Motivation

OTEL requires `http.route` attribute for server-side metrics.

#### Modifications

- Add `HttpContextKeys.HTTP_ROUTE` key.

#### Result

Router implementation can assign `http.route` value to either request context or `AsyncContext`. Actual implementations for `servicetalk-http-router-predicate` and `servicetalk-http-router-jersey` will be in follow-up PRs.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
@@ -16,6 +16,7 @@
 package io.servicetalk.http.api;
 
 import io.servicetalk.client.api.TransportObserverConnectionFactoryFilter;
+import io.servicetalk.concurrent.api.AsyncContext;
 import io.servicetalk.context.api.ContextMap;
 import io.servicetalk.context.api.ContextMap.Key;
 import io.servicetalk.transport.api.ConnectionInfo;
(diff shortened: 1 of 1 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpContextKeys.java
```diff
@@ -68,6 +69,17 @@ public final class HttpContextKeys {
...
+    /**
+     * The key used by server-side routers to assign an
+     * <a href="https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/#http-route">http.route</a>
+     * attribute for incoming requests.
+     * <p>
+     * Router implementation are responsible for assigning this key to either {@link HttpRequestMetaData#context()}
+     * (preferred) or {@link AsyncContext} if {@link HttpRequestMetaData} is not available in the scope of the router.
+     * This key can be used later by observability extensions to assign an attribute for emitted metrics or traces.
+     */
+    public static final Key<String> HTTP_ROUTE = newKey("http.route", String.class);
```
daschl, line 81: should we follow the other key names and use `HTTP_ROUTE` here as well? I know semantic conventions uses `http.route`, but technically the OTEL implementations need to pick their own attribute name anyways when actually storing it on an otel span, right?

idelpivnitskiy: I looked at all other keys we have across the codebase, and it doesn't look like we have an established pattern for `name` argument. Reminder that it doesn't change runtime (it can be renamed at any time), it's only used for logging. Then I thought that it might be easier for logs to see a key named similarly to the standardized attribute for easier correlation of what this key is for. 

> technically the OTEL implementations need to pick their own attribute name anyways when actually storing it on an otel span, right?

I thought we will use the name defined by semantic conventions for both metrics and spans.

PR #3295, opened 2025-07-16
Add `ConnectionInfo.connectionId()`

idelpivnitskiy:
#### Motivation

It's a frequent case that users would like to parse connection id from `ConnectionInfo.toString()` to correlate it with other logs and events. We can simplify their life if we provide a dedicated method to grab id.

#### Modifications

- Add `ConnectionInfo.connectionId()` with default implementation.
- Implement `ConnectionInfo.connectionId()` for all existing implementations.
- Enhance `ConnectionContextToStringTest` for http and grpc to assert expected behavior.

#### Result

Users can grab connection id from `ConnectionInfo` instead of parsing `ConnectionInfo.toString()`.

### DIFF
--- a/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionInfo.java
+++ b/servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionInfo.java
@@ -46,10 +46,7 @@ public interface ConnectionInfo {
      * @return String representation of an identifier for this connection (can be globally non-unique).
      */
     default String connectionId() {
-        final long hashCode = System.identityHashCode(this);
-        final int high = (int) (hashCode >>> 32);
-        final int low = (int) (hashCode & 0xFFFFFFFFL);
-        return String.format("0x%08x", high ^ low);
+        return String.format("0x%08x", System.identityHashCode(this));
     }
 
     /**


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionInfo.java
```diff
@@ -25,6 +25,33 @@
...
+     * uniqueness can be ensured by using a combination of the current identifier with {@link #localAddress()} and
+     * {@link #remoteAddress()}.
+     *
+     * @return String representation of an identifier for this connection (can be globally non-unique).
+     */
+    default String connectionId() {
+        final long hashCode = System.identityHashCode(this);
+        final int high = (int) (hashCode >>> 32);
+        final int low = (int) (hashCode & 0xFFFFFFFFL);
+        return String.format("0x%08x", high ^ low);
```
bryce-anderson: I'm not sure what's going on here. I believe `System.identityHashCode(..)` returns an `int`. That means the `high` part will always be 0 and you could just do `String.format("0x%08x", System.identityHashCode(this))`;

bryce-anderson (approved):
Nice improvement.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionInfo.java
idelpivnitskiy: Good catch! For some reason, I thought hashCode is always `long`

PR #3296, opened 2025-07-17
grpc opentelemetry module

aleqi200:
#### Motivation
The existing Opentelemetry integration treats GRPC as an HTTP metric/tag. There's a convention for RPC/grpc described here https://opentelemetry.io/docs/specs/semconv/rpc/grpc/

#### Modifications
- Created a new module for opentelemetry grpc as a lifecycle observer for client/server

#### Result
- A new module that have metrics/span for RPC conventions
- This uses incubation APIS which can be unstable


aleqi200 (commented):
LGTM

servicetalk-opentelemetry-http/build.gradle
```diff
@@ -58,9 +65,69 @@ dependencies {
   testImplementation "org.mockito:mockito-junit-jupiter:$mockitoCoreVersion"
   testImplementation "org.slf4j:slf4j-api:$slf4jVersion"
 
+  // gRPC test dependencies
+  testImplementation enforcedPlatform("io.grpc:grpc-bom:$grpcVersion")
```
idelpivnitskiy: Looks like this configuration was copied from `servicetalk-grpc-netty` module without considering what we actually need here and what not. We don't need Google's gRPC here.

servicetalk-opentelemetry-http/build.gradle
```diff
@@ -58,9 +65,69 @@ dependencies {
...
+  testImplementation 'com.google.guava:guava:33.3.1-android'
+  testImplementation 'com.google.protobuf:protobuf-java:3.25.5'
```
idelpivnitskiy: `protobuf-java` version should be supplied via `protobuf-bom`

servicetalk-opentelemetry-http/build.gradle
```diff
@@ -58,9 +65,69 @@ dependencies {
...
+  testImplementation 'io.grpc:grpc-api:1.69.0'
+  testImplementation 'io.opentelemetry:opentelemetry-sdk-common:1.48.0'
```
idelpivnitskiy: This already comes from `opentelemetry-bom`, we should not have any hardcoded versions in `build.gradle` files

servicetalk-opentelemetry-http/build.gradle
```diff
@@ -58,9 +65,69 @@ dependencies {
...
+
+        task.inputs
+                .file(pluginJar)
+                .withNormalizer(ClasspathNormalizer)
+                .withPropertyName("servicetalkPluginJar")
+                .withPathSensitivity(PathSensitivity.RELATIVE)
+      }
+
+      task.plugins {
+        grpc {}
```
idelpivnitskiy: We don't need Google's gRPC here. Here is what we need:
```patch
Index: servicetalk-opentelemetry-http/build.gradle
IDEA additional info:
<+>UTF-8
diff --git a/servicetalk-opentelemetry-http/build.gradle b/servicetalk-opentelemetry-http/build.gradle
--- a/servicetalk-opentelemetry-http/build.gradle	(revision 75b813468cf6917badcb6835090eb94315fedd8c)
+++ b/servicetalk-opentelemetry-http/build.gradle	(date 1753473280303)
@@ -41,19 +41,27 @@
   implementation "io.opentelemetry:opentelemetry-context"
 
   testImplementation enforcedPlatform("com.fasterxml.jackson:jackson-bom:$jacksonVersion")
+  testImplementation enforcedPlatform("com.google.protobuf:protobuf-bom:$protobufVersion")
   testImplementation enforcedPlatform("org.junit:junit-bom:$junit5Version")
   testImplementation testFixtures(project(":servicetalk-concurrent-internal"))
   testImplementation testFixtures(project(":servicetalk-http-netty"))
   testImplementation testFixtures(project(":servicetalk-log4j2-mdc-utils"))
   testImplementation testFixtures(project(":servicetalk-transport-netty-internal"))
-  testImplementation project(':servicetalk-client-api')
+  testImplementation project(":servicetalk-data-protobuf")
+  testImplementation project(":servicetalk-encoding-api")
+  testImplementation project(":servicetalk-grpc-api")
+  testImplementation project(":servicetalk-grpc-netty")
+  testImplementation project(":servicetalk-client-api")
   testImplementation project(":servicetalk-data-jackson")
   testImplementation project(":servicetalk-http-netty")
+  testImplementation project(":servicetalk-router-api")
   testImplementation project(":servicetalk-serializer-api")
   testImplementation project(":servicetalk-test-resources")
   testImplementation project(":servicetalk-utils-internal")
   testImplementation "com.fasterxml.jackson.core:jackson-annotations"
   testImplementation "com.fasterxml.jackson.core:jackson-databind"
+  testImplementation "com.google.protobuf:protobuf-java"
+  testImplementation "io.opentelemetry:opentelemetry-sdk-common"
   testImplementation "io.opentelemetry:opentelemetry-sdk-trace"
   testImplementation "io.opentelemetry:opentelemetry-sdk-testing"
   testImplementation "io.opentelemetry.semconv:opentelemetry-semconv"
@@ -65,27 +73,9 @@
   testImplementation "org.mockito:mockito-junit-jupiter:$mockitoCoreVersion"
   testImplementation "org.slf4j:slf4j-api:$slf4jVersion"
 
-  // gRPC test dependencies
-  testImplementation enforcedPlatform("io.grpc:grpc-bom:$grpcVersion")
-  testImplementation 'com.google.guava:guava:33.3.1-android'
-  testImplementation 'com.google.protobuf:protobuf-java:3.25.5'
-  testImplementation 'io.grpc:grpc-api:1.69.0'
-  testImplementation 'io.opentelemetry:opentelemetry-sdk-common:1.48.0'
-  testImplementation "io.grpc:grpc-protobuf"
-  testImplementation "io.grpc:grpc-stub"
-  testImplementation "io.opentelemetry.semconv:opentelemetry-semconv"
-  testImplementation project(':servicetalk-encoding-api')
-  testImplementation project(':servicetalk-router-api')
-  testImplementation project(":servicetalk-grpc-api")
-  testImplementation project(":servicetalk-grpc-netty")
-  testImplementation project(":servicetalk-data-protobuf")
-  testImplementation project(":servicetalk-grpc-protobuf")
-
-  testRuntimeOnly enforcedPlatform("org.apache.logging.log4j:log4j-bom:$log4jVersion")
   testRuntimeOnly project(":servicetalk-opentelemetry-asynccontext")
   testRuntimeOnly "io.opentelemetry.instrumentation:opentelemetry-log4j-context-data-2.17-autoconfigure:" +
           "$opentelemetryInstrumentationVersion-alpha"
-  testRuntimeOnly "org.apache.logging.log4j:log4j-slf4j-impl"
 }
 
 protobuf {
@@ -99,16 +89,18 @@
   //// REMOVE if outside of ServiceTalk gradle project
 
   plugins {
-    grpc {
-      artifact = "io.grpc:protoc-gen-grpc-java:$grpcVersion"
-    }
     servicetalk_grpc {
-      // use gradle project local grpc-protoc dependency
+      //// REMOVE if outside of ServiceTalk gradle project - use "artifact" as demonstrated below
+      //// "path" is used only because we want to use the gradle project local version of the plugin.
       path = pluginJar.path
+      //// REMOVE if outside of ServiceTalk gradle project - use "artifact" as demonstrated below
+
+      // artifact = "io.servicetalk:servicetalk-grpc-protoc:$serviceTalkVersion:all@jar"
     }
   }
   generateProtoTasks {
     all().each { task ->
+      //// REMOVE if outside of ServiceTalk gradle project
       io.servicetalk.gradle.plugin.internal.ProjectUtils.addGeneratedProtoDependsOn(project, task, true)
       if (task.isTest) {
         task.dependsOn(":servicetalk-grpc-protoc:buildExecutable") // use gradle project local grpc-protoc dependency
@@ -119,13 +111,12 @@
                 .withPropertyName("servicetalkPluginJar")
                 .withPathSensitivity(PathSensitivity.RELATIVE)
       }
+      //// REMOVE if outside of ServiceTalk gradle project
 
       task.plugins {
-        grpc {}
         servicetalk_grpc {
           outputSubDir = "java"
-          // this will eventually become the default behavior
-          option "defaultServiceMethods=true"
+          option "skipDeprecated=true"
         }
       }
     }
```

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/AbstractOpenTelemetryFilter.java
```diff
@@ -28,12 +28,14 @@
...
 import io.servicetalk.http.api.StreamingHttpResponse;
 
+import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 
 abstract class AbstractOpenTelemetryFilter implements HttpExecutionStrategyInfluencer {
     static final OpenTelemetryOptions DEFAULT_OPTIONS = new OpenTelemetryOptions.Builder().build();
     static final String INSTRUMENTATION_SCOPE_NAME = "io.servicetalk";
+    static final AttributeKey<String> PEER_SERVICE = AttributeKey.stringKey("peer.service");
```
idelpivnitskiy: Consider adding a link to official description of this attribute: https://opentelemetry.io/docs/specs/semconv/registry/attributes/peer/

After reading their description, tbh I don't understand why this attribute is valuable if it's equal to `service.name` 🤔

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestInfo.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Nullable
+    SocketAddress getLocalAddress() {
+        return connectionInfo != null ? connectionInfo.localAddress() : null;
+    }
+
+    @Nullable
+    SocketAddress getRemoteAddress() {
+        return connectionInfo != null ? connectionInfo.remoteAddress() : null;
+    }
```
idelpivnitskiy: Not used and can be safely removed:
```suggestion
```

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ScopeTracker.java
```diff
@@ -27,95 +25,37 @@
...
-        responseCompleteCause = throwable;
-        if (STATE_UPDATER.compareAndSet(this, IDLE, RESPONSE_COMPLETE)) {
-            // nothing to do: it's up to the request to finish now.
-        } else if (STATE_UPDATER.compareAndSet(this, REQUEST_COMPLETE, FINISHED)) {
-            instrumenter.end(context, requestMetaData, responseMetaData, throwable);
-        }
-    }
-
-    Single<StreamingHttpResponse> track(Single<StreamingHttpResponse> responseSingle) {
+    public Single<StreamingHttpResponse> track(Single<StreamingHttpResponse> responseSingle) {
```
idelpivnitskiy: Keep it pkg-private:
```suggestion
    Single<StreamingHttpResponse> track(Single<StreamingHttpResponse> responseSingle) {
```

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ScopeTracker.java
```diff
@@ -27,95 +25,37 @@
...
-
-    private final Context context;
-    private final HttpRequestMetaData requestMetaData;
-    private final Instrumenter<HttpRequestMetaData, HttpResponseMetaData> instrumenter;
-
-    @Nullable
-    private Throwable responseCompleteCause;
-    private volatile int state;
-
+final class ScopeTracker extends AbstractScopeTracker<HttpResponseMetaData> {
```
idelpivnitskiy: Should we name it `HttpScopeTracker`?

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ScopeTracker.java
```diff
@@ -27,95 +25,37 @@
...
 
-    void onResponseMeta(final HttpResponseMetaData metaData) {
-        this.responseMetaData = metaData;
-    }
-
     @Override
-    public void onComplete() {
-        assert responseMetaData != null : "can't have succeeded without capturing metadata first";
-        responseFinished(null);
+    protected void finishSpan(@Nullable Throwable error) {
```
idelpivnitskiy:
```suggestion
    void finishSpan(@Nullable Throwable error) {
```

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcScopeTracker.java
```diff
@@ -0,0 +1,108 @@
...
+        // same that the users see.
+        return responseSingle
+                // We want to make sure we capture our response and add our trailers listener first.
+                .map(this::captureTrailers)
+                .liftSync(new AfterFinallyHttpOperator(this, true));
+    }
+
+    private StreamingHttpResponse captureTrailers(StreamingHttpResponse response) {
+        this.responseMetaData = response;
+        return response.transform(new TrailersCapturingTransformer());
```
idelpivnitskiy: `transform` is intended for use-cases that want to modify trailers or for users who prefer convenience, as it comes with side-effects of resetting internal flags and de-optimizing certain paths. We should use `transformMessageBody` instead and look for `HttpHeaders` object in that publisher instead, as it gives access to what we need without affecting anything else

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/AbstractScopeTracker.java
```diff
@@ -0,0 +1,117 @@
...
+
+    protected final Context context;
+    protected final RequestInfo requestInfo;
+    protected final Instrumenter<RequestInfo, T> instrumenter;
+
+    @Nullable
+    protected Throwable responseCompleteCause;
+    private volatile int state;
+
+    protected AbstractScopeTracker(boolean isClient, Context context, StreamingHttpRequest request,
```
idelpivnitskiy: Prefer the minimal possible scope. Most items on this class can be private, some pkg-private, no need in protected or public modifiers

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/AbstractScopeTracker.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param responseSingle the response Single to track
+     * @return the tracked response Single
+     */
+    public abstract Single<StreamingHttpResponse> track(Single<StreamingHttpResponse> responseSingle);
+
+    private static final class CancelledRequestException extends Exception {
+        private static final long serialVersionUID = 6357694797622093267L;
+        static final CancelledRequestException INSTANCE = new CancelledRequestException();
+
+        CancelledRequestException() {
```
idelpivnitskiy:
```suggestion
        private CancelledRequestException() {
```

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/AbstractScopeTracker.java
```diff
@@ -0,0 +1,117 @@
...
+     * <p>
+     * Implementations should set up response metadata capture and apply the necessary
+     * operators to track the response lifecycle.
+     *
+     * @param responseSingle the response Single to track
+     * @return the tracked response Single
+     */
+    public abstract Single<StreamingHttpResponse> track(Single<StreamingHttpResponse> responseSingle);
+
+    private static final class CancelledRequestException extends Exception {
```
idelpivnitskiy: optional or for follow-up: `java.util.concurrent.CancellationException` is well known for this type of use-cases. I understand that you can not use it directly because it does not allow to disable suppressions or stack-trace, but we can at least make the name consistent. Consider names like `RequestCancellationException` or `ExchangeCancellationException`

servicetalk-opentelemetry-http/src/test/proto/test_service.proto
```diff
@@ -0,0 +1,37 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+option java_multiple_files = false;
```
idelpivnitskiy: it's much easier to read and understand generated code when this flag is set to `true`

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServiceTalkHttpAttributesGetter.java
```diff
@@ -143,20 +148,20 @@ public String getUrlFull(final HttpRequestMetaData request) {
...
-        public Integer getServerPort(HttpRequestMetaData metaData) {
-            final HostAndPort effectiveHostAndPort = metaData.effectiveHostAndPort();
+        public Integer getServerPort(RequestInfo requestInfo) {
+            final HostAndPort effectiveHostAndPort = requestInfo.getMetadata().effectiveHostAndPort();
             if (effectiveHostAndPort != null) {
                 return effectiveHostAndPort.port();
             }
             // No port. See if we can infer it from the scheme.
-            String scheme = metaData.scheme();
+            String scheme = requestInfo.getMetadata().scheme();
```
idelpivnitskiy: (can be follow-up) Now that we may optionally have `ConnectionInfo`, should we prefer extraction info from there (if available) and fallback to request meta-data if not?

servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryGrpcServiceFilterTest.java
```diff
@@ -0,0 +1,287 @@
...
+
+        // Create gRPC client with unified OpenTelemetry HTTP requester filter
+        // The filter will automatically detect gRPC requests and handle them appropriately
+        client = GrpcClients.forAddress(serverHostAndPort(serverContext))
+                .initializeHttp(builder -> builder.appendClientFilter(new OpenTelemetryHttpRequesterFilter(
+                        otelTesting.getOpenTelemetry(), "test-client",
+                        new OpenTelemetryOptions.Builder().build())))
+                .build(new ClientFactory());
+
+        blockingClient = GrpcClients.forAddress(serverHostAndPort(serverContext))
```
idelpivnitskiy: Do you need 2 different client instances? Alternatively, you can convert existing `client` using `asBlockingClient()` method

servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryGrpcServiceFilterTest.java
```diff
@@ -0,0 +1,287 @@
...
+                .initializeHttp(builder -> builder.appendClientFilter(new OpenTelemetryHttpRequesterFilter(
+                        otelTesting.getOpenTelemetry(), "test-client",
+                        new OpenTelemetryOptions.Builder().build())))
+                .buildBlocking(new ClientFactory());
+    }
+
+    private SpanData findSpanByKind(SpanKind kind) {
```
idelpivnitskiy:
```suggestion
    private static SpanData findSpanByKind(SpanKind kind) {
```

servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryGrpcServiceFilterTest.java
```diff
@@ -0,0 +1,287 @@
...
+        // Verify client span
+        SpanData clientSpan = findSpanByKind(SpanKind.CLIENT);
+        assertThat(clientSpan.getName()).isEqualTo("opentelemetry.grpc.Tester/test");
+        assertThat(clientSpan.getInstrumentationScopeInfo().getName()).isEqualTo("io.servicetalk");
+
+        // Verify they're part of the same trace
+        assertThat(serverSpan.getTraceId()).isEqualTo(clientSpan.getTraceId());
+
+        // Verify gRPC attributes are set correctly
+        assertThat(serverSpan.getAttributes().get(AttributeKey.stringKey("rpc.system")))
```
idelpivnitskiy: Should we also assert `network.*` attributes? https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/

servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryGrpcServiceFilterTest.java
```diff
@@ -0,0 +1,287 @@
...
+                .isEqualTo("grpc");
+        assertThat(serverSpan.getAttributes().get(AttributeKey.stringKey("rpc.service")))
+                .isEqualTo("opentelemetry.grpc.Tester");
+        assertThat(serverSpan.getAttributes().get(AttributeKey.stringKey("rpc.method")))
+                .isEqualTo("test");
+        assertThat(serverSpan.getAttributes().get(AttributeKey.longKey("rpc.grpc.status_code")))
```
idelpivnitskiy: Consider moving these assertions into a method and reusing them for all other request API variants

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestHeadersPropagatorSetter.java
```diff
@@ -16,23 +16,22 @@
...
-    static final TextMapSetter<HttpRequestMetaData> INSTANCE = new RequestHeadersPropagatorSetter();
+    static final TextMapSetter<RequestInfo> INSTANCE = new RequestHeadersPropagatorSetter();
 
     private RequestHeadersPropagatorSetter() {
     }
 
     @Override
-    public void set(@Nullable final HttpRequestMetaData carrier, final String key, final String value) {
-        if (carrier != null) {
-            HeadersPropagatorSetter.INSTANCE.set(carrier.headers(), key, value);
```
idelpivnitskiy: If `HeadersPropagatorSetter` is not used anymore, consider removing it

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestHeadersPropagatorGetter.java
```diff
@@ -16,30 +16,29 @@
...
     }
 
     @Override
     @Nullable
-    public String get(@Nullable HttpRequestMetaData carrier, final String key) {
+    public String get(@Nullable RequestInfo carrier, final String key) {
         if (carrier == null) {
             return null;
         }
-        return HeadersPropagatorGetter.INSTANCE.get(carrier.headers(), key);
```
idelpivnitskiy: Why not using `HeadersPropagatorGetter` consistently for both cases?

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServiceTalkHttpAttributesGetter.java
```diff
@@ -36,58 +39,59 @@
...
             final HttpResponseMetaData httpResponseMetaData,
             final String name) {
         return getHeaderValues(httpResponseMetaData.headers(), name);
     }
 
     @Override
     public final String getNetworkProtocolName(
-            final HttpRequestMetaData request, @Nullable final HttpResponseMetaData response) {
+            final RequestInfo request, @Nullable final HttpResponseMetaData response) {
         return HTTP_SCHEME;
```
idelpivnitskiy: Unrelated, but need to capture so that we don't forget: the spec says that this attribute should be present only if it's not `http`: https://opentelemetry.io/docs/specs/semconv/http/http-spans/

On the other hand, we should add other `network.*` attributes that we currently don't have

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestHeadersPropagatorGetter.java
```diff
@@ -16,30 +16,29 @@
...
 package io.servicetalk.opentelemetry.http;
 
-import io.servicetalk.http.api.HttpRequestMetaData;
-
 import io.opentelemetry.context.propagation.TextMapGetter;
 
 import javax.annotation.Nullable;
 
-final class RequestHeadersPropagatorGetter implements TextMapGetter<HttpRequestMetaData> {
+final class RequestHeadersPropagatorGetter implements TextMapGetter<RequestInfo> {
```
idelpivnitskiy: Would be great if all HTTP related classes will start with `Http` prefix to make it clear in comparison to `Grpc*` classes

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcCapturedHeadersExtractor.java
```diff
@@ -0,0 +1,103 @@
...
+        // Check both response headers and trailers for metadata
+        if (response != null) {
+            for (CapturedHeader entry : capturedResponseHeaders) {
+                // First check trailers (more likely for gRPC metadata)
+                CharSequence headerValue = null;
+                if (response.getTrailers() != null) {
+                    headerValue = response.getTrailers().get(entry.headerName);
+                }
+                // Fallback to response headers if not found in trailers
+                if (headerValue == null && response.getResponseMetaData() != null) {
```
idelpivnitskiy: Taking into account that users configured capturedResponse**Headers** on the `OpenTelemetryOptions`, I think we should look only at headers. Maybe we can also look at trailers, but only as a fallback. But if we do, then why we don't look at trailers for HTTP as well?

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcClientAttributesExtractor.java
```diff
@@ -0,0 +1,78 @@
...
+    public void onEnd(AttributesBuilder attributes, Context context, RequestInfo request,
+                      @Nullable GrpcTelemetryStatus response, @Nullable Throwable error) {
+        // Apply pure gRPC/RPC semantic conventions
+        GrpcSemanticAttributesExtractor.INSTANCE.onEnd(attributes, context, request, response, error);
+
+        // Handle captured headers
+        capturedHeadersExtractor.onEnd(attributes, context, request, response, error);
+    }
+
+    private void addClientServerAttributes(AttributesBuilder attributes, RequestInfo request) {
```
idelpivnitskiy:
```suggestion
    private static void addClientServerAttributes(AttributesBuilder attributes, RequestInfo request) {
```
hint: there is `Method can be made 'static'` inspection in IntelliJ IDEA to help capture these cases

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcClientAttributesExtractor.java
```diff
@@ -0,0 +1,78 @@
...
+        // Add client-specific server address/port if available
+        addClientServerAttributes(attributes, request);
+
+        // Handle captured headers
+        capturedHeadersExtractor.onStart(attributes, parentContext, request);
+    }
+
+    @Override
+    public void onEnd(AttributesBuilder attributes, Context context, RequestInfo request,
+                      @Nullable GrpcTelemetryStatus response, @Nullable Throwable error) {
```
idelpivnitskiy: here and in other places the class name does not match with param name, can we find a way to align?

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSemanticAttributesExtractor.java
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Extractor for gRPC semantic convention attributes using stable OpenTelemetry APIs.
+ * <p>
+ * This class manually implements RPC semantic conventions as defined by OpenTelemetry,
+ * avoiding dependencies on alpha/incubator APIs.
+ */
+final class GrpcSemanticAttributesExtractor implements AttributesExtractor<RequestInfo, GrpcTelemetryStatus> {
+
+    // RPC semantic convention attribute keys
```
idelpivnitskiy, line 41: Consider linking to https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/ and https://opentelemetry.io/docs/specs/semconv/rpc/grpc/ either here or in javadoc

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSemanticAttributesExtractor.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final AttributeKey<String> RPC_SYSTEM = AttributeKey.stringKey("rpc.system");
+    private static final AttributeKey<String> RPC_SERVICE = AttributeKey.stringKey("rpc.service");
+    private static final AttributeKey<String> RPC_METHOD = AttributeKey.stringKey("rpc.method");
+    private static final AttributeKey<Long> RPC_GRPC_STATUS_CODE = AttributeKey.longKey("rpc.grpc.status_code");
+
+    static final AttributeKey<String> NETWORK_PEER_ADDRESS = AttributeKey.stringKey("network.peer.address");
+    static final AttributeKey<Long> NETWORK_PEER_PORT = AttributeKey.longKey("network.peer.port");
```
idelpivnitskiy:
```suggestion
    private static final AttributeKey<String> NETWORK_PEER_ADDRESS = AttributeKey.stringKey("network.peer.address");
    private static final AttributeKey<Long> NETWORK_PEER_PORT = AttributeKey.longKey("network.peer.port");
```

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSemanticAttributesExtractor.java
```diff
@@ -0,0 +1,115 @@
...
+    public void onEnd(AttributesBuilder attributes, Context context, RequestInfo request,
+                      @Nullable GrpcTelemetryStatus response, @Nullable Throwable error) {
+        // Extract gRPC status code from response headers/trailers (trailers first per gRPC spec)
+        if (response != null) {
+            extractGrpcStatusCode(response, attributes);
+        }
+    }
+
+    private static void extractServiceAndMethod(String path, AttributesBuilder attributes) {
+        if (path == null || path.isEmpty()) {
```
idelpivnitskiy: This is a non-null attribute, null check can be removed

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSemanticAttributesExtractor.java
```diff
@@ -0,0 +1,115 @@
...
+            return;
+        }
+
+        // gRPC path format: /service.name/MethodName
+        // Remove leading slash if present
+        String normalizedPath = path.startsWith("/") ? path.substring(1) : path;
```
idelpivnitskiy: The first character can be checked with `path.charAt(0) == '/'`. Also, consider using indexes to avoid an extra `substring` here, you can just do `substring(1, lastSlashIndex)` later when you extract `service`

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSemanticAttributesExtractor.java
```diff
@@ -0,0 +1,115 @@
...
+ * This class manually implements RPC semantic conventions as defined by OpenTelemetry,
+ * avoiding dependencies on alpha/incubator APIs.
+ */
+final class GrpcSemanticAttributesExtractor implements AttributesExtractor<RequestInfo, GrpcTelemetryStatus> {
+
+    // RPC semantic convention attribute keys
+    private static final AttributeKey<String> RPC_SYSTEM = AttributeKey.stringKey("rpc.system");
+    private static final AttributeKey<String> RPC_SERVICE = AttributeKey.stringKey("rpc.service");
+    private static final AttributeKey<String> RPC_METHOD = AttributeKey.stringKey("rpc.method");
+    private static final AttributeKey<Long> RPC_GRPC_STATUS_CODE = AttributeKey.longKey("rpc.grpc.status_code");
```
idelpivnitskiy, line 52: Is there any difference in published spans based on an attribute type or all of them look like strings anyway? Even though the spec defines this attribute of type `int`, if it doesn't affect anything else we can keep it as `string` to avoid overhead of conversion and boxing

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcTelemetryStatus.java
```diff
@@ -0,0 +1,105 @@
...
+     * @return parsed gRPC status code as long, or Long.MIN_VALUE if not found or invalid
+     */
+    private long parseGrpcStatusCode() {
+        String statusString = null;
+
+        // Check trailers first (gRPC HTTP/2 specification compliance)
+        if (trailers != null) {
+            CharSequence statusFromTrailers = trailers.get(GRPC_STATUS_HEADER_NAME);
+            if (statusFromTrailers != null) {
+                statusString = statusFromTrailers.toString();
```
idelpivnitskiy: If you still decide to stick to `long`, there is no need to convert to `String`, we have `CharSequences.parseLong` method to avoid allocations in this case

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSemanticAttributesExtractor.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private static void extractPeerAddress(RequestInfo requestInfo, AttributesBuilder attributes) {
+        ConnectionInfo connectionInfo = requestInfo.getConnectionInfo();
+        if (connectionInfo != null) {
+            SocketAddress peerResolvedAddress = connectionInfo.remoteAddress();
+            if (peerResolvedAddress instanceof InetSocketAddress) {
+                InetSocketAddress inetSocketAddress = (InetSocketAddress) peerResolvedAddress;
+                attributes.put(
+                        NETWORK_PEER_ADDRESS, inetSocketAddress.getAddress().getHostAddress());
+                attributes.put(NETWORK_PEER_PORT, inetSocketAddress.getPort());
```
idelpivnitskiy: We can also extract [network.transport](https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/) and [network.type](https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/) in this case.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSemanticAttributesExtractor.java
```diff
@@ -0,0 +1,115 @@
...
+            } else {
+                // This is unlikely since the resolved form is almost always an `InetSocketAddress`, and
```
idelpivnitskiy: Note that ST also supports Unix Domain Sockets. If address is an instance of `io.servicetalk.transport.api.DomainSocketAddress`, we should record [network.transport](https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/) as `unix` instead of `tcp`

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcServerAttributesExtractor.java
```diff
@@ -0,0 +1,60 @@
...
+        );
+    }
+
+    @Override
+    public void onStart(AttributesBuilder attributes, Context parentContext, RequestInfo request) {
+        // Apply pure gRPC/RPC semantic conventions
+        GrpcSemanticAttributesExtractor.INSTANCE.onStart(attributes, parentContext, request);
+
+        // Handle captured headers
+        capturedHeadersExtractor.onStart(attributes, parentContext, request);
```
idelpivnitskiy: It looks like we can make `GrpcSemanticAttributesExtractor` an abstract class that will contain almost all logic, with a small difference in `GrpcClientAttributesExtractor.onStart`

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSpanNameExtractor.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public String extract(RequestInfo request) {
+        String path = request.getMetadata().path();
+        if (path == null || path.isEmpty()) {
+            return "grpc.request";
+        }
+
+        // gRPC path format: /service.name/MethodName
+        // Remove leading slash for span name
+        if (path.startsWith("/")) {
```
idelpivnitskiy: it has cheaper runtime cost:
```suggestion
        if (path.charAt(0) == '/') {
```

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSpanNameExtractor.java
```diff
@@ -0,0 +1,49 @@
...
+
+    static final SpanNameExtractor<RequestInfo> INSTANCE = new GrpcSpanNameExtractor();
+
+    private GrpcSpanNameExtractor() {
+    }
+
+    @Override
+    public String extract(RequestInfo request) {
+        String path = request.getMetadata().path();
+        if (path == null || path.isEmpty()) {
```
idelpivnitskiy: Here and in all other places for grpc: consider switching to `requestTarget()` instead of `path()` which avoids expensive parsing of the internal `requestTarget` value (it's known that for gRPC we don't expect any other format) and skip null checks

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSpanStatusExtractor.java
```diff
@@ -0,0 +1,80 @@
...
+            // - UNIMPLEMENTED (12)
+            // - INTERNAL (13)
+            // - UNAVAILABLE (14)
+            // - DATA_LOSS (15)
+            // - UNAUTHENTICATED (16)
+
+            if (grpcStatusCode == 0) {
+                // OK (0) - leave span status unset (default OK)
+            } else {
+                // Any non-OK gRPC status is considered an error
```
idelpivnitskiy: Look at https://opentelemetry.io/docs/specs/semconv/rpc/grpc/#grpc-status, it says that on the server-side some statuses should not set error flag

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcTelemetryStatus.java
```diff
@@ -0,0 +1,105 @@
...
+
+    GrpcTelemetryStatus(@Nullable HttpResponseMetaData responseMetaData,
+                        @Nullable HttpHeaders trailers) {
+        this.responseMetaData = responseMetaData;
+        this.trailers = trailers;
+        this.statusCode = parseGrpcStatusCode();
+    }
+
+    @Nullable
+    HttpResponseMetaData getResponseMetaData() {
```
idelpivnitskiy: Comment for all other places: in ST codebase we do not use `get` prefix

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcTelemetryStatus.java
```diff
@@ -0,0 +1,105 @@
...
+package io.servicetalk.opentelemetry.http;
+
+import io.servicetalk.http.api.HttpHeaders;
+import io.servicetalk.http.api.HttpResponseMetaData;
+
+import javax.annotation.Nullable;
+
+final class GrpcTelemetryStatus {
+
+    private static final String GRPC_STATUS_HEADER_NAME = "grpc-status";
```
idelpivnitskiy: We should use `CharSequence GRPC_STATUS = newAsciiString("grpc-status")` for more efficient internal comparison

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/HttpInstrumentationHelper.java
```diff
@@ -0,0 +1,186 @@
...
+
+        return new HttpInstrumentationHelper(instrumenter, true);
+    }
+
+    private static final class DeferredHttpClientAttributesExtractor implements
+            AttributesExtractor<RequestInfo, io.servicetalk.http.api.HttpResponseMetaData> {
+
+        private final AttributesExtractor<RequestInfo, io.servicetalk.http.api.HttpResponseMetaData> delegate;
+
+        private DeferredHttpClientAttributesExtractor(OpenTelemetryOptions openTelemetryOptions) {
```
idelpivnitskiy:
```suggestion
        DeferredHttpClientAttributesExtractor(OpenTelemetryOptions openTelemetryOptions) {
```
bcz it's used outside of itself

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSpanStatusExtractor.java
```diff
@@ -0,0 +1,80 @@
...
+    private GrpcSpanStatusExtractor() {
+    }
+
+    @Override
+    public void extract(
+            SpanStatusBuilder spanStatusBuilder,
+            RequestInfo requestInfo,
+            @Nullable GrpcTelemetryStatus response,
+            @Nullable Throwable error) {
+        if (error != null) {
```
idelpivnitskiy: Another case for an error is a non-200 HTTP response status. grpc spec itself allows 415 response code + buggy proxies may return 5xx when they handle grpc 🫠

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcInstrumentationHelper.java
```diff
@@ -0,0 +1,207 @@
...
+        }
+        return true;
+    }
+
+    private static final class DeferredGrpcClientAttributesExtractor implements
+            AttributesExtractor<RequestInfo, GrpcTelemetryStatus> {
+
+        private final AttributesExtractor<RequestInfo, GrpcTelemetryStatus> delegate;
+
+        private DeferredGrpcClientAttributesExtractor(OpenTelemetryOptions openTelemetryOptions) {
```
idelpivnitskiy:
```suggestion
        DeferredGrpcClientAttributesExtractor(OpenTelemetryOptions openTelemetryOptions) {
```

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcInstrumentationHelper.java
```diff
@@ -0,0 +1,207 @@
...
+        if (seqLength < GRPC_CONTENT_TYPE.length()) {
+            return false;
+        }
+        if (seqLength == GRPC_CONTENT_TYPE.length()) {
+            return GRPC_CONTENT_TYPE.contentEquals(charSequence);
+        }
+        // Start at the end since many content types start with 'application/' so we're more likely
+        // to be able to abort early by checking that the prefix ends with 'grpc'.
+        for (int i = GRPC_CONTENT_TYPE.length() - 1; i >= 0; i--) {
+            if (charSequence.charAt(i) != GRPC_CONTENT_TYPE.charAt(i)) {
```
idelpivnitskiy: I'm afraid the content-type values are case-insensitive. For that we have handy utilities in `CharSequences`, like `equalsIgnoreCaseLower` for chars and `contentEqualsIgnoreCase` for `CharSequence`. If you go that route, it will be worth switching to `newAsciiString` for `GRPC_CONTENT_TYPE` constant

idelpivnitskiy:
With that many comments it will be handy if you don't mark them as "resolved", leave them open for me to verify

bryce-anderson:
I'll address the comments since a lot of it came from me.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/AbstractScopeTracker.java
bryce-anderson: fwiw, this doesn't matter: the parent can access the private constructors of inner classes.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/AbstractScopeTracker.java
bryce-anderson: ExchangeCancellationException looks fine.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServiceTalkHttpAttributesGetter.java
bryce-anderson: We can do it now.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/ServiceTalkHttpAttributesGetter.java
bryce-anderson: The first part is already taken care of by the instrumentation-api: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/semconv/http/HttpCommonAttributesExtractor.java#L121

As for the latter, I'll add them here since I'm also refactoring the `RequestInfo` aspect.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/RequestHeadersPropagatorGetter.java
bryce-anderson: Broader point taken, but this particular class is used by both http and grpc.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcCapturedHeadersExtractor.java
bryce-anderson: gRPC tends to have materially different semantics than HTTP despite using it as the underlying transport (eg grpc-status header) so I can see a case, but we can start with the same behavior for now and adjust later if we want.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcClientAttributesExtractor.java
bryce-anderson: I got a bunch, maybe there are more.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSemanticAttributesExtractor.java
bryce-anderson: My understanding is that it is boxed differently over the wire and therefore the type may be meaningful for the backend. Only marginally informative, attribute keys of the same name but different type do not match in the tests.

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSemanticAttributesExtractor.java
bryce-anderson: I left `network.transport` out deliberately since it seem it will essentially always be `tcp`. I'm less confident about `network.type`: does it add enough value to justify its overhead?

bryce-anderson: Based on your comment below about domain sockets, I decided to add them both.

servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryGrpcFilterTest.java
```diff
@@ -0,0 +1,284 @@
...
+        assertThat(otelTesting.getSpans()).hasSize(2); // client + server spans
+        // Verify they're part of the same trace
+        assertThat(findSpanByKind(SpanKind.CLIENT).getTraceId())
+                .isEqualTo(findSpanByKind(SpanKind.SERVER).getTraceId());
+    }
+
+    private void assertGRpcAttributes(SpanKind spanKind, String methodName, long statusCode) {
+        SpanData spanData = findSpanByKind(spanKind);
+        assertThat(spanData.getName()).isEqualTo("opentelemetry.grpc.Tester/" + methodName);
+        assertThat(spanData.getInstrumentationScopeInfo().getName()).isEqualTo("io.servicetalk");
```
idelpivnitskiy: We can use constant from `AbstractOpenTelemetryFilter`:
```suggestion
        assertThat(spanData.getInstrumentationScopeInfo().getName()).isEqualTo(INSTRUMENTATION_SCOPE_NAME);
```

servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryGrpcFilterTest.java
```diff
@@ -0,0 +1,284 @@
...
+        assertThat(spanData.getAttributes().get(AttributeKey.longKey("rpc.grpc.status_code")))
+                .isEqualTo(statusCode);
+
+        // TODO: right now we don't have a way to get the servers address from this filters position in the client.
+        if (spanKind == SpanKind.SERVER) {
+            assertThat(spanData.getAttributes().get(AttributeKey.stringKey("network.peer.address")))
+                    .isEqualTo("127.0.0.1");
+            // hard to tell what it is, but it shouldn't be null
+            assertThat(spanData.getAttributes().get(AttributeKey.longKey("network.peer.port")))
+                    .isNotNull();
```
idelpivnitskiy: It's possible that in some env the local address will use IPv6. Consider using `serverAddress` for these assertions, similar to what you do above for `server.address` and `server.port`

servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryGrpcFilterTest.java
```diff
@@ -0,0 +1,284 @@
...
+import javax.annotation.Nullable;
+
+import static io.servicetalk.concurrent.api.Single.succeeded;
+import static io.servicetalk.concurrent.internal.DeliberateException.DELIBERATE_EXCEPTION;
+import static io.servicetalk.transport.netty.internal.AddressUtils.localAddress;
+import static io.servicetalk.transport.netty.internal.AddressUtils.serverHostAndPort;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@SuppressWarnings("PMD.AvoidUsingHardCodedIP")
```
idelpivnitskiy: Then you should be able to remove this suppression

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcClientAttributesExtractor.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private static void addServerAddressAndPort(AttributesBuilder attributesBuilder, RequestInfo requestInfo) {
+        // Add server.address and server.port for client spans (per RPC semantic conventions)
+        HostAndPort hostAndPort = requestInfo.request().effectiveHostAndPort();
+        if (hostAndPort == null) {
+            // On the client side we lazily populate the attributes because adding the host header happens last.
+            // For HTTP/2, this host header will get turned into an ':authority' header by the ServiceTalk internals
+            // and that is what we typically observer after the request.
+            CharSequence authority = requestInfo.request().headers().get(AUTHORITY);
```
idelpivnitskiy, line 58: We should somehow remember to remove this path after we flip order of `HostHeaderHttpRequesterFilter`

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcInstrumentationHelper.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    // TODO: maybe there is a more efficient way to do this comparison.
+    private static boolean startsWithPrefix(CharSequence charSequence) {
+        int seqLength = charSequence.length();
+        if (seqLength < GRPC_CONTENT_TYPE.length()) {
+            return false;
+        }
+        if (seqLength == GRPC_CONTENT_TYPE.length()) {
+            return GRPC_CONTENT_TYPE.contentEquals(charSequence);
```
idelpivnitskiy: There is `CharSequences.contentEqualsIgnoreCase(...)` utility to help here

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcInstrumentationHelper.java
```diff
@@ -0,0 +1,208 @@
...
+     *
+     * @param request the HTTP request to examine
+     * @return true if this should be treated as a gRPC request, false for HTTP
+     */
+    boolean isGrpcRequest(StreamingHttpRequest request) {
+        CharSequence contentType = request.headers().get(CONTENT_TYPE);
+        return contentType != null && startsWithPrefix(contentType);
+    }
+
+    // TODO: maybe there is a more efficient way to do this comparison.
```
idelpivnitskiy: Is this TODO still relevant?

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcInstrumentationHelper.java
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * Helper class that encapsulates gRPC-specific OpenTelemetry instrumentation logic.
+ * <p>
+ * This helper handles the creation of gRPC instrumenters and provides methods to track
+ * gRPC requests with proper span lifecycle management and gRPC semantic conventions.
+ */
+final class GrpcInstrumentationHelper {
+
+    private static final String GRPC_CONTENT_TYPE = "application/grpc";
```
idelpivnitskiy: Now that you switched to `CharSequences`, using `newAsciiString` will give more benefits

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcSpanNameExtractor.java
```diff
@@ -0,0 +1,51 @@
...
+final class GrpcSpanNameExtractor implements SpanNameExtractor<RequestInfo> {
+
+    static final SpanNameExtractor<RequestInfo> INSTANCE = new GrpcSpanNameExtractor();
+
+    private GrpcSpanNameExtractor() {
+    }
+
+    @Override
+    public String extract(RequestInfo requestInfo) {
+        requestInfo.request().requestTarget();
```
idelpivnitskiy: leftover?

servicetalk-opentelemetry-http/src/main/java/io/servicetalk/opentelemetry/http/GrpcClientAttributesExtractor.java
bryce-anderson: I added a comment to `DefaultSingleAddressHttpClientBuilder` to point here to this workaround. Maybe there is a better way?

idelpivnitskiy (approved):
🚀

bryce-anderson (approved):

PR #3297, opened 2025-07-17
`TransportObserver`: let users easily skip uninterested callbacks

idelpivnitskiy:
Motivation:

`TransportObserver` and all other observers related to it, like `ConnectionObserver`, etc. have quite a lot of methods but users are not interested in all of them at the same time. However, the current interface definition forces users to maintain a lot of boilerplate code.

Modifications:
- Add default implementations for all `ConnectionObserver` methods and for methods of other related observers.
- Remove boilerplate from existing `TransportObserver` implementations.
- `DefaultNettyConnection`: skip wiring `ReadObserver` if it's noop.

Result:

`TransportObserver` implementations that are interested only in subset of callbacks will be able to remove boilerplate code.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NoopTransportObserver.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NoopTransportObserver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2020 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2020, 2025 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,7 +15,6 @@
  */
 package io.servicetalk.transport.netty.internal;
 
-import io.servicetalk.transport.api.ConnectionInfo;
 import io.servicetalk.transport.api.ConnectionObserver;
 import io.servicetalk.transport.api.ConnectionObserver.DataObserver;
 import io.servicetalk.transport.api.ConnectionObserver.MultiplexedObserver;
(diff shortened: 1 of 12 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NoopTransportObserver.java
```diff
@@ -43,260 +45,111 @@ private NoopTransportObserver() {
...
     /**
      * Noop version of {@link ConnectionObserver}.
      */
     public static final class NoopConnectionObserver implements ConnectionObserver {
 
-        public static final ConnectionObserver INSTANCE = new NoopConnectionObserver();
+        // Here and in all other INSTANCE cases below we call original methods to get the reference of the pkg-private
+        // NOOP variants from transport-api module. That way we can make sure that checking for NOOP reference can be
+        // used as a way to skip observability wiring in cases where it adds overhead.
+        public static final ConnectionObserver INSTANCE = NoopTransportObserver.INSTANCE.onNewConnection();
```
bryce-anderson, line 59: This is kind of hairy to understand. I suppose that the fallout is that we unnecessarily feed events to a noop observer so it's not so bad.

bryce-anderson (approved):
So much glorious red... A great improvement, thank you.

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/NoopTransportObserver.java
bryce-anderson: I guess I should have noted what I wanted from this: it's just a comment for now. Maybe we can find a better way in a follow up. I think you suggested we could allow returning null, but that deserves more thought and a separate PR.

idelpivnitskiy: Right, I didn't want to go with the `null` changes right here. We should find a better way later, and removal of this deprecated method will force us to do something about it

PR #3304, opened 2025-07-24
Improve javadoc for observers

idelpivnitskiy:
#### Motivation

Improve level of details for `[Http|Grpc]LifecycleObserver`s and `ConnectionObserver`.

#### Modifications

- Provide more details for interfaces and callbacks.

#### Result

Easier for implementers to understand observer behavior.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcLifecycleObserver.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcLifecycleObserver.java
@@ -20,22 +20,32 @@ import io.servicetalk.http.api.HttpHeaders;
 import io.servicetalk.http.api.HttpLifecycleObserver;
 import io.servicetalk.http.api.HttpRequestMetaData;
 import io.servicetalk.http.api.HttpResponseMetaData;
+import io.servicetalk.http.api.HttpServerBuilder;
+import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
 import io.servicetalk.http.api.StreamingHttpClientFilterFactory;
+import io.servicetalk.http.api.StreamingHttpConnectionFilterFactory;
+import io.servicetalk.http.api.StreamingHttpServiceFilterFactory;
 import io.servicetalk.transport.api.IoExecutor;
 
 /**
  * An observer interface that provides visibility into gRPC lifecycle events.
  * <p>
- * In order to deliver events at accurate time, callbacks on this interface can be invoked from the {@link IoExecutor}.
+ * To deliver events at accurate time, callbacks on this interface can be invoked from the {@link IoExecutor}.
  * Implementation of this observer <b>must</b> be non-blocking. If the
  * consumer of events may block (uses a blocking library or
- * <a href="https://logging.apache.org/log4j/2.x/manual/async.html">logger configuration is not async</a>), it has to
+ * <a href="https://logging.apache.org/log4j/2.x/manual/async.html">logger configuration is not async</a>), it must
  * offload publications to another {@link Executor} <b>after</b> capturing timing of events. If blocking code is
  * executed inside callbacks without offloading, it will negatively impact {@link IoExecutor}.
  * <p>
- * To install this observer for the server use {@link GrpcServerBuilder#lifecycleObserver(GrpcLifecycleObserver)}.
- * For the client use {@code io.servicetalk.grpc.netty.GrpcLifecycleObserverRequesterFilter} as argument to
- * {@link io.servicetalk.http.api.SingleAddressHttpClientBuilder#appendClientFilter(StreamingHttpClientFilterFactory)}
+ * To install this observer for the server use
+ * {@link GrpcServerBuilder#lifecycleObserver(GrpcLifecycleObserver) lifecycleObserver} method or
+ * {@link HttpServerBuilder#appendNonOffloadingServiceFilter(StreamingHttpServiceFilterFactory)
+ * appendNonOffloadingServiceFilter} with
+ * {@code io.servicetalk.grpc.netty.GrpcLifecycleObserverServiceFilter} which can be configured using
+ * {@link GrpcServerBuilder#initializeHttp(GrpcServerBuilder.HttpInitializer)}. For the client use either
+ * {@link SingleAddressHttpClientBuilder#appendClientFilter(StreamingHttpClientFilterFactory) appendClientFilter} or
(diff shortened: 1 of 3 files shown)


servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -217,15 +230,15 @@ default void proxyConnectComplete(Object responseMsg) {
     interface SecurityHandshakeObserver {
 
         /**
-         * Callback when the handshake is failed.
+         * Callback when the handshake fails.
```
daschl, line 246: While we are in here, I think there is an opportunity to clarify on the exact timing of the callbacks. Maybe it's because I am not a native speaker, but for me the "active" tense makes it a little unclear when the callback is invoked exactly. I am posting it here, but I think other callbacks have the same semantics (for example see the next two).

I would propose instead of `Callback when the handshake fails`: `Callback is executed after the handshake has failed`

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
```diff
@@ -169,14 +178,18 @@ default void connectionWritabilityChanged(boolean isWritable) {
 
     /**
      * Callback when the connection is closed due to an {@link Throwable error}.
+     * <p>
+     * Connections can be closed at any time, even before they are fully established.
```
daschl: Consider here and with the next docblock to `@link` and mention that the open callback will not have been called if this happens

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLifecycleObserver.java
```diff
@@ -182,26 +197,28 @@ default void onRequestCancel() {
     /**
      * An observer interface that provides visibility into events associated with a single HTTP response.
      * <p>
-     * The response is considered complete when one of the terminal events is invoked. It's guaranteed only one terminal
-     * event will be invoked per response.
+     * The response is considered complete when one of the terminal events is invoked. It is guaranteed only one
```
daschl: It is guaranteed `that` only..?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLifecycleObserver.java
```diff
@@ -117,26 +127,28 @@ default void onExchangeFinally() {
...
 
         /**
-         * Callback when subscriber requests {@code n} items of the request payload body.
+         * Callback when the subscriber requests {@code n} items of the request payload body.
          * <p>
-         * May be invoked multiple times. Helps to track when items are requested and when they are
-         * {@link #onRequestData(Buffer) delivered}.
+         * May be invoked multiple times and concurrently with other callbacks on this observer. Therefore, it should
+         * have its own isolated state or should be synchronized if the state is shared with other callbacks.
+         * It can help track when items are requested and when they are {@link #onRequestData(Buffer) delivered}.
```
daschl: It can help `to` track ..?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcLifecycleObserver.java
```diff
@@ -93,8 +104,8 @@ default void onRequestTrailers(HttpHeaders trailers) {
     /**
      * An observer interface that provides visibility into events associated with a single gRPC response.
      * <p>
-     * The response is considered complete when one of the terminal events is invoked. It's guaranteed only one terminal
-     * event will be invoked per response.
+     * The response is considered complete when one of the terminal events is invoked. It is guaranteed only one
```
daschl: It is guaranteed `that` only ?

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcLifecycleObserver.java
```diff
@@ -73,7 +84,7 @@ default GrpcResponseObserver onResponse(HttpResponseMetaData responseMetaData) {
     /**
      * An observer interface that provides visibility into events associated with a single gRPC request.
      * <p>
-     * The request is considered complete when one of the terminal events is invoked. It's guaranteed only one terminal
+     * The request is considered complete when one of the terminal events is invoked. It is guaranteed only one terminal
```
daschl: It is guaranteed that only ?

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLifecycleObserver.java
idelpivnitskiy: I originally had `to` there, but AI proofreading dropped it 😄 
Asked a different model again:
> Both "It can help track" and "It can help to track" are grammatically correct and convey the same meaning. The addition of "to" before "track" is a stylistic choice and doesn't change the fundamental meaning. In most cases, you can use either phrase interchangeably.

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
idelpivnitskiy: I got your point, but made changes in a slightly different way:
1. moved "terminal" clarification from the top of the interface to each "closed" method and clarified that any callback after that doesn't make sense
2. moved `connectionWritabilityChanged` up to read/write/flush events to make "established" and "closed" callbacks next to each other in the interface structure
3. added clarification that only one "established" callback can be invoked

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpLifecycleObserver.java
idelpivnitskiy: I added it back 😄

servicetalk-transport-api/src/main/java/io/servicetalk/transport/api/ConnectionObserver.java
idelpivnitskiy: As another non-native speaker who doesn't feel confident with grammar, I'm sure I will mess it up. WDYT if we merge as-is and then you can do a follow-up to adjust as necessary?

I would like to avoid mentioning before/after because it's different for every callback in practice, but should not matter for implementations of these observers. Sticking to "when" makes it more flexible. There is also a chance we can change from before to after or from after to before in the future

daschl (approved):

PR #3307, opened 2025-07-30
Don't apply SO_KEEPALIVE socket option for Unix Domain Sockets transport

idelpivnitskiy:
Motivation:

Adjustment for #2676.
Unix Domain Sockets transport does not support `SO_KEEPALIVE` socket option. Users see `Unknown channel option 'SO_KEEPALIVE' for channel` warning log every time a new connection is created.

Modifications:

- Inspect address type before applying `SO_KEEPALIVE` socket option.
- Add missed javadoc for `TcpServerConfig`.
- Remove unused `TcpClient.defaultConfig()` method.

Result:

No warning for users of UDS. Verified by inspecting `HttpUdsTest` logs.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
@@ -104,7 +104,7 @@ class DefaultHttpServerBuilder implements HttpServerBuilder {
     DefaultHttpServerBuilder(SocketAddress address) {
         this.address = address;
         // Skip default options not supported by this address type
-        if (!(address instanceof DomainSocketAddress || address instanceof io.netty.channel.unix.DomainSocketAddress)) {
+        if (!(address instanceof DomainSocketAddress)) {
             socketOption(SO_KEEPALIVE, true);
         }
     }
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
@@ -141,7 +141,7 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
         this.address = requireNonNull(address);
         config = new HttpClientConfig();
         // Skip default options not supported by this address type
-        if (!(address instanceof DomainSocketAddress || address instanceof io.netty.channel.unix.DomainSocketAddress)) {
+        if (!(address instanceof DomainSocketAddress)) {
             socketOption(SO_KEEPALIVE, true);
         }
         executionContextBuilder = new HttpExecutionContextBuilder();
--- a/servicetalk-http-netty/build.gradle
+++ b/servicetalk-http-netty/build.gradle
@@ -46,7 +46,6 @@ dependencies {
   implementation "io.netty:netty-common"
   implementation "io.netty:netty-handler"
   implementation "io.netty:netty-transport"
-  implementation "io.netty:netty-transport-native-unix-common"
   implementation "org.slf4j:slf4j-api:$slf4jVersion"
 
   runtimeOnly project(":servicetalk-loadbalancer-experimental-provider")


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
```diff
@@ -101,6 +103,10 @@ class DefaultHttpServerBuilder implements HttpServerBuilder {
     // Do not use this ctor directly, HttpServers is the entry point for creating a new builder.
     DefaultHttpServerBuilder(SocketAddress address) {
         this.address = address;
+        // Skip default options not supported by this address type
+        if (!(address instanceof DomainSocketAddress || address instanceof io.netty.channel.unix.DomainSocketAddress)) {
+            socketOption(SO_KEEPALIVE, true);
```
idelpivnitskiy, line 108: I'm not super excited about the taken approach because if we ever have any new protocol based on TCP but not on HTTP, then this default will be lost unless manually preserved there. It will also affect existing users of `TcpConnect`, which is an internal API, but who knows, maybe there is a use-case for that in the wild. The only reason I took it is that I don't need to change public API of `TcpClientConfig` and `TcpServerConfig`.

Alternatively, I can keep it in `AbstractTcpConfig` but then I will have to deprecate default constructors and create a new one that eigher takes `address` or at least its `Class<?>`. WDYT?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
```diff
@@ -138,6 +140,10 @@ final class DefaultSingleAddressHttpClientBuilder<U, R> implements SingleAddress
             final U address, final ServiceDiscoverer<U, R, ? extends ServiceDiscovererEvent<R>> serviceDiscoverer) {
         this.address = requireNonNull(address);
         config = new HttpClientConfig();
+        // Skip default options not supported by this address type
+        if (!(address instanceof DomainSocketAddress || address instanceof io.netty.channel.unix.DomainSocketAddress)) {
```
idelpivnitskiy: Another discovery is that even if users use `io.servicetalk.transport.api.DomainSocketAddress` to start a server, the `HttpServerContext.localAddress()` returns `io.netty.channel.unix.DomainSocketAddress`. See `HttpUdsTest` as an example. The same will happen for `ConnectionInfo` that we use for `network.transport` attribute in metrics and tracing. This is because internally we map ST type to Netty type before bootstrapping Netty (see `io.servicetalk.transport.netty.internal.BuilderUtils.toNettyAddress(Object)` but never map it back for `ConnectionInfo` or `ServerContext` types that return local/remote address.

How do you think, what should we do? Leave as-is and account for 2 types like here or do proper mapping to ST type in a follow-up PR?

bryce-anderson: Good find! I'd vote for doing a proper mapping, however we may be breaking users who are downcasting. I suspect the risk is small, but think it's worth pondering.
If we do it as a pre-PR, does that means we can avoid the second typecheck? I suspect that is what you meant.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
bryce-anderson: I think this is probably fine for now unless we think there will be a new TCP protocol in our future.

bryce-anderson (approved):
I'm fine with this as is, but think it's probably best long term to fix the accidental emission of the netty DomainSocketAddress type.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
idelpivnitskiy: Yes, but the order doesn't matter. I can remove the 2nd check from these 2 files in that follow-up if we decide to do proper mapping.

You made a good point that if we change behavior, it may break existing users. What if they already look for Netty's `io.netty.channel.unix.DomainSocketAddress` type?

bryce-anderson: Then they'll be broken. The question is whether this is something anybody will be doing? I suspect it's very rare but idk.

idelpivnitskiy: I agree, UDS is a rare use-case. Ok, I will work on that change

daschl: I would also vote for doing a proper mapping back and forth, it feels cleaner and we do not leak an implementation detail. Since its likely a rare use-case, calling it out in the release notes should suffice

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultHttpServerBuilder.java
daschl: Yes, let's keep it as-is. One special case is maintainable, if we end up having more then we might need to re-think our approach

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DefaultSingleAddressHttpClientBuilder.java
idelpivnitskiy: Made changes in #3309

bryce-anderson (approved):

daschl (approved):

PR #3309, opened 2025-08-01
Always return `DomainSocketAddress` for Unix Domain Socket transport

idelpivnitskiy:
#### Motivation

In #3307 we observed that when users create server or client with `io.servicetalk.transport.api.DomainSocketAddress` type, after bootstrap they see `io.netty.channel.unix.DomainSocketAddress` type returned from our `ServerContext` and `ConnectionInfo` API. This is unexpected behavior. Users don't expect to operate with a different type at runtime.

#### Modifications

- Enhance `HttpUdsTest` and `GrpcUdsTest` to assert expected behavior.
- Add `BuilderUtils.fromNettyAddress(SocketAddress)` utility to mirror behavior of `BuilderUtils.toNettyAddress(Object)` for necessary cases.
- Use this new utility to convert `SocketAddress` type to satisfy test assertions.

#### Result

Users consistently see `io.servicetalk.transport.api.DomainSocketAddress` type returned from all our public API if they use Unix Domain Socket transport.

### DIFF
--- a/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/EarlyConnectionContext.java
+++ b/servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/EarlyConnectionContext.java
@@ -27,6 +27,7 @@ import java.net.SocketOption;
 import javax.annotation.Nullable;
 import javax.net.ssl.SSLSession;
 
+import static io.servicetalk.transport.netty.internal.BuilderUtils.fromNettyAddress;
 import static io.servicetalk.transport.netty.internal.SocketOptionUtils.getOption;
 
 final class EarlyConnectionContext extends NettyChannelListenableAsyncCloseable implements ConnectionContext {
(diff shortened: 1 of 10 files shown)


daschl (approved):
LGMT - it's a little unfortunate that the conversions are spread out over many different files in the overrides, makes me think that eventually we'll forget one spot in the future. But I also don't have a better idea how to fix it right now.

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/EarlyConnectionContext.java
```diff
@@ -57,12 +58,12 @@ public String connectionId() {
 
     @Override
     public SocketAddress localAddress() {
-        return channel.localAddress();
+        return fromNettyAddress(channel.localAddress());
```
bryce-anderson, line 61: Does it make sense to cache the result, or to save it as a field on initialization? The latter may be risky if we're constructing these before connections are established, so caching may be the only option.

I'm not strongly opinionated, especially considering how infrequently domain sockets are used, so feel free to ignore.

bryce-anderson (approved):

servicetalk-tcp-netty-internal/src/main/java/io/servicetalk/tcp/netty/internal/EarlyConnectionContext.java
idelpivnitskiy: I was considering caching but then decided not to mess with it as it has low value. UDS is rare and the only use-case where caching can give some benefits (address accessed multiple times) I know about, we can cache inside the observer implementation.

PR #3311, opened 2025-08-01
Propagate HTTP/2 streamId to request context

idelpivnitskiy:
#### Motivation

Multiplexed protocols, such as HTTP/2 and HTTP/3, utilize virtual streams for each request. The stream ID is unknown upfront and is assigned by the protocol at the time client sends metadata (headers). Currently, there is no way for client-side to know what the stream ID assigned for a request was.

#### Modifications

- Introduce `HttpContextKeys.STREAM_ID` key as a way to query stream ID value from request context, and adjust H2 codec to do so.
- Add tests to validate new behavior.

#### Result

`STREAM_ID` value is assigned for a request context as soon as request metadata is sent. Users can extract it and correlate with other logs and events. The ID is updated on every retry of the same request instance.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/AbstractDelegatingHttpRequest.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/AbstractDelegatingHttpRequest.java
@@ -103,12 +103,14 @@ abstract class AbstractDelegatingHttpRequest implements PayloadInfo, HttpRequest
         return original.context();
     }
 
+    @Nullable
     @Deprecated
     @Override
     public ContentCodec encoding() {
         return original.encoding();
     }
 
+    @Nullable
     @Override
     public BufferEncoder contentEncoding() {
         return original.contentEncoding();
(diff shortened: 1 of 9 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/AbstractDelegatingHttpRequest.java
```diff
@@ -103,12 +103,14 @@ public ContextMap context() {
...
     }
 
+    @Nullable
     @Deprecated
     @Override
     public ContentCodec encoding() {
         return original.encoding();
     }
 
+    @Nullable
```
bryce-anderson, line 113: Good catch.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/StreamingHttpRequests.java
```diff
@@ -65,12 +68,40 @@ public static StreamingHttpRequest newRequest(
...
         return new DefaultStreamingHttpRequest(method, requestTarget, version, headers, null, null, null, allocator,
                 payload, forTransportReceive(requireTrailerHeader, version, headers), headersFactory);
     }
+
+    /**
+     * Creates a new {@link StreamingHttpRequest} which is read from the transport. If the request contains
+     * <a href="https://tools.ietf.org/html/rfc7230#section-4.4">trailers</a> then the passed {@code payload}
+     * {@link Publisher} must also emit {@link HttpHeaders} before completion.
+     *
+     * @param metaData the {@link HttpRequestMetaData} of the request parsed by the transport layer.
```
bryce-anderson: We should make the comment more clear that the constituent parts of this parameter are shared with the newly minted `StreamingHttpRequest`. I know the other factory methods behave the same but imo it's less obvious with this one. It wouldn't hurt to make their docs more clear as well.

bryce-anderson (approved):

PR #3331, opened 2025-08-25
Avoid deadlock for `subscribeOn()` in case of unexpected exception

idelpivnitskiy:
Motivation:

If one of the delegate async sources throws an unexpected exception from `subscribe` operation, the offloaded path of `subscribeOn` operator can cause a deadlock because that exception is not delivered to the original `Subscriber`.

Modifications:

- Add `SubscribeOnTest` to demonstrate the described scenario.
- Add `try-catch` around each offloaded `handleSubscribe` call to handle unexpected exceptions.
- Fix javadoc formatting and links for internal methods.

Result:

No deadlock if delegate operators throw an unexpected exception from `subscribe` method after offloading.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncSingleOperator.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncSingleOperator.java
@@ -26,7 +26,6 @@ import org.slf4j.LoggerFactory;
 import java.util.function.BooleanSupplier;
 import javax.annotation.Nullable;
 
-import static io.servicetalk.concurrent.internal.SubscriberUtils.deliverErrorFromSource;
 import static io.servicetalk.concurrent.internal.SubscriberUtils.safeCancel;
 import static io.servicetalk.concurrent.internal.SubscriberUtils.safeOnError;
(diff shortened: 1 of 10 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/TaskBasedAsyncSingleOperator.java
```diff
@@ -76,7 +77,22 @@ final io.servicetalk.concurrent.Executor executor() {
     @Override
     void handleSubscribe(final Subscriber<? super T> subscriber,
                          final CapturedContext capturedContext, final AsyncContextProvider contextProvider) {
-        original.delegateSubscribe(subscriber, capturedContext, contextProvider);
+        try {
+            original.delegateSubscribe(subscriber, capturedContext, contextProvider);
+        } catch (Throwable t) {
```
idelpivnitskiy: For now I added this try-catch block for all calls to `super.handleSubscribe(...)` to satisfy tests. After thinking about it more, it feels like we should do that only for `executor().execute(...)` runnable, but not for cases that call this `super` method without offloading. Those cases likely should be caught by a subscriber. WDYT?

bryce-anderson: I think I agree with you, although the distinction is pretty small.

PR #3339, opened 2025-09-23
Upgrade Gradle from 8.14 to 9.1.0

daschl:
This commit contains the necessary modifications to migrate to a new major gradle version, it contains the following fixes and enhancements:

 - Update gradle wrapper distribution to 9.1.0
 - Update shadow plugin from com.github.johnrengelman.shadow to com.gradleup.shadow 8.3.9
 - Remove deprecated CLASSPATH handling from gradlew scripts
 - Modernize Gradle configuration syntax (enabled=false, group/description assignments)
 - Refactor dependency locking configuration
 - Add required test runtime dependencies (junit-platform-launcher, XmlParser)
 - Fix task dependencies for Jersey modules and protobuf generation
 - Update SpotBugs task configuration for Gradle 9 compatibility

### DIFF
--- a/.github/workflows/ci-prb.yml
+++ b/.github/workflows/ci-prb.yml
@@ -13,7 +13,7 @@ on:
       - '**.txt'
 jobs:
   build:
-    name: PRB JDK ${{ matrix.java }}
+    name: JDK ${{ matrix.java }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/.github/workflows/ci-prq.yml
+++ b/.github/workflows/ci-prq.yml
@@ -12,7 +12,7 @@ on:
       - '*.txt'
 jobs:
   quality:
-    name: PRQ JDK ${{ matrix.java }} Toolchain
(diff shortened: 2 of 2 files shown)


servicetalk-http-router-jersey3-jakarta10-internal/build.gradle
```diff
@@ -63,6 +63,7 @@ task copySourcesForJersey3(type: Copy) {
 tasks.processResources.dependsOn(copySourcesForJersey3)
 tasks.sourcesJar.dependsOn(copySourcesForJersey3)
 tasks.compileJava.dependsOn(copySourcesForJersey3)
+tasks.processTestResources.dependsOn(copySourcesForJersey3)
```
bryce-anderson, line 69: I see a bunch of these: is it because something that we had gotten transiently before for this task is no longer inherited?

daschl: gradle 9 is now less forgiving on this, I think it has been forgotten before and now it fails because it is more explicit.

PR #3344, opened 2025-09-30
Update dependency-analysis-gradle-plugin 2.6.1 -> 3.0.4

idelpivnitskiy:


build.gradle
```diff
@@ -81,6 +81,10 @@ if (JavaVersion.current().isJava11Compatible()) {
           // We import it as api dependency via :servicetalk-annotations
           exclude("com.google.code.findbugs:jsr305")
         }
+        onCompileOnly {
+          // For backward compatibility
+          exclude(":servicetalk-annotations") // FIXME: 0.43 - move to compileOnly
```
idelpivnitskiy: How do you guys think, can we move it to `compileOnly` scope now or should we keep it as `implementation` for 0.42.x branch?

daschl: I'm not 100% sure on the impact, could it break downstream users who rely on servicetalk-annotations be on the classpath somehow now to include it manually? If so, I would keep it in 0.42.x as implementation to be safe.

idelpivnitskiy: This is a compile-only annotation in practice but we currently define it as `implementation` which means compile + runtime. I was worried that there is a chance someone relies on this dependency to be pulled transitively, but now I realized that if our users see it as `runtime` they can not use it in their code without defining it as a dependency themselves. So, it seems like changing it to `compileOnly` shouldn't affect anyone. I will give it a try and we can return it back if assumptions were wrong and someone is affected.

idelpivnitskiy: Hm, I tried and got many build failures. It looks like a false positive because this annotation is marked with `RetentionPolicy.RUNTIME` and not `SOURCE`. So, we should keep it as-is

daschl: 👍

idelpivnitskiy: Do you plan on giving me an approval? 😄

daschl: yes sir 🫡

PR #3352, opened 2025-10-18
Align grpc-status mappings with grpc-java and official gRPC docs

idelpivnitskiy:
Motivation:

When a server or proxy responds with an HTTP response instead of grpc response, our current behavior is not aligned with gRPC documentation and grpc-java behavior:
1. https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md
2. https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/GrpcUtil.java

Modifications:
1. Enhance `ProtocolCompatibilityTest.clientH2ReturnStatus` to assert exact behavior for different scenarios: missing `content-type`, missing `grpc-status`, non-200 response status.
2. Align our `GrpcUtils.fromHttpStatus` with grpc-java `GrpcUtil.httpStatusToGrpcCode`.
3. Adjust our `GrpcUtils.validateResponseAndGetPayload` to order validation like in grpc-java: first check `content-type`, then respect `grpc-status`, then fallback to http status.

Result:

ServiceTalk behavior is aligned with gRPC docs and grpc-java.

### DIFF
--- a/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
+++ b/servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
@@ -264,8 +264,20 @@ final class GrpcUtils {
                 .setMessage("Unexpected HTTP status code received: " + status).build());
     }
 
-    // See https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md and
-    // https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/GrpcUtil.java
+    // ServiceTalk provides semantically better mappings than the original spec:
+    // https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md
+    //
+    // Following grpc-java's precedent (431 → INTERNAL), we map additional HTTP status codes to their semantic gRPC
+    // equivalents rather than defaulting to UNKNOWN:
+    //   407 Proxy Authentication Required → UNAUTHENTICATED (like 401)
+    //   408 Request Timeout → DEADLINE_EXCEEDED (timeout semantic)
+    //   412 Precondition Failed → FAILED_PRECONDITION (exact semantic match)
+    //   417 Expectation Failed → FAILED_PRECONDITION (similar to 412)
+    //   501 Not Implemented → UNIMPLEMENTED (method not supported)
+    //
+    // While this diverges from grpc-java's UNKNOWN fallback, it provides better error classification for clients.
+    // grpc-java itself diverges from the spec with 431 → INTERNAL mapping, acknowledging the spec is not comprehensive.
+    // See https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/GrpcUtil.java
     static GrpcStatusCode fromHttpStatus(final HttpResponseStatus status) {
         final int statusCode = status.code();
         final GrpcStatusCode grpcStatusCode;
--- a/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
+++ b/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
@@ -653,7 +653,6 @@ class ProtocolCompatibilityTest {
                             stClient.clientStreamingCall(Publisher.from(request)) :
                             stClient.scalarCall(request))
                             .toFuture().get();
-                    stClient.scalarCall(request).toFuture().get();
                 } catch (ExecutionException e) {
                     logger.debug("servicetalk-grpc failed for httpCode={} with", httpCode, e.getCause());
                     stCode = ((GrpcStatusException) e.getCause()).status().code().value();


servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/HttpToGrpcStatusVariant.java
```diff
@@ -0,0 +1,191 @@
...
+            } else if (httpCode == BAD_REQUEST.code()) {
+                assertStatusesMatch(httpCode, stCode, grpcJavaCode, INTERNAL);
+            } else if (httpCode == UNAUTHORIZED.code()) {
+                assertStatusesMatch(httpCode, stCode, grpcJavaCode, UNAUTHENTICATED);
+            } else if (httpCode == FORBIDDEN.code()) {
+                assertStatusesMatch(httpCode, stCode, grpcJavaCode, PERMISSION_DENIED);
+            } else if (httpCode == NOT_FOUND.code()) {
+                assertStatusesMatch(httpCode, stCode, grpcJavaCode, UNIMPLEMENTED);
+            } else if (httpCode == PROXY_AUTHENTICATION_REQUIRED.code()) {
+                assertStatusesDiffer(httpCode, stCode, UNAUTHENTICATED, grpcJavaCode, UNKNOWN);
```
idelpivnitskiy, line 82: WDYT about differences like this one? It looks to me like https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md is not comprehensive and we can do a little bit better job with mappings, but on the other hand we won't be 1-1 consistent with their behavior. The only reason I kept these differences for now is that they are also not 100% aligned with the docs in grpc-java. There is one status code which mapping is not aligned with the docs: https://github.com/grpc/grpc-java/blob/9662f0fdccf1cb93a056236b0c93ca4bc4adfbf5/core/src/main/java/io/grpc/internal/GrpcUtil.java#L306-L308

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -318,13 +295,12 @@ static <Resp> Publisher<Resp> validateResponseAndGetPayload(final StreamingHttpR
...
                                                                 final GrpcStreamingDeserializer<Resp> deserializer,
                                                                 final String httpPath) {
-        validateStatusCode(response.status()); // gRPC protocol requires 200, don't look further if this check fails.
         // In case of an empty response, gRPC-server may return only one HEADER frame with endStream=true. Our
         // HTTP1-based implementation translates them into response headers so we need to look for a grpc-status in both
         // headers and trailers. Since this is streaming response and we have the headers now, we check for the
         // grpc-status here first. If there is no grpc-status in headers, we look for it in trailers later.
         final HttpHeaders headers = response.headers();
-        validateContentType(headers, expectedContentType);
+        validateContentType(response, expectedContentType);
```
idelpivnitskiy, line 315: I realized that these validations will result in a leaked undrained `StreamingHttpResponse` when they are triggered 😢 
I will send a follow-up to address that after this PR is merged

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
```diff
@@ -592,36 +640,25 @@ void clientH2ReturnStatus() throws Exception {
...
                 } catch (ExecutionException e) {
+                    logger.debug("grpc-java failed for httpCode={} with", httpCode, e.getCause());
                     grpcJavaCode = ((StatusRuntimeException) e.getCause()).getStatus().getCode().value();
                 }
                 try {
+                    (streamingCall ?
+                            stClient.clientStreamingCall(Publisher.from(request)) :
+                            stClient.scalarCall(request))
+                            .toFuture().get();
                     stClient.scalarCall(request).toFuture().get();
```
daschl: should this line be removed since it is covered by the line before?

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/HttpToGrpcStatusVariant.java
daschl: I think if we can be more specific here it is good, since it provides more context to the caller of why it is failing. Since In all these cases the response would be "failure" anyways, I don't think we need to preserve 1:1 behavioral compatibility here.

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
```diff
@@ -281,20 +256,24 @@ static GrpcStatus setStatus(final HttpHeaders trailers, final Throwable cause, f
...
         if (status.code() == OK.code()) {
             return;
         }
         throw GrpcStatusException.of(Status.newBuilder().setCode(fromHttpStatus(status).value())
-                .setMessage("HTTP status code: " + status).build());
+                .setMessage("Unexpected HTTP status code received: " + status).build());
     }
 
+    // See https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md and
+    // https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/GrpcUtil.java
```
daschl: Would it make sense here to document where we (and why) diverge from the spec and the grpc-java implementation? 

something along the lines of 

```
  ServiceTalk provides semantically better mappings than the minimal spec:
  https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md
  
   Following grpc-java's precedent (431→INTERNAL), we map additional HTTP status
   codes to their semantic gRPC equivalents rather than defaulting to UNKNOWN:
     407 Proxy Authentication Required → UNAUTHENTICATED (like 401)
     408 Request Timeout → DEADLINE_EXCEEDED (timeout semantic)
     412 Precondition Failed → FAILED_PRECONDITION (exact semantic match)
     417 Expectation Failed → FAILED_PRECONDITION (similar to 412)
     501 Not Implemented → UNIMPLEMENTED (method not supported)
  
   While this diverges from grpc-java's UNKNOWN fallback, it provides better
   error classification for clients. grpc-java itself diverges from the spec
   with 431→INTERNAL, acknowledging the spec is not comprehensive.
```

servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/ProtocolCompatibilityTest.java
idelpivnitskiy: Good catch, it's a leftover

daschl (approved):

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/GrpcUtils.java
bryce-anderson: Good eye.

idelpivnitskiy: Fixed in https://github.com/apple/servicetalk/pull/3354

PR #3358, opened 2025-10-30
http-api: edit MultiMap.getKeys() to return a LinkedHashSet to prevent non-deterministic behaviors in MultiMap.hashCode()

lawrencewang49:
#### Motivation
`headersWithSameNamesAndValuesShouldBeEquivalent` in `AbstractHttpHeadersTest` included assertions on `hashCode()` that were failing unpredictably:
- `headers1.hashCode() != headers1.hashCode()`
- `headers2.hashCode() != headers2.hashCode()`
- `headers1.hashCode() != headers2.hashCode()`

These failures occur because the hashCode() implementation is non-deterministic for these header objects. The assertions were invalid and caused unreliable test failures with nondex. The existing hashCode() implementation for header objects depended on the iteration order of an internal HashMap. Since HashMap does not guarantee stable ordering, two logically equivalent header instances could produce different hash codes depending on their bucket layout. 

To guarantee that hashCode() is invariant with respect to key iteration order, we need an order-independent hashing strategy.

#### Modifications
- Reworked hashCode() to compute a per-key sub-hash that incorporates the key and all of its associated values.
- Replaced the previous ```31 * result + elementHash accumulation``` with: ```itemHashCode = 31 * itemHashCode + hashCodeForValue(value)```, and summing all per-key itemHashCodes into the final result with ```result += itemHashCode;``` at the end
- Ensures that the final hashCode only depends on the set of keys and values, not on their iteration order.

#### Result
- hashCode() is now deterministic across logically identical header instances.
- Prevents inconsistent behavior when headers are used in hashed collections.
- No functional or API changes beyond making hashing stable and predictable.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiMap.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiMap.java
@@ -441,7 +441,7 @@ abstract class MultiMap<K, V> {
             int itemHashCode = hashCode(key);
             final Iterator<? extends V> valueItr = getValues(key);
             while (valueItr.hasNext()) {
-                // Note that for each item, this algorith makes the hashcode order dependent on it's elements
+                // Note that for each item, this algorith makes the hashcode order dependent on its elements
                 itemHashCode = 31 * itemHashCode + hashCodeForValue(valueItr.next());
             }
             result += itemHashCode;


idelpivnitskiy:
Hi @lawrencewang49,
Thank you for the contribution and interest in our project!

Interesting, I haven't seen `headersWithSameNamesAndValuesShouldBeEquivalent()` test failures before. Could you provide  a bit more details on how you reproduced this issue (including the error log and stack-trace), and why do you think `io.servicetalk.http.api.MultiMap.hashCode()` is non-deterministic?

lawrencewang49:
@idelpivnitskiy Yes!
To reproduce:
1) Add the following to the top of `build.gradle` in the `settings-http-api` module
```
plugins {
    id 'edu.illinois.nondex' version '2.1.7'
}
```
2) Add `apply plugin: 'edu.illinois.nondex'` to the bottom of `build.gradle` also in the `settings-http-api` module
3) Ran `./gradlew :settings-http-api:nondexTest --tests "io.servicetalk.http.api.DefaultHttpHeadersTest.headersWithSameNamesAndValuesShouldBeEquivalent" -DnondexRuns=10` in the terminal in the root repository.

Running that command resulted in the following output for a few runs:
```
DefaultHttpHeadersTest > headersWithSameNamesAndValuesShouldBeEquivalent() FAILED
    org.opentest4j.AssertionFailedError: expected: <-868629791> but was: <829772141>
        at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
        at app//org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
        at app//org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:531)
        at app//io.servicetalk.http.api.AbstractHttpHeadersTest.headersWithSameNamesAndValuesShouldBeEquivalent(AbstractHttpHeadersTest.java:439)
```
and similar output for another run:
```
DefaultHttpHeadersTest > headersWithSameNamesAndValuesShouldBeEquivalent() FAILED
    org.opentest4j.AssertionFailedError: expected: <829772141> but was: <-868629791>
        at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
        at app//org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
        at app//org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:531)
        at app//io.servicetalk.http.api.AbstractHttpHeadersTest.headersWithSameNamesAndValuesShouldBeEquivalent(AbstractHttpHeadersTest.java:439)
```

The reason I believe that it's non-deterministic is because the hashCode() inconsistency arises from `io.servicetalk.http.api.MultiMap.hashCode()`, which iterates over keys returned by `getKeys()`. That method constructs a new HashSet on each call. Since HashSet is backed by a HashMap and does not guarantee the same iteration order of the set over time, the final hash value can vary even for identical headers. This makes the `io.servicetalk.http.api.MultiMap.hashCode()`function used in the hash code assertions in headersWithSameNamesAndValuesShouldBeEquivalent() non-deterministic.

Equality checks remain valid since equals() doesn’t depend on key iteration order, only the hash codes vary.

idelpivnitskiy:
Heh, that's a great catch! Thanks a lot for detailed explanation.
It's actually important to preserve deterministic `hashCode()` behavior. Would you like to rebase and try fixing `MultiMap` instead of removing assertions?

lawrencewang49:
> Heh, that's a great catch! Thanks a lot for detailed explanation. It's actually important to preserve deterministic `hashCode()` behavior. Would you like to rebase and try fixing `MultiMap` instead of removing assertions?

Done!

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiMap.java
```diff
@@ -144,7 +144,7 @@ final Set<? extends K> getKeys() {
             return emptySet();
         }
         // Overall iteration order does not need to be preserved.
-        final Set<K> names = new HashSet<>((int) (size() / .75), .75f);
+        final Set<K> names = new LinkedHashSet<>((int) (size() / .75), .75f);
```
bryce-anderson: I don't know that this actually fixes our issue since `LinkedHashSet` provides _insertion_ ordering, not total ordering. Being a hash-map like data structure, the size of the hash array can change the ordering of buckets and thus iteration order.
We could go two ways: either use something like a TreeSet, which will give us both a Set and the correct iteration order, or we could modify the `hashCode()` method. We could do the latter by changing the general algorithm from `result = 31 * result + hash(newContributor)` to simply `result += hash(newContributor)`. This will fix our iteration order problem (since the sum will be the sum regardless of iteration order) but will reduce hashcode entropy, but maybe that's fine.

bryce-anderson: @lawrencewang49 , this is what I'm thinking:
```patch
diff --git a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiMap.java b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiMap.java
index 07a5f641a..4ffd18e97 100644
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiMap.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiMap.java
@@ -433,12 +433,18 @@ abstract class MultiMap<K, V> {
             return 0;
         }
         int result = HASH_CODE_SEED;
+        // Because we use a hashmap internally, our key iteration order may not be consistent across two
+        // instances even if they have same contents because the bucket ordering may be different.
+        // To make the hashcode work correctly we make a sub-hashcode for each key and its associated values,
+        // and then we simply sum them up to because the sum is invariant over key iteration order.
         for (final K key : getKeys()) {
-            result = 31 * result + hashCode(key);
+            int itemHashCode = hashCode(key);
             final Iterator<? extends V> valueItr = getValues(key);
             while (valueItr.hasNext()) {
-                result = 31 * result + hashCodeForValue(valueItr.next());
+                // Note that for each item, this algorith makes the hashcode order dependent on it's elements
+                itemHashCode = 31 * itemHashCode + hashCodeForValue(valueItr.next());
             }
+            result += itemHashCode;
         }
         return result;
     }
```

idelpivnitskiy: To me it looks like `LinkedHashSet` will solve the problem with hashCode and also give predictable iteration order if users use `names()` directly, which is also nice for the users experience. Looking at how it's done in Netty, we will be consistent if we use `LinkedHashSet`:

names: https://github.com/netty/netty/blob/98ef8b9c1d45115eb37a1f9270a38bc7cdfc3a07/codec-base/src/main/java/io/netty/handler/codec/DefaultHeaders.java#L313-L325

hashCode: https://github.com/netty/netty/blob/98ef8b9c1d45115eb37a1f9270a38bc7cdfc3a07/codec-base/src/main/java/io/netty/handler/codec/DefaultHeaders.java#L981-L996

idelpivnitskiy: @lawrencewang49 I chatted with @bryce-anderson more offline and he explained me what I was missing. With `LinkedHashSet`, the `hashCode()` will be consistent only for the same instance of `HttpHeaders`. If we have two different instances that were constructed with exact same headers but in different order, then we will get different hashes, which is unexpected. 

His patch will solve the problem without dependency on underlying data structure. WDYT?

lawrencewang49: Hi @bryce-anderson @idelpivnitskiy, so sorry for the long delay! I got busy with midterms and haven't had the chance to check my emails recently.

Thanks for the detailed explanation and fix! @bryce-anderson's proposal makes sense and I'm on board with his changes!

I’m happy to update the PR to incorporate that approach if you guys are also okay with it.

idelpivnitskiy: Yes please, go ahead

servicetalk-http-api/src/main/java/io/servicetalk/http/api/MultiMap.java
```diff
@@ -433,12 +433,18 @@ public int hashCode() {
...
+        // instances even if they have same contents because the bucket ordering may be different.
+        // To make the hashcode work correctly we make a sub-hashcode for each key and its associated values,
+        // and then we simply sum them up to because the sum is invariant over key iteration order.
         for (final K key : getKeys()) {
-            result = 31 * result + hashCode(key);
+            int itemHashCode = hashCode(key);
             final Iterator<? extends V> valueItr = getValues(key);
             while (valueItr.hasNext()) {
-                result = 31 * result + hashCodeForValue(valueItr.next());
+                // Note that for each item, this algorith makes the hashcode order dependent on it's elements
```
bryce-anderson: I believe this is my fault, I'm sorry. 😢 
```suggestion
                // Note that for each item, this algorith makes the hashcode order dependent on its elements
```

lawrencewang49: Done! Sorry, I should've proofread 😅

lawrencewang49:
@bryce-anderson done! no problem, it was great learning more about this framework and being able to contribute something to it!

lawrencewang49:
@idelpivnitskiy Thanks for letting me know! I will get back to you on this, as another student may have been assigned to work on identifying nondeterministic test results for this repository in our class. If you guys would like, you are free to open an issue and implement the changes yourselves!

Issue #3356, opened 2025-10-29
Flaky ProtocolCompatibilityTest.serviceTalkToServiceTalkErrorViaServerFilter


bryce-anderson:
https://github.com/apple/servicetalk/actions/runs/18923384907/job/54024965876?pr=3355#step:8:969

```
> Task :servicetalk-grpc-netty:test

ProtocolCompatibilityTest > serviceTalkToServiceTalkErrorViaServerFilter(boolean, boolean, String) > [7] false, true, gzip FAILED
    java.lang.AssertionError: Detected pending requests, possible response leak
    Expected: is <0>
         but: was <1>
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at io.servicetalk.grpc.netty.ProtocolCompatibilityTest$ResponseLeakValidator.assertNoPendingRequests(ProtocolCompatibilityTest.java:1903)
        at io.servicetalk.grpc.netty.ProtocolCompatibilityTest.finalChecks(ProtocolCompatibilityTest.java:218)
```

PR #3359, opened 2025-10-30
`GrpcExceptionMapperServiceFilter`: log `GrpcStatusException` as expected

idelpivnitskiy:
Motivation:

`GrpcStatusException` is created by users when they map any issue into appropriate grpc-status. It's common to through this exception type from intermediate HTTP filters to avoid the need to construct gRPC response manually.

Modifications:

- Adjust `GrpcExceptionMapperServiceFilter` to log `GrpcStatusException` at debug level instead of error.
- Simplify other use-cases that already use `GrpcStatusException.serverCatchAllShouldLog`.

Result:

Less noisy logs for already handled exceptions.


bryce-anderson (approved):
PR looks good, it seems like CI is building with another manifestation of https://github.com/apple/servicetalk/issues/3356.

PR #3364, opened 2025-11-05
Upgrade Gradle from 8.14 to 9.2.0

daschl:
This commit contains the necessary modifications to migrate to a new major gradle
version, it contains the following fixes and enhancements:

 - Update gradle wrapper distribution to 9.2.0
 - Update shadow plugin from com.github.johnrengelman.shadow to com.gradleup.shadow 8.3.9
 - Remove deprecated CLASSPATH handling from gradlew scripts
 - Modernize Gradle configuration syntax (enabled=false, group/description assignments)
 - Refactor dependency locking configuration
 - Add required test runtime dependencies (junit-platform-launcher, XmlParser)
 - Fix task dependencies for Jersey modules and protobuf generation
 - Update SpotBugs task configuration for Gradle 9 compatibility
 - Update CI to pick up JDK 17 for gradle (not the main jvm)
 - Move from JDK 24 to JDK 25 LTS testing

More info on the last point: Gradle 9 can only run with JDK 17 or later, so there
are multiple ways to approach this. Using language toolchains would likely be the
right call, but causing a lot of migration pain. The easier version is to run the
gradle wrapper and daemon with jdk 17 but everything else with the configured
java version. To do this, we can specify the gradle home variable and always load
jdk 17 into the CI workflow.

### DIFF
--- a/.github/workflows/ci-prb.yml
+++ b/.github/workflows/ci-prb.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        java: [ 8, 11, 17, 21, 24 ]
+        java: [ 8, 11, 17, 21, 25 ]
         os: [ ubuntu-latest ]
         os_label: [ ubuntu ]
         include:


build.gradle
```diff
@@ -123,20 +123,10 @@ subprojects {
...
-      resolutionStrategy.activateDependencyLocking()
-    }
-    runtimeClasspath {
-      resolutionStrategy.activateDependencyLocking()
-    }
-    jmhCompileClasspath {
-      resolutionStrategy.activateDependencyLocking()
-    }
-    jmhRuntimeClasspath {
+  def lockableConfigs = ["annotationProcessor", "compileClasspath", "runtimeClasspath", "jmhCompileClasspath",
```
bryce-anderson, line 126: Nice touch.

servicetalk-data-protobuf-jersey3-jakarta10/build.gradle
```diff
@@ -73,6 +73,7 @@ task copySourcesForJersey3(type: Copy) {
 tasks.processResources.dependsOn(copySourcesForJersey3)
 tasks.sourcesJar.dependsOn(copySourcesForJersey3)
 tasks.compileJava.dependsOn(copySourcesForJersey3)
+tasks.generateProto.dependsOn(copySourcesForJersey3)
```
bryce-anderson, line 76: What triggered this new line? Gradle is no longer as aggressive at copying stuff?

bryce-anderson (approved):

servicetalk-data-protobuf-jersey3-jakarta10/build.gradle
daschl: Yeah gradle 9 starts to fail the build if this is missing - looks like gradle 8 was more lenient with non-explicit task dependencies

PR #3385, opened 2026-01-06
Fix gaps in AsyncContext propagation when services API converted

idelpivnitskiy:
Motivation:

This is a follow-up for #3382 to add more test coverage and cover remaining gaps.

Modifications:
1. Implement `ServiceApiConversionsAsyncContextTest` that adds integration testing for all combinations of service API conversions.
2. Remove `StreamingHttpServiceToBlockingHttpServiceTest` and `StreamingHttpServiceToBlockingStreamingHttpServiceTest` - it's easy to miss a case when mocks are used, (1) adds comprehensive coverage.
3. Adjust `AsyncContextHttpFilterVerifier` to make it more flexible and reusable.
4. Add `.shareContextOnSubscribe()` as necessary.

Result:

Correct AsyncContext propagation on the server-side when users convert services to/from all API variants.

Issue #3384, opened 2026-01-06
JerseyRouterTestSuite.AsyncContextTest were affected by context propagation improvements


idelpivnitskiy:
In #3170 we accidentally disabled all `JerseyRouterTestSuite` and restored them back in #3383. However, due to improvements we made in #3218 and #3221 assertions inside `JerseyRouterTestSuite.AsyncContextTest` trigger errors. 

For JAX-RS endpoints that do not have payload body`AsyncContextAssertionFilter` expects that keys are not set during request payload draining. Those assertions worked correctly because we were relying on old buggy behavior when context was not propagated correctly through all parts of request-response chain. Now that the context propagates as expected, we see keys when `DrainingStreamingHttpServiceFilter` auto-drains the request body after executing the endpoint.

To make behavior correct, we need to explicitly drain request payload when we find the route and see that it does not have incoming payload body. We should also throw an exception if draining detects some incoming bytes.

PR #3386, opened 2026-01-06
Fix gaps in `AsyncContext` propagation for Jersey router

idelpivnitskiy:
#### Motivation
In #3383 we had to skip `AsyncContextTest` for `BaseJerseyRouterTestSuite` because the context propagation did not pass all `AsyncContextAssertionFilter` checks. Now that we addressed some gaps for API conversions in #3385, we can re-enabled `AsyncContextTest` and fix remaining gaps inside the Jersey router.

#### Modifications
1. Uncomment skipped `AsyncContextTest` in `BaseJerseyRouterTestSuite`, enhance it to test with all `RouterApi` variants, and adjust assertions to temporarily workaround #3384.
2. Fix remaining gaps in Jersey router found by these tests.

#### Result
`AsyncContext` is correctly propagated through the Jersey router.

### DIFF
--- a/servicetalk-http-router-jersey-internal/src/main/java/io/servicetalk/http/router/jersey/internal/BufferPublisherInputStream.java
+++ b/servicetalk-http-router-jersey-internal/src/main/java/io/servicetalk/http/router/jersey/internal/BufferPublisherInputStream.java
@@ -87,6 +87,9 @@ public final class BufferPublisherInputStream extends InputStream {
     public void offloadSourcePublisher(final HttpExecutionStrategy executionStrategy, final Executor executor) {
         if (inputStream == EMPTY_INPUT_STREAM) {
             publisher = executionStrategy.isMetadataReceiveOffloaded() || executionStrategy.isDataReceiveOffloaded() ?
+                    // We only need to add shareContextOnSubscribe() if we decide to offload. Otherwise, it's already
+                    // shared before the `publisher` was wrapped with BufferPublisherInputStream in
+                    // DefaultJerseyStreamingHttpRouter.
                     publisher.publishOn(executor, IoThreadFactory.IoThread::currentThreadIsIoThread)
                             .shareContextOnSubscribe() : publisher;
         } else {
--- a/servicetalk-http-router-jersey/src/testFixtures/java/io/servicetalk/http/router/jersey/resources/AsyncContextResources.java
+++ b/servicetalk-http-router-jersey/src/testFixtures/java/io/servicetalk/http/router/jersey/resources/AsyncContextResources.java
@@ -16,6 +16,7 @@
 package io.servicetalk.http.router.jersey.resources;
 
 import io.servicetalk.buffer.api.Buffer;
+import io.servicetalk.buffer.api.BufferAllocator;
 import io.servicetalk.concurrent.api.AsyncContext;
 import io.servicetalk.concurrent.api.Completable;
 import io.servicetalk.concurrent.api.Publisher;
@@ -198,10 +199,11 @@ public class AsyncContextResources {
     @Path("/publisherEchoSync")
     @Consumes(TEXT_PLAIN)
     @Produces(TEXT_PLAIN)
-    public Publisher<Buffer> publisherEchoSync(Publisher<Buffer> in) throws Exception {
+    public Publisher<Buffer> publisherEchoSync(Publisher<Buffer> in) {
         AsyncContext.put(K1, V1);
-        in.ignoreElements().shareContextOnSubscribe().toFuture().get();
-        return Publisher.from(ctx.executionContext().bufferAllocator().fromUtf8("foo"))
+        final BufferAllocator allocator = ctx.executionContext().bufferAllocator();
+        return in.ignoreElements()
+                .concat(Publisher.from(allocator.fromUtf8("foo")).shareContextOnSubscribe())
                 .beforeOnComplete(() -> AsyncContext.put(K3, V3));
     }
 


servicetalk-http-netty/src/testFixtures/java/io/servicetalk/http/netty/AsyncContextHttpFilterVerifier.java
```diff
@@ -281,8 +281,9 @@ public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
                                     assertSameContext(current, errorQueue);
                                 }
                                 assertAsyncContext(K1, lazyPayload ? V1 : null, errorQueue);
-                                assertAsyncContext(K2, null, errorQueue);
-                                assertAsyncContext(K3, null, errorQueue);
+                                // FIXME: uncomment the next lines, see https://github.com/apple/servicetalk/issues/3384
+                                // assertAsyncContext(K2, null, errorQueue);
+                                // assertAsyncContext(K3, null, errorQueue);
```
idelpivnitskiy, line 286: I don't plan addressing these `FIXME`s in this PR, it will be a follow-up for one of the next releases to auto-drain incoming requests if Jersey endpoint does not have incoming arguments

servicetalk-http-router-jersey/src/testFixtures/java/io/servicetalk/http/router/jersey/resources/AsyncContextResources.java
```diff
@@ -199,9 +198,9 @@ public Publisher<Buffer> publisherEcho(Publisher<Buffer> in) {
     @Path("/publisherEchoSync")
     @Consumes(TEXT_PLAIN)
     @Produces(TEXT_PLAIN)
-    public Publisher<Buffer> publisherEchoSync(Publisher<Buffer> in) {
+    public Publisher<Buffer> publisherEchoSync(Publisher<Buffer> in) throws Exception {
         AsyncContext.put(K1, V1);
-        in.toCompletionStage().toCompletableFuture().join();
+        in.ignoreElements().shareContextOnSubscribe().toFuture().get();
```
idelpivnitskiy: Had to change to `toFuture()` here because `.toCompletionStage()` converts publisher to single, which does not allow us to share the context 😕

servicetalk-http-router-jersey-internal/src/main/java/io/servicetalk/http/router/jersey/internal/BufferPublisherInputStream.java
```diff
@@ -87,7 +87,8 @@ public int read(final byte[] b, final int off, final int len) throws IOException
     public void offloadSourcePublisher(final HttpExecutionStrategy executionStrategy, final Executor executor) {
         if (inputStream == EMPTY_INPUT_STREAM) {
             publisher = executionStrategy.isMetadataReceiveOffloaded() || executionStrategy.isDataReceiveOffloaded() ?
-                    publisher.publishOn(executor, IoThreadFactory.IoThread::currentThreadIsIoThread) : publisher;
+                    publisher.publishOn(executor, IoThreadFactory.IoThread::currentThreadIsIoThread)
+                            .shareContextOnSubscribe() : publisher;
```
daschl, line 94: just to double check: we only need to share if offloaded?

daschl (approved):

servicetalk-http-router-jersey-internal/src/main/java/io/servicetalk/http/router/jersey/internal/BufferPublisherInputStream.java
bryce-anderson: +1: In my minds eye, it should be the same regardless.

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/DefaultJerseyStreamingHttpRouter.java
```diff
@@ -253,8 +253,8 @@ private void handle0(final HttpServiceContext serviceCtx, final StreamingHttpReq
         req.headers().forEach(h ->
                 containerRequest.getHeaders().add(h.getKey().toString(), h.getValue().toString()));
 
-        final BufferPublisherInputStream entityStream = new BufferPublisherInputStream(req.payloadBody(),
-                publisherInputStreamQueueCapacity);
+        final BufferPublisherInputStream entityStream = new BufferPublisherInputStream(
+                req.payloadBody().shareContextOnSubscribe(), publisherInputStreamQueueCapacity);
```
bryce-anderson, line 257: This feels like the same 'sharing doesn't make sense for iterator behavior due to buffering', or is it somehow different?

servicetalk-http-router-jersey/src/testFixtures/java/io/servicetalk/http/router/jersey/resources/AsyncContextResources.java
bryce-anderson: I don't think it will matter at all, but will this get flagged as an API breakage during the release? I think technically throwing the new exception breaks the API. If so, It seems easy enough to just wrap it in a runtime exception.

bryce-anderson (approved):
Only two small questions that are non-blocking for me.

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/DefaultJerseyStreamingHttpRouter.java
idelpivnitskiy: Yes, in case Jersey decides to go through forEach case with an iterator, it does it before invoking the endpoint. If we don't share there, any modifications made by a filter during request payload body processing won't be visible for the endpoint.

Actually, the same is true for `DefaultBlockingStreamingHttpRequest` that we discussed y'day. We haven't considered a case when a filter can set the value (like request payload body checksum) that the endpoint may want to read after processing the payload body.

servicetalk-http-router-jersey-internal/src/main/java/io/servicetalk/http/router/jersey/internal/BufferPublisherInputStream.java
idelpivnitskiy: The original publisher is unwrapped from `BufferPublisherInputStream` which already has `shareContextOnSubscribe()` before it was wrapped. I will add a comment to clarify this

servicetalk-http-router-jersey/src/testFixtures/java/io/servicetalk/http/router/jersey/resources/AsyncContextResources.java
idelpivnitskiy: thanks, updated

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/DefaultJerseyStreamingHttpRouter.java
bryce-anderson: That's true re filter setting the value. That one is more clear cut, other than if it sets a value per item it's still racy behavior with respect to which value(s) be read by the hander.

bryce-anderson (approved):

PR #3403, opened 2026-01-14
Update SpotBugs 4.7.3 -> 4.9.8

idelpivnitskiy:
Motivation:

New version does not require us to disable spotbugs on newer JDKs anymore.

Modifications:
- Remove logic that disables SpotBugs task with newer JDK versions.
- Use `AtomicLong` instead of `long` to advance time in tests to avoid `AT_NONATOMIC_OPERATIONS_ON_SHARED_VARIABLE` warning.
- Make classes `final` if they throw from constructor to avoid `CT_CONSTRUCTOR_THROW` warning complaining that those are vulnerable to Finalizer attacks.
- Merge `AbstractContentCodec` and `NettyChannelContentCodec` into a single final class for the same reason.
- Make inner classes `static` when possible.
- Suppress some false positives, explaining in comments why.
- Remove leftover `@Timeout` annotations in tests.

Result:

SpotBugs works with all versions of JVM, newer version -> better validation.


daschl (approved):
great cleanup!

PR #3407, opened 2026-01-23
Add support for H1 total header size limit

daschl:
This changeset ensures that in addition to checking each individual header for its length, the total amount is also accumulated and checked against a limit.

The default for this limit is 8k, similar to H2 as well as what netty does for its decoder, but with one caveat: to ease deployment and not break users in a bugfix release, the code will only warn by default to let users adjust. When you do not see any errors for your use case, flip the system property. In the future, this will be set to throw by default.

Test cases added and adjusted to also cover trailers.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfig.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfig.java
@@ -18,7 +18,7 @@ package io.servicetalk.http.netty;
 import io.servicetalk.http.api.HttpClient;
 import io.servicetalk.http.api.HttpProtocolConfig;
 
-import static io.servicetalk.http.netty.H1ProtocolConfigBuilder.DEFAULT_MAX_TOTAL_HEADER_LENGTH;
+import static io.servicetalk.http.netty.H1ProtocolConfigBuilder.DEFAULT_MAX_TOTAL_HEADER_FIELDS_LENGTH;
 
 /**
  * Configuration for <a href="https://tools.ietf.org/html/rfc7230">HTTP/1.1</a> protocol.
@@ -57,16 +57,16 @@ public interface H1ProtocolConfig extends HttpProtocolConfig {
     int maxStartLineLength();
 
     /**
-     * Get the maximum total length of the HTTP
-     * <a href="https://tools.ietf.org/html/rfc7230#section-3.1">start line</a> and all
+     * Get the maximum total allowed length of all HTTP
      * <a href="https://tools.ietf.org/html/rfc7230#section-3.2">header fields</a> combined.
      * <p>
-     * This limit protects against memory exhaustion attacks where an attacker sends many small headers.
+     * This limit protects against memory exhaustion attacks where an attacker sends many small headers which
+     * individually are below the {@link #maxHeaderFieldLength()} limit.
      *
-     * @return maximum total length of start line and all header fields combined
+     * @return maximum total allowed length of all header fields combined
      */
-    default int maxTotalHeaderLength() { // TODO 0.43 - remove default method
-        return DEFAULT_MAX_TOTAL_HEADER_LENGTH;
+    default int maxTotalHeaderFieldsLength() { // TODO 0.43 - remove default method
+        return DEFAULT_MAX_TOTAL_HEADER_FIELDS_LENGTH;
     }
 
     /**
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
@@ -16,6 +16,7 @@
 package io.servicetalk.http.netty;
 
 import io.servicetalk.http.api.DefaultHttpHeadersFactory;
+import io.servicetalk.http.api.Http2Settings;
 import io.servicetalk.http.api.HttpClient;
 import io.servicetalk.http.api.HttpHeaders;
 import io.servicetalk.http.api.HttpHeadersFactory;
@@ -32,12 +33,12 @@ public final class H1ProtocolConfigBuilder {
 
     private static final H1SpecExceptions DEFAULT_H1_SPEC_EXCEPTIONS = new H1SpecExceptions.Builder().build();
 
-    static final int DEFAULT_MAX_TOTAL_HEADER_LENGTH = 8192;
+    static final int DEFAULT_MAX_TOTAL_HEADER_FIELDS_LENGTH = 8192;
 
     private int maxPipelinedRequests = 1;
     private int maxStartLineLength = 4096;
     private int maxHeaderFieldLength = 8192;
-    private int maxTotalHeaderLength = DEFAULT_MAX_TOTAL_HEADER_LENGTH;
+    private int maxTotalHeaderFieldsLength = DEFAULT_MAX_TOTAL_HEADER_FIELDS_LENGTH;
     private HttpHeadersFactory headersFactory = DefaultHttpHeadersFactory.INSTANCE;
     private int headersEncodedSizeEstimate = 256;
     private int trailersEncodedSizeEstimate = 256;
@@ -92,8 +93,7 @@ public final class H1ProtocolConfigBuilder {
     }
 
     /**
-     * Sets the maximum total length of the HTTP
-     * <a href="https://tools.ietf.org/html/rfc7230#section-3.1">start line</a> and all
+     * Sets the maximum total allowed length of all HTTP
      * <a href="https://tools.ietf.org/html/rfc7230#section-3.2">header fields</a> combined.
      * <p>
      * This limit protects against memory exhaustion attacks where an attacker sends many small headers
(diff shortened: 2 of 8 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfig.java
```diff
@@ -54,6 +54,17 @@ default String alpnId() {
...
+    /**
+     * Get the maximum total length of the HTTP
+     * <a href="https://tools.ietf.org/html/rfc7230#section-3.1">start line</a> and all
+     * <a href="https://tools.ietf.org/html/rfc7230#section-3.2">header fields</a> combined.
+     * <p>
+     * This limit protects against memory exhaustion attacks where an attacker sends many small headers.
+     *
+     * @return maximum total length of start line and all header fields combined
+     */
+    int maxTotalHeaderLength();
```
bryce-anderson: I think this is technically a breaking change and our ABI checker may be sad: can we add a reasonable default implementation, maybe 8k?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -786,6 +815,21 @@ private boolean parseAllHeaders(final ByteBuf buffer, final HttpHeaders headers,
                 consumeCRLF(buffer, lfIndex);
                 return true;
             }
+
+            final int headerLineSize = nonControlIndex - buffer.readerIndex() + 1;
+            currentHeaderLength += headerLineSize;
+            if (currentHeaderLength > maxTotalHeaderLength) {
+                if (totalHeaderLengthWarnOnly) {
+                    LOGGER.warn("HTTP headers exceed maxTotalHeaderLength limit: {} > {} bytes. This request " +
```
daschl: @bryce-anderson @idelpivnitskiy wdyt think -- should we keep this at WARN or move to DEBUG? I want to avoid causing log spamming, but still we need to notify it somehow.

bryce-anderson: Maybe make it once-per-connection?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfig.java
```diff
@@ -54,6 +56,19 @@ default String alpnId() {
      */
     int maxStartLineLength();
 
+    /**
+     * Get the maximum total length of the HTTP
+     * <a href="https://tools.ietf.org/html/rfc7230#section-3.1">start line</a> and all
```
idelpivnitskiy: Why does the new limit include start line? Looks like in Netty `maxStartLineLength` and `maxHeaderFieldLength` are fully independent from each other and in order to smooth the migration, we need to preserve similar behavior but a different name for the 2nd one

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfig.java
```diff
@@ -54,6 +56,19 @@ default String alpnId() {
...
+     * <a href="https://tools.ietf.org/html/rfc7230#section-3.2">header fields</a> combined.
+     * <p>
+     * This limit protects against memory exhaustion attacks where an attacker sends many small headers.
+     *
+     * @return maximum total length of start line and all header fields combined
+     */
+    default int maxTotalHeaderLength() { // TODO 0.43 - remove default method
```
idelpivnitskiy: Consider highlighting in the name that it's for all headers by using plural form in all places

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
```diff
@@ -88,6 +91,29 @@ public H1ProtocolConfigBuilder maxStartLineLength(final int maxStartLineLength)
...
+     * Sets the maximum total length of the HTTP
+     * <a href="https://tools.ietf.org/html/rfc7230#section-3.1">start line</a> and all
+     * <a href="https://tools.ietf.org/html/rfc7230#section-3.2">header fields</a> combined.
+     * <p>
+     * This limit protects against memory exhaustion attacks where an attacker sends many small headers
+     * that individually pass validation but collectively consume excessive memory.
+     * <p>
+     * <b>Note:</b> a decoder will close the connection with {@code TooLongFrameException} if the total
+     * header block size exceeds this value. The default matches HTTP/2's
+     * <a href="https://tools.ietf.org/html/rfc7540#section-6.5.2">SETTINGS_MAX_HEADER_LIST_SIZE</a>.
```
idelpivnitskiy, line 104: Consider adding a link to `H2ProtocolConfigBuilder#initialSettings(Http2Settings)` so that users can relate how to configure it for h2

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java
```diff
@@ -88,6 +91,29 @@ public H1ProtocolConfigBuilder maxStartLineLength(final int maxStartLineLength)
...
+     * <p>
+     * If property is set to {@code io.servicetalk.http.netty.maxTotalHeaderLengthWarnOnly=true} (which is the default),
+     * then ServiceTalk will only emit a warning instead of throwing - this is to ease initial rollout of this limit.
```
idelpivnitskiy: Consider mentioning that in the future releases we will enforce this limit as error instead of warning by default

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -127,8 +131,27 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
     private static final int MAX_ALLOWED_CHARS_TO_SKIP = CHUNK_DELIMETER_SIZE * 2; // Max allowed prefacing CRLF to skip
     private static final int MAX_ALLOWED_CHARS_TO_SKIP_PLUS_ONE = MAX_ALLOWED_CHARS_TO_SKIP + 1;
 
+    // FIXME: 0.43 - remove this system property after validation in production
+    // This property allows to enable a "warn-only" mode for the new total header size limit.
+    // When enabled, exceeding maxTotalHeaderLength logs a warning instead of throwing TooLongFrameException.
+    // This is useful for gradual rollout to identify legitimate traffic that exceeds the limit.
+    private static final String TOTAL_HEADER_LENGTH_WARN_ONLY_PROPERTY =
+            "io.servicetalk.http.netty.maxTotalHeaderLengthWarnOnly";
```
idelpivnitskiy: Should we somehow highlight in the name that it's a temporary property that will be removed in the future major releases or fine to keep as-is?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -127,8 +131,27 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
...
+    protected static final boolean TOTAL_HEADER_LENGTH_WARN_ONLY = Boolean.parseBoolean(System
+            .getProperty(TOTAL_HEADER_LENGTH_WARN_ONLY_PROPERTY, "true"));
+
+    static {
+        if (TOTAL_HEADER_LENGTH_WARN_ONLY) {
+            LOGGER.debug("Total header length enforcement is in WARN-ONLY mode. " +
+                    "Requests exceeding maxTotalHeaderLength will log a warning but not be rejected. " +
+                    "Set -D{}=false for production hardening.", TOTAL_HEADER_LENGTH_WARN_ONLY_PROPERTY);
+        }
```
idelpivnitskiy, line 151: Consider always logging the value of the property in `-D{}: {}` format at debug level, similar to how we do it in other places (for example, in `DefaultDnsServiceDiscovererBuilder`).

I agree that when it's set to `true`, we need more users attentions with further description which is worth to have at `warn` or at least `info` levels. Something like:

```java
if (TOTAL_HEADER_LENGTH_WARN_ONLY) {
    LOGGER.warn("-D{}: {}. Requests exceeding the limit won't be rejected. Users should flip this property to false and adjust their limit as necessary.",
        TOTAL_HEADER_LENGTH_WARN_ONLY_PROPERTY, TOTAL_HEADER_LENGTH_WARN_ONLY);
} else {
    LOGGER.debug("-D{}: {}", TOTAL_HEADER_LENGTH_WARN_ONLY_PROPERTY, TOTAL_HEADER_LENGTH_WARN_ONLY);
}
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
idelpivnitskiy: The more we log, higher chance ppl will notice. But to avoid overflowing logs, I agree with Bryce that logging once per connection can be a reasonable balance

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -786,6 +815,21 @@ private boolean parseAllHeaders(final ByteBuf buffer, final HttpHeaders headers,
                 consumeCRLF(buffer, lfIndex);
                 return true;
             }
+
+            final int headerLineSize = nonControlIndex - buffer.readerIndex() + 1;
+            currentHeaderLength += headerLineSize;
+            if (currentHeaderLength > maxTotalHeaderLength) {
+                if (totalHeaderLengthWarnOnly) {
+                    LOGGER.warn("HTTP headers exceed maxTotalHeaderLength limit: {} > {} bytes. This request " +
+                                    "would be rejected when warn-only mode is disabled. Disable -D{}=true to " +
```
idelpivnitskiy: This might be easier to understand:
```suggestion
                                    "would be rejected when warn-only mode is disabled. Set -D{}=false to " +
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -127,8 +131,27 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
...
     private static final int MAX_ALLOWED_CHARS_TO_SKIP_PLUS_ONE = MAX_ALLOWED_CHARS_TO_SKIP + 1;
 
+    // FIXME: 0.43 - remove this system property after validation in production
+    // This property allows to enable a "warn-only" mode for the new total header size limit.
+    // When enabled, exceeding maxTotalHeaderLength logs a warning instead of throwing TooLongFrameException.
+    // This is useful for gradual rollout to identify legitimate traffic that exceeds the limit.
+    private static final String TOTAL_HEADER_LENGTH_WARN_ONLY_PROPERTY =
+            "io.servicetalk.http.netty.maxTotalHeaderLengthWarnOnly";
+    protected static final boolean TOTAL_HEADER_LENGTH_WARN_ONLY = Boolean.parseBoolean(System
+            .getProperty(TOTAL_HEADER_LENGTH_WARN_ONLY_PROPERTY, "true"));
```
idelpivnitskiy, line 142: Consider adding `FIXME: 0.43 - flip default value`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
daschl: I think it might be fine to keep it this way if we for the future major release just flip the property instead of completely removing it?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfig.java
daschl: Aah you are right, I misread that part! I'll fix it up.

PR #3454, opened 2026-03-17
Add on-demand support for jacoco code coverage

daschl:
#### Motivation
Adds support to run jacoco code coverage on demand if needed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,7 @@
  */
 
 import io.servicetalk.gradle.plugin.internal.Versions
+import org.gradle.testing.jacoco.plugins.JacocoCoverageReport
 
 if (!repositories) {
   allprojects {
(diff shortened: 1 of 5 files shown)


bryce-anderson (approved):
Seems to work for me.

build.gradle
```diff
@@ -93,6 +94,52 @@ dependencyAnalysis {
 
 quality.dependsOn buildHealth
 
+apply plugin: "jacoco-report-aggregation"
```
idelpivnitskiy: Optionally: just because the number of lines is large and they are under different code blocks (not a single one), consider moving this code to something like `./gradle/jacoco.gradle` file and then use `apply` here to import it

idelpivnitskiy (approved):

bryce-anderson (approved):

PR #3456, opened 2026-03-17
`ConnectionObserver` does not see enriched `ConnectException`

idelpivnitskiy:
Motivation:

`TcpConnector` enriches `ConnectException`s with their retryable wrappers, but `ConnectionObserver.connectionClosed(Throwable)` does not see those.

Modifications:

1. Enhance `ConnectionAcceptingNettyHttpServerTest` to ensure `ConnectionObserver` sees the same retryable exception type as callers on connect timeouts.
2. `TcpTransportObserverErrorsTest`: to ensure connection refused error is retryable.
3. Adjust `ConnectionObserverInitializer` and `TcpConnector` to propagate consistent exception types.

Result:

`ConnectionObserver.connectionClosed(Throwable)` sees the same exception type as the caller.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptingNettyHttpServerTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptingNettyHttpServerTest.java
@@ -174,5 +174,10 @@ class ConnectionAcceptingNettyHttpServerTest extends AbstractNettyHttpServerTest
         public void connectionClosed(final Throwable error) {
             observed.add(error);
         }
+
+        @Override
+        public void connectionClosed() {
+            observed.add(new IllegalStateException("Unexpected call to connectionClosed()"));
+        }
     }
 }
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptingNettyHttpServerTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptingNettyHttpServerTest.java
@@ -16,20 +16,27 @@
 package io.servicetalk.http.netty;
 
 import io.servicetalk.client.api.ConnectTimeoutException;
+import io.servicetalk.client.api.TransportObserverConnectionFactoryFilter;
 import io.servicetalk.concurrent.api.Single;
 import io.servicetalk.http.api.HttpServerBuilder;
 import io.servicetalk.http.api.SingleAddressHttpClientBuilder;
 import io.servicetalk.http.api.StreamingHttpClient;
 import io.servicetalk.http.api.StreamingHttpRequest;
 import io.servicetalk.http.api.StreamingHttpResponse;
+import io.servicetalk.transport.api.ConnectionObserver;
 import io.servicetalk.transport.api.HostAndPort;
+import io.servicetalk.transport.api.RetryableException;
+import io.servicetalk.transport.api.TransportObserver;
 
 import org.junit.jupiter.api.Test;
 
 import java.net.InetSocketAddress;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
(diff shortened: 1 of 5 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ConnectionObserverInitializer.java
```diff
@@ -156,30 +158,21 @@ public void handlerAdded(final ChannelHandlerContext ctx) {
...
                     addedCloseListener = true;
-                    Throwable t = channelError(future.channel());
-                    if (t == null) {
-                        t = future.cause();
-                    }
-                    if (t == null) {
-                        observer.connectionClosed();
-                    } else {
-                        observer.connectionClosed(t);
-                    }
```
bryce-anderson, line 179: No longer observing connection failures here feels like a behavior change, although I don't know how significant it is. It seems like the real problem is that we want the `TcpConnector` to _translate the error_ before propagating it, similar to a `Single.onErrorMap`.
Unfortunately something like `onErrorMap` doesn't exist for `ChannelPromise`. If we had a `DelegatingChannelPromise` we could override the `[try|set]Failure` methods to do that conversion for us and move our error translation code into that. I copied the netty `DelegatingChannelPromiseNotifier` and overrode those methods and all your test modifications pass.

However, interestingly, the ConnectionObserverEventOrderTest test breaks because the `connectionClosed` event moves to the last position, which I haven't looked into.

idelpivnitskiy: > No longer observing connection failures here feels like a behavior change

Why do you think so? The observer callback just moved from here to `TcpConnector`, but ordering of events and the event itself is preserved

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/ConnectionAcceptingNettyHttpServerTest.java
```diff
@@ -137,4 +156,23 @@ private void assertConnectionRequestSucceeds(final StreamingHttpRequest request)
...
+        @Override
+        public ConnectionObserver onNewConnection(@Nullable final Object localAddress, final Object remoteAddress) {
+            return this;
+        }
+
+        @Override
+        public void connectionClosed(final Throwable error) {
+            observed.add(error);
+        }
+    }
```
daschl, line 182: consider overriding connectionClosed() and call fail() so that it would not pass silently if behavior regresses.

daschl (approved):

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/ConnectionObserverInitializer.java
bryce-anderson: nm, I saw that `ConnectionObserverInitializer` is public, but just noticed it's for an `internal` package.
That aside, it does feel like we're starting to conflate the responsibilities of two classes. Do you want to keep the original netty exception propagating back along the connect call, or would you prefer to propagate the enriched exception?

idelpivnitskiy: I feel like the enriched one is the best to propagate bcz it carries more meta-data (better msg and retryable marker), while still preserving the original one in the cause. I agree that the responsibility gets split, which is not nice but at the same time `ConnectionObserver` events are reported from different places (not only from `ConnectionObserverInitializer`). I'm open for any suggestions or follow-up improvements to consolidate those

bryce-anderson: Talked with Idel about this offline. My suggestion is encapsulated in #3459.

bryce-anderson:
This is superseded by #3459.

PR #3461, opened 2026-03-23
Defer connectionEstablished observer callback and H2 auto-read until

daschl:
after connection acceptors complete

#### Motivation
Two related issues stem from the same root cause — actions that should happen after all connection acceptors complete instead happen when the connection Single resolves in `tryCompleteSubscriber`/`completeSubscriber`:

 1. `connectionEstablished`/`multiplexedConnectionEstablished` observer callbacks fire before Late/Legacy `ConnectionAcceptors` evaluate. This means connection time measurement excludes acceptor time, and the callback fires even if the acceptor later rejects the connection.

 2. HTTP/2 server enables auto-read in `AbstractH2ParentConnection#handlerAdded()` before acceptors evaluate. If the client's HTTP/2 preface and request arrive in the same TCP segment, the server processes the request before the acceptor can reject — meaning `StreamingHttpService.handle()` gets invoked for connections that should have been rejected.

#### Modifications
 1. Add a `deferConnectionEstablished` parameter to `DefaultNettyConnection.initChannel()` and guard the `connectionEstablished` call in `completeSubscriber()`. Add a method for callers to invoke after acceptors pass.

 2. Add a `deferAutoRead` parameter to `AbstractH2ParentConnection`. When true, `handlerAdded()` no longer enables auto-read. The server passes true; the client passes false (unchanged behavior).

 3. Add new tests covering observer callback timing and H2 auto-read deferral across H1, H2 prior-knowledge, and H2 ALPN code paths.

#### Result
` connectionEstablished/multiplexedConnectionEstablished` observer
 callbacks now fire after all connection acceptors complete, and only
 when the connection is accepted. HTTP/2 auto-read is deferred until
 the connectionConsumer runs, preventing requests from reaching the
 service handler before acceptors can reject. Client-side behavior is
 unchanged.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DeferredServerChannelBinder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DeferredServerChannelBinder.java
@@ -69,11 +69,14 @@ final class DeferredServerChannelBinder {
 
         return TcpServerBinder.bind(listenAddress, tcpConfig, executionContext, connectionAcceptor, channelInit,
                 serverConnection -> {
-                    // Start processing requests on http/1.1 connection:
+                    // Notify connection established after all acceptors have completed:
                     if (serverConnection instanceof NettyHttpServerConnection) {
+                        ((NettyHttpServerConnection) serverConnection).notifyConnectionEstablished();
                         ((NettyHttpServerConnection) serverConnection).process(true);
+                    } else if (serverConnection instanceof H2ServerParentConnectionContext) {
+                        ((H2ServerParentConnectionContext) serverConnection)
+                                .notifyConnectionEstablishedAndEnableAutoRead();
                     }
-                    // Nothing to do otherwise as h2 uses auto read on the parent channel
                 }, earlyConnectionAcceptor, lateConnectionAcceptor)
                 .map(delegate -> {
                     LOGGER.debug("Started HTTP server with ALPN for address {}", delegate.listenAddress());
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DeferredServerChannelBinder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DeferredServerChannelBinder.java
@@ -71,11 +71,13 @@ final class DeferredServerChannelBinder {
                 serverConnection -> {
                     // Notify connection established after all acceptors have completed:
                     if (serverConnection instanceof NettyHttpServerConnection) {
-                        ((NettyHttpServerConnection) serverConnection).notifyConnectionEstablished();
-                        ((NettyHttpServerConnection) serverConnection).process(true);
+                        ((NettyHttpServerConnection) serverConnection).notifyConnectionEstablishedAndProcess();
                     } else if (serverConnection instanceof H2ServerParentConnectionContext) {
                         ((H2ServerParentConnectionContext) serverConnection)
                                 .notifyConnectionEstablishedAndEnableAutoRead();
+                    } else {
+                        throw new IllegalStateException("Unexpected connection type: " +
+                                serverConnection.getClass().getName());
                     }
                 }, earlyConnectionAcceptor, lateConnectionAcceptor)
                 .map(delegate -> {
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
@@ -44,6 +44,7 @@ import io.servicetalk.transport.netty.internal.NoopTransportObserver.NoopMultipl
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPipeline;
+import io.netty.channel.EventLoop;
 import io.netty.handler.codec.http2.Http2SettingsFrame;
 import io.netty.handler.codec.http2.Http2StreamChannel;
 import org.slf4j.Logger;
@@ -66,6 +67,10 @@ import static java.util.Objects.requireNonNull;
 final class H2ServerParentConnectionContext extends H2ParentConnectionContext implements ServerContext {
     private static final Logger LOGGER = LoggerFactory.getLogger(H2ServerParentConnectionContext.class);
     private final SocketAddress listenAddress;
+    @Nullable
+    private ConnectionObserver observer;
+    @Nullable
+    private DefaultH2ServerParentConnection parentConnectionHandler;
     private H2ServerParentConnectionContext(final Channel channel, final HttpExecutionContext executionContext,
                                             final FlushStrategy flushStrategy,
                                             final long idleTimeoutMs,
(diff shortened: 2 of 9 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DeferredServerChannelBinder.java
```diff
@@ -69,11 +69,14 @@ static Single<HttpServerContext> bind(final HttpExecutionContext executionContex
...
                 serverConnection -> {
-                    // Start processing requests on http/1.1 connection:
+                    // Notify connection established after all acceptors have completed:
                     if (serverConnection instanceof NettyHttpServerConnection) {
+                        ((NettyHttpServerConnection) serverConnection).notifyConnectionEstablished();
                         ((NettyHttpServerConnection) serverConnection).process(true);
+                    } else if (serverConnection instanceof H2ServerParentConnectionContext) {
+                        ((H2ServerParentConnectionContext) serverConnection)
+                                .notifyConnectionEstablishedAndEnableAutoRead();
                     }
```
bryce-anderson, line 81: Now that I think we've covered both cases, should we have an `else` that probably blows up since I think it means an unexpected state?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -185,10 +188,14 @@ private static Single<NettyHttpServerConnection> initChannel(final Channel chann
...
-                HTTP_1_1, observer, false, __ -> false)
-                .map(conn -> new NettyHttpServerConnection(conn, service,
-                        HTTP_1_1, h1Config.headersFactory(),
-                        config.allowDropTrailersReadFromTransport())),
+                HTTP_1_1, observer, false, __ -> false, true)
+                .map(conn -> {
+                    final NettyHttpServerConnection serverConn = new NettyHttpServerConnection(conn, service,
+                            HTTP_1_1, h1Config.headersFactory(),
+                            config.allowDropTrailersReadFromTransport());
+                    serverConn.observer = observer;
```
bryce-anderson: fwiw, I think this would be better as a constructor parameter that can be null and then in the case of the stream channels you can say why it's null.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/DeferredServerChannelBinder.java
```diff
@@ -69,11 +69,14 @@ static Single<HttpServerContext> bind(final HttpExecutionContext executionContex
 
         return TcpServerBinder.bind(listenAddress, tcpConfig, executionContext, connectionAcceptor, channelInit,
                 serverConnection -> {
-                    // Start processing requests on http/1.1 connection:
+                    // Notify connection established after all acceptors have completed:
                     if (serverConnection instanceof NettyHttpServerConnection) {
+                        ((NettyHttpServerConnection) serverConnection).notifyConnectionEstablished();
                         ((NettyHttpServerConnection) serverConnection).process(true);
```
bryce-anderson: It looks like `notifyConnectionEstablished()` and `process(true)` are always called together and it _feels_ like it has kind of the same meaning as `notifyConnectionEstablishedAndEnableAutoRead()` which is "notify the observer and start processing". Can we unify these into one method call?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
```diff
@@ -87,6 +92,44 @@ public SocketAddress listenAddress() {
...
+     * Thread-safety invariant — this method may run on an offload executor thread (e.g., when a
+     * {@link LateConnectionAcceptor} requires offloading). The {@code observer}
+     * and {@code parentConnectionHandler} fields were written during {@link #initChannel} (on the event loop)
+     * and are safely published here via the reactive chain's {@code onSuccess} delivery, which provides a
+     * happens-before edge.
+     * <p>
+     * The actual work is always dispatched to the event loop to ensure that the {@code multiplexedObserver}
+     * write and the subsequent {@code setAutoRead(true)} (which triggers {@code channel.read()} inline)
+     * both execute on the same thread that will read the observer when frames arrive. This makes the
+     * visibility guarantee trivial — single-threaded access, no memory ordering concerns.
```
bryce-anderson: fwiw, these comments seem excessive, especially for an internal class.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
```diff
@@ -87,6 +92,44 @@ public SocketAddress listenAddress() {
...
+        final EventLoop eventLoop = nettyChannel().eventLoop();
+        if (eventLoop.inEventLoop()) {
+            doNotifyAndEnableAutoRead();
+        } else {
+            eventLoop.execute(this::doNotifyAndEnableAutoRead);
+        }
+    }
+
+    private void doNotifyAndEnableAutoRead() {
+        if (observer != null && parentConnectionHandler != null) {
```
bryce-anderson: I think the `parentConnectionHandler != null` is better as an assert in the if block.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ServerParentConnectionContext.java
```diff
@@ -66,6 +67,10 @@
 final class H2ServerParentConnectionContext extends H2ParentConnectionContext implements ServerContext {
     private static final Logger LOGGER = LoggerFactory.getLogger(H2ServerParentConnectionContext.class);
     private final SocketAddress listenAddress;
+    @Nullable
+    private ConnectionObserver observer;
+    @Nullable
+    private DefaultH2ServerParentConnection parentConnectionHandler;
     private H2ServerParentConnectionContext(final Channel channel, final HttpExecutionContext executionContext,
```
bryce-anderson, line 74:
```suggestion
    private DefaultH2ServerParentConnection parentConnectionHandler;

private H2ServerParentConnectionContext(final Channel channel, final HttpExecutionContext executionContext,
```

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/NettyHttpServer.java
```diff
@@ -309,6 +318,17 @@ void process(final boolean handleMultipleRequests) {
                     .subscribe(new ErrorLoggingHttpSubscriber(this));
         }
 
+        /**
+         * Notifies the observer that the connection has been established. Called from the connection consumer
+         * after all connection acceptors have passed.
+         */
+        void notifyConnectionEstablished() {
+            if (observer != null && connection instanceof DefaultNettyConnection) {
```
bryce-anderson: What if `connection` is not an instance of `DefaultNettyConnection`? I think we should refine the type of `connection` itself.

bryce-anderson (commented):
It feels like the underlying problem is that we setup our pipelines too early. If that's true and how to address it might b a lot more work than we want to tackle.

bryce-anderson (approved):

PR #3466, opened 2026-04-28
Add UBI9 Devcontainers

chrisvest:
Motivation:
Devcontainers are useful for running tests locally in a predictable environment, and for doing stuff in a Linux environment even when your local system is Windows or macOS.

Modification:
Add devcontainer files.
The devcontainer is based on RedHat UBI9, because I think that's more representative of a typical prod environment than Ubuntu is, despite CI using Ubuntu.

Result:
Easy local Linux-based dev workflow.

### DIFF
--- a/.devcontainer/ubi9/Dockerfile
+++ b/.devcontainer/ubi9/Dockerfile
@@ -25,4 +25,4 @@ RUN bash -c "source /root/.sdkman/bin/sdkman-init.sh && \
     rm -rf /root/.sdkman/tmp/*"
 
 RUN echo 'export JAVA_HOME="/root/.sdkman/candidates/java/current"' >> ~/.bashrc
-RUN echo 'PATH=/jdk/bin:$PATH' >> ~/.bashrc
+RUN echo 'export PATH=$JAVA_HOME/bin:$PATH' >> ~/.bashrc
--- a/.devcontainer/ubi9/Dockerfile
+++ b/.devcontainer/ubi9/Dockerfile
@@ -1,0 +1,28 @@
+FROM redhat/ubi9
+
+ARG java_version=17.0.19-zulu
+ENV JAVA_VERSION=$java_version
+
+RUN dnf install -y \
+    procps-ng \
+    git \
+    openssl \
+    tar \
+    unzip \
+    wget \
+    zip \
+    zlib-devel \
+    && dnf clean all
+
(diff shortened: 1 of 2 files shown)


.devcontainer/ubi9/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM redhat/ubi9
+
+ARG java_version=17.0.19-zulu
```
idelpivnitskiy, line 3: IIUC, it should be possible to use `17-zulu` to grab the latest instead of a specific version

.devcontainer/ubi9/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+# Installing Java removing some unnecessary SDKMAN files
+RUN bash -c "source /root/.sdkman/bin/sdkman-init.sh && \
+    sdk list java | grep zulu && \
+    yes | sdk install java $JAVA_VERSION && \
+    rm -rf /root/.sdkman/archives/* && \
+    rm -rf /root/.sdkman/tmp/*"
+
+RUN echo 'export JAVA_HOME="/root/.sdkman/candidates/java/current"' >> ~/.bashrc
+RUN echo 'PATH=/jdk/bin:$PATH' >> ~/.bashrc
```
idelpivnitskiy: Should it be `$JAVA_HOME/bin` instead of `/jdk/bin`?

.devcontainer/ubi9/devcontainer.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "servicetalk-ubi9",
+  "containerEnv": {
```
idelpivnitskiy, line 3: Would be nice to have some short docs either in core `README.md` or in `.devcontainer/README.md` on how to use it

.devcontainer/ubi9/Dockerfile
chrisvest: Sadly not, I tried that first.

.devcontainer/ubi9/devcontainer.json
chrisvest: This depends on the combination of IDE and container runtime. It'll be better to just let people find the most up to date info for their environment on their own, than trying to write yet another guide to devcontainers.

idelpivnitskiy (approved):
Thank you!

PR #3471, opened 2026-05-04
Adjust maxTotalHeaderFieldsLength when maxHeaderFieldLength is larger

daschl:
This changeset ensures that when maxHeaderFieldLength is explicitly or implicitly configured larger than maxTotalHeaderFieldsLength, the value of maxTotalHeaderFieldsLength is set to match in order to not cause misleading issues at runtime.

We still WARN though, since this likely points towards a misconfiguration and should be addressed eventually by the application code.


idelpivnitskiy (approved):
LGTM, but I also think that there are quite low chances someone will notice a one time warning log. For server-side it might be even more beneficial to throw. Either way works for me

daschl:
> LGTM, but I also think that there are quite low chances someone will notice a one time warning log. For server-side it might be even more beneficial to throw. Either way works for me

I see. My take on this is that it is not a "showstopper" misconfiguration in the sense that you have a limit, it's just larger and we can be lenient to accept and still operate with it.

bryce-anderson:
Post merge :shipit:

PR #3472, opened 2026-05-04
Add ability to unpack `[Http|GrpcLifecycleObservers]`

idelpivnitskiy:
Motivation:

`HttpLifecycleObservers` and `GrpcLifecycleObservers` offer `combine(...)` method to combine multiple observers into a single one. Also, `AbstractLifecycleObserverHttpFilter`s do not allow to check which observers are installed. This makes it hard for `HttpProviders` to intercept client/server configuration.

Modifications:
1. Add `unpack(...)` method for `HttpLifecycleObservers` and `GrpcLifecycleObservers` as reverse operation to `combine(...)`.
2. Add `AbstractLifecycleObserverHttpFilter.lifecycleObservers()` method.

Result:

`HttpProviders` can detect what observers are installed for client or server.


daschl (approved):
lgtm! nice work on the recursion aspect.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/HttpLifecycleObservers.java
```diff
@@ -69,4 +75,35 @@ public static HttpLifecycleObserver combine(final HttpLifecycleObserver first, f
...
+     * If the provided {@code observer} was created using one of the {@link #combine(HttpLifecycleObserver,
+     * HttpLifecycleObserver) combine} methods, this method recursively extracts all individual observers that were
+     * combined. Otherwise, returns a singleton list containing the provided observer.
+     *
+     * @param observer {@link HttpLifecycleObserver} to unpack
+     * @return a {@link List} of leaf {@link HttpLifecycleObserver}s
+     * @see #combine(HttpLifecycleObserver, HttpLifecycleObserver)
+     * @see #combine(HttpLifecycleObserver, HttpLifecycleObserver, HttpLifecycleObserver...)
+     */
+    public static List<HttpLifecycleObserver> unpack(final HttpLifecycleObserver observer) {
```
bryce-anderson, line 91: Does this need to be public right now? I think we'd need to move them, but otherwise it seems like they could be package private. I see the symmetry, but ideally we could pass all the observers together as a list instead of making a tree structure out of them.

servicetalk-grpc-utils/src/main/java/io/servicetalk/grpc/utils/GrpcLifecycleObservers.java
```diff
@@ -69,4 +75,35 @@ public static GrpcLifecycleObserver combine(final GrpcLifecycleObserver first, f
...
+     * If the provided {@code observer} was created using one of the {@link #combine(GrpcLifecycleObserver,
+     * GrpcLifecycleObserver) combine} methods, this method recursively extracts all individual observers that were
+     * combined. Otherwise, returns a singleton list containing the provided observer.
+     *
+     * @param observer {@link GrpcLifecycleObserver} to unpack
+     * @return a {@link List} of leaf {@link GrpcLifecycleObserver}s
+     * @see #combine(GrpcLifecycleObserver, GrpcLifecycleObserver)
+     * @see #combine(GrpcLifecycleObserver, GrpcLifecycleObserver, GrpcLifecycleObserver...)
+     */
+    public static List<GrpcLifecycleObserver> unpack(final GrpcLifecycleObserver observer) {
```
bryce-anderson, line 91: Same question re does it need to be public right now.

bryce-anderson (approved):
I would prefer to leave the unpacking private for now, but I'm not strongly opinionated.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/HttpLifecycleObservers.java
idelpivnitskiy: > Does this need to be public right now?

Yes, there are a few reasons:
1. `unpack` method should live in `servicetalk-http-utils` module to get pkg-private access to `BiHttpLifecycleObserver`, but it's called from another module `servicetalk-http-netty`
2. Providers that intercept `HttpServerBuilder.lifecycleObserver(HttpLifecycleObserver)` still need a way to unpack it.

> ideally we could pass all the observers together as a list instead of making a tree structure out of them

I parked that for the future as we did not resolve all questions about API when we discussed it last time. Even if new public API will accept a list in the future and return a list:
1. The ship has sailed and we still need to support current API until we can remove it
2. Internally I would still expect it to use `BiHttpLifecycleObserver` to avoid dealing with lists in `AbstractLifecycleObserverHttpFilter`

servicetalk-grpc-utils/src/main/java/io/servicetalk/grpc/utils/GrpcLifecycleObservers.java
idelpivnitskiy: Should live in `servicetalk-grpc-utils` and accessed from `servicetalk-grpc-netty`

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/HttpLifecycleObservers.java
bryce-anderson: Ah yes, I didn't consider that we'd lose access to `BiHttpLifecycleObserver`. That settles it then.

PR #3483, opened 2026-05-12
http-router-jersey: guard EndpointEnhancingRequestFilter against concurrent shutdown

daschl:
### Motivation

HK2's ServiceLocatorImpl.shutdown() calls context.shutdown() (which sets
RequestScope.isActive = false) outside its write lock, then re-acquires
the lock to set state = SHUTDOWN. In that window, an in-flight request
reaching EndpointEnhancingRequestFilter causes HK2 to throw:

```
  IllegalStateException: Could not find an active context for
      org.glassfish.jersey.process.internal.RequestScoped
```

HK2 wraps this in a MultiException (via Utilities.createService()) before
it reaches the filter. This is a known HK2 design trade-off (observed through
4.0.1): context shutdown is done intentionally outside the write lock to avoid
deadlocks, leaving a window where the scope is inactive but the locator is not
yet fully shut down. The workaround must live in ServiceTalk permanently.

### Modifications

- EndpointEnhancingRequestFilter.filter(): catch MultiException around
  ctxRefProvider.get() — the actual exception HK2 throws in this window —
  and abort with 503 Service Unavailable if requestScope.isActive() is false.
  RequestScope.isActive is a one-way latch so if it is false in the catch
  the exception was caused by the shutdown race; otherwise re-throw.
- Add RouterConcurrentShutdownTest reproducing the exact production exception
  and asserting a graceful 503 response.
- Wire RouterConcurrentShutdownTest into BaseJerseyRouterTestSuite.

### Result

Requests during concurrent shutdown receive a 503 instead of propagating
an internal HK2 MultiException.

### DIFF
--- a/servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
+++ b/servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
@@ -65,6 +65,7 @@ import static io.servicetalk.http.router.jersey.internal.RequestProperties.setRe
 import static java.lang.Integer.MAX_VALUE;
 import static java.util.Objects.requireNonNull;
 import static javax.ws.rs.core.Response.Status.NO_CONTENT;
+import static javax.ws.rs.core.Response.Status.SERVICE_UNAVAILABLE;
 import static javax.ws.rs.core.Response.noContent;
 
 /**
@@ -97,7 +98,27 @@ final class EndpointEnhancingRequestFilter implements ContainerRequestFilter {
     @Override
     public void filter(final ContainerRequestContext requestCtx) {
         // If we don't have a ConnectionContext then the request isn't from ServiceTalk and need not be filtered.
-        if (ctxRefProvider.get().get() != null) {
+        final Ref<ConnectionContext> ctxRef;
+        try {
+            ctxRef = ctxRefProvider.get();
+        } catch (RuntimeException e) {
+            // HK2 ServiceLocatorImpl.shutdown() (observed through 4.0.1) deactivates request-scope contexts
+            // outside its write lock before flipping its own state to SHUTDOWN, so an in-flight request can
+            // hit two failure modes:
+            //   1) MultiException wrapping ISE("Could not find an active context for RequestScoped"),
+            //      thrown by Utilities.createService while locator state is still RUNNING.
+            //   2) Raw ISE("<locator> has been shut down"), thrown by ServiceLocatorImpl.checkState()
+            //      once state == SHUTDOWN.
+            // RequestScope.isActive is a one-way latch flipped during shutdown, so it is the authoritative
+            // signal that we are racing against locator teardown — it covers both windows. If false, abort
+            // with 503; otherwise re-throw, the failure was unrelated.
+            if (!requestScope.isActive()) {
+                requestCtx.abortWith(Response.status(SERVICE_UNAVAILABLE).build());
+                return;
+            }
+            throw e;
+        }
+        if (ctxRef.get() != null) {
             enhancedEndpointCache.enhance(requestScope, ctxRefProvider, routeStrategiesConfigProvider,
                     (UriRoutingContext) requestCtx.getUriInfo());
         }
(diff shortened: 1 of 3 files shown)


servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
```diff
@@ -96,8 +97,27 @@ final class EndpointEnhancingRequestFilter implements ContainerRequestFilter {
 
     @Override
     public void filter(final ContainerRequestContext requestCtx) {
+        // HK2 bug (unfixed through 4.0.1): ServiceLocatorImpl.shutdown() calls context.shutdown()
+        // (setting RequestScope.isActive = false) outside its write lock, before state = SHUTDOWN
+        // is set. In that window, ctxRefProvider.get() throws:
+        //   IllegalStateException: Could not find an active context for RequestScoped
```
idelpivnitskiy: Consider adding a link to HK2 bug if it's public, I could not find it

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
```diff
@@ -96,8 +97,27 @@ final class EndpointEnhancingRequestFilter implements ContainerRequestFilter {
...
+        if (!requestScope.isActive()) {
+            requestCtx.abortWith(Response.status(SERVICE_UNAVAILABLE).build());
```
idelpivnitskiy: 503 typically means that the server is temporarily down or overloaded. If it's caused by a bug, will 500 work better in this case?

idelpivnitskiy (approved):
LGTM, just a question on the status code:

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
```diff
@@ -97,7 +99,23 @@ final class EndpointEnhancingRequestFilter implements ContainerRequestFilter {
     @Override
     public void filter(final ContainerRequestContext requestCtx) {
         // If we don't have a ConnectionContext then the request isn't from ServiceTalk and need not be filtered.
-        if (ctxRefProvider.get().get() != null) {
+        final Ref<ConnectionContext> ctxRef;
+        try {
+            ctxRef = ctxRefProvider.get();
+        } catch (MultiException e) {
```
bryce-anderson: fwiw, I think you can catch a generic throwable and see if an `IllegalStateException` is in the cause chain. That will probably be more generic and make it so you don't need to import the hk2 API. Otoh, if you're reasonably certain it's always going to be a MultiException this is fine.

bryce-anderson (approved):
I'm not strongly opinionated about MultiException vs ISE, so feel free to ship as is.

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
daschl: Actually, I rephrased it now -- I think it is just the way it is intended to work -- I think 503 should work well with load balancers when the shutdown happens

servicetalk-http-router-jersey/src/main/java/io/servicetalk/http/router/jersey/EndpointEnhancingRequestFilter.java
daschl: after reconsidering, I think it is just intended behavior so I adjusted the PR to reflect that wording

gradle.properties
```diff
@@ -45,6 +45,7 @@ javaxJaxbApiVersion=2.3.3
 javaxJaxbCoreVersion=2.3.0.1
 javaxJaxbImplVersion=2.3.9
 javaxInjectVersion=2.6.1
+hk2ApiVersion=2.6.1
```
idelpivnitskiy: If you decide to keep hk2 dependency for `MultiException`, do we need different hk2 versions for different variants of Jersey: 2, 3, 4?

bryce-anderson (approved):

PR #3508, opened 2026-06-16
http-netty: Cap HTTP/1.1 chunk-size at Integer.MAX_VALUE

daschl:
This changeset aligns the chunk-size handling with netty's parser  and cleans up the implementation a little bit w.r.t int/long.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
@@ -433,7 +433,7 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
                 }
                 onDataSeen();
                 final int lfIndex = crlfIndex(longLFIndex);
-                long chunkSize = getChunkSize(buffer, lfIndex);
+                int chunkSize = getChunkSize(buffer, lfIndex);
                 consumeCRLF(buffer, lfIndex);
                 this.chunkSize = chunkSize;
                 if (chunkSize == 0) {
@@ -444,7 +444,7 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
                 // fall-through
             }
             case READ_CHUNKED_CONTENT: {
-                final int toRead = min((int) min(Integer.MAX_VALUE, chunkSize), buffer.readableBytes());
+                final int toRead = (int) min(chunkSize, buffer.readableBytes());
                 if (toRead == 0) {
                     return;
                 }
@@ -836,7 +836,7 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
         }
     }
 
-    private static long getChunkSize(final ByteBuf buffer, final int lfIndex) {
+    private static int getChunkSize(final ByteBuf buffer, final int lfIndex) {
         if (lfIndex - 2 < buffer.readerIndex()) {
             throw new DecoderException("Chunked encoding specified but chunk-size not found");
         }
@@ -844,7 +844,13 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
                 lfIndex - 1 - buffer.readerIndex(), US_ASCII));
     }
 
-    private static long getChunkSize(String hex) {
+    /**
+     * Parses an HTTP/1.1 chunk-size token. Stops at the first {@code ';'}, whitespace, or
+     * ISO control character so any chunk-extension is ignored. Visible for testing the
+     * {@code value < 0} guard which is otherwise unreachable through the public flow given
+     * the {@link #MAX_HEX_CHARS_FOR_LONG} line cap.
+     */
+    static int getChunkSize(String hex) {
         hex = hex.trim();
         for (int i = 0; i < hex.length(); ++i) {
             char c = hex.charAt(i);
@@ -853,9 +859,13 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
                 break;
             }
         }
-
         try {
-            return parseUnsignedLong(hex, 16);
+            final long value = parseUnsignedLong(hex, 16);
+            if (value < 0 || value > Integer.MAX_VALUE) {
+                throw new NumberFormatException("Chunk size " + Long.toUnsignedString(value)
+                        + " exceeds Integer.MAX_VALUE");
+            }
+            return (int) value;
         } catch (NumberFormatException cause) {
             throw invalidChunkSize(hex, cause);
         }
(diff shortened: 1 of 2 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -844,21 +843,62 @@ private static long getChunkSize(final ByteBuf buffer, final int lfIndex) {
...
             char c = hex.charAt(i);
             if (c == ';' || isWhitespace(c) || isISOControl(c)) {
-                hex = hex.substring(0, i);
-                break;
+                if (i == 0) {
+                    throw invalidChunkSize(hex, new NumberFormatException("Empty chunk size"));
+                }
+                return result;
+            }
+            final int digit = decodeHexNibble(c);
```
idelpivnitskiy: The reason why Netty reads input digit-by-digit is because they don't do `toString` conversion. In our case, we already use a string, so it's much easier to stick to `parseUnsignedLong` with a follow-up check to ensure limits. If you'd like to optimize this part for digit-by-digit read, then we should skip the `toString`  part to take full advantage of it. I would propose that as a follow-up optimization with a proper benchmark to ensure complexity is worth the benefit

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -844,21 +843,62 @@ private static long getChunkSize(final ByteBuf buffer, final int lfIndex) {
...
+     * an out-of-range chunk-size or producing a negative {@code int} after a cast in
+     * {@code READ_CHUNKED_CONTENT}. Mirrors {@code netty-4.2.15.Final}
+     * {@code HttpObjectDecoder#getChunkSize}.
+     *
+     * <p>Stops at the first {@code ';'}, whitespace, or ISO control character so any
+     * trailing chunk-extension is ignored. Throws when the input is empty, contains a
+     * non-hex character, or its numeric value cannot be represented as a non-negative
+     * {@code int}.
+     */
     private static long getChunkSize(String hex) {
```
idelpivnitskiy: If we now ensure the chunk fits into `int`, consider changing the type to `int` here and for `this.chunkSize`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -844,21 +843,62 @@ private static long getChunkSize(final ByteBuf buffer, final int lfIndex) {
...
+            throw invalidChunkSize(hex, new NumberFormatException("Empty chunk size"));
+        }
+        long result = 0;
         for (int i = 0; i < hex.length(); ++i) {
             char c = hex.charAt(i);
             if (c == ';' || isWhitespace(c) || isISOControl(c)) {
-                hex = hex.substring(0, i);
-                break;
+                if (i == 0) {
+                    throw invalidChunkSize(hex, new NumberFormatException("Empty chunk size"));
```
idelpivnitskiy: Consider adding tests for this and `hex.isEmpty()` cases

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
daschl: `this.chunkSize` is also used for fixed length content size storage, so this won't work unless we refactor the code to use two different variables there I think

idelpivnitskiy (approved):
Thank you!

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -836,15 +836,19 @@ private void ensureTotalHeaderFieldsLimit(final String name, final int headerLin
...
         return getChunkSize(buffer.toString(buffer.readerIndex(),
                 lfIndex - 1 - buffer.readerIndex(), US_ASCII));
     }
 
-    private static long getChunkSize(String hex) {
+    /**
+     * Parses an HTTP/1.1 chunk-size token. Stops at the first {@code ';'}, whitespace, or
+     * ISO control character so any chunk-extension is ignored.
+     */
+    private static int getChunkSize(String hex) {
```
bryce-anderson: I think this is wrong: I don't see the problem with emitting a chunk size larger than Integer.MAX_VALUE. This is just to get the length of the current 'frame', not how much data we're actually going to emit on the next chunk. For example, say we get a chunk size of `Integer.MAX_VALUE * 2`. Clearly this won't fit in a single buffer- but it doesn't need to: `buffer.readableBytes()` is an `int` (and positive) so at most will emit that many bytes. This then gets subtracted from the `this.chunkSize` and we wait for more data. Rinse and repeat.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -853,9 +857,12 @@ private static long getChunkSize(String hex) {
                 break;
             }
         }
-
         try {
-            return parseUnsignedLong(hex, 16);
```
bryce-anderson, line 858: I think this was fine before just because we limit how long our hex string can be (so it can never be negative). In the case that it ever changes to be the full 64 bits, I think we need to also check if the parsed value is negative and throw if so. Kind of belt and suspenders...

bryce-anderson (changes requested):
I'm not sure what we're trying to do here. I think the commit message is at least wrong: I think we're following the protocol correctly.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
bryce-anderson: I may have used too strong of language, _wrong_ isn't quite right. I do think is a step back in capability in that we were able to correctly read large chunks before. Otoh, 2+ GiB chunks are likely not the norm so it's likely okay if we want to make this a policy decision. If we want to move forward with this strategy we should rectify the commit message to reflect that we're tightening our chunk limit to 2 GiB.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpObjectDecoderTest.java
```diff
@@ -639,6 +639,102 @@ void chunkedNoTrailersInvalidChunkSize(boolean crlf) {
         assertThat(channel.inboundMessages(), is(not(empty())));
     }
 
+    /**
+     * Validates that chunk-size values exceeding {@link Integer#MAX_VALUE} are rejected.
+     * Previously {@link Long#parseUnsignedLong} accepted them and a defensive cast in
+     * {@code READ_CHUNKED_CONTENT} silently clamped to {@link Integer#MAX_VALUE}, leaving
+     * the decoder waiting forever for body bytes that would not arrive.
```
bryce-anderson: I think this is outdated.

bryce-anderson (approved):

idelpivnitskiy (commented):
Reminder to please-please-please don't use force pushes on any PRs that got reviewed. Now I don't know what changed since my previous approval 🤷

PR #3509, opened 2026-06-16
http-netty: Enforce RFC 9112 Transfer-Encoding correctness

daschl:
#### Motivation
  HttpObjectDecoder.readHeaders does not validate Transfer-Encoding against the RFC 9112 section 6.1 rules: Transfer-Encoding on a non-HTTP/1.1 message is silently accepted, a TE list where chunked is not the final coding is silently treated as chunked-encoded regardless of role, and a message with both Transfer-Encoding and Content-Length is processed per Transfer-Encoding alone without closing the connection afterward as section 6.1 requires.

  #### Modifications
  - Reject any message that carries Transfer-Encoding with a protocol version other than HTTP/1.1, throwing a DecoderException.
  - When chunked is detected, walk the Transfer-Encoding values and require the last one to end with "chunked (case-insensitive). If chunked is present but not final, reject requests but frame responses by reading until connection close, per RFC 9112 section 6.3 item 4. If chunked never appears, leave Content-Length untouched.
  - When Transfer-Encoding governs framing and Content-Length is also present, strip Content-Length and force Connection: close (preserving any existing Connection tokens) so the connection isn't reused, per RFC 9112 section 6.1.
  - Add parameterized tests covering protocol-version handling, the full range of comma-separated and multi-header Transfer-Encoding shapes, and the close-after-respond behavior.

  #### Result
  Transfer-Encoding desync attacks via mismatched protocol version or mismatched coding ordering in a request are rejected with a clear DecoderException. Messages with both Transfer-Encoding and Content-Length are processed per Transfer-Encoding alone and the connection is closed after responding. Legitimate HTTP/1.1 + chunked, gzip,chunked, or TE-without-chunked traffic continues to decode unchanged.

##### a) ServiceTalk as a server, decoding a malformed client request

  | Malformation | Outcome | Connection |
  |---|---|---|
  | `Transfer-Encoding` on a non-HTTP/1.1 request | Rejected (`DecoderException`) | Closed |
  | `Transfer-Encoding: chunked, gzip` (chunked not final) | Rejected (`DecoderException`, "chunked must be the final Transfer-Encoding coding") | Closed |
  | `Transfer-Encoding: gzip` alone (chunked never appears) | **Now rejected** (`DecoderException`, same message) — deliberate deviation from RFC 9112, which would tolerate this and let Content-Length govern | Closed |
  | `Transfer-Encoding: chunked, chunked` (repeated chunked token) | Accepted as ordinary chunked framing — deliberate, tested, RFC-literal reading (only the *last* coding must be chunked) | Stays open |
  | `Transfer-Encoding: chunked` + `Content-Length` both present | Accepted — Content-Length stripped, `Connection: close` forced (added, existing tokens preserved) | Closed once both inbound and outbound payloads for this exchange complete (graceful close) |
  | Multiple `Content-Length` values that disagree | Rejected (`IllegalArgumentException` → wrapped `DecoderException`) | Closed |
  | Multiple `Content-Length` values that agree | Rejected too — no tolerance for duplicates, even identical ones | Closed |

##### b) ServiceTalk as a client, decoding a malformed response

  | Malformation | Outcome | Connection |
  |---|---|---|
  | `Transfer-Encoding` on a non-HTTP/1.1 response | Rejected (`DecoderException`) | Closed |
  | `Transfer-Encoding: chunked, gzip` (chunked not final) | **Now rejected** (`DecoderException`) — deliberate deviation from RFC 9112 §6.3 item 4, which would let this fall back to close-delimited framing; we reject it unconditionally instead, matching the request-side treatment | Closed |
  | `Transfer-Encoding: gzip` alone (chunked never appears) | **Now rejected** (`DecoderException`, same message) — same deliberate deviation, matching the request-side treatment | Closed |
  | `Transfer-Encoding: chunked, chunked` | Accepted — same as (a), deliberate/RFC-literal, not a bug | Stays open |
  | `Transfer-Encoding: chunked` + `Content-Length` both present | Accepted — Content-Length stripped, `Connection: close` injected into the decoded response object your client code sees, `protocolClosingInbound` fires | Closed once both inbound/outbound for this exchange complete |
  | Multiple disagreeing/duplicate `Content-Length` | Rejected, same as (a) | Closed |

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
@@ -148,6 +148,14 @@ public final class HeaderUtils {
      * The values of all {@link HttpHeaderNames#TRANSFER_ENCODING} headers are interpreted as comma-separated values,
      * with spaces between values trimmed. If any of these values is  {@link HttpHeaderValues#CHUNKED}, this method
      * return {@code true}, otherwise it returns {@code false}.
+     * <p>
+     * This checks for {@link HttpHeaderValues#CHUNKED} <b>anywhere</b> among the values, regardless of position.
+     * Per <a href="https://www.rfc-editor.org/rfc/rfc9112.html#section-6.1">RFC 9112, Section 6.1</a>, {@code
+     * chunked} only governs the actual message framing when it is the <b>final</b> coding (e.g. {@code chunked} is
+     * final in {@code "gzip, chunked"} but not in {@code "chunked, gzip"}). This method does not distinguish the
+     * two: it returns {@code true} for both. Callers that need to know whether {@code chunked} framing actually
+     * applies to the message (as opposed to merely being named somewhere in the header) must perform that
+     * additional check themselves.
      *
      * @param headers The {@link HttpHeaders} to check.
      * @return {@code} true if {@code headers} indicates {@code transfer-encoding} {@code chunked}, {@code false}
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
@@ -1110,11 +1110,15 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
         if (valueLength == chunkedLength) {
             return contentEqualsIgnoreCase(value, CHUNKED);
         }
-        final char before = value.charAt(valueLength - chunkedLength - 1);
-        if (before == ',' || before == ' ' || before == '\t') {
-            return regionMatches(value, true, valueLength - chunkedLength, CHUNKED, 0, chunkedLength);
+        // List elements are comma-separated (RFC 9110 section 5.6.1): skip any OWS (SP/HTAB) immediately
+        // preceding "chunked", then require a comma. A bare SP/HTAB with no comma (e.g. "gzip chunked") is not
+        // a valid element separator and must not be mistaken for "chunked" being a distinct, final coding.
+        int sepIndex = valueLength - chunkedLength - 1;
+        while (sepIndex >= 0 && (value.charAt(sepIndex) == ' ' || value.charAt(sepIndex) == '\t')) {
+            sepIndex--;
         }
-        return false;
+        return sepIndex >= 0 && value.charAt(sepIndex) == ','
+                && regionMatches(value, true, valueLength - chunkedLength, CHUNKED, 0, chunkedLength);
     }
 
     private static boolean isObsText(final byte value) {
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
@@ -54,6 +54,7 @@ import io.netty.handler.codec.http.HttpExpectationFailedEvent;
 import io.netty.util.AsciiString;
 import io.netty.util.ByteProcessor;
 
+import java.util.Iterator;
 import javax.annotation.Nullable;
 
 import static io.netty.handler.codec.http.HttpConstants.COLON;
@@ -72,7 +73,9 @@ import static io.servicetalk.http.api.HttpHeaderNames.SEC_WEBSOCKET_KEY1;
 import static io.servicetalk.http.api.HttpHeaderNames.SEC_WEBSOCKET_KEY2;
 import static io.servicetalk.http.api.HttpHeaderNames.SEC_WEBSOCKET_LOCATION;
 import static io.servicetalk.http.api.HttpHeaderNames.SEC_WEBSOCKET_ORIGIN;
+import static io.servicetalk.http.api.HttpHeaderNames.TRANSFER_ENCODING;
 import static io.servicetalk.http.api.HttpHeaderNames.UPGRADE;
+import static io.servicetalk.http.api.HttpHeaderValues.CHUNKED;
 import static io.servicetalk.http.api.HttpProtocolVersion.HTTP_1_0;
 import static io.servicetalk.http.api.HttpProtocolVersion.HTTP_1_1;
 import static io.servicetalk.http.api.HttpRequestMethod.GET;
@@ -748,10 +751,29 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
         if (isContentAlwaysEmpty(message)) {
             removeTransferEncodingChunked(message.headers());
             return State.SKIP_CONTROL_CHARS;
-        } else if (isTransferEncodingChunked(message.headers())) {
-            if (contentLength >= 0L && HTTP_1_1.equals(message.version())) {
-                // Remove the received content-length header(s), keep only "transfer-encoding: chunked"
-                // See https://tools.ietf.org/html/rfc7230#section-3.3.3, item 3.
+        }
+        // RFC 9112 section 6.1: Transfer-Encoding is HTTP/1.1-only.
+        if (message.headers().contains(TRANSFER_ENCODING) && !HTTP_1_1.equals(message.version())) {
+            throw new StacklessDecoderException("Transfer-Encoding is only allowed in HTTP/1.1");
+        }
+        if (isTransferEncodingChunked(message.headers())) {
+            // RFC 9112 section 6.1: chunked must be the final coding.
+            final Iterator<? extends CharSequence> encodingIt =
+                    message.headers().valuesIterator(TRANSFER_ENCODING);
+            CharSequence lastValue = encodingIt.next();
+            while (encodingIt.hasNext()) {
+                lastValue = encodingIt.next();
+            }
+            final int vLen = lastValue.length();
+            final int chunkedValueLength = CHUNKED.length();
+            if (vLen < chunkedValueLength ||
+                    !AsciiString.regionMatches(lastValue, true, vLen - chunkedValueLength,
+                            CHUNKED, 0, chunkedValueLength)) {
+                throw new StacklessDecoderException(
+                        "chunked must be the last encoding present in the Transfer-Encoding header");
+            }
+            if (contentLength >= 0L) {
+                // https://tools.ietf.org/html/rfc7230#section-3.3.3, item 3.
                 message.headers().remove(CONTENT_LENGTH);
                 this.contentLength = Long.MIN_VALUE;
             }
@@ -1074,6 +1096,10 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
 
     static final class StacklessDecoderException extends DecoderException {
         private static final long serialVersionUID = 7611225180490304156L;
+
+        StacklessDecoderException(final String message) {
+            super(message);
+        }
 
         StacklessDecoderException(final String message, final Throwable cause) {
             super(message, cause);
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
@@ -63,8 +63,10 @@ import static io.netty.handler.codec.http.HttpConstants.HT;
 import static io.netty.handler.codec.http.HttpConstants.LF;
 import static io.netty.handler.codec.http.HttpConstants.SP;
 import static io.netty.util.ByteProcessor.FIND_LF;
+import static io.servicetalk.buffer.api.CharSequences.contentEqualsIgnoreCase;
 import static io.servicetalk.buffer.api.CharSequences.emptyAsciiString;
 import static io.servicetalk.buffer.api.CharSequences.newAsciiString;
+import static io.servicetalk.buffer.api.CharSequences.regionMatches;
 import static io.servicetalk.buffer.netty.BufferUtils.newBufferFrom;
 import static io.servicetalk.concurrent.internal.FlowControlUtils.addWithOverflowProtection;
 import static io.servicetalk.http.api.HeaderUtils.isTransferEncodingChunked;
@@ -757,34 +759,30 @@ abstract class HttpObjectDecoder<T extends HttpMetaData> extends ByteToMessageDe
             if (!HTTP_1_1.equals(message.version())) {
                 throw new StacklessDecoderException("Transfer-Encoding is only allowed in HTTP/1.1");
             }
-            // RFC 9112 section 6.3: a request that has Transfer-Encoding without chunked as the
-            // final coding cannot be reliably framed and MUST be rejected. For responses the
-            // legacy "read-until-connection-close" framing still applies, so only enforce on
-            // requests.
-            if (!isTransferEncodingChunked(message.headers())) {
-                if (isDecodingRequest()) {
-                    throw new StacklessDecoderException(
-                            "Transfer-Encoding without chunked is not allowed in requests");
-                }
-            } else {
-                // RFC 9112 section 6.1: chunked must be the final coding.
-                final Iterator<? extends CharSequence> encodingIt =
-                        message.headers().valuesIterator(TRANSFER_ENCODING);
-                CharSequence lastValue = encodingIt.next();
-                while (encodingIt.hasNext()) {
-                    lastValue = encodingIt.next();
-                }
-                if (!endsWithChunkedToken(lastValue)) {
-                    throw new StacklessDecoderException(
-                            "chunked must be the last encoding present in the Transfer-Encoding header");
-                }
-                if (contentLength >= 0L) {
-                    // https://tools.ietf.org/html/rfc7230#section-3.3.3, item 3.
-                    message.headers().remove(CONTENT_LENGTH);
-                    this.contentLength = Long.MIN_VALUE;
-                }
+            // RFC 9112 section 6.3: Transfer-Encoding overrides Content-Length. Drop any received
+            // Content-Length now so it can never be used to frame the body, in any branch below.
+            if (contentLength >= 0L) {
+                // https://tools.ietf.org/html/rfc7230#section-3.3.3, item 3.
+                message.headers().remove(CONTENT_LENGTH);
+                this.contentLength = Long.MIN_VALUE;
+            }
+            // RFC 9112 section 6.1: chunked must be the final coding. There may be multiple
+            // Transfer-Encoding header lines, so check the last value of the last line.
+            final Iterator<? extends CharSequence> encodingIt =
+                    message.headers().valuesIterator(TRANSFER_ENCODING);
+            CharSequence lastValue = encodingIt.next();
+            while (encodingIt.hasNext()) {
+                lastValue = encodingIt.next();
+            }
+            if (endsWithChunkedToken(lastValue)) {
                 return State.READ_CHUNK_SIZE;
             }
+            // RFC 9112 section 6.3: chunked is not the final coding. A request cannot be framed
+            // reliably and MUST be rejected; a response is framed by reading until connection close.
+            if (isDecodingRequest()) {
+                throw new StacklessDecoderException("chunked must be the final Transfer-Encoding coding in requests");
+            }
+            return State.READ_VARIABLE_LENGTH_CONTENT;
         }
         if (contentLength >= 0L) {
             return State.READ_FIXED_LENGTH_CONTENT;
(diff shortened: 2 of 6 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -747,12 +750,36 @@ private State readHeaders(final ByteBuf buffer) {
...
-        } else if (isTransferEncodingChunked(message.headers())) {
-            if (contentLength >= 0L && HTTP_1_1.equals(message.version())) {
-                // Remove the received content-length header(s), keep only "transfer-encoding: chunked"
-                // See https://tools.ietf.org/html/rfc7230#section-3.3.3, item 3.
-                message.headers().remove(CONTENT_LENGTH);
-                this.contentLength = Long.MIN_VALUE;
+        }
+        // RFC 9112 section 6.1: Transfer-Encoding is HTTP/1.1-only.
+        if (message.headers().contains(TRANSFER_ENCODING) && !HTTP_1_1.equals(message.version())) {
+            throw new DecoderException("Transfer-Encoding is only allowed in HTTP/1.1 or newer");
```
idelpivnitskiy: The message says "or newer" but the check has strict equals comparison. Should we allow any version in between 1.1 but below 2.0?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -747,12 +750,36 @@ private State readHeaders(final ByteBuf buffer) {
...
+        }
+        if (isTransferEncodingChunked(message.headers())) {
+            if (HTTP_1_1.equals(message.version())) {
```
idelpivnitskiy: Same question here

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -747,12 +750,36 @@ private State readHeaders(final ByteBuf buffer) {
...
+                // RFC 9112 section 6.1: chunked must be the final coding.
+                final Iterator<? extends CharSequence> encodingIt =
+                        message.headers().valuesIterator(TRANSFER_ENCODING);
```
idelpivnitskiy: If we care only about the `lastValue`, should we just parse backward for 7 symbols case-insensitively? Also, does it apply for all other cases that use `HeaderUtils.isTransferEncodingChunked`? If yes, we may need to enhance that utility instead of adding an extra check here

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
daschl: good catch, dropped the or newer since this is what netty also does. (exact match)

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
daschl: I think the code as-is aligns with netty and should be fine? If we find it to be a problem we can revisit, but we are not allocating I think on the iterator.

bryce-anderson (approved):

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpObjectDecoderTest.java
```diff
@@ -703,6 +703,103 @@ void chunkedNoContentNoTrailers(boolean crlf) {
...
+            arguments.add(Arguments.of("chunked, gzip", true, crlf));
+            // Same with whitespace before comma - rejected.
+            arguments.add(Arguments.of("chunked , gzip", true, crlf));
+            // Multi-header: gzip first, chunked last - accepted.
+            arguments.add(Arguments.of("gzip" + br + "Transfer-Encoding: chunked", false, crlf));
+            // Multi-header: chunked first, gzip last (length 4) - rejected.
+            arguments.add(Arguments.of("chunked" + br + "Transfer-Encoding: gzip", true, crlf));
+            // Multi-header: chunked first, deflate last (length 7, same as "chunked") - rejected.
+            // This case demonstrates the gap closed vs Netty's permissive check.
+            arguments.add(Arguments.of("chunked" + br + "Transfer-Encoding: deflate", true, crlf));
```
idelpivnitskiy: Could you please also add a case for `TE: chunked then TE: ` (empty)?

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -747,10 +750,29 @@ private State readHeaders(final ByteBuf buffer) {
...
+                    message.headers().valuesIterator(TRANSFER_ENCODING);
+            CharSequence lastValue = encodingIt.next();
+            while (encodingIt.hasNext()) {
+                lastValue = encodingIt.next();
+            }
+            final int vLen = lastValue.length();
+            final int chunkedValueLength = CHUNKED.length();
+            if (vLen < chunkedValueLength ||
+                    !AsciiString.regionMatches(lastValue, true, vLen - chunkedValueLength,
+                            CHUNKED, 0, chunkedValueLength)) {
```
idelpivnitskiy: This will treat `Transfer-Encoding: notchunked` as `chunked` 😄 
`isTransferEncodingChunked` has the same bug

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -747,10 +750,29 @@ private State readHeaders(final ByteBuf buffer) {
...
-        } else if (isTransferEncodingChunked(message.headers())) {
-            if (contentLength >= 0L && HTTP_1_1.equals(message.version())) {
-                // Remove the received content-length header(s), keep only "transfer-encoding: chunked"
-                // See https://tools.ietf.org/html/rfc7230#section-3.3.3, item 3.
+        }
+        // RFC 9112 section 6.1: Transfer-Encoding is HTTP/1.1-only.
+        if (message.headers().contains(TRANSFER_ENCODING) && !HTTP_1_1.equals(message.version())) {
+            throw new StacklessDecoderException("Transfer-Encoding is only allowed in HTTP/1.1");
+        }
+        if (isTransferEncodingChunked(message.headers())) {
```
idelpivnitskiy: I think we should apply the new validation for any message that contains `Transfer-Encoding` header as the spec says: 
> If any transfer coding other than chunked is applied to a request's content, the sender MUST apply chunked as the final transfer coding to ensure that the message is properly framed

Please add tests for `Transfer-Encoding: gzip` with and without `Content-Length`

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
idelpivnitskiy: ah, nvm about "should we just parse backward" part. the iterator is actually to ensure we read the last header entry in case there are multiple `Transfer-Encoding` headers.

I'm still worried about `HeaderUtils.isTransferEncodingChunked` part. If this rule about the last encoding always to be "chunked" is universal, should we incorporate those checks into `isTransferEncodingChunked` instead of validating only in the decoder? What about cases when users use ServiceTalk to write a request with `Transfer-Encoding: chunked` followed by `Transfer-Encoding: gzip`? Our write side logic may incorrectly treat it as "chunked"

idelpivnitskiy (changes requested):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -751,33 +751,41 @@ private State readHeaders(final ByteBuf buffer) {
...
             }
-            if (contentLength >= 0L) {
-                // https://tools.ietf.org/html/rfc7230#section-3.3.3, item 3.
-                message.headers().remove(CONTENT_LENGTH);
-                this.contentLength = Long.MIN_VALUE;
+            // RFC 9112 section 6.3: a request that has Transfer-Encoding without chunked as the
+            // final coding cannot be reliably framed and MUST be rejected. For responses the
+            // legacy "read-until-connection-close" framing still applies, so only enforce on
+            // requests.
+            if (!isTransferEncodingChunked(message.headers())) {
```
idelpivnitskiy: Imho we can skip this check and go straight to the `else` case as it doesn't matter if "chunked" is present in the values or not, we still need to validate that it's the last encoding

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -1052,6 +1082,26 @@ static boolean isVCHAR(final byte value) {
...
+        final int chunkedLen = CHUNKED.length();
+        if (vLen < chunkedLen ||
+                !AsciiString.regionMatches(value, true, vLen - chunkedLen, CHUNKED, 0, chunkedLen)) {
+            return false;
+        }
+        if (vLen == chunkedLen) {
+            return true;
+        }
+        final char before = value.charAt(vLen - chunkedLen - 1);
+        return before == ',' || before == ' ' || before == '\t';
```
idelpivnitskiy: `regionMatches` is an expensive call, consider deferring it until necessary while prioritizing the most probably happy path. we should also use our `CharSequences` instead of netty's `AsciiString`:
```suggestion
        if (vLen < chunkedLen) {
            return false;
        }
        if (vLen == chunkedLen) {
            return CharSequences.contentEqualsIgnoreCase(value, CHUNKED);
        }
        final char before = value.charAt(vLen - chunkedLen - 1);
        if (before == ',' || before == ' ' || before == '\t') {
            return CharSequences.regionMatches(value, true, vLen - chunkedLen, CHUNKED, 0, chunkedLen);
        }
        return false;
```

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpObjectDecoderTest.java
```diff
@@ -786,6 +786,157 @@ void chunkedNoContentNoTrailers(boolean crlf) {
...
+    private static Collection<Arguments> transferEncodingChunkedLastArgs() {
+        final List<Arguments> arguments = new ArrayList<>();
+        for (boolean crlf : new boolean[] {true, false}) {
+            final String br = crlf ? "\r\n" : "\n";
+            // Single value: chunked alone - accepted.
+            arguments.add(Arguments.of("chunked", false, crlf));
+            // Comma-separated, chunked last - accepted.
+            arguments.add(Arguments.of("gzip, chunked", false, crlf));
+            // No space after comma - still accepted (chunked still last).
+            arguments.add(Arguments.of("gzip,chunked", false, crlf));
```
idelpivnitskiy: consider adding at least one case for comma separated case-insensitive match

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpObjectDecoderTest.java
```diff
@@ -786,6 +786,157 @@ void chunkedNoContentNoTrailers(boolean crlf) {
...
+    }
+
+    private static Collection<Arguments> transferEncodingChunkedLastArgs() {
+        final List<Arguments> arguments = new ArrayList<>();
+        for (boolean crlf : new boolean[] {true, false}) {
+            final String br = crlf ? "\r\n" : "\n";
+            // Single value: chunked alone - accepted.
+            arguments.add(Arguments.of("chunked", false, crlf));
+            // Comma-separated, chunked last - accepted.
+            arguments.add(Arguments.of("gzip, chunked", false, crlf));
```
idelpivnitskiy: consider also covering `\t`

idelpivnitskiy (commented):
Logic lgtm, a few minor suggestions, and we can handle `HeaderUtils.isTransferEncodingChunked` in a separate PR

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
daschl: I think this would cause a behavioral difference, since according to the RFC the request and response behavior is different. I'll come up with an alternative proposal that represents this differently.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
```diff
@@ -148,6 +148,14 @@ static String toString(final HttpHeaders headers,
...
      * with spaces between values trimmed. If any of these values is  {@link HttpHeaderValues#CHUNKED}, this method
      * return {@code true}, otherwise it returns {@code false}.
+     * <p>
+     * This checks for {@link HttpHeaderValues#CHUNKED} <b>anywhere</b> among the values, regardless of position.
+     * Per <a href="https://www.rfc-editor.org/rfc/rfc9112.html#section-6.1">RFC 9112, Section 6.1</a>, {@code
+     * chunked} only governs the actual message framing when it is the <b>final</b> coding (e.g. {@code chunked} is
+     * final in {@code "gzip, chunked"} but not in {@code "chunked, gzip"}). This method does not distinguish the
+     * two: it returns {@code true} for both. Callers that need to know whether {@code chunked} framing actually
+     * applies to the message (as opposed to merely being named somewhere in the header) must perform that
+     * additional check themselves.
```
bryce-anderson, line 158: This function is staring to feel like a foot-gun. I think chunked _must_ be last or else it's a protocol error (or it will probably turn into a protocol error 😂).

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -748,15 +753,46 @@ private State readHeaders(final ByteBuf buffer) {
...
+                if (chunkedIsFinal) {
+                    return State.READ_CHUNK_SIZE;
+                }
+                // RFC 9112 section 6.3: chunked is not the final coding. A request cannot be framed
+                // reliably and MUST be rejected; a response is framed by reading until connection close.
+                if (isDecodingRequest()) {
+                    throw new StacklessDecoderException(
+                            "chunked must be the final Transfer-Encoding coding in requests");
+                }
+                return State.READ_VARIABLE_LENGTH_CONTENT;
```
bryce-anderson: I think this is technically right per the RFC, but I suspect the best thing to do would be to abort the connection. Why would the server ever send a `transfer-encoding: chunked, gzip` with the intent for the length to be dictated by the lifetime of the connection? The most common scenario would be that the header just got added in the wrong order and it's really chunked framed, so we'll read a corrupt payload.

Exploring what others do, many implementations such as golang, nginx, envoy, etc don't accept anything other than a simple "chunked". See [here](https://github.com/golang/go/blob/d90bf6d72263ef6eb0e26b1f6b3fd8322dcfba98/src/net/http/transfer.go#L650-L659) for an example.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -1054,6 +1090,29 @@ static boolean isVCHAR(final byte value) {
...
     }
 
+    /**
+     * Checks if the given value ends with exactly {@code chunked} (case-insensitive).
+     *
+     * @param value the value to check
+     * @return returns {@code true} if {@code value} ends with the {@code chunked} token (case-insensitive),
+     * either as the entire value or as the final comma-separated token.
+     */
+    private static boolean endsWithChunkedToken(final CharSequence value) {
```
bryce-anderson: I think we technically can get scenarios where we have "chunked, " that are technically correct. Maybe what we need to do is a more robust splitting of the header values. The overwhelming most common case will be a simple "chunked" which we can fast path.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -1054,6 +1090,29 @@ static boolean isVCHAR(final byte value) {
...
+        final int valueLength = value.length();
+        final int chunkedLength = CHUNKED.length();
+        if (valueLength < chunkedLength) {
+            return false;
+        }
+        if (valueLength == chunkedLength) {
+            return contentEqualsIgnoreCase(value, CHUNKED);
+        }
+        final char before = value.charAt(valueLength - chunkedLength - 1);
+        if (before == ',' || before == ' ' || before == '\t') {
```
bryce-anderson: I think this means we also accept "gzip chunked", which feels overly permissive.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HeaderUtils.java
daschl: that is true, although I think to properly fix it we also need to fix all callsites/codepaths to ensure they don't blow up "late" where previously it would let it pass through. I think this is out of scope for this pr.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
daschl: fast path should be in place on line 1111 below.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
daschl: fixed

bryce-anderson (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -748,15 +756,47 @@ private State readHeaders(final ByteBuf buffer) {
...
+            // There may be multiple Transfer-Encoding header lines; only the last coding of the last line
+            // decides whether chunked framing applies, so a single pass tracks the last value seen.
+            final Iterator<? extends CharSequence> encodingIt = message.headers().valuesIterator(TRANSFER_ENCODING);
+            CharSequence lastValue = encodingIt.next();
+            while (encodingIt.hasNext()) {
+                lastValue = encodingIt.next();
+            }
+            // RFC 9112 section 6.1: chunked must be the final coding to apply chunked framing. Deliberate
+            // deviation from RFC 9112 section 6.3, which would otherwise let some non-chunked-final shapes
+            // fall back to Content-Length (e.g. plain "gzip") or to framing a response by connection close
```
bryce-anderson, line 768: One thing worth noting here is that while we reject `transfer-encoding: gzip` we will accept `transfer-encoding: gzip, chunked` and more oddly 
```
transfer-encoding: chunked
transfer-encoding: gzip
transfer-encoding: chunked
```
I suspect the last one is potentially a problem.

idelpivnitskiy: Hm, yes, section 6.1 says:
> A sender MUST NOT apply the chunked transfer coding more than once to a message body (i.e., chunking an already chunked message is not allowed).

This is getting more complicated 😞

daschl: reworked the parsing logic for it

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -1054,6 +1085,58 @@ static boolean isVCHAR(final byte value) {
...
+        boolean chunkedSeen = false;
+        boolean lastTokenIsChunked = false;
+        while (encodingIt.hasNext()) {
+            final CharSequence value = encodingIt.next();
+            final int valueLength = value.length();
+            int tokenStart = 0;
+            // List elements are comma-separated (RFC 9110 section 5.6.1); walk tokens left-to-right, trimming
+            // any OWS (SP/HTAB) at their edges, so a bare SP/HTAB with no comma (e.g. "gzip chunked") is not
+            // mistaken for two distinct tokens.
+            for (int i = 0; i <= valueLength; i++) {
```
bryce-anderson, line 1109: This is obviously subjective but I find this kind of hard to read. I think this for loop basically boils down to a big `indexOf(value, ',')` call. Maybe this would be easier to read as a more classical two pointer walking algorithm:
```
int tokenStart = 0;
int commaIdx;
do {
  commaIdx = indexOf(value, ',', tokenStart);
  int start = tokenStart;
  int end = commaIdx < 0 ? valueLength : commaIdx;
  // Trim OWS ...
  if (end > start) { // ignore empty list elements
    lastTokenIsChunked = end - start == chunkedLength &&
        regionMatches(value, true, start, CHUNKED, 0, chunkedLength);
    if (lastTokenIsChunked) {
      if (chunkedSeen) {
        throw new StacklessDecoderException(
                                "chunked transfer coding must not be applied more than once");
      }
      chunkedSeen = true;
    }
  }
  tokenStart = commaIdx + 1;
} while (commaIdx >= 0);
```

bryce-anderson (approved):
Looks correct to me. Feel free to ignore my readability suggestion - the only behavior is to ignore empty elements and I don't think it matters in real life anyway.

idelpivnitskiy (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
daschl: I merged this since it has been going for too long - I'll look into it for a follow-up 👍 thanks!

PR #3510, opened 2026-06-16
http-netty: Reject HTTP/1.1 messages with both Transfer-Encoding and Content-Length

daschl:
#### Motivation
HttpObjectDecoder.readHeaders silently strips Content-Length when an
  HTTP/1.1 message carries both Transfer-Encoding and Content-Length.
  RFC 9112 section 6.1 considers the combination a request-smuggling
  vector and obsoletes the RFC 7230 strip-and-continue handling.

#### Modifications
  - Reject the combination by default with a DecoderException.
  - Expose H1SpecExceptions.Builder#allowTransferEncodingWithContentLength
    to opt back into RFC 7230 lenient handling. When enabled, strip
    Content-Length and (for requests only, matching Netty's
    HttpUtil.setKeepAlive call) force Connection: close.
  - Plumb the flag through H1SpecExceptions, HttpRequestDecoder /
    HttpResponseDecoder constructors, and the NettyHttpServer /
    HttpClientChannelInitializer call sites.
  - Update chunkedWithContentLength to assert strict rejection; add
    chunkedWithContentLengthLenient to cover the opt-in path.

#### Result
  HTTP/1.1 + TE + CL is rejected with a clear DecoderException that
  closes the connection. Operators with a non-compliant peer can
  opt back into legacy strip-and-close via
  allowTransferEncodingWithContentLength.


idelpivnitskiy (commented):
The latest version of the spec permits this behavior, clarifying:
> the server MUST close the connection after responding to such a request to avoid the potential attacks

PR #3522, opened 2026-06-18
ci: fix publications of junit test reports

idelpivnitskiy:
Motivation:

The Test Reports do not work for quite a long time. Also, `scacap/action-surefire-report` is now deprecated.

Modifications:
- Replace `scacap/action-surefire-report` with the more popular `EnricoMi/publish-unit-test-result-action`, which is also used by pkl.
- Replace custom `dawidd6/action-download-artifact` action with official `actions/download-artifact` from GitHub.
- Update permissions.
- Publush Test Reports only if pipelines were not cancelled.

Result:

jUnit reports are published on GitHub, making it easier to see the results instead of manual investigation of the build logs.

### DIFF
--- a/.github/workflows/ci-prb-reports.yml
+++ b/.github/workflows/ci-prb-reports.yml
@@ -8,15 +8,15 @@ on:
       - completed
 jobs:
   tests:
+    if: ${{ github.event.workflow_run.conclusion == 'success' || github.event.workflow_run.conclusion == 'failure' }}
     permissions:
-      contents: read
       actions: read
       checks: write
     runs-on: ubuntu-latest
     strategy:
       # Matrix should be coordinated with ci-prb.yml.
       matrix:
-        java: [ 8, 11, 17, 21, 24 ]
+        java: [ 8, 11, 17, 21, 25 ]
         os: [ ubuntu-latest ]
         os_label: [ ubuntu ]
         include:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-prb-reports.yml
```diff
@@ -25,18 +25,16 @@ jobs:
             os_label: macos
     steps:
       - name: Download Artifacts
-        uses: dawidd6/action-download-artifact@b6e2e70617bc3265edd6dab6c906732b2f1ae151 # v21
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
```
bryce-anderson, line 28: Do you want to also update ci-prq-reports.yml?

.github/workflows/ci-prb-reports.yml
```diff
@@ -8,15 +8,15 @@ on:
...
-      contents: read
       actions: read
       checks: write
     runs-on: ubuntu-latest
     strategy:
       # Matrix should be coordinated with ci-prb.yml.
       matrix:
-        java: [ 8, 11, 17, 21, 24 ]
+        java: [ 8, 11, 17, 21, 25 ]
         os: [ ubuntu-latest ]
```
bryce-anderson, line 20: Not related, but since you're here, do you want to remove the `matrix.os` or use it in the `runs-on`?

bryce-anderson: Nevermind, this is more messy than I thought.

.github/workflows/ci-prb-reports.yml
idelpivnitskiy: this will come in a follow-up PR focussed on prq, here only trying to fix prb-reports

bryce-anderson: Okay, so long as it gets consolidated.

bryce-anderson (approved):

PR #3523, opened 2026-06-19
http-netty: Close connection after responding to TE+CL HTTP/1.1 messages

daschl:
#### Motivation
RFC 9112 section 6.1 states:

> A server MAY reject a request that contains both Content-Length
> and Transfer-Encoding or process such a request in accordance
> with the Transfer-Encoding alone. Regardless, the server MUST
> close the connection after responding to such a request to
> avoid the potential attacks.

#### Modifications
Up until now, we are not following this rule, so this changeset explicitly closes the connection if the condition is identified.

The tests are adjusted to reflect/test the new behavior.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
@@ -54,6 +54,7 @@ import io.netty.handler.codec.http.HttpExpectationFailedEvent;
 import io.netty.util.AsciiString;
 import io.netty.util.ByteProcessor;
 
+import java.util.Iterator;
 import javax.annotation.Nullable;
 
 import static io.netty.handler.codec.http.HttpConstants.COLON;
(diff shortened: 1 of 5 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpObjectDecoder.java
```diff
@@ -778,9 +780,12 @@ private State readHeaders(final ByteBuf buffer) {
                             "chunked must be the last encoding present in the Transfer-Encoding header");
                 }
                 if (contentLength >= 0L) {
-                    // https://tools.ietf.org/html/rfc7230#section-3.3.3, item 3.
+                    // RFC 9112 section 6.1: forbidden combination. Process per Transfer-Encoding
+                    // alone, force Connection: close - the shouldClose(message) check below
+                    // signals the close handler so the connection is torn down after this exchange.
                     message.headers().remove(CONTENT_LENGTH);
                     this.contentLength = Long.MIN_VALUE;
+                    message.headers().set(CONNECTION, CLOSE);
```
idelpivnitskiy: The request may already have `connection` header wither with `close` value or with other values, like `keep-alive`,  `upgrade`, etc. Consider using `HeaderUtils.isConnectionClose(headers)` utility to check first, then use `add` instead of `set` to avoid override for already defined values. Test coverage for these cases would be appreciated

PR #3524, opened 2026-06-22
http-netty: Do not store sensitive headers in HTTP/2 HPACK dynamic table

igavrysh:
#### Motivation
Per RFC 7541 §7.1.3, sensitive headers should be marked as never-indexed to prevent storage in the HPACK dynamic table. Without this, intermediaries on a long-lived HTTP/2 connection can retain indexed copies of headers like `Authorization`, `Cookie`, `Set-Cookie`, and `Proxy-Authorization`.

#### Modifications
Update `DEFAULT_SENSITIVITY_DETECTOR` to recognize and flag sensitive headers as never-indexed.

#### Result
`DEFAULT_SENSITIVITY_DETECTOR` now correctly flags sensitive headers, preventing them from being added to the HPACK dynamic table.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaderNames.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaderNames.java
@@ -215,7 +215,7 @@ public final class HttpHeaderNames {
      */
     public static final CharSequence DPOP = newAsciiString("dpop");
     /**
-     * {@code "dpop"}
+     * {@code "dpop-nonce"}
      *
      * @see <a href="https://tools.ietf.org/html/rfc9449#name-authorization-server-provid">RFC 9449, section 8</a>
      */
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaderNames.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaderNames.java
@@ -209,6 +209,18 @@ public final class HttpHeaderNames {
      */
     public static final CharSequence DATE = newAsciiString("date");
     /**
+     * {@code "dpop"}
+     *
+     * @see <a href="https://tools.ietf.org/html/rfc9449#section-4.1">RFC 9449, section 4.1</a>
+     */
+    public static final CharSequence DPOP = newAsciiString("dpop");
+    /**
+     * {@code "dpop"}
+     *
+     * @see <a href="https://tools.ietf.org/html/rfc9449#name-authorization-server-provid">RFC 9449, section 8</a>
+     */
+    public static final CharSequence DPOP_NONCE = newAsciiString("dpop-nonce");
+    /**
      * {@code "etag"}
      *
      * @see <a href="https://tools.ietf.org/html/rfc7232#section-2.3">RFC7232, section 2.3</a>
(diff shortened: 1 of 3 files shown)


idelpivnitskiy (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ProtocolConfigBuilder.java
```diff
@@ -39,7 +47,29 @@
...
+
+    private static final CharSequence API_KEY_SUFFIX = "api-key";
+    private static final CharSequence AUTHENTICATE_SUFFIX = "authenticate";
+    private static final CharSequence SIGNATURE_SUFFIX = "signature";
+    private static final CharSequence TOKEN_SUFFIX = "token";
+    private static final CharSequence TOKEN_ID_SUFFIX = "token-id";
+
+    // Marks common auth/credential headers as <a href="https://tools.ietf.org/html/rfc7541#section-7.1.3">sensitive</a>
+    // so they aren't stored in the HPACK dynamic table; includes the
+    // <a href="https://datatracker.ietf.org/doc/html/rfc9449#section-4.1">DPoP</a> proof header.
```
bryce-anderson, line 59: Feel free to ignore: I think we'll technically be overly conservative - there are likely some '*-token' headers that are fine, so consider adding a note that we're erring on the safe side.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/HttpHeaderNames.java
```diff
@@ -208,6 +208,18 @@ public final class HttpHeaderNames {
...
      */
     public static final CharSequence DATE = newAsciiString("date");
+    /**
+     * {@code "dpop"}
+     *
+     * @see <a href="https://tools.ietf.org/html/rfc9449#section-4.1">RFC 9449, section 4.1</a>
+     */
+    public static final CharSequence DPOP = newAsciiString("dpop");
+    /**
+     * {@code "dpop"}
```
bryce-anderson: Is this supposed to be `{@code "dpop-nonce"}`?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpProtocolConfigTest.java
```diff
@@ -203,4 +209,75 @@ void maxTotalHeaderFieldsLengthWorksWhenEqual() {
...
+
+    @Test
+    void defaultDetectorIsCaseInsensitive() {
+        BiPredicate<CharSequence, CharSequence> detector = h2Default().headersSensitivityDetector();
+        assertTrue(detector.test("AUTHORIZATION", "Bearer abc"));
+        assertTrue(detector.test("Authorization", "Bearer abc"));
+        assertTrue(detector.test("proxy-authorization", "abc"));
+        assertTrue(detector.test("x-original-authorization", "abc"));
+        assertTrue(detector.test("x-forwarded-authorization", "abc"));
+        assertTrue(detector.test("Set-Cookie", "sid=1; Path=/"));
```
bryce-anderson, line 270: Maybe `set-cookie2` as well?

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpProtocolConfigTest.java
```diff
@@ -203,4 +209,75 @@ void maxTotalHeaderFieldsLengthWorksWhenEqual() {
         assertEquals(32 * 1024, config.maxHeaderFieldLength());
         assertEquals(32 * 1024, config.maxTotalHeaderFieldsLength());
     }
+
+    @Test
+    void defaultDetectorMarksSensitiveHeaders() {
+        BiPredicate<CharSequence, CharSequence> detector = h2Default().headersSensitivityDetector();
+        assertTrue(detector.test("authorization", "Bearer abc"));
```
bryce-anderson, line 218: What do you think about also testing against `AsciiString` in addition to `String`? We could make a wrapper that wraps the original BiPredicate and tests the value twice - once with the String and once with an AsciiString, and makes sure they have the same value then returns the result.

igavrysh: added `bothCharSequenceTypes` local static method for testing both char seqs

bryce-anderson (approved):

PR #3526, opened 2026-06-23
http-netty: validate HTTP/1 request-target

Nadav0077:
#### Motivation
HTTP/1.x request targets are written directly into the request line during encoding. Raw control characters are not valid in that position and should be rejected before bytes are written to the transport.

The inbound decoder already rejects request targets with control characters; this change applies the same expectation to outbound HTTP/1.x encoding.

#### Modifications
- Validate the HTTP/1.x request target before serializing the initial line.
- Reject request targets that contain raw control characters.
- Add `HttpRequestEncoderTest` coverage for CR, LF, CRLF, form-feed, and DEL.

#### Result
Invalid HTTP/1.x request targets fail during encoding instead of producing malformed start lines. Valid percent-encoded request targets are unchanged.


Nadav0077:
Closing this draft in favor of #3527, which uses the same commit from a clean branch name.

PR #3527, opened 2026-06-23
http-netty: validate HTTP/1 request-target

Nadav0077:
#### Motivation
HTTP/1.x request targets are written directly into the request line during encoding. Raw control characters are not valid in that position and should be rejected before bytes are written to the transport.

The inbound decoder already rejects request targets with control characters; this change applies the same expectation to outbound HTTP/1.x encoding.

#### Modifications
- Validate the HTTP/1.x request target before serializing the initial line.
- Reject request targets that contain raw control characters.
- Add `HttpRequestEncoderTest` coverage for CR, LF, CRLF, form-feed, and DEL.

#### Result
Invalid HTTP/1.x request targets fail during encoding instead of producing malformed start lines. Valid percent-encoded request targets are unchanged.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestEncoder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestEncoder.java
@@ -38,6 +38,7 @@ import io.servicetalk.http.netty.HttpResponseDecoder.Signal;
 import io.servicetalk.transport.netty.internal.CloseHandler;
 import io.servicetalk.transport.netty.internal.DefaultNettyConnection.CancelWriteUserEvent;
 import io.servicetalk.transport.netty.internal.DefaultNettyConnection.ContinueUserEvent;
+import io.servicetalk.utils.internal.IllegalCharacterException;
 
 import io.netty.channel.ChannelHandlerContext;
 
@@ -54,6 +55,7 @@ import static java.util.Objects.requireNonNull;
 final class HttpRequestEncoder extends HttpObjectEncoder<HttpRequestMetaData> {
     private static final char SLASH = '/';
     private static final char QUESTION_MARK = '?';
+    private static final char DEL = 0x7f;
     private static final int SLASH_AND_SPACE_SHORT = (SLASH << 8) | SP;
     private static final int SPACE_SLASH_AND_SPACE_MEDIUM = (SP << 16) | SLASH_AND_SPACE_SHORT;
 
@@ -107,6 +109,7 @@ final class HttpRequestEncoder extends HttpObjectEncoder<HttpRequestMetaData> {
         message.method().writeTo(stBuffer);
 
         String uri = message.requestTarget();
+        validateRequestTarget(uri);
 
         if (uri.isEmpty()) {
             // Add " / " as absolute path if uri is not present.
@@ -147,6 +150,17 @@ final class HttpRequestEncoder extends HttpObjectEncoder<HttpRequestMetaData> {
         // if this happens just force http/1.1 to avoid generating an invalid request.
         (message.version().major() == 1 ? message.version() : HTTP_1_1).writeTo(stBuffer);
         stBuffer.writeShort(CRLF_SHORT);
+    }
+
+    private static void validateRequestTarget(final String requestTarget) {
+        // This guards HTTP/1.x request-line framing only. Preserve existing behavior for non-ASCII
+        // request-target characters: they are written via UTF-8 and do not encode to SP/CTL/DEL bytes.
+        for (int i = 0; i < requestTarget.length(); ++i) {
+            final char c = requestTarget.charAt(i);
+            if (c <= SP || c == DEL) {
+                throw new IllegalCharacterException((byte) c, "request-target character (no SP / CTL / DEL)");
+            }
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestEncoder.java
```diff
@@ -149,6 +151,16 @@ protected void encodeInitialLine(ChannelHandlerContext ctx, Buffer stBuffer, Htt
         stBuffer.writeShort(CRLF_SHORT);
     }
 
+    private static void validateRequestTarget(final String requestTarget) {
+        for (int i = 0; i < requestTarget.length(); ++i) {
+            final char c = requestTarget.charAt(i);
+            if (c <= 0x1f || c == DEL) {
+                throw new IllegalArgumentException(
+                        "Invalid HTTP/1.x request-target: contains prohibited control character at index " + i);
```
daschl: consider using `IllegalCharacterException` which is already used in the `HttpObjectDecoder` to convey similar messages.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestEncoder.java
```diff
@@ -149,6 +151,16 @@ protected void encodeInitialLine(ChannelHandlerContext ctx, Buffer stBuffer, Htt
         stBuffer.writeShort(CRLF_SHORT);
     }
 
+    private static void validateRequestTarget(final String requestTarget) {
+        for (int i = 0; i < requestTarget.length(); ++i) {
+            final char c = requestTarget.charAt(i);
+            if (c <= 0x1f || c == DEL) {
```
daschl: If we tighten the validation, I think it also makes sense to include `SP (0x20)` .. so instead of `0x1f` we could include `0x20`.

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpRequestEncoderTest.java
```diff
@@ -162,6 +162,18 @@ void contentLengthNoTrailersHeaderWhiteSpaceThrowByDefault() {
         assertFalse(channel.finishAndReleaseAll());
     }
 
+    @ParameterizedTest(name = "{index}")
+    @ValueSource(strings = {"/safe\r\nX-Injected: yes", "/safe\nX-Injected: yes", "/safe\rX-Injected: yes",
+            "/safe\f/target", "/safe\u007Ftarget"})
```
daschl: Good start, maybe this can be extended further:

```
      "/safe\r\nX-Injected: yes",
      "/safe\nX-Injected: yes",
      "/safe\rX-Injected: yes",
      "/safe\f/target",
      "/safe\u007Ftarget",
      // additions:
      "/safe\u0000target",   // NUL — canonical C-string termination attack
      "/safe\ttarget",       // HTAB — explicitly forbidden by RFC 9110 §5.5
      "/safe\u000Btarget",   // VT  — completes the CR/LF/VT/FF whitespace family
      "/path with space",    // SP (0x20) — request-line delimiter, RFC 9112 §3.2 violation
      "\u0001/safe",         // control at index 0 — boundary, ensures the loop covers the prefix
      "/safe\u001F",         // 0x1F — upper boundary of the rejected control range
```

daschl (changes requested):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestEncoder.java
Nadav0077: Good idea, done 👍

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestEncoder.java
Nadav0077: Thanks, done 👍

servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/HttpRequestEncoderTest.java
Nadav0077: Updated with the additional request target cases, thanks.

daschl (approved):

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestEncoder.java
```diff
@@ -149,6 +152,15 @@ protected void encodeInitialLine(ChannelHandlerContext ctx, Buffer stBuffer, Htt
         stBuffer.writeShort(CRLF_SHORT);
     }
 
+    private static void validateRequestTarget(final String requestTarget) {
```
bryce-anderson, line 155: I think _technically_ we're not doing the complete validation since we allow non-ASCII chars. I think we should continue to due so just in case someone is depending on it and it's not a realistic smuggling vector, but we should probably add a note that it's deliberate.

bryce-anderson (approved):
Approved pending a comment about deliberately allowing chars >= 0x80.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/HttpRequestEncoder.java
Nadav0077: Added a note. Thanks

bryce-anderson (approved):

PR #3528, opened 2026-06-23
Validate HTTP header values by default

### LINKED ISSUE #3022: Bare CR improperly allowed in HTTP/1.1 header values
# The bug

ServiceTalk allows bare CR within HTTP/1.1 header values. This is disallowed by the RFCs.

RFC 9110:
> Field values containing CR, LF, or NUL characters are invalid and dangerous, due to the varying ways that implementations might parse and interpret those characters; a recipient of CR, LF, or NUL within a field value MUST either reject the message or replace each of those characters with SP before further processing or forwarding of that message.

RFC 9112:
> A sender MUST NOT generate a bare CR (a CR character not immediately followed by LF) within any protocol elements other than the content. A recipient of such a bare CR MUST consider that element to be invalid or replace each bare CR with SP before processing the element or forwarding the message.

ServiceTalk doesn't enforce this rule.

# To reproduce

1. Start a ServiceTalk-based HTTP/1.1 server that echos back the parsed components of the request, such as [this one](https://github.com/narfindustries/http-garden/blob/main/images/servicetalk/GardenServer.java).
2. Send it a request with a bare CR in a header value, and extract the interpreted header value from the response:
```sh
printf 'GET / HTTP/1.1\r\nTest: a\ra\r\n\r\n' \
    | nc localhost 80 \
    | grep '"headers"' \
    | jq '.["headers"][0][1]' \
    | sed 's/"//g' \
    | base64 -d \
    | xxd
```
3. Observe that the CR made it into the header value:
```
00000000: 610d 61                                  a.a
```

# What other HTTP implementations do

Reject the request with a 400:
- AIOHTTP                                                                                     - Apache httpd
- FastHTTP
- Go net/http
- H2O
- HAProxy
- Hyper
- Hypercorn
- Jetty
- Libevent
- Lighttpd
- Nginx
- Node.js
- Passenger
- Puma
- Tomcat
- Unicorn
- Uvicorn
- Waitress
- WEBrick

Close the connection without responding:
- Mongoose
- Netty

Translate the CR to a space:
- Libsoup
- Twisted
- LiteSpeed

Let it through (i.e., have the same bug as ServiceTalk):
- Cheroot
- Gunicorn
- Tornado
- OpenWrt uhttpd
- OpenBSD httpd

# Platform
```
Linux 399194a6dcd8 6.9.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 05 Jul 2024 22:11:24 +0000 x86_64 GNU/Linux
```
ServiceTalk main @ 8ae9e7bec4cff5a32fe6374c5b7d531507f2cbbe

Nadav0077:
#### Motivation
HTTP header field values should not contain illegal characters. Currently, validation was not enabled by default for performance reasons, but we reconsidered it in favor of safe behavior by default.

#### Modifications
- Enable headers values validation for `DefaultHttpHeadersFactory` and deprecated `H2HeadersFactory` by default.
- Provide a `io.servicetalk.http.api.temporaryDefaultValidateHeaderValues` system property to help users rollback pre-existing default behavior.
- Ensure null values are rejected before validation happens.
- Enhance testing.

#### Result
Invalid HTTP header values fail during insertion into `HttpHeaders` collection.

Fixes: #3022

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeaders.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeaders.java
@@ -575,7 +575,11 @@ final class DefaultHttpHeaders extends MultiMap<CharSequence, CharSequence> impl
     }
 
     @Override
-    protected CharSequence validateValue(final CharSequence value) {
+    protected CharSequence validateValue(final CharSequence name, @Nullable final CharSequence value) {
+        if (value == null) {
+            throw new IllegalArgumentException(
+                    "Null values are not allowed, encountered null for header name: " + name);
+        }
         if (validateValues) {
             validateHeaderValue(value);
         }
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeaders.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeaders.java
@@ -576,7 +576,7 @@ final class DefaultHttpHeaders extends MultiMap<CharSequence, CharSequence> impl
 
     @Override
     protected CharSequence validateValue(final CharSequence value) {
-        if (validateValues) {
+        if (value != null && validateValues) {
             validateHeaderValue(value);
         }
         return value;
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeaders.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeaders.java
@@ -576,7 +576,7 @@ final class DefaultHttpHeaders extends MultiMap<CharSequence, CharSequence> impl
 
     @Override
     protected CharSequence validateValue(final CharSequence value) {
-        if (value != null && validateValues) {
+        if (validateValues) {
             validateHeaderValue(value);
         }
         return value;
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeadersFactory.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeadersFactory.java
@@ -32,7 +32,8 @@ public final class DefaultHttpHeadersFactory implements HttpHeadersFactory {
             "io.servicetalk.http.api.temporaryDefaultValidateHeaderValues";
     private static final boolean DEFAULT_VALIDATE_VALUES = defaultValidateValues();
 
-    public static final HttpHeadersFactory INSTANCE = new DefaultHttpHeadersFactory(true, true, DEFAULT_VALIDATE_VALUES);
+    public static final HttpHeadersFactory INSTANCE =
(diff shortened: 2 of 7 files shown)


idelpivnitskiy:
Hi @Nadav0077,
Thank you for your contribution!

As your PR description currently says, users already have an option to enable header values validation via `DefaultHttpHeadersFactory` settings. The benefit of validating at that layer, compare to the encoder level, is that users get an exception right away, which makes it much easier to figure out which header value is invalid. If validation happens in the encoder, then the caller stack-trace is lost and users won't be able to see where the invalid header is coming from.

I understand the point that there could be other implementations of `HttpHeaders` container, but to our knowledge there are none outside this repo and if there are some, it might be responsibility of that implementation to enforce validation.

For our `DefaultHttpHeadersFactory` and `H2HeadersFactory` (deprecated), we debated for a long time if values validation should be enabled by default or not due to perf overhead (see #3022 as an examples). Recently we did some analysis and decided to enable it by default.

Would you like to rework your PR to enable `validateValues` by default instead of changing the encoder? Because it's a behavior change that may cause unexpected side-effects on our users, we typically provide a temporary system property to revert pre-existing behavior, you can see an example of our pattern here: https://github.com/apple/servicetalk/blob/b8a7631f475da7997b7e8eea4922da55ecfe2a38/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H1ProtocolConfigBuilder.java#L43-L55

Let me know what you think. Happy to discuss!

Nadav0077:
@idelpivnitskiy 
Thanks, that makes sense to me. I can rework this to enable validateValues by default in the built in header factories, add the temporary opt out property following the existing pattern, and drop the encoder level validation. What do you think?

idelpivnitskiy:
Yes please, will appreciate your help!
Please reuse the same system property name for `DefaultHeadersFactory` and deprecated `H2HeadersFactory`. They are in different modules, so you may need to duplicate logic. That's fine, we will clean it up when we remove deprecated classes

idelpivnitskiy:
@Nadav0077 we plan to cut a release in the 2nd half of this week, and would like to include the new default there. Would you be open to update the PR today-tomorrow or should we send another one? Thank you!

Nadav0077:
@idelpivnitskiy Hey, sorry, I have a busy week..
Pushed, can you review it please ?:)
Thanks

servicetalk-http-api/src/main/java/io/servicetalk/http/api/DefaultHttpHeaders.java
```diff
@@ -576,7 +576,7 @@ protected CharSequence validateKey(@Nullable final CharSequence name) {
 
     @Override
     protected CharSequence validateValue(final CharSequence value) {
-        if (validateValues) {
+        if (value != null && validateValues) {
```
idelpivnitskiy: Our API does not support null values, we expect NPE if users pass `null`:
```suggestion
        if (validateValues) {
```

idelpivnitskiy (approved):
Thank you!

PR #3529, opened 2026-06-23
dns-discovery-netty: clear min TTL state for removed SRV targets

Nadav0077:
#### Motivation
`MinTtlCache` keeps minimum TTL bookkeeping separately from the delegate DNS cache. When an SRV target is removed, `DefaultDnsClient` removes the corresponding A-record publisher, but the min-TTL bookkeeping for that hostname can remain behind.

That side state should follow the lifecycle of the per-target A-record publisher.

#### Modifications
- Add a local min-TTL cleanup method to `MinTtlCache`.
- Clear min-TTL state when a specific hostname is cleared.
- Clear min-TTL state when an SRV target is removed and its A-record publisher is cancelled.
- Add focused coverage for hostname-specific min-TTL cleanup.

#### Result
Removed SRV targets no longer leave stale min-TTL state behind, while the shared delegate DNS cache is not cleared for unrelated users of the same hostname.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
@@ -345,7 +345,9 @@ final class DefaultDnsClient implements DnsClient {
                     // Unwrap the list so we can use it in SrvInactiveCombinerOperator below.
                     return from(((SrvInactiveEvent<HostAndPort, InetSocketAddress>) srvEvent).aggregatedEvents);
                 } else {
-                    final ARecordPublisher aPublisher = aRecordMap.remove(srvEvent.address().hostName());
+                    final String srvHostName = srvEvent.address().hostName();
+                    final ARecordPublisher aPublisher = aRecordMap.remove(srvHostName);
+                    ttlCache.clearMinTtl(srvHostName);
                     if (aPublisher != null) {
                         aPublisher.cancelAndFail0(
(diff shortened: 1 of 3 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -345,8 +345,10 @@ public Publisher<Collection<ServiceDiscovererEvent<InetSocketAddress>>> dnsSrvQu
                     // Unwrap the list so we can use it in SrvInactiveCombinerOperator below.
                     return from(((SrvInactiveEvent<HostAndPort, InetSocketAddress>) srvEvent).aggregatedEvents);
                 } else {
-                    final ARecordPublisher aPublisher = aRecordMap.remove(srvEvent.address().hostName());
+                    final String srvHostName = srvEvent.address().hostName();
+                    final ARecordPublisher aPublisher = aRecordMap.remove(srvHostName);
                     if (aPublisher != null) {
+                        ttlCache.clearMinTtl(srvHostName);
```
bryce-anderson: Should this match the removal of the `ARecordPublisher`, eg be unconditional?

Nadav0077: Good point, updated so min-TTL state is cleared whenever we process the SRV target removal while publisher cancel remains conditional on a publisher being present.

bryce-anderson (approved):

idelpivnitskiy (approved):
Thank you!

PR #3531, opened 2026-06-23
Update PMD 6.55.0 -> 7.25.0

idelpivnitskiy:


servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
```diff
@@ -297,21 +297,23 @@ public <Req, Resp> BlockingStreamingClientCall<Req, Resp> newBlockingStreamingCa
...
                 throw GrpcStatusException.fromThrowable(cause);
             }
+            try {
+                extractResponseContext(response, metadata);
+                return validateResponseAndGetPayload(response.toStreamingResponse(), responseContentType,
+                        client.executionContext().bufferAllocator(), readGrpcMessageEncodingRaw(response.headers(),
+                                deserializerIdentity, deserializers, GrpcStreamingDeserializer::messageEncoding),
+                        httpRequest.requestTarget()).toIterable();
+            } catch (Throwable t) {
+                response.messageBody().iterator().close();
```
daschl, line 314: could this throw a NPE if response has not been set before?

daschl (approved):

servicetalk-grpc-api/src/main/java/io/servicetalk/grpc/api/DefaultGrpcClientCallFactory.java
idelpivnitskiy: no, if `client.request(...)` does not throw, it always returns a response object, its API contract is non-nullable

PR #3533, opened 2026-06-25
utils-internal: add OS detection (isLinux, isOsx, normalizedOs)

daschl:
#### Motivation
ServiceTalk has its own PlatformDependent class and so we should use it to detect the OS instead of falling back to Netty's.

#### Modifications
Adds OS detection to ServiceTalk's PlatformDependent, ported from io.netty.util.internal.PlatformDependent's.

Exposes isLinux(), isOsx(), and normalizedOs(); no isWindows() — ST has no Windows-specific branches. Migrates four call sites that previously reached into Netty's internal package, and adds a parity test in transport-netty-internal to catch silent drift on future Netty bumps.

### DIFF
--- a/servicetalk-traffic-resilience-http/build.gradle
+++ b/servicetalk-traffic-resilience-http/build.gradle
@@ -43,7 +43,7 @@ dependencies {
   testImplementation project(":servicetalk-concurrent-test-internal")
   testImplementation project(":servicetalk-http-netty")
   testImplementation project(":servicetalk-test-resources")
-  testImplementation "io.netty:netty-common"
+  testRuntimeOnly "io.netty:netty-common"
   testImplementation "org.junit.jupiter:junit-jupiter-api"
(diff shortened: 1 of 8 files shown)


servicetalk-traffic-resilience-http/build.gradle
```diff
@@ -43,7 +43,7 @@ dependencies {
   testImplementation project(":servicetalk-concurrent-test-internal")
   testImplementation project(":servicetalk-http-netty")
   testImplementation project(":servicetalk-test-resources")
-  testImplementation "io.netty:netty-common"
+  testRuntimeOnly "io.netty:netty-common"
```
bryce-anderson: Do we still need this at all?

bryce-anderson (approved):

servicetalk-traffic-resilience-http/build.gradle
daschl: nope 👍

PR #3546, opened 2026-07-01
http-security-jersey: fix fail-open BasicAuth Jersey filter with non-null anonymous principal

igavrysh:
#### Motivation 
`servicetalk-http-security-jersey` was flagged as fail-open: `@BasicAuthenticated` resources silently received unauthenticated requests when `BasicAuthHttpServiceFilter` was missing upstream (or when `AsyncContext` was disabled). The Jersey security-context
  filter had no reliable signal to tell an authenticated request apart from one that never went through Basic authentication, so it defaulted to letting the request through.

#### Modifications
  * `BasicAuthHttpServiceFilter` now publishes an `AUTHENTICATED` marker (`ContextMap.Key<Boolean>`) into `AsyncContext` on every successful authentication, independent of whether a userInfo key is configured.

  * `AbstractBasicAuthSecurityContextFilter.filter()` now calls `abortWith(401)` when the `AUTHENTICATED `marker is absent from `AsyncContext`, and emits a one-time SLF4J warning to surface the misconfiguration (missing upstream filter or disabled AsyncContext).

 * The anonymous-principal behavior is retained: when a request is authenticated upstream but no user info is published, `NoUserInfoBuilder`'s default installs a `BasicAuthSecurityContext` carrying `ANONYMOUS_PRINCIPAL` rather than proceeding with a null principal. The no-arg `forNameBinding()` / `forGlobalBinding()` builders therefore remain fully supported and non-deprecated.

*  Tests updated to assert the fail-closed contract while keeping the anonymous-principal expectations; new `BasicAuthFilterFailCloseTest` covers the end-to-end behavior:
   * missing upstream filter -> `401`
   * key mismatch -> proceeds with null principal
   * no-user-info default → anonymous principal

`slf4j-api` added to the base module and three Jakarta variants.

#### Result
`@BasicAuthenticated` resources now reliably fail closed with `401` whenever `AsyncContext` lacks the `AUTHENTICATED` flag, while an authenticated request with no published user info still carries the anonymous principal as before.

Behavioral change: deployments that relied on the previous fail-open behavior — using these filters without installing `BasicAuthHttpServiceFilter` upstream, or with `AsyncContext` disabled — will now receive 401 instead of reaching the resource. Install `BasicAuthHttpServiceFilter` upstream of the Jersey router and ensure `AsyncContext` is enabled.


servicetalk-http-security-jersey/src/main/java/io/servicetalk/http/security/auth/basic/jersey/AbstractBasicAuthSecurityContextFilter.java
```diff
@@ -38,10 +48,36 @@ abstract class AbstractBasicAuthSecurityContextFilter<UserInfo> implements Conta
 
     @Override
     public void filter(final ContainerRequestContext requestCtx) {
+        if (!Boolean.TRUE.equals(AsyncContext.get(BasicAuthHttpServiceFilter.AUTHENTICATED))) {
+            // The request was not authenticated by a BasicAuthHttpServiceFilter. Either that filter is not installed
+            // upstream of the Jersey router, or AsyncContext is disabled. Fail closed rather than exposing a
+            // @BasicAuthenticated resource to an unauthenticated caller.
+            if (NOT_AUTHENTICATED_WARNED.compareAndSet(false, true)) {
```
bryce-anderson: tbh, I think this probably should log each time. It's a very serious issue and it's going to fail all requests so we might as well be more noisy about it so it's easy to pinpoint.

servicetalk-http-security-jersey/src/main/java/io/servicetalk/http/security/auth/basic/jersey/AbstractBasicAuthSecurityContextFilter.java
```diff
@@ -38,10 +48,36 @@ abstract class AbstractBasicAuthSecurityContextFilter<UserInfo> implements Conta
...
+                        "and that AsyncContext is enabled. Further occurrences will not be logged.");
+            }
+            requestCtx.abortWith(Response.status(Response.Status.UNAUTHORIZED).build());
+            return;
+        }
+
         final SecurityContext securityContext = securityContext(requestCtx);
         if (securityContext != null) {
             requestCtx.setSecurityContext(securityContext);
+        } else if (userInfoKey != null && NO_USER_INFO_WARNED.compareAndSet(false, true)) {
```
bryce-anderson, line 64: This one seems reasonable to throttle since it's still letting the request through.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/auth/BasicAuthHttpServiceFilter.java
```diff
@@ -382,6 +396,10 @@ private Single<StreamingHttpResponse> onAuthenticated(final HttpServiceContext c
                                                               final StreamingHttpRequest request,
                                                               final StreamingHttpResponseFactory factory,
                                                               final UserInfo userInfo) {
+            // Always mark the request as authenticated so that downstream components can distinguish an
+            // authenticated request that carries no user info from a request that never went through Basic auth.
+            AsyncContext.put(AUTHENTICATED, Boolean.TRUE);
+            request.context().put(AUTHENTICATED, Boolean.TRUE);
```
bryce-anderson: I don't think we're getting anything from this today, right? If so I think we should just use the AsyncContext. I get there may be some users who would prefer to read the context entry but when that day comes I think can start adding it.

servicetalk-http-security-jersey/src/main/java/io/servicetalk/http/security/auth/basic/jersey/BasicAuthSecurityContextFilters.java
```diff
@@ -112,20 +112,27 @@ public abstract static class UserInfoBuilder<UserInfo> extends AbstractBuilder<U
...
 
         NoUserInfoBuilder() {
             super(BasicAuthSecurityContextFilters::asSecurityContextFunction,
-                    BasicAuthSecurityContextFilters::newAnonymousSecurityContext);
+                    requestCtx -> null);
         }
     }
 
-    // Visible for testing
-    static final Principal ANONYMOUS_PRINCIPAL = () -> "ANONYMOUS";
```
bryce-anderson, line 127: I'm starting to wonder if we should keep this. With the new context key we _guarantee_ that before someone can fetch this `ANONYMOUS_PRINCIPAL` from the SecurityContext they must have passed through the `BasicAuthHttpServiceFilter`. They are thus authenticated, and the uses of this are simply anonymous because we decided not to extract a meaningful identifier.
Keeping it would preserve all the behavior we currently have _except_ the possibility of accidentally forgetting to install the BasicAuthHttpServiceFilter.

I'm not strongly opinionated about it: if we want to drop this too I'm fine with it, although unless there is a strong case for coupling the two I think we could make it a separate commit so it feels more deliberate.

servicetalk-http-utils/src/main/java/io/servicetalk/http/utils/auth/BasicAuthHttpServiceFilter.java
igavrysh: removed

servicetalk-http-security-jersey/src/main/java/io/servicetalk/http/security/auth/basic/jersey/BasicAuthSecurityContextFilters.java
igavrysh: added back `ANONYMOUS_PRINCIPAL`

servicetalk-http-security-jersey/src/test/java/io/servicetalk/http/security/auth/basic/jersey/BasicAuthFilterFailCloseTest.java
```diff
@@ -0,0 +1,290 @@
...
+
+    /**
+     * When the filter is configured with a {@code userInfoKey} but
+     * {@code AsyncContext} contains no {@code UserInfo} (e.g. because
+     * {@code BasicAuthHttpServiceFilter} is not installed upstream), the filter
+     * must fail closed. The expected behavior is to call
+     * {@link ContainerRequestContext#abortWith(Response)} with a {@code 401}.
+     */
+    @Test
+    void nameBindingFilterAbortsWith401WhenUserInfoMissing() throws IOException {
```
bryce-anderson: I think this test is redundant with BasicAuthSecurityContextFiltersTest.notAuthenticatedAbortsWith401.

servicetalk-http-security-jersey/src/test/java/io/servicetalk/http/security/auth/basic/jersey/BasicAuthFilterFailCloseTest.java
```diff
@@ -0,0 +1,290 @@
...
+        assertThat(respCaptor.getValue().getStatus(),
+                is(equalTo(Response.Status.UNAUTHORIZED.getStatusCode())));
+    }
+
+    /**
+     * Same as above but for the global-binding variant — the fail-close path is
+     * shared via {@link AbstractBasicAuthSecurityContextFilter}.
+     */
+    @Test
+    void globalBindingFilterAbortsWith401WhenUserInfoMissing() throws IOException {
```
bryce-anderson: I think this test is redundant with BasicAuthSecurityContextFiltersTest.notAuthenticatedAbortsWith401.

servicetalk-http-security-jersey/src/test/java/io/servicetalk/http/security/auth/basic/jersey/BasicAuthFilterFailCloseTest.java
```diff
@@ -0,0 +1,290 @@
...
+                is(equalTo(Response.Status.UNAUTHORIZED.getStatusCode())));
+    }
+
+    /**
+     * The {@code NoUserInfoBuilder} default installs a {@link SecurityContext} whose principal is the
+     * {@code ANONYMOUS_PRINCIPAL}. When the request was authenticated upstream but no user info is published, the
+     * filter still associates the request with the anonymous principal rather than rejecting it.
+     */
+    @Test
+    void noUserInfoDefaultInstallsAnonymousPrincipalWhenAuthenticated() throws IOException {
```
bryce-anderson: I think this is redundant with authenticatedNoUserInfoInstallsAnonymousPrincipal

servicetalk-http-security-jersey/src/test/java/io/servicetalk/http/security/auth/basic/jersey/BasicAuthFilterFailCloseTest.java
```diff
@@ -0,0 +1,290 @@
...
+ *       {@link SecurityContext} whose principal is the anonymous principal, so an already-authenticated request is
+ *       never rejected for lacking user info.</li>
+ *   <li>When the request <em>was</em> authenticated but no user info is available under a configured key (for example
+ *       a key mismatch), the filter proceeds with a {@code null} user principal instead of rejecting an
+ *       already-authenticated request.</li>
+ * </ol>
+ */
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+class BasicAuthFilterFailCloseTest {
```
bryce-anderson: Presuming we remove the mock tests, I think this could be better named as `BasicAuthFilterMisconfigurationTest`.

bryce-anderson (commented):
The production changes look good, just a few minor nits in the testing. The comments are a bit verbose: maybe they could be condensed.

servicetalk-http-security-jersey/src/test/java/io/servicetalk/http/security/auth/basic/jersey/BasicAuthFilterFailCloseTest.java
igavrysh: renamed

servicetalk-http-security-jersey/src/main/java/io/servicetalk/http/security/auth/basic/jersey/AbstractBasicAuthSecurityContextFilter.java
igavrysh: added `NO_USER_INFO_WARNED.compareAndSet(false, true)`

bryce-anderson (approved):

igavrysh:
> The production changes look good, just a few minor nits in the testing. The comments are a bit verbose: maybe they could be condensed.

Compressed comments

idelpivnitskiy (approved):

PR #3547, opened 2026-07-02
Add the unpack method to TransportObservers

miroswan:
The unpack method was recently added to `HttpLifecycleObservers` to enable detection of a specific observer implementation within the observer tree created by `HttpLifecycleObservers.combine`. This change adds similar functionality to `TransportObservers`.

#### Motivation
This is useful when integrating metric observers into ServiceTalk services.

#### Modifications
The `unpack` method was added to `TransportObservers`

#### Result
The `unpack` method returns a `List<TransportObserver>` of all found observers
decorated from the root observer.


idelpivnitskiy (approved):
Thank you!

PR #3552, opened 2026-07-08
DnsServiceDiscovererObservers: add `combine` and `unpack` methods

miroswan:
#### Motivation

Adding `BiDnsServiceDiscovererObserver`,  `DnsServiceDiscovererObservers`, and `CatchAllDnsServiceDiscovererObserver` to enable safe composition of DNS observers. 

#### Modifications

The public API is the static utility class `DnsServiceDiscovererObservers`. Its implementation closely matches `TransportObservers` and `HttpLifecycleObservers`. 

#### Result

This API can help facilitate the integration of DNS metric observers.

#### Caveat

This PR does _not_ integrate `CatchAllDnsServiceDiscovererObserver` into the default DNS observer implementation. This work is reserved for a subsequent PR.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObservers.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObservers.java
@@ -38,7 +38,7 @@ public final class DnsServiceDiscovererObservers {
      * @return a safe version of the passed {@link DnsServiceDiscovererObserver} that catches and logs all exceptions,
      * but does not rethrow them.
      */
-    public static DnsServiceDiscovererObserver asSafeObserver(final DnsServiceDiscovererObserver observer) {
+    static DnsServiceDiscovererObserver asSafeObserver(final DnsServiceDiscovererObserver observer) {
         if (observer instanceof CatchAllDnsServiceDiscovererObserver) {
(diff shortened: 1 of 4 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DnsServiceDiscovererObservers.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Converts passed {@link DnsServiceDiscovererObserver} to a safe version that catches and logs all exceptions, but
+     * does not rethrow them.
+     *
+     * @param observer {@link DnsServiceDiscovererObserver} to convert
+     * @return a safe version of the passed {@link DnsServiceDiscovererObserver} that catches and logs all exceptions,
+     * but does not rethrow them.
+     */
+    public static DnsServiceDiscovererObserver asSafeObserver(final DnsServiceDiscovererObserver observer) {
```
idelpivnitskiy: I would propose we hide this method for now, until it's necessary to make it public. Most likely, we can just use it internally in a follow-up when we start using catch-all wrapper, and users don't have to deal with it themselves:
```suggestion
    static DnsServiceDiscovererObserver asSafeObserver(final DnsServiceDiscovererObserver observer) {
```

idelpivnitskiy (approved):
Thank you!

PR #3554, opened 2026-07-08
buffer-api: fix `getUnsignedShort()` mask

fallintoplace:
#### Motivation:
`AbstractBuffer.getUnsignedShort(int)` masked with `0xfff` instead of `0xffff`, so any short value with any of the upper 4 bits set (`0x1000..0xffff`) returned the wrong number — e.g. `0xffff` read as `4095` instead of `65535`.  

#### Modifications:
- `AbstractBuffer.getUnsignedShort`: mask with `0xffff`.
- Add `ReadOnlyByteBufferTest.getUnsignedShort` covering values that trip the old mask.

#### Result:
`Buffer.getUnsignedShort(int)` returns the correct unsigned value for the full 16-bit domain on read-only buffers.


idelpivnitskiy (approved):
Great find, thank you!

PR #3557, opened 2026-07-08
buffer-api: reject negative readable byte counts

fallintoplace:
#### Motivation

Reject negative readable byte counts before checking whether enough bytes are available. Without this guard, calls like `skipBytes(-1)` can pass the readable-bytes check and move `readerIndex` backward.

#### Modifications

- Enhanced `AbstractBuffer.checkReadableBytes0` to ensure `minimumReadableBytes` is not negative.
- Added a regression test that verifies `skipBytes(-1)` throws and leaves `readerIndex` unchanged.

#### Result

Users get an exception if they try to use negative values for readable bytes.


idelpivnitskiy (approved):
Thank you!

idelpivnitskiy:
@fallintoplace looks like this change broke `NettyCompressionStreamingSerializerTest` variants that use `DEFAULT_RO_ALLOCATOR`

idelpivnitskiy (approved):
Great findings!

PR #3558, opened 2026-07-08
buffer-api: fix zero-length `BufferInputStream` reads

fallintoplace:
#### Motivation

`BufferInputStream.read(byte[], off, len)` returned `-1` for zero-length reads when the backing buffer was already exhausted. `InputStream` expects zero-length reads to return `0` after validating the array and range arguments.

#### Modifications

- Validates the array/range before checking EOF, returns `0` for `len == 0`
- Adds regression coverage for zero-length EOF reads plus invalid arguments at EOF

The bounds check is kept local instead of using `Objects.checkFromIndexSize` because the module still targets Java 8 bytecode.

#### Result

`BufferInputStream` implementation behavior is aligned with `InputStream` javadoc.

### DIFF
--- a/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/BufferInputStream.java
+++ b/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/BufferInputStream.java
@@ -42,7 +42,7 @@ final class BufferInputStream extends InputStream {
     public int read(byte[] b, int off, int len) {
         requireNonNull(b);
         if ((off | len) < 0 || len > b.length - off) {
-            throw new IndexOutOfBoundsException();
+            throw new IndexOutOfBoundsException("off=" + off + ", len=" + len + ", b.length=" + b.length);
         }
         if (len == 0) {
             return 0;
--- a/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/BufferInputStream.java
+++ b/servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/BufferInputStream.java
@@ -40,6 +40,13 @@ final class BufferInputStream extends InputStream {
 
     @Override
     public int read(byte[] b, int off, int len) {
+        requireNonNull(b);
+        if ((off | len) < 0 || len > b.length - off) {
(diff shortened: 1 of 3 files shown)


servicetalk-buffer-api/src/main/java/io/servicetalk/buffer/api/BufferInputStream.java
```diff
@@ -40,6 +40,13 @@ public int read() {
 
     @Override
     public int read(byte[] b, int off, int len) {
+        requireNonNull(b);
+        if ((off | len) < 0 || len > b.length - off) {
+            throw new IndexOutOfBoundsException();
```
idelpivnitskiy:
```suggestion
            throw new IndexOutOfBoundsException("off=" + off + ", len=" + len + ", b.length=" + b.length);
```

servicetalk-buffer-api/src/test/java/io/servicetalk/buffer/api/BufferInputStreamTest.java
```diff
@@ -69,6 +70,27 @@ void skipMoreThanIntegerMax() throws Exception {
...
+    }
+
+    @Test
+    void readValidatesBoundsAtEnd() {
+        InputStream empty = new BufferInputStream(DEFAULT_RO_ALLOCATOR.fromAscii(""));
+        byte[] bytes = new byte[1];
+        assertThrows(IndexOutOfBoundsException.class, () -> empty.read(bytes, -1, 0));
+        assertThrows(IndexOutOfBoundsException.class, () -> empty.read(bytes, 0, -1));
+        assertThrows(IndexOutOfBoundsException.class, () -> empty.read(bytes, 1, 1));
+    }
```
idelpivnitskiy, line 97:
```suggestion
        assertThrows(IndexOutOfBoundsException.class, () -> empty.read(bytes, 0, Integer.MAX_VALUE));
        assertThrows(IndexOutOfBoundsException.class, () -> empty.read(bytes, Integer.MAX_VALUE, 1));
    }
```

idelpivnitskiy (approved):
Thank you!

PR #3560, opened 2026-07-08
buffer-api: fix read-only byte buffer slice offsets

### LINKED ISSUE #3559: Sliced `ReadOnlyByteBuffer` does not account for the slice offset
Reproducer, add in `CharSequencesTest`:

```java
    @Test
    public void slice() {
        Buffer buffer = DEFAULT_RO_ALLOCATOR.fromAscii("text42text");
        assertThat((char) buffer.getByte(0), is('t'));
        assertThat((char) buffer.slice(4, 2).getByte(0), is('4'));
    }
```

The 2nd assertion fails with `java.lang.AssertionError`: 
```
Expected: is "4"
     but: was "t"
```

This affects all Buffer(s) allocated using `ReadOnlyBufferAllocator` from buffer-api. Same test with `ServiceTalkBufferAllocator` (`BufferAllocators` in buffer-netty) works correctly. Consider adding a similar test to `ReadOnlyBufferTest` or `NettyBufferTest` to demonstrate that.

fallintoplace:
#### Motivation

`ReadOnlyByteBuffer` kept the original `ByteBuffer` position and limit, so sliced or partial wrapped buffers still read from the original absolute offset.

#### Modifications
- Normalizes read-only buffer storage around zero-based `ByteBuffer` slices while preserving byte order.
- Makes `toNioBuffer(index, length)` return a zero-position view for the requested sub-region.
- Enhance tests.

#### Result

Fixes #3559

#### Behavior change

`ReadOnlyBufferAllocators.wrap(byte[], int, int)` and `wrap(ByteBuffer)` now normalize the returned `Buffer`'s coordinates to start at index `0` (previously, `readerIndex()` reflected the source `ByteBuffer`'s absolute position). Callers reading via `readerIndex()`/`writerIndex()` are unaffected; callers using absolute indexing tied to the original `ByteBuffer` position must update their offsets.

### DIFF
--- a/servicetalk-buffer-api/src/test/java/io/servicetalk/buffer/api/ReadOnlyByteBufferTest.java
+++ b/servicetalk-buffer-api/src/test/java/io/servicetalk/buffer/api/ReadOnlyByteBufferTest.java
@@ -67,6 +67,39 @@ class ReadOnlyByteBufferTest {
     }
 
     @Test
+    void sliceStartsAtSliceOffset() {
+        Buffer buffer = DEFAULT_RO_ALLOCATOR.fromAscii("text42text");
+        Buffer slice = buffer.slice(4, 2);
+
+        assertEquals(0, slice.readerIndex());
+        assertEquals(2, slice.writerIndex());
+        assertEquals(2, slice.capacity());
+        assertEquals('4', (char) slice.getByte(0));
+        assertEquals('2', (char) slice.getByte(1));
+    }
+
+    @Test
+    void wrapPartialArrayStartsAtOffset() {
+        Buffer buffer = DEFAULT_RO_ALLOCATOR.wrap("text42text".getBytes(US_ASCII), 4, 2);
+
+        assertEquals(0, buffer.readerIndex());
+        assertEquals(2, buffer.writerIndex());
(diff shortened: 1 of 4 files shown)


servicetalk-buffer-api/src/test/java/io/servicetalk/buffer/api/ReadOnlyByteBufferTest.java
```diff
@@ -66,6 +66,39 @@ void getUnsignedShort() {
         assertEquals(0xf234, buffer1.getUnsignedShort(buffer1.readerIndex() + 2));
     }
 
+    @Test
+    void sliceStartsAtSliceOffset() {
+        Buffer buffer = DEFAULT_RO_ALLOCATOR.fromAscii("text42text");
```
idelpivnitskiy: It would be great to have a similar set of tests in `servicetalk-buffer-netty` module to ensure consistency in behavior with buffers allocated by `BufferAllocators`, not only `ReadOnlyBufferAllocators`

idelpivnitskiy (commented):
Thanks! Please add more tests for buffer-netty:

idelpivnitskiy (approved):
Thank you!

PR #3565, opened 2026-07-10
Use CatchAll observer within DefaultDnsClient to centralize exception handling

miroswan:
#### Motivation

`DefaultDnsClient` held a nullable `DnsServiceDiscovererObserver` and implemented exception safety around it. Every place that called into the observer had to null-check the observer and its returned discovery and resolution observers, and wrap each callback in its own try/catch to log and swallow exceptions. This spread the same defensive boilerplate across the class and duplicated logic that `CatchAllDnsServiceDiscovererObserver` now provides.

The transport side already solved this by defaulting to a Noop observer and wraps user observers with `asSafeObserver`, so call sites can treat the observer as always present and always safe. The DNS client did not previously follow that pattern.

#### Modifications

* Extracted the Noop observer implementations out of `CatchAllDnsServiceDiscovererObserver` into a new top-level `NoopDnsServiceDiscovererObserver` (with nested `NoopDnsDiscoveryObserver` and `NoopDnsResolutionObserver`), so they can be reused as defaults.
* Made `DefaultDnsClient.observer` non-null. It is now assigned via a `safeObserver(...)` helper that maps a null user observer to `NoopDnsServiceDiscovererObserver.INSTANCE` and otherwise wraps it with `DnsServiceDiscovererObservers.asSafeObserver(...)`. This was inspired what we do on the transport side.
* Removed the dead null checks and try/catch blocks from `DefaultDnsClient`.
* Dropped `@Nullable` from the internal discovery/resolution observer fields and parameters that can no longer be null.

#### Result

`DefaultDnsClient` now aligns with the transport side. The observer is always present and always exception-safe, so call sites no longer repeat null checks and try/catch logic.

No public API change. All touched classes are package-private, including the new `NoopDnsServiceDiscovererObserver`. The constructor still accepts a nullable observer.

No behavior change for users. A user observer that throws is still caught, logged, and swallowed, and the client proceeds without further callbacks. The only internal difference is that it proceeds with a Noop observer instead of a null one. The `beforeFinally` terminal handler is now always attached (driving the Noop when no observer is configured), which has negligible overhead and no functional impact.

### DIFF
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/CatchAllDnsServiceDiscovererObserver.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/CatchAllDnsServiceDiscovererObserver.java
@@ -15,16 +15,14 @@
  */
 package io.servicetalk.dns.discovery.netty;
 
-import io.servicetalk.dns.discovery.netty.DnsServiceDiscovererObserver.DnsDiscoveryObserver;
-import io.servicetalk.dns.discovery.netty.DnsServiceDiscovererObserver.DnsResolutionObserver;
-import io.servicetalk.dns.discovery.netty.DnsServiceDiscovererObserver.ResolutionResult;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.function.Supplier;
 import java.util.function.UnaryOperator;
 
+import static io.servicetalk.dns.discovery.netty.NoopDnsServiceDiscovererObserver.NoopDnsDiscoveryObserver;
+import static io.servicetalk.dns.discovery.netty.NoopDnsServiceDiscovererObserver.NoopDnsResolutionObserver;
 import static io.servicetalk.utils.internal.ThrowableUtils.addSuppressed;
 import static java.util.Objects.requireNonNull;
 
@@ -99,49 +97,6 @@ final class CatchAllDnsServiceDiscovererObserver implements DnsServiceDiscoverer
         @Override
         public void resolutionCompleted(final ResolutionResult result) {
             safeReport(() -> observer.resolutionCompleted(result), observer, "resolution completed");
-        }
-    }
-
-    private static final class NoopDnsDiscoveryObserver implements DnsDiscoveryObserver {
-
-        static final DnsDiscoveryObserver INSTANCE = new NoopDnsDiscoveryObserver();
-
-        private NoopDnsDiscoveryObserver() {
-            // Singleton
-        }
-
-        @Override
-        public DnsResolutionObserver onNewResolution(final String name) {
-            return NoopDnsResolutionObserver.INSTANCE;
-        }
-
-        @Override
-        public void discoveryCancelled() {
-            // noop
-        }
-
-        @Override
-        public void discoveryFailed(final Throwable cause) {
-            // noop
-        }
-    }
-
-    private static final class NoopDnsResolutionObserver implements DnsResolutionObserver {
-
-        static final DnsResolutionObserver INSTANCE = new NoopDnsResolutionObserver();
-
-        private NoopDnsResolutionObserver() {
-            // Singleton
-        }
-
-        @Override
-        public void resolutionFailed(final Throwable cause) {
-            // noop
-        }
-
-        @Override
-        public void resolutionCompleted(final ResolutionResult result) {
-            // noop
         }
     }
 
--- a/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/CatchAllDnsServiceDiscovererObserver.java
+++ b/servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/CatchAllDnsServiceDiscovererObserver.java
@@ -15,14 +15,16 @@
  */
 package io.servicetalk.dns.discovery.netty;
 
+import io.servicetalk.dns.discovery.netty.NoopDnsServiceDiscovererObserver.NoopDnsDiscoveryObserver;
+import io.servicetalk.dns.discovery.netty.NoopDnsServiceDiscovererObserver.NoopDnsResolutionObserver;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.function.Supplier;
 import java.util.function.UnaryOperator;
+import javax.annotation.Nullable;
 
-import static io.servicetalk.dns.discovery.netty.NoopDnsServiceDiscovererObserver.NoopDnsDiscoveryObserver;
-import static io.servicetalk.dns.discovery.netty.NoopDnsServiceDiscovererObserver.NoopDnsResolutionObserver;
 import static io.servicetalk.utils.internal.ThrowableUtils.addSuppressed;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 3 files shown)


servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -268,31 +260,20 @@ public Publisher<Collection<ServiceDiscovererEvent<InetAddress>>> dnsQuery(final
             ARecordPublisher pub = new ARecordPublisher(address, discoveryObserver);
             Publisher<? extends Collection<ServiceDiscovererEvent<InetAddress>>> events =
                     recoverWithInactiveEvents(pub, false, nxInvalidation);
-            return discoveryObserver == null ? events : events.beforeFinally(new TerminalSignalConsumer() {
+            return events.beforeFinally(new TerminalSignalConsumer() {
```
bryce-anderson: Now we unconditionally add the `beforeFinally` even if it's going to be events to nowhere. Could we change the equality check to look for the noop observer singleton and skip adding callbacks if it's noop?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -357,31 +338,20 @@ public Publisher<Collection<ServiceDiscovererEvent<InetSocketAddress>>> dnsSrvQu
             }, srvConcurrency)
             .liftSync(SrvInactiveCombinerOperator.EMIT);
 
-            return discoveryObserver == null ? events : events.beforeFinally(new TerminalSignalConsumer() {
```
bryce-anderson, line 360: Same question re equality check for the noop instance.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
bryce-anderson: I I think this whole block can how be extracted and reused into something like
```
 private static <T> Publisher<T> observeEvents(final DnsDiscoveryObserver discoveryObserver,
                                                  final Publisher<T> events) {
        return discoveryObserver == NoopDnsServiceDiscovererObserver.NoopDnsDiscoveryObserver.INSTANCE ? events :
                events.beforeFinally(new TerminalSignalConsumer() {
                    @Override
                    public void onComplete() {
                        // this event will never be triggered
                    }

                    @Override
                    public void onError(final Throwable cause) {
                        discoveryObserver.discoveryFailed(cause);
                    }

                    @Override
                    public void cancel() {
                        discoveryObserver.discoveryCancelled();
                    }
                });
    }
```

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -874,8 +832,8 @@ private void handleResolveDone0(final Future<DnsAnswer<T>> addressFuture,
                         ttlNanos = maxTTLNanos;
                     }
                     final List<ServiceDiscovererEvent<T>> events = calculateDifference(activeAddresses, addresses,
-                            comparator(), resolutionObserver == null ? null : (nAvailable, nMissing) ->
-                                    reportResolutionResult(resolutionObserver, new DefaultResolutionResult(
+                            comparator(), (nAvailable, nMissing) ->
+                                    resolutionObserver.resolutionCompleted(new DefaultResolutionResult(
```
bryce-anderson: Same thing re checking for the noop instance to elide the rest of the work.

bryce-anderson (commented):
Overall it looks good, just a few small suggestions. I'm not strongly opinionated about the equality checks, maybe it's better to just let them go.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
miroswan: Yeah, that's reasonable. I thought the performance implications here were negligible, justifying a simpler implementation.  But swapping the ternary null check for equality on Noop is reasonable. It's guaranteed to not change the performance if we take your approach. 

Changes incoming.

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/CatchAllDnsServiceDiscovererObserver.java
```diff
@@ -47,129 +47,131 @@ DnsServiceDiscovererObserver observer() {
...
 
     private static void safeReport(final Runnable runnable, final Object observer, final String eventName,
+                                   @Nullable final String serviceDiscovererId, final String targetName,
                                    final Throwable original) {
         try {
             runnable.run();
         } catch (Throwable unexpected) {
             addSuppressed(unexpected, original);
-            LOGGER.warn("Unexpected exception from {} while reporting a {} event", observer, eventName, unexpected);
+            LOGGER.warn("Unexpected exception from {} while reporting a {} event{}",
```
bryce-anderson: It feels like the formatting is off. From my reading of this it appears the expanded string will be
`"Unexpected exception from {observer} while reporting a {eventName} event{" (unique name)}"`. I think we at least want a space in there (I realize it's in the construction of the value but that is weird), and maybe some descriptor: right now it almost reads as the event description. Maybe it would be better as a prefix, eg `"{idName}: Unexpected exception from {} while reporting a {} event"`

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/CatchAllDnsServiceDiscovererObserver.java
```diff
@@ -47,129 +47,131 @@ DnsServiceDiscovererObserver observer() {
...
+    private static void safeReport(final Runnable runnable, final Object observer, final String eventName,
+                                   @Nullable final String serviceDiscovererId, final String targetName,
+                                   final Object result) {
+        try {
+            runnable.run();
+        } catch (Throwable unexpected) {
+            LOGGER.warn("Unexpected exception from {} while reporting a {} event{}: {}",
+                    observer, eventName, idName(serviceDiscovererId, targetName), result, unexpected);
+        }
+    }
```
bryce-anderson: This overload is dangerously close in type signature to the preexisting method with the same name - can we rename it so we don't have to worry about an exception being passed in here?

servicetalk-dns-discovery-netty/src/main/java/io/servicetalk/dns/discovery/netty/DefaultDnsClient.java
```diff
@@ -246,55 +246,46 @@ MinTtlCache ttlCache() {
...
-        return defer(() -> {
-            final DnsDiscoveryObserver discoveryObserver = newDiscoveryObserver(address);
-            ARecordPublisher pub = new ARecordPublisher(address, discoveryObserver);
-            Publisher<? extends Collection<ServiceDiscovererEvent<InetAddress>>> events =
-                    recoverWithInactiveEvents(pub, false, nxInvalidation);
-            return discoveryObserver == null ? events : events.beforeFinally(new TerminalSignalConsumer() {
+    private static <T> Publisher<T> observeDiscovery(final DnsDiscoveryObserver discoveryObserver,
+                                                     final Publisher<T> events) {
+        // If the observer is noop there is nothing to report, so avoid wrapping with a beforeFinally that would only
+        // deliver events to nowhere.
```
bryce-anderson: This comment is probably unnecessary: it's reasonably obvious what the intent is.

bryce-anderson (approved):

PR #3570, opened 2026-07-15
opentelemetry-http: stabilize lifecycle cancellation test

### LINKED ISSUE #3561: Flaky OpenTelemetryHttpRequesterFilterTest
https://github.com/apple/servicetalk/actions/runs/28782507944/job/85341265611?pr=3549
```
OpenTelemetryHttpRequesterFilterTest > transportObserver(ResultType, boolean) > transportObserver(ResultType, boolean) [5]: resultType=CANCEL, useHttp2=true FAILED
    java.lang.AssertionError: [span [GET] attribute keys] 
    Expecting UnmodifiableSet:
      [server.port,
        url.full,
        onResponseCancel,
        error.type,
        server.address,
        peer.service,
        network.protocol.version,
        onNewExchange,
        onRequest,
        onRequestComplete,
        http.request.method]
    to contain:
      [onExchangeFinally]
    but could not find the following element(s):
      [onExchangeFinally]
        at io.opentelemetry.sdk.testing.assertj.SpanDataAssert.hasAttribute(SpanDataAssert.java:285)
        at io.opentelemetry.sdk.testing.assertj.SpanDataAssert.hasAttribute(SpanDataAssert.java:275)
        at io.servicetalk.opentelemetry.http.OpenTelemetryHttpRequesterFilterTest.lambda$null$14(OpenTelemetryHttpRequesterFilterTest.java:470)
        at io.opentelemetry.sdk.testing.assertj.TraceAssert.hasSpansSatisfyingExactly(TraceAssert.java:65)
        at io.opentelemetry.sdk.testing.assertj.TraceAssert.hasSpansSatisfyingExactly(TraceAssert.java:49)
        at io.servicetalk.opentelemetry.http.OpenTelemetryHttpRequesterFilterTest.lambda$transportObserver$15(OpenTelemetryHttpRequesterFilterTest.java:466)
        at io.opentelemetry.sdk.testing.assertj.TracesAssert.hasTracesSatisfyingExactly(TracesAssert.java:108)
        at io.opentelemetry.sdk.testing.assertj.TracesAssert.hasTracesSatisfyingExactly(TracesAssert.java:93)
        at io.servicetalk.opentelemetry.http.OpenTelemetryHttpRequesterFilterTest.transportObserver(OpenTelemetryHttpRequesterFilterTest.java:465)
```

arnabnandy7:
#### Motivation

`OpenTelemetryHttpRequesterFilterTest.transportObserver` was flaky when cancelling an HTTP/2 request. The test assumed that `onExchangeFinally()` always runs before the OpenTelemetry span is exported. During cancellation, request termination may complete asynchronously, allowing the span to end before the final lifecycle callback adds its test attribute.

Fixes #3561.

#### Modifications

- Added a latch to `TestHttpLifecycleObserver` to observe `onExchangeFinally()` deterministically.
- Updated the cancellation test to await the final lifecycle callback.
- Removed the assertion that the final callback’s attribute must appear on an already-ended span.
- Preserved validation of the cancellation attribute and the callback’s active span context.

#### Result

The cancellation test no longer depends on timing or span mutability after export. All `transportObserver` HTTP/1.1 and HTTP/2 combinations pass.

There are no production behavior, API, or compatibility changes; only test code is affected.

### DIFF
--- a/servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequesterFilterTest.java
+++ b/servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequesterFilterTest.java
@@ -84,6 +84,7 @@ import static io.servicetalk.opentelemetry.http.TestUtils.clientBuilder;
 import static io.servicetalk.opentelemetry.http.TestUtils.sleep;
 import static io.servicetalk.transport.netty.internal.AddressUtils.serverHostAndPort;
 import static java.util.Collections.singletonList;
+import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
@@ -326,6 +327,7 @@ class OpenTelemetryHttpRequesterFilterTest {
         final boolean transportFailure = resultType == ResultType.TRANSPORT_ERROR;
         OpenTelemetry openTelemetry = otelTesting.getOpenTelemetry();
         BlockingQueue<Error> errors = new LinkedBlockingQueue<>();
+        TestHttpLifecycleObserver lifecycleObserver = new TestHttpLifecycleObserver(errors);
         TransportObserver transportObserver = new TransportObserver() {
 
             final AtomicReference<Span> span = new AtomicReference<>();
@@ -418,7 +420,7 @@ class OpenTelemetryHttpRequesterFilterTest {
                         .build())
                 .appendClientFilter(new TestTracingClientLoggerFilter(TRACING_TEST_LOG_LINE_PREFIX))
                 .appendClientFilter(new HttpLifecycleObserverRequesterFilter(
-                        new TestHttpLifecycleObserver(errors)))
+                        lifecycleObserver))
                 .appendConnectionFactoryFilter(
                         new TransportObserverConnectionFactoryFilter<>(transportObserver)).build()) {
            // This is necessary to let the load balancer become ready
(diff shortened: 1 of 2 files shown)


bryce-anderson:
I imagine this fixes the flaky test, but I'm a bit concerned about why it's flaky in the first place. Eg, are we papering over a real problem?

arnabnandy7:
@bryce-anderson Thanks for raising this. I traced the cancellation path to make sure the change is not hiding a production issue.

`HttpLifecycleObserver.onExchangeFinally()` is invoked only after both the request and response sides terminate. When an HTTP/2 request is cancelled, the response side can terminate first, causing the OpenTelemetry client span to end and be exported while the request side is still completing asynchronously. Once the request side terminates, `onExchangeFinally()` runs with the expected span context, but `setAttribute()` has no effect because that span has already ended. Whether the attribute appears in the exported span therefore depends on the relative timing of those two events.

This is consistent with the requester filter documentation, which notes that callbacks executing after the tracing filter closes its scope will not be able to modify the span.

The change still verifies that:

- `onExchangeFinally()` is invoked, using a bounded deterministic wait.
- The callback sees the same valid span context, through the existing identity/error checks.
- `onResponseCancel()` executes early enough to add its attribute to the exported span.

It only removes the assumption that `onExchangeFinally()` can always mutate the span after cancellation. No production behavior is changed. I’ll also add a comment in the cancellation test explaining why its final callback is verified separately from the exported span attributes.

bryce-anderson:
Sorry for the delay, I haven't forgotten about you @arnabnandy7. I'll take another look this week. The thing I want to check whether it's okay that the span has already ended before the `onExchangeFinally()`. OnExchangeFinally seems like the ideal place to end the span, but it's been a bit and my guess is there are "reasons" (maybe good, maybe bad 😅).

servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequesterFilterTest.java
```diff
@@ -461,13 +463,13 @@ public void connectionClosed() {
                     Future<HttpResponse> result = client.request(client.get("/slow")).toFuture();
                     TestUtils.sleep(20);
                     result.cancel(true);
+                    assertTrue(lifecycleObserver.awaitExchangeFinally(10, SECONDS));
```
bryce-anderson, line 468: Probably worth a note that `onResponseFinally()` isn't guaranteed to fire before the span completes because it measures the totality of the dispatch while the span only measures request sent to response received. Even in this case with an empty body we have some async operations to track that the empty body has fully terminated that can race with the response completion.

bryce-anderson (approved):
After digging through this it looks good to me. I'd like to add a comment so it's more clear in the future since it _feels_ right that `onExchangeFinally()` should trigger before span completion, even though thats wrong. Alternatively, maybe testing that onExchangeFinally can modify the span is wrong since it's not generally true: I'd also be open to removing the assertion altogether and documenting why that's the case.

servicetalk-opentelemetry-http/src/test/java/io/servicetalk/opentelemetry/http/OpenTelemetryHttpRequesterFilterTest.java
arnabnandy7: Added the necessary documentation as suggested.

bryce-anderson (approved):

bryce-anderson:
Thank you @arnabnandy7!

arnabnandy7:
@bryce-anderson I might look into rest of the flaky issues as well, when time permits.

PR #3578, opened 2026-07-20
Align how we log all system property values

idelpivnitskiy:
Motivation:

Align all props to use consistent pattern.

Modifications:

- Use `-D{}={}` pattern instead of `-D{}: {}` to help systems with key-value log layout to show properties correctly as a single line.
- Add standard `DANGEROUS_CONFIG_WARNING` prefix where necessary.
- Update formatting and wording to keep things consistent.

Result:

System properties are logged in consistent format and easier to parse for key-value logging systems.

### DIFF
--- a/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CloseHandler.java
+++ b/servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CloseHandler.java
@@ -55,7 +55,7 @@ public abstract class CloseHandler {
                 LOGGER.info("Error while parsing {}", logLevelPropertyStr, cause);
             }
         }
-        LOGGER.debug("{}={}", logLevelPropertyStr, logLevel);
+        LOGGER.debug("-D{}={}", logLevelPropertyStr, logLevel);
         CLOSE_HANDLER_LOG_LEVEL = logLevel;
     }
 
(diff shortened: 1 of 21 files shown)


servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CloseHandler.java
```diff
@@ -55,7 +55,7 @@ public abstract class CloseHandler {
                 LOGGER.info("Error while parsing {}", logLevelPropertyStr, cause);
             }
         }
-        LOGGER.debug("{}={}", logLevelPropertyStr, logLevel);
+        LOGGER.debug("-D{}={}", logLevelPropertyStr, logLevel);
```
bryce-anderson, line 58: Feel free to ignore: Not something you did in this PR, but in the throw case, `logLevel` will be null. While you're here do you want to change it to something else or is null obvious enough?

bryce-anderson (approved):

servicetalk-transport-netty-internal/src/main/java/io/servicetalk/transport/netty/internal/CloseHandler.java
idelpivnitskiy: Should be ok as-is, this is a super rate feature that is likely used only by us

PR #3586, opened 2026-07-28
Update Netty 4.1.136 -> 4.2.16

idelpivnitskiy:
#### Motivation
We've been running all tests with 4.2.x snapshots for a long time, we are ready to adopt 4.2.x.

#### Modifications
- Update Netty 4.1.136 -> 4.2.16 and regenerate lockfiles.
- Update transport-netty-internal module to use new Netty 4.2.x event loop API.
- Add io_uring as transitive dependency for all our users.
- Update `ci-netty-snapshot.yml` workflows to use Netty snapshots only to run tests, while the source code is always compiled with version defined in `gradle.properties`. This can be tested locally with `ORG_GRADLE_PROJECT_testNettyVersion=4.1.136.Final ./gradlew test`.
- Adjust `BufferAllocatorsTest`  to be resilient to 4.1 and 4.2 differences.
- Remove test dependency on `netty-codec` from `http-api`.
- Added PRB pipeline with Netty 4.1.x to always ensure compatibility.

#### Result
Compile and ship with Netty 4.2.x, while remain compatible with 4.1.x at runtime.


servicetalk-buffer-netty/src/test/java/io/servicetalk/buffer/netty/BufferAllocatorsTest.java
```diff
@@ -241,10 +242,9 @@ void testFromSequenceHeap(BufferAllocator allocator) {
     }
 
     private void assertBuffer(BufferAllocator allocator, Buffer buffer) {
-        // JDK24+ disables unsafe by default, which means allocators like PREFER_DIRECT will also fall back to
-        // heap allocations if unsafe is not explicitly enabled via a JVM arg.
-        final boolean direct = allocator != PREFER_HEAP_ALLOCATOR && PlatformDependent.hasUnsafe();
-        assertBuffer(buffer, direct);
+        // Try allocating a ByteBuf to check if the underlying allocator uses direct or heap memory:
+        assertThat(allocator, is(instanceOf(ByteBufAllocator.class)));
+        assertBuffer(buffer, ((ByteBufAllocator) allocator).buffer().isDirect());
```
bryce-anderson, line 247: Does this allocated buffer need to be released?

.github/workflows/ci-netty-snapshot.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     name: Netty ${{ matrix.netty }} Snapshot
     runs-on: ${{ matrix.os }}
     env:
-      ORG_GRADLE_PROJECT_nettyVersion: ${{ matrix.netty }}
+      ORG_GRADLE_PROJECT_testNettyVersion: ${{ matrix.netty }}
```
bryce-anderson, line 21: I know you're just inverting this behavior that already exists, but should we run the "other" netty as part of our PR CI matrix? As 4.2 and 4.1 naturally begin diverge the risk of breaking compatibility with 4.1.x may start to get significant.

bryce-anderson (approved):

chrisvest (approved):

servicetalk-buffer-netty/src/test/java/io/servicetalk/buffer/netty/BufferAllocatorsTest.java
chrisvest: `ServiceTalkBufferAllocator` always wrap its buffers in `Unreleasable*` implementations.

.github/workflows/ci-netty-snapshot.yml
idelpivnitskiy: I like the idea, added in https://github.com/apple/servicetalk/pull/3586/commits/4ff8bebfeb60506109a84dbdc0b232902831461e

bryce-anderson (approved):

PR #3587, opened 2026-07-28
Fix int overflow when sizing `FromInputStreamPublisher` buffer

idelpivnitskiy:
Motivation

`FromInputStreamPublisher` sized its read buffer with `min(available + 1, readChunkSize)`, which can overflow and result in throwing `NegativeArraySizeException`.

Modifications

- Use `addWithOverflowProtection(available, 1)`.
- Add a test that reproduces the issue.

Result

Such streams no longer fail with `NegativeArraySizeException`.

### DIFF
--- a/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
+++ b/servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
@@ -202,7 +202,7 @@ final class FromInputStreamPublisher<T> extends SubscribablePublisher<T> {
             final int readChunkSize = mapper.maxBufferSize();
             final byte[] buffer;
             if (readByte >= 0) {
-                buffer = new byte[min(available + 1, readChunkSize)];
+                buffer = new byte[min(addWithOverflowProtection(available, 1), readChunkSize)];
                 buffer[writeIdx++] = (byte) readByte;
             } else {
                 buffer = new byte[min(available, readChunkSize)];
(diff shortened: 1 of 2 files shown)


servicetalk-concurrent-api/src/main/java/io/servicetalk/concurrent/api/FromInputStreamPublisher.java
```diff
@@ -202,7 +202,7 @@ private int readAvailableAndEmit(final int available, final int readByte) throws
             final int readChunkSize = mapper.maxBufferSize();
             final byte[] buffer;
             if (readByte >= 0) {
-                buffer = new byte[min(available + 1, readChunkSize)];
+                buffer = new byte[min(addWithOverflowProtection(available, 1), readChunkSize)];
```
prakash-sandeep, line 205: In the regular case, the value of `available` into the `fillBuffer` call below, tracks the original value of the available value because the first byte is being put into the allocated buffer. When we clamp it here, shouldn't we model the same behavior? The problem is `available` is immutable here! My concern is if there is a potential for an off-by-one scenario later in the pipeline?

idelpivnitskiy: `available` isn't a capacity for `fillBuffer`, it's a "non-blocking" budget on the way in, and on the way out it's the unconsumed remainder. Writes are bounded separately by `min(buffer.length - writeIdx, available)`, and the loop exits on `writeIdx == buffer.length`, so a larger available just yields a positive remainder. That remainder is only checked against `END_OF_FILE`, and the next iteration re-queries `stream.available()`, so nothing carries forward. The else branch already passes unclamped available with a clamped buffer, so both are consistent now. 

If you have a sequence in mind that you think would break, send it over and I'll add it as a test case.

prakash-sandeep: Thanks the clarity. This makes sense now. I don't have a specific sequence in mind.

prakash-sandeep (approved):

PR #3599, opened 2026-08-07
http-netty: validate the HTTP/2 ':path' pseudo-header

daschl:
#### Motivation
Nothing validated ':path' characters, so a peer could send CR, LF, NUL or SP in it and we would hand it to the service and re-emit it on an HTTP/1.1 downgrade - the header-injection route RFC 9113, 8.2.1 calls out as request smuggling. Outbound we sent an empty ':path' where 8.3.1 requires "/", and HPACK silently mangled any non-ASCII target.

#### Modifications:
- Validate ':path' in both duplex handlers: non-empty, VCHAR only (0x21-0x7e), per RFC 9113, 8.3.1.
- Inbound violations become a PROTOCOL_ERROR stream error; outbound ones fail the write before the value reaches HPACK.
- Send "/" for an empty request-target, as HTTP/1.x already does.

#### Result:
An illegal ':path' no longer crosses the HTTP/2 boundary in either direction, and an empty request-target is sent as "/" on both protocols.

A raw non-ASCII target now fails instead of being silently corrupted; use path(String) or requestTarget(String, Charset) to percent-encode, which the request-target API already requires. HTTP/1.x is unchanged.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1Utils.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1Utils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Apple Inc. and the ServiceTalk project authors
+ * Copyright © 2019, 2026 Apple Inc. and the ServiceTalk project authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@ import org.slf4j.LoggerFactory;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import javax.annotation.Nullable;
 
 import static io.netty.handler.codec.http.HttpHeaderNames.TE;
 import static io.netty.handler.codec.http.HttpHeaderValues.TRAILERS;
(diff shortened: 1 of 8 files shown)


bryce-anderson (approved):
Looks good other than some of the comments are a bit excessive.

servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/H2ToStH1Utils.java
```diff
@@ -71,6 +72,51 @@ static void h2HeadersSanitizeForH1(Http2Headers h2Headers) {
...
+        final int len = path.length();
+        if (len == 0) {
+            // CONNECT omits :path entirely and never reaches here. An outbound request with no path component sends
+            // "/" instead of an empty value, so only a peer can trigger this.
+            return "':path' must not be empty";
+        }
+        // Not a table indexed by c & 0xff: that would alias a char above one octet onto a legal one (U+0161 -> 'a').
+        for (int i = 0; i < len; ++i) {
+            final char c = path.charAt(i);
+            if (c < 0x21 || c > 0x7e) {
```
idelpivnitskiy: Consider reusing `HttpObjectDecoder.isVCHAR` static utility

idelpivnitskiy (approved):

PR #3613, opened 2026-08-17
Deflake processor GC tests

### LINKED ISSUE #3392: SingleProcessorTest failure
https://github.com/apple/servicetalk/actions/runs/20862371073/job/59945016811?pr=3389

```
> Task :servicetalk-concurrent-api:test

SingleProcessorTest > synchronousCancelStillAllowsForGC() FAILED
    org.opentest4j.AssertionFailedError: expected: <java.lang.ref.WeakReference@59d35978> but was: <null>
        at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
        at app//org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
        at app//org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1145)
        at app//io.servicetalk.concurrent.api.SingleProcessorTest.synchronousCancelStillAllowsForGC(SingleProcessorTest.java:198)
```

n200534:
## Summary

- replace the processor tests' one-shot GC timing assumption with bounded reference-queue retries
- keep the processor strongly reachable while waiting, so the test still detects subscriber retention
- share the GC assertion between `SingleProcessorTest` and `CompletableProcessorTest`

## Root cause

The tests assumed that one best-effort `System.gc()` request and reference-queue processing would complete within roughly 400 ms. GC and weak-reference enqueueing have no such timing guarantee, so `ReferenceQueue.remove(100)` could time out even when the processor no longer retained the subscriber.

## Verification

- targeted `SingleProcessorTest.synchronousCancelStillAllowsForGC`: passed
- forced stress run: 50/50 passed
- clean, uncached `servicetalk-concurrent-api` suite: 3,293 tests, 9 skipped, 0 failures, 0 errors
- Checkstyle, SpotBugs, and PMD: passed

Fixes #3392

### DIFF
--- a/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/GcTestUtils.java
+++ b/servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/GcTestUtils.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright © 2026 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.concurrent.api;
+
+import java.lang.ref.Reference;
+import java.lang.ref.ReferenceQueue;
+import java.lang.ref.WeakReference;
+
+import static org.junit.jupiter.api.Assertions.assertSame;
+
+final class GcTestUtils {
+    private static final int MAX_GC_ATTEMPTS = 10;
+    private static final long QUEUE_WAIT_MILLIS = 1_000;
+
+    private GcTestUtils() {
+        // No instances.
+    }
+
+    static <T> void assertEventuallyEnqueued(Object keepAlive, WeakReference<T> expected,
+                                             ReferenceQueue<T> queue) throws InterruptedException {
+        Reference<? extends T> actual = null;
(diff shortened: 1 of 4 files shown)


servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/GcTestUtils.java
```diff
@@ -0,0 +1,46 @@
...
+    private GcTestUtils() {
+        // No instances.
+    }
+
+    static <T> void assertEventuallyEnqueued(Object keepAlive, WeakReference<T> expected,
+                                             ReferenceQueue<T> queue) throws InterruptedException {
+        Reference<? extends T> actual = null;
+        // Keep the owner alive while collecting the referent. Otherwise, collecting the owner and all its fields could
+        // make this assertion pass even if the owner still retains the referent.
+        synchronized (keepAlive) {
```
bryce-anderson: If the JIT decides to elide the lock, do we lose our keep-alive behavior? We could probably achieve the same thing by including it in the `assertSame` message via something like `+ "parent: " + keepAlive`.

It's too bad that Reference.reachabilityFence didn't come into existence until java9. 🫤

bryce-anderson (commented):
Thank you for the PR! Just one question.

servicetalk-concurrent-api/src/test/java/io/servicetalk/concurrent/api/GcTestUtils.java
n200534: Yes, that’s a valid concern. If the JIT elides the lock, we should not rely on the synchronized block alone to keep the parent reachable. Since Reference.reachabilityFence is unavailable on Java 8, I removed the synchronization and added keepAlive to the eagerly evaluated assertSame message. This creates an explicit use after the GC loop and keeps the parent reachable until the assertion. The tests pass with this change.

bryce-anderson (approved):
Thank you again @n200534!

PR #3622, opened 2026-08-28
Fix flaky H2ResponseCancelTest

arnabnandy7:
#### Motivation

`H2ResponseCancelTest.testServerClosesStreamNotConnection()` is flaky because closing an HTTP/2 stream can produce either an `H2StreamResetException` or a `ClosedChannelException`, depending on whether the reset frame or the closed child channel is observed first.

Closes #3425.

#### Modifications

- Accept either `H2StreamResetException` or `ClosedChannelException`.
- Document why both outcomes are valid.

#### Result

The test tolerates both valid race outcomes while continuing to verify that the parent HTTP/2 connection remains open and reusable.

There are no production behavior, API, or compatibility changes.

### DIFF
--- a/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
+++ b/servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
@@ -60,6 +60,7 @@ import static io.servicetalk.http.netty.HttpProtocol.HTTP_2;
 import static io.servicetalk.http.netty.TestServiceStreaming.SVC_ECHO;
 import static io.servicetalk.utils.internal.ThrowableUtils.throwException;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.anyOf;
 import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
@@ -148,7 +149,9 @@ class H2ResponseCancelTest extends AbstractNettyHttpServerTest {
 
         ExecutionException e = assertThrows(ExecutionException.class,
                 () -> makeRequest(newRequest(connection, "close")));
-        assertThat(e.getCause(), instanceOf(H2StreamResetException.class));
+        // The client may observe either the reset frame or the closed child channel, depending on which event wins.
+        assertThat(e.getCause(), anyOf(instanceOf(H2StreamResetException.class),
+                instanceOf(ClosedChannelException.class)));
 
         // Mak sure we can use the same connection for future requests:
         assertSerializedResponse(makeRequest(newRequest(connection, "ok")), HTTP_2_0, OK, "ok");


servicetalk-http-netty/src/test/java/io/servicetalk/http/netty/H2ResponseCancelTest.java
```diff
@@ -148,7 +149,9 @@ void testServerClosesStreamNotConnection() throws Exception {
 
         ExecutionException e = assertThrows(ExecutionException.class,
                 () -> makeRequest(newRequest(connection, "close")));
-        assertThat(e.getCause(), instanceOf(H2StreamResetException.class));
+        // The client may observe either the reset frame or the closed child channel, depending on which event wins.
+        assertThat(e.getCause(), anyOf(instanceOf(H2StreamResetException.class),
+                instanceOf(ClosedChannelException.class)));
```
idelpivnitskiy, line 154: Hi @arnabnandy7,
Thank you for the PR. Were you able to reproduce this behavior locally? I tried but could not cause this test to fail on macos. 

The expectation is that client should always receive `H2StreamResetException` in this case. While I agree that there could be a race, it would be interesting to understand why it's happening in case there is something inside ServiceTalk internals we can improve to make it deterministic.

If you managed to reproduce it, it would be helpful to capture wire and h2-frame logging (no user-data required) for the client as shown in this example: https://github.com/apple/servicetalk/blob/37ff504c877dc1eb2973a2eaaa09162a2f5f1695/servicetalk-examples/http/debugging/src/main/java/io/servicetalk/examples/http/debugging/DebuggingExampleClient.java#L169-L196
Those logs will help us understand what events are happening when it fails.

arnabnandy7: > Were you able to reproduce this behavior locally? I tried but could not cause this test to fail on macos.

Hi @idelpivnitskiy, thank you for the review. I have not reproduced the failure locally, the change was based on the failure reported in #3425 and the fact that `StacklessClosedChannelException` extends `ClosedChannelException`. 

I agree that relaxing the assertion could mask an internal ordering issue. I’ll investigate with wire and HTTP/2 frame logging enabled and update the PR if I can reproduce it.

PS: I don’t have access to macOS, but the failure linked in #3425 occurred on the `JDK 21 ubuntu` GitHub Actions job, so it does not appear to be macOS-specific. I’ll try to reproduce it in an Ubuntu/JDK 21 environment with wire and HTTP/2 frame logging enabled and share the event ordering if it reproduces.

arnabnandy7: Hi @idelpivnitskiy, thank you for the review.

I tested this on Ubuntu 22.04 with JDK 21, matching the environment of the original CI failure. I first enabled wire and HTTP/2 frame logging. The logs showed the server sending `RST_STREAM(CANCEL)` and the client receiving it before the child stream closed.

I then ran 1,000 sequential close/reset/reuse cycles with the original strict `H2StreamResetException` assertion. All 1,000 cycles passed, no `ClosedChannelException` was observed, and the parent HTTP/2 connection remained reusable throughout.

Based on these results, I agree that relaxing the assertion is not sufficiently justified and could mask an internal ordering issue. I have not been able to reproduce the failure reported in #3425. I suggest keeping the strict assertion and investigating the event ordering if the failure occurs again with diagnostic logging enabled.

idelpivnitskiy: Thank you, makes sense, we will keep an eye on it

Issue #2251, opened 2022-06-15
Remove `MethodHandles` from `OptimizedHttp2FrameCodecBuilder`


idelpivnitskiy:
In #2250 we started to use `MethodHandles` for `OptimizedHttp2FrameCodecBuilder` to avoid breaking users that may have older netty version in the classpath. After a few netty releases (at least 6) we should be able to remove the `MethodHandles` and use this API directly.

PR #3630, opened 2026-09-02
Remove MethodHandles from HTTP/2 codec builder

arnabnandy7:
#### Motivation

Remove the obsolete `MethodHandle` compatibility logic from `OptimizedHttp2FrameCodecBuilder`, as described in #2251. The required Netty APIs have been available since Netty 4.1.78.Final and 4.1.100.Final, while ServiceTalk now depends on Netty 4.2.17.Final.

#### Modifications

- Invoke `Http2FrameCodecBuilder#flushPreface(boolean)` directly.
- Invoke `Http2FrameCodecBuilder#decoderEnforceMaxRstFramesPerWindow(int, int)` directly.
- Remove the associated method-handle lookup, fallback, and invocation code.
- Preserve the existing preface-flushing and RST-frame protection configuration.

#### Result

`OptimizedHttp2FrameCodecBuilder` no longer uses `MethodHandle` for APIs available in the supported Netty version.

There are no API or intended behavior changes. The change removes compatibility with Netty versions older than the APIs used, which are already older than ServiceTalk's current Netty dependency.

### DIFF
--- a/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptimizedHttp2FrameCodecBuilder.java
+++ b/servicetalk-http-netty/src/main/java/io/servicetalk/http/netty/OptimizedHttp2FrameCodecBuilder.java
@@ -34,13 +34,10 @@ import io.netty.handler.codec.http2.UniformStreamByteDistributor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.lang.invoke.MethodHandle;
-import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


idelpivnitskiy:
Hi @arnabnandy7,
Thank you for the contribution! Unfortunately, we won't be able to accept this one. We saw cases when users pin older version of Netty with later versions of ServiceTalk, so we try to preserve backward compatibility regardless of what version we use. This is why we keep testing with both Netty versions: https://github.com/apple/servicetalk/blob/main/.github/workflows/ci-netty-snapshot.yml#L26

When we are ready to go to the next "major" version, which is likely to be 1.0.0 instead of 0.43.0, we will clean those up.

PR #3631, opened 2026-09-03
Add option to disable thread-interrupt-on-cancel for blocking services

daschl:
#### Motivation
`BlockingStreamingHttpService`/`BlockingHttpService` handlers always had their thread `Thread.interrupt()`ed when the client cancelled the response, even if the thread was blocked on unrelated work with no connection to the cancelled exchange. That interrupt can outlive the call it landed in and fail a second, unrelated blocking operation on the same thread later on. Handlers had no way to opt out and rely on cooperative cancellation instead.

#### Modifications
- Added `HttpServerBuilder#interruptBlockingServiceOnCancel(boolean)`, applying to both `listenBlocking` and `listenBlockingStreaming`. Defaults to `true` (unchanged behavior).
- Extended the same toggle to `HttpPredicateRouterBuilder` and `GrpcServerBuilder`, since routes registered through either are converted to a streaming service before an `HttpServerBuilder` ever sees them, so the top-level setting alone can't reach them.
- Extracted the shared cancellation-interrupt handling used by the blocking-to-streaming adapters into a small internal utility.
- Documented the new behavior and its trade-offs on the affected service and builder APIs.

#### Result
Default behavior is unchanged. Embedders can now opt a server, router, or gRPC service out of the interrupt, so a handler's unrelated blocking work runs to completion instead of being interrupted. For blocking-streaming services, a cancelled response is still observed the next time the handler writes to it, failing with a plain `IOException`; a handler blocked on something else at that moment (e.g. still reading the request) will not observe that cancellation. Blocking (aggregated) services have no such mechanism at all, so disabling the interrupt means their handler won't observe cancellation until it returns.

Covered by new tests verifying both settings' actual behavior at the unit level for the HTTP adapters and the predicate router, and end-to-end for gRPC via a client-deadline-triggered cancellation; a separate real-transport HTTP test confirms the setting doesn't disrupt normal request handling.


servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingToStreamingService.java
```diff
@@ -17,27 +17,69 @@
...
+        return request.toRequest().flatMap(req -> (interruptOnCancel ?
+                fromCallable(() -> original.handle(ctx, req, ctx.responseFactory()).toStreamingResponse()) :
+                uninterruptible(ctx, req)).shareContextOnSubscribe());
+    }
+
+    /**
+     * Same as the {@code interruptOnCancel} branch of {@link #handle}, except the handling thread is never
+     * {@link Thread#interrupt() interrupted} on cancellation.
+     */
+    private Single<StreamingHttpResponse> uninterruptible(final HttpServiceContext ctx,
```
daschl, line 55: I don't like that this is duplicated logic, maybe we can make the fromCallable or the internal call parameterized

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -67,9 +71,11 @@ public Single<StreamingHttpResponse> handle(final HttpServiceContext ctx,
         return new SubscribableSingle<StreamingHttpResponse>() {
             @Override
             protected void handleSubscribe(final Subscriber<? super StreamingHttpResponse> subscriber) {
```
daschl, line 73: this down here is the meat of it all, rest is plumbing

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -139,7 +147,14 @@ public void cancel() {
...
                 } catch (Throwable cause) {
-                    tiCancellable.setDone(cause);
+                    if (tiCancellable != null) {
+                        tiCancellable.setDone(cause);
+                    } else if (cause instanceof InterruptedException) {
+                        // Mirrors ThreadInterruptingCancellable#setDone(Throwable): clear a stale interrupt flag
+                        // before the (likely pooled) thread is reused, in case something other than cancellation
+                        // (e.g. executor shutdown) interrupted it during handling.
+                        Thread.interrupted();
+                    }
```
bryce-anderson, line 157: Would it make sense to assert that we've reset the interrupt status when we exit?
Alternatively, we could always clear the interrupted flag on exit (like a finally block that encompasses everything) since we will have swallowed any InterruptedException at that point.

servicetalk-http-api/src/main/java/io/servicetalk/http/api/BlockingStreamingToStreamingService.java
```diff
@@ -52,11 +53,14 @@
 final class BlockingStreamingToStreamingService extends AbstractServiceAdapterHolder {
     private static final HttpExecutionStrategy DEFAULT_STRATEGY = OFFLOAD_RECEIVE_META_STRATEGY;
     private final BlockingStreamingHttpService original;
+    private final boolean interruptOnCancel;
 
     BlockingStreamingToStreamingService(final BlockingStreamingHttpService original,
-                                        final HttpExecutionStrategy strategy) {
+                                        final HttpExecutionStrategy strategy,
+                                        final boolean interruptOnCancel) {
```
bryce-anderson, line 60: I feel like the fundamental problem is that we're swallowing an `InterruptedException` without clearing the threads interrupted flag. If we solve that problem, which we shoapt in the first place, but that could have downstream problems namely that the compute resource isn't actually released yet and any downstream cancellation can't propagate.

daschl:
see https://github.com/apple/servicetalk/pull/3635

PR #3632, opened 2026-09-04
http: allow opting out of interrupt-on-cancel for blocking services

daschl:
#### Motivation

When a response is cancelled, for example because the client disconnected or a timeout fired, ServiceTalk interrupts the thread running a `listenBlocking` or `listenBlockingStreaming` service. There is no way to opt out. Applications whose blocking code cannot tolerate an interrupt, or that use libraries which respond to one by closing shared resources, have no recourse.

#### Modifications

- Add `HttpContextKeys.INTERRUPT_BLOCKING_SERVICE_ON_CANCEL`, a request context key that disables the interrupt when set to `false`.
- Honor it in the blocking and blocking-streaming service adapters.
- Add `InterruptBlockingServiceOnCancelHttpServiceFilter` to `http-utils` for setting it, which also covers routers and gRPC services because they build on the same adapters.

#### Result

Blocking services can opt out of interrupt-on-cancel per request. The default is unchanged, so absent the key the thread is still interrupted.

With the interrupt disabled a blocking-streaming service still observes cancellation cooperatively, because its payload writer is terminated and the next write fails. An aggregated blocking service has no such signal and runs to completion, so the option should only be used by services that cannot block indefinitely.

No API compatibility impact; the context key and the filter are additions.

### DIFF
--- a/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcServiceInterruptOnCancelTest.java
+++ b/servicetalk-grpc-netty/src/test/java/io/servicetalk/grpc/netty/GrpcServiceInterruptOnCancelTest.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright © 2026 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


daschl:
see https://github.com/apple/servicetalk/pull/3635

PR #3635, opened 2026-09-08
http: allow opting out of interrupt-on-cancel for blocking services

daschl:
#### Motivation

When a response is cancelled — a client disconnect, a timeout, or a gRPC
deadline — ServiceTalk interrupts the thread running a `listenBlocking` or
`listenBlockingStreaming` service, and there is no way to opt out. That interrupt
can land in blocking work unrelated to the cancelled exchange, and the usual
`catch (InterruptedException e) { restore; rethrow; }` idiom relays it into every
later blocking call on that thread.

The interrupt is also redundant: a `BlockingStreamingHttpService` already learns
the peer is gone when its next `HttpPayloadWriter#write` throws `IOException`.
Only the interrupt needs to go, and the cancel must keep propagating so that
signal survives — which is why a service filter alone cannot do this.

#### Modifications

- Add `DisableInterruptOnCancelHttpServiceFilter`, a singleton service filter. It
  is the only public addition; the request-context key it sets is
  package-private.
- Honor that key in the blocking and blocking-streaming service adapters.

#### Result

Append `DisableInterruptOnCancelHttpServiceFilter.INSTANCE` ahead of a blocking
service and its thread is no longer interrupted when the response is cancelled.
This covers routers and gRPC services, which build on the same adapters.

Cancellation stays observable: a `BlockingStreamingHttpService` still sees `write`
fail. An aggregated `BlockingHttpService` has no such signal and runs to
completion, so nothing releases a thread that blocks indefinitely.

The default is unchanged, and the key is read only by the blocking adapters, so
`listenStreaming` and asynchronous services are unaffected. No API compatibility
impact.

### DIFF
--- a/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DisableInterruptOnCancelHttpServiceFilter.java
+++ b/servicetalk-http-api/src/main/java/io/servicetalk/http/api/DisableInterruptOnCancelHttpServiceFilter.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright © 2026 Apple Inc. and the ServiceTalk project authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.servicetalk.http.api;
+
+import io.servicetalk.concurrent.api.Single;
+import io.servicetalk.context.api.ContextMap.Key;
+
+import static io.servicetalk.context.api.ContextMap.Key.newKey;
+import static io.servicetalk.http.api.HttpExecutionStrategies.offloadNone;
+import static java.lang.Boolean.FALSE;
+
+/**
+ * Stops the thread running a blocking service from being {@link Thread#interrupt() interrupted} when the response is
+ * cancelled. Without this filter the thread is interrupted.
+ * <p>
+ * Applies to {@link HttpServerBuilder#listenBlocking(BlockingHttpService)} and
+ * {@link HttpServerBuilder#listenBlockingStreaming(BlockingStreamingHttpService)}, including routers and gRPC
+ * services built on the same adapters. Append it ahead of the service it governs.
+ * <p>
+ * Cancellation is still observable by a {@link BlockingStreamingHttpService}, because its payload writer is
+ * terminated and the next {@link HttpPayloadWriter#write(Object) write} throws {@link java.io.IOException}. A
+ * {@link BlockingHttpService} has no such signal and runs to completion, so nothing releases a thread that blocks
+ * indefinitely.
+ */
+public final class DisableInterruptOnCancelHttpServiceFilter implements StreamingHttpServiceFilterFactory {
+
+    /**
+     * Instance of {@link DisableInterruptOnCancelHttpServiceFilter}.
+     */
+    public static final StreamingHttpServiceFilterFactory INSTANCE = new DisableInterruptOnCancelHttpServiceFilter();
+
+    static final Key<Boolean> INTERRUPT_ON_CANCEL = newKey("INTERRUPT_ON_CANCEL", Boolean.class);
+
+    private DisableInterruptOnCancelHttpServiceFilter() {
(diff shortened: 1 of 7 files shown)


servicetalk-http-api/src/main/java/io/servicetalk/http/api/DisableInterruptOnCancelHttpServiceFilter.java
```diff
@@ -0,0 +1,80 @@
...
+ * cancelled. Without this filter the thread is interrupted.
+ * <p>
+ * Applies to {@link HttpServerBuilder#listenBlocking(BlockingHttpService)} and
+ * {@link HttpServerBuilder#listenBlockingStreaming(BlockingStreamingHttpService)}, including routers and gRPC
+ * services built on the same adapters. Append it ahead of the service it governs.
+ * <p>
+ * Cancellation is still observable by a {@link BlockingStreamingHttpService}, because its payload writer is
+ * terminated and the next {@link HttpPayloadWriter#write(Object) write} throws {@link java.io.IOException}. A
+ * {@link BlockingHttpService} has no such signal and runs to completion, so nothing releases a thread that blocks
+ * indefinitely.
```
idelpivnitskiy, line 36: Now I got why a filter that simply suppresses cancellation is not going to work...
After looking at 2 variants of the PR, I lean towards the following options:
1. Having only one behavior, either always interrupt or never interrupt. Interrupting a thread seems to be a better option as otherwise users can block forever the handle method forever. Open for never interrupt, especially if all other implementations follow that approach. I remember you checked Jetty, could you please look at how Tomcat, JBoss/WildFly, Undertow, Vert.x/Quarkus, WebSphere, etc. behave?
2. Having a system property to control it instead of API or a filter. Users usually have only one listening port. So, having a single property to control classpath should be good enough.

bryce-anderson: Michael and I chatted about this yesterday  and came to roughly the same conclusions as well. There are no major frameworks that interrupt the thread. What we landed on was a simple flag that says whether the service handler thread gets interrupted and flipping the default to 'no interrupt'. It sounds like you're 👍 for that plan as well.

idelpivnitskiy: We had a chat with Michael this morning too, and it seems like we are going in circles 😞 

One thing we discussed is that `BlockingStreamingHttpService` and `BlockingHttpService` should have consistent behavior. While `BlockingStreamingHttpService` gives users some ways to infer cancellation through read/write exceptions, `BlockingHttpService` doesn't. We again came back to the idea of the filter that will let us control it for `BlockingStreamingHttpService` only.

Then later today I went to research this area. Research confirmed that no other implementations interrupt the handler, including all servlet containers and other frameworks like Armeria, Helidon, Vert.x, etc. Most of them log the event for visibility and run WatchDog to detect unreasonably long running tasks. The main reason - it's hard to ensure interrupt flag does not leak. The recent #3637 demonstrates that and fixes some, but there are still more holes left.

With the new info, I tend to agree with the approach that you two discussed earlier: flipping the default to 'no interrupt' and adding system props to opt-in if necessary. We can even have 2 props to control `BlockingStreamingHttpService` and `BlockingHttpService` separately. For `Single.fromCallable` / `Completable.fromCallable` we can add overloads that will take a boolean, similar to Java `Future.cancel` API with its `mayInterruptIfRunning` flag.

If we ever go through servlet impl, it has `AsyncContext` API that we will need to find a way to bridge. As we currently have a parallel need to observe connection/stream state (active or not), `HttpServiceContext` might be a good way to expose both - the underlying channel state and reason, such as cancellation/timeout/etc.

@daschl wdyt? if there are more concerns left, we can get all together on a call to discuss

idelpivnitskiy: Some more fixes in #3638

idelpivnitskiy: Another interesting finding is that `java.util.concurrent.ThreadPoolExecutor` actually takes care of a leaked interrupted flag in its `runWorker` method. This is behavior starting from JDK7. `ForkJoinPool` didn't do it before, started in JDK25.

Also, Tomcat has an opt-in feature `interruptThreadThreshold` in `StuckThreadDetectionValve` and Open Liberty has `<requestTiming hungRequestThreshold="120s" interruptHungRequest="true"/>` config. So, not sure how "it's hard to ensure interrupt flag does not leak" is strong. 

Servlet spec does not clarify behavior, neither prescribes nor prohibits interruptions. Still, there are many lines of code written with assumption of industry standard behavior.

PR #3642, opened 2026-09-17
Add AGENTS.md and a testing skill for coding agents

daschl:
#### Motivation

We are starting to add agent-facing guidance to the repository so that AI coding assistants produce changes that match our conventions on the first attempt instead of after a round of review. Testing is the natural place to start: it is the largest and most convention-heavy surface in the repo, and the conventions here are not the JUnit defaults.

Nothing documented them. An agent inferring conventions from 1019 test files reliably guesses wrong in ways that break the build or the review:

- AssertJ instead of Hamcrest (538 test files use Hamcrest, 4 use AssertJ)
- `var` or `List.of` in test code, which compiles at Java 8
- `@Timeout`, when a 10s/30s global default already applies
- `@Nested` / `@DisplayName`, which appear 6 and 0 times respectively
- a missing Apache license header, which fails Checkstyle

`AGENTS.md` is used rather than a vendor-specific file so the guidance is not tied to one tool.

#### Modifications

- Added `AGENTS.md` at the repository root: module naming, build and verify commands, the constraints that break the build, and the license header for a new file. It defers anything longer than a sentence to a skill.
- Added `.agents/skills/servicetalk-testing/` with `SKILL.md` and two references, `reactive-sources.md` and `transport-and-netty.md`. A single skill matches the repository, which has no unit/integration split: one source set, one `*Test.java` suffix, no tags, one Gradle task.
- Every rule is backed by a count measured against the current tree or by a cited file, and every snippet is copied verbatim from a real test. Rules that the corpus did not support were dropped rather than softened.

#### Result

An agent has a single entry point and a short path to the conventions it needs. Documentation only: no production, test, or build changes, and `quality` is unaffected.


.agents/skills/servicetalk-testing/SKILL.md
```diff
@@ -0,0 +1,283 @@
...
+  class running beside yours, which then looks flaky for no reason.
+- **Clear a thread interrupt flag in `finally`.** Pooled threads carry it into
+  unrelated tests:
+
+```java
+} finally {
+    // A leaked interrupt would otherwise bleed into the next test on this thread.
+    Thread.interrupted();
+}
+```
```
bryce-anderson, line 140: Is this really true? This reads as "every test needs to end in a finally block".

.agents/skills/servicetalk-testing/SKILL.md
```diff
@@ -0,0 +1,283 @@
...
+
+## 5. Pick your harness
+
+| What you are testing | Harness | Read |
+|---|---|---|
+| A `Publisher`/`Single`/`Completable` operator | `TestPublisher` + `TestPublisherSubscriber` | [references/reactive-sources.md](references/reactive-sources.md), then `TimeoutPublisherTest` |
+| A codec or Netty `ChannelHandler` | `EmbeddedChannel` — no sockets | [references/transport-and-netty.md](references/transport-and-netty.md), then `HttpRequestEncoderTest` |
+| A filter, client, or server end to end | Real bound server plus a real client | [references/transport-and-netty.md](references/transport-and-netty.md), then `BlockingStreamingInputStreamTest` |
```
bryce-anderson, line 148: Some of these are feeling like they should be module level AGENTS.md content.

.agents/skills/servicetalk-testing/SKILL.md
```diff
@@ -0,0 +1,283 @@
...
+| A plain value type or utility | A flat JUnit test, no harness | `RoundRobinSelectorTest` |
+
+Reading the named exemplar is faster and more accurate than reading prose.
+Do that before writing.
+
+## 6. Waiting for asynchronous results
+
+`.toFuture().get()` with **no** timeout is the norm — 804 of 817 call sites.
```
bryce-anderson, line 156: The number of call sites will quickly go stale.

.agents/skills/servicetalk-testing/SKILL.md
```diff
@@ -0,0 +1,283 @@
...
+That is deliberate: the global test timeout fires, and
+`TimeoutTracingInfoExtension` (registered automatically for every test) dumps
+every thread's stack just before JUnit interrupts the test. Do not add timed
+overloads to match a different codebase's style.
+
+**Never sleep and then assert** on an asynchronous side effect. Poll a real
+signal instead — a `CountDownLatch`, a `BlockingQueue`, or an observer. Several
+recent commits exist only to make this change, for example
+"opentelemetry-http: await span export in filter tests instead of sleeping" and
+"Deflake processor GC tests".
```
bryce-anderson, line 166: Do the examples here help? And the `recent` will go stale eventually.

bryce-anderson (approved):
Looks like a good start!

